PR #7, opened 2017-11-06
feat(front50): adding front50 support for intents

emjburns:
- Adding support for front50
- changing get by status API to use parameter `status` instead of `statuses` for filtering by status. still is a list.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/IntentRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/IntentRepository.kt
@@ -17,11 +17,11 @@ package com.netflix.spinnaker.keel
 
 interface IntentRepository {
 
-  fun upsertIntent(intent: Intent<IntentSpec>)
+  fun upsertIntent(intent: Intent<IntentSpec>): Intent<IntentSpec>
 
   fun getIntents(): List<Intent<IntentSpec>>
 
-  fun getIntents(statuses: List<IntentStatus>): List<Intent<IntentSpec>>
+  fun getIntents(status: List<IntentStatus>): List<Intent<IntentSpec>>
 
   fun getIntent(id: String): Intent<IntentSpec>?
 }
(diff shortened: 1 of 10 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/IntentRepository.kt
```diff
@@ -17,11 +17,13 @@ package com.netflix.spinnaker.keel
 
 interface IntentRepository {
 
-  fun upsertIntent(intent: Intent<IntentSpec>)
+  fun createIntent(intent: Intent<IntentSpec>): Intent<IntentSpec>
```
robzienert: Do we need `createIntent` still if `upsertIntent` is a thing?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/controllers/v1/IntentController.kt
```diff
@@ -36,7 +37,21 @@ class IntentController
   private val traceRepository: TraceRepository
 ) {
 
-  @RequestMapping(method = arrayOf(RequestMethod.PUT))
+  private val log = LoggerFactory.getLogger(javaClass)
+
+  @RequestMapping(method = arrayOf(RequestMethod.GET))
+  fun getIntents(@QueryParam("status") status: Array<IntentStatus>? ): List<Intent<IntentSpec>> {
+    log.info("status: ${status}")
```
robzienert: Let's remove this log line.

emjburns: oops, debugging.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/IntentRepository.kt
emjburns: Good point. Removing.

PR #9, opened 2017-11-13
feat(intents): compare actual vs desired state before updating

emjburns:
* Add annotation for computed parameters
* Ignore annotated parameters when comparing state
* Start of an application converter

### DIFF
--- a/keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Application.kt
+++ b/keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Application.kt
@@ -17,18 +17,20 @@ package com.netflix.spinnaker.keel.front50.model
 
 import com.fasterxml.jackson.annotation.JsonAnyGetter
 import com.fasterxml.jackson.annotation.JsonAnySetter
+import com.netflix.spinnaker.keel.front50.annotations.Computed
 
 data class Application(
   val name: String,
   val description: String,
   val email: String,
-  val updateTs: String,
-  val createTs: String,
+  @Computed val updateTs: String? = null,
+  @Computed val createTs: String? = null,
   val platformHealthOnly: Boolean,
-  val platformHealthOnlyShowOverride: Boolean
+  val platformHealthOnlyShowOverride: Boolean,
+  val owner: String
 ) {
 
-  private val details: MutableMap<String, Any?> = mutableMapOf()
+  val details: MutableMap<String, Any?> = mutableMapOf()
 
   @JsonAnySetter
   fun set(name: String, value: Any?) {
@@ -37,4 +39,19 @@ data class Application(
 
   @JsonAnyGetter
   fun details() = details
+
+  override fun toString(): String {
+    return "Application(name='$name', description='$description', email='$email', " +
+      "updateTs='$updateTs', createTs='$createTs', platformHealthOnly=$platformHealthOnly, " +
+      "platformHealthOnlyShowOverride=$platformHealthOnlyShowOverride, " +
+      "owner=$owner, details=$details)"
+  }
+
+  /*
+   * Need to identify additional ignored properties that are in
+   * the 'details' map
+   */
+  fun computedPropertiesToIgnore(): List<String> {
+    return listOf("user","lastModifiedBy","requiredGroupMembership")
+  }
 }
--- a/keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/ApplicationIntent.kt
+++ b/keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/ApplicationIntent.kt
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.annotation.JsonTypeName
 import com.github.jonpeterson.jackson.module.versioning.JsonVersionedModel
 import com.netflix.spinnaker.keel.Intent
 import com.netflix.spinnaker.keel.IntentSpec
+import com.netflix.spinnaker.keel.front50.annotations.Computed
 
 private const val KIND = "Application"
 private const val CURRENT_SCHEMA = "1"
@@ -43,7 +44,6 @@ abstract class BaseApplicationSpec : IntentSpec {
   abstract val name: String
   abstract val description: String?
   abstract val email: String
-  abstract val lastModifiedBy: String?
   abstract val owner: String
   abstract val chaosMonkey: ChaosMonkeySpec?
   abstract val enableRestartRunningExecutions: Boolean
@@ -52,7 +52,6 @@ abstract class BaseApplicationSpec : IntentSpec {
   abstract val appGroup: String?
   abstract val cloudProviders: String?
   abstract val accounts: String?
-  abstract val user: String?
   abstract val dataSources: DataSourcesSpec
   abstract val requiredGroupMembership: List<String>
   abstract val group: String?
@@ -134,7 +133,6 @@ data class ApplicationSpec(
   override val name: String,
   override val description: String?,
   override val email: String,
-  override val lastModifiedBy: String?,
   override val owner: String,
   override val chaosMonkey: ChaosMonkeySpec?,
   override val enableRestartRunningExecutions: Boolean,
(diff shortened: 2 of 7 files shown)


keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Application.kt
```diff
@@ -17,18 +17,20 @@ package com.netflix.spinnaker.keel.front50.model
...
+import com.netflix.spinnaker.keel.front50.annotations.Computed
 
 data class Application(
   val name: String,
   val description: String,
   val email: String,
-  val updateTs: String,
-  val createTs: String,
+  @Computed val updateTs: String = "",
+  @Computed val createTs: String = "",
```
robzienert: Do you think we should just make these nullable instead?

keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Application.kt
```diff
@@ -37,4 +39,35 @@ data class Application(
 
   @JsonAnyGetter
   fun details() = details
+
+  override fun toString(): String {
+    return "Application(name='$name', description='$description', email='$email', updateTs='$updateTs', createTs='$createTs', platformHealthOnly=$platformHealthOnly, platformHealthOnlyShowOverride=$platformHealthOnlyShowOverride, owner=$owner, details=$details)"
```
robzienert: Let's newline this at col 80.

keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/ApplicationIntent.kt
```diff
@@ -143,24 +142,24 @@ data class ApplicationSpec(
...
+  @Computed override val requiredGroupMembership: List<String> = listOf(),
   override val group: String?,
   override val providerSettings: Map<String, Map<String, Any>>,
   override val trafficGuards: List<TrafficGuardSpec>,
   override val platformHealthOnlyShowOverride: Boolean,
   override val platformHealthOnly: Boolean,
   override val notifications: NotificationSpec?
-) : BaseApplicationSpec()
+) : BaseApplicationSpec() {
+  private val log = LoggerFactory.getLogger(javaClass)
```
robzienert: Not sure we need this.

keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/processors/ApplicationIntentProcessor.kt
```diff
@@ -79,8 +83,51 @@ class ApplicationIntentProcessor
...
+    val ignoredCurrentStateParameters: MutableList<String> = mutableListOf()
+    Application::class.primaryConstructor!!.parameters.filter { param ->
+        param.findAnnotation<Computed>()?.ignore == true
+      }.mapTo(ignoredCurrentStateParameters) { param ->
+        param.name.toString()
+      }
+    ignoredCurrentStateParameters.addAll(currentState.computedPropertiesToIgnore())
+
+    val ignoredDesiredStateParameters: MutableList<String> = mutableListOf()
+    val specPrimaryConstructor = NetflixApplicationSpec::class.primaryConstructor!!
```
robzienert: What if Google comes along with their own extension of `BaseApplicationSpec`? Is there some way we can abstract this behind something like... `fun <T : BaseApplicationSpec> getPrimarySpecConstructor<T>(): T`? (I think that syntax is slightly wrong).

At any rate, good step. We'll probably want to find a way to make all of this abstracted so we can use it across all of the intent processors.

keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/processors/ApplicationIntentProcessor.kt
```diff
@@ -79,8 +83,51 @@ class ApplicationIntentProcessor
...
+    ignoredDesiredStateParameters.forEach { param ->
+      desiredStateMap.remove(param)
+    }
+    desiredStateMap.put("name", desiredStateMap["name"].toString().toLowerCase())
+
+    var matching = true
+    desiredStateMap.forEach { key, value ->
+      val curVal = currentStateMap[key]
+      if (curVal != value ) {
+        log.info("$key has drifted from \"$value\" to \"$curVal\" for application ${desiredState.name}")
```
robzienert: Probably a good message to have, but would suggestion `debug` level instead.

keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/processors/converters/ApplicationConverter.kt
```diff
@@ -0,0 +1,90 @@
...
+      set("requiredGroupMembership", spec.requiredGroupMembership)
+      set("group", spec.group)
+      set("providerSettings", spec.providerSettings)
+      set("trafficGuards", spec.trafficGuards)
+      set("platformHealthOnlyShowOverride", spec.platformHealthOnlyShowOverride)
+      set("platformHealthOnly", spec.platformHealthOnly)
+      set("notifications", spec.notifications)
+    }
+
+    if (spec is NetflixApplicationSpec) {
```
robzienert: We should chat tomorrow about how to start supporting extension / cloud-provider specific specs - I had to do similar things in the security group stuff. Definitely going to become a sore spot for keel pretty quickly.

keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Application.kt
```diff
@@ -37,4 +39,35 @@ data class Application(
...
+   */
+  fun computedPropertiesToIgnore(): List<String> {
+    return listOf("user","lastModifiedBy","requiredGroupMembership")
+  }
+
+  /*
+   * Need to flatten all objects in details to top level objects to
+   * match the Application Spec
+   */
+  fun convertToMap(): MutableMap<String, Any?> {
```
robzienert: I wonder if this wouldn't make more sense being in the converter?

robzienert:
I squeezed in my PR before yours - you'll need to rebase.

keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Application.kt
emjburns: I was torn. On the one hand I like to have the logic closely tied with the class, but on the other we do have a converter.. If it makes more sense to you to have it in the converter I will move it.

keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/processors/converters/ApplicationConverter.kt
emjburns: Yup, I agree. My thought is to get a starting implementation in first, so we don't launch a million orchestrations, then refactor this when we have a better plan.

keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Application.kt
emjburns: When we give the params a default value we don't have to pass them into the constructor as 'null'. Seems nicer, for testing and creating new things? I don't particularly care though.

robzienert: I'd rather see them be `String?` signatures with default values of `null`. At least with null, we know the value isn't valid because it isn't set.

keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Application.kt
robzienert: Makes more sense to have it in the converter. I was torn on that as well, but if we're going to have a converter and that's where all other conversions are happening, seems to make sense to keep it all in one place.

keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/processors/converters/ApplicationConverter.kt
robzienert: 💯 Definitely not something blocking this PR.

keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/processors/ApplicationIntentProcessor.kt
emjburns: Ah forgot to abstract that part. Good suggestion, will address in the future, have added a great TODO. Merging for now so we stop bombing orca with tasks.

PR #12, opened 2017-11-14
feat(core): Intent-level execution window support

robzienert:
We support execution windows in Orca at the stage-level, but not at the entire orchestration/pipeline level. This adds support for execution windows around converge times, which can be helpful for ensuring deployments or infrastructure changes don't occur in peak times.

Primarily a showcase for how Attributes & Filters work.

cc @emjburns @robfletcher

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/attribute/attributes.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/attribute/attributes.kt
@@ -36,3 +36,22 @@ class PriorityAttribute(value: IntentPriority) : Attribute<IntentPriority>("Prio
 
 @JsonTypeName("Enabled")
 class EnabledAttribute(value: Boolean) : Attribute<Boolean>("Enabled", value)
+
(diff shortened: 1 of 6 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/attribute/attributes.kt
```diff
@@ -36,3 +36,22 @@ class PriorityAttribute(value: IntentPriority) : Attribute<IntentPriority>("Prio
 
 @JsonTypeName("Enabled")
 class EnabledAttribute(value: Boolean) : Attribute<Boolean>("Enabled", value)
+
+@JsonTypeName("ExecutionWindow")
+class ExecutionWindowAttribute(value: ExecutionWindow) : Attribute<ExecutionWindow>("ExecutionWindow", value) {
```
robzienert: I think I might need to rethink the `Attribute` interface a little bit. It'd be way nicer to have something more like:

```
class ExecutionWindowAttribute(days: List<Int>, timeWindows: List<TimeWindow>) ...
```

PR #14, opened 2017-11-17
feat(intent): Security group ingress rule support

robzienert:
Not all-inclusive, but a start with the most common case.

### DIFF
--- a/keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/processors/SecurityGroupIntentProcessor.kt
+++ b/keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/processors/SecurityGroupIntentProcessor.kt
@@ -60,9 +60,9 @@ class SecurityGroupIntentProcessor
       return ConvergeResult(listOf(), ConvergeReason.UNCHANGED.reason)
     }
 
-    if (missingUpstreamGroups(intent.spec)) {
-      // TODO rz - Should return _what_ security groups are missing
-      return ConvergeResult(listOf(), "Some upstream security groups are missing")
+    val missingGroups = missingUpstreamGroups(intent.spec)
+    if (missingGroups.isNotEmpty()) {
+      return ConvergeResult(listOf(), "Some upstream security groups are missing: $missingGroups")
     }
 
     return ConvergeResult(listOf(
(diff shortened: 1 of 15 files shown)


keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/processors/SecurityGroupIntentProcessor.kt
```diff
@@ -73,21 +73,20 @@ class SecurityGroupIntentProcessor
...
+      // TODO rz - It'd be rad to enumerate what actually changed in the reason if it's a delta, rather than giving
+      // a generic message.
       ConvergeReason.CHANGED.reason
     )
   }
 
   private fun getSecurityGroups(spec: SecurityGroupSpec): Set<SecurityGroup> {
-    // TODO rz - Quite dislike that any new cloudprovider would need to edit keel for declarative support... how to fix?
-    // Maybe this is the sacrifice we need to make to start statically typing all of our models?
     if (spec is AmazonSecurityGroupSpec) {
```
emjburns: Any thoughts about how we're gonna handle this for other types of security groups?

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaIntentLauncher.kt
```diff
@@ -40,25 +40,32 @@ open class OrcaIntentLauncher
...
-        .also {
-          if (it.orchestrations.isNotEmpty()) {
-            applicationEventPublisher.publishEvent(ConvergenceRequiredEvent(intent))
-          }
-        }.orchestrations.map {
-        log.info("Launching orchestration for intent (kind: ${intent.kind})")
+      val result = intentProcessor(intentProcessors, intent).converge(intent)
+
+      if (result.orchestrations.isEmpty()) {
+        log.info("Not converging state for intent (id: ${intent.id}, reason: ${result.reason})")
```
emjburns: Should be structured logs now?

keel-intents/src/main/kotlin/com/netflix/spinnaker/keel/intents/processors/SecurityGroupIntentProcessor.kt
robzienert: Each cloud provider will need to bring their own implementation at this point, just as they need to bring their own implementation into orca stages.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaIntentLauncher.kt
robzienert: been updated with the rebase. :)

emjburns:
LGTM

PR #16, opened 2017-11-17
chore(*): Converting to structured logs where applicable

robzienert:
FYI @robfletcher @emjburns -- Any logs you do from here on out, structured. `value("intent", intentId)` should be used any place that you're logging about a specific intent (should make tracing intent behavior in logs pretty clear). Prefer the intent ID over `kind`, since the ID will include the intent kind anyway.


emjburns:
Ok sounds good! LGTM

PR #22, opened 2017-11-21
feat(core): Generic computed functionality; security group diffing support

robzienert:
Builds on the work that @emjburns did for the `@Computed` annotation, moving some of the functionality into a utility `StateInspector` class and implemented for both `ApplicationIntent` and `SecurityGroupIntent`.

While it all works, I don't think it's as good as it could be. I'm running into some type erasure issues (and fun reified generics issues) that prevent me from runtime processing the computed annotations on Specs, so I created an interface that provides additional ignoreable fields. I'd really like to get rid of this, since it'll make auto-generating API docs / json schemas substantially more difficult.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/state/StateInspector.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/state/StateInspector.kt
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.keel.state
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import com.fasterxml.jackson.core.type.TypeReference
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.keel.annotation.Computed
+import org.slf4j.LoggerFactory
+import kotlin.reflect.KClass
+import kotlin.reflect.KFunction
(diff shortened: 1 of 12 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/state/StateInspector.kt
```diff
@@ -0,0 +1,107 @@
...
+  val name: String,
+  val mutator: (v: Any?) -> Any?
+)
+
+/**
+ * Performs basic state diffing to determine whether or not system state needs to be converged on the desired spec.
+ */
+class StateInspector(
+  private val objectMapper: ObjectMapper
+) {
```
robzienert: If we find value, we'll also be able to open up an API offering granular diffs of resource vs desired state.

keel-intent/src/main/kotlin/com/netflix/spinnaker/keel/intent/processor/SecurityGroupIntentProcessor.kt
```diff
@@ -103,9 +108,37 @@ class SecurityGroupIntentProcessor
...
+
+    val statePairs = mutableListOf<Pair<SecurityGroup, SecurityGroup>>()
+    desired.forEach { d ->
+      val key = "${d.accountName}/${d.region}/${d.name}"
+      currentState.forEach { c ->
+        if ("${c.accountName}/${c.region}/${c.name}" == key) {
+          statePairs.add(Pair(c, d))
+        }
+      }
+    }
```
robzienert: Determining whether or not the system needs to be updated for security groups is a little more involved since the relationship is `1 Intent -> n [system resource]`, so we have to create state pairs based on a key namespace. Something to keep in mind for other composite intents, like load balancers.

keel-intent/src/main/kotlin/com/netflix/spinnaker/keel/intent/processor/converter/SecurityGroupConverter.kt
```diff
@@ -52,8 +49,31 @@ class SecurityGroupConverter(
...
+                portRanges = it.portRanges.map { SecurityGroup.SecurityGroupRulePortRange(it.startPort, it.endPort) },
+                securityGroup = SecurityGroup.SecurityGroupRuleReference(
+                  name = it.name,
+                  accountName = it.account,
+                  region = it.region
+                )
+              )
+              else -> TODO(reason = "${it.javaClass.simpleName} has not been implemented yet")
+            }
+          },
```
robzienert: I'll be working on refactoring `keel-intent` to be cloud-provider specific soon.

PR #25, opened 2017-11-25
feat(*): NotificationPolicy support

robzienert:
Adds the ability for people to define notification policies for various events within Keel. Currently, the policy is only attached to individual Intents, but I want to be able to define Redis-backed policies so people could subscribe to an event stream that matches particular labels (e.g. "send a slack notification into `#spinnaker-events` whenever an intent with the label `spinnaker` has failed").


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/controllers/v1/IntentController.kt
```diff
@@ -73,7 +78,12 @@ class IntentController
       // TODO rz - Add updateStatus method to intentRepository
       throw NotImplementedError("soft-deleting intents is not currently supported")
     }
-    intentRepository.deleteIntent(id)
+    intentRepository.getIntent(id)
+      .takeIf { it != null }
+      ?.run {
+        applicationEventPublisher.publishEvent(AfterIntentDeleteEvent(this))
+        intentRepository.deleteIntent(id)
```
emjburns: Are these in the wrong order? Seems like the event should happen after the delete has been kicked off.

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/EventNotificationListener.kt
```diff
@@ -0,0 +1,75 @@
...
+  private val registry: Registry
+) {
+
+  private val log = LoggerFactory.getLogger(javaClass)
+
+  private val notificationsId = registry.createId("echo.notifications")
+
+  // TODO rz - Need to introduce some templating into this, I think? Stuff like {{intentId}} etc
+  @EventListener(IntentAwareEvent::class)
+  fun onIntentAwareEvent(e: IntentAwareEvent) {
```
emjburns: variable names not very specific here..

keel-scheduler/src/main/kotlin/com/netflix/spinnaker/keel/scheduler/handler/ConvergeIntentHandler.kt
```diff
@@ -60,7 +61,7 @@ class ConvergeIntentHandler
 
     if (clock.millis() > message.timeoutTtl) {
       log.warn("Intent timed out, canceling converge for {}", value("intent", message.intent.id))
-      applicationEventPublisher.publishEvent(ConvergenceTimeoutEvent(message))
+      applicationEventPublisher.publishEvent(IntentConvergeTimeoutEvent(message.intent))
```
emjburns: What's the reason that all these `.publishEvent`s happen before the action is taken instead of after?

emjburns:
LGTM

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/controllers/v1/IntentController.kt
robzienert: "after" events are supposed to become "before" the actions they relate to. It sounds wrong because it is. :(

robzienert: I'm not even sure how this happened, but at least I was consistent.

emjburns: 😂  you were very consistent

PR #26, opened 2017-11-29
feat(activity): monitor and emit metrics on orch status

emjburns:
We'd like to track, by intent kind, whether orchestrations for an intent succeed or fail. This is a first pass at that.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/memory/MemoryIntentActivityRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/memory/MemoryIntentActivityRepository.kt
@@ -17,13 +17,16 @@ package com.netflix.spinnaker.keel.memory
 
 import com.netflix.spinnaker.keel.IntentActivityRepository
 import org.slf4j.LoggerFactory
+import java.util.concurrent.ConcurrentHashMap
 import javax.annotation.PostConstruct
 
 class MemoryIntentActivityRepository : IntentActivityRepository {
 
   private val log = LoggerFactory.getLogger(javaClass)
 
-  private val orchestrations: MutableMap<String, MutableList<String>> = mutableMapOf()
+  private val currentOrchestrations: ConcurrentHashMap<String, MutableSet<String>> = ConcurrentHashMap()
+
+  private val orchestrations: ConcurrentHashMap<String, MutableSet<String>> = ConcurrentHashMap()
 
   @PostConstruct
   fun init() {
@@ -31,11 +34,16 @@ class MemoryIntentActivityRepository : IntentActivityRepository {
   }
 
   override fun addOrchestration(intentId: String, orchestrationId: String) {
+    val orchestrationUUID = parseOrchestrationId(orchestrationId)
     if (!orchestrations.containsKey(intentId)) {
-      orchestrations[intentId] = mutableListOf()
+      orchestrations[intentId] = mutableSetOf()
     }
-    if (orchestrations[intentId]?.contains(orchestrationId) == false) {
-      orchestrations[intentId]?.add(orchestrationId)
+    if (orchestrations[intentId]?.contains(orchestrationUUID) == false) {
+      orchestrations[intentId]?.add(orchestrationUUID)
+      if (!currentOrchestrations.containsKey(intentId)) {
+        currentOrchestrations.put(intentId, mutableSetOf())
+      }
+      currentOrchestrations[intentId]?.add(orchestrationUUID)
     }
   }
 
@@ -43,5 +51,29 @@ class MemoryIntentActivityRepository : IntentActivityRepository {
     orchestrations.forEach { addOrchestration(intentId, it) }
   }
 
+  override fun getCurrent(intentId: String): List<String> {
+    return currentOrchestrations.getOrDefault(intentId, listOf<String>()).toList()
+  }
+
+  override fun upsertCurrent(intentId: String, orchestrations: List<String>) {
+    if (!currentOrchestrations.containsKey(intentId)) {
+      currentOrchestrations.put(intentId, mutableSetOf())
+    }
+    currentOrchestrations[intentId]?.addAll(orchestrations.toMutableSet())
+  }
+
+  override fun upsertCurrent(intentId: String, orchestration: String) {
+    upsertCurrent(intentId, listOf(orchestration))
+  }
+
+  override fun removeCurrent(intentId: String, orchestrationId: String) {
+    currentOrchestrations[intentId]?.remove(orchestrationId)
+  }
+
+  override fun removeCurrent(intentId: String) {
+    currentOrchestrations.remove(intentId)
+  }
+
   override fun getHistory(intentId: String) = orchestrations.getOrDefault(intentId, listOf<String>()).toList()
+
 }
--- a/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaExecutionStatus.kt
+++ b/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaExecutionStatus.kt
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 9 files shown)


keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaService.kt
```diff
@@ -16,16 +16,18 @@
...
 // TODO Origin needs to be set on executions
+// origin is used for dynamic routing by orca to different clouddriver instances
 interface OrcaService {
 
   @POST("/ops")
   @Headers("Content-Type: application/context+json")
   fun orchestrate(@Body request: OrchestrationRequest): TaskRefResponse
+
+  @GET("/tasks/{id}")
+  fun getTask(@Path("id") id: String): Map<String, Any>
```
robzienert: I'd like to avoid dealing with maps as responses. You could just create a `TaskResponse` data class below like I did for `TaskRefResponse`, defining only the fields we care about:

```kotlin
data class TaskResponse(
  val status: String
  // etc...
)
```

Doing so should make interacting with the client a lot cleaner, and keeps the spread of the evil maps-of-maps pattern isolated to existing services as much as possible.

keel-scheduler/src/main/kotlin/com/netflix/spinnaker/keel/scheduler/handler/MonitorOrchestrationsHandler.kt
```diff
@@ -0,0 +1,92 @@
...
+import org.springframework.stereotype.Component
+import java.time.Duration
+
+
+/**
+ * This handler follows up on the launched orchestrations, and records
+ * success and failure by kind.
+ */
+
+@Component
```
robzienert: rm extra newline here

keel-scheduler/src/main/kotlin/com/netflix/spinnaker/keel/scheduler/handler/MonitorOrchestrationsHandler.kt
```diff
@@ -0,0 +1,92 @@
...
+
+  override fun handle(message: MonitorOrchestrations) {
+    val orchestrationIds = intentActivityRepository.getCurrent(message.intentId)
+    orchestrationIds.forEach {
+      val task = orcaService.getTask(it)
+      val orcaExecutionStatus = parseStatus(task)
+
+      when (orcaExecutionStatus) {
+        OrcaExecutionStatus.NOT_STARTED, OrcaExecutionStatus.RUNNING -> {
+          log.debug("orchestration $it for intent ${message.intentId} is RUNNING")
```
robzienert: Use structured logs where possible (and consistently). For example:

```
log.debug("orchestration for intent is running (intentId: {}, taskId: {})", value("intentId", message.intentId), value("taskId", it))
```

etc... This will make things more searchable.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaExecutionStatus.kt
```diff
@@ -0,0 +1,37 @@
...
+  REDIRECT,
+  STOPPED,
+  SKIPPED;
+}
+
+/*
+  COMPLETED = Collections.unmodifiableList(Arrays.asList(SUCCEEDED, STOPPED, SKIPPED, TERMINAL, FAILED_CONTINUE));
+  SUCCESSFUL = Collections.unmodifiableList(Arrays.asList(SUCCEEDED, STOPPED, SKIPPED));
+  FAILURE = Collections.unmodifiableList(Arrays.asList(TERMINAL, STOPPED, FAILED_CONTINUE));
+ */
```
robzienert: rm

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaService.kt
emjburns: AMAZING

keel-scheduler/src/main/kotlin/com/netflix/spinnaker/keel/scheduler/handler/MonitorOrchestrationsHandler.kt
emjburns: Yes I forgot, good reminder!

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaExecutionStatus.kt
emjburns: done

keel-scheduler/src/main/kotlin/com/netflix/spinnaker/keel/scheduler/handler/MonitorOrchestrationsHandler.kt
```diff
@@ -0,0 +1,82 @@
...
+          intentActivityRepository.removeCurrent(message.intentId, it)
+          registry.counter(orchestrationsStatusId.withTag("kind", message.kind).withTag("status", "success")).increment()
+        }
+        orcaExecutionStatus.isFailure() -> {
+          log.debug("orchestration for intent has failed (intentId: {}, taskId: {})", value("intent", message.intentId), value("task", it))
+          intentActivityRepository.removeCurrent(message.intentId, it)
+          registry.counter(orchestrationsStatusId.withTag("kind", message.kind).withTag("status", "failure")).increment()
+        }
+        else -> {
+          log.debug("orchestration $it for intent ${message.intentId} has unhandled status $orcaExecutionStatus")
```
asher: May be useful to bump registry.counter w/status=unknown in this case to help identify occurrences if it should never actually happen.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaService.kt
```diff
@@ -16,18 +16,30 @@
...
   val ref: String
 )
+
+data class TaskDetailResponse(
+  val id: String,
+  val name: String,
+  val application: String,
+  val buildTime: String,
+  val startTime: String,
+  val endTime: String,
```
asher: Since we get endTime from orca, emitting a intent.orchestrations.duration metric could be useful. Especially if it's calculated from a keel specific start time (i.e. using the score/insertion time of orchestration keys in the IntentActivity zsets) instead of the orca start time.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/memory/MemoryIntentActivityRepository.kt
```diff
@@ -23,25 +23,57 @@ class MemoryIntentActivityRepository : IntentActivityRepository {
 
   private val log = LoggerFactory.getLogger(javaClass)
 
-  private val orchestrations: MutableMap<String, MutableList<String>> = mutableMapOf()
+  private val currentOrchestrations: MutableMap<String, MutableSet<String>> = mutableMapOf()
```
cfieber: I haven't looked specifically at how IntentActivityRepository gets used but I suspect these should be ConcurrentHashMaps

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaService.kt
emjburns: I definitely think this might be useful in the future, but I'd like to add this once we want to measure this. 

Currently only the redis implementation keeps track of score, so I suspect I should do more  work on the in memory implementation to have it match the redis one more closely.

I'll add a todo for myself.

keel-scheduler/src/main/kotlin/com/netflix/spinnaker/keel/scheduler/handler/MonitorOrchestrationsHandler.kt
emjburns: Good idea. I've added this metric with an additional tag for the execution status. As these come in we can look at them and decide whether they're a success, failure, or something we need to handle differently.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/memory/MemoryIntentActivityRepository.kt
emjburns: Changed.

PR #29, opened 2017-12-06
feat(dryRun): improve dry run messaging a bit

emjburns:
Trying to improve the messaging around dry run. Have added the type of change to be made, a summary (which I want to act like a list of steps that were taken while putting together the tasks submitted to orca), and the diff. 

My hope is that intent creators will add messaging to the summary so someone who submits this intent will be able to see something like:
```
Generating job for upsertSecurityGroup,
Adding port range support,
Creating dependent security group blah
Generating job for upsertSecurityGroup blah
Adding ingress rules for security group blah
```
along with the diff to tell what the final value of the fields are going to be.

I think this needs to be massaged as we continue development. Suggestions welcome!

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/dryrun/ChangeSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/dryrun/ChangeSummary.kt
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.keel.dryrun
+
+import com.netflix.spinnaker.keel.state.FieldState
+
+data class ChangeSummary(
+  private val summary: MutableList<String> = mutableListOf()
+) {
+  var type: ChangeType = ChangeType.NO_CHANGE
+  var diff: List<FieldState> = emptyList()
+
+  fun addMessage(msg: String) {
+    summary.add(msg)
+  }
+
+  override fun toString(): String {
+    // TODO eb: more readable friendly format for logs/response?
+    var msg = "ChangeSummary(type=$type, summary=$summary"
+    if (!listOf(ChangeType.CREATE, ChangeType.NO_CHANGE, ChangeType.FAILED_PRECONDITIONS).contains(type)) msg += ", diff=$diff"
+    msg += ")"
+    return msg
+  }
+}
+
+enum class ChangeType {
+  CREATE,
+  UPDATE,
+  DELETE,
+  NO_CHANGE,
+  FAILED_PRECONDITIONS
+}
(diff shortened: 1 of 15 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/dryrun/ChangeSummary.kt
```diff
@@ -0,0 +1,47 @@
...
+  override fun toString(): String {
+    // TODO eb: more readable friendly format for logs/response?
+    var msg = "ChangeSummary(type=$type, summary=$summary"
+    if (!listOf(ChangeType.CREATE, ChangeType.NO_CHANGE).contains(type)) msg += ", diff=$diff"
+    msg += ")"
+    return msg
+  }
+
+
+}
```
robzienert: A couple things here:

1. Don't think we need this to be `open`.
2. Why all `var`? Seems to me like we could switch this to being a `data class` and have the properties immutable? If someone needs to change it, they could just perform a copy.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaIntentLauncher.kt
```diff
@@ -44,30 +49,30 @@ open class OrcaIntentLauncher
...
       if (result.orchestrations.isEmpty()) {
-        log.info("Not converging state for intent (id: ${intent.id}, reason: ${result.reason})")
+        log.info("Not converging state for intent ${intent.id}, ${result.changeSummary}")
       } else {
         applicationEventPublisher.publishEvent(ConvergenceRequiredEvent(intent))
       }
 
       val orchestrationIds = result.orchestrations.map {
-        log.info("Launching orchestration for intent (id: ${intent.id}, reason: ${result.reason})")
+        log.info("Launching orchestration for intent ${intent.id}, ${result.changeSummary}")
```
robzienert: Oops. This log (and the one above) should get `StructuredArguments`-ified.

keel-intent/src/main/kotlin/com/netflix/spinnaker/keel/intent/processor/SecurityGroupIntentProcessor.kt
```diff
@@ -69,27 +81,30 @@ class SecurityGroupIntentProcessor
...
+    if (currentStateUpToDate(intent.id, currentState, intent.spec, changeSummary)) {
+      changeSummary.addMessage(ConvergeReason.UNCHANGED.reason)
+      return ConvergeResult(listOf(), changeSummary)
     }
 
     val missingGroups = missingUpstreamGroups(intent.spec)
     if (missingGroups.isNotEmpty()) {
-      return ConvergeResult(listOf(), "Some upstream security groups are missing: $missingGroups")
+      changeSummary.addMessage("Some upstream security groups are missing: $missingGroups")
+      changeSummary.type = ChangeType.UNHANDLED
```
robzienert: Not sure `UNHANDLED` is the right thing here. Maybe something like... `FAILED_PRECONDITIONS` or something? `UNHANDLED` feels like a short-coming of the intent processor / something not being implemented, which this case is not.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/dryrun/ChangeSummary.kt
emjburns: I'll change to data class. But, the reason they're var is so this object can have defaults but be built/updated as we know what's changing. Would you prefer to create a new object each time instead of updating fields on this one?

keel-intent/src/main/kotlin/com/netflix/spinnaker/keel/intent/processor/SecurityGroupIntentProcessor.kt
emjburns: `FAILED_PRECONDITIONS` ise much better. Changing.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaIntentLauncher.kt
emjburns: Fixed

PR #30, opened 2017-12-08
feat(controller): normal upsert response format matchs dry run

emjburns:
Now the format returned by this API is the same for dry run vs no dry run: List<Map>


robzienert:
GR8!

PR #42, opened 2018-01-18
feat(logging): convergence logging now in intentActivityRepo

emjburns:
In order to make logging more manageable for each intent for each convergence, I've moved everything but the `com.netflix.spinnaker.q.QueueProcessor   : [] Received message ConvergeIntent(` messages to the intentActivityRepository. 

Hitting `v1/intents/{intentId}/convergenceLog` will give you all the messages for each intent with the format 
```{
"intentId": "Application:emilykeeltest",
"changeType": "NO_CHANGE",
"orchestrations": [],
"messages": [
  "System state matches desired state"
],
"diff": [],
"actor": "keel:scheduledConvergence",
"timeMs": 1516214128706
}
```

Each message is identified by its timestamp, which is created when the log message is created in `OrcaIntentLauncher`. This message is created right after orchestrations are launched in orca.

`v1/intents/{intentId}/convergenceLog/{timeMs}` is a permalink to the specific log message, while it exists. 

The number of messages kept is configurable. I've set it to keep logs for about 6 hours right now (a message is posted for each intent convergence every 30 seconds).

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/IntentActivityRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/IntentActivityRepository.kt
@@ -15,6 +15,9 @@
  */
 package com.netflix.spinnaker.keel
 
+import com.netflix.spinnaker.keel.dryrun.ChangeType
+import com.netflix.spinnaker.keel.state.FieldState
+
 /**
  * TODO rz - refactor IntentActivityRepository to use this instead
  *
@@ -28,6 +31,25 @@ data class IntentActivity(
   val orchestrations: List<String>,
   val reason: String,
   val actor: String
+)
+
+/**
+ * @param intentId The ID of the intent that is being evaluated.
+ * @param changeType The type of change that took place this cycle.
+ * @param orchestrations A resultant (if any) list of orchestration IDs from the intent.
+ * @param messages Human-friendly messages about the change.
+ * @param diff The diff between current and desired state.
+ * @param actor Who (or what) initiated the operation.
+ * @param timestampMillis The timestamp in millis the record was created.
+ */
+data class IntentConvergenceRecord(
+  val intentId: String,
+  val changeType: ChangeType,
+  val orchestrations: List<String>?,
+  val messages: List<String>?,
+  val diff: Set<FieldState>,
+  val actor: String,
+  val timestampMillis: Long
 )
 
 interface IntentActivityRepository {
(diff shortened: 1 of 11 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/IntentActivityRepository.kt
```diff
@@ -30,6 +33,24 @@ data class IntentActivity(
...
+ * @param actor Who (or what) initiated the operation.
+ */
+data class IntentConvergenceRecord(
+  val intentId: String,
+  val changeType: ChangeType,
+  val orchestrations: List<String>?,
+  val messages: List<String>?,
+  val diff: Set<FieldState>,
+  val actor: String,
+  val timeMs: Long
```
ajordens: If we're going full on javadoc ... do we want to include a definition of `timeMs`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/IntentActivityRepository.kt
```diff
@@ -48,6 +69,18 @@ interface IntentActivityRepository {
...
+
+  fun getConvergenceLog(intentId: String): List<IntentConvergenceRecord>
+
+  /**
+   * Permalink to a specific log message, identified by timeMs
+   * @param intentId The ID of the intent that is being evaluated.
+   * @param timeMs The timestamp of the log message,
+   *  used as the unique identifier of the message.
+   */
+  fun getConvergenceLogEntry(intentId: String, timeMs: Long): List<IntentConvergenceRecord>
```
ajordens: Slightly confusing name ... get*Entry implies a single result but a `List` is returned?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/controllers/v1/IntentController.kt
```diff
@@ -116,6 +116,13 @@ class IntentController
 
   @RequestMapping(value = "/{id}/traces", method = arrayOf(RequestMethod.GET))
   fun getIntentTrace(@PathVariable("id") id: String) = traceRepository.getForIntent(id)
+
+  @RequestMapping(value = "/{id}/convergenceLog", method = arrayOf(RequestMethod.GET))
```
ajordens: Is `convergenceLog` being overly verbose ... could we just call it `/logs`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/IntentActivityRepository.kt
emjburns: I'll change to timestampMillis also following our discussion on unit naming

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/IntentActivityRepository.kt
emjburns: There shouldn't be more than one here. Changing to return the first object if there is more than one with the same timestamp and then log a message . This shouldn't happen, and if it does it indicates that something is wrong with the queue processing.

PR #44, opened 2018-01-19
feat(intent): Scaffolding pipeline intent spec

robzienert:
I don't really expect we'll be able to get keel to strongly type pipelines this quarter. Still made some schema changes from a keel perspective, however. Namely, boolean flags were thrown into a `flags` bucket, and feature toggles were thrown into a `properties` bucket. I type-casted the few things that I think will be sticking around for quite some time, but otherwise a map seems a good fit in here.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/annotation/ForcesNew.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/annotation/ForcesNew.kt
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/annotation/ForcesNew.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.kjetland.jackson.jsonSchema.annotations.JsonSchemaInject
+
+/**
+ * Flags IntentSpec parameters that cannot change once created without
+ * recreating the intent.
+ */
+@JsonSchemaInject(json = "{ \"forcesNew\": true }")
+@Target(AnnotationTarget.VALUE_PARAMETER, AnnotationTarget.FIELD)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class ForcesNew(val value: Boolean = true)
```
emjburns: Why don't you call this `Immutable`?

robzienert: I originally had called it this, but I wanted to communicate the side effects of changing the value. It's also consistent with Terraform's taxonomy, which I thought may help some people onboard. Do you have strong feelings here?

emjburns: Oh so you're communicating that if the value changes, it will make a new object? That's fine then. I thought `Immutable` might be clearer but it sounds like it won't be.

PR #48, opened 2018-01-24
feat(front50): Adding internal rate limiter

robzienert:
While I'm working through https://github.com/spinnaker/keiko/pull/15, adding this so we don't thrash front50 as much. Nice insurance to have anyhow.

### DIFF
--- a/keel-front50/src/main/kotlin/com/netflix/spinnaker/config/Front50RateLimitProperties.kt
+++ b/keel-front50/src/main/kotlin/com/netflix/spinnaker/config/Front50RateLimitProperties.kt
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


keel-front50/src/main/kotlin/com/netflix/spinnaker/config/Front50RateLimitProperties.kt
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.config
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+
+@ConfigurationProperties("front50.rateLimit")
```
emjburns: Can you add some units to this config option in the name?

robzienert: There is a unit: `requestsPerSecond`. The annotation here is the prefix of the properties this class defines.

PR #54, opened 2018-01-26
fix(elb): Various fixes and refactoring to the ELB intents

ajordens:
Moved implementations to `keel-intent-aws` artifact and
have opted to use AWS SDK models wherever possible.

The intent process is still a WIP.


ajordens:
@robzienert -- Still not quite complete but addresses a number of things that I stumbled over when trying to get things working.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
```diff
@@ -35,8 +35,12 @@ interface CloudDriverService {
                        @Path("region") region: String,
                        @Query("vpcId") vpcId: String): SecurityGroup?
 
-  @GET("/securityGroups/{account}")
-  fun getSecurityGroups(@Path("account") account: String): Collection<SecurityGroup>
+  @GET("/securityGroups/{account}/{provider}")
```
ajordens: This endpoint doesn't actually return `SecurityGroups`!

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
```diff
@@ -50,11 +54,14 @@ interface CloudDriverService {
...
+  fun getCredential(@Path("account") account: String) : Credential
+
   @GET("/{provider}/loadBalancers/{account}/{region}/{name}")
-  fun getElasticLoadBalancer(
+  fun getLoadBalancer(
     @Path("provider") provider: String,
     @Path("account") account: String,
     @Path("region") region: String,
     @Path("name") name: String
-  ): Set<ElasticLoadBalancer>
```
ajordens: `ElasticLoadBalancer` is AWS terminology, we should use cloudprovider-agnostic terms when fetching direct from clouddriver.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
```diff
@@ -25,10 +26,25 @@ class MemoryCloudDriverCache(
...
+    }
+
+  override fun securityGroupSummaryBy(account: String, region: String, id: String): SecurityGroupSummary =
+    securityGroupSummaries.getOrNotFound(
+      "$account:$region:$id",
+      "Security group with id $id not found in the $account account and $region region"
+    ) {
+      val credential = credentialBy(account)
+      cloudDriver
+        .getSecurityGroupSummaries(account, credential.type, region)
```
ajordens: This is ridiculously inefficient ... I will fix this.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/LoadBalancer.kt
```diff
@@ -0,0 +1,20 @@
+package com.netflix.spinnaker.keel.clouddriver.model
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter
+import com.fasterxml.jackson.annotation.JsonAnySetter
+import com.fasterxml.jackson.annotation.JsonIgnore
+
+class LoadBalancer {
+  @JsonIgnore
+  private val properties = mutableMapOf<String, Any>()
```
ajordens: If you look at the "model" in `clouddriver`, it's empty.

My approach here is to `objectMapper.convert()` these to the appropriate class from the aws sdk.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/SecurityGroupSummary.kt
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.keel.clouddriver.model
+
+data class SecurityGroupSummary(
+  val name: String,
+  val id: String
+//  val moniker: Moniker
```
ajordens: this should be here but we have some garbage data with invalid monikers! (ie. monikers without `app`)

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCacheTest.kt
```diff
@@ -48,20 +44,21 @@ object MemoryCloudDriverCacheTest {
   )
 
   @Test
-  fun `security groups are looked up from CloudDriver`() {
-    whenever(cloudDriver.getSecurityGroups("prod")) doReturn securityGroups
+  fun `security group summaries are looked up from CloudDriver`() {
```
ajordens: these are somewhat lame tests against mocks but they were pre-existing so I've updated for now

keel-intent-aws/src/main/kotlin/com/netflix/spinnaker/keel/intent/aws/loadbalancer/ApplicationLoadBalancerIntent.kt
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.keel.intent.aws.loadbalancer
+
+import com.fasterxml.jackson.annotation.JsonTypeName
+import com.netflix.spinnaker.keel.intent.LoadBalancerSpec
+
+@JsonTypeName("aws.ApplicationLoadBalancer")
+data class ApplicationLoadBalancerSpec(
```
ajordens: placeholder ... same with `NetworkLoadBalancer`.

keel-intent-aws/src/main/kotlin/com/netflix/spinnaker/keel/intent/aws/loadbalancer/ApplicationLoadBalancerIntent.kt
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.keel.intent.aws.loadbalancer
+
+import com.fasterxml.jackson.annotation.JsonTypeName
+import com.netflix.spinnaker.keel.intent.LoadBalancerSpec
+
+@JsonTypeName("aws.ApplicationLoadBalancer")
```
ajordens: Not a huge fan of this particular `JsonTypeName` but I don't know if there's anything better ... jackson doesn't support deserializing from multiple properties.

keel-intent-aws/src/main/kotlin/com/netflix/spinnaker/keel/intent/aws/loadbalancer/ClassicLoadBalancerConverter.kt
```diff
@@ -0,0 +1,131 @@
...
+                .withInstanceProtocol(it.instanceProtocol.toString())
+                .withSSLCertificateId(it.sslCertificateId)
+            )
+        }
+      )
+  }
+
+  override fun convertFromState(state: LoadBalancerDescription): ClassicLoadBalancerSpec =
+    state.run {
+      val vpc = cloudDriver.networkBy(vpcId!!)
```
ajordens: I don't think this network logic is totally correct (this existed previous) so I'll need to revisit and ensure we do the right things with vpc / subnet names.

keel-intent-aws/src/main/kotlin/com/netflix/spinnaker/keel/intent/aws/loadbalancer/ClassicLoadBalancerIntent.kt
```diff
@@ -0,0 +1,129 @@
...
+data class ClassicListener(
+  val protocol: Protocol,
+  val loadBalancerPort: Int,
+  val instanceProtocol: Protocol,
+  val instancePort: Int,
+  val sslCertificateId: String? = null
+)
+
+/*
+"certificates": [],
```
ajordens: Garbage .. will remove

keel-intent-aws/src/main/kotlin/com/netflix/spinnaker/keel/intent/aws/loadbalancer/ClassicLoadBalancerIntentProcessor.kt
```diff
@@ -0,0 +1,105 @@
...
+  override fun supports(intent: Intent<IntentSpec>) = intent.spec is ClassicLoadBalancerSpec
+
+  override fun converge(intent: LoadBalancerIntent): ConvergeResult {
+    val changeSummary = ChangeSummary(intent.id())
+    val spec = intent.spec as ClassicLoadBalancerSpec
+
+    log.info("Converging state for {}", value("intent", intent.id()))
+
+    getLoadBalancer(spec)?.let {
+      val differences = Maps.difference(
```
ajordens: Not quite sure how best to detect differences ... I don't really want to do explicit checks against each particular key/value.

keel-intent-aws/src/main/kotlin/com/netflix/spinnaker/keel/intent/aws/loadbalancer/ApplicationLoadBalancerIntent.kt
robzienert: I would like to explore simplifying this, because I definitely don't like the idea of repeating this information. I'd like to explore a custom `TypeIdResolver` implementation that we use in combination with custom serializer/deserializer. It's not supported via vanilla Jackson, but a combination of the three looks promising.

keel-intent-aws/src/main/kotlin/com/netflix/spinnaker/keel/intent/aws/loadbalancer/ClassicLoadBalancerIntentProcessor.kt
robzienert: I wrote `StateInspector` to build up diffs; I didn't know about this Guava class. StateInspector groups diffs by root-level element, although I want to have more granular control.

From an end-user's perspective, if I'm working with a loadbalancer's healthcheck, I'd just like to see the diff of the healthcheck if something as changed, not an individual property. However, if I'm in some other part, I might just want to see the individual property change. It seems contextual to me... in an ideal world I'd like to have some utility to allow developers to decide what units are more/less granular.

The fact we're generating diffs is the important bit at this point. I'd like to see what we and early adopters find most useful before getting fancy, though.

keel-intent-aws/src/main/kotlin/com/netflix/spinnaker/keel/intent/aws/loadbalancer/ApplicationLoadBalancerIntent.kt
robzienert: Agreed. I'd like to POC some janky code to get a feel of what it'd take. I think what you did here was good enough for now.

PR #85, opened 2018-06-12
chore(build): use Gradle 4.8 + new dependency locking mechanism

robfletcher:


keel-scheduler/keel-scheduler.gradle
```diff
@@ -1,7 +1,9 @@
 dependencies {
   compile project(":keel-core")
   compile project(":keel-orca")
-  compile "com.netflix.spinnaker.keiko:keiko-redis-spring:2.+"
+  compile("com.netflix.spinnaker.keiko:keiko-redis-spring:2.+") {
+    exclude module: "kork-core"
+  }
```
robfletcher: This is the only weird thing I ran into. Without this exclusion `keel-scheduler` was pulling in an older version of `kork-core` via `keiko-spring` instead of the newer one from `keel-core`.

robzienert: Interesting. I was under the impression that Keiko had no dependencies on other spinnaker libs... in that case, I'm totally going to bring `kork-core` into `keiko-spring` so I can delete the duplicated object mapper configurer.

robfletcher: It's already in there. Just a quite old version:

```> gradle :keiko-spring:dependencyInsight --configuration compile --dependency kork-core

Parallel execution is an incubating feature.
Inferred project: keiko, version: 2.2.0-SNAPSHOT
Publication nebula not found in project :.
com.netflix.spinnaker.kork:kork-core:1.110.0
   variant "default"
\--- compile

A web-based, searchable dependency report is available by adding the --scan option.

BUILD SUCCESSFUL in 4s
1 actionable task: 1 executed
```

robfletcher: I don't know why Gradle was prioritizing that over the newer version.

PR #104, opened 2018-08-22
Adding asset containers & partials

robzienert:
Will allow for stuff like security group rules to be defined separately from the root security group.

### DIFF
--- a/keel-api/src/main/proto/Asset.proto
+++ b/keel-api/src/main/proto/Asset.proto
@@ -14,6 +14,18 @@ message Asset {
   google.protobuf.Any spec = 4;
 }
 
+message PartialAsset {
+  AssetId id = 1;
+  AssetId root = 2;
+  TypeMetadata typeMetadata = 3;
+  google.protobuf.Any spec = 4;
+}
+
+message AssetContainer {
+  Asset asset = 1;
+  repeated PartialAsset partialAssets = 2;
+}
+
 message AssetId {
   string value = 1;
 }
--- a/keel-api/src/main/proto/AssetPlugin.proto
+++ b/keel-api/src/main/proto/AssetPlugin.proto
@@ -9,15 +9,16 @@ import "google/protobuf/empty.proto";
 import "Asset.proto";
 
 service AssetPlugin {
-  rpc current (Asset) returns (CurrentResponse);
+  rpc current (AssetContainer) returns (CurrentResponse);
 
-  rpc plan (Asset) returns (PlanResponse);
+  rpc plan (AssetContainer) returns (PlanResponse);
 
-  rpc converge (Asset) returns (ConvergeResponse);
+  rpc converge (AssetContainer) returns (ConvergeResponse);
 }
 
 message CurrentResponse {
-  Asset asset = 1;
+  Asset desired = 1;
+  Asset current = 2;
 }
 
 message PlanResponse {
--- a/keel-api/src/test/kotlin/com/netflix/spinnaker/keel/api/TestAssetPlugin.kt
+++ b/keel-api/src/test/kotlin/com/netflix/spinnaker/keel/api/TestAssetPlugin.kt
@@ -5,11 +5,14 @@ import com.netflix.spinnaker.keel.api.plugin.CurrentResponse
 import io.grpc.stub.StreamObserver
 
 class TestAssetPlugin : AssetPluginGrpc.AssetPluginImplBase() {
-  override fun current(request: Asset, responseObserver: StreamObserver<CurrentResponse>) {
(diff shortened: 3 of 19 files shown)


keel-api/src/main/proto/Asset.proto
```diff
@@ -14,6 +14,18 @@ message Asset {
   google.protobuf.Any spec = 4;
 }
 
+message PartialAsset {
+  AssetId id = 1;
```
robfletcher: why does this have its own id?

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/api/TestAssetPlugin.kt
```diff
@@ -5,11 +5,11 @@ import com.netflix.spinnaker.keel.api.plugin.CurrentResponse
 import io.grpc.stub.StreamObserver
 
 class TestAssetPlugin : AssetPluginGrpc.AssetPluginImplBase() {
-  override fun current(request: Asset, responseObserver: StreamObserver<CurrentResponse>) {
+  override fun current(request: AssetContainer, responseObserver: StreamObserver<CurrentResponse>) {
```
robfletcher: why does it need the partials in order to get the current state?

keel-api/src/main/proto/AssetPlugin.proto
```diff
@@ -9,15 +9,15 @@ import "google/protobuf/empty.proto";
...
-  rpc plan (Asset) returns (PlanResponse);
+  rpc plan (AssetContainer) returns (PlanResponse);
 
-  rpc converge (Asset) returns (ConvergeResponse);
+  rpc converge (AssetContainer) returns (ConvergeResponse);
 }
 
 message CurrentResponse {
-  Asset asset = 1;
+  AssetContainer assetContainer = 1;
```
robfletcher: surely the plugin is never going to respond with any partials

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/api/TestAssetPlugin.kt
robfletcher: Feels like we should really just pass the `AssetId`.

keel-api/src/main/proto/Asset.proto
robzienert: It needs to be addressable. Partial assets will be saved / deleted independently from assets.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/api/TestAssetPlugin.kt
robzienert: Oh, here's why: Plugins won't know what to do with an `AssetId`. In the case of the AwsAssetPlugin, it's built entirely around the ability to inspect the request to decide what to do with it.

robfletcher: That's fair.

keel-api/src/main/proto/Asset.proto
robfletcher: Ok, I guess that makes sense.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/api/TestAssetPlugin.kt
robfletcher: It doesn't need the partials, though. Right?

keel-rule-engine/src/main/kotlin/com/netflix/spinnaker/keel/persistence/AssetRepository.kt
```diff
@@ -19,12 +22,23 @@ interface AssetRepository {
   fun get(id: AssetId): Asset?
 
   /**
-   * Persists an asset.
+   * Fetches the ids of any assets that depend (directly) on [id].
+   * Get a partial asset
    */
-  fun store(asset: Asset)
+  fun getPartial(id: AssetId): PartialAsset?
```
robfletcher: Does it make sense to have a distinct id type?

robzienert: You tell me. I don't think it matters, but I'm happy to add it in.

robfletcher: I guess let's keep it the same for now & consider it separately from this PR.

robzienert:
This is now perfect.

PR #111, opened 2018-08-29
feat(aws): SecurityGroup and SecurityGroupRuleAsset support

robzienert:


keel-aws/src/main/kotlin/com/netflix/spinnaker/keel/aws/AmazonAssetPlugin.kt
```diff
@@ -61,31 +75,11 @@ class AmazonAssetPlugin(
...
-                "accountName" to spec.accountName
-              )
-            )),
-            OrchestrationTrigger("1")
-          ))
+        securityGroupHandler().let {
+          val flattenedAsset = it.flattenAssetContainer(request)
+          val spec: SecurityGroup = flattenedAsset.spec.unpack()
+          it.converge(flattenedAsset.id.value, spec)
+        }
```
robfletcher: Doesn't look like this ever sends a response.

PR #159, opened 2018-09-27
Prevent multiple converge requests for same asset

robfletcher:
See #158


robfletcher:
Don't merge. Re-thinking

PR #162, opened 2018-09-28
Package rules in proto response from current endpoint

robfletcher:


keel-ec2/src/main/kotlin/com/netflix/spinnaker/keel/ec2/asset/AmazonSecurityGroupHandler.kt
```diff
@@ -124,18 +130,67 @@ class AmazonSecurityGroupHandler(
...
-            it.region = region
-            it.vpcName = vpcId?.let { cloudDriverCache.networkBy(it).name }
-            it.description = description
+          .also { builder ->
+            builder.name = name
+            builder.accountName = accountName
+            builder.region = region
+            builder.vpcName = vpcId?.let { cloudDriverCache.networkBy(it).name }
+            builder.description = description
+            inboundRules.forEach { rule ->
```
robfletcher: This whole bit with the inbound rules was missing so the current and desired states never match.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/SecurityGroup.kt
```diff
@@ -41,7 +41,8 @@ data class SecurityGroup(
   data class SecurityGroupRuleReference(
     val name: String,
     val accountName: String?,
-    val region: String?
+    val region: String?,
+    val vpcId: String?
```
robfletcher: Need to resolve the VPC to map to the proto response

keel-ec2/src/main/kotlin/com/netflix/spinnaker/keel/ec2/asset/AmazonSecurityGroupHandler.kt
```diff
@@ -124,18 +130,67 @@ class AmazonSecurityGroupHandler(
...
+              builder.inboundRuleList.add(
+                SecurityGroupRule.newBuilder().also { builder ->
+                  val referencedGroup = rule.securityGroup
+                  val cidrRange = rule.range
+                  when {
+                    referencedGroup != null && (referencedGroup.vpcId == vpcId && referencedGroup.name == name) -> {
```
robfletcher: I believe this is the correct condition for a self-referencing rule

keel-ec2/src/main/kotlin/com/netflix/spinnaker/keel/ec2/asset/AmazonSecurityGroupHandler.kt
```diff
@@ -124,18 +130,67 @@ class AmazonSecurityGroupHandler(
...
+                      builder.selfReferencingRuleBuilder.also { ruleBuilder ->
+                        ruleBuilder.protocol = rule.protocol
+                        rule.portRanges?.forEach {
+                          ruleBuilder.portRangeList.add(it.toProto())
+                        }
+                      }
+                    }
+                    referencedGroup != null && (referencedGroup.accountName != accountName || referencedGroup.region != region) -> {
```
robfletcher: If account or region differ from the parent security group, it's a cross-region rule.

keel-ec2/src/main/kotlin/com/netflix/spinnaker/keel/ec2/asset/AmazonSecurityGroupHandler.kt
robzienert: 🤦‍♂️ oops

keel-ec2/src/main/kotlin/com/netflix/spinnaker/keel/ec2/asset/AmazonSecurityGroupHandler.kt
robzienert: This looks sane, yes.

robfletcher: Specifically no need to check account / region as the vpcId would be different if those were.

PR #169, opened 2018-10-18
feat(ec2): Change cardinality of port ranges on security group rules

robfletcher:
Previously a rule could have multiple port ranges but this didn't really reflect what was happening under the hood on EC2 and it made it hard to diff current and desired models. For example, a security group would appear to have a diff because River shows it as having a single rule with 2 port ranges where the desired state is two rules with a port range each. Now we normalize the data coming out of River and require the desired model to have a single port range per rule so there's no ambiguity.


keel-ec2/src/main/kotlin/com/netflix/spinnaker/keel/ec2/asset/AmazonSecurityGroupHandler.kt
```diff
@@ -99,9 +99,9 @@ class AmazonSecurityGroupHandler(
             "vpcId" to spec.vpcName,
             "description" to spec.description,
             "securityGroupIngress" to portRangeRuleToJob(spec),
-            "ipIngress" to spec.inboundRuleList.filter { it.hasCidrRule() }.flatMap {
-              convertCidrRuleToJob(it)
-            },
+            "ipIngress" to spec.inboundRuleList
+              .filter { it.hasCidrRule() }
+              .flatMap { convertCidrRuleToJob(it) },
```
robfletcher: Hopefully this makes sense. This is the part where it's creating the actual change via River.

keel-ec2/src/main/kotlin/com/netflix/spinnaker/keel/ec2/asset/SecurityGroupRuleComparator.kt
```diff
@@ -114,66 +114,7 @@ private fun SecurityGroup.canonicalize(): SecurityGroup {
...
-}
-
-private fun CidrRule.canonicalize(): CidrRule {
-  val originalPortRangeList = portRangeList
-  return toBuilder()
-    .apply {
-      clearPortRange()
-      originalPortRangeList.sortedWith(PortRangeComparator).forEach {
-        addPortRange(it)
+        addInboundRule(originalRule)
```
robfletcher: Canonicalization is much easier now as I just have to order the rules not the rules _and_ the port ranges inside them.

keel-ec2/src/main/kotlin/com/netflix/spinnaker/keel/ec2/asset/EC2TypeConverter.kt
```diff
@@ -33,7 +33,11 @@ class EC2TypeConverter(
         builder.region = securityGroup.region
         builder.vpcName = securityGroup.vpcId?.let { cloudDriverCache.networkBy(it).name }
         builder.description = securityGroup.description
-        securityGroup.inboundRules.forEach { rule ->
+        securityGroup.inboundRules
+          .flatMap { rule ->
+            rule.portRanges?.map { rule to it } ?: listOf(rule to null)
+          }
```
robfletcher: Here's where I map across the multiple port ranges River may have in its model of the rule and turn them into multiple distinct rules

PR #173, opened 2018-10-19
Redis asset store implementation

robfletcher:


keel-redis/src/main/kotlin/com/netflix/spinnaker/keel/redis/RedisAssetRepository.kt
```diff
@@ -0,0 +1,207 @@
...
+    }
+  }
+
+  companion object {
+    private const val INDEX_SET = "keel:assets"
+    private const val ASSET_HASH = "keel:asset:%s"
+    private const val DEPENDENCIES_SET = "keel:asset:%s:dependencies"
+    private const val DEPENDS_ON_SET = "keel:asset:%s:dependsOn"
+    private const val PARTIALS_SET = "keel:asset:%s:partials"
+    private const val STATE_SORTED_SET = "keel:asset:%s:state"
```
robzienert: Let's get these hashed' `{keel:asset:%s}` plz, so we can keep dynomite as a viable option.

robfletcher: Ultimately I'd just like to do this with SQL but I wanted a quick impl.

robzienert: Yeah, that's fine but it shouldn't prevent us from having a good enough implementation for OSS. The last thing we want is a Front50 Redis situation.

robfletcher: Totally agree. This will probably live for a while.

PR #191, opened 2019-01-22
Deserialization bug fix and dependency updates

robfletcher:


keel-api/src/main/resources/banner.txt
```diff
@@ -0,0 +1,9 @@
+ ██ ▄█▀▓█████ ▓█████  ██▓
```
emjburns: <3

PR #195, opened 2019-02-01
Dependency handling without lockfiles

robfletcher:


gradle.properties
```diff
@@ -1,5 +1,8 @@
 kapt.use.worker.api=true
-
-bootVersion=1.5.14.RELEASE
+bootVersion=1.5.19.RELEASE
```
robfletcher: Have to manage this here because `spinnaker.version("springBoot")` can't be used on a Gradle plugin.

keel-orca/keel-orca.gradle
```diff
@@ -19,11 +19,11 @@ apply plugin: "java-library"
...
-  implementation "com.fasterxml.jackson.core:jackson-annotations:[2.9.6,)"
+  implementation "com.fasterxml.jackson.core:jackson-databind"
+  implementation "com.fasterxml.jackson.core:jackson-annotations"
   implementation "org.jetbrains.kotlin:kotlin-reflect"
 
-  api "com.fasterxml.jackson.module:jackson-module-kotlin:[2.9.6,)"
-  api "com.github.jonpeterson:jackson-module-model-versioning:+"
-  api "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:[2.9.6,)"
+  api "com.fasterxml.jackson.module:jackson-module-kotlin"
+  api "com.github.jonpeterson:jackson-module-model-versioning:latest.release"
```
cfieber: I'd be hesitant about putting `latest.release` on anything without some sort of locking in place because it makes the builds unstable and also noticably slower, but especially something considered part of the module api

build.gradle
```diff
@@ -61,30 +72,15 @@ subprojects {
   
   dependencies {
     implementation platform("org.springframework.boot:spring-boot-dependencies:$bootVersion")
+    testImplementation platform("org.junit:junit-bom:${spinnaker.version('jupiter')}")
```
robfletcher: This means we can omit the version on all the JUnit dependencies as it gets them from this BOM.

keel-orca/keel-orca.gradle
robfletcher: Yeah, I'll do that. I was being lazy.

robfletcher: Specifically with `retrofit2-kotlin-coroutines-adapter` that won't be necessary for long as there's an approved PR in Retrofit to get that capability merged in core.

PR #196, opened 2019-02-06
chore(terminology): Asset -> Resource

robfletcher:


robfletcher:
As agreed in NYC meeting

PR #198, opened 2019-02-11
Cluster annealing episode 1

robfletcher:


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterName.kt
```diff
@@ -0,0 +1,18 @@
+package com.netflix.spinnaker.keel.api.ec2
+
+/**
+ * Frigga style name.
+ */
+data class ClusterName(
```
emjburns: What's the relationship between this and Moniker?

robfletcher: Pretty much the same thing. I'm still thinking this out. I'm not sure if it's better for the user to specify the cluster name using separate values like this or for us to parse it out of a single name string.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -12,63 +13,137 @@ import com.netflix.spinnaker.keel.api.ec2.SecurityGroup
...
+              "loadBalancers" to spec.loadBalancerNames,
+              "targetGroups" to spec.targetGroups,
+              "account" to spec.accountName
+            )
+          )),
+          OrchestrationTrigger(resourceName.toString())
+        ))
+        .await()
+    }
+    log.info("Started task {} to upsert security group", taskRef.ref)
```
cfieber: `s/security group/cluster/`

PR #200, opened 2019-02-15
Removed k8s as a core dependency

robfletcher:


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/bus/ResourceEventListener.kt
```diff
@@ -0,0 +1,30 @@
...
+
+  @EventListener
+  fun handle(event: ResourceEvent) {
+    log.info("Received event {}", event)
+    when (event.type) {
+      CREATE,
+      UPDATE ->
+        resourceRepository.store(event.resource)
+      DELETE ->
+        resourceRepository.delete(event.resource.metadata.name)
```
robfletcher: This basically replaces everything we were using k8s for

keel-k8s/src/main/kotlin/com/netflix/spinnaker/keel/k8s/KubernetesResourceEventListener.kt
```diff
@@ -1,75 +0,0 @@
...
-import com.netflix.spinnaker.keel.events.ResourceEventType.UPDATE
-import io.kubernetes.client.ApiException
-import io.kubernetes.client.apis.CustomObjectsApi
-import io.kubernetes.client.models.V1DeleteOptions
-import org.slf4j.LoggerFactory
-import org.springframework.context.event.EventListener
-import org.springframework.stereotype.Component
-
-@Component
-class KubernetesResourceEventListener(
```
robfletcher: Previous listener impl that pushed resource changes to k8s API

PR #202, opened 2019-02-16
Resize ASG if only diff is capacity

robfletcher:
I built this on top of #201 so you can ignore the first commit for now. It will get rebased out.

I'd like to do some tweaking of the use of the diff library as the method I'm using is designed for comparing new & old versions of the same object. In our case that's not quite valid as either one could be newer (either desired state changed, or actual resource in the cloud changed out of band).

This is a proof-of-concept though so I haven't really tried to be too clean here. I like that the check for what changed is strongly typed rather than just grepping strings or something.

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -24,11 +24,13 @@ import com.netflix.spinnaker.keel.model.Job
 import com.netflix.spinnaker.keel.model.OrchestrationRequest
 import com.netflix.spinnaker.keel.model.OrchestrationTrigger
 import com.netflix.spinnaker.keel.orca.OrcaService
+import com.netflix.spinnaker.keel.plugin.ResourceConflict
 import com.netflix.spinnaker.keel.plugin.ResourceHandler
 import com.netflix.spinnaker.keel.retrofit.isNotFound
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
+import org.javers.core.diff.Diff
 import org.slf4j.LoggerFactory
 import retrofit2.HttpException
 import java.time.Clock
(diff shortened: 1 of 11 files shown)


robfletcher:
I built this on top of #201 so you can ignore the first commit for now. It will get rebased out.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -124,19 +73,122 @@ class ClusterHandler(
...
-  override fun delete(resourceName: ResourceName, spec: SecurityGroup) {
+  /**
+   * @return `true` if the only changes in the diff are to capacity.
+   */
+  private fun Diff?.isCapacityOnly(): Boolean =
+    this != null && changes.all {
+      it.affectedObject.run {
+        isPresent && get() is Capacity
+      }
+    }
```
robfletcher: I like that this check is strongly typed (because the objects being diffed were).

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandlerTests.kt
```diff
@@ -190,21 +194,52 @@ internal object ClusterHandlerTest : JUnit5Minutests {
...
 
-          expectThat(firstValue.job.first()["source"])
-            .isA<Map<String, Any>>()
-            .and {
-              hasEntry("account", activeServerGroupResponse.accountName)
-              hasEntry("region", activeServerGroupResponse.region)
-              hasEntry("asgName", activeServerGroupResponse.asg.autoScalingGroupName)
+        val diff = javers.compare(resource.spec.withDoubleCapacity().withDifferentInstanceType(), resource.spec)
+
+        test("annealing clones the current server group") {
```
robfletcher: Tests that if there are more changes than just capacity it does a "createServerGroup"

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandlerTests.kt
```diff
@@ -190,21 +194,52 @@ internal object ClusterHandlerTest : JUnit5Minutests {
...
-      test("annealing a diff clones the current server group") {
-        whenever(orcaService.orchestrate(any())) doReturn CompletableDeferred(TaskRefResponse("/tasks/1"))
+      context("the diff is only in capacity") {
 
-        converge(request.metadata.name, spec)
+        val diff = javers.compare(resource.spec.withDoubleCapacity(), resource.spec)
 
-        argumentCaptor<OrchestrationRequest>().apply {
-          verify(orcaService).orchestrate(capture())
+        test("annealing resizes the current server group") {
```
robfletcher: Tests that if the only thing that changed was capacity we do a "resizeServerGroup"

robfletcher:
I'd like to do some tweaking of the use of the diff library as the method I'm using is designed for comparing new & old versions of the same object. In our case that's not quite valid as either one could be newer (either desired state changed, or actual resource in the cloud changed out of band).

This is a proof-of-concept though so I haven't really tried to be too clean here. I like that the check for what changed is strongly typed rather than just grepping strings or something.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -42,80 +44,27 @@ class ClusterHandler(
...
-            "region" to spec.location.region,
-            "asgName" to ancestorServerGroup
-          )
-          job["copySourceCustomBlockDeviceMappings"] = true
-        }
+      val spec = resource.spec
+      val job = when {
+        diff.isCapacityOnly() -> spec.resizeServerGroupJob()
+        else -> spec.createServerGroupJob()
+      }
```
robfletcher: Determines here which type of job to submit to Orca

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -124,19 +73,124 @@ class ClusterHandler(
...
     }
     log.info("Started task {} to upsert cluster", taskRef.ref)
   }
 
-  override fun delete(resourceName: ResourceName, spec: SecurityGroup) {
+  /**
+   * @return `true` if the only changes in the diff are to capacity.
+   */
+  private fun Diff?.isCapacityOnly(): Boolean =
+    this != null && affectedObjects.all { it is Capacity }
```
robfletcher: Strongly-typed check for where the changes in the diff are

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
emjburns: That is cool!

robfletcher: Right?

PR #204, opened 2019-02-26
fix(travis): update script

emjburns:
Copying a working script from echo to see if it will build master


robfletcher:
I copied this & added the missing thing it needed (telling `compileTestKotlin` to target Java 1.8)

emjburns:
Wonderful thank you.

PR #205, opened 2019-02-26
Support annealing security group ingress rules

robfletcher:


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -84,14 +84,18 @@ class ClusterHandler(
   /**
    * @return `true` if the only changes in the diff are to capacity.
    */
-  private fun Diff?.isCapacityOnly(): Boolean =
-    this != null && affectedObjects.all { it is Capacity }
+  private fun DiffNode?.isCapacityOnly(): Boolean =
+    this != null && affectedRootPropertyTypes.all { it == Capacity::class.java }
```
robfletcher: Different handling here because I swapped out the diff lib

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -69,12 +71,14 @@ class SecurityGroupHandler(
                 "cloudProvider" to CLOUD_PROVIDER,
                 "name" to spec.name,
                 "regions" to listOf(spec.region),
-                "vpcId" to spec.vpcName,
+                "vpcId" to cloudDriverCache.networkBy(spec.vpcName, spec.accountName, spec.region).id,
```
robfletcher: Oops, I was sending VPC name not id.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -132,7 +136,34 @@ class SecurityGroupHandler(
...
+              response.inboundRules.flatMap { rule ->
+                val ingressGroup = rule.securityGroup
+                val ingressRange = rule.range
+                when {
+                  ingressGroup != null -> rule.portRanges?.map { portRange ->
+                    ReferenceSecurityGroupRule(
+                      Protocol.valueOf(rule.protocol!!.toUpperCase()),
+                      if (ingressGroup.name != response.name || ingressGroup.accountName != response.accountName) ingressGroup.name else null,
+                      if (ingressGroup.accountName != response.accountName) ingressGroup.accountName else null,
+                      if (ingressGroup.vpcId != response.vpcId) cloudDriverCache.networkBy(ingressGroup.vpcId!!).name else null,
```
robfletcher: This is kinda nasty. Basically if the desired state is a self referencing or non-cross-account rule you can omit these fields. But if we include them in the converted response from River we'll get a false delta and keep no-op upserting the security group.

I think long term I'd like to define new sub-classes of `SecurityGroupRule` for the self-reference and in-account cases.

keel-plugin/keel-plugin.gradle
```diff
@@ -4,8 +4,8 @@ apply from: "$rootDir/gradle/junit5.gradle"
 
 dependencies {
   api project(":keel-core")
-  api "org.javers:javers-core:$javersVersion"
-
+  api "de.danielbechler:java-object-diff:0.95"
```
robfletcher: Different diff lib that can handle `Set` properties

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt
```diff
@@ -58,7 +57,7 @@ interface ResourceHandler<T : Any> : KeelPlugin {
    * Implement this method and [create] if you need to handle create and update in different ways.
    * Otherwise just implement [upsert].
    */
-  fun update(resource: Resource<T>, diff: Diff = DiffBuilder.empty()) {
+  fun update(resource: Resource<T>, diff: DiffNode = DiffNode.newRootNode()) {
```
robfletcher: Would like to get rid of these defaults but it makes simple unit tests a pain.

PR #215, opened 2019-02-28
SQL resource repository

### LINKED ISSUE #211: Use SQL rather than Redis

robfletcher:
Fixes #211


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
```diff
@@ -0,0 +1,47 @@
...
+    .registerULIDModule()
+    .enable(WRITE_DATES_AS_TIMESTAMPS)
+    .enable(WRITE_DATES_WITH_ZONE_ID)
+    .enable(WRITE_DATE_KEYS_AS_TIMESTAMPS)
+    .disable(FAIL_ON_UNKNOWN_PROPERTIES)
+    .apply {
+      dateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZZ").apply {
+        timeZone = TimeZone.getDefault()
+      }
+    }
```
robfletcher: I created a standard configured `ObjectMapper` as we were replicating this all over the place.

keel-redis/src/main/kotlin/com/netflix/spinnaker/keel/redis/RedisResourceRepository.kt
```diff
@@ -58,30 +59,30 @@ class RedisResourceRepository(
     }
 
   override fun store(resource: Resource<*>) {
-    redisClient.withCommandsClient<Long?> { redis: JedisCommands ->
+    redisClient.withCommandsClient<Long> { redis: JedisCommands ->
```
emjburns: Is it even worth keeping redis around if we think sql is a better fit for the data? Or, are we going to maintain and test both to fit with our story of "you can still run on redis we just don't"

robfletcher: That was my assumption but I was asking myself the same question, yeah.

emjburns: Hmm. Let's keep redis for now, and support both ways. I think it makes sense to have an easy redis story as well as a recommended sql story.

emjburns:
LGTM

keel-redis/src/main/kotlin/com/netflix/spinnaker/keel/redis/RedisResourceRepository.kt
robfletcher: Yeah, I think I'd be tempted to abandon it if it was time consuming to maintain, but it's not so far.

robzienert: We want to keep redis as an option for now.

keel-sql/src/main/resources/db/changelog/20190227-initial-schema.yml
```diff
@@ -0,0 +1,92 @@
...
+              value: " engine innodb"
+      rollback:
+        - dropTable:
+            tableName: resources
+  - changeSet:
+      id: create-resource-state-table
+      author: fletch
+      changes:
+        - createTable:
+            tableName: resource_state
```
asher: Is this table going to remain update only? If you plan on deleting rows from here in the future, doing so via a `ulid` based primary key column is preferable. (Deleting rows other than via a unique key can effectively block inserts)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -0,0 +1,212 @@
...
+      )
+      .from(RESOURCE)
+      .where(field("name").eq(name.value))
+      .fetch()
+      .intoResultSet()
+      .run {
+        if (next()) {
+          toResource(specType)
+        } else {
+          throw NoSuchResourceException(name)
```
robzienert: Interested to know why you prefer throwing exceptions for not found instead of a nullable type?

keel-sql/src/main/resources/db/changelog/20190227-initial-schema.yml
robfletcher: I’d like to update it to use the uid. Was just doing it this way as the repository interface was already there.

robfletcher: But yeah, it’s an append only table

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
robfletcher: In this case it’s a truly exceptional condition. There’s no reason this method should ever get called with an invalid name.

PR #217, opened 2019-03-01
handle an unparseable request body

robfletcher:


robfletcher:
Ran into this when I threw my old request body at your updated controller. We get a better HTTP status / message now.

PR #221, opened 2019-03-01
endpoint to list supported resource kinds

robfletcher:


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/KindController.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.netflix.spinnaker.keel.plugin.ResourceHandler
+import com.netflix.spinnaker.keel.yaml.APPLICATION_YAML_VALUE
+import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping(path = ["/kinds"])
+class KindController(val plugins: List<ResourceHandler<*>>) {
```
emjburns: Do you like to have args on the same line if it's only one?

robfletcher: I wrap them if the line goes over whatever length

PR #222, opened 2019-03-01
Handle default security groups

### LINKED ISSUE #213: Removing security group name from a cluster desired state does not delete it
Should red/black the cluster resulting in a new server group without the security group attached.

robfletcher:
We should _always_ assign `nf-infrastructure` and `nf-datacenter`, and--if there are no other security group names--an application specific security group.

Closes #213


emjburns:
I feel like this should be in keel-nflx, because it's a netflix specific opinion

robfletcher:
> I feel like this should be in keel-nflx, because it's a netflix specific opinion

I don't disagree. I think it might be an idea to have a `ResourceValidator` interface that is separate from the `ResourceHandler`. There could be 0..n validators for each type of resource and we'd just run them all. Each one could reject and/or mutate the resource before it gets persisted.

robfletcher:
Raised #225 to address ^

PR #224, opened 2019-03-02
Async diff/actuation via eventing

### LINKED ISSUE #218: Event based diff/merge loop
Raise an event to go check resource state vs desired. That way we can trigger the same event on resource upsert for instant action.

robfletcher:
Closes #218 

See #226 for next steps with this


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/Main.kt
```diff
@@ -52,6 +53,7 @@ private val DEFAULT_PROPS = mapOf(
   ]
 )
 @Import(PlatformComponents::class)
+@EnableAsync
```
robfletcher: This turns on the async capabilities for the whole app.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceActuationScheduler.kt
```diff
@@ -0,0 +1,26 @@
...
+  private val resourceRepository: ResourceRepository,
+  private val publisher: ApplicationEventPublisher
+) {
+
+  @Scheduled(fixedDelayString = "\${keel.resource.monitoring.frequency.ms:60000}")
+  fun validateManagedResources() {
+    log.debug("Starting scheduled validation…")
+    resourceRepository.allResources { (name, apiVersion, kind) ->
+      log.debug("Requesting validation of {}", name)
+      publisher.publishEvent(ResourceCheckEvent(name, apiVersion, kind))
```
robfletcher: This is now just firing an event to request a resource be checked.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceActuator.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.netflix.spinnaker.keel.plugin.supporting
+import de.danielbechler.diff.ObjectDifferBuilder
+import de.danielbechler.diff.node.DiffNode
+import org.slf4j.LoggerFactory
+import org.springframework.context.event.EventListener
+import org.springframework.scheduling.annotation.Async
+import org.springframework.stereotype.Component
+
+@Component
+class ResourceActuator(
```
robfletcher: This is the diff/actuate logic that used to be in `ResourceStateMonitor` but now it just handles one resource at a time via an event.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceActuator.kt
```diff
@@ -0,0 +1,84 @@
...
+  private val resourceRepository: ResourceRepository,
+  private val handlers: List<ResourceHandler<*>>
+) {
+
+  private val differ = ObjectDifferBuilder.buildDefault()
+
+  @Async
```
robfletcher: This is pretty neat because it means that even though the event dispatch is synchronous, this handler runs async. In other words it won't block that loop in `ResourceActuationScheduler`. This work will get run on a thread pool.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourcePersister.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.netflix.spinnaker.keel.events.ResourceEventType.UPDATE
+import com.netflix.spinnaker.keel.persistence.ResourceRepository
+import com.netflix.spinnaker.keel.plugin.ResourceHandler
+import com.netflix.spinnaker.keel.plugin.supporting
+import org.slf4j.LoggerFactory
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.stereotype.Component
+
+@Component
+class ResourcePersister(
```
robfletcher: I just moved this class from `keel-api` to `keel-plugin` so it sits with the others involved in this workflow. Seemed more logical there.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt
```diff
@@ -83,6 +83,17 @@ interface ResourceHandler<T : Any> : KeelPlugin {
...
 
 /**
- * Represents some kind of conflict that prevents the plugin from interacting with the resource.
+ * Searches a list of `ResourceHandler`s and returns the first that supports [apiVersion] and
+ * [kind].
+ *
+ * @throws UnsupportedKind if no appropriate handlers are found in the list.
  */
-class ResourceConflict(message: String) : RuntimeException(message)
+internal fun List<ResourceHandler<*>>.supporting(
```
robfletcher: We were doing this in a couple of different places so I made an extension function here that can just get the handler for an `apiVersion`/`kind` from a list of handlers.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt
```diff
@@ -83,6 +83,17 @@ interface ResourceHandler<T : Any> : KeelPlugin {
...
+  apiVersion: ApiVersion,
+  kind: String
+): ResourceHandler<*> =
+  find { it.apiVersion == apiVersion && it.supportedKind.first.singular == kind }
+    ?: throw UnsupportedKind(apiVersion, kind)
```
robfletcher: Previously we were returning `null` and then the caller threw this exception. I figured it made more sense to just do it here so the caller can assume a non-null return.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/spring/ApplicationEventResourceCheckListener.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.netflix.spinnaker.keel.annealing.ResourceActuator
+import org.slf4j.LoggerFactory
+import org.springframework.context.event.EventListener
+import org.springframework.scheduling.annotation.Async
+
+open class ApplicationEventResourceCheckListener(
+  private val resourceActuator: ResourceActuator
+) {
+  @Async
+  @EventListener(ResourceCheckEvent::class)
```
robfletcher: This is pretty neat because it means that even though the event dispatch is synchronous, this handler runs async. In other words it won't block that loop in `ResourceCheckScheduler`. Instead this work will get run on a thread pool.

keel-plugin/src/test/kotlin/com/netflix/spinnaker/keel/annealing/spring/SpringBasedResourceCheckQueueTests.kt
```diff
@@ -0,0 +1,39 @@
...
+class SpringBasedResourceCheckQueueTests {
+
+  @Autowired
+  lateinit var resourceCheckQueue: ResourceCheckQueue
+
+  @MockBean
+  lateinit var resourceActuator: ResourceActuator
+
+  @Test
+  fun `can enqueue a resource check via Spring`() {
```
robfletcher: Tests that the spring wiring works and resource checks get propagated correctly.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/config/AnnealingConfiguration.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
+@Configuration
+class AnnealingConfiguration {
+
+  @Bean
+  @ConditionalOnMissingBean(ResourceCheckQueue::class)
```
robfletcher: Only register the Spring-based `ResourceCheckQueue` if there isn't another one registered. So we'll fall back to this locally / until we get an SQS/whatever backed implementation.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/config/AnnealingConfiguration.kt
```diff
@@ -0,0 +1,25 @@
...
+  fun resourceCheckQueue(publisher: ApplicationEventPublisher) =
+    ApplicationEventResourceCheckQueue(publisher)
+
+  @Bean
+  @ConditionalOnBean(ApplicationEventResourceCheckQueue::class)
```
robfletcher: Only register the listener if the `ResourceCheckQueue` is the Spring one. I tried to put the listener method in the queue class but Spring doesn't like it because the event handler method has to be on the interface.

PR #227, opened 2019-03-02
Base database indices on uid instead of name

robfletcher:


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
```diff
@@ -62,13 +60,7 @@ class ResourceController(
     // TODO: we need to take the resource type as well so we can actually parse and validate here
     // if you're creating a resource you don't need to pass the name
     log.info("Creating: $submittedResource")
-    val resource = Resource(
-      metadata = ResourceMetadata(name = ResourceName("undefined")),
-      kind = submittedResource.kind,
-      apiVersion = submittedResource.apiVersion,
-      spec = submittedResource.spec
-    )
```
robfletcher: No longer creating a resource with dummy fields here.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceMetadata.kt
```diff
@@ -23,17 +23,17 @@ import de.huxhorn.sulky.ulid.ULID
 data class ResourceMetadata(
   val name: ResourceName,
   @JsonProperty(defaultValue = "0") val resourceVersion: Long = 0,
-  val uid: ULID.Value? = null,
+  val uid: ULID.Value,
```
robfletcher: `uid` no longer nullable.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -16,8 +16,13 @@
...
 
 import com.netflix.spinnaker.keel.api.Resource
+import com.netflix.spinnaker.keel.api.SubmittedResource
+import de.huxhorn.sulky.ulid.ULID
 
-data class ResourceEvent(
-  val type: ResourceEventType,
-  val resource: Resource<*>
-)
+sealed class ResourceEvent
```
robfletcher: Replaced a single event class with a `type` field with 3 different event sub-types as I want each one to have a different payload.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -44,20 +58,9 @@ class ClusterHandler(
...
-    return Resource(
-      apiVersion = resource.apiVersion,
-      kind = resource.kind,
-      metadata = metadata,
-      spec = cluster
-    )
-  }
+  override fun generateName(spec: Cluster) = ResourceName(
+    "ec2:cluster:${spec.location.accountName}:${spec.location.region}:${spec.moniker.cluster}"
+  )
```
robfletcher: Now each handler just needs to implement a `generateName` and optionally a `validate`. I know we may be moving this logic to a separate type but it still makes sense to simplify it.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt
```diff
@@ -29,10 +34,51 @@ interface ResourceHandler<T : Any> : KeelPlugin {
...
+  /**
+   * Validates the resource spec, and generates a metadata header, and applies any defaults /
+   * normalization.
+   *
+   * Implementors should not generally need to override this method.
+   *
+   * @return a hydrated `Resource` with a name generated by convention, a `uid`, default values
+   * applied, etc.
+   */
+  fun validate(resource: SubmittedResource<*>): Resource<T> {
```
robfletcher: The `ResourceHandler` interface can do the default stuff - constructing metadata & parsing the spec. Don't need to re-implement it in every handler.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt
```diff
@@ -29,10 +34,51 @@ interface ResourceHandler<T : Any> : KeelPlugin {
...
+   * Validates the resource and applies any defaults / normalization.
+   *
+   * @return [resource] or a copy of [resource] that may have been changed, for example in order to
+   * set default values.
    */
-  fun validate(resource: Resource<*>, generateName: Boolean): Resource<T>
+  @Suppress("UNCHECKED_CAST")
+  fun validate(resource: Resource<Any>): Resource<T> = resource.copy(
+    spec = objectMapper.convertValue(resource.spec, supportedKind.second)
+  ) as Resource<T>
```
robfletcher: This method can be overridden if you want custom validation, normalization, etc.

keel-sql/src/main/resources/db/changelog/20190227-initial-schema.yml
```diff
@@ -53,8 +53,8 @@ databaseChangeLog:
             tableName: resource_state
             columns:
               - column:
-                  name: name
-                  type: varchar(255)
+                  name: uid
+                  type: char(26)
```
robfletcher: Asher was of the opinion this was a better schema and I think it's more logical to use the primary key of the `resource` table.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
emjburns: Makes sense!

PR #239, opened 2019-03-06
New kork-web and workarounds for REST tests

robfletcher:


keel-spring-test-support/src/main/kotlin/com/netflix/spinnaker/keel/redis/spring/DisabledSecurityConfiguration.kt
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Disables security for all endpoints for use in [org.springframework.test.web.servlet.MockMvc]
+ * type tests.
+ *
+ * You're supposed to be able to use [org.springframework.security.test.context.support.WithMockUser]
+ * but I couldn't get it to work with our security setup.
+ */
+@Configuration
+@Order(0)
+class SecurityDisabledConfiguration : WebSecurityConfigurerAdapter() {
```
robfletcher: This is the test config class we could add to kork. At the same time, it feels like there's a right way to do this.

PR #242, opened 2019-03-07
New coroutines support in Retrofit core lib

robfletcher:
The new coroutine support in Retrofit differs a bit from the call adapter we've been using. Instead of returning `Deferred<*>` the interface methods are just defined as `suspend`.


robfletcher:
Updated to handle the new `Front50Service` and associated stuff.

PR #246, opened 2019-03-08
2-phase create for security groups with self-ref ingress

### LINKED ISSUE #235: Cannot include a self-ref ingress rule on a new security group
Actuation fails because the referenced security group does not exist during the operation that is creating it.

robfletcher:
You can't include a self-referential ingress rule when creating a new security group as the referenced group must exist already. Instead I'm now filtering them out on create which means a delta will be detected later and the ingress added.

Fixes #235

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
@@ -58,39 +58,38 @@ class SecurityGroupHandler(
   override fun current(resource: Resource<SecurityGroup>): SecurityGroup? =
     cloudDriverService.getSecurityGroup(resource.spec)
 
-  override fun upsert(resource: Resource<SecurityGroup>, diff: DiffNode?) {
+  override fun create(resource: Resource<SecurityGroup>) {
     val taskRef = runBlocking {
       resource.spec.let { spec ->
         orcaService
           .orchestrate(OrchestrationRequest(
-            "Upsert security group ${spec.name} in ${spec.accountName}/${spec.region}",
+            "Create security group ${spec.name} in ${spec.accountName}/${spec.region}",
             spec.application,
-            "Upsert security group ${spec.name} in ${spec.accountName}/${spec.region}",
-            listOf(Job(
-              "upsertSecurityGroup",
-              mapOf(
-                "application" to spec.application,
-                "credentials" to spec.accountName,
-                "cloudProvider" to CLOUD_PROVIDER,
-                "name" to spec.name,
-                "regions" to listOf(spec.region),
-                "vpcId" to cloudDriverCache.networkBy(spec.vpcName, spec.accountName, spec.region).id,
-                "description" to spec.description,
-                "securityGroupIngress" to spec.inboundRules.mapNotNull {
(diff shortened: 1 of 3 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -58,39 +58,38 @@ class SecurityGroupHandler(
   override fun current(resource: Resource<SecurityGroup>): SecurityGroup? =
     cloudDriverService.getSecurityGroup(resource.spec)
 
-  override fun upsert(resource: Resource<SecurityGroup>, diff: DiffNode?) {
+  override fun create(resource: Resource<SecurityGroup>) {
```
robfletcher: Instead of implementing `upsert` I now implement separate `create` and `update` on the handler.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -190,6 +178,68 @@ class SecurityGroupHandler(
...
+        "cloudProvider" to CLOUD_PROVIDER,
+        "name" to name,
+        "regions" to listOf(region),
+        "vpcId" to cloudDriverCache.networkBy(vpcName, accountName, region).id,
+        "description" to description,
+        "securityGroupIngress" to inboundRules
+          // we have to do a 2-phase create for self-referencing ingress rules as the referenced
+          // security group must exist prior to the rule being applied. We filter then out here and
+          // the subsequent diff will apply the additional group(s).
+          .filterNot { it is SelfReferenceRule }
```
robfletcher: Filter out self-referencing rules when submitting the task to create a new security group to Orca

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt
```diff
@@ -210,112 +213,203 @@ internal object SecurityGroupHandlerTests : JUnit5Minutests {
...
-      before {
-        orcaService.stub {
-          on {
-            orchestrate(any())
-          } doAnswer {
-            CompletableDeferred(TaskRefResponse("/tasks/${randomUUID()}"))
+    sequenceOf(
+      "create" to SecurityGroupHandler::create,
+      "update" to SecurityGroupHandler::update.partially3(DiffNode.newRootNode())
+    )
```
robfletcher: I want to run all the tests contained in this loop for both `create` and `update` without copy 'n pasting all the code

robfletcher:
The diff view is not doing a wonderful job with the test here. Basically I moved the 3 existing tests for upserting a security group into a loop so it runs them against both the `create` and `update` methods on the handler (they should behave the same). I then added two new tests outside that loop for the differential behavior between `create` and `update` when there's a self-referential ingress rule on the desired model.

robfletcher:
Yeah, I’m not 100% happy with everything about that minutest library, but it seems to work better than Spek and be a bit less verbose than stock JUnit.

That `SecurityGroupHandlerTests` in particular could use breaking up.

PR #247, opened 2019-03-08
fix(csrf): ensure csrf is disabled for API

cfieber:


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/Main.kt
```diff
@@ -88,6 +92,13 @@ class KeelApplication {
   @ConditionalOnMissingBean(ResourceHandler::class)
   fun noResourcePlugins(): List<ResourceHandler<*>> = emptyList()
 
+  @Bean
+  fun csrfDisable() = @Order(Ordered.HIGHEST_PRECEDENCE) object: WebSecurityConfigurerAdapter() {
```
cfieber: needed an explicit value for `@Order` and it also seemed to matter that it was high precidence (I picked `12345` initially and that I guess didn't happen soon enough to properly disable csrf)

PR #248, opened 2019-03-08
SQS backed diff/actuate

robfletcher:


gradle.properties
```diff
@@ -4,6 +4,7 @@ coroutinesVersion=1.1.1
 hikariVersion=2.5.1
 kotlinVersion=1.3.21
 liquibaseVersion=3.6.3
+localstackVersion=1.10.6
```
robfletcher: Travis is not having this so I'll try to find an alternative.

keel-sqs/src/main/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckQueue.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.amazonaws.services.sqs.AmazonSQS
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.config.SqsProperties
+import com.netflix.spinnaker.keel.annealing.ResourceCheckQueue
+import com.netflix.spinnaker.keel.api.ApiVersion
+import com.netflix.spinnaker.keel.api.ResourceName
+import javax.annotation.PostConstruct
+
+internal class SqsResourceCheckQueue(
+  private val sqs: AmazonSQS,
```
robfletcher: I'm not actually creating this bean anywhere. Trying to figure that out.

keel-sqs/src/main/kotlin/com/netflix/spinnaker/config/SqsProperties.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.config
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+
+@ConfigurationProperties("sqs")
+class SqsProperties {
+  var queueARN: String = ""
```
emjburns: Do we also need to specify the topic ARN?

emjburns: Or, are you just creating a queue and pushing messages and reading from the queue?

keel-sqs/src/test/kotlin/com/netflix/spinnaker/keel/sqs/LocalStackSqsConfiguration.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import com.amazonaws.services.sqs.AmazonSQS
+import com.amazonaws.services.sqs.AmazonSQSClient
+import com.netflix.spinnaker.keel.serialization.configuredObjectMapper
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.testcontainers.containers.localstack.LocalStackContainer
+
+@Configuration
+class LocalStackSqsConfiguration {
```
robfletcher: Just split this out into its own file

keel-sqs/src/test/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckListenerTests.kt
```diff
@@ -0,0 +1,127 @@
...
+import com.nhaarman.mockitokotlin2.doThrow
+import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.never
+import com.nhaarman.mockitokotlin2.reset
+import com.nhaarman.mockitokotlin2.stub
+import com.nhaarman.mockitokotlin2.verify
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+
+internal class SqsResourceCheckListenerTests : JUnit5Minutests {
```
robfletcher: Faster running unit tests that don't deal with any concurrency / coroutine management and just do a single poll on the SQS queue.

keel-sqs/src/main/kotlin/com/netflix/spinnaker/config/SqsConfiguration.kt
```diff
@@ -53,5 +53,7 @@ class SqsConfiguration {
     objectMapper: ObjectMapper,
     actuator: ResourceActuator
   ): Any =
-    SqsResourceCheckListener(sqsClient, queueARN, objectMapper, actuator)
+    sqsClient.createQueue(queueARN.name).run {
```
robfletcher: Feel like there might be a better place to do the queue creation than here but probably ok for now.

keel-sqs/src/test/kotlin/com/netflix/spinnaker/keel/sqs/SqsIntegrationTests.kt
```diff
@@ -0,0 +1,60 @@
...
+@SpringBootTest(
+  classes = [LocalStackSqsConfiguration::class, SqsConfiguration::class],
+  properties = [
+    "sqs.enabled=true",
+    "sqs.account-name=test",
+    "sqs.queue-arn=arn:aws:sqs:ap-south-1:530822479253:keel-test-resource_check"
+  ],
+  webEnvironment = NONE
+)
+internal class SqsIntegrationTests {
```
robfletcher: Integration test that uses localstack to stand up an SQS API and make sure the queue read / write work in congress.

Thoughts on whether this has value? It's pretty slow due to localstack pulling down a fucking giant Docker image.

emjburns: I don't feel like it has a ton of value. What cases would this docker image version of SQS catch that we wouldn't catch another way? Seems like increased build time is a win regardless.

PR #249, opened 2019-03-09
Experimental stab at trying to get keel-api working with Spring webflux

robfletcher:
Everything _seems_ to work. ¯\_(ツ)_/¯

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/KindController.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/KindController.kt
@@ -2,22 +2,38 @@ package com.netflix.spinnaker.keel.rest
 
 import com.netflix.spinnaker.keel.plugin.ResourceHandler
 import com.netflix.spinnaker.keel.yaml.APPLICATION_YAML_VALUE
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers.IO
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.reactor.flux
 import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
+import reactor.core.publisher.Flux
 
 @RestController
 @RequestMapping(path = ["/kinds"])
 class KindController(val plugins: List<ResourceHandler<*>>) {
+
+  @ExperimentalCoroutinesApi
   @GetMapping(produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE])
-  fun get(): List<Map<String, Any>> =
-    plugins
-      .groupBy { it.apiVersion }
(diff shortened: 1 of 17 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/KindController.kt
```diff
@@ -2,22 +2,35 @@ package com.netflix.spinnaker.keel.rest
...
+          )
+        }
+        .forEach {
+          send(it)
+        }
+      close()
+    }
+
+  private val scope: CoroutineScope
+    get() = CoroutineScope(IO)
```
robfletcher: Really no idea if this is the right coroutine dispatcher to use.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/KindController.kt
```diff
@@ -2,22 +2,38 @@ package com.netflix.spinnaker.keel.rest
...
+        .map { (apiVersion, plugin) ->
+          mapOf(
+            "api-version" to apiVersion,
+            "kinds" to plugin.map { it.supportedKind.first.singular }
+          )
+        }
+        .forEach {
+          send(it)
+        }
+      close()
```
robfletcher: This is kinda bullshit because the content I'm `forEach`ing here is in no way a reactive stream of any kind.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
```diff
@@ -51,46 +55,53 @@ class ResourceController(
 ) {
 
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
+  private val scope: CoroutineScope
+    get() = CoroutineScope(IO)
```
robfletcher: All the endpoints in this controller _are_ doing database IO so this seems like the right dispatcher.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -41,51 +41,51 @@ interface ResourceRepository {
   /**
    * Invokes [callback] once with each registered resource.
    */
-  fun allResources(callback: (ResourceHeader) -> Unit)
+  suspend fun allResources(callback: (ResourceHeader) -> Unit)
```
robfletcher: I really like the fact that you need to be explicit about blocking when using `ResourceRepository` now. Forces you to think about handling IO.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/KindController.kt
cfieber: What are the other scope options? Agree this doesn't seem to be IO

robfletcher: Here are the options: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/index.html

PR #251, opened 2019-03-12
fix(dependencies): transitive dependency cleanup

cfieber:
We were picking up our spring-context/spring-boot-autoconfigure via keel-retrofit exporting
kork-web as an api level dependency. Moved some of keel-retrofit's dependencies to
implementation scope and added explicit dependencies where needed.

Switched use of Guava CacheBuilder to Caffeine CacheBuilder in ClouddriverInMemoryCache.

### DIFF
--- a/keel-clouddriver/keel-clouddriver.gradle
+++ b/keel-clouddriver/keel-clouddriver.gradle
@@ -7,8 +7,11 @@ dependencies {
   implementation "com.fasterxml.jackson.core:jackson-databind"
   implementation "com.fasterxml.jackson.core:jackson-annotations"
   implementation "com.fasterxml.jackson.module:jackson-module-kotlin"
-  implementation "com.github.jonpeterson:jackson-module-model-versioning:1.2.2"
+  implementation "com.github.ben-manes.caffeine:caffeine"
   implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310"
+  implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion"
+  implementation "org.springframework:spring-context"
+  implementation "org.springframework.boot:spring-boot-autoconfigure"
 
   testImplementation project(":keel-retrofit-test-support")
   testImplementation "com.squareup.retrofit2:retrofit-mock:${spinnaker.version('retrofit2')}"
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/config/ClouddriverConfiguration.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/config/ClouddriverConfiguration.kt
@@ -32,13 +32,11 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.context.ApplicationContext
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
-import org.springframework.context.annotation.Import
 import retrofit2.Retrofit
 import retrofit2.converter.jackson.JacksonConverterFactory
 
 @Configuration
 @ConditionalOnProperty("clouddriver.enabled")
-@Import(RetrofitConfiguration::class)
 class ClouddriverConfiguration {
 
   @Bean
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
@@ -15,8 +15,8 @@
  */
 package com.netflix.spinnaker.keel.clouddriver
 
-import com.google.common.cache.Cache
-import com.google.common.cache.CacheBuilder
+import com.github.benmanes.caffeine.cache.Cache
+import com.github.benmanes.caffeine.cache.Caffeine
 import com.netflix.spinnaker.keel.clouddriver.model.Credential
 import com.netflix.spinnaker.keel.clouddriver.model.Network
 import com.netflix.spinnaker.keel.clouddriver.model.SecurityGroupSummary
@@ -30,27 +30,27 @@ class MemoryCloudDriverCache(
   private val cloudDriver: CloudDriverService
 ) : CloudDriverCache {
 
-  private val securityGroupSummariesById = CacheBuilder.newBuilder()
+  private val securityGroupSummariesById = Caffeine.newBuilder()
     .maximumSize(1000)
     .expireAfterWrite(30, SECONDS)
     .build<String, SecurityGroupSummary>()
 
-  private val networks = CacheBuilder.newBuilder()
+  private val networks = Caffeine.newBuilder()
     .maximumSize(1000)
     .expireAfterWrite(30, SECONDS)
     .build<String, Network>()
 
-  private val availabilityZones = CacheBuilder.newBuilder()
+  private val availabilityZones = Caffeine.newBuilder()
     .maximumSize(1000)
     .expireAfterWrite(30, SECONDS)
     .build<String, Set<String>>()
 
-  private val credentials = CacheBuilder.newBuilder()
+  private val credentials = Caffeine.newBuilder()
     .maximumSize(100)
     .expireAfterWrite(1, HOURS)
     .build<String, Credential>()
 
-  private val subnets = CacheBuilder.newBuilder()
+  private val subnets = Caffeine.newBuilder()
     .maximumSize(1000)
     .expireAfterWrite(30, SECONDS)
     .build<String, Subnet>()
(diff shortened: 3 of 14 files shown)


keel-clouddriver/keel-clouddriver.gradle
```diff
@@ -7,8 +7,13 @@ dependencies {
   implementation "com.fasterxml.jackson.core:jackson-databind"
   implementation "com.fasterxml.jackson.core:jackson-annotations"
   implementation "com.fasterxml.jackson.module:jackson-module-kotlin"
+  implementation "com.github.ben-manes.caffeine:caffeine"
   implementation "com.github.jonpeterson:jackson-module-model-versioning:1.2.2"
   implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310"
+  implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion"
+  implementation "org.slf4j:slf4j-api"
```
cfieber: slf4j needs to be explicitly declared now in a bunch of spots...

we could theoretically add this as an implementation dependency in the `subprojects` block in build.gradle (though that only right now has the bom imports so this would potentially add to the dependency graph if something didn't actually do any logging...)

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/config/ClouddriverConfiguration.kt
```diff
@@ -32,13 +32,11 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.context.ApplicationContext
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
-import org.springframework.context.annotation.Import
 import retrofit2.Retrofit
 import retrofit2.converter.jackson.JacksonConverterFactory
 
 @Configuration
 @ConditionalOnProperty("clouddriver.enabled")
-@Import(RetrofitConfiguration::class)
```
cfieber: removed this where it was referenced - all it does is configure a RetrofitLogLevel bean which we don't actually use anywhere

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
```diff
@@ -117,7 +117,7 @@ class MemoryCloudDriverCache(
           .map { it.availabilityZone }
           .toSet()
       }
-    }
+    }!!
```
cfieber: Caffeine is actually annotated with @Nullable on its interface - I went with `!!` instead of `?: emptySet()` because the loading function should always return a Set and it seems like an assertion type failure if it doesn't rather than silently running and returning emptySet

keel-deliveryconfig-plugin/keel-deliveryconfig-plugin.gradle
```diff
@@ -7,6 +7,10 @@ dependencies {
   implementation project(":keel-plugin")
   implementation project(":keel-front50")
 
+  implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion"
```
cfieber: much boiler

very plate

Not sure there is a better option really though without reinventing `spinnaker.group('kotlinySpringyCommonDependencies')`

emjburns: Are the kotlin spring dependencies significantly different than the normal ones? I'd be delighted with a group if there are more than a couple.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
emjburns: That feels right.

keel-clouddriver/keel-clouddriver.gradle
robfletcher: I think even if something doesn't do any logging you generally want the API available when making changes to code.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/config/ClouddriverConfiguration.kt
robfletcher: D'oh. Cruft from before I switched over to Retrofit 2. Good spot

keel-deliveryconfig-plugin/keel-deliveryconfig-plugin.gradle
robfletcher: I tend to never use the groups because I'm fussy about what I import (as you can tell from the fact you're fixing my shitty dependency tangle in this PR) and I have to go refer to spinnaker/spinnaker-dependencies in order to understand what they do. If we want to group them we could do an `apply(from: "$rootDir/gradle/kotling-spring-common.gradle")` or something.

keel-front50/keel-front50.gradle
```diff
@@ -7,8 +7,11 @@ dependencies {
   implementation "com.fasterxml.jackson.core:jackson-databind"
   implementation "com.fasterxml.jackson.core:jackson-annotations"
   implementation "com.fasterxml.jackson.module:jackson-module-kotlin"
-  implementation "com.github.jonpeterson:jackson-module-model-versioning:1.2.2"
   implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310"
+  implementation "com.github.jonpeterson:jackson-module-model-versioning:1.2.2"
```
robfletcher: Are we actually using this guy?

keel-retrofit/keel-retrofit.gradle
```diff
@@ -19,9 +19,9 @@ apply plugin: "kotlin-spring"
 dependencies {
   api "com.squareup.retrofit2:retrofit:${spinnaker.version('retrofit2')}"
   api "com.squareup.retrofit2:converter-jackson:${spinnaker.version('retrofit2')}"
-  api("com.netflix.spinnaker.kork:kork-web:${spinnaker.version('kork')}")
   api "com.jakewharton.retrofit:retrofit2-kotlin-coroutines-adapter:0.9.2"
-  api "org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion"
 
+  implementation "com.netflix.spinnaker.kork:kork-web:${spinnaker.version('kork')}"
```
robfletcher: In the long run it would be nice to separate the stuff Retrofit needs from the general Spring web MVC config inside kork but that's a job for another day.

robfletcher:
What's the advantage of the Caffeine cache builder? Just curious.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/config/ClouddriverConfiguration.kt
robfletcher: I had to create a new config key just for that bean as well since they changed one of the log level enum names in Retrofit 2.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
ben-manes: Yes. This is because sometimes absence is a valid response (e.g. user not found when looking up by email). It matches `Map.computeIfAbsent` syntax, so better match. Since you know it can never happen, this is perfect.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
```diff
@@ -117,7 +117,7 @@ class MemoryCloudDriverCache(
...
 
   override fun subnetBy(subnetId: String): Subnet =
     subnets.getOrNotFound(subnetId, "Subnet with id $subnetId not found") {
```
ben-manes: I believe `getOrNotFound` could be implemented using a `cache.get(key, function)` like your usages above.

ben-manes:
> What's the advantage of the Caffeine cache builder? Just curious.

@robfletcher You might enjoy this blog series ([part 1](http://highscalability.com/blog/2016/1/25/design-of-a-modern-cache.html), [part 2](http://highscalability.com/blog/2019/2/25/design-of-a-modern-cachepart-deux.html)). I suspect in this case it might be due to improved concurrency and active maintenance.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
cfieber: good point, thanks!

keel-front50/keel-front50.gradle
cfieber: looks like we are not, removing it

PR #252, opened 2019-03-13
refactor(config): Use constructor args for config rather than public var

robfletcher:


keel-retrofit/src/main/kotlin/com/netflix/spinnaker/keel/retrofit/KeelRetrofitConfiguration.kt
```diff
@@ -35,21 +35,21 @@ import java.util.concurrent.TimeUnit.MILLISECONDS
...
 @Import(OkHttp3ClientConfiguration::class)
 @EnableConfigurationProperties
-class KeelRetrofitConfiguration {
-
-  private val log = LoggerFactory.getLogger(javaClass)
-
+class KeelRetrofitConfiguration(
   @Value("\${ok-http-client.connection-pool.max-idle-connections:5}")
-  var maxIdleConnections = 5
+  val maxIdleConnections: Int = 5,
```
emjburns: Does this actually let you set the property in the config? I thought it had to be var to be resettable

robfletcher: Why would it ever get re-set?

emjburns: you've given it a default value of 5 but you can override it by setting `ok-http-client.connection-pool.max-idle-connections`

emjburns: I thought it had to be `var` for setting by property to actually work with a default value

cfieber: I think it works in a constructor as val because its either injected or it uses the default.

It would be good to verify that changing the associated properties actually changes the supplied values

PR #253, opened 2019-03-13
feat(validation): support for custom validation

emjburns:
Creates and wires in validators to all handlers. Validators should be able to run in any order and may mutate the resource.

Another PR coming internally for doing the netflix-specific security group validation

Processes validation in `ResourceHandler<>`


keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandler.kt
```diff
@@ -9,17 +9,22 @@ import com.netflix.spinnaker.keel.api.deliveryconfig.DeliveryConfig
...
 import kotlinx.coroutines.runBlocking
+import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import retrofit2.HttpException
 
 class DeliveryConfigHandler(
   private val front50Service: Front50Service,
-  override val objectMapper: ObjectMapper
+  override val objectMapper: ObjectMapper,
+  override val validators: List<ResourceValidator<DeliveryConfig>>
```
robfletcher: Would it make sense to wire these into `ResourceActuator` rather than each individual handler?

keel-ec2-plugin/keel-ec2-plugin.gradle
```diff
@@ -4,7 +4,7 @@ apply from: "$rootDir/gradle/junit5.gradle"
 
 dependencies {
   implementation project(":keel-retrofit")
-  implementation project(":keel-plugin")
+  api project(":keel-plugin")
```
robfletcher: Is this so something in `keel-nflx` can see it?

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/InvalidResourceException.kt
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.exceptions
+
+import com.netflix.spinnaker.keel.api.Resource
+
+
+sealed class InvalidResourceException(override val message: String?) : RuntimeException(message)
```
robfletcher: Do these get handled anywhere?

keel-ec2-plugin/keel-ec2-plugin.gradle
emjburns: Yep - so that custom validation applied in keel-ec2-netflix can have these dependencies

keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandler.kt
emjburns: Discussed offline: going to move to the `ResourcePersister`

keel-ec2-plugin/keel-ec2-plugin.gradle
cfieber: more pedantic than anything but I tend to sort the dependencies by scope (api, implementation, testImplementation) then the dependencies within a scope by project then alphabetically

if that sounds reasonable then just bumping this to the start of the list would be appreciated!

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt
```diff
@@ -24,8 +24,12 @@ import com.netflix.spinnaker.keel.api.ResourceName
 import com.netflix.spinnaker.keel.api.SubmittedResource
 import com.netflix.spinnaker.keel.api.randomUID
 import de.danielbechler.diff.node.DiffNode
+import jdk.nashorn.internal.runtime.regexp.joni.Config.log
```
cfieber: [log](https://www.youtube.com/embed/8-9scNP5KWk)

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/DefaultResourceValidator.kt
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package com.netflix.spinnaker.keel.plugin
+
+import com.netflix.spinnaker.keel.api.Resource
+import org.springframework.stereotype.Component
+import kotlin.reflect.KClass
+
+@Component
+class DefaultResourceValidator : ResourceValidator<Any> {
```
cfieber: I'm not sure this provides anything useful other than ensuring there is at least one ResourceValidator

The default implementation of validate in ResourceHandler does an objectMapper conversion of the spec to the generic type which would be useful if we could get it in here (but the typing bit of that might be tricky to pull off)

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceValidator.kt
```diff
@@ -0,0 +1,32 @@
...
+
+package com.netflix.spinnaker.keel.plugin
+
+import com.netflix.spinnaker.keel.api.Resource
+import kotlin.reflect.KClass
+
+/**
+ * A resource validator throws a [FailedValidationException] when a resource is invalid
+ */
+interface ResourceValidator<T : Any> {
```
cfieber: I think `T : Any` is redundant (could just use `<T>`)

robfletcher: no, because then it's equivalent to `<T: Any?>`

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/DefaultResourceValidator.kt
emjburns: Removing it, you are right. Was just using it for testing.

keel-ec2-plugin/keel-ec2-plugin.gradle
emjburns: I will make an attempt to sort things alphabetically in the future and fix here :)

keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandler.kt
emjburns: @cfieber discussed offline with @robfletcher: Wiring into the `ResourcePersister` actually made things clunky because the naming filling in of metadata still needed to happen with all the knowledge that the handlers had (https://github.com/spinnaker/keel/blob/master/keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt#L48). Instead I've simplified the wiring and all the validation happens in the base `ResourceHandler` class.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/InvalidResourceException.kt
emjburns: Now they're handled better in `ResourceController`

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
```diff
@@ -102,4 +104,21 @@ class ResourceController(
     log.error(e.message)
     return mapOf("message" to (e.cause?.message ?: e.message))
   }
+
+  @ExceptionHandler(FailedValidationException::class)
+  @ResponseStatus(BAD_REQUEST)
```
robfletcher: If this exception type means "I could parse it but it was logically wrong" (e.g. a cluster with a region that doesn't exist) this should be `UNPROCESSABLE_ENTITY`.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceNormalizer.kt
```diff
@@ -0,0 +1,39 @@
...
+
+package com.netflix.spinnaker.keel.plugin
+
+import com.netflix.spinnaker.keel.api.ApiVersion
+import com.netflix.spinnaker.keel.api.Resource
+
+/**
+ * A resource normalizer throws a [FailedValidationException] when a resource is invalid
+ */
+interface ResourceNormalizer<T : Any> {
```
cfieber: could also theoretically name this `ResourceTransformer` or `ResourceInterceptor` but naming sucks and I'm not proposing you rename anything

PR #256, opened 2019-03-13
Added some SQS configuration

robfletcher:


keel-sqs/src/main/kotlin/com/netflix/spinnaker/config/SqsProperties.kt
```diff
@@ -5,4 +5,6 @@ import org.springframework.boot.context.properties.ConfigurationProperties
 @ConfigurationProperties("sqs")
 class SqsProperties {
   var queueARN: String = ""
+  var visibilityTimeoutSeconds: Int = 60
```
robfletcher: Determines how long we have to process the message before it will potentially be visible to another `receiveMessages` call. It looks like the default is zero so we really need to use this to avoid instances reading duplicate messages.

keel-sqs/src/main/kotlin/com/netflix/spinnaker/config/SqsProperties.kt
```diff
@@ -5,4 +5,6 @@ import org.springframework.boot.context.properties.ConfigurationProperties
...
+  var waitTimeSeconds: Int = 10
```
robfletcher: How long the `receiveMessages` call will block for (shorter if a message arrives). We're in a coroutine scope when we read from the queue so there's no point busy-waiting.

keel-sqs/src/main/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckListener.kt
```diff
@@ -39,17 +43,16 @@ internal class SqsResourceCheckListener(
 
   internal fun handleMessages() {
     with(sqsClient) {
-      receiveMessage(queueUrl)
+      receiveMessages()
         .also {
-          log.debug("Got {} messages", it.messages.size)
+          if (it.isNotEmpty()) log.debug("Got {} messages: {}", it.size, it.map(Message::getMessageId))
```
robfletcher: Logging is pretty noisy currently. I've also added the message ids so I can validate there's no duplicate reading going on on multiple instances in the Keel cluster.

keel-sqs/src/main/kotlin/com/netflix/spinnaker/config/SqsProperties.kt
emjburns: In echo I'm using 30 seconds. 60 seems good.

keel-sqs/src/main/kotlin/com/netflix/spinnaker/config/SqsProperties.kt
emjburns: I'm using 5 in echo

robfletcher: But you're probably not using coroutines

PR #257, opened 2019-03-14
Persist state history for resources

robfletcher:
Persists the resource state every time it is checked. I'm also persisting the instance that did the check in the SQL implementation. Can put some metrics around that later perhaps to help track utilization.

I'd like to change the resource state stuff in a future PR so it's persisting a log rather than just the current state.


keel-api/keel-api.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
   api project(":keel-core")
   api project(":keel-plugin")
 
+  implementation project(":keel-eureka")
```
robfletcher: New module

robfletcher:
Actually I think persisting the instance id is bullshit. Let me rework this a bit.

cfieber:
seems like it would be useful info for trying to debug things later if nothing else..

robfletcher:
That's why I started doing it but then I thought what we should really do is emit a spectator metric.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
```diff
@@ -86,10 +90,21 @@ class ResourceController(
...
-    log.info("Deleting: $name")
+    log.debug("Deleting: $name")
     return resourcePersister.handle(ResourceDeleted(name))
   }
 
+  @GetMapping(
+    path = ["/{name}/history"],
+    produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  fun history(@PathVariable("name") name: ResourceName): List<ResourceStateHistoryEntry> {
```
robfletcher: New endpoint to get state history

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceStateHistoryEntry.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.keel.persistence
+
+import java.time.Instant
+
+data class ResourceStateHistoryEntry(
+  val state: ResourceState,
+  val timestamp: Instant
```
robfletcher: I figure we can add more to this type. The details of the detected diff, for example.

robfletcher:
I'm leaving the Eureka stuff in because the next thing I want to do is Eureka activation so I'll just end up re-implementing it.

PR #259, opened 2019-03-14
Better coroutine usage in SQS listener

robfletcher:
Took some tips from https://jivimberg.io/blog/2019/02/23/sqs-consumer-using-kotlin-coroutines/

### DIFF
--- a/keel-sqs/src/main/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckListener.kt
+++ b/keel-sqs/src/main/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckListener.kt
@@ -10,13 +10,20 @@ import com.fasterxml.jackson.module.kotlin.readValue
 import com.netflix.spinnaker.config.SqsProperties
 import com.netflix.spinnaker.keel.annealing.ResourceActuator
 import com.netflix.spinnaker.keel.api.ResourceName
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers.Default
+import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.cancelAndJoin
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.channels.ReceiveChannel
+import kotlinx.coroutines.channels.SendChannel
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.yield
 import org.slf4j.LoggerFactory
+import java.lang.Thread.currentThread
 import javax.annotation.PostConstruct
 import javax.annotation.PreDestroy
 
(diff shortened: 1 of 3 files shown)


keel-sqs/src/main/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckListener.kt
```diff
@@ -27,38 +34,48 @@ internal class SqsResourceCheckListener(
...
 
   @PostConstruct
   fun startListening() {
-    listenerJob = scope.launch {
-      while (isActive) {
-        handleMessages()
+    rootJob = scope.launch {
+      val channel = Channel<Message>()
+      launchMessageReceiver(channel)
+      repeat(10) {
```
cfieber: maybe throw this number in a `@Value` or `@ConfigurationProperties`?

keel-sqs/src/main/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckListener.kt
```diff
@@ -27,38 +34,48 @@ internal class SqsResourceCheckListener(
...
-            log.error("Resource Actuator threw {}: {}", e.javaClass.simpleName, e.message)
+          channel.send(it)
+        }
+    }
+  }
+
+  private fun CoroutineScope.launchWorker(channel: ReceiveChannel<Message>) = launch {
+    runUntilCancelled {
+      for (message in channel) {
+        try {
```
cfieber: I wonder if its worth doing a call to SQS to set the visibilityTimeout here before starting to work on the message to ensure that it is still a valid message handle and that you have time remaining to work on it

I don't know if there is a case where your channel gets backed up if all the workers are busy and the message takes too long to dispatch to a worker

Probably pretty theoretical for the scale at which this queue would be operating

cfieber: also if we don't care about two different consumers processing the same message than this probably doesn't matter either

keel-sqs/src/main/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckListener.kt
```diff
@@ -27,38 +34,48 @@ internal class SqsResourceCheckListener(
...
-          try {
-            with(it.parse()) {
-              actuator.checkResource(name.let(::ResourceName), apiVersion, kind)
-            }
-            deleteMessage(it)
-          } catch (e: JsonProcessingException) {
-            log.error("Could not parse message payload: {}", it.body)
-          } catch (e: Throwable) {
-            log.error("Resource Actuator threw {}: {}", e.javaClass.simpleName, e.message)
+          channel.send(it)
```
cfieber: is there a version of send that will fail if it can't enqueue the message within some timeframe? would probably be better than blocking indefinitely here, because we could just reset the visibilityTimeout to zero and let another instance receive the message if so

keel-sqs/src/main/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckListener.kt
robfletcher: Yeah. I meant to. Then I forgotted.

keel-sqs/src/main/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckListener.kt
robfletcher: Yeah, the operations are currently idempotent in theory. The SQS queue just lets work propagate across the cluster. I still need to get the scheduler to be smarter because currently all the instances are going to schedule checks for all the things!

keel-sqs/src/main/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckListener.kt
robfletcher: This is a coroutine channel send, not an SQS operation.

robfletcher:
Yes, discovery is the next step. I have thoughts.

robfletcher:
I would also like to use the async SQS client so that we're not making blocking calls. TODO.

keel-sqs/src/main/kotlin/com/netflix/spinnaker/keel/sqs/SqsResourceCheckListener.kt
robfletcher: I don't see any override of this method, unfortunately.

PR #261, opened 2019-03-15
feat(deliveryConfig): actuating on dynamic image

cfieber:
this "works" in that it will detect a newer image and trigger annealing
of a cluster, however it needs a solid overhaul to handle resource
references, hence the pretty much zero tests currently

this orphans the front50 module currently, but I didn't remove it as
I suspect we may still want that at some point

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ClusterActiveServerGroup.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ClusterActiveServerGroup.kt
@@ -19,7 +19,7 @@ data class ClusterActiveServerGroup(
 data class LaunchConfig(
   val ramdiskId: String?,
   val ebsOptimized: Boolean,
-  val imageId: String,
+  val imageId: String?,
   val instanceType: String,
   val keyName: String,
   val iamInstanceProfile: String,
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/NamedImage.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/NamedImage.kt
@@ -1,0 +1,8 @@
+package com.netflix.spinnaker.keel.clouddriver.model
+
+data class NamedImage(
+  val imageName: String,
+  val attributes: Map<String, Any?>,
+  val tagsByImageId: Map<String, Map<String, String?>?>,
+  val accounts: Set<String>,
+  val amis: Map<String, List<String>?>)
--- a/keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/deliveryconfig/DeliveryConfig.kt
+++ b/keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/deliveryconfig/DeliveryConfig.kt
@@ -6,5 +6,7 @@ import com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL
 @JsonInclude(NON_NULL)
 data class DeliveryConfig(
   val name: String,
-  val application: String
+  val application: String,
+  val deliveryArtifacts: List<Map<String, Any>> = emptyList(),
+  val deliveryEnvironments: List<DeliveryEnvironment> = emptyList()
 )
--- a/keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/deliveryconfig/DeliveryEnvironment.kt
+++ b/keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/deliveryconfig/DeliveryEnvironment.kt
@@ -1,0 +1,25 @@
+package com.netflix.spinnaker.keel.api.deliveryconfig
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL
+import com.netflix.spinnaker.keel.api.ApiVersion
+import com.netflix.spinnaker.keel.api.ResourceMetadata
+
+data class DeliveryEnvironment(
+  val name: String,
+  val packageRef: ChildResource,
+  val targets: List<ChildResource>
+)
+
+@JsonInclude(NON_NULL)
+data class ChildResource(
+  val apiVersion: ApiVersion,
+  val kind: String,
+  val spec: Map<String, Any?>,
+  val metadata: Map<String, Any?>?
+) {
+  @get:JsonIgnore
+  val resourceMetadata : ResourceMetadata?
+    get() = metadata?.let { ResourceMetadata(it) }
+}
--- a/keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandler.kt
+++ b/keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandler.kt
@@ -1,26 +1,33 @@
 package com.netflix.spinnaker.keel.deliveryconfig.resource
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.convertValue
+import com.netflix.spinnaker.keel.annealing.ResourcePersister
 import com.netflix.spinnaker.keel.api.Resource
 import com.netflix.spinnaker.keel.api.ResourceKind
 import com.netflix.spinnaker.keel.api.ResourceName
 import com.netflix.spinnaker.keel.api.SPINNAKER_API_V1
+import com.netflix.spinnaker.keel.api.SubmittedResource
+import com.netflix.spinnaker.keel.api.deliveryconfig.ChildResource
 import com.netflix.spinnaker.keel.api.deliveryconfig.DeliveryConfig
-import com.netflix.spinnaker.keel.front50.Front50Service
-import com.netflix.spinnaker.keel.front50.model.Delivery
+import com.netflix.spinnaker.keel.api.deliveryconfig.DeliveryEnvironment
+import com.netflix.spinnaker.keel.events.ResourceCreated
+import com.netflix.spinnaker.keel.events.ResourceUpdated
+import com.netflix.spinnaker.keel.persistence.ResourceRepository
+import com.netflix.spinnaker.keel.persistence.get
 import com.netflix.spinnaker.keel.plugin.ResourceHandler
+import com.netflix.spinnaker.keel.plugin.ResourceHandler.ResourceDiff
 import com.netflix.spinnaker.keel.plugin.ResourceNormalizer
-import com.netflix.spinnaker.keel.retrofit.isNotFound
-import de.danielbechler.diff.node.DiffNode
 import kotlinx.coroutines.runBlocking
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
-import retrofit2.HttpException
+import org.springframework.beans.factory.ObjectFactory
 
 class DeliveryConfigHandler(
-  private val front50Service: Front50Service,
   override val objectMapper: ObjectMapper,
-  override val normalizers: List<ResourceNormalizer<*>>
+  override val normalizers: List<ResourceNormalizer<*>>,
+  private val resourcePersisterProvider: ObjectFactory<ResourcePersister>,
+  private val resourceRepositoryProvider: ObjectFactory<ResourceRepository>
 ) : ResourceHandler<DeliveryConfig> {
   override val log: Logger by lazy { LoggerFactory.getLogger(javaClass) }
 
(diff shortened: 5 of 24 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -61,26 +62,58 @@ class ClusterHandler(
     }
 
   override fun upsert(resource: Resource<Cluster>, diff: DiffNode?) {
-    val taskRef = runBlocking {
+    runBlocking {
       val spec = resource.spec
       val job = when {
         diff.isCapacityOnly() -> spec.resizeServerGroupJob()
-        else -> spec.createServerGroupJob()
+        else -> spec.createServerGroupJob().let { job ->
```
emjburns: General style comment - I dislike chaining big functions with the `.let{}` / `when` etc. I prefer something like 
```
else -> {
  val job = spec.createServerGroupJob()
  when {
    job["amiName"] != null -> job
   resource.metadata.data["packageRef"] != null ->  {
     //a function for what you're doing with the if/else
   }
  }
}
```
I find the very chained style pretty hard to read and reason about when you're looking through someone else's code.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageHandler.kt
```diff
@@ -0,0 +1,54 @@
...
+  override val log: Logger by lazy { LoggerFactory.getLogger(javaClass) }
+  override val apiVersion = SPINNAKER_API_V1.subApi("ec2")
+  override val supportedKind = ResourceKind(
+    apiVersion.group,
+    "namedImage",
+    "namedImage"
+  ) to NamedImage::class.java
+
+  override fun generateName(spec: NamedImage) = ResourceName(
+    "ec2:image:${spec.account}:${spec.name}"
```
emjburns: Do we need region in here?

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/NamedImage.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.keel.clouddriver.model
+
+class NamedImage(
+  val imageName: String,
+  val attributes: Map<String, Any?>,
+  val tagsByImageId: Map<String, Map<String, String?>?>,
+  val accounts: Set<String>,
+  val amis: Map<String, List<String>?>)
```
emjburns: region here?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageHandler.kt
cfieber: the NamedImage is multiregion (amis is a map of region -> list of amiId)

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
cfieber: I've reworked this to pull more stuff out into functions but I would really welcome any more `this is too hard to read/reason about` feedback on where I landed. This kotlin thing is still pretty shiny for me so I am not entirely sure which features I like or dislike for terseness vs readability

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ClusterActiveServerGroup.kt
```diff
@@ -19,7 +19,7 @@ data class ClusterActiveServerGroup(
 data class LaunchConfig(
   val ramdiskId: String?,
   val ebsOptimized: Boolean,
-  val imageId: String,
+  val imageId: String?,
```
cfieber: imageId becomes nullable to indicate that the cluster handler expects metadata supplying the image.

as we discussed in person, this is almost certainly the wrong thing where the spec ends up always being out of sync with the current state of the world, and the upsert becomes responsible for no-oping when the metadata supplied image is correct

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/NamedImage.kt
cfieber: I'm not super happy with the typing and naming of these attributes but they are modelling the response from clouddriver's find endpoint. If this sticks around longer term then I'd like to make sure it is more clear that this is class is multi-regional

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/NamedImage.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.keel.clouddriver.model
+
+class NamedImage(
```
cfieber: I should probably make this a data class...

keel-deliveryconfig-plugin/keel-deliveryconfig-plugin.gradle
```diff
@@ -6,6 +6,7 @@ dependencies {
   implementation project(":keel-retrofit")
   implementation project(":keel-plugin")
   implementation project(":keel-front50")
+  implementation "javax.inject:javax.inject:1"
```
cfieber: needed for `Provider<BeanType>` to break circular dependency

spring probably has a native equivalent but I somewhat prefer the generic annotations

keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/deliveryconfig/DeliveryConfig.kt
```diff
@@ -1,10 +1,21 @@
...
 
 @JsonInclude(NON_NULL)
 data class DeliveryConfig(
   val name: String,
-  val application: String
-)
+  val application: String,
+  val deliveryArtifacts: List<Map<String, Any>> = emptyList(),
+  val deliveryEnvironments: List<DeliveryEnvironment> = emptyList(),
+  @get:JsonAnyGetter val otherAttributes: MutableMap<String, Any> = mutableMapOf()
```
cfieber: this should be an immutable map, I'm not sure how I feel about json any setters and data classes yet (I could go the route of a constructor that takes Map<String, Any> but that feels similarly bad

keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/deliveryconfig/DeliveryEnvironment.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.netflix.spinnaker.keel.api.ApiVersion
+import com.netflix.spinnaker.keel.api.ResourceMetadata
+
+data class DeliveryEnvironment(
+  val name: String,
+  val packageRef: PackageRef,
+  val targets: List<Target>
+)
+
+interface SubResource {
```
cfieber: the interface + base class + typealiases in here exist because at one point I thought I would have additional attributes on the different types, but in hindsight I think those don't make sense with the model of treating things directly as resources (I'd pull any attributes up to the DeliveryEnvironment level)

cfieber: so I can and probably will eventually just make this a single type, but I'd prefer to leave it for now and address it in the follow up change where we start exploring a better way to do type referencing

keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandler.kt
```diff
@@ -35,39 +42,98 @@ class DeliveryConfigHandler(
     "deliveryConfig:${spec.name}"
   )
 
+  private val resourcePersister: ResourcePersister by lazy { resourcePersisterProvider.get() }
```
cfieber: defers the bean resolution from the app context until it is actually used, breaks the circular dependency of resourcePersister needs resource handlers but this is a resource handler and it needs the resourcePersister

keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandler.kt
```diff
@@ -35,39 +42,98 @@ class DeliveryConfigHandler(
...
-        val delivery = front50Service.deliveryById(resource.spec.name).await()
-        log.debug("Fetched $delivery")
-        DeliveryConfig(delivery.id, delivery.application)
-      } catch (e: HttpException) {
-        if (!e.isNotFound) {
-          log.error("Fetching delivery from front50 failed", e)
-          throw e
-        }
-        null
+    if (resource.spec.isResolved) {
```
cfieber: ensure that if any of the sub resources components in the spec are not yet being managed (don't have populated metadata) that we resubmit them. currently does this dumbly by just bombing out with 'current state is null' and could potentially be smarter, but the upsert doesn't look at the diff at all anyway so I think this is ok

keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandler.kt
```diff
@@ -35,39 +42,98 @@ class DeliveryConfigHandler(
...
     }
 
-
   override fun delete(resource: Resource<DeliveryConfig>) {
-    runBlocking {
-      log.debug("deleting delivery ${resource.spec.name} in application ${resource.spec.application}")
-      front50Service.deleteDelivery(resource.spec.application, resource.spec.name).await()
-      log.debug("deleted delivery ${resource.spec.name} in application ${resource.spec.application}")
-    }
+    TODO("Not implemented")
```
cfieber: I'm sure this will be easy.....

keel-deliveryconfig-plugin/src/test/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandlerTests.kt
```diff
@@ -48,23 +46,9 @@ internal class DeliveryConfigHandlerTests : JUnit5Minutests {
...
-      test("the current model is null") {
-        expectThat(current(resource)).isNull()
-      }
-
-      test("upserting a deliveryconfig persists to front50") {
-        upsert(resource)
-        verify { front50Service.upsertDelivery(spec.name, Delivery(spec.name, spec.application)) }
-      }
-    }
+    //TODO(cfieber) - some tests would be good
```
cfieber: This line probably won't be here in 2020

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -60,34 +78,103 @@ class ClusterHandler(
...
     }
 
-  override fun upsert(resource: Resource<Cluster>, diff: DiffNode?) {
-    val taskRef = runBlocking {
+  private fun Resource<Cluster>.imageIdFromMetadata(): String? =
+    packageResource()?.let { pkg ->
+      imageIdSuppliers.find { it.supports(pkg) }?.imageIdForCluster(pkg, this.spec)
+    }
+
+  interface ImageIdSupplier {
```
cfieber: this is just inline in the class currently but sort of mocks out how I might see making the imageIdForCluster logic pluggable by the resource type for the image

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -60,34 +78,103 @@ class ClusterHandler(
...
-    this != null && affectedRootPropertyTypes.all { it == Capacity::class.java }
+  private fun ResourceDiff<Cluster>?.isCapacityOnly(): Boolean =
+    this != null && diff.affectedRootPropertyTypes.all { it == Capacity::class.java }
+
+  private fun ResourceDiff<Cluster>?.isImageOnly(): Boolean =
+    this != null && this.diff.imageChanged && this.diff.nodePaths.size == 1
+
+  private val imageIdPath = NodePath.with("launchConfiguration", "imageId")
+
+  private val DiffNode.imageChanged: Boolean
```
cfieber: feels like we might want to think about how we could hide away this object diffing library, I suspect we'll end up with many similar bits of code to this as we keep building out keel, it might be nice if we had some layer in front of it to contain the dependency

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -148,7 +238,7 @@ class ClusterHandler(
         "cluster" to moniker.cluster
       ),
       "amiName" to launchConfiguration.imageId,
-      "reason" to "Diff detected at ${clock.instant().iso()}",
+      "reason" to "ResourceDiff detected at ${clock.instant().iso()}",
```
cfieber: oops, thanks intellij

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt
```diff
@@ -196,7 +197,7 @@ internal object SecurityGroupHandlerTests : JUnit5Minutests {
 
     sequenceOf(
       "create" to SecurityGroupHandler::create,
-      "update" to SecurityGroupHandler::update.partially3(DiffNode.newRootNode())
+      "update" to SecurityGroupHandler::update.partially3(ResourceDiff(UpsertFixture().resource.spec))
```
cfieber: @robfletcher calling `UpsertFixture()` here seems wrong, is there a better way to reference the fixture here to get the resource?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -114,6 +117,7 @@ class SqlResourceRepository(
         field("kind") to resource.kind,
         field("name") to resource.metadata.name.value,
         field("resource_version") to resource.metadata.resourceVersion,
+        field("metadata") to objectMapper.writeValueAsString(resource.metadata.data),
```
cfieber: this is just the extra data bits of the metadata, currently we would drop those on the floor

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
emjburns: I agree

keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandler.kt
emjburns: It kind of feels like the handlers should have access to the resource persister. Maybe there's a better way - emitting some sort of resource event that then gets picked up by the API and persists from there?

keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/deliveryconfig/DeliveryEnvironment.kt
robfletcher: I find this very confusing

keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/deliveryconfig/DeliveryConfig.kt
cfieber: removing it entirely, it was mostly there for when I was expecting to pull this from front50

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/NamedImage.kt
cfieber: oversight, fixed!

keel-deliveryconfig-plugin/keel-deliveryconfig-plugin.gradle
cfieber: sure, dropped this and switched to spring ObjectFactory which is the equivalent

keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandler.kt
```diff
@@ -35,39 +42,98 @@ class DeliveryConfigHandler(
...
+                t.getOrCreateResource().updateMetadata(mapOf("packageResource" to pkg))
+              }
+            )
+          }
+        }
+      )))
     }
   }
 
+  private fun DeliveryEnvironment.refreshFromRepository() =
```
emjburns: This way of doing things is one reason why I love kotlin

keel-deliveryconfig-plugin/src/test/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandlerTests.kt
```diff
@@ -48,23 +46,9 @@ internal class DeliveryConfigHandlerTests : JUnit5Minutests {
 
   fun tests() = rootContext<DeliveryConfigHandler> {
     fixture {
-      DeliveryConfigHandler(front50Service, objectMapper, normalizers)
+      DeliveryConfigHandler(objectMapper, normalizers, ObjectFactory { resourcePersister }, ObjectFactory { resourceRepository })
```
emjburns: pls mind line length

PR #262, opened 2019-03-16
Discovery activation for scheduled checks and SQS listener

robfletcher:


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/Main.kt
```diff
@@ -55,7 +54,8 @@ private val DEFAULT_PROPS = mapOf(
   "spring.config.name" to "spinnaker,\${spring.application.name}",
   "spring.profiles.active" to "\${netflix.environment},local",
   // TODO: not sure why we need this when it should get loaded from application.properties
-  "spring.main.allow-bean-definition-overriding" to "true"
+  "spring.main.allow-bean-definition-overriding" to "true",
+  "spring.groovy.template.check-template-location" to "false"
```
emjburns: What does this do?

robfletcher: I was getting log noise on startup

PR #267, opened 2019-03-21
Use actual MySQL instead of H2 for SQL tests

robfletcher:
Would love any thoughts on this. I ran into a compatibility issue the other day that got me looking at alternatives to using H2 in MySQL "compatibility" mode. This is using `testcontainers` but is considerably faster than the test I wrote for SQS that was using a very large docker image.

### DIFF
--- a/keel-sql/src/main/resources/db/changelog/20190320-create-lock-table.yml
+++ b/keel-sql/src/main/resources/db/changelog/20190320-create-lock-table.yml
@@ -4,7 +4,7 @@ databaseChangeLog:
       author: fletch
       changes:
         - createTable:
-            tableName: lock
+            tableName: cluster_lock
             columns:
               - column:
                   name: name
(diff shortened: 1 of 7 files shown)


keel-sql/src/main/resources/db/changelog/20190320-create-lock-table.yml
```diff
@@ -4,7 +4,7 @@ databaseChangeLog:
       author: fletch
       changes:
         - createTable:
-            tableName: lock
+            tableName: cluster_lock
```
robfletcher: Testcontainers found another bug: `lock` is a reserved word in MySQL but H2 allowed it.

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/TestDatabase.kt
```diff
@@ -64,23 +59,25 @@ internal fun initDatabase(jdbcUrl: String): DSLContext {
...
+    .apply {
+      meta()
+        .schemas
+        .filter { it.name == schema }
+        .flatMap(Schema::getTables)
+        .filterNot {
+          it.name in setOf(
+            databaseChangeLogTableName,
+            databaseChangeLogLockTableName
+          )
```
robfletcher: D'oh! If I truncate these tables the following test will fall over trying to re-run the Liquibase changeset.

PR #268, opened 2019-03-22
Added a simple metric counter when checking resources

robfletcher:


keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceActuator.kt
```diff
@@ -4,35 +4,41 @@ import com.netflix.spinnaker.keel.api.ApiVersion
 import com.netflix.spinnaker.keel.api.Resource
 import com.netflix.spinnaker.keel.api.ResourceName
 import com.netflix.spinnaker.keel.persistence.ResourceRepository
-import com.netflix.spinnaker.keel.persistence.ResourceState.Diff as Different
+import com.netflix.spinnaker.keel.persistence.ResourceState.Diff
```
robfletcher: WTF Cam?

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryListener.kt
```diff
@@ -0,0 +1,38 @@
...
+
+  @EventListener(ResourceChecked::class)
+  fun onResourceChecked(event: ResourceChecked) {
+    scope.launch {
+      try {
+        spectator.counter(
+          resourceCheckedCounterId
+            .withTag("resource_name", event.name.value)
+            .withTag("resource_state", event.state.name)
+            .withTag("instance_id", instanceIdSupplier.get())
```
robfletcher: Can't remember if I actually need to tag the instance id or if that is done automagically.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceActuator.kt
```diff
@@ -4,35 +4,41 @@ import com.netflix.spinnaker.keel.api.ApiVersion
...
 import org.slf4j.LoggerFactory
+import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Component
 
 @Component
 class ResourceActuator(
   private val resourceRepository: ResourceRepository,
-  private val handlers: List<ResourceHandler<*>>
+  private val handlers: List<ResourceHandler<*>>,
+  private val publisher: ApplicationEventPublisher
```
robfletcher: I'll rather raise a simple event and handle it out of band that inject `Registry` here.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceActuator.kt
cfieber: loool I was building you a DifferenceEngine

cfieber: actually I created a class originally called Diff that I subsequently renamed but I forgot where I did this thing sorry

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryListener.kt
cfieber: the instanceId should already be added automatically

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceActuator.kt
robfletcher: I figured you were importing something from the diff lib at one point, yeah.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryListener.kt
robfletcher: Yeah, just checked Orca & couldn't see it. Also appears like camel case rather than snake case is the convention for tag names.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceActuator.kt
cfieber: I could get behind this strategy, lets giv'er a go

is the event publisher we are using async and multi-threaded? (I feel like that is maybe a spring config type of thing..) If we are using it for metrics then I think it shouldn't be a single threaded event processing queue as it would kind of suck for metrics to get stuck around some other slow event handler

keel-api/src/main/kotlin/com/netflix/spinnaker/config/EventsConfiguration.kt
```diff
@@ -0,0 +1,22 @@
...
+  fun applicationEventTaskExecutor() =
+    ThreadPoolTaskExecutor().apply {
+      threadNamePrefix = "event-pool-"
+    }
+
+  @Bean
+  fun applicationEventMulticaster(applicationEventTaskExecutor: TaskExecutor) =
+    SimpleApplicationEventMulticaster().apply {
+      setTaskExecutor(applicationEventTaskExecutor)
+    }
```
robfletcher: This is the magic required so that event listeners use a different thread. The way I've configured it there's an initial pool of 1 thread with an unbounded max size.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationEventTests.kt
```diff
@@ -0,0 +1,66 @@
...
+    val testThread = Thread.currentThread()
+
+    publisher.publishEvent(TestEvent(this))
+
+    val eventThread = listener.awaitInvoked(Duration.ofMillis(100))
+
+    expectThat(eventThread)
+      .isNotEqualTo(testThread)
+      .get { name }
+      .startsWith("event-pool-")
```
robfletcher: Thought it was worth testing the events actually do get handled on another thread since it's dependent on bean-naming magic.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceActuator.kt
robfletcher: It is following 112e577

Issue #165, opened 2018-10-02
Resource deletion


robfletcher:
Do we want the ability for users to define a desired state of “this resource MUST NOT exist” and delete it if it ever does?

luispollo:
@robfletcher Can you please add a description for this? What are "assets" in this context, "resources"?

robfletcher:
Issue showing its age. Asset was the original name for what we now call resource. I’m increasingly convinced this is something we don’t need.

emjburns:
Closing for now - this doesn't seem to be required for beta, and it's easy enough to create a new issue.

PR #271, opened 2019-04-13
Basic resource event tracking

robfletcher:
I removed the existing resource state stuff and instead persist events. The responsibility for that is in `ResourcePersister` (CRUD events) and `ResourceActuator` (diff detection / resolution events).

Currently the events contain only metadata. The next step will be to flesh them out with useful information, such as diff trees.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
```diff
@@ -61,9 +57,9 @@ class ResourceController(
     produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
   )
   @ResponseStatus(CREATED)
-  fun create(@RequestBody submittedResource: SubmittedResource<*>): Resource<*> {
+  fun create(@RequestBody submittedResource: SubmittedResource<Any>): Resource<out Any> {
     log.debug("Creating: $submittedResource")
-    return resourcePersister.handle(ResourceCreated(submittedResource))
+    return resourcePersister.create(submittedResource)
```
robfletcher: This `handle(some type of event)` formation is cruft from when `ResourcePersister` was consuming events from the Spring event bus. It makes more sense to have regular ol' individual methods now.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -255,14 +197,20 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
...
           expectThrows<NoSuchResourceException> {
             subject.get<Map<String, Any>>(resource.metadata.name)
           }
         }
+
+        test("events for the resource are also deleted") {
+          expectThrows<NoSuchResourceException>() {
+            subject.eventHistory(resource.metadata.uid)
+          }
+        }
```
robfletcher: Deleting effectively functions as a "stop managing this and clean up" which is pretty useful for testing stuff out in staging.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/events/ResourceEventSerializationTests.kt
```diff
@@ -0,0 +1,119 @@
...
+import strikt.api.expectThat
+import strikt.assertions.isEqualTo
+import strikt.jackson.has
+import strikt.jackson.path
+import strikt.jackson.textValue
+import java.time.Clock
+import java.time.Instant
+import java.time.ZoneId
+
+internal class ResourceEventSerializationTests : JUnit5Minutests {
```
robfletcher: Polymorphism +  Jackson deserialization is always a barrel of laughs so I put some tests around it.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourcePersister.kt
```diff
@@ -1,44 +1,52 @@
...
+    handlers.supporting(resource.apiVersion, resource.kind)
+      .normalize(resource)
+      .also(resourceRepository::store)
+      .also { resourceRepository.appendHistory(ResourceCreated(it, clock)) }
+      .also { queue.scheduleCheck(it) }
+
+  fun update(resource: Resource<Any>): Resource<out Any> {
+    val existing = resourceRepository.get<Any>(resource.metadata.uid)
+
+    return if (existing.spec == resource.spec) {
```
robfletcher: Eventually this will become where we diff the resources & store the delta in the event. For now straight equality check will do.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -128,115 +125,66 @@ class SqlResourceRepository(
...
+        .values(
+          event.uid.toString(),
+          event.timestamp.let(Timestamp::from),
+          objectMapper.writeValueAsString(event)
+        )
+        .execute()
     }
   }
 
   override fun delete(name: ResourceName) {
```
cfieber: Thoughts on a model where we have a deleted flag on the resource but don't remove it from the data store? Was thinking it might make for a better auditing story?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
```diff
@@ -0,0 +1,49 @@
...
+      resourceRepository
+        .eventHistory(resource.metadata.uid)
+        .filter { it.timestamp > since }
+    }
+  }
+
+  // TODO: would be nice to make this common with ResourceController
+  @ExceptionHandler(NoSuchResourceException::class)
+  @ResponseStatus(NOT_FOUND)
+  fun onNotFound(e: NoSuchResourceException) {
```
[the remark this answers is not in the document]
robfletcher: I’ll take a look

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
robfletcher: Yeah at some point we’ll want to do that. I think for now the key thing is just to be able to get rid of stuff so we can do repeated demos!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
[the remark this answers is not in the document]
robfletcher: Going to make a separate PR to align some of the exception types.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
robfletcher: I'm going to go ahead and merge this. We have #165 to track a proper deletion solution. I agree that keeping audit history and just flagging a resource as inactive is probably the right thing, but I think it's best handled as its own PR.

PR #272, opened 2019-04-16
Persist diffs when delta is detected or resource is updated

robfletcher:
The event we persist when updating a resource or when a delta between desired & actual state is detected now has a diff attached so when we look at the history we can see what happened.

See #152 

Long term I'm not sure if we want to consider storing every version of a resource as it gets updated but I figured that would be expensive when generating the event stream as we'd be recreating the diff for each version.


emjburns:
What does it look like when you get the history for a resource?

PR #279, opened 2019-04-19
Initial persistence for delivery artifacts

robfletcher:


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -29,8 +31,3 @@ data class EchoArtifactEvent(
   val payload: ArtifactEvent,
   val eventName: String
 )
-
-data class ArtifactEvent(
```
robfletcher: Just moved this somewhere more central

keel-artifact/keel-artifact.gradle.kts
```diff
@@ -0,0 +1,12 @@
...
+  id("kotlin-spring")
+}
+
+dependencies {
+  implementation(project(":keel-core"))
+  implementation("org.springframework:spring-context")
+
+  testImplementation("dev.minutest:minutest")
+  testImplementation("io.strikt:strikt-core")
+}
```
robfletcher: Not sure we really need a separate sub-module here

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
```diff
@@ -0,0 +1,18 @@
...
+)
+
+enum class ArtifactType {
+  DEB
+}
+
+data class DeliveryArtifactVersion(
+  val artifact: DeliveryArtifact,
+  val version: String,
+  val provenance: URI
```
robfletcher: Can add to this model as & when we need more things

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -0,0 +1,32 @@
...
+
+@Component
+class ArtifactListener(
+  private val artifactRepository: ArtifactRepository
+) {
+  @EventListener(ArtifactEvent::class)
+  fun onArtifactEvent(event: ArtifactEvent) {
+    log.info("Received artifact event: {}", event)
+    event.artifacts.forEach {
+      val registeredArtifact = artifactRepository.get(it.name, ArtifactType.valueOf(it.type))
```
emjburns: This assumes that the delivery config with artifact info has already been saved, and so we have some concept of what artifacts we care about?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -0,0 +1,111 @@
...
+              DeliveryArtifactVersion(
+                artifact,
+                "1.0",
+                URI("https://my.jenkins.master/job/${artifact.name}-release/1")
+              )
+            )
+          }
+        }
+
+        // TODO: should we throw an exception instead?
```
emjburns: I don't think so, because we may get the same message more than once, and that should be nicely handled

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -0,0 +1,17 @@
...
+import com.netflix.spinnaker.keel.api.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.DeliveryArtifactVersion
+
+interface ArtifactRepository {
+
+  fun store(artifact: DeliveryArtifact)
+  fun store(artifactVersion: DeliveryArtifactVersion)
+
+  // TODO: should this just return boolean?
+  fun get(name: String, type: ArtifactType): DeliveryArtifact?
```
emjburns: Maybe there should be another function to just see if the artifact exists.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
robfletcher: It's checked prior to this

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
robfletcher: Yeah, agreed

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
robfletcher: Oh I see what you mean, I need to add an endpoint for that

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -1,36 +1,53 @@
...
-    log.debug("Received artifact events for ${echoArtifactEvent.payload.artifacts.map { it.reference }}")
+    log.debug(
+      "Received artifact events {} for {}",
+      echoArtifactEvent.eventName,
+      echoArtifactEvent.payload.artifacts.map { it.reference }
+    )
     publisher.publishEvent(echoArtifactEvent.payload)
   }
+
+  @PostMapping(path = ["/register"], consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE])
```
robfletcher: I don't love this URL. Would like to change the one for the echo events

PR #280, opened 2019-04-20
record an atlas event if we add an artifact version

robfletcher:


keel-artifact/keel-artifact.gradle.kts
```diff
@@ -5,6 +5,7 @@ plugins {
 
 dependencies {
   implementation(project(":keel-core"))
+  implementation(project(":keel-plugin"))
```
robfletcher: I feel like I need to rethink where some things live in the keel packages

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -27,6 +30,11 @@ class ArtifactListener(
           if (artifactRepository.isRegistered(artifact.name, artifact.type)) {
             log.info("Registering version {} of {} {}", version, artifact.name, artifact.type)
             artifactRepository.store(this)
+              .also { wasAdded ->
+                if (wasAdded) {
+                  publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.type))
```
robfletcher: Only record telemetry if we actually add a new version

PR #282, opened 2019-04-29
Auto bake

robfletcher:
This attempts to reconcile a desired state consisting of a package and a bake spec by looking up:

1. The current base AMI corresponding to the OS & label (retrieved from Bakery’s fast properties via Mahe)
2. The current artifact (fetched from Igor by package & version)
3. Existing AMIs (fetched from The Clouddriver)

Limitations

* Only works for Rocket right now
* Assumes Netflix bakery (might want to move some of this to `keel-nflx`)
* Only handles the latest version of packages

Next step is to figure out how to tie this up with clusters via/replacing the existing `NamedImageHandler` bits.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -62,7 +61,7 @@ class ArtifactController(
   fun listVersions(
     @PathVariable name: String,
     @PathVariable type: ArtifactType
-  ): List<DeliveryArtifactVersion> =
+  ): List<String> =
```
robfletcher: Simplified this signature so we’re just storing & returning `appversion` strings.

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/ami/BaseImageCache.kt
```diff
@@ -0,0 +1,57 @@
...
+import com.netflix.spinnaker.keel.bakery.api.BaseLabel
+import com.netflix.spinnaker.keel.mahe.DynamicPropertyService
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import org.slf4j.LoggerFactory
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+
+@Component
+class BaseImageCache(
```
robfletcher: Looks up & caches latest base AMIs by OS / label

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,131 @@
...
+    with(resource.spec) {
+      val latestVersion = artifactRepository
+        .versions(DeliveryArtifact(artifactName, DEB))
+        .first()
+      Image(
+        baseAmiVersion = baseImageCache.getBaseImage(baseOs, baseLabel),
+        // TODO: there must be a canonical way of composing this
+        appVersion = latestVersion,
+        regions = regions
+      )
```
robfletcher: Desired state of an image is the specific base AMI version, specific app version, and the regions

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,131 @@
...
+      cloudDriver.images("aws", resource.spec.artifactName)
+        .await()
+        .firstOrNull()
+        ?.let {
+          // TODO: this is pretty crude to say the least
+          val tags = it.tagsByImageId.values.first()!!
+          Image(
+            tags.getValue("base_ami_version")!!,
+            tags.getValue("appversion")!!,
+            it.amis.keys
```
robfletcher: Current state is latest baked AMI fetched from The Clouddriver

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+  override fun upsert(
+    resource: Resource<ImageSpec>,
+    resourceDiff: ResourceDiff<Image>?
+  ): List<TaskRef> {
+    val taskRef = runBlocking {
+      val (_, application, version) = Regex("([\\w_]+)-(.+?)/.*")
+        .find(resourceDiff!!.source.appVersion)!!.groupValues
+      val artifact = igorService.getArtifact(application, version)
```
robfletcher: Before baking fetch the full artifact from Igor.

We _could_ store this when we initially get it via Rocket events but there’s some consensus that we need a mechanism to catch up on any potentially missed events and I kind of don’t want to replicate a ton of data from elsewhere.

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,131 @@
...
+      log.info("baking new image for {}", resource.spec.artifactName)
+      orcaService.orchestrate(
+        OrchestrationRequest(
+          name = "Bake image for ${resource.spec.artifactName}",
+          application = resource.spec.artifactName, // TODO: is this always the same? Does it matter?
+          description = "Bake image for ${resource.spec.artifactName}",
+          job = listOf(
+            Job(
+              "bake",
```
robfletcher: Upsert via a bake

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,131 @@
...
+    TODO("not implemented")
+  }
+
+  override val log: Logger by lazy { LoggerFactory.getLogger(javaClass) }
+}
+
+/**
+ * The resolved representation of an [ImageSpec].
+ */
+data class Image(
```
robfletcher: Represents the current / desired state when resolved against whatever versions of the package and base AMI exist

keel-deliveryconfig-plugin/src/test/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandlerTests.kt
```diff
@@ -29,8 +29,7 @@ internal class DeliveryConfigHandlerTests : JUnit5Minutests {
     "deliveryconfig",
     ResourceMetadata(
       name = ResourceName("foo"),
-      uid = idGenerator.nextValue(),
-      resourceVersion = 1234L
+      uid = idGenerator.nextValue()
```
robfletcher: We never use the `resourceVersion` property so I killed it off.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceActuator.kt
```diff
@@ -42,6 +42,8 @@ class ResourceActuator(
     val plugin = handlers.supporting(apiVersion, kind)
     val type = plugin.supportedKind.second
     val resource = resourceRepository.get(name, type)
+
+    val desired = plugin.desired(resource)
```
robfletcher: We can now generate a desired state based on a spec. In the case of an image this means figuring out the specific app and base AMI versions

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResolvableResourceHandler.kt
```diff
@@ -0,0 +1,161 @@
...
+import com.netflix.spinnaker.keel.exceptions.InvalidResourceStructureException
+import org.slf4j.Logger
+
+/**
+ * @param S the spec type.
+ * @param R the resolved model type.
+ *
+ * If those two are the same, use [ResourceHandler] instead.
+ */
+interface ResolvableResourceHandler<S : Any, R : Any> : KeelPlugin {
```
robfletcher: Super-interface of `ResourceHandler` where the spec and desired/current types may be different

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt
```diff
@@ -15,155 +15,15 @@
...
-    TODO("Not implemented")
-  }
+interface ResourceHandler<T : Any> : ResolvableResourceHandler<T, T> {
 
   /**
-   * Delete a resource as the desired state is that it should no longer exist.
+   * This method is a no-op when the model and resolved types are the same.
    */
-  fun delete(resource: Resource<T>)
+  override fun desired(resource: Resource<T>): T = resource.spec
```
robfletcher: In the `ResourceHandler` interface spec type and current/desired type are the same.

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+  override fun current(resource: Resource<ImageSpec>): Image? =
+    runBlocking {
+      cloudDriver.images("aws", resource.spec.artifactName)
+        .await()
+        .firstOrNull()
+        ?.let {
+          // TODO: this is pretty crude to say the least
+          val tags = it.tagsByImageId.values.first()!!
```
robfletcher: Is there ever a situation where the image would not be tagged? How should we handle that?

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,131 @@
...
+          Image(
+            tags.getValue("base_ami_version")!!,
+            tags.getValue("appversion")!!,
```
robfletcher: Null checks where are because the declared map type is `<String, String?>` I'm pretty sure these 2 specifically should never be null.

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+  override fun upsert(
+    resource: Resource<ImageSpec>,
+    resourceDiff: ResourceDiff<Image>?
+  ): List<TaskRef> {
+    val taskRef = runBlocking {
+      val (_, application, version) = Regex("([\\w_]+)-(.+?)/.*")
+        .find(resourceDiff!!.source.appVersion)!!.groupValues
```
robfletcher: Just realized I can't assume this is non-null. Fixing now.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -21,22 +19,19 @@ class ArtifactListener(
...
-          if (artifactRepository.isRegistered(artifact.name, artifact.type)) {
-            log.info("Registering version {} of {} {}", version, artifact.name, artifact.type)
-            artifactRepository.store(this)
-              .also { wasAdded ->
-                if (wasAdded) {
-                  publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.type))
-                }
-              }
+      val artifact = DeliveryArtifact(it.name, ArtifactType.valueOf(it.type))
+      // TODO: should be able to construct this with Frigga or something, apparently, also it might
```
emjburns: I think you can only parse with Frigga, but I might be mistaken.

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,135 @@
...
+    apiVersion.group,
+    "image",
+    "images"
+  ) to ImageSpec::class.java
+
+  override fun generateName(spec: ImageSpec): ResourceName =
+    ResourceName("bakery:image:${spec.artifactName}")
+
+  override fun desired(resource: Resource<ImageSpec>): Image =
+    with(resource.spec) {
```
emjburns: In the future, do you image this could support a strategy? Like the spec might say "latest" or specify a version, and then the desired would resolve to the actual version?

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
emjburns: For now I think we should assume that all images baked by us are tagged by the bakery. When we come to un-netflix this, we can re evaluate

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
emjburns: We have images which have no tags. I'd probably say do some sort of check here to discard those, instead of failing hard? Unless we're prefiltering and looking for only things that have tags or only the base ami

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,135 @@
...
+    resourceDiff: ResourceDiff<Image>
+  ): List<TaskRef> {
+    val taskRef = runBlocking {
+      val (_, application, version) = resourceDiff.desired.appVersion.let { appVersion ->
+        Regex("([\\w_]+)-(.+?)/.*")
+          .find(appVersion)
+          ?.groupValues
+          ?: throw IllegalStateException("Could not parse app version $appVersion")
+      }
+      val artifact = igorService.getArtifact(application, version)
```
emjburns: FYI, the this method on igor can be configured to supply artifacts from not only rocket (e.g. artifactory) which pivotal is working on. So this is less netflix-specific than the find image from tags stuff

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
robfletcher: I'll have a dig. I don't think I currently have Frigga on the classpath.

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
robfletcher: I _think_ so. I haven't given it a whole lot of thought.

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
robfletcher: RIght, I hard-coded it for now, but I saw the provider thing in the URL. I need to figure out how we'd know what type it was here. I guess we'll have to persist that when the event comes in.

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,135 @@
...
+      log.info("baking new image for {}", resource.spec.artifactName)
+      orcaService.orchestrate(
+        OrchestrationRequest(
+          name = "Bake image for ${resource.spec.artifactName}",
+          application = resource.spec.artifactName, // TODO: is this always the same? Does it matter?
```
emjburns: I don't know if we want to use artifact name here - the name most likely won't correspond to an app (think of things like redis that are common or packages that have a different name from their app, like if our package was `keel-nflx`). Either, we should have someone define an app in their config to run tasks in, or we should just run it in keel for now.

An alternative for finding the app for an image (for netflix) is to build an endpoint in swabbie. Swabbie already has a cache of these things and is pulling internal data to see if an image is used anywhere in any app. We could easily expose this to get that application. This doesn't solve the "many apps use one package" case though.

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -0,0 +1,135 @@
...
+    TODO("not implemented")
+  }
+
+  override val log: Logger by lazy { LoggerFactory.getLogger(javaClass) }
+}
+
+/**
+ * The resolved representation of an [ImageSpec].
+ */
+data class Image(
```
emjburns: Should we also be storing whether this is an ami or a docker image? Or should that be a refactor for a later date?

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
robfletcher: Makes sense.

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
robfletcher: I think that would be a completely different resource type and handler.

robfletcher: But even if not, I'd intend to add that when we need it rather than trying to anticipate & getting it wrong.

keel-bakery/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
robfletcher: Right now we don't have a general way of tying resources to applications. I'll make it bake as Keel for now & we can revisit when we decide to tackle that issue.

emjburns:
Overall I think this looks great. I like the change in the resource handler to have a ResolveableResourceHandler. This definitely lines up with what @cfieber and I have talked about where you may store something in a spec that's not what you want to diff, and then have that desired state calculated at the right low level when you need to diff it

robfletcher:
I still think there’s more thought we want to do around the idea of converting a spec to a concrete desired state but I definitely think this takes us in the right direction.

PR #285, opened 2019-05-02
Record metrics for errors on checking resources

robfletcher:


keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/ResourceState.kt
```diff
@@ -13,8 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.netflix.spinnaker.keel.persistence
+package com.netflix.spinnaker.keel.telemetry
```
robfletcher: Moved this enum because it's only used for telemetry

PR #289, opened 2019-05-16
Simpler actuation loop

robfletcher:
I ripped out all the SQS and cluster locking stuff in favor of a method on the repository that can just return a batch of resources to check in least recently checked order updating their checked times as it goes. This is a lot simpler and should distribute the work better.


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -220,11 +226,104 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
     }
   }
 
+  data class CheckLockFixture<T : ResourceRepository>(
```
robfletcher: Tests for new functionality in resource repository

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -78,6 +79,20 @@ interface ResourceRepository {
...
   fun appendHistory(event: ResourceEvent)
+
+  /**
+   * Returns between zero and [limit] resources that have not been checked (i.e. returned by this
+   * method) in at least [minTimeSinceLastCheck].
+   *
+   * This method is _not_ intended to be idempotent, subsequent calls are expected to return
+   * different values.
+   */
+  fun nextResourcesDueForCheck(minTimeSinceLastCheck: Duration, limit: Int): Collection<ResourceHeader>
```
robfletcher: New method to get a batch of resources to check

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -78,6 +79,20 @@ interface ResourceRepository {
...
+
+  /**
+   * Marks [resource] as due for checking so that it will be returned by [nextResourcesDueForCheck].
+   */
+  fun markCheckDue(resource: Resource<*>)
```
robfletcher: Actively mark a resource as due for a check (called after update)

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceCheckScheduler.kt
```diff
@@ -17,13 +12,8 @@ import java.time.Duration
 @Component
 class ResourceCheckScheduler(
   private val resourceRepository: ResourceRepository,
-  private val resourceCheckQueue: ResourceCheckQueue,
-  private val lock: Lock,
-  private val publisher: ApplicationEventPublisher,
-  @Value("\${keel.resource.monitoring.frequency.ms:60000}")
-  private val frequencyMs: Long
+  private val resourceActuator: ResourceActuator
```
robfletcher: Look how much simpler

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourceCheckScheduler.kt
```diff
@@ -38,29 +28,20 @@ class ResourceCheckScheduler(
...
-        resourceRepository.allResources { (_, name, apiVersion, kind) ->
-          resourceCheckQueue.scheduleCheck(name, apiVersion, kind)
+  @Scheduled(fixedDelayString = "\${keel.resource.check.frequency.ms:1000}")
+  fun checkResources() {
+    if (enabled) {
+      log.debug("Starting scheduled validation…")
+      resourceRepository
+        .nextResourcesDueForCheck(Duration.ofMinutes(1), 1)
+        .forEach { (_, name, apiVersion, kind) ->
+          resourceActuator.checkResource(name, apiVersion, kind)
```
robfletcher: I should do some coroutining here at some point so this loop doesn't block.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -141,18 +143,53 @@ class SqlResourceRepository(
     }
   }
 
+  override fun nextResourcesDueForCheck(minTimeSinceLastCheck: Duration, limit: Int): Collection<ResourceHeader> {
```
robfletcher: Would love @asher's take on the viability of this implementation

settings.gradle.kts
```diff
@@ -34,7 +34,6 @@ include(
   "keel-retrofit-test-support",
   "keel-spring-test-support",
   "keel-sql",
-  "keel-sqs",
```
robfletcher: bye Felicia

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/annealing/ResourcePersister.kt
```diff
@@ -42,9 +42,11 @@ class ResourcePersister(
     return if (diff.hasChanges()) {
       log.debug("Resource {} updated: {}", normalized.metadata.name, diff.toDebug(normalized.spec, existing.spec))
       normalized
-        .also(resourceRepository::store)
-        .also { resourceRepository.appendHistory(ResourceUpdated(it, diff.toUpdateJson(normalized.spec, existing.spec), clock)) }
-        .also { queue.scheduleCheck(it) }
+        .also {
+          resourceRepository.store(it)
+          resourceRepository.appendHistory(ResourceUpdated(it, diff.toUpdateJson(it.spec, existing.spec), clock))
+          resourceRepository.markCheckDue(it)
```
robfletcher: Instead of explicitly checking a resource after an update I just bump it to the front of the line.

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepositoryTests.kt
```diff
@@ -28,4 +42,58 @@ internal object SqlResourceRepositoryTests : ResourceRepositoryTests<SqlResource
...
+    }
+
+    after { flush() }
+
+    /**
+     * I'd like to have this test in the superclass but the in-memory implementation is not designed
+     * to be thread safe and the Redis implementation needs too much work to make it thread safe
+     * given it's not our preferred target.
+     */
+    context("many threads are checking simultaneously") {
```
robfletcher: I think this satisfies me that we won't see the same resources getting selected for checking by multiple instances in the Keel cluster.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -141,18 +143,53 @@ class SqlResourceRepository(
...
+    val now = clock.instant()
+    val cutoff = now.minus(minTimeSinceLastCheck).let(Timestamp::from)
+    return jooq.inTransaction {
+      select(UID, API_VERSION, KIND, NAME)
+        .from(RESOURCE)
+        .where(LAST_CHECKED.isNull.or(LAST_CHECKED.lessOrEqual(cutoff)))
+        .orderBy(LAST_CHECKED)
+        .limit(limit)
+        .forUpdate()
```
asher: I suspect this would result in deadlocks w/mysql's default isolation level but I think it will be ok in `READ_COMMITTED` which kork-sql should already be setting on connections by default (it's already the default postgres isolation level.) I'd still expect lock wait contention between concurrently scheduled keel instances which may constrain how large the `limit` variable here can be set to. I can imagine a world where we have a resource table per type, with checks scheduled independently behind a short lived distributed lock per type to get better throughput, though it would be nice if small limits avoid the need for this without throughput being an issue.

To help ascertain how viable this is, we can do some benchmarking via an expanded version of `SqlResourceRepositoryTests.parallelCheckingTests()` with millions of rows and iterations testing increasing `limit` values. It would also be good to include some concurrent inserts of new resources to validate that insert performance isn't impacted. This is going to need some db configuration to actually get useful lock wait timing data. It could also be run against an Aurora instance that has Perf Insights enabled, in which case we'd be able to graph lock wait times in great detail.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -141,18 +143,53 @@ class SqlResourceRepository(
     }
   }
 
+  override fun nextResourcesDueForCheck(minTimeSinceLastCheck: Duration, limit: Int): Collection<ResourceHeader> {
+    val now = clock.instant()
+    val cutoff = now.minus(minTimeSinceLastCheck).let(Timestamp::from)
+    return jooq.inTransaction {
+      select(UID, API_VERSION, KIND, NAME)
+        .from(RESOURCE)
+        .where(LAST_CHECKED.isNull.or(LAST_CHECKED.lessOrEqual(cutoff)))
```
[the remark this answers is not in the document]
robfletcher: I couldn’t get Jooq to cooperate but I guess I could just default it on the column.

keel-sql/src/main/resources/db/changelog/20190515-add-last-checked-timestamp.yml
```diff
@@ -0,0 +1,18 @@
+databaseChangeLog:
+  - changeSet:
+      id: add-last-checked-timestamp
+      author: fletch
+      changes:
+        - addColumn:
+            tableName: resource
+            columns:
+              - column:
+                  name: last_checked
```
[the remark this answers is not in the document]
robfletcher: Damn. I meant to do that and forgot.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
robfletcher: I was thinking of using a limit of 1 to start with!

keel-sql/src/main/resources/db/changelog/20190515-add-last-checked-timestamp.yml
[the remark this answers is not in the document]
robfletcher: Fixed

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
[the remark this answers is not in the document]
robfletcher: Figured it out. MySQL is stupid. We should use Postgres.

PR #292, opened 2019-05-17
Skip actuation when Orca is already busy

robfletcher:


keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -142,6 +142,11 @@ class ImageHandler(
     TODO("not implemented")
   }
 
+  override fun actuationInProgress(name: ResourceName) =
```
emjburns: why does this function need to be in all the handlers? can't it be pulled up?

robfletcher: Not all handlers necessarily use Orca. I would like to pull it up to a common interface or something but I haven't quite figured out where it should live.

PR #293, opened 2019-05-17
Propagate kork-bom

robfletcher:


marchello2000:
seems reasonable to me, but i defer to @cfieber for expert review

build.gradle.kts
```diff
@@ -19,44 +22,49 @@ plugins {
...
   group = "com.netflix.spinnaker.keel"
 }
 
 subprojects {
-  apply(plugin = "nebula.kotlin")
-
-  repositories {
-    jcenter()
-  }
+  if (name != "keel-bom") {
```
robfletcher: I hate this, but this stuff has to not be applied to the `keel-bom` project as it's not compatible with the `java-platform` plugin.

build.gradle.kts
```diff
@@ -19,44 +22,49 @@ plugins {
...
 
-  dependencies {
-    "implementation"(platform("com.netflix.spinnaker.kork:kork-bom:${property("korkVersion")}"))
+    repositories {
+      jcenter()
+    }
 
-    "implementation"("org.slf4j:slf4j-api")
+    dependencies {
+      "implementation"(platform("com.netflix.spinnaker.kork:kork-bom:${property("korkVersion")}"))
```
robfletcher: Version has to be specified here or it's missing in the generated bom and then Gradle can't parse it.

keel-bom/keel-bom.gradle.kts
```diff
@@ -0,0 +1,50 @@
...
+    publications {
+      named<MavenPublication>("nebula") {
+        from(components["javaPlatform"])
+      }
+    }
+  }
+}
+
+dependencies {
+  api(platform("com.netflix.spinnaker.kork:kork-bom:${property("korkVersion")}"))
```
robfletcher: Version has to be specified here or it's missing in the generated bom and then Gradle can't parse it.

keel-bom/keel-bom.gradle.kts
```diff
@@ -0,0 +1,49 @@
...
+  constraints {
+    api(project(":keel-api"))
```
cfieber: I feel like you could possibly programmatically generate this list by getting the rootProject and enumerating the subprojects excluding this project - might be worth a look so that forgetting to add a new module to this list doesn't become a thing?

PR #301, opened 2019-05-22
Deserialize security group rules without type tokens

robfletcher:


keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -19,7 +19,7 @@ class ArtifactListener(
   fun onArtifactEvent(event: ArtifactEvent) {
     log.info("Received artifact event: {}", event)
     event.artifacts.forEach {
-      val artifact = DeliveryArtifact(it.name, ArtifactType.valueOf(it.type))
+      val artifact = DeliveryArtifact(it.name, ArtifactType.valueOf(it.type.toUpperCase()))
```
robfletcher: Noticed a bunch of errors in the logs for this.

PR #305, opened 2019-05-22
feat(image): provide image with jenkins information

emjburns:
It's pretty easy to add another ImageProvider. Added a preliminary jenkins one here - it relies on our special netflix "appversion" tag and "build_host" tag that every image gets tagged with when it bakes.

Not loving the way you have to specify this detail (build_host is a full url), but this is good beginning functionality for testing these things out and working with some of our alpha adopters.

This looks like:

```
"launchConfiguration": {
  "imageProvider": {
    "packageName": "deb-sample-app-server",
    "buildHost": "https://my.jenkins.url/",
    "buildName": "MY-JENKINS-job-name",
    "buildNumber": "92"
  },
```

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/Dates.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/Dates.kt
@@ -1,0 +1,55 @@
+/*
+ *
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.clouddriver
+
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+
+class Dates {
+  companion object {
+    private val formats: List<DateTimeFormatter> = listOf(
+      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSS'Z'"),
+      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ssZ"),
+      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ"),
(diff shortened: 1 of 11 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -98,11 +99,26 @@ class ClusterHandler(
...
+        log.info("Image found for {}: {}", artifactName, namedImage)
 
         val amis = namedImage.amis[region] ?: throw NoImageFoundForRegion(artifactName, region)
         return amis.first() // todo eb: when are there multiple?
       }
+      is JenkinsJobImageProvider -> {
+        val namedImage = runBlocking {
+          imageService.getNamedImageFromJenkinsInfo(
+            imageProvider.packageName,
+            "test",
```
asher: This should be a dynamic property that just defaults to test (same with other occurrences in the class.)

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -62,4 +62,42 @@ class ImageService(
       .await()
       .lastOrNull()
   }
+
+  suspend fun getNamedImageFromJenkinsInfo(packageName: String, account: String, buildHost: String, buildName: String, buildNumber: String): NamedImage? {
+    return cloudDriverService.namedImages(packageName, account)
+      .await()
+      .lastOrNull { namedImage ->
```
asher: I see that this is due to clouddriver's image search endpoint lexically sorting images by name but don't think that guarantees the last element is actually the most recent package. Lexical sorting isn't great for all versioning schemes. I don't think this PR should change, just a note that this could theoretically result in incorrect artifact delivery unless we modify clouddriver to sort by a timestamp if possible.

emjburns: Good note, I'll address this because we have creation date in returned in NamedImage and that will be a really annoying bug to troubleshoot later.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/Dates.kt
```diff
@@ -0,0 +1,55 @@
...
+ *
+ */
+package com.netflix.spinnaker.keel.clouddriver
+
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+
+class Dates {
+  companion object {
+    private val formats: List<DateTimeFormatter> = listOf(
```
robfletcher: Is this because our images are tagged inconsistently? Because… 😨

emjburns: This is pulled from swabbie because I spent ~10 minutes trying to parse the date w/o it and then got very annoyed and took what we're already doing.

PR #307, opened 2019-05-23
Fixed update resource API endpoint

robfletcher:


keel-deliveryconfig-plugin/src/main/kotlin/com/netflix/spinnaker/keel/deliveryconfig/resource/DeliveryConfigHandler.kt
```diff
@@ -114,13 +114,13 @@ class DeliveryConfigHandler(
...
-          this.copy(
-            metadata = this.metadata?.let { it + extraMetadata }
-          ))).asBaseSubResource()
+//  private fun ChildResource.updateMetadata(extraMetadata: Map<String, Any?>) =
+//    resourcePersister
+//      .update(
+//        objectMapper.convertValue(
+//          this.copy(
+//            metadata = this.metadata?.let { it + extraMetadata }
+//          ))).asBaseSubResource()
```
robfletcher: Had to disable some things in this class, but it's about to change anyway. I notice this didn't break any tests 😠

PR #309, opened 2019-05-31
Stop scheduler halting when Retrofit experiences an issue

robfletcher:


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceCheckScheduler.kt
```diff
@@ -34,6 +36,7 @@ class ResourceCheckScheduler(
   @Scheduled(fixedDelayString = "\${keel.resource-check.frequency:PT1S}")
   fun checkResources() {
     if (enabled) {
+      publisher.publishEvent(ScheduledCheckStarting)
```
emjburns: Why the change to publishing an event here?

robfletcher: Needed something for the test to listen to to confirm the scheduler does not hang.

I figure tacking a health check to this probably makes sense too.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/integration/SchedulingResilienceTests.kt
```diff
@@ -0,0 +1,88 @@
...
+  lateinit var service: DummyRetrofitService
+
+  @Test
+  fun `retrofit call completes even if an interceptor throws an exception`() {
+    val latch = CountDownLatch(1)
+    GlobalScope.launch {
+      try {
+        service.greet().await()
+      } finally {
+        latch.countDown()
```
robfletcher: Don't care if the method returns or throws an exception. Just ensuring it doesn't hang.

keel-retrofit/src/main/kotlin/com/netflix/spinnaker/keel/retrofit/KeelRetrofitConfiguration.kt
```diff
@@ -82,18 +83,22 @@ class KeelRetrofitConfiguration {
...
             v.ifPresent {
               requestBuilder.addHeader(k, it)
             }
           }
       }
-      AuthenticatedRequest.allowAnonymous { chain.proceed(requestBuilder.build()) }
+      try {
+        AuthenticatedRequest.allowAnonymous { chain.proceed(requestBuilder.build()) }
+      } catch (ex: RuntimeException) {
+        throw IOException(ex)
```
robfletcher: This is a workaround we could remove once kork#305 is merged

PR #310, opened 2019-06-03
Gauge tracking drift in resource check times

robfletcher:


keel-api/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.context.annotation.Configuration
+import org.springframework.context.annotation.Scope
+import org.springframework.core.Ordered.HIGHEST_PRECEDENCE
+import org.springframework.core.annotation.Order
+import org.springframework.security.config.annotation.web.builders.HttpSecurity
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter
+import java.time.Clock
+
+@Configuration
+class DefaultConfiguration {
```
robfletcher: Had to move this stuff out of `Main` or the new health indicator got a circular bean dependency on the `Clock`. I was thinking about doing this anyway so w/e.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceCheckHealthIndicator.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.stereotype.Component
+import java.time.Clock
+import java.time.Duration
+import java.time.Instant
+import java.time.format.DateTimeFormatter.ISO_LOCAL_DATE_TIME
+
+@Component
+class ResourceCheckHealthIndicator(
+  private val clock: Clock,
+  @Value("\${keel.health.resource-check.threshold:PT1M}") private val healthyThreshold: Duration
```
robfletcher: We'll start failing health checks if the scheduled check does not happen at least once per minute.

robfletcher:
It occurs to me that this should not cause an unhealthy signal if the instance is disabled and thus not running checks. Any idea what the right thing to do there would be @cfieber ?

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceCheckHealthIndicator.kt
```diff
@@ -0,0 +1,36 @@
...
+) : AbstractHealthIndicator() {
+
+  private var lastCheck: Instant = clock.instant()
+
+  override fun doHealthCheck(builder: Health.Builder) {
+    with(builder) {
+      if (lastCheck >= clock.instant() - healthyThreshold) {
+        up()
+      } else {
+        down()
```
asher: Won't this result in an `ApplicationDown` event that in-turn disables the `checkResources` actuation loop, which would in turn prevent this health check from ever recovering? If actuation cycles are missed for a minute due to a transient db or network issue, we'd want keel to recover on its own once that issue clears.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryListener.kt
```diff
@@ -3,14 +3,31 @@ package com.netflix.spinnaker.keel.telemetry
...
 ) {
+  private var lastResourceCheck: Instant = Instant.EPOCH
+
+  @PostConstruct
+  fun registerMeters() {
+    spectator.gauge(RESOURCE_CHECK_DRIFT_GAUGE, this) {
+      Duration
+        .between(lastResourceCheck, clock.instant())
+        .toMillis()
+        .toDouble()
```
robfletcher: Better than a failing health check

PR #312, opened 2019-06-03
chore(securityGroup): use Moniker, refactor tests

emjburns:
* `SecurityGroup` has a `moniker` instead of `name` for consistency  
* Using one Moniker because having several moniker classes all defining different things is confusing!
* Changing some package structure.
* `internal object SecurityGroupHandlerTests` -> `internal class SecurityGroupHandlerTests` so we can run the test in Intellij


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -23,8 +23,8 @@ import com.netflix.spinnaker.keel.api.ec2.cluster.Dependencies
 import com.netflix.spinnaker.keel.api.ec2.cluster.Health
 import com.netflix.spinnaker.keel.api.ec2.cluster.LaunchConfigurationSpec
 import com.netflix.spinnaker.keel.api.ec2.cluster.Location
-import com.netflix.spinnaker.keel.api.ec2.cluster.Moniker
 import com.netflix.spinnaker.keel.api.ec2.cluster.Scaling
+import com.netflix.spinnaker.keel.clouddriver.model.Moniker
```
robfletcher: The idea with the package structure was that it would clearly delineate things that were part of the Keel API vs things that were part of the Clouddriver API. Maybe we should move this somewhere common?

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupTests.kt
```diff
@@ -19,8 +25,10 @@ internal object SecurityGroupTests : JUnit5Minutests {
   fun diffTests() = rootContext<SecurityGroup> {
     fixture {
       SecurityGroup(
-        application = "fnord",
-        name = "fnord-ext",
+        moniker = Moniker(
+          app = "fnord",
+          stack = "ext"
+        ),
```
robfletcher: Can security groups have a _detail_ value?

robfletcher: Yes

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
emjburns: Sounds good.

PR #316, opened 2019-06-10
feat(ec2): ClassicLoadBalancer support

asher:


asher:
This is a functional implementation to anneal the EC2 CLB resource type. ALB's have a fairly different data model to CLB's and I find the way ALB's were implemented in clouddriver as an extension to ELB's to be inelegant due to the constraints of marrying the two resource types. I wanted to write an initial CLB implementation as a standalone type without concern for building base EC2 loadBalancer classes suitable for both.

This code will likely be refactored as I think about implementing ALB's but I'm interested in thoughts on that and general review.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClassicLoadBalancerHandler.kt
```diff
@@ -0,0 +1,218 @@
...
+        getClassicLoadBalancer(
+          CLOUD_PROVIDER,
+          spec.location.accountName,
+          spec.location.region,
+          spec.moniker.name
+        )
+          .await()
+          .firstOrNull()
+          .run {
+            if (this == null) {
```
emjburns: I think there's some fancier kotlin you can do with the `?` and maybe `.let` so you don't have to have this first null check.

emjburns:
Seems reasonable to me (both this PR and your thoughts about refactoring in the future).

cfieber:
I agree that it is worth keeping CLB and ALB as different resources and looking for opportunities to extract some utility code out of the CLB handler if it makes sense but otherwise not coupling them

Issue #317, opened 2019-06-14
Bug: Diffing doesn't work when a class changes.


emjburns:
When you have in interface (say `ImageProvider`) that has subclasses representing different types, the differ doesn't detect this.

I'll commit this test later, but for example the base cluster has a launch config provided by an ami id, and this tests that if we switch that to an artifact, that it correctly picks up a change. Spoiler alert, this test fails.

(in `ClusterHandlerTests`)
```
context("testing diffing") {
      test("diff exists when specs have different image providers") {
        val lc1 = LaunchConfigurationSpec(
          imageProvider = LatestFromPackageImageProvider(DeliveryArtifact("test-art", ArtifactType.DEB)),
          instanceType = "r4.8xlarge",
          ebsOptimized = false,
          iamRole = "keelRole",
          keyPair = "keel-key-pair",
          instanceMonitoring = false
        )
        val changedSpec = spec.copy(launchConfiguration = lc1)
        val diff = differ.compare(spec, changedSpec)
        expectThat(diff.hasChanges()).isEqualTo(true)
      }
    }
```

robfletcher:
Caused by SQiShER/java-object-diff#186

PR #319, opened 2019-06-17
feat(tags): tag resources that keel manages with an entity tag

emjburns:
There's a bug in our diffing: https://github.com/spinnaker/keel/issues/317. The first section of this PR comments out the "check if things have changed on update" portion of the code so that changes can be made to these resources.

I've created a keel tagging plugin that looks through resources keel manages and makes sure they all have tags. On a delete, an application event is emitted and that tag is removed. 

This will be used for showing that a resource is managed in the UI, and linking from an entity id to a keel id to surface diff things.

I'd love feedback on this approach.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/tags/EntityTag.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/tags/EntityTag.kt
@@ -17,28 +17,31 @@
  */
 package com.netflix.spinnaker.keel.tags
 
+/**
+ * The representation of an entity tag.
+ * The value can be any object.
+ */
 data class EntityTag(
   val value: Any,
   val namespace: String,
   val valueType: String,
-  val category: String,
   val name: String
 )
 
-data class KeelEntityTag(
-  val value: TagValue?,
-  val namespace: String = "keel",
-  val valueType: String = "object",
-  val category: String = "notice",
-  val name: String = KEEL_TAG_NAME
-)
-
+/**
+ * The value of a keel entity tag.
+ * Entity tags can be of any value, but for keel tags we want
+ * an object of this form.
+ */
 data class TagValue(
   val message: String,
   val keelResourceId: String,
   val type: String
 )
 
+/**
+ * Represents the id of an entity as used by the entity tags api
+ */
 data class EntityRef(
   val entityType: String,
   val entityId: String,
@@ -52,6 +55,9 @@ data class EntityRef(
   }
 }
 
+/**
+ * A response containing all entity tags for an object
+ */
 data class EntityTags(
   val id: String,
   val idPattern: String,
@@ -60,6 +66,10 @@ data class EntityTags(
   val entityRef: EntityRef
 )
 
+/**
+ * Metadata about entity tags for an object.
+ * Used only in the [EntityTags] response
+ */
 data class TagsMetadata(
   val name: String,
   val lastModified: Long,
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/tags/EntityTag.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/tags/EntityTag.kt
@@ -1,0 +1,71 @@
+/*
+ *
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.tags
+
+data class EntityTag(
+  val value: Any,
+  val namespace: String,
+  val valueType: String,
+  val category: String,
+  val name: String
+)
+
+data class KeelEntityTag(
+  val value: TagValue?,
+  val namespace: String = "keel",
+  val valueType: String = "object",
+  val category: String = "notice",
+  val name: String = KEEL_TAG_NAME
+)
+
+data class TagValue(
+  val message: String,
+  val keelResourceId: String,
+  val type: String
+)
+
+data class EntityRef(
+  val entityType: String,
+  val entityId: String,
+  val application: String,
+  val region: String,
+  val account: String, // must be account number
+  val cloudProvider: String
+) {
+  fun generateId(): String {
+    return "$cloudProvider:$entityType:$entityId:$account:$region"
+  }
+}
+
+data class EntityTags(
+  val id: String,
+  val idPattern: String,
+  val tags: List<EntityTag>,
+  val tagsMetadata: List<TagsMetadata>,
+  val entityRef: EntityRef
+)
+
+data class TagsMetadata(
+  val name: String,
+  val lastModified: Long,
+  val lastModifiedBy: String,
+  val created: Long,
+  val createdBy: String
+)
+
+const val KEEL_TAG_NAME = "spinnaker_ui_notice:managed_by_spinnaker"
--- a/keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandlerTests.kt
+++ b/keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandlerTests.kt
@@ -183,24 +183,6 @@ internal class ClusterHandlerTests : JUnit5Minutests {
       confirmVerified(orcaService)
     }
 
-    // todo emjburns: enable this test when diffing is fixed https://github.com/spinnaker/keel/issues/317
-
-//    context("testing diffing") {
-//      test("diff exists when specs have different image providers") {
-//        val lc1 = LaunchConfigurationSpec(
-//          imageProvider = LatestFromPackageImageProvider(DeliveryArtifact("test-art", ArtifactType.DEB)),
-//          instanceType = "r4.8xlarge",
-//          ebsOptimized = false,
-//          iamRole = "keelRole",
-//          keyPair = "keel-key-pair",
-//          instanceMonitoring = false
-//        )
-//        val changedSpec = spec.copy(launchConfiguration = lc1)
-//        val diff = differ.compare(spec, changedSpec)
-//        expectThat(diff.hasChanges()).isEqualTo(true)
-//      }
-//    }
-
     context("the cluster does not exist or has no active server groups") {
       before {
         coEvery { cloudDriverService.activeServerGroup() } throws RETROFIT_NOT_FOUND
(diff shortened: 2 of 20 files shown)


keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/KeelTagSpec.kt
```diff
@@ -0,0 +1,58 @@
...
+  val tagState: TagState
+)
+
+data class TaggedResource(
+  val keelId: String,
+  val entityRef: EntityRef,
+  val relevantTag: EntityTag?
+)
+
+sealed class TagState
```
robfletcher: Am I right in thinking we desire a tag on a managed resource, and remove it from anything that gets deleted?

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -0,0 +1,203 @@
...
+        type = "notice"
+      ),
+      namespace = KEEL_TAG_NAMESPACE,
+      valueType = "object",
+      category = "notice",
+      name = KEEL_TAG_NAME
+    )
+    )
+
+  private fun syncAccounts() {
```
robfletcher: Could you just do this with a different `@Schedule` annotation? It would make it easier to test.

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -0,0 +1,203 @@
...
+import org.springframework.context.event.EventListener
+import org.springframework.scheduling.annotation.Scheduled
+import java.time.LocalDateTime
+import java.time.ZoneOffset
+import java.util.concurrent.TimeUnit
+
+/**
+ * A scheduled job that checks to make sure each resource keel is managing
+ * has an entity tag indicating this, as well as linking the entity tag id
+ * to the keel id.
```
robfletcher: If we do this, I'm not clear why we need to model tags as their own resource type. Shouldn't we _either_ periodically check all our resources for tags, _or_ do something after a resource is created or deleted?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/tags/EntityTag.kt
```diff
@@ -0,0 +1,71 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.tags
+
+data class EntityTag(
```
robfletcher: It would be great to have some doc comments explaining what these classes are for, I don't find them very intuitive to understand.

cfieber:
not necessarily in scope for this PR, but one thing I've wondered about is how we handle the desired state being that a resource no longer exists. At the same time, we don't want to clean up a resource just because we don't have a declarative definition for it (stuff managed outside keel, or not onboarded to declarative yet)

For example, say we create a transient cluster to run a canary then want to ensure it doesn't exist as part of the steady state.

One thought I had is if we have these tags, we can tell that keel created a resource and therefor should be responsible for ensuring it is destroyed if there is no longer a resource definition present (in which case maybe we don't want to get rid of tags unless someone deregisters a resource from declarative management.

Alternatively maybe the "there should be no resource" is an actual resource with an empty definition.

Anyway I'll actually look more at the code now...

robfletcher:
I'm not sure about the lifecycle here. If Keel, for example, creates a cluster resource, it should maintain the tag every time it RBs a new server group. Do we want to do that out-of-band of the upsert task or just always tag everything Keel creates?

Similarly on delete, it feels like we could just remove the tag from the resource (given the current strategy of delete = no longer manage) do we really want to be on the hook for monitoring the tag state of that resource until the end of time?

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  @EventListener(ApplicationDown::class)
+  fun onApplicationDown() {
+    log.info("Application down, disabling scheduled resource tagging")
+    enabled = false
+  }
+
+  @Scheduled(fixedDelayString = "\${keel.resource-tagger.frequency:PT10S}")
+  fun checkResources() {
```
cfieber: this looks like it will handle the case of ensuring a tag exists for every resource in keel

Do we also want to ensure a resource exists in keel for every tag?

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/KeelTagSpec.kt
emjburns: yes

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
emjburns: Yep, changing

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
emjburns: As we discussed offline, I'm going to switch to doing something on create and delete + daily removing old "tag not desired" states. 

We can revisit in the future if there's a better optimization.

Re: clouddriver entity tags performance: currently you need elastic search for entity tags, and the correct way is going through clouddriver. Adam does not think there's a performance gain to doing a bulk update of the tags.

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
emjburns: refactored to do things on create and delete instead of this polling. 

That might be a good future check. Can you think of any cases where this might bite us?

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/KeelTagSpec.kt
```diff
@@ -0,0 +1,58 @@
...
+
+data class TagDesired(
+  val tag: EntityTag
+) : TagState()
+
+/**
+ * Desire no keel tags
+ */
+data class TagNotDesired(
+  val blah: String = "wtf"
```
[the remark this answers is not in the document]
emjburns: oops, good catch.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/tags/EntityTag.kt
emjburns: Adding, and removing unneeded ones!

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/KeelTagHandler.kt
```diff
@@ -0,0 +1,202 @@
...
+
+  /**
+   * Orca Job for removing Keel tags
+   */
+  private fun TaggedResource.removeTagJob() =
+    mapOf(
+      "type" to "deleteEntityTags",
+      "application" to entityRef.application,
+      "description" to "Removing entity tag for $keelId",
+      "tags" to listOf(relevantTag?.name),
```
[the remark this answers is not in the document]
emjburns: This just does a partial delete. It only removes the keel tag.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandlerTests.kt
```diff
@@ -183,6 +183,24 @@ internal class ClusterHandlerTests : JUnit5Minutests {
...
+//          ebsOptimized = false,
+//          iamRole = "keelRole",
+//          keyPair = "keel-key-pair",
+//          instanceMonitoring = false
+//        )
+//        val changedSpec = spec.copy(launchConfiguration = lc1)
+//        val diff = differ.compare(spec, changedSpec)
+//        expectThat(diff.hasChanges()).isEqualTo(true)
+//      }
+//    }
```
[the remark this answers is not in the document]
emjburns: Ya, that would make sense. I'll remove these tests here, and when we fix diffing we can add a comprehensive test around it.

emjburns:
Fixes #297 

I think the overall question of how do we know something used to be declaratively managed but now we don't want it to exist is interesting, but shouldn't be solved with tags. I think that an empty resource might be a better solution, but I vote we solve that later.

This PR tags the actual resource submitted - so `cluster`, `securitygroup` and `loadbalancer`. We don't need to worry about the server group and carrying the tag forward.

This PR also won't add tags to anything that is already managed by keel - meaning we will need to destroy and re-create the resources. I've also updated this PR to just use create and delete events, and to sweep for old tags (tag resources in `TagNotDesired` state) every day (configurable) and remove them from our database.

@cfieber @robfletcher care to take another look?

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/KeelTagHandler.kt
```diff
@@ -0,0 +1,189 @@
...
+
+  override fun generateName(spec: KeelTagSpec) = ResourceName("tag:keel-tag:${spec.keelId}")
+
+  override suspend fun desired(resource: Resource<KeelTagSpec>): TaggedResource {
+    when (resource.spec.tagState) {
+      is TagDesired -> {
+        val desiredTag = (resource.spec.tagState as TagDesired).tag
+        return TaggedResource(resource.spec.keelId, resource.spec.entityRef, desiredTag)
+      }
+      else -> {
```
robfletcher: If `TagState` is a sealed class you should be able to specify both options and not need an else.

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -0,0 +1,188 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  @Scheduled(fixedDelayString = "\${keel.resource-tagger.account-sync-frequency:PT600S}")
+  private fun syncAccounts() {
+    val now = LocalDateTime.now().toEpochSecond(ZoneOffset.UTC)
+    if (now - accountsUpdateFrequencyS > accountsUpdateTimeS) {
```
robfletcher: Is there any need to check this given the method is called on a schedule by Spring?

emjburns: How would I test that?

robfletcher: It should just always sync the accounts when the method is called, no?

PR #322, opened 2019-06-17
use latest spinnaker gradle plugins

robfletcher:


robfletcher:
@cfieber Not sure if this is why Keel is not getting autobump PRs (it's in the list in spinnaker/spinnaker-gradle-plugin).

emjburns:
LGTM

PR #326, opened 2019-06-19
fix(tagging): only tag real resources

emjburns:
We only want to tag resources that we create, not things that we use internally like images or tags.


keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -67,9 +67,11 @@ class ResourceTagger(
     "ec2" to "aws"
   )
 
+  private val taggableResources = listOf("cluster", "securityGroup", "clb")
```
emjburns: @asher you'll need to add new resources here if you want them tagged

PR #329, opened 2019-06-19
Detect change in sub-type of polymorphic properties

robfletcher:
Also encapsulate use of java-object-diff library better.

The `java-object-diff` library internally uses several different types of `Differ`, one for primitives, one for collections, one for maps, and one for "beans". That last one is missing the logic that detects that the base and working states of a property are of different types. I've written a custom differ that delegates to the `BeanDiffer` for all other cases but marks this case as `CHANGED`. I think this is sufficient for what we want (it will at least detect there is a diff). It could probably work a little better so that it attaches the sub-tree of added/removed properties but I'm not sure that's actually desperately important.

I'm going to work this up into a PR for the original library but since that has had no commit activity in 18 months I'm not expecting much response.


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -55,7 +48,7 @@ class ResourceActuator(
       val resource = resourceRepository.get(name, type)
 
       val (desired, current) = plugin.resolve(resource)
-      val diff = differ.compare(desired, current)
+      val diff = ResourceDiff(desired, current)
```
robfletcher: Instead of using the diff library directly we just use `ResourceDiff` everywhere

keel-core/keel-core.gradle.kts
```diff
@@ -24,9 +24,10 @@ dependencies {
   api("com.fasterxml.jackson.module:jackson-module-kotlin")
   api("com.fasterxml.jackson.datatype:jackson-datatype-jsr310")
   api("de.huxhorn.sulky:de.huxhorn.sulky.ulid")
-  api("de.danielbechler:java-object-diff")
   api("com.netflix.spinnaker.kork:kork-artifacts")
 
+  implementation("de.danielbechler:java-object-diff")
```
robfletcher: Made this not get exported so other modules can't directly reference types from it

keel-tagging-plugin/src/test/kotlin/com/netflix/spinnaker/keel/tagging/KeelTagSpecTests.kt
```diff
@@ -0,0 +1,48 @@
...
+            )
+          )
+        )
+        ResourceDiff(
+          resource,
+          resource.copy(tagState = TagNotDesired(0L))
+        )
+      }
+
+      SKIP - test("a diff is detected") {
```
robfletcher: Test skipped until #317 is fixed

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/diff/ResourceDiffTests.kt
```diff
@@ -47,13 +54,13 @@ internal class DiffTests : JUnit5Minutests {
...
     }
 
-    context("same parent two children") {
-//      test("diffing works") {
-//        val obj1: Parent = Child1("common", "myprop")
-//        val obj2: Parent = Child2("common", "anotherProp")
-//        val resourceDiff = ResourceDiff(obj1, obj2)
-//        expectThat(resourceDiff.diff.hasChanges()).isEqualTo(true)
-//      }
+    SKIP - context("delta is the type of a polymorphic property") {
```
robfletcher: Test skipped until #317 is fixed

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/diff/ResourceDiff.kt
```diff
@@ -0,0 +1,75 @@
...
+    when (this) {
+      null -> null
+      else -> mapper.convertValue(this)
+    }
+
+  companion object {
+    val differ: ObjectDiffer = ObjectDifferBuilder
+      .startBuilding()
+      .apply {
+        //        differs().register(PolymorphismAwareDifferFactory)
```
emjburns: Needed?

emjburns: If it's a reference to that todo below, fine with leaving it

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/diff/ResourceDiffTests.kt
emjburns: woah I didn't know that this was a thing you could do! cool.

Also, shouldn't this PR fix #317 ?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/diff/ResourceDiff.kt
robfletcher: Working on it in subsequent commit

keel-core/keel-core.gradle.kts
robfletcher: Actually can't do this (yet)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/diff/ResourceDiff.kt
```diff
@@ -0,0 +1,111 @@
...
+    when (this) {
+      null -> null
+      else -> mapper.convertValue(this)
+    }
+
+  companion object {
+    val differ: ObjectDiffer = ObjectDifferBuilder
+      .startBuilding()
+      .apply {
+        differs().register(PolymorphismAwareDifferFactory(this))
```
robfletcher: Here I'm injecting a custom differ that handles objects that are not the same type, delegating to the regular differ in all other cases

keel-tagging-plugin/src/test/kotlin/com/netflix/spinnaker/keel/tagging/KeelTagSpecTests.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.netflix.spinnaker.keel.tags.EntityTag
+import com.netflix.spinnaker.keel.tags.KEEL_TAG_NAME
+import com.netflix.spinnaker.keel.tags.TagValue
+import de.danielbechler.diff.node.DiffNode.State.CHANGED
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import strikt.api.expectThat
+import strikt.assertions.isEqualTo
+
+internal class KeelTagSpecTests : JUnit5Minutests {
```
robfletcher: Proves this works in the specific case of `KeelTagSpec`

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/diff/ResourceDiffTests.kt
```diff
@@ -15,45 +15,44 @@
...
-    context("same parent two children") {
-//      test("diffing works") {
-//        val obj1: Parent = Child1("common", "myprop")
-//        val obj2: Parent = Child2("common", "anotherProp")
-//        val resourceDiff = ResourceDiff(obj1, obj2)
-//        expectThat(resourceDiff.diff.hasChanges()).isEqualTo(true)
-//      }
+  fun tests() = rootContext {
+    context("delta is the type of a polymorphic property") {
+      test("the delta is detected") {
```
robfletcher: Proves this works in the general case

PR #331, opened 2019-06-20
feat(ec2): ApplicationLoadBalancer support

asher:


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
```diff
@@ -0,0 +1,211 @@
...
+  override val objectMapper: ObjectMapper,
+  override val normalizers: List<ResourceNormalizer<*>>
+) : ResourceHandler<ApplicationLoadBalancer> {
+  override val log: Logger by lazy { LoggerFactory.getLogger(javaClass) }
+
+  override val apiVersion = SPINNAKER_API_V1.subApi("ec2")
+  override val supportedKind = ResourceKind(
+    apiVersion.group,
+    "alb",
+    "albs"
```
robfletcher: Pedantic but would it make sense to use something more verbose here? We currently have `cluster`, `security-group` and `image`. Is `application-load-balancer` too long? I like the clarity myself.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/normalizers/ApplicationLoadBalancerNormalizer.kt
```diff
@@ -0,0 +1,75 @@
...
+) : ResourceNormalizer<ApplicationLoadBalancer> {
+  override val apiVersion = SPINNAKER_API_V1.subApi("ec2")
+  override val supportedKind = "alb"
+
+  override fun normalize(resource: Resource<*>): Resource<ApplicationLoadBalancer> {
+    @Suppress("UNCHECKED_CAST")
+    val r = resource as Resource<ApplicationLoadBalancer>
+
+    if (r.spec.targetGroups.any { it.name.length > 32 }) {
+      throw IllegalStateException("targetGroup names have a 32 character limit")
```
robfletcher: We definitely need to think about how to model this kind of thing in the schema

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
asher: I like keeping it pithy but I'm fine with verbose, will update `clb` to `classic-load-balancer` as well.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/normalizers/ApplicationLoadBalancerNormalizer.kt
asher: Looks like precondition checks can be added to a data class' init block, going to test that out.

robfletcher: I wasn't suggesting you need to do it as part of this PR, just thinking aloud.

PR #332, opened 2019-06-20
feat(auth): add required service account for all resources

emjburns:
* Adding fiat to keel.
* Adding a required new field when submitting resources, `metadata`, which must contain a `securityAccount`. Is this good? Give me opinions. I'm not sure if this should be required.
* Checking when you create/update/delete if you (`X-SPINNAKER-USER`) have access to the service account you specify.
* Adding this service account to the normal metadata for all resources.

TODO: use that service account for actuation.

### DIFF
--- a/keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
+++ b/keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
@@ -68,7 +68,8 @@ internal class ResourceActuatorTests : JUnit5Minutests {
         kind = "foo",
         metadata = mapOf(
           "name" to "resource1",
-          "uid" to randomUID()
+          "uid" to randomUID(),
+          "serviceAccount" to "keel@spinnaker"
         ),
         spec = DummyResourceSpec("whatever")
       )
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
@@ -1,6 +1,8 @@
 package com.netflix.spinnaker.config
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.fiat.shared.EnableFiatAutoConfig
+import com.netflix.spinnaker.filters.AuthenticatedRequestFilter
 import com.netflix.spinnaker.keel.persistence.ArtifactRepository
 import com.netflix.spinnaker.keel.persistence.ResourceRepository
 import com.netflix.spinnaker.keel.persistence.ResourceVersionTracker
@@ -12,15 +14,14 @@ import com.netflix.spinnaker.keel.serialization.configuredObjectMapper
 import de.huxhorn.sulky.ulid.ULID
 import org.springframework.beans.factory.config.ConfigurableBeanFactory.SCOPE_PROTOTYPE
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
+import org.springframework.boot.web.servlet.FilterRegistrationBean
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.context.annotation.Scope
-import org.springframework.core.Ordered.HIGHEST_PRECEDENCE
-import org.springframework.core.annotation.Order
-import org.springframework.security.config.annotation.web.builders.HttpSecurity
-import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter
+import org.springframework.core.Ordered
 import java.time.Clock
 
+@EnableFiatAutoConfig
 @Configuration
 class DefaultConfiguration {
   @Bean
@@ -52,9 +53,9 @@ class DefaultConfiguration {
   fun noResourcePlugins(): List<ResolvableResourceHandler<*, *>> = emptyList()
 
   @Bean
-  fun csrfDisable() = @Order(HIGHEST_PRECEDENCE) object : WebSecurityConfigurerAdapter() {
-    override fun configure(http: HttpSecurity) {
-      http.csrf().disable()
-    }
+  fun authenticatedRequestFilter(): FilterRegistrationBean<AuthenticatedRequestFilter> {
+    val frb = FilterRegistrationBean(AuthenticatedRequestFilter(true))
+    frb.order = Ordered.HIGHEST_PRECEDENCE
+    return frb
   }
 }
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
@@ -1,0 +1,52 @@
+/*
+ *
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.rest
+
+import com.netflix.spinnaker.fiat.shared.FiatPermissionEvaluator
+import com.netflix.spinnaker.keel.api.ResourceName
+import com.netflix.spinnaker.keel.api.serviceAccount
+import com.netflix.spinnaker.keel.persistence.NoSuchResourceException
+import com.netflix.spinnaker.keel.persistence.ResourceRepository
+import org.springframework.security.core.Authentication
+import org.springframework.security.core.context.SecurityContextHolder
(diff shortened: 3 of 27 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
```diff
@@ -0,0 +1,51 @@
...
+  private val resourceRepository: ResourceRepository
+) {
+  fun userCanModifyResource(name: String): Boolean {
+    try {
+      val resource = resourceRepository.get(ResourceName(name), Any::class.java)
+      return userCanModifySpec(resource.metadata.serviceAccount)
+    } catch (e: NoSuchResourceException) {
+      // todo eb: should we say "unauthorized" if the resource doesn't exist?
+      // or, should we let someone see that it doesn't exist, because no action would happen?
+      return false
```
emjburns: Please weigh in

keel-api/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
```diff
@@ -52,9 +53,9 @@ class DefaultConfiguration {
   fun noResourcePlugins(): List<ResolvableResourceHandler<*, *>> = emptyList()
 
   @Bean
```
emjburns: `csrfDisable()` is not needed and was making it so that fiat never got a chance to handle requests.
`authenticatedRequestFilter()` adds the Spinnaker filter we need to correctly parse the headers like `X-SPINNAKER-USER` and add them to the `mdc`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
robfletcher: I vote for unauthorized

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
```diff
@@ -30,9 +30,9 @@ data class Resource<T : Any>(
 
 /**
  * External representation of a resource that would be submitted to the API
- * It doesn't need to contain metadata
  */
 data class SubmittedResource<T : Any>(
+  val metadata: SubmittedMetadata,
```
emjburns: Should this be required of all resources? Should we allow no service account? I'm inclined to say we should require service account since we will always be doing things on behalf of a user.

keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -68,7 +68,8 @@ internal class ResourceActuatorTests : JUnit5Minutests {
         kind = "foo",
         metadata = ResourceMetadata(
           name = ResourceName("resource1"),
-          uid = randomUID()
+          uid = randomUID(),
+          serviceAccount = "keel@spinnaker"
```
robfletcher: I like that this is in metadata

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceMetadata.kt
```diff
@@ -40,3 +42,7 @@ data class ResourceMetadata(
     ) + data)
       .toString()
 }
+
+data class SubmittedMetadata(
+  val serviceAccount: String
```
robfletcher: I feel like this is one case where a good old `Map<String, Any>` might actually be appropriate

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
```diff
@@ -75,13 +80,16 @@ internal class ResourceControllerTests {
   @Test
   fun `can create a resource as YAML`() {
     every { resourcePersister.create(any()) } returns resource
+    every { authorizationSupport.userCanModifySpec("keel@spinnaker") } returns true
```
robfletcher: Are there equivalent tests for where this would return `false`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceMetadata.kt
```diff
@@ -22,6 +22,7 @@ import de.huxhorn.sulky.ulid.ULID
 data class ResourceMetadata(
   val name: ResourceName,
   val uid: UID,
+  val serviceAccount: String,
```
robfletcher: Is this OSS compatible?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -185,6 +188,7 @@ class SqlResourceRepository(
 
     private val UID = field<String>("uid")
     private val API_VERSION = field<String>("api_version")
+    private val SERVICE_ACCOUNT = field<String>("service_account")
```
robfletcher: Does this need its own column in the database? Will we ever be indexing on this?

keel-tagging-plugin/src/test/kotlin/com/netflix/spinnaker/keel/tagging/KeelTagHandlerTests.kt
```diff
@@ -109,7 +109,8 @@ internal class KeelTagHandlerTests : JUnit5Minutests {
     "keel-tag",
     ResourceMetadata(
       name = ResourceName(keelId),
-      uid = randomUID()
+      uid = randomUID(),
+      serviceAccount = "keel@spinaker"
```
robfletcher: I approve of the @HORSE_SPINAKER spelling

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceMetadata.kt
emjburns: Ya, this is just good old fiat service accounts.

keel-tagging-plugin/src/test/kotlin/com/netflix/spinnaker/keel/tagging/KeelTagHandlerTests.kt
emjburns: 😆 oops

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
robfletcher: I think so, at least for now.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
emjburns: Removing!

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
emjburns: I'll add one.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceMetadata.kt
emjburns: It's easier to do auth checking when this is an actual class with actual values. I am having trouble getting it to work with a map, so I'm just gonna leave it as is for now.

emjburns:
The last thing I have left to do on this PR is to figure out how to migrate existing things we have saved in the database.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
asher: Could be confusing if unauthorized is returned when calling /delete on an already deleted resource though? That's the only controller method using `userCanModifyResource` directly.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
```diff
@@ -30,6 +30,7 @@ data class Resource<T : Any>(
     require(kind.isNotEmpty()) { "resource kind must be defined" }
     require(metadata["uid"].isValidULID()) { "resource uid must be a valid ULID" }
     require(metadata["name"].isValidName()) { "resource name must be a valid name" }
+    require(metadata["serviceAccount"].isValidServiceAccount()) { "serviceAccount must be a valid email" }
```
asher: Fiat doesn't require service account names to be in the form of an email address. We have at least one that isn't internally. I think this can just be dropped. Since ```@PreAuthorize("@authorizationSupport.userCanModifySpec(#submittedResource.metadata.serviceAccount)")``` is on the create controller, that should guarantee that  `serviceAccount` is valid according to fiat regardless of formatting.

emjburns: Oo good point. I'll update.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
emjburns: Hmm. Good point. I think a 404 might be more expected here.

robfletcher: We should 100% say 404 if it doesn't exist

PR #335, opened 2019-06-21
Proposal: use a map for metadata rather than a dedicated type

robfletcher:
I think this is simpler and allows more convenient access to the `uid` and `name` properties (via extension methods on `Resource`). As we add more things to the metadata I can see that class becoming super unwieldy, especially in tests.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
```diff
@@ -37,3 +39,9 @@ data class SubmittedResource<T : Any>(
   val kind: String,
   val spec: T
 )
+
+val <T : Any> Resource<T>.uid: UID
+  get() = metadata.getValue("uid").toString().let(ULID::parseULID)
```
cfieber: would it be worth adding the `set()` equivalents on these as some helper to keep the metatdata String key from propagating everywhere and ensuring the correct types land in the map?

robfletcher: I think I'd be down with them being strings in the map. It makes the Jackson serialization story easier.

robfletcher: I agree that the concern with this change is the loss of type safety, though. I think the use of extension properties mitigates that enough. The `name` and `uid` metadata values should only ever be set once (when the resource is first persisted) so I don't know that we need a setter.

cfieber: I saw a few spots but only in tests where we are populating the map which is what led me to ask but I agree

Issue #223, opened 2019-03-02
Prevent creation of duplicate resources


robfletcher:
If a resource already exists with the name we generate for a new one we should respond with an error.

luispollo:
@robfletcher I thought this was fixed recently? Maybe it was @asher or @emjburns?

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

luispollo:
@robfletcher @emjburns Ping?

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

PR #339, opened 2019-06-25
feat(create): if resource exists, second create is an update

emjburns:
This will make sure that we don't get mismatching uids

### DIFF
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
@@ -9,6 +9,7 @@ import com.netflix.spinnaker.keel.events.CreateEvent
 import com.netflix.spinnaker.keel.events.DeleteEvent
 import com.netflix.spinnaker.keel.events.ResourceCreated
 import com.netflix.spinnaker.keel.events.ResourceUpdated
+import com.netflix.spinnaker.keel.persistence.NoSuchResourceException
 import com.netflix.spinnaker.keel.persistence.ResourceRepository
 import com.netflix.spinnaker.keel.persistence.get
 import com.netflix.spinnaker.keel.plugin.ResolvableResourceHandler
@@ -29,9 +30,16 @@ class ResourcePersister(
     handlers.supporting(resource.apiVersion, resource.kind)
       .normalize(resource)
       .also {
-        resourceRepository.store(it)
-        resourceRepository.appendHistory(ResourceCreated(it, clock))
-        publisher.publishEvent(CreateEvent(it.name))
+        try {
+          // If resource exists, we should update instead
+          val existing = resourceRepository.get(it.name, Any::class.java)
+          return update(existing.name, resource)
+        } catch (e: NoSuchResourceException) {
+          resourceRepository.store(it)
+          resourceRepository.appendHistory(ResourceCreated(it, clock))
+          publisher.publishEvent(CreateEvent(it.name))
+        }
+        return it
       }
 
   fun update(name: ResourceName, updated: SubmittedResource<Any>): Resource<out Any> {
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
@@ -26,23 +26,29 @@ class ResourcePersister(
   private val clock: Clock,
   private val publisher: ApplicationEventPublisher
 ) {
+  fun upsert(resource: SubmittedResource<Any>): Resource<out Any> =
+    handlers.supporting(resource.apiVersion, resource.kind)
+      .normalize(resource)
+      .also {
+        if (it.name.isRegistered()) {
+          return update(it.name, resource)
+        } else {
+          return create(resource)
+        }
+      }
+
   fun create(resource: SubmittedResource<Any>): Resource<out Any> =
     handlers.supporting(resource.apiVersion, resource.kind)
       .normalize(resource)
       .also {
-        try {
-          // If resource exists, we should update instead
-          val existing = resourceRepository.get(it.name, Any::class.java)
-          return update(existing.name, resource)
-        } catch (e: NoSuchResourceException) {
-          resourceRepository.store(it)
-          resourceRepository.appendHistory(ResourceCreated(it, clock))
-          publisher.publishEvent(CreateEvent(it.name))
-        }
-        return it
+        log.debug("Creating $it")
+        resourceRepository.store(it)
+        resourceRepository.appendHistory(ResourceCreated(it, clock))
+        publisher.publishEvent(CreateEvent(it.name))
       }
 
   fun update(name: ResourceName, updated: SubmittedResource<Any>): Resource<out Any> {
+    log.debug("Updating $name")
     val handler = handlers.supporting(updated.apiVersion, updated.kind)
     val existing = resourceRepository.get(name, Any::class.java)
     val resource = existing.copy(spec = updated.spec)
@@ -70,6 +76,15 @@ class ResourcePersister(
         resourceRepository.delete(name)
         publisher.publishEvent(DeleteEvent(name))
       }
+
+  private fun ResourceName.isRegistered(): Boolean {
+    try {
+      resourceRepository.get(this, Any::class.java)
+      return true
+    } catch (e: NoSuchResourceException) {
+      return false
+    }
+  }
 
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 }
--- a/keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
+++ b/keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
@@ -284,6 +284,12 @@ internal class ResourceActuatorTests : JUnit5Minutests {
 private fun <T : Iterable<E>, E> Assertion.Builder<T>.second(): Assertion.Builder<E> =
   get("second element %s") { toList()[1] }
 
-internal data class DummyResource(val state: String)
+internal data class DummyResource(
+  val state: String,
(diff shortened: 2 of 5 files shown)


robfletcher:
Does this fix #223?

robfletcher:
If we assume a duplicate create is an update does that risk users overwriting each others’ resource definitions?

emjburns:
I don't know if it "fixes" that issue. If we want to reject a second create, then this PR is wrong. Maybe it's better to create a separate upsert method, which will either create or update? I'm torn on what the right thing to do is.

cfieber:
also torn-ish, what is the use case where you are seeing duplicate updates?

robfletcher:
I think the _controller_ should respond with HTTP 409 if the resource (with a clashing name) already exsists. I'm open to the `store` method on the repository being an upsert.

robfletcher:
Although we may have further complications via delivery config. What should be the behavior if 2 different delivery config manifests specify clashing resources?

asher:
How about replacing the `/resources/{update, create}` controller methods with an `upsert` and refactoring `ResourcePersister` accordingly? Thinking of a gitops workflow where resource defs are persisted to keel via a post commit hook or CI system, this should be achievable with dead simple glue code that has no knowledge or concern of whether a resource was previously persisted.

robfletcher:
But what happens if 2 users define conflicting definitions of the same resource?

emjburns:
Do you think that two people with the same service account membership would often "stomp" each other in this way? Maybe it would be enough to let them know that we did an update instead of a create, and also offer some way to disregard an update / roll back to a previous version. I agree with Asher, it should be really easy to post to keel from CI, and having to do _either_ a create or update is problematic for that.

asher:
I think requiring separate create/update methods only prevents accidental resource conflicts with users who are aware of and actively looking to catch occurrences. As in, I can imagine users scripting around the distinction and automatically calling create then update on every resource file to avoid complexity, run via a CI system that is rarely viewed by humans. The problem is real, but not sure if we can do better than assert user/group ACLs against resources.

robfletcher:
Yeah, I think Asher's point is a good one. Maybe we need a different way to prevent multiple users defining the same thing. I _do_ think that's a concern for a couple of reasons:

1. I can see naïve users defining shared resources (e.g. `nf-infrastructure` security group) particularly when mapping from an imperatively managed app.
2. It could be hellishly hard to debug.

emjburns:
@asher @robfletcher @cfieber Changing so that the controller and resource persister has upsert instead of create/update to support git-ops workflows.

I'll keep thinking about how to support someone accidentally stomping on a resource.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
```diff
@@ -25,6 +26,18 @@ class ResourcePersister(
...
 ) {
+  fun upsert(resource: SubmittedResource<Any>): Resource<out Any> =
+    handlers.supporting(resource.apiVersion, resource.kind)
+      .normalize(resource)
+      .also {
+        try {
+          val existing = resourceRepository.get(it.name, Any::class.java)
+          return update(existing.name, resource)
+        } catch (e: NoSuchResourceException) {
+          return create(resource)
```
robfletcher: Does it make sense to propagate the upsert concept all the way to the `ResourceRepository`? Alternatively having an `isThisRegistered(name) : Boolean` method. Catching exceptions feels bad.

keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersisterTests.kt
```diff
@@ -111,12 +111,17 @@ internal class ResourcePersisterTests : JUnit5Minutests {
         context("after an update") {
           before {
             resourcesDueForCheck()
-            update(DummyResourceSpec("kthxbye"))
+            subject.upsert(SubmittedResource(
+              metadata = SubmittedMetadata("keel@spinnaker"),
```
robfletcher: I think `SubmittedMetadata` can be done away with now the `Resource` class just uses a `Map<String, Any?>`.

keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -284,6 +284,12 @@ internal class ResourceActuatorTests : JUnit5Minutests {
 private fun <T : Iterable<E>, E> Assertion.Builder<T>.second(): Assertion.Builder<E> =
   get("second element %s") { toList()[1] }
 
-internal data class DummyResource(val state: String)
-
-internal data class DummyResourceSpec(val state: String)
+internal data class DummyResource(
+  val state: String,
+  val data: String = "some data"
+)
```
robfletcher: Curious what the value is of the additional field here?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
```diff
@@ -75,14 +61,13 @@ class ResourceController(
...
     consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
     produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
   )
   @PreAuthorize("@authorizationSupport.userCanModifySpec(#resource.metadata.serviceAccount)")
-  fun update(@PathVariable("name") name: ResourceName, @RequestBody resource: SubmittedResource<Any>): Resource<out Any> {
-    log.debug("Updating: $resource")
-    return resourcePersister.update(name, resource)
+  fun upsert(@RequestBody resource: SubmittedResource<Any>): Resource<out Any> {
+    log.debug("Upserting: $resource")
+    return resourcePersister.upsert(resource)
```
robfletcher: Would it make sense to distinguish between a `200` and `201` response? Not really a big deal but it might help us debug things.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
```diff
@@ -75,14 +61,13 @@ class ResourceController(
   }
 
   @PutMapping(
```
robfletcher: Not sure what the semantically correct HTTP verb is for an upsert but I suspect `POST` might be more appropriate. `PUT` is specifically for updating an existing thing IIRC.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
```diff
@@ -75,14 +61,13 @@ class ResourceController(
...
-    path = ["/{name}"],
```
robfletcher: Specifically I think `PUT` would be right if the user was specifying the name (i.e. the full URL of the resource's `GET`)

keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
emjburns: with this dummy resource we calculate name based on `state`. When you change the state, it changes the name, which now will just persist a new resource. So, I had to add another field so that we could properly test an update situation.

robfletcher: Gotcha. That makes sense.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
emjburns: (as discussed offline) I'll do a method for now.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
emjburns: How would I pass that back to the controller? I'm happy to do that.

keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersisterTests.kt
emjburns: I had some trouble getting the auth function to work with a map. Overall, it seemed simpler to just have this data class.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
robfletcher: Yeah, that's a good question.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
emjburns: I think it should be `PUT`: https://stackoverflow.com/questions/18470588/in-rest-is-post-or-put-best-suited-for-upsert-operation

robfletcher: I read the same thing, but specifically this jumped out:
> provided that in both cases the URI is known in advance
which is not the case here

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
emjburns: I'll add a log statement for now.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
emjburns: 👍

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
emjburns: 👍

PR #346, opened 2019-07-19
Persist delivery config and associated things transactionally

robfletcher:
… at least when using SQL.


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -10,39 +10,40 @@ import com.netflix.spinnaker.keel.persistence.metamodel.Tables.DELIVERY_ARTIFACT
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.DELIVERY_ARTIFACT_VERSION
 import org.jooq.DSLContext
 import org.slf4j.LoggerFactory
+import org.springframework.transaction.annotation.Propagation.REQUIRED
+import org.springframework.transaction.annotation.Transactional
 
-class SqlArtifactRepository(
+open class SqlArtifactRepository(
```
robfletcher: Still trying to figure out why I had to do this when the `kotlin-spring` plugin is applied and it works in `ResourcePersister`.

robfletcher: Seems to be something to do with the fact the annotations are on methods defined in the interface.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
```diff
@@ -17,27 +23,53 @@ import com.netflix.spinnaker.keel.plugin.supporting
...
+  private val deliveryConfigRepository: DeliveryConfigRepository,
+  private val artifactRepository: ArtifactRepository,
   private val resourceRepository: ResourceRepository,
   private val handlers: List<ResolvableResourceHandler<*, *>>,
   private val clock: Clock,
   private val publisher: ApplicationEventPublisher
 ) {
-  fun upsert(resource: SubmittedResource<Any>): Resource<out Any> =
+  @Transactional(propagation = REQUIRED)
+  fun upsert(deliveryConfig: SubmittedDeliveryConfig): DeliveryConfig =
```
robfletcher: This is the key method that needs to be transactional.

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/tx/DeliveryConfigTransactionTests.kt
```diff
@@ -0,0 +1,229 @@
...
+  webEnvironment = MOCK,
+  properties = [
+    "sql.enabled=true",
+    "sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename",
+    "sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename",
+    "spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver",
+    "eureka.default-to-up=false"
+  ]
+)
+internal class DeliveryConfigTransactionTests : JUnit5Minutests {
```
robfletcher: Here are a bunch of tests to make sure delivery config is persisted all or nothing

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/tx/DeliveryConfigTransactionTests.kt
```diff
@@ -0,0 +1,229 @@
...
+
+  @SpykBean
+  lateinit var artifactRepository: ArtifactRepository
+
+  @SpykBean
+  lateinit var resourceRepository: ResourceRepository
+
+  @SpykBean
```
robfletcher: First time I've ever really used spies. I want the persistence to actually work, but also to be able to force exceptions at certain places when I want to

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
asher: `SqlDeliveryConfigRepository.store()` doesn't appear to be transaction though, so I feel like I'm missing something?

robfletcher: It doesn't need to be to operate in the transaction that this method creates.

asher: I missed that spring-tx is managing transaction boundaries here 👍

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -29,43 +29,35 @@ class SqlDeliveryConfigRepository(
...
-            .set(ENVIRONMENT.DELIVERY_CONFIG_UID, uid)
-            .set(ENVIRONMENT.NAME, environment.name)
+      }
+      environments.forEach { environment ->
+        val environmentUID = randomUID().toString()
+        jooq.insertInto(ENVIRONMENT)
+          .set(ENVIRONMENT.UID, environmentUID)
+          .set(ENVIRONMENT.DELIVERY_CONFIG_UID, uid)
+          .set(ENVIRONMENT.NAME, environment.name)
+          .onDuplicateKeyIgnore()
```
asher: Not sure about environments always getting a new uid here. This store method is called for delivery config upsert, so it seems like the environment table is likely to get lots of "duplicate" rows, where `delivery_config_uid` and `name` are the same, but with a new `uid` per config upsert.

robfletcher:
Good spot, yeah. Will fix.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -132,30 +143,48 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
...
+          storeArtifacts()
+          storeResources()
+          store()
+          store()
+        }
+
+        test("the environments are not duplicated") {
+          getByName()
+            .succeeded()
+            .get { environments }.hasSize(deliveryConfig.environments.size)
```
robfletcher: Test to ensure environments are not re-created

PR #348, opened 2019-07-22
Vetos and allowing applications to opt out

emjburns:
Created a new "veto" thing that vetos applications from being checked.

I've created a general veto enforcer, which loops through all vetos that are registered and checks to see if any of them vetos the resource. A deny always overrides an allowed (if not, it's really hard to tell who has the 'authority' to deny or allow a resource, and who overrides another resource).

I've created a veto controller which allows you to pass messages/state to the policies (like opting out an application using the `ApplicationVeto`). 

To opt out an application you'll do:
```
POST /vetos/ApplicationVeto

{
    "application": "keeldemo",
    "optedOut": true
  }
```

I'd love feedback on whether this is a good approach. I still need to implement a real SQL backing for the `ApplicationVeto` but I wanted to get general feedback on my approach first.


keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -276,6 +281,36 @@ internal class ResourceActuatorTests : JUnit5Minutests {
...
+              "serviceAccount" to "keel@spinnaker"
+            ),
+            spec = DummyResourceSpec("badapp")
+          )
+          before {
+            resourceRepository.store(badresource)
+            resourceRepository.appendHistory(ResourceCreated(badresource))
+            coEvery { plugin1.desired(badresource) } returns DummyResource(badresource.spec.state)
+            coEvery { plugin1.current(badresource) } returns DummyResource(badresource.spec.state)
+            coEvery { plugin1.update(badresource, any()) } returns listOf(TaskRef("/tasks/${randomUID()}"))
```
robfletcher: Why do you need to stub these if it's not going to check that resource? I would have thought it should not call them (might make a good additional test).

keel-policy/src/main/kotlin/com/netflix/spinnaker/keel/policy/PolicyController.kt
```diff
@@ -0,0 +1,84 @@
...
+@RestController
+@RequestMapping(path = ["/policies"])
+class PolicyController(
+  val policies: List<Policy>
+) {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  @GetMapping(
+    path = [""],
```
robfletcher: Can omit that when empty

keel-policy/src/main/kotlin/com/netflix/spinnaker/keel/policy/PolicyController.kt
```diff
@@ -0,0 +1,84 @@
...
+  fun getActivePolicies(): List<String> = policies.map { it.name() }
+
+  @GetMapping(
+    path = ["/{name}/message-format"],
+    produces = [MediaType.APPLICATION_JSON_VALUE]
+  )
+  fun getPolicyMessageFormat(@PathVariable name: String): Map<String, Any> {
+    val policy = policies.find { it.name().equals(name, true) }
+    if (policy == null) {
+      throw PolicyNotFoundException(name)
```
robfletcher: Is there a handler for this so it will respond with 404 rather than 500?

keel-policy/src/main/kotlin/com/netflix/spinnaker/keel/policy/PolicyController.kt
```diff
@@ -0,0 +1,84 @@
...
+    path = [""],
+    produces = [MediaType.APPLICATION_JSON_VALUE]
+  )
+  fun getActivePolicies(): List<String> = policies.map { it.name() }
+
+  @GetMapping(
+    path = ["/{name}/message-format"],
+    produces = [MediaType.APPLICATION_JSON_VALUE]
+  )
+  fun getPolicyMessageFormat(@PathVariable name: String): Map<String, Any> {
```
robfletcher: What's this endpoint for? I'm not understanding.

keel-policy/src/main/kotlin/com/netflix/spinnaker/keel/policy/PolicyEnforcer.kt
```diff
@@ -0,0 +1,47 @@
...
+@Component
+class PolicyEnforcer(
+  val policies: List<Policy>
+) {
+
+  fun canCheck(name: ResourceName): PolicyResponse {
+    policies.forEach { policy ->
+      val response = policy.check(name)
+      if (!response.allowed) {
+        return response
```
robfletcher: We're only getting the result of the first policy to veto the check. Would there be any value in collecting all of them so we can get a holistic picture of what's going on?

keel-policy/src/main/kotlin/com/netflix/spinnaker/keel/policy/exceptions/MalformedMessageException.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.policy.exceptions
+
+class MalformedMessageException(
```
robfletcher: Probably need a handler for this to set an appropriate HTTP status (400 looks like) if it causes a request to fail.

keel-policy/src/main/kotlin/com/netflix/spinnaker/keel/policy/PolicyEnforcer.kt
emjburns: Maybe, probably, but I'm going to leave that as a future feature.

keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
emjburns: Did this before I figured out what test I actually wanted to create. Good catch!

cfieber:
per our round the table discussion I like this as a veto concept!

emjburns:
Updated this to be a `veto` concept instead of using the word "policy". Logic is the same.

PR #350, opened 2019-07-22
Removed Redis implementations of things

robfletcher:
Our opinion is that Redis is not an ideal persistence target and we don't have any intention of using so trying to support it going forward is an unnecessary burden on our efforts.


emjburns:
LGTM!

PR #352, opened 2019-07-31
Apply constraints when selecting an artifact version for an environment

robfletcher:
Basically it works like this:

* environments may have 0…n constraints
* each `Constraint` model type has a corresponding `ConstraintEvaluator`
* environments are regularly checked and artifact versions that pass an environment’s constraints are marked as “approved” in `PromotionRepository`
* when we calculate the desired state of a cluster resource, we use the latest approved version for that environment (see `ImageResolver` with logic previously directly in `ClusterHandler`)
* after a cluster is actuated we mark the artifact version as “successfully deployed” in that environment
* there's a simple constraint that just checks "was this artifact version ever successfully deployed in environment _x_?"


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
```diff
@@ -0,0 +1,44 @@
+package com.netflix.spinnaker.keel.constraints
+
+import com.netflix.spinnaker.keel.api.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+import com.netflix.spinnaker.keel.api.DependsOnConstraint
+import com.netflix.spinnaker.keel.persistence.PromotionRepository
+import org.springframework.stereotype.Component
+
+@Component
+class DependsOnConstraintEvaluator(
```
robfletcher: This is kinda the simplest possible constraint. It just checks if an artifact version is already in another (named) environment. Not was ever there, just is there now.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/PromotionRepository.kt
```diff
@@ -0,0 +1,20 @@
+package com.netflix.spinnaker.keel.persistence
+
+import com.netflix.spinnaker.keel.api.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+
+interface PromotionRepository {
```
robfletcher: The idea here is that we can easily record what version of a particular artifact is currently in a particular environment. In theory we could also do this via CloudDriver but then we'd need to look at a particular resource. Still debating with myself whether that would actually be better.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -76,46 +67,9 @@ class ClusterHandler(
     "ec2:cluster:${spec.location.accountName}:${spec.location.region}:${spec.moniker.name}"
   )
 
-  private suspend fun resolveImageId(imageProvider: ImageProvider, region: String): String {
```
robfletcher: Pulled this out into the new `ImageResolver` class

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -161,6 +120,26 @@ class ClusterHandler(
       .getCorrelatedExecutions(name.value)
       .isNotEmpty()
 
+  override suspend fun onActuationComplete(resource: Resource<ClusterSpec>, current: Cluster) {
```
robfletcher: This method on `ClusterHandler` lets me update the artifact version known to exist in an environment via `PromotionRepository`. I'm reverse engineering the version from the image id in the current model.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -0,0 +1,89 @@
...
+    .associateBy { it.constraintType }
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  // TODO: we may (probably will) want to make the parameter type more generic
+  suspend fun resolveImageId(resource: Resource<ClusterSpec>): String {
+    when (val imageProvider = resource.spec.launchConfiguration.imageProvider) {
+      is IdImageProvider -> {
+        return imageProvider.imageId
+      }
```
robfletcher: Logic here unchanged

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -0,0 +1,89 @@
...
+        val artifactName = artifact.name
+        val account = dynamicConfigService.getConfig("images.default-account", "test")
+        val namedImage = cloudDriverService
+          .namedImages(artifactName, account)
+          .sortedWith(NamedImageComparator)
+          .let { images ->
+            if (images.isEmpty()) {
+              throw NoImageFound(artifactName)
+            } else if (deliveryConfig == null || environment == null || environment.constraints.isEmpty()) {
+              images.last()
```
robfletcher: If the resource is not part of an environment, or no constraints are on that environment, just use the latest image (as before)

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -0,0 +1,89 @@
...
+            } else {
+              images.lastOrNull { image ->
+                constraintEvaluators.all {
+                  it.value.canPromote(artifact, image.appVersion, deliveryConfig, environment.name)
+                }
+              }
```
robfletcher: … otherwise find the newest image that satisfies all constraints on the environment

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolverTests.kt
```diff
@@ -0,0 +1,269 @@
...
+import io.mockk.every
+import io.mockk.mockk
+import kotlinx.coroutines.runBlocking
+import strikt.api.expectCatching
+import strikt.api.expectThat
+import strikt.assertions.failed
+import strikt.assertions.isA
+import strikt.assertions.isEqualTo
+
+internal class ImageResolverTests : JUnit5Minutests {
```
robfletcher: There were no tests for this logic before and now it's complex enough to require them (part of the reason I extracted it out of `ClusterHandler` into its own class)

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResolvableResourceHandler.kt
```diff
@@ -173,6 +173,12 @@ interface ResolvableResourceHandler<S : Any, R : Any> : KeelPlugin {
    * @return `true` if this plugin is still busy running a previous actuation, `false` otherwise.
    */
   suspend fun actuationInProgress(name: ResourceName): Boolean = false
+
+  /**
+   * Called when a resource returns to a valid state. There is no guarantee this will only be called
+   * once. Default implementation is a no-op.
+   */
+  suspend fun onActuationComplete(resource: Resource<S>, current: R) {}
```
robfletcher: Hook we can call on an handler to do stuff when a resource returns to a valid state

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
cfieber: I feel like eventually we'll need to support that an artifact was in another environment even if it isn't still (but that it wasn't rolled back out of that environment...) Thinking of a rapidly changing test/staging env and a slower changing prod env, you might want to not necessarily be promoting the thing that is currently there (maybe there is still some validation in flight) but rather something that was successfully there. Once you factor in some sort of manual approval constraint and deployment windows as well I could see this being a common thing.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/PromotionRepository.kt
cfieber: re: comment above, I kind of like us tracking the states independently of what clouddriver says is currently there - gives us opportunities to also track whether an environment went into a bad state due to a change and whether an artifact was rolled back (or fixed foward possiblity). 

Not sure we could get that level of metadata on an artifact's lifecycle from clouddriver without essentially building a repository over there

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
robfletcher: Definitely. I think this is the smallest possible step in that direction.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
```diff
@@ -0,0 +1,44 @@
...
+      "No environment named $targetEnvironment exists in the configuration ${deliveryConfig.name}"
+    }
+
+    val constraint = target
+      .constraints
+      .filterIsInstance<DependsOnConstraint>()
+      .firstOrNull()
+
+    return if (constraint == null) {
+      true
```
robfletcher: I would like to genericize this behavior. If the environment doesn't _have_ my supported constraint type, I should just approve any version.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -43,25 +47,49 @@ class ResourceCheckScheduler(
   @Scheduled(fixedDelayString = "\${keel.resource-check.frequency:PT1S}")
   fun checkResources() {
     if (enabled) {
-      log.debug("Starting scheduled validation…")
-      publisher.publishEvent(ScheduledCheckStarting)
+      log.debug("Starting scheduled resource validation…")
+      publisher.publishEvent(ScheduledResourceCheckStarting)
```
robfletcher: Changed event name because there are scheduled checks for the environment now too

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -43,25 +47,49 @@ class ResourceCheckScheduler(
...
-      log.debug("Scheduled validation complete")
+      log.debug("Scheduled resource validation complete")
     } else {
-      log.debug("Scheduled validation disabled")
+      log.debug("Scheduled resource validation disabled")
     }
   }
 
+  @Scheduled(fixedDelayString = "\${keel.environment-check.frequency:PT1S}")
+  fun checkEnvironments() {
```
robfletcher: Scheduled environment check

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.stereotype.Component
+
+@Component
+class EnvironmentPromotionChecker(
+  private val artifactRepository: ArtifactRepository,
+  private val promotionRepository: PromotionRepository,
+  private val constraints: List<ConstraintEvaluator<*>>
+) {
+
+  suspend fun checkEnvironments(deliveryConfig: DeliveryConfig) {
```
robfletcher: Check for promotable artifacts in each environment

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
```diff
@@ -34,13 +34,6 @@ data class ResourceCheckSkipped(
   )
 }
 
-data class LockAttempt(
-  val success: Boolean
-) : TelemetryEvent()
-
-val LockAttemptSucceeded = LockAttempt(true)
-val LockAttemptFailed = LockAttempt(false)
-
```
robfletcher: We were no longer using these. Unrelated to this PR but thought I might as well tidy it up

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryPeriodicallyCheckedTests.kt
```diff
@@ -0,0 +1,27 @@
+package com.netflix.spinnaker.keel.persistence
+
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+import com.netflix.spinnaker.keel.api.Environment
+
+abstract class DeliveryConfigRepositoryPeriodicallyCheckedTests<S : DeliveryConfigRepository> :
```
robfletcher: These tests apply to both resource and delivery config repositories

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -225,111 +219,6 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
     }
   }
 
-  data class CheckLockFixture<T : ResourceRepository>(
```
robfletcher: Moved elsewhere (into the new tests that are common with `DeliveryConfigRepository`)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -191,17 +191,23 @@ data class ResourceDeltaResolved(
   override val apiVersion: ApiVersion,
   override val kind: String,
   override val name: String,
-  override val timestamp: Instant
+  override val timestamp: Instant,
+  val desired: Any,
+  val current: Any
```
robfletcher: Added these new fields to the event to make the new listener simpler

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/actuation/ArtifactPromotionListener.kt
```diff
@@ -0,0 +1,56 @@
...
+@ConditionalOnBean(CloudDriverService::class)
+class ArtifactPromotionListener(
+  private val deliveryConfigRepository: DeliveryConfigRepository,
+  private val promotionRepository: PromotionRepository,
+  private val cloudDriverService: CloudDriverService
+) {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  @EventListener(ResourceDeltaResolved::class)
+  fun onDeltaResolved(event: ResourceDeltaResolved) {
```
robfletcher: This listener is responsible for marking an artifact version as "has been deployed to environment x"

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
robfletcher: It does this now

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -43,25 +47,49 @@ class ResourceCheckScheduler(
...
+      publisher.publishEvent(ScheduledEnvironmentCheckStarting)
+
+      val job = launch {
+        deliveryConfigRepository
+          .launchForEachItem {
+            environmentPromotionChecker.checkEnvironments(it)
+          }
+      }
+
+      runBlocking { job.join() }
```
emjburns: This means we simultaneously check each resource then we wait for them all to finish, right?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryPromotionRepository.kt
```diff
@@ -0,0 +1,59 @@
+package com.netflix.spinnaker.keel.persistence.memory
+
+import com.netflix.spinnaker.keel.api.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+import com.netflix.spinnaker.keel.persistence.PromotionRepository
+
+class InMemoryPromotionRepository : PromotionRepository {
+
+  private val approvedVersions = mutableMapOf<Triple<DeliveryArtifact, DeliveryConfig, String>, String>()
+  private val deployedVersions = mutableMapOf<Triple<DeliveryArtifact, DeliveryConfig, String>, MutableList<String>>()
```
emjburns: is there a benefit to having this as a `Triple` as opposed to making a data class to hold these three related things together?

emjburns: It also seems like a lot of data to store (a duplicate of the whole delivery config). Maybe we need to generate ids from delivery configs and just store that? Can be a future optimization.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/actuation/ArtifactPromotionListener.kt
```diff
@@ -0,0 +1,56 @@
...
+        cloudDriverService.namedImages(deployedImage, desired.location.accountName, desired.location.region)
+      }
+      check(amis.size == 1) {
+        "Found ${amis.size} images for image id $deployedImage when 1 was expected"
+      }
+      val appVersion = amis.first().appVersion
+      val deliveryConfig = deliveryConfigRepository.deliveryConfigFor(event.uid)
+      val environment = deliveryConfigRepository.environmentFor(event.uid)
+      if (deliveryConfig == null || environment == null) {
+        log.warn("Resource ${event.name} is not part of a delivery environment")
```
emjburns: is this `warn` worthy?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/actuation/ArtifactPromotionListener.kt
emjburns: This is cool. Is this only handling "environment has only one cluster" right now?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -0,0 +1,82 @@
...
+          .namedImages(artifactName, account)
+          .sortedWith(NamedImageComparator)
+          .firstOrNull() ?: throw NoImageFound(artifactName)
+
+        log.info("Image found for {}: {}", artifactName, namedImage)
+
+        return namedImage.amis[resource.spec.location.region]?.first()
+          ?: throw NoImageFoundForRegion(artifactName, resource.spec.location.region) // todo eb: when are there multiple?
+      }
+//      is JenkinsImageProvider -> {
```
emjburns: do you want this to be commented out?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -57,7 +57,7 @@ class SqlArtifactRepository(
         .where(DELIVERY_ARTIFACT.UID.eq(DELIVERY_ARTIFACT_VERSION.DELIVERY_ARTIFACT_UID))
         .and(DELIVERY_ARTIFACT.NAME.eq(artifact.name))
         .and(DELIVERY_ARTIFACT.TYPE.eq(artifact.type.name))
-        .orderBy(DELIVERY_ARTIFACT_VERSION.VERSION.desc())
+        .orderBy(DELIVERY_ARTIFACT_VERSION.VERSION.desc()) // TODO: this is not going to work
```
emjburns: is this a future todo or something you wanted to address in this PR?

keel-sql/src/main/resources/db/changelog/20190726-artifact-versions-in-environments.yml
```diff
@@ -0,0 +1,61 @@
...
+              - column:
+                  name: approved_at
+                  type: timestamp
+                  constraints:
+                    nullable: true
+              - column:
+                  name: deployed_at
+                  type: timestamp
+                  constraints:
+                    nullable: true
```
emjburns: this needs the modify sql statement

keel-sql/src/main/resources/db/databaseChangeLog.yml
```diff
@@ -20,3 +20,12 @@ databaseChangeLog:
   - include:
       file: changelog/20190804-create-application-veto-table.yml
       relativeToChangelogFile: true
+  - include:
+      file: changelog/20190724-environment-constraints.yml
```
emjburns: worth renaming these changelog files to have today's date?

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
robfletcher: Yeah, because the `@Scheduled` annotation specifies a delay from when the function completes and when it will be called again, so here we want to wait until all the checks are complete. Otherwise subsequent checks would start to overlap each other.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryPromotionRepository.kt
robfletcher: Right now they don't have ids. I've been thinking about whether we want to add that but I figured it could be a future optimization. There are other issues; do we want `DeliveryConfig`, `Environment` and `Resource` to be all extensions of the same thing? Maybe they all need `metadata`?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/actuation/ArtifactPromotionListener.kt
robfletcher: I shouldn't happen I don't think. Would we ever have an artifact-y cluster that wasn't part of an environment?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
robfletcher: Damn, sorry. Meant to write some tests for this and forgot.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
robfletcher: Future. Semver strings are not alphabetically sortable so I don't think this is a great way to get "latest version first". It's kind of an orthogonal concern to this PR, though.

keel-sql/src/main/resources/db/changelog/20190726-artifact-versions-in-environments.yml
robfletcher: Fixing…

keel-sql/src/main/resources/db/databaseChangeLog.yml
robfletcher: Ok, I did

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
robfletcher: I put it back (still needs tests!)

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/actuation/ArtifactPromotionListener.kt
robfletcher: Nope, it should work for multiple clusters. It's listening for when each cluster is resolved so it will mark a version as deployed when the cluster is. There's currently no coordination of "artifact in cluster a needs to be version aligned with artifact in cluster b".

robfletcher: Not sure if there's a case for > 1 cluster using the same artifact. I'm not catering for that currently. I guess that _can_ be a thing (e.g. we have special purpose stacks). Need to give that some thought.

robfletcher: I think that might be handled by the notion of environments being "stable" (i.e. all resources being in line with desired states). I decided to handle that separately as it needs us to start persisting the current state a little better than the current `resource_event` table. I think this PR is fit for purpose for simpler use-cases without it.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
emjburns: 👍 just making sure I understood the coroutines `join()` bit

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryPromotionRepository.kt
emjburns: hmm okay. seems like a good thing to come back to in the future if it's a problem.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/actuation/ArtifactPromotionListener.kt
emjburns: Ok sounds good! This all makes sense

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/actuation/ArtifactPromotionListener.kt
emjburns: How does this interact with declarative infra (so just having a cluster no environment) but it has an image provided by an artifact? This would be like just using keel to deploy to test.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryPromotionRepository.kt
robfletcher: Also, this is just the in-memory repository so it doesn't need to be super optimal.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
emjburns: Can we do this where each constraint has a "handles" function which is the type of constraint it has an opinion on, and it only gets asked for an opinion if the env has a constraint of its type?

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -43,25 +47,49 @@ class ResourceCheckScheduler(
...
+      log.debug("Scheduled environment validation complete")
+    } else {
+      log.debug("Scheduled environment validation disabled")
+    }
+  }
+
+  private fun <T : Any> PeriodicallyCheckedRepository<T>.launchForEachItem(block: suspend CoroutineScope.(T) -> Unit) {
+    itemsDueForCheck(Duration.ofMinutes(resourceCheckMinAgeMinutes), resourceCheckBatchSize)
```
cfieber: what happens if the block throws an exception?

cfieber: do we need to catch/log in here?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/actuation/ArtifactPromotionListener.kt
cfieber: our cluster model is still regional so 1 cluster per region all using the same artifact (assuming artifact is generically representing the global image (container or imageName + base that we baked)) seems like a legit common state to be considering

robfletcher: Raised #361 to track this

PR #353, opened 2019-08-02
moved resource last checked timestamp to own table

robfletcher:


robfletcher:
@emjburns as discussed previously. By the way, I was mistaken about us needing to sort timestamps in descending order. We want oldest first.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
```diff
@@ -94,7 +94,6 @@ class ResourcePersister(
         .also {
           resourceRepository.store(it)
           resourceRepository.appendHistory(ResourceUpdated(it, diff.toUpdateJson(), clock))
-          resourceRepository.markCheckDue(it)
```
robfletcher: Doing this in the `store` method now

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -310,7 +310,7 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
         }
       }
 
-      test("if a resource was explicitly marked due for checking is it returned") {
+      test("after a resource is updated is it returned") {
```
robfletcher: See?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -156,13 +167,7 @@ class SqlResourceRepository(
...
-    jooq.update(RESOURCE)
-      // MySQL is stupid and won't let you insert a zero valued TIMESTAMP
-      .set(mapOf(RESOURCE.LAST_CHECKED to EPOCH.plusSeconds(1).atZone(clock.zone).toLocalDateTime()))
-      .where(RESOURCE.UID.eq(resource.uid.toString()))
-      .execute()
-  }
-
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  private fun Instant.toLocal() = atZone(clock.zone).toLocalDateTime()
```
robfletcher: Keep doing this everywhere so made it an extension

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -89,6 +91,12 @@ class SqlResourceRepository(
       .onDuplicateKeyUpdate()
       .set(updatePairs)
       .execute()
+    jooq.insertInto(RESOURCE_LAST_CHECKED)
```
emjburns: Do we need to specify that this should happen as a transaction (or are we specifying that in some what that I don't understand)?

robfletcher: The transaction boundary is in `ResourcePersister`. I want to add `@Transactional(REQUIRED)` or `@Transactional(SUPPORTS)` to the relevant methods in the repository classes, but only adding it on the `SQL*` ones messed with the `spring-kotlin` Gradle plugin and adding it on the interface seems meaningless when in-memory, etc. probably couldn't be guaranteed to abide by it.

robfletcher: Raised #354 to track this

PR #356, opened 2019-08-05
feat(application): restart with schemas, add application as indexable field

emjburns:
This PR removes all the database change history, defines our current database structure in new files, and adds `application` to the resource table as a required and indexable field.

This seemed less complex than figuring out how to do a migration given that our schemas are still changing. This will require a wipe of the database when we deploy. 

This PR also adds an `/application` controller where you can hit `GET /application/{application}?includeDetails=true` to figure out if an app has managed resources. This enables us to conditionally show a "opt out" button in the application config as a break-the-glass "please stop managing my resources". If you don't pass the `includeDetails` flag (it defaults to false) you just get the `hasManagedResources` field.

Response looks like:
``` json
{
  "hasManagedResources": true,
  "resources": [
    "ec2:cluster:test:us-west-1:emburnstest-managed-reference",
    "ec2:securityGroup:test:us-west-2:emburnstest-managed",
    "ec2:classic-load-balancer:test:us-east-1:emburnstest-managed-clb",
    "ec2:application-load-balancer:test:us-east-1:emburnstest-managed-wowalb",
    "bakery:image:deb-sample-app-server"
  ]
}
```


erikmunson:
thank youuuuu 💯

one question/idea: could we start out with just a _count_ of the number of resources that are being managed, and add an additional endpoint or param later to send back a full list of resource names? I don't know of a good case for using the actual resource names right now in the UI and I'd love to keep the payload slim given it'll get hit for every app config load.

emjburns:
@erikmunson I threw in resource names just to see if it was useful. Thanks for the feedback! I'll make 2 apis - one for the boolean does it have resources, and one to get resources for an app. That'll be better.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Applicationed.kt
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.api
+
+/**
+ * An object which is associated with an application by the field application
+ */
+interface Applicationed {
+  val application: String
```
robfletcher: This seems like a different approach to what we've done with typed metadata before (`name`, `uid`, `serviceAccount`). Curious why the difference.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
```diff
@@ -63,6 +64,9 @@ val <T : Any> Resource<T>.name: ResourceName
 val <T : Any> Resource<T>.serviceAccount: String
   get() = metadata.getValue("serviceAccount").toString()
 
+val <T : Any> Resource<T>.application: String
+  get() = metadata.getValue("application").toString()
```
robfletcher: Does this make it implement the `Applicationed` interface? Seems like it duck-types to it.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
```diff
@@ -63,6 +64,25 @@ class InMemoryResourceRepository(
       }
     } ?: throw NoSuchResourceUID(uid)
 
+  override fun hasManagedResources(application: String): Boolean {
+    resources.forEach { (_, resource) ->
+      if (resource.application == application) {
+        return true
+      }
+    }
+    return false
```
robfletcher: ```resources.any { it.application == application }```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
```diff
@@ -63,6 +64,25 @@ class InMemoryResourceRepository(
...
+  }
+
+  override fun getByApplication(application: String): List<String> {
+    val matches = mutableListOf<String>()
+    resources.forEach { (_, resource) ->
+      if (resource.application == application) {
+        matches.add(resource.name.toString())
+      }
+    }
+    return matches
```
robfletcher: ```resources.filter …```?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -73,14 +74,38 @@ class SqlResourceRepository(
...
+      .select(RESOURCE.NAME)
+      .from(RESOURCE)
+      .where(RESOURCE.APPLICATION.eq(application))
+      .fetchArray(RESOURCE.NAME)
+      .let { results ->
+        if (results.isNotEmpty()) {
+          return true
+        }
+      }
+    return false
```
robfletcher: Think you can just do `jooq.selectOne().from…` rather than having to filter results.

emjburns: I tried this, and if you say `jooq.selectOne()` and there is more than one resource it throws an error. The normal `.fetch()` always returns a record and therefor is never null. What I'm doing here definitely doesn't seem great though. I'll keep thinking

emjburns:
@robfletcher I ported all our schemas into a new changelog so we can start fresh (and because I made a big change to the resource table and didn't want to migrate)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
emjburns: I found a better way thanks to Zed! `selectCount()`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Applicationed.kt
emjburns: The main difference is this is for the spec, before it gets turned into a resource. This is used to pull the app value into the top level metadata so that we can use it elsewhere. This is used in `resolvableResourceHandler.normalize()`. I didn't want to add another required field to the submittedMetadata when that information is already present in the resource body, just in two different places (some things don't have a moniker, like images).

I can't think of a more elegant way to do this, @cfieber got any suggestions.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
```diff
@@ -1,9 +1,12 @@
...
+
 data class ImageSpec(
   val artifactName: String,
   val baseLabel: BaseLabel,
   val baseOs: String,
   val regions: Set<String>,
-  val storeType: StoreType
-)
+  val storeType: StoreType,
+  override val application: String // the application an image is baked in
```
cfieber: this might be tricky - there are definitely cases of generic-ish images that are used in multiple applications (and maybe even baked in multiple different applications pipelines). Not sure what to suggest here though..

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
```diff
@@ -1,9 +1,12 @@
 package com.netflix.spinnaker.keel.bakery.api
 
+import com.netflix.spinnaker.keel.api.Applicationed
```
cfieber: spent a while trying to get the right english teacher words to call out `Application` -> `Applicationed` but I fell short so I suspect my dad is going to disown me (gerund? past particple? past gerund? maybe Applicationed is the past gerunded form of Application)

Maybe HasApplication as the interface instead? I dunno, naming is hard

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Applicationed.kt
cfieber: I like this approach (maybe hung up on the interface name)

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResolvableResourceHandler.kt
```diff
@@ -60,10 +62,17 @@ interface ResolvableResourceHandler<S : Any, R : Any> : KeelPlugin {
         e
       )
     }
+    val app = when (spec) {
+      is Monikered -> spec.moniker.app
```
cfieber: worth just making Monikered extend Applicationed?

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/KeelTagSpec.kt
```diff
@@ -34,8 +35,9 @@ import com.netflix.spinnaker.keel.tags.EntityTag
 data class KeelTagSpec(
   val keelId: String,
   val entityRef: EntityRef,
-  val tagState: TagState
-)
+  val tagState: TagState,
+  override val application: String // this is found from entity ref
```
cfieber: feels like for a case where we have the application source object in the constructor we should get rid of application as a constructor arg and move the `val application = entityRef.application` into the constructor itself

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -93,10 +93,12 @@ class ResourceTagger(
...
     if (event.resourceName.shouldTag()) {
       log.debug("Persisting no tag desired for resource {} because it is no longer managed", event.resourceName.toString())
+      val entityRef = event.resourceName.toEntityRef()
       val spec = KeelTagSpec(
         keelId = event.resourceName.toString(),
-        entityRef = event.resourceName.toEntityRef(),
-        tagState = TagNotDesired(startTime = clock.millis())
+        entityRef = entityRef,
+        tagState = TagNotDesired(startTime = clock.millis()),
+        application = entityRef.application
```
cfieber: specifically to comment above on KeelTagSpec, this seems redundant to supply here

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResolvableResourceHandler.kt
```diff
@@ -60,10 +62,17 @@ interface ResolvableResourceHandler<S : Any, R : Any> : KeelPlugin {
...
+      is Applicationed -> spec.application
+      else -> "noapp" // todo eb: is this the right way to handle this?
```
cfieber: feels like if app is really optional then maybe modelling it as such on metadata would be better. Right now I'm domain squatting on the `noapp` application in our spinnaker just because I'm a jerk

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
cfieber: The Applicationed/Monikered interfaces are being used on the way in to generate this value on metadata - but I think the fallback to `noapp` to ensure we have a value here isn't ideal

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Applicationed.kt
robfletcher: Yeah, I like it. And we could implement the interface still backed by a metadata field (for JSON consistency, interface field would be ignored).

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
robfletcher: Would that be an issue? Pretty sure the application field is irrelevant as far as the bakery is concerned, it just determines where the bake task show up in the Spinnaker UI.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
robfletcher: I like `HasApplication` FWIW.

emjburns: 😆 changing to `HasApplication` because it seems less bad

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
cfieber: discussed this in person with @emjburns and I think I get where we are at with this now (this is where we will ask to bake a newly detected artifact vs where we expect it to have come from)

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResolvableResourceHandler.kt
emjburns: I'd expect nothing less of you.

I'm changing this to throw an exception here to indicate there's something wrong with the plugin if it allows a resource which doesn't have an application.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
emjburns: Yep, adding this so that we bake artifacts where people want them. This will be used for generating bake tasks. I should actually do that...

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
emjburns: removing the fallback and throwing an exception instead.

Issue #355, opened 2019-08-05
Actually use service account when converging on resources


emjburns:
Right now we require this information but don't use it to pass identity along. We need to use the service account to make all mutating changes.

PR #358, opened 2019-08-07
feat(auth): use service account for CRUD

emjburns:
Closes https://github.com/spinnaker/keel/issues/355

Remove `X-SPINNAKER-USER=keel@spinnaker.io` as the default header. Instead, use the service account provided with every resource.

Also, inject "user = Spinnaker" into the `job` payload because it turns out this is how you get a nice  task UI `User = Spinnaker (delivery-engineering@netflix.com)`


emjburns:
I think I need to figure out how to conditionally set the `X-SPINNAKER-USER=keel@spinnaker.io` for requests that don’t set it on their own. With this always being set the individual requests couldn't override it. I'll follow up on that tomorrow.

robfletcher:
I think you can use an annotation on a parameter in a Retrofit interface to set a header value.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaService.kt
```diff
@@ -29,7 +30,7 @@ interface OrcaService {
 
   @POST("/ops")
   @Headers("Content-Type: application/context+json")
-  suspend fun orchestrate(@Body request: OrchestrationRequest): TaskRefResponse
+  suspend fun orchestrate(@Header("X-SPINNAKER-USER") serviceAccount: String, @Body request: OrchestrationRequest): TaskRefResponse
```
emjburns: This is where we add the header.

emjburns:
Updated all clouddriver calls to have a header as well. In some calls (like those we use to populate a cache) I'm choosing to use the default service account for now (what we were using before). I _think_ this is ok because we're just reading things. Anywhere we have access to the resource we're using the serviceAccount to perform those reads.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
```diff
@@ -90,6 +111,7 @@ interface CloudDriverService {
 
   @GET("/aws/images/find")
   suspend fun namedImages(
+    @Header("X-SPINNAKER-USER") serviceAccount: String,
```
robfletcher: Would this one be a good candidate for the default?

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandlerTest.kt
```diff
@@ -158,7 +158,7 @@ internal class ApplicationLoadBalancerHandlerTests : JUnit5Minutests {
         every { securityGroupByName(vpc.account, vpc.region, sg1.name) } returns sg1
       }
 
-      coEvery { orcaService.orchestrate(any()) } returns TaskRefResponse("/tasks/${UUID.randomUUID()}")
+      coEvery { orcaService.orchestrate("keel@spinnaker", any()) } returns TaskRefResponse("/tasks/${UUID.randomUUID()}")
```
robfletcher: Do we care?

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
emjburns: Sometimes we use this where we have the service account, sometimes we use this where we don't have it. I chose not to put the default in here because we can pass in the right one sometimes

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandlerTest.kt
emjburns: ehhhh maybe? I'm not sure.

PR #364, opened 2019-08-13
Tidy up event classes and handling

robfletcher:


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -65,45 +62,37 @@ class ResourceActuator(
...
+          // TODO: not sure this logic belongs here
           val lastEvent = resourceRepository.eventHistory(resource.uid).first()
           if (lastEvent is ResourceDeltaDetected || lastEvent is ResourceActuationLaunched) {
-            val event = ResourceDeltaResolved(
-              resource = resource,
-              current = current,
-              clock = clock
-            )
-            resourceRepository.appendHistory(event)
-            publisher.publishEvent(event) // TODO: refactor pending to unify how all this stuff with events works
```
robfletcher: This was the case that really triggered me wanting to do this refactor. It seems crazy for this class to be responsible for appending to the resource history _and_ raising an event for the benefit of another listener. Instead the resource history can just be managed by a dedicated listener responding to the same events.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
```diff
@@ -75,8 +74,7 @@ class ResourcePersister(
       .also {
         log.debug("Creating $it")
         resourceRepository.store(it)
-        resourceRepository.appendHistory(ResourceCreated(it, clock))
-        publisher.publishEvent(CreateEvent(it.name))
+        publisher.publishEvent(ResourceCreated(it, clock))
```
robfletcher: We already had this `ResourceCreated` event and `CreateEvent` seemed to just duplicate it. Listeners can just use the same event type.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -96,16 +111,35 @@ data class ResourceUpdated(
...
+    resource.kind,
+    resource.name.value,
+    clock.instant()
+  )
+}
 
-  constructor(resource: Resource<*>, delta: Map<String, Any?>) : this(resource, delta, clock)
+abstract class ResourceCheckResult : ResourceEvent() {
+  abstract val state: ResourceState
+  override val ignoreRepeatedInHistory = true
```
robfletcher: Added this intermediate abstract class (which is the thing Atlas telemetry is interested in).

robfletcher: Also we don't want > 1 of these in a row in the events table.

PR #366, opened 2019-08-14
Test factories

robfletcher:
We do a lot of creating of resources, etc. for unit tests and they contain numerous mandatory fields that tests are often not interested in. Makes sense to have some factory methods that can do some of that repetitive initialization.


emjburns:
LGTM, this is so much better.

robfletcher:
Okay, this ended up branching out to the creation of a base type for all resource specs. The reason for this is that while creating the resource test factory methods it became apparent that generating the resource name was a bit convoluted. It made sense to move the source of truth from the handler to the resource spec itself. 

Since all resources must also have an `application` property, I merged the existing `HasApplication` interface into the new `ResourceSpec`.

The `SubmittedResource` type does not enforce that its `spec` has to be a `ResourceSpec` at compile time. This is because of the difficulty in having the REST endpoint in `ResourceController` identify the concrete type. That's done programmatically in the controller method.

emjburns:
This is too big for me to spot any problems but overall I like the direction and I think that the way you've done name and `ResourceSpec` is much much better.

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/Resources.kt
```diff
@@ -0,0 +1,107 @@
...
+import java.util.UUID
+
+val TEST_API = SPINNAKER_API_V1.subApi("test")
+
+fun resource(
+  apiVersion: ApiVersion = TEST_API,
+  kind: String = "whatever",
+  name: String = randomString(),
+  application: String = "fnord"
+): Resource<DummyResourceSpec> =
```
robfletcher: First variant of test factory, lets you easily build a `Resource<DummyResourceSpec>` which is probably ideal for most tests that are not dealing with a specific resource type. You can call this with no parameters to get a nice random resource.

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/Resources.kt
```diff
@@ -0,0 +1,107 @@
...
+        name = spec.name,
+        application = application
+      )
+    }
+
+fun <T : Monikered> resource(
+  apiVersion: ApiVersion = TEST_API,
+  kind: String = "whatever",
+  spec: T
+): Resource<T> = resource(
```
robfletcher: Second variant of test factory, knows how to get a name from a `Monikered` spec.

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/Resources.kt
```diff
@@ -0,0 +1,107 @@
...
+  application = spec.application
+)
+
+fun <T : ResourceSpec> resource(
+  apiVersion: ApiVersion = TEST_API,
+  kind: String = "whatever",
+  spec: T,
+  name: String = spec.name,
+  application: String = "fnord"
+): Resource<T> =
```
robfletcher: Third variant of test factory, you supply the spec, it (optionally) does the rest. The other variants call this one.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceSpec.kt
```diff
@@ -0,0 +1,26 @@
+package com.netflix.spinnaker.keel.api
+
+/**
+ * Implemented by all resource specs.
+ */
+interface ResourceSpec {
```
robfletcher: Common interface for all resource specs.

PR #377, opened 2019-08-19
feat(artifact): improvements in artifact registering

emjburns:
The big change: when you register an artifact version, we automatically get the latest version from igor and persist that.

The bigger change: when you submit an image yaml and that artifact isn't registered we automatically register it for you.

Another change: when you register an artifact, it no longer tells you if that artifact is already registered.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
@@ -3,6 +3,7 @@ package com.netflix.spinnaker.keel.rest
 import com.netflix.spinnaker.keel.api.ArtifactType
 import com.netflix.spinnaker.keel.api.DeliveryArtifact
 import com.netflix.spinnaker.keel.events.ArtifactEvent
+import com.netflix.spinnaker.keel.events.ArtifactRegisteredEvent
 import com.netflix.spinnaker.keel.persistence.ArtifactAlreadyRegistered
 import com.netflix.spinnaker.keel.persistence.ArtifactRepository
 import com.netflix.spinnaker.keel.persistence.NoSuchArtifactException
@@ -52,6 +53,7 @@ class ArtifactController(
   @ResponseStatus(CREATED)
   fun register(@RequestBody artifact: DeliveryArtifact) {
     log.debug("Registering {} artifact {}", artifact.type, artifact.name)
+    publisher.publishEvent(ArtifactRegisteredEvent(name = artifact.name, type = artifact.type))
     artifactRepository.register(artifact)
   }
 
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
@@ -53,8 +53,7 @@ class ArtifactController(
   @ResponseStatus(CREATED)
   fun register(@RequestBody artifact: DeliveryArtifact) {
     log.debug("Registering {} artifact {}", artifact.type, artifact.name)
-    publisher.publishEvent(ArtifactRegisteredEvent(name = artifact.name, type = artifact.type))
-    artifactRepository.register(artifact)
+    publisher.publishEvent(ArtifactRegisteredEvent(artifact))
   }
 
   @GetMapping(
--- a/keel-api/src/test/kotlin/com/netflix/spinnaker/keel/integration/AuthPropagationTests.kt
+++ b/keel-api/src/test/kotlin/com/netflix/spinnaker/keel/integration/AuthPropagationTests.kt
@@ -27,7 +27,8 @@ import strikt.assertions.isNotNull
   classes = [KeelApplication::class],
   properties = [
     "clouddriver.enabled=true",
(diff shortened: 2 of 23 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -52,7 +53,7 @@ class ArtifactController(
   @ResponseStatus(CREATED)
   fun register(@RequestBody artifact: DeliveryArtifact) {
     log.debug("Registering {} artifact {}", artifact.type, artifact.name)
-    artifactRepository.register(artifact)
```
emjburns: I've moved this into the `ArtifactListener` because I thought it made sense to centralize registering an artifact.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/integration/AuthPropagationTests.kt
```diff
@@ -27,7 +27,8 @@ import strikt.assertions.isNotNull
   classes = [KeelApplication::class],
   properties = [
     "clouddriver.enabled=true",
-    "clouddriver.base-url=http://localhost:8080"
+    "clouddriver.base-url=http://localhost:8080",
+    "igor.base-url=http://localhost:8081"
```
emjburns: Since the `artifactRepostiory` is `igor`, it needs to be included in all the tests.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/ArtifactControllerTests.kt
```diff
@@ -59,24 +60,6 @@ internal class ArtifactControllerTests {
       .andExpect(status().isCreated)
   }
 
-  @Test
```
emjburns: I removed this behavior so I removed this test. My thought was that people are going to register their artifacts by submitting an image yaml file. I think we need a better way to indicate that two apps care about the same image.

keel-igor/src/main/kotlin/com/netflix/spinnaker/igor/ArtifactService.kt
```diff
@@ -6,9 +6,14 @@ import retrofit2.http.Path
 
 interface ArtifactService {
 
-  @GET("/artifacts/rocket/{application}/{version}")
+  @GET("/artifacts/rocket/{packageName}/{version}")
```
emjburns: this is actually the package name and not the application. renamed the parameter so that was clear.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -35,5 +40,30 @@ class ArtifactListener(
...
+    val artifact = event.artifact
+    if (artifactRepository.isRegistered(artifact.name, artifact.type)) {
+      log.debug("Artifact {} is already registered", artifact)
+      return
+    }
+    artifactRepository.register(artifact)
+    runBlocking {
+      artifactService
+        .getVersions(artifact.name)
+        .firstOrNull()
```
robfletcher: That's definitely the right order, right? It's not the kind of mistake I'd ever make, but I thought it was worth double-checking.

keel-igor/src/main/kotlin/com/netflix/spinnaker/config/IgorConfiguration.kt
```diff
@@ -13,7 +13,7 @@ import retrofit2.Retrofit
 import retrofit2.converter.jackson.JacksonConverterFactory
 
 @Configuration
-@ConditionalOnProperty("igor.enabled")
+@ConditionalOnProperty("igor.base-url")
```
cfieber: I propose we just get rid of this and make keel expect that an Iger is available

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/EventControllerTests.kt
```diff
@@ -51,9 +51,10 @@ import java.time.Instant
 @SpringBootTest(
   classes = [KeelApplication::class, MockEurekaConfiguration::class, MockTimeConfiguration::class],
   properties = [
-    "clouddriver.baseUrl=https://localhost:8081",
-    "orca.baseUrl=https://localhost:8082",
-    "front50.baseUrl=https://localhost:8083"
+    "clouddriver.base-url=https://localhost:8081",
```
cfieber: we might be able to add a src/test/resources/application.properties with all these values in it and not have to add them to all the spring based tests

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: 😆 Yep, from this endpoint the versions come back sorted with newest first.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/EventControllerTests.kt
emjburns: Figured it out, thanks for the suggestion! That is so much better.

PR #378, opened 2019-08-19
add application to Atlas tags, track orchestrations

robfletcher:


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryListener.kt
```diff
@@ -71,6 +73,19 @@ class TelemetryListener(
     lastResourceCheck = clock.instant()
   }
 
+  @EventListener(ResourceActuationLaunched::class)
+  fun onResourceActuationLaunched(event: ResourceActuationLaunched) {
```
robfletcher: Ping Atlas when we launch an orchestration so we can graph this

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -58,6 +59,7 @@ sealed class ResourceEvent {
   abstract val apiVersion: ApiVersion
   abstract val kind: String
   abstract val name: String
+  abstract val application: String
```
robfletcher: Added `application` to events so we can use that as a tag in Atlas

PR #380, opened 2019-08-19
Only look at artifact events we support

robfletcher:
I think I accidentally backed this out by screwing up a merge / rebase.


robfletcher:
See #371

PR #381, opened 2019-08-20
fix(artifact): parse artifact correctly

emjburns:
The artifact name is not necessarily the app name. For example, the "appVersion" tags for keel are of the form `keel-0.231.0-h146.a7234bf` but my super great sample app has "appVersion" tags the form `deb-sample-app-server-0.0.1~rc.103-h104.72b8e24`. This PR uses frigga to parse the package name so that we correctly get `keel` as well as `deb-sample-app-server`.


emjburns:
@robfletcher you do but it's in a different place and I _think_ not applicable but it's on my list to look into.

PR #382, opened 2019-08-20
fix(test): delay loading clouddriver creds

emjburns:
this was causing the `AuthPropagationTest` to sporadically fail because it was making an unexpected request to clouddriver.


emjburns:
I think the right solution to this is to add `@EnableAutoConfiguration(exclude = [TaskSchedulingAutoConfiguration::class]) to `AuthPropagationTests`, which I'll explore in the future. Merging this for now to get the build reliably unbroken.

PR #385, opened 2019-08-21
ensure we don't end up with empty string bits of names

### LINKED ISSUE #324: Resources with empty strings in Moniker fields have malformed names
If a resource is submitted with stack or detail set to "" rather than null (or not present in the yaml) the resource name is generated incorrectly. 

We should probably normalize empty strings to null.

This might make sense: https://stackoverflow.com/a/36780112/565049

robfletcher:
Fixes #324

### DIFF
--- a/keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
+++ b/keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
@@ -123,6 +123,38 @@ internal class ResourceControllerTests {
   }
 
   @Test
+  fun `empty string fields are normalized to null`() {
+    every { resourcePersister.upsert<DummyResourceSpec>(any()) } returns resource
+    every { authorizationSupport.userCanModifySpec("keel@spinnaker", any()) } returns true
+
+    val request = post("/resources")
+      .accept(APPLICATION_JSON)
+      .contentType(APPLICATION_JSON)
+      .content(
+        """{
+          |  "apiVersion": "test.spinnaker.netflix.com/v1",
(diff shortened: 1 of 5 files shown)


keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
```diff
@@ -118,6 +118,38 @@ internal class ResourceControllerTests {
...
+      .contentType(APPLICATION_JSON)
+      .content(
+        """{
+          |  "apiVersion": "test.spinnaker.netflix.com/v1",
+          |  "kind": "whatever",
+          |  "metadata": {
+          |    "serviceAccount": "keel@spinnaker"
+          |  },
+          |  "spec": {
+          |    "data": ""
```
robfletcher: Post a spec with an empty string field

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
```diff
@@ -118,6 +118,38 @@ internal class ResourceControllerTests {
...
+          |}"""
+          .trimMargin()
+      )
+    mvc
+      .perform(request)
+      .andExpect(status().isOk)
+
+    verify {
+      resourcePersister.upsert(match<SubmittedResource<out ResourceSpec>> {
+        (it.spec as? DummyResourceSpec)?.data == null
```
robfletcher: Ensure it gets persisted as `null`

luispollo:
@robfletcher Closing this since it's been stable for many months. Feel free to reopen if needed.

PR #393, opened 2019-08-25
parse ResourceSpec at REST boundary

robfletcher:


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
```diff
@@ -41,10 +44,15 @@ data class Resource<T : ResourceSpec>(
 /**
  * External representation of a resource that would be submitted to the API
  */
-data class SubmittedResource<T : Any>(
+data class SubmittedResource<T : ResourceSpec>(
```
robfletcher: This is the key thing with this PR. Being able to strongly type the spec at the REST boundary so we don't have to do any messing about converting things.

keel-api/src/main/kotlin/com/netflix/spinnaker/config/YamlConfiguration.kt
```diff
@@ -15,12 +15,10 @@
  */
 package com.netflix.spinnaker.config
 
-import com.netflix.spinnaker.keel.yaml.YamlHttpMessageConverter
-import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 
 @Configuration
 class YamlConfiguration {
```
cfieber: This doesn't do anything now.. any reason to keep it around?

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
```diff
@@ -83,7 +82,7 @@ internal class DeliveryConfigControllerTests : JUnit5Minutests {
                   apiVersion = SPINNAKER_API_V1.subApi("test"),
                   kind = "whatever",
                   metadata = SubmittedMetadata("keel@spinnaker"),
-                  spec = DummyResourceSpec(data = "resource in test").asMap()
+                  spec = DummyResourceSpec(data = "resource in test") // .asMap()
```
cfieber: Remove the commented out bits in here

keel-api/src/main/kotlin/com/netflix/spinnaker/config/YamlConfiguration.kt
robfletcher: Good spot. Fixed

PR #407, opened 2019-09-03
Handle multiple AMIs for appversion (different AMIs in each region)

robfletcher:
I know this should not strictly be necessary but it's pretty easy to make Keel resilient to this case

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
@@ -59,10 +59,13 @@ class ImageService(
   }
 
   /**
-   * Get the latest named image for a package
+   * Get the latest named image for a package.
+   *
+   * @param region if supplied the latest image in this region is returned, if `null` the latest
+   * image regardless of region.
    */
-  suspend fun getLatestNamedImage(packageName: String, account: String): NamedImage? =
-    cloudDriverService.namedImages(DEFAULT_SERVICE_ACCOUNT, "$packageName-", account)
+  suspend fun getLatestNamedImage(packageName: String, account: String, region: String? = null): NamedImage? =
+    cloudDriverService.namedImages(DEFAULT_SERVICE_ACCOUNT, "$packageName-", account, region)
       .sortedWith(NamedImageComparator)
       .lastOrNull {
         AppVersion.parseName(it.appVersion).packageName == packageName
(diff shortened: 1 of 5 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -59,10 +59,13 @@ class ImageService(
...
-   * Get the latest named image for a package
+   * Get the latest named image for a package.
+   *
+   * @param region if supplied the latest image in this region is returned, if `null` the latest
+   * image regardless of region.
    */
-  suspend fun getLatestNamedImage(packageName: String, account: String): NamedImage? =
-    cloudDriverService.namedImages(DEFAULT_SERVICE_ACCOUNT, "$packageName-", account)
+  suspend fun getLatestNamedImage(packageName: String, account: String, region: String? = null): NamedImage? =
+    cloudDriverService.namedImages(DEFAULT_SERVICE_ACCOUNT, "$packageName-", account, region)
```
robfletcher: CloudDriver already allows filtering by region

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -47,12 +48,12 @@ class ImageResolver(
         }
         val account = dynamicConfigService.getConfig("images.default-account", "test")
         val namedImage = imageService
-          .getLatestNamedImage(artifactName, account) ?: throw NoImageFound(artifactName)
+          .getLatestNamedImage(artifactName, account, region) ?: throw NoImageFound(artifactName)
```
robfletcher: supplying the cluster resource's region here means we'll get a filtered list of AMIs back from River

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -296,7 +296,8 @@ data class ResourceCheckError(
   override val name: String,
   override val application: String,
   override val timestamp: Instant,
-  val exception: Throwable
+  val exceptionType: Class<Throwable>,
+  val exceptionMessage: String?
```
robfletcher: Semi-related change. I noticed the resource history has big blobs of stack trace when there are resource check errors. Seems undesirable to store that. I'll need to flush DB before rolling this out.

emjburns:
Is there a case where we deploy a cluster with a different different image in each region?

robfletcher:
> Is there a case where we deploy a cluster with a different different image in each region?

I don't believe so yet. Once we get to the concept of regional roll-outs of new appversions I think that will be a thing.

PR #410, opened 2019-09-04
Limit number of events in history

robfletcher:


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
```diff
@@ -31,13 +31,13 @@ class EventController(
   )
   fun eventHistory(
     @PathVariable("name") name: ResourceName,
-    @RequestParam("since", defaultValue = "1970-01-01T00:00:00Z") since: Instant
+    @RequestParam("maxAge", defaultValue = "P3D") maxAge: Period,
```
srekapalli: Do we need to consider Pagination/InfiniteScrolling type approach here or not needed for the size of the data set ?

erikmunson: just to pile onto what Sairam said — ideally I expect the UI will want some kind of `before` / `after` semantic to enable paging (as opposed to page 1/2/3 etc). I don't take issue with this `maxAge` filter existing, provided it will peacefully coexist with the necessary filters for infinite scroll-y type paging later when needed.

emjburns:
You'll need to make the corresponding change here: https://github.com/spinnaker/gate/blob/de6efc207288e9008511b7042d50784f1c3f3af4/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/HistoryController.java#L37

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
robfletcher: Yeah, we'll possibly want it eventually. In that case I think we'd add a `page=whatever` param that would offset the results.

asher: Should make sure it's token based, i.e. a ULID and order.

robfletcher: That would involve adding ULIDs to the events

robfletcher: (which I'm not against doing)

robfletcher: maybe I should drop the max age and just have a default limit? Would be simple to add a "from ULID" to that in future. Currently I'm just trying to cut down on the volume of the responses. > 1MB of YAML is not a good thing.

robfletcher: I revised the PR to remove the `maxAge` option and just have a `limit` (defaulting to 10). We can easily enough add a ULID to each event and a `before` parameter in a future PR.

erikmunson: Thanks @robfletcher!

PR #412, opened 2019-09-04
fix(imageName): replace ~ with _

emjburns:
Artifact versions have ~ in them for dev artifacts, like `keeldemo-0.0.1~dev.1-h2.da2282b`. The corresponding image names replace ~ with an _. This PR replaces any ~s in the artifact version it finds so it can appropriately find the right image.

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
@@ -1,5 +1,8 @@
 package com.netflix.spinnaker.keel.ec2.resource
 
+import com.netflix.spinnaker.keel.api.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.NoImageFound
 import com.netflix.spinnaker.keel.api.NoImageFoundForRegion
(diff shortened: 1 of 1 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -46,6 +46,10 @@ class ImageResolver(
         } else {
           artifact.name
         }
+
+        // image names have _ instead of ~
+        artifactName.replace("~", "_")
```
robfletcher: That's not going to reassign the value, it returns the amended value. I think you should chain ```.let { it.replace("~", "_") }``` after the if/else.

emjburns: 🤦‍♀ lol.

I'd prefer not to chain `.let` after an if/else - I think it's pretty confusing to read. Are you ok with just reassigning the variable?

emjburns: Otherwise, I can just create a function that does the figuring out of the artifact name and pull the logic there.

robfletcher: Yeah, a function sounds even better

emjburns: Ok, updated.

PR #413, opened 2019-09-04
rename resource name to resource id

### LINKED ISSUE #411: Keel name should instead be ID
We should allow users to supply a name so that there is something unique identifying a resource in the metadata. Kubernetes tooling relies on a name in the metadata, and since we currently generate it that causes a problem. 

We should name our auto-generated field ID and use that instead of ID all over keel.

robfletcher:
In discussion with @emjburns it's apparent we want the concept of a resource "name" to be user assignable (at least in part in order to use Kustomize). This PR renames the existing `name` metadata entry in resources to `id`. It's a consistent, hierarchical, system-assigned, but human readable identifier. 

I've also made that consistently the thing we use to identify resources rather than also maintaining a UID. We had some overloaded methods on `ResourceRepository`, for example, and were using `uid` to access resource history but `name` (now `id`) to access most everything else. This was annoying and inconsistent in a few places. `SqlResourceRepository` still assigns a UID that is used as the primary key on the tables, but it's no longer exposed via the `Resource` class (except through `metadata`). That's the same as `DeliveryConfig` and `Environment`. It was also easier than updating all the reference tables to use the `id` property instead which would be a much more complex schema migration and probably require us to dump all data out of the DB before pushing this change out.

Fixes #411

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
@@ -32,7 +32,7 @@ data class Resource<T : ResourceSpec>(
   init {
     require(kind.isNotEmpty()) { "resource kind must be defined" }
     require(metadata["uid"].isValidULID()) { "resource uid must be a valid ULID" }
-    require(metadata["name"].isValidName()) { "resource name must be a valid name" }
+    require(metadata["id"].isValidId()) { "resource id must be a valid id" }
     require(metadata["serviceAccount"].isValidServiceAccount()) { "serviceAccount must be a valid service account" }
     require(metadata["application"].isValidApplication()) { "application must be a valid application" }
   }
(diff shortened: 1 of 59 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
```diff
@@ -31,14 +30,34 @@ data class Resource<T : ResourceSpec>(
...
+    if (javaClass != other?.javaClass) return false
+
+    other as Resource<*>
+
+    if (apiVersion != other.apiVersion) return false
+    if (kind != other.kind) return false
+    if (spec != other.spec) return false
+
+    return true
+  }
```
robfletcher: I'm not real happy with this, but several tests check for equality of a `Resource` (or an `Environment` or `DeliveryConfig` containing resources) before and after persistence and the fact that the "uid" metadata field is assigned at persistence time means that breaks. Open to any suggestions on improving that.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -88,19 +70,24 @@ open class SqlResourceRepository(
...
+      .fetchOne(RESOURCE.UID)
+      ?: randomUID().toString()
+
     val updatePairs = mapOf(
       RESOURCE.API_VERSION to resource.apiVersion.toString(),
       RESOURCE.KIND to resource.kind,
-      RESOURCE.NAME to resource.name.value,
-      RESOURCE.METADATA to objectMapper.writeValueAsString(resource.metadata),
+      RESOURCE.ID to resource.id.value,
+      RESOURCE.METADATA to objectMapper.writeValueAsString(resource.metadata + ("uid" to uid)),
```
robfletcher: This is the thing that screws with `Resource` equality

PR #417, opened 2019-09-05
added a new method for looking up AMI for exact app version

robfletcher:


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -112,3 +131,5 @@ class ImageService(
     return true
   }
 }
+
+private fun AppVersion.toImageName() = "$packageName-$version-h$buildNumber.$commit"
```
emjburns: ay ay ay I wish there was a better way

robfletcher: Yeah, I'm sure I'm missing some fuck-awful edge case, but I don't believe any of those fields are allowed to be `null`.

PR #423, opened 2019-09-06
fix(deliveryConfig): event for registering artifact

emjburns:
Publish an event for registering artifacts in the delivery config so that we automatically fetch the first version of them in [this function](https://github.com/spinnaker/keel/blob/e713b7281144aaa7a064cf4249804e2754131a08/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt#L47). This keeps the logic of registering in one place.

### DIFF
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
@@ -134,6 +134,7 @@ class ResourcePersister(
 
   private fun DeliveryArtifact.register() {
     if (!artifactRepository.isRegistered(name, type)) {
+      artifactRepository.register(this)
       publisher.publishEvent(ArtifactRegisteredEvent(this))
     }
   }
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
@@ -11,6 +11,7 @@ import com.netflix.spinnaker.keel.api.SubmittedDeliveryConfig
 import com.netflix.spinnaker.keel.api.SubmittedResource
 import com.netflix.spinnaker.keel.api.id
 import com.netflix.spinnaker.keel.diff.ResourceDiff
+import com.netflix.spinnaker.keel.events.ArtifactRegisteredEvent
 import com.netflix.spinnaker.keel.events.ResourceCreated
 import com.netflix.spinnaker.keel.events.ResourceDeleted
 import com.netflix.spinnaker.keel.events.ResourceUpdated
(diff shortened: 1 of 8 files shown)


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
```diff
@@ -133,7 +134,7 @@ class ResourcePersister(
 
   private fun DeliveryArtifact.register() {
     if (!artifactRepository.isRegistered(name, type)) {
-      artifactRepository.register(this)
```
emjburns: this is the real change

emjburns:
@robfletcher it's the method I linked in my description...

robfletcher:
So we were doing it twice?

emjburns:
? No, in this PR I remove the register from the resource persister, and instead publish the event. The event listener for that event first registers the artifact and then fetches the latest version for that artifact. The registration is moved from the `ResourcePersister` to the `ArtifactListener`

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/tx/DeliveryConfigTransactionTests.kt
```diff
@@ -143,27 +143,6 @@ internal class DeliveryConfigTransactionTests : JUnit5Minutests {
       }
     }
 
-    context("an artifact attached to the delivery config fails to persist") {
```
emjburns: This is no longer an assumption - if an artifact fails to be persisted, we won't fail to persist the whole config.

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/tx/DeliveryConfigTransactionTests.kt
```diff
@@ -177,11 +156,6 @@ internal class DeliveryConfigTransactionTests : JUnit5Minutests {
         expectThat(resourceRepository.allResourceNames())
           .isEmpty()
       }
-
```
emjburns: This is also no longer an assumption: if the delivery config fails the artifact still might be registered. That's okay.

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/tx/DeliveryConfigTransactionTests.kt
robfletcher: Is that bad? You'd have a bunch of resources and no artifact? I think the inverse is fine… having an artifact nothing references but this seems like it might cause problems.

emjburns: reverting this behavior after testing + in person conversations

emjburns:
Reverting most of this, instead I'm just adding the event publishing so that we get the registering existing versions behavior.

PR #427, opened 2019-09-09
feat(cluster): cluster converge with unhealthy instances

cfieber:
Changes the cluster task to use the rollback on failure flag, and times out the stage faster (30mins instead of 2hrs).

Rollback on failure only applies to instances failing to come up healthy, this change attempts to leave the cluster in
a previous-known-good but also needs-to-reconverge state so that we will get retries on deployment via keel attempting
to converge the cluster repeatedly.

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -130,8 +130,17 @@ class ClusterHandler(
       "delayBeforeDisableSec" to 0,
       "delayBeforeScaleDownSec" to 0,
       "maxRemainingAsgs" to 2,
+      // the 2hr default timeout sort-of? makes sense in an imperative
+      // pipeline world where maybe within 2 hours the environment around
+      // the instances will fix itself and the stage will succeed. Since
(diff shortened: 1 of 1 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -130,8 +130,17 @@ class ClusterHandler(
...
       "maxRemainingAsgs" to 2,
+      // the 2hr default timeout sort-of? makes sense in an imperative
+      // pipeline world where maybe within 2 hours the environment around
+      // the instances will fix itself and the stage will succeed. Since
+      // we are telling the red/black strategy to roll back on failure,
+      // this will leave us in a position where we will instead keep
+      // reattempting to clone the server group because the rollback
+      // on failure of instances to come up will leave us in a non
+      // converged state...
+      "stageTimeoutMs" to Duration.ofMinutes(30).toMillis(),
```
asher: Can we make this configurable?

cfieber: this is tucked within a whole block of "lets make the parameters to the deploy strategy configurable". I definitely forsee this being something we could override but I didn't go there first pass

emjburns:
That seems fine to me.

PR #430, opened 2019-09-11
fix(imagehandler): exclude unnecessary image regions

cfieber:
If an image exists in more regions than desired, consider the
current state only to contain the regions in the desired state.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
@@ -72,7 +72,9 @@ class ImageHandler(
 
   override suspend fun current(resource: Resource<ImageSpec>): Image? =
     with(resource) {
-      imageService.getLatestImage(spec.artifactName, "test")
+      imageService.getLatestImage(spec.artifactName, "test")?.let {
+        it.copy(regions = it.regions.intersect(resource.spec.regions))
+      }
     }
 
   private fun DeliveryArtifact.findLatestVersion(): String =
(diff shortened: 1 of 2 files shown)


keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -84,6 +84,18 @@ class ImageHandler(
     resource: Resource<ImageSpec>,
     resourceDiff: ResourceDiff<Image>
   ): List<Task> {
+
```
emjburns: Did you put this in here (vs where we detect the diff) because the general diff detection logic is generic, so there's no place for this kind of "ignore the diff in this situation" logic?

cfieber: I think I was working in the old mindset of the `current` method should return the actual current state of the world, but in hindsight I'd be better overriding desired to trim the set of regions

emjburns: Ya I think that would be slightly better

robfletcher: I concur

cfieber: Updated to modify the current state call instead

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -72,7 +72,9 @@ class ImageHandler(
 
   override suspend fun current(resource: Resource<ImageSpec>): Image? =
     with(resource) {
-      imageService.getLatestImage(spec.artifactName, "test")
+      imageService.getLatestImage(spec.artifactName, "test")?.let {
+        it.copy(regions = it.regions.intersect(resource.spec.regions))
+      }
```
robfletcher: ❤️

PR #432, opened 2019-09-11
feat(artifact): mark artifact deployed if already out

emjburns:
Current state:
- for artifacts to "progress" through environments, keel has to deploy them (it's not enough that they're already running in that environment)
- this is annoying for testing and annoying for someone switching to declarative

This pr:
- emits an event with the ami deployed to a cluster every time we get the current state of the cluster
- a new listener reacts to those events and checks if the version running has been approved for that env but not deployed
- if so, the listener marks that version as successfully deployed in the environment. 

I also change `keel.resource-check.min-age-minutes` to `keel.resource-check.min-age-duration` (where duration is something like `60s` or `1m`) because that makes it easier to quicken up the checks locally and I'm impatient.

### DIFF
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
@@ -24,7 +24,7 @@ class CheckScheduler(
   private val deliveryConfigRepository: DeliveryConfigRepository,
   private val resourceActuator: ResourceActuator,
   private val environmentPromotionChecker: EnvironmentPromotionChecker,
-  @Value("\${keel.resource-check.min-age-minutes:1}") private val resourceCheckMinAgeMinutes: Long,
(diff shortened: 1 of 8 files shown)


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -24,7 +24,7 @@ class CheckScheduler(
   private val deliveryConfigRepository: DeliveryConfigRepository,
   private val resourceActuator: ResourceActuator,
   private val environmentPromotionChecker: EnvironmentPromotionChecker,
-  @Value("\${keel.resource-check.min-age-minutes:1}") private val resourceCheckMinAgeMinutes: Long,
+  @Value("\${keel.resource-check.min-age-duration:60s}") private val resourceCheckMinAgeDuration: Duration,
```
robfletcher: I am delighted this is better typed. We can adjust the config in keel-nflx for non-local envs.

PR #440, opened 2019-09-13
Wrap resource resolution errors in standard exceptions

robfletcher:


keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -223,7 +230,33 @@ internal class ResourceActuatorTests : JUnit5Minutests {
             }
           }
 
-          context("plugin throws an exception on resource resolution") {
+          context("plugin throws an exception in current state resolution") {
+            before {
+              coEvery { plugin1.desired(resource) } returns DummyResource(resource.spec)
+              coEvery { plugin1.current(resource) } throws RuntimeException("o noes")
```
emjburns: this is my favorite exception

PR #449, opened 2019-09-17
feat(artifact): optionally specify status in image and cluster

emjburns:
If you only want to bake or deploy an artifact with a certain status, you should be able to do that. 

This PR allows you to optionally add a list of statuses (`FINAL`, `CANDIDATE`, or `SNAPSHOT`) to the `ImageSpec` and the `ArtifactImageProvider` section of the `ClusterSpec`.

I've verified that when you have resources in the database and you upgrade to this version everything serializes with the default value correctly. Until you get new artifacts in the database with status recorded you won't be able to deploy any new versions, but that's expected (we don't know the status of the old versions, so they don't meet the requirement of being `in` the default list of statuses.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
@@ -3,7 +3,6 @@ package com.netflix.spinnaker.keel.rest
 import com.netflix.spinnaker.keel.api.ArtifactType
 import com.netflix.spinnaker.keel.api.DeliveryArtifact
 import com.netflix.spinnaker.keel.events.ArtifactEvent
-import com.netflix.spinnaker.keel.events.ArtifactRegisteredEvent
 import com.netflix.spinnaker.keel.persistence.ArtifactAlreadyRegistered
 import com.netflix.spinnaker.keel.persistence.ArtifactRepository
 import com.netflix.spinnaker.keel.persistence.NoSuchArtifactException
@@ -12,7 +11,6 @@ import org.slf4j.LoggerFactory
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.http.HttpStatus.ACCEPTED
 import org.springframework.http.HttpStatus.CONFLICT
-import org.springframework.http.HttpStatus.CREATED
 import org.springframework.http.HttpStatus.NOT_FOUND
 import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
 import org.springframework.web.bind.annotation.ExceptionHandler
@@ -38,22 +36,7 @@ class ArtifactController(
   )
   @ResponseStatus(ACCEPTED)
   fun submitArtifact(@RequestBody echoArtifactEvent: EchoArtifactEvent) {
-    log.debug(
-      echoArtifactEvent.eventName,
-      echoArtifactEvent.payload.artifacts.map { it.name }
-    )
     publisher.publishEvent(echoArtifactEvent.payload)
-  }
-
-  // todo eb: do we need this endpoint anymore? we auto-register.
-  @PostMapping(
-    consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
-    produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
-  )
-  @ResponseStatus(CREATED)
-  fun register(@RequestBody artifact: DeliveryArtifact) {
-    log.debug("Registering {} artifact {}", artifact.type, artifact.name)
-    publisher.publishEvent(ArtifactRegisteredEvent(artifact))
   }
 
   @GetMapping(
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
@@ -39,13 +39,13 @@ class ArtifactController(
   @ResponseStatus(ACCEPTED)
   fun submitArtifact(@RequestBody echoArtifactEvent: EchoArtifactEvent) {
     log.debug(
-      "Received artifact events {} for {}",
       echoArtifactEvent.eventName,
       echoArtifactEvent.payload.artifacts.map { it.name }
     )
     publisher.publishEvent(echoArtifactEvent.payload)
   }
 
+  // todo eb: do we need this endpoint anymore? we auto-register.
   @PostMapping(
     consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
     produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
@@ -1,6 +1,7 @@
 package com.netflix.spinnaker.keel.artifact
 
 import com.netflix.spinnaker.igor.ArtifactService
+import com.netflix.spinnaker.keel.api.ArtifactStatus
 import com.netflix.spinnaker.keel.api.ArtifactType
 import com.netflix.spinnaker.keel.api.DeliveryArtifact
 import com.netflix.spinnaker.keel.events.ArtifactEvent
@@ -26,14 +27,13 @@ class ArtifactListener(
     event
       .artifacts
       .filter { it.type.toUpperCase() in artifactTypeNames }
-      .forEach {
-        val artifact = it.toDeliveryArtifact()
-        // TODO: should be able to construct this with Frigga or something, apparently, also it might
-        //  make sense to have a method that does this on the Kork class rather than here
-        val version = "${it.name}-${it.version}"
+      .forEach { korkArtifact ->
+        val artifact = korkArtifact.toDeliveryArtifact()
+        val version = "${korkArtifact.name}-${korkArtifact.version}"
+        val status = artifactStatus(korkArtifact)
         if (artifactRepository.isRegistered(artifact.name, artifact.type)) {
-          log.info("Registering version {} of {} {}", version, artifact.name, artifact.type)
-          artifactRepository.store(artifact, version)
+          log.info("Registering version {} ({}) of {} {}", version, status, artifact.name, artifact.type)
+          artifactRepository.store(artifact, version, status)
             .also { wasAdded ->
               if (wasAdded) {
                 publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.type))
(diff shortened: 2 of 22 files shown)


keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -54,22 +54,36 @@ class ArtifactListener(
...
-          artifactRepository.store(artifact, version)
+          val status = artifactStatus(artifactService.getArtifact(artifact.name, firstVersion))
+          log.debug("Storing latest version {} ({}) for registered artifact {}", version, status, artifact)
+          artifactRepository.store(artifact, version, status)
         }
     }
 
+  /**
+   * Parses the status from a kork artifact, and throws an error if [releaseStatus] isn't
+   * present in [metadata]
```
emjburns: This might be an assumption we have to re-work later, but I think it's ok for now.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -18,17 +19,26 @@ interface ArtifactRepository {
...
 
-  fun versions(artifact: DeliveryArtifact): List<String>
+  /**
+   * @returns the versions we have for an artifact, optionally filtering by status if provided
+   */
+  fun versions(
+    artifact: DeliveryArtifact,
+    statuses: List<ArtifactStatus> = enumValues<ArtifactStatus>().toList()
+  ): List<String>
+  // todo eb: if the list matches the default value, should we also include nulls? probably.
```
emjburns: Open question: should we allow `null` status to be okay if the list of statuses contains every possible value, or should we ignore nulls? This is only really relevant for the transition time going from the previous version to a new artifact being sent to us on this version. I'm inclined to just ignore this case.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/image/ImageProvider.kt
```diff
@@ -41,7 +42,8 @@ data class IdImageProvider(
  */
 @JsonDeserialize(using = JsonDeserializer.None::class)
 data class ArtifactImageProvider(
-  val deliveryArtifact: DeliveryArtifact
+  val deliveryArtifact: DeliveryArtifact,
+  val artifactStatuses: List<ArtifactStatus> = enumValues<ArtifactStatus>().toList()
```
emjburns: Default value here so you don't have to specify if you want all versions.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -48,7 +48,7 @@ class ImageResolver(
       }
     }
 
-  private suspend fun resolveFromImageId(resource: Resource<ClusterSpec>, imageProvider: IdImageProvider): Pair<String, String> {
+  private suspend fun resolveFromImageId(imageProvider: IdImageProvider): Pair<String, String> {
```
emjburns: wasn't used

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
```diff
@@ -9,6 +10,7 @@ data class ImageSpec(
   val baseOs: String,
   val regions: Set<String>,
   val storeType: StoreType,
+  val artifactStatuses: List<ArtifactStatus> = enumValues<ArtifactStatus>().toList(),
```
emjburns: Default value here so you don't have to specify if you want all versions.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -39,13 +39,13 @@ class ArtifactController(
   @ResponseStatus(ACCEPTED)
   fun submitArtifact(@RequestBody echoArtifactEvent: EchoArtifactEvent) {
     log.debug(
-      "Received artifact events {} for {}",
```
robfletcher: I R confused. Why is this line gone?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -39,13 +39,13 @@ class ArtifactController(
...
       echoArtifactEvent.eventName,
       echoArtifactEvent.payload.artifacts.map { it.name }
     )
     publisher.publishEvent(echoArtifactEvent.payload)
   }
 
+  // todo eb: do we need this endpoint anymore? we auto-register.
```
robfletcher: Nope. Let's kill it.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
robfletcher: Conversely we could just make the default value be `null` and then not restrict the query.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -60,35 +62,43 @@ class SqlArtifactRepository(
...
       .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifact.uid))
       .orderBy(ENVIRONMENT_ARTIFACT_VERSIONS.APPROVED_AT.desc())
-      .limit(1)
-      .fetchOne(0, String::class.java)
+      .fetch()
+      .getValues(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
+    val versionsWithCorrectStatus = versions(artifact, statuses)
+
+    // return the latest version that has been approved with the correct status
+    return approvedVersions.intersect(versionsWithCorrectStatus).firstOrNull()
```
robfletcher: Is there a reason we can't just add one more table join to the query rather than having to do this with the intersect of 2 different query results?

keel-sql/src/main/resources/db/changelog/20190913-add-status-to-artifact-version.yml
```diff
@@ -0,0 +1,48 @@
...
+            tableName: delivery_artifact_version
+      rollback:
+        - createIndex:
+            indexName: delivery_artifact_version_uid_idx
+            tableName: delivery_artifact_version
+            columns:
+              - column:
+                  name: delivery_artifact_uid
+  - changeSet:
+      id: status-to-artifact-version-create-better-index
```
robfletcher: Isn't the version still unique and the status is just supplementary?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ArtifactEvent.kt
```diff
@@ -8,6 +9,10 @@ data class ArtifactEvent(
   val details: Map<String, Any>?
 )
 
+/**
+ * This event optionally takes [statuses] so that we can fetch the most relevant first artifact.
+ */
 data class ArtifactRegisteredEvent(
-  val artifact: DeliveryArtifact
+  val artifact: DeliveryArtifact,
+  val statuses: List<ArtifactStatus> = enumValues<ArtifactStatus>().toList()
```
robfletcher: I don't understand why the statuses are needed here

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
emjburns: There's another almost identical statement that provides the actual useful detail

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: I consulted with @asher and he advised doing it this way. Also, my sql is not that great, and I found the syntax for joining to be extremely confusing.

keel-sql/src/main/resources/db/changelog/20190913-add-status-to-artifact-version.yml
emjburns: ya, but @asher advised creating the index this way because it can be used for both types of queries.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ArtifactEvent.kt
emjburns: It's in the class comment. For more information, when we register the artifact we fetch the latest version. I want to make sure we fetch the latest version with the correct status (or a status we care about) so we can deploy right away if needed.

robfletcher: How do we know what statuses we'd want to consider in this scenario?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
robfletcher: But this looks like just the message, not the entire log statement

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
emjburns: Do you think that's better than having the default be everything?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
emjburns: Oops :( fail. I'll fix!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ArtifactEvent.kt
emjburns: We publish the artifact register event in the `ImageHandler`, so we know what statuses we need to consider (it's in the `ImageSpec` now).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
robfletcher: In the unlikely event of a new status type being added it would continue to work!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ArtifactEvent.kt
robfletcher: Okay, I see, so this is just to ensure that the new resource gets an image it can use.

emjburns: Exactly.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
asher: I think I was thinking more in terms of multiple queries being ok and sometimes preferred where everything fetched is needed. The intersect is a sign of fetching more from the DB than we need which is better to avoid. For joins, I prefer to be explicit about the join type and related columns, makes it easier to understand index needs - I'd have to look at `EXPLAIN` output to check if the query in `versions()` is efficient. I need to get up to speed on how the schema has evolved around environments, I'm not currently sure how `ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION` differs from `DELIVERY_ARTIFACT_VERSION.VERSION` so I'm hesitant to write an example query here but would be happy to work it out in intellj w/you along with how to validate index usage and efficiency.

robfletcher: Those two things hold the same value. `DELIVERY_ARTIFACT_VERSION` holds all known versions per artifact. `ENVIRONMENT_ARTIFACT_VERSIONS` holds the versions that have been approved and/or deployed to a given environment.

emjburns: I made it bester

PR #453, opened 2019-09-18
Multi region clusters

robfletcher:


robfletcher:
This is what a cluster definition in YAML looks like right now: ```yaml
---
moniker:
  app: "fnord"
  stack: "test"
imageProvider:
  deliveryArtifact:
    name: "fnord"
    type: "DEB"
locations:
  accountName: "test"
  regions:
  - region: "us-east-1"
    subnet: "internal (vpc0)"
    availabilityZones:
    - "us-east-1a"
    - "us-east-1b"
    - "us-east-1c"
  - region: "us-west-2"
    subnet: "internal (vpc0)"
    availabilityZones:
    - "us-west-2a"
    - "us-west-2b"
    - "us-west-2c"
defaults:
  launchConfiguration:
    instanceType: "m5.large"
    ebsOptimized: true
    iamRole: "fnordInstanceProfile"
    instanceMonitoring: false
  capacity:
    min: 2
    max: 2
    desired: 2
overrides:
  us-east-1:
    launchConfiguration:
      iamRole: "fnordEastInstanceProfile"
      keyPair: "fnord-keypair-325719997469-us-east-1"
    capacity:
      min: 5
      max: 5
      desired: 5
  us-west-2:
    launchConfiguration:
      keyPair: "fnord-keypair-325719997469-us-west-2"```

I'd like to use `@JsonUnwrapped` to flatten the `defaults` field into the top level but then I cannot for the life of me get Jackson to parse that format back in.

robfletcher:
@cfieber I'd be in favor of supporting image name rather than AMI id.

luispollo:
> @cfieber I'd be in favor of supporting image name rather than AMI id.

Names have the added advantage of working across regions too (not sure that was a motivation). For example, kops [does this for their approved AMIs](https://github.com/kubernetes/kops/blob/master/docs/images.md) and it makes life much easier. Their names include well-known aliases for accounts using a DNS format (e.g. `netflix.com/my-awesome-ami`).

robfletcher:
@luispollo yeah, having people specify a particular AMI is a small minority use-case (but it is important for some teams), having it work multi-region is exactly the issue.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -0,0 +1,76 @@
...
+      tags = overrides[it.region]?.tags ?: defaults.tags ?: emptyMap()
+    )
+  }
+    .toSet()
+}
+
+data class ClusterSpec(
+  override val moniker: Moniker,
+  val imageProvider: ImageProvider,
+  val locations: ClusterLocations,
```
luispollo: Newbie question: this is `locations`, plural, but not a list, and then `ClusterLocations` below is mapped to a single account. Are all server groups assumed to be in the same account in a cluster, or was this meant to be a list?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -0,0 +1,76 @@
...
+}
+
+data class ClusterLocations(
+  val accountName: String,
+  val regions: Set<ClusterRegion>
+)
+
+data class ClusterRegion(
+  val region: String,
+  val subnet: String,
```
luispollo: Another one: one single subnet, but multiple AZs? How does that map out?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
robfletcher: all server groups in a cluster need to be in the same account

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
robfletcher: AFAIK subnets are region-wide. Most teams allow their server groups to balance across all available AZs but there are some teams that pin to particular AZs or even have a separate server group per AZ.

luispollo: Hmm... From the [AWS docs](https://docs.aws.amazon.com/vpc/latest/userguide/working-with-vpcs.html#AddaSubnet)
> Availability Zone: Optionally choose an Availability Zone in which your subnet will reside, or leave the default No Preference to let AWS choose an Availability Zone for you.

I totally understand balancing across AZs -- that's a recommended best practices. Which is why I wasn't sure what the single `subnet` property would do. Judging by the UI, it doesn't look like you manually create/specify subnets when creating a server group, only AZs. So where would this value come from?

Sorry for all the questions... 😅

robfletcher: Questions are good! I _think_ (and you probably want to check with somebody who knows the AWS environment better than me) that the same logical subnet name (which is what this field contains) maps to different subnet ids in each AZ under the hood. Whatever the exact story is there, it's handled in CloudDriver.

luispollo: Ah! I think you're right:
Screen Shot 2019-09-19 at 4 14 24 PM

robfletcher: @luispollo I think it's a valid question whether we can expect OSS users to have things set up that way

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -0,0 +1,191 @@
...
+import com.netflix.spinnaker.keel.api.Monikered
+import com.netflix.spinnaker.keel.api.ResourceSpec
+import com.netflix.spinnaker.keel.ec2.resource.ResolvedImages
+import com.netflix.spinnaker.keel.model.Moniker
+import java.time.Duration
+
+/**
+ * Transforms a [ClusterSpec] into a concrete model of server group desired states.
+ */
+fun ClusterSpec.resolve(resolvedImages: ResolvedImages): Set<ServerGroup> {
```
robfletcher: This is a "turn Keel spec into resolved desired state" function. I wanted it to be separate here rather than buried in `ClusterHandler` for ease of testing.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/actuation/ArtifactPromotionListener.kt
```diff
@@ -23,24 +23,25 @@ class ArtifactPromotionListener(
...
-        log.warn("Resource ${event.resourceId} is not part of a delivery environment")
-      } else {
-        log.info("Marking {} as successfully deployed to {}'s {} environment", appVersion, deliveryConfig.name, environment.name)
-        artifactRepository.markAsSuccessfullyDeployedTo(
-          deliveryConfig,
-          desired.launchConfiguration.imageProvider.deliveryArtifact,
-          appVersion,
-          environment.name
-        )
+      current.forEach { serverGroup ->
```
robfletcher: The current model can now include > 1 server group, so I'm just looping across them. Possibly we need to give this more thought.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -55,62 +58,65 @@ class ServerGroupHandler(
...
     group = apiVersion.group,
-    singular = "server-group",
-    plural = "server-groups"
-  ) to ServerGroupSpec::class.java
+    singular = "cluster",
+    plural = "clusters"
+  ) to ClusterSpec::class.java
 
-  override suspend fun desired(resource: Resource<ServerGroupSpec>): ServerGroup =
+  override suspend fun desired(resource: Resource<ClusterSpec>): Set<ServerGroup> =
```
robfletcher: spec for cluster potentially resolves to multiple server groups

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -55,62 +58,65 @@ class ServerGroupHandler(
...
-    val spec = resourceDiff.desired
-    val job = when {
-      resourceDiff.isCapacityOnly() -> spec.resizeServerGroupJob(resource.serviceAccount)
-      else -> spec.createServerGroupJob(resource.serviceAccount)
-    }
+    resource: Resource<ClusterSpec>,
+    resourceDiff: ResourceDiff<Set<ServerGroup>>
+  ): List<Task> =
+    resourceDiff
+      .toIndividualDiffs()
```
robfletcher: I thought it's easiest to understand what's going on if we handle diffs for regional server groups individually

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -55,62 +58,65 @@ class ServerGroupHandler(
...
-          spec.moniker.app,
-          description,
-          listOf(Job(job["type"].toString(), job)),
-          OrchestrationTrigger(resource.id.toString())
-        ))
-      .also { log.info("Started task {} to upsert server group", it.ref) }
-      // TODO: ugleee
-      .let { listOf(Task(id = it.taskId, name = description)) }
-  }
+        // TODO: this needs to be done async
```
robfletcher: oh yeah, I should resolve this before the PR is complete. It's not super high priority, though.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -247,68 +253,80 @@ class ServerGroupHandler(
...
-              suspendedProcesses = asg.suspendedProcesses.map { ScalingProcess.valueOf(it) }.toSet()
-            ),
-            tags = asg.tags.associateBy(Tag::key, Tag::value).filterNot { it.key in DEFAULT_TAGS }
-          )
+  private suspend fun CloudDriverService.getServerGroups(resource: Resource<ClusterSpec>): Set<ServerGroup> =
+    coroutineScope {
+      resource.spec.locations.regions.map {
+        async {
+          try {
+            activeServerGroup(
```
robfletcher: For simplicity I've currently made this code use the same River endpoint as before so we're making _n_ calls to River per cluster. There may be a way to optimize this later but I consider that out of scope for this PR.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -32,32 +34,42 @@ class ImageResolver(
...
-      is ArtifactImageProvider -> {
-        resolveFromArtifact(resource, imageProvider)
-      }
-      is JenkinsImageProvider -> {
-        resolveFromJenkinsJob(resource, imageProvider)
-      }
-      else -> {
-        throw UnsupportedStrategy(imageProvider::class.simpleName.orEmpty(), ImageProvider::class.simpleName.orEmpty())
-      }
+  suspend fun resolveImageId(resource: Resource<ClusterSpec>): ResolvedImages =
```
robfletcher: Note change of return type. This method can resolve the images for each region. I'm happier that there's a proper class for this as well.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -72,47 +84,65 @@ class ImageResolver(
...
 
+  private fun NamedImage.toResolvedImage(regions: Collection<String>) =
+    amis
+      .filterKeys { it in regions }
+      .mapValues { (_, amis) -> amis?.first() }
+      .filterNotNullValues()
+      .also {
+        if (it.size < regions.size) {
+          throw NoImageFoundForRegions(imageName, regions - it.keys)
+        }
```
robfletcher: Detect once if we are missing an image for one or more regions rather than potentially throwing an exception for the first detected missing image and then failing to detect any further ones.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -72,47 +84,65 @@ class ImageResolver(
...
+      }
+
+  @Suppress("UNCHECKED_CAST")
+  private fun <K, V> Map<out K, V?>.filterNotNullValues(): Map<K, V> =
+    filterValues { it != null } as Map<K, V>
```
robfletcher: Why is this not a thing in the Kotlin stdlib?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -72,47 +84,65 @@ class ImageResolver(
...
+
   private inline fun <reified T> DynamicConfigService.getConfig(configName: String, defaultValue: T) =
     getConfig(T::class.java, configName, defaultValue)
 }
+
+data class ResolvedImages(
+  val appVersion: String,
+  val imagesByRegion: Map<String, String>
+)
```
robfletcher: This is the model `ImageResolver` now returns.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpecTests.kt
```diff
@@ -0,0 +1,200 @@
...
+import strikt.api.expect
+import strikt.api.expectThat
+import strikt.assertions.all
+import strikt.assertions.contains
+import strikt.assertions.hasSize
+import strikt.assertions.isEqualTo
+import strikt.assertions.propertiesAreEqualTo
+import java.time.Duration
+
+internal class ClusterSpecTests : JUnit5Minutests {
```
robfletcher: These tests are basically for the cluster -> set<server group> resolution

keel-sql/src/main/resources/db/changelog/20190924-rename-server-group-back-to-cluster-sigh.yml
```diff
@@ -0,0 +1,46 @@
+databaseChangeLog:
+  - changeSet:
+      id: rename-server-group-back-to-cluster-sigh
```
robfletcher: I think this is not going to work given the structure of a `ClusterSpec` is different to old `ServerGroupSpec`. I think we're actually going to have to delete any server group  resources and recreate them as clusters.

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -71,16 +71,11 @@ class ResourceTagger(
 
   private val entityTypeTransforms = mapOf(
     "classic-load-balancer" to "loadbalancer",
-    "application-load-balancer" to "loadbalancer",
-    "server-group" to "cluster" // this is a hack
+    "application-load-balancer" to "loadbalancer"
```
robfletcher: Backing this out as we don't need to treat clusters in a special way any more

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -200,14 +195,20 @@ class ResourceTagger(
       account
     }
 
+    val (region, resourceId) = if (suffix.contains(":")) {
+      suffix.split(":", limit = 2).toList()
+    } else {
+      listOf("*", suffix)
+    }
```
robfletcher: Cluster resource ids no longer include a region. In the entity tag that should be reflected as `*` for the region field.

keel-sql/src/main/resources/db/databaseChangeLog.yml
```diff
@@ -32,9 +32,6 @@ databaseChangeLog:
   - include:
       file: changelog/20190904-rename-name-to-id.yml
       relativeToChangelogFile: true
-  - include:
-      file: changelog/20190917-rename-cluster-to-server-group.yml
-      relativeToChangelogFile: true
```
robfletcher: I removed this previous migration as it no longer serves any purpose. Given the change in structure between `ServerGroupSpec` and `ClusterSpec` I don't think there's any point in trying to migrate existing resources, we should just re-create them.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/actuation/ArtifactPromotionListener.kt
robfletcher: Raised #465 to track this. I think this PR is big enough already!

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
```diff
@@ -32,32 +34,42 @@ class ImageResolver(
...
+      imageName = imageProvider.imageId,
+      account = null,
+      region = null
+    )
     check(images.size == 1) {
       "Expected a single image for AMI id ${imageProvider.imageId} but found ${images.size}"
     }
-    return imageProvider.imageId to images.first().appVersion
+
+    // TODO: this is not going to work for multiple regions
```
emjburns: Want to raise an issue to track this?

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/ServerGroupMonikerTests.kt
```diff
@@ -8,34 +8,34 @@ import strikt.assertions.isEqualTo
 
 internal class ServerGroupMonikerTests : JUnit5Minutests {
```
emjburns: should the class be renamed?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ImageResolver.kt
robfletcher: There is one: #461

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/ServerGroupMonikerTests.kt
robfletcher: Fixed

PR #475, opened 2019-09-27
Tidy up determination of source server group name when annealing cluster

robfletcher:
Sorry, this is a bit of a grab-bag of various bits of tidying preliminary to #465 (making environments require all regions to be deployed before considering an artifact version successfully deployed). In the interests of keeping things simple I thought I'd get this in first so the subsequent PR is easier to parse.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/model/Moniker.kt
```diff
@@ -18,14 +18,14 @@ package com.netflix.spinnaker.keel.model
...
+import com.netflix.frigga.Names
 
 @JsonInclude(NON_NULL)
 data class Moniker(
   val app: String,
   val stack: String? = null,
   val detail: String? = null,
-  val sequence: String? = null,
-  val cluster: String? = null
+  val sequence: String? = null
```
robfletcher: The `cluster` field as I understand it is always derived from the others, so it doesn't seem to make sense for us to store it. We never use it anyway.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ServerGroup.kt
```diff
@@ -17,18 +17,30 @@
...
+  @get:ObjectDiffProperty(inclusion = EXCLUDED)
+  val name: String,
   val location: Location,
   val launchConfiguration: LaunchConfiguration,
   val capacity: Capacity = Capacity(1, 1, 1),
   val dependencies: Dependencies = Dependencies(),
   val health: Health = Health(),
   val scaling: Scaling = Scaling(),
   val tags: Map<String, String> = emptyMap()
-) : Monikered {
```
robfletcher: `ServerGroup` is not a `ResourceSpec` i.e. we do not allow people to manage a server group by itself, it's the resolved representation of a single region of a `ClusterSpec`. Therefore it should not extend `Monikered` as that in turn extends `ResourceSpec`.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ServerGroup.kt
```diff
@@ -17,18 +17,30 @@
...
-  override val id = "${location.accountName}:${location.region}:${moniker.name}"
```
robfletcher: `ServerGroup` should definitely not have an `id`, this is just going to confuse someone in the future.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/image/ArtifactVersionDeployed.kt
```diff
@@ -17,12 +17,14 @@
...
 
+import com.netflix.spinnaker.keel.api.ResourceId
+
 /**
  * An event fired while we check a server group that lets us know
  * what version of software is running in that server group
  */
-data class ArtifactAlreadyDeployedEvent(
-  val resourceId: String,
+data class ArtifactVersionDeployed(
```
robfletcher: I want to move to this event being the sole trigger for a version being considered successfully deployed so renaming it seemed the right thing.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -89,8 +88,8 @@ class ClusterHandler(
         .map { diff ->
           val spec = diff.desired
           val job = when {
-            diff.isCapacityOnly() -> spec.resizeServerGroupJob(resource.serviceAccount)
-            else -> spec.createServerGroupJob(resource.serviceAccount)
+            diff.isCapacityOnly() -> diff.resizeServerGroupJob()
+            else -> diff.createServerGroupJob()
```
robfletcher: service account was being passed in here so that we could look up the current server group name from CloudDriver. That is now on the `current` property of the diff (we just looked it up milliseconds before so doing it again seems a waste).

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -133,130 +132,114 @@ class ClusterHandler(
...
+      "region" to current.location.region,
+      "serverGroupName" to current.name
+    )
+  }
 
   override suspend fun delete(resource: Resource<ClusterSpec>) {
     TODO("not implemented")
   }
 
-  private suspend fun CloudDriverService.getAncestorServerGroup(spec: ServerGroup, serviceAccount: String): ActiveServerGroup? =
```
robfletcher: No longer need to look this up as we have it already from having determined current cluster state.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -115,7 +115,7 @@ data class ClusterSpec(
     health: HealthSpec?,
     scaling: Scaling?,
     tags: Map<String, String>?,
-    overrides: Map<String, ServerGroupSpec>
+    overrides: Map<String, ServerGroupSpec> = emptyMap()
```
robfletcher: One thing I noticed when creating cluster definitions is I needed to include `overrides: {}` in the YAML. I suspect overrides will only rarely need to be specified so not requiring users to specify it at all if they don't need to is the right thing to do.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/ServerGroupTests.kt
```diff
@@ -1,77 +0,0 @@
...
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.module.kotlin.readValue
-import com.netflix.spinnaker.keel.serialization.configuredYamlMapper
-import dev.minutest.junit.JUnit5Minutests
-import dev.minutest.rootContext
-import strikt.api.expectThat
-import strikt.assertions.isEqualTo
-import java.time.Duration
-
-internal object ServerGroupTests : JUnit5Minutests {
```
robfletcher: This test is just cruft. We have the relevant coverage in `ClusterSpecTests`

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -287,14 +270,22 @@ class ClusterHandler(
       }
         .mapNotNull { it.await() }
         .toSet()
+//        .also { them ->
```
asher: Looks like this can be fully removed?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
emjburns: yes this is definitely better

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
robfletcher: That's actually the change I want to make next. Didn't mean to commit it.

PR #477, opened 2019-09-30
represent a cluster's server groups as a Map not a Set

robfletcher:
Diff doesn't work as nicely when comparing a `Set<ServerGroup>` as it does with a `Map<String, ServerGroup>`. In particular ignored fields still register as a diff, which causes false deltas in this case.

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -17,6 +17,7 @@ import com.netflix.spinnaker.keel.api.ec2.Scaling
 import com.netflix.spinnaker.keel.api.ec2.ScalingProcess
 import com.netflix.spinnaker.keel.api.ec2.ServerGroup
 import com.netflix.spinnaker.keel.api.ec2.TerminationPolicy
+import com.netflix.spinnaker.keel.api.ec2.byRegion
 import com.netflix.spinnaker.keel.api.ec2.moniker
 import com.netflix.spinnaker.keel.api.ec2.resolve
 import com.netflix.spinnaker.keel.api.id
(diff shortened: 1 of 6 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -57,7 +58,7 @@ class ClusterHandler(
   private val publisher: ApplicationEventPublisher,
   override val objectMapper: ObjectMapper,
   override val normalizers: List<ResourceNormalizer<*>>
-) : ResolvableResourceHandler<ClusterSpec, Set<ServerGroup>> {
+) : ResolvableResourceHandler<ClusterSpec, Map<String, ServerGroup>> {
```
robfletcher: Handler now resolves `ClusterSpec` to a map by region. This diffs much better as we can tell which region is missing and get a field-by-field diff. Also see #473

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/ServerGroupDiffTests.kt
```diff
@@ -0,0 +1,108 @@
...
+import de.danielbechler.diff.node.DiffNode.State.CHANGED
+import de.danielbechler.diff.node.DiffNode.State.UNTOUCHED
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import org.apache.commons.lang3.RandomStringUtils.randomAlphanumeric
+import org.apache.commons.lang3.RandomStringUtils.randomNumeric
+import strikt.api.expectThat
+import strikt.assertions.isEqualTo
+
+internal class ServerGroupDiffTests : JUnit5Minutests {
```
robfletcher: Some tests to see if delta detection works as expected

PR #478, opened 2019-09-30
feat(securityGroup): multi-region resource definitions

asher:
- This is a breaking change for securityGroup resource definitions that standardizes on a `locations` object as used in `ClusterSpec` to contain account/region assignments. A single-region securityGroup called "keeldemo" that was previously defined in part as:
```
---
kind: "security-group"
metadata:
  serviceAccount: "delivery@spinnaker"
spec:
  moniker:
    app: keeldemo
  accountName: test
  region: us-west-2
...
```
becomes:
```
---
kind: "security-group"
metadata:
  serviceAccount: "delivery@spinnaker"
spec:
  moniker:
    app: keeldemo
  locations:
    accountName: test
    regions:
    - us-west-2
...
```

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroup.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroup.kt
@@ -1,33 +1,20 @@
-/*
- * Copyright 2018 Netflix, Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
 package com.netflix.spinnaker.keel.api.ec2
 
-import com.netflix.spinnaker.keel.api.Monikered
 import com.netflix.spinnaker.keel.model.Moniker
-import de.danielbechler.diff.inclusion.Inclusion.EXCLUDED
+import de.danielbechler.diff.inclusion.Inclusion
 import de.danielbechler.diff.introspection.ObjectDiffProperty
 
 data class SecurityGroup(
-  override val moniker: Moniker,
-  val accountName: String,
-  val region: String,
+  val moniker: Moniker,
+  val location: Location,
   val vpcName: String?,
-  @get:ObjectDiffProperty(inclusion = EXCLUDED)
+  @get:ObjectDiffProperty(inclusion = Inclusion.EXCLUDED)
   val description: String?,
   val inboundRules: Set<SecurityGroupRule> = emptySet()
-) : Monikered {
-  override val id = "$accountName:$region:${moniker.name}"
+) {
+
+  data class Location(
+    val accountName: String,
+    val region: String
+  )
 }
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupSpec.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupSpec.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.keel.api.ec2
+
+import com.netflix.spinnaker.keel.api.MultiRegion
+import com.netflix.spinnaker.keel.model.Moniker
+import de.danielbechler.diff.inclusion.Inclusion.EXCLUDED
+import de.danielbechler.diff.introspection.ObjectDiffProperty
+
+data class SecurityGroupSpec(
+  override val moniker: Moniker,
+  val locations: Locations,
+  val vpcName: String?,
+  val description: String?,
+  val inboundRules: Set<SecurityGroupRule> = emptySet(),
+  val overrides: Map<String, SecurityGroupOverride> = emptyMap()
+) : MultiRegion {
+  override val id = "${locations.accountName}:${moniker.name}"
+
+  override val regionalIds = locations.regions.map { region ->
+    "${locations.accountName}:$region:${moniker.name}"
+  }.sorted()
+
+  data class Locations(
+    val accountName: String,
+    val regions: Set<String>
+  )
+}
+
+data class SecurityGroupOverride(
+  val vpcName: String? = null,
+  @get:ObjectDiffProperty(inclusion = EXCLUDED)
+  val description: String? = null,
+  val inboundRules: Set<SecurityGroupRule>? = null
+)
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
@@ -25,6 +25,7 @@ import com.netflix.spinnaker.keel.api.ec2.CrossAccountReferenceRule
 import com.netflix.spinnaker.keel.api.ec2.PortRange
 import com.netflix.spinnaker.keel.api.ec2.ReferenceRule
 import com.netflix.spinnaker.keel.api.ec2.SecurityGroup
+import com.netflix.spinnaker.keel.api.ec2.SecurityGroupSpec
 import com.netflix.spinnaker.keel.api.ec2.SecurityGroupRule
 import com.netflix.spinnaker.keel.api.ec2.SecurityGroupRule.Protocol
 import com.netflix.spinnaker.keel.api.ec2.SelfReferenceRule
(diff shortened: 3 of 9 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -63,150 +66,185 @@ class SecurityGroupHandler(
...
+          when (diff.current) {
+            null -> {
+              job = spec.toCreateJob()
+              verb = "Create"
+            }
+            else -> {
+              job = spec.toUpdateJob()
+              verb = "Update"
+            }
+          }
```
asher: A diff may include a mix of create and updates (i.e. when a new region is added to an existing resource definition); this ensures a newly added region gets the create treatment, required to support self-referential ingress rules.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -63,150 +66,185 @@ class SecurityGroupHandler(
...
+        .filter { diff -> diff.hasChanges() }
+        .map { diff ->
+          val spec = diff.desired
+          val job: Job
+          val verb: String
+
+          when (diff.current) {
+            null -> {
+              job = spec.toCreateJob()
+              verb = "Create"
```
emjburns: nit: `Creating` and `Updating`? `Creating security group using task: blah`

emjburns: oh you also use the verb below. never mind, it's fine

emjburns:
Does tagging work with this multi-region security group? You can check by locally creating things, seeing if the tag is actually on the resource, and then making sure the resource id is correct (easiest way is to check the url when you click on the UI buttons and then do a get locally).

asher:
> Does tagging work with this multi-region security group? You can check by locally creating things, seeing if the tag is actually on the resource, and then making sure the resource id is correct (easiest way is to check the url when you click on the UI buttons and then do a get locally).

A single entity tag is created for multi-region resources, with a glob in the region field of the tag id. Opened https://github.com/spinnaker/keel/issues/481 for UI support on tags like this. Alternatively, we could change keel to generate regional tags on individual resources, but I like the simplicity of a 1:1 mapping between a `KeelResourceId` and a tag id.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -283,7 +306,7 @@ class SecurityGroupHandler(
         "vpcId" to cloudDriverCache.networkBy(
           vpcName,
           account,
-          spec.region
+          spec.locations.regions.first()
```
robfletcher: Doesn't this need to map to all the regions, not just the first?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -63,150 +66,185 @@ class SecurityGroupHandler(
...
-          resource.serviceAccount,
-          OrchestrationRequest(
-            description,
-            spec.moniker.app,
-            description,
-            listOf(spec.toCreateJob()),
-            OrchestrationTrigger(resource.id.toString())
+  override suspend fun desired(resource: Resource<SecurityGroup>): Map<String, SecurityGroup> =
+    with(resource.spec) {
+      locations.regions.map { region ->
```
robfletcher: You can use `associateWith` rather than `map…toMap` I think

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
asher: A `SecurityGroup` spec which may address multiple regions always resolves to a `Map<String, SecurityGroup>` of single region resources. I didn't want to have a combination of `regions: Set<String>` and `region: String` properties in the spec but the assumption is that regions is a single item set regardless of what's in the spec everywhere you see `regions.first()`. Definitely needs some additional tests to validate.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
asher: Went ahead and made verb a Pair with both tenses.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -53,7 +56,7 @@ class SecurityGroupHandler(
   private val orcaService: OrcaService,
   override val objectMapper: ObjectMapper,
   override val normalizers: List<ResourceNormalizer<*>>
-) : ResourceHandler<SecurityGroup> {
+) : ResolvableResourceHandler<SecurityGroup, Map<String, SecurityGroup>> {
```
robfletcher: Should we use different types here like the difference between `ClusterSpec` and `ServerGroup`? It seems confusing that the handler can return multiple single-region security groups, but their type makes it look like they could contain multiple regions.

asher: Went ahead and refactored this to split out `SecurityGroupSpec` (multi-region) from `SecurityGroup` (single-region). After doing it, I agree this is cleaner and makes sense going forward.

robfletcher: Awesome

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroup.kt
```diff
@@ -1,33 +1,22 @@
...
-  val region: String,
+  val location: SecurityGroup.Location,
   val vpcName: String?,
-  @get:ObjectDiffProperty(inclusion = EXCLUDED)
+  @get:ObjectDiffProperty(inclusion = Inclusion.EXCLUDED)
   val description: String?,
   val inboundRules: Set<SecurityGroupRule> = emptySet()
 ) : Monikered {
-  override val id = "$accountName:$region:${moniker.name}"
+  override val id = "${location.accountName}:${location.region}:${moniker.name}"
```
robfletcher: I think this class should no longer extend `ResourceSpec` if it's now the resolved form. It shouldn't need an id. You can't have a `Resource<SecurityGroup>`, you must have `Resource<SecurityGroupSpec>`. I made the same change to `ServerGroup` when I did multi-region clusters.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupSpec.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.netflix.spinnaker.keel.api.MultiRegion
+import com.netflix.spinnaker.keel.model.Moniker
+import de.danielbechler.diff.inclusion.Inclusion.EXCLUDED
+import de.danielbechler.diff.introspection.ObjectDiffProperty
+
+data class SecurityGroupSpec(
+  override val moniker: Moniker,
+  val locations: Locations,
+  val vpcName: String?,
+  @get:ObjectDiffProperty(inclusion = EXCLUDED)
```
robfletcher: Shouldn't need this as we won't be diffing these directly. We'll diff the `SecurityGroup` it resolves to.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -53,160 +57,190 @@ class SecurityGroupHandler(
   private val orcaService: OrcaService,
   override val objectMapper: ObjectMapper,
   override val normalizers: List<ResourceNormalizer<*>>
-) : ResourceHandler<SecurityGroup> {
+) : ResolvableResourceHandler<SecurityGroupSpec, Map<String, SecurityGroup>> {
```
robfletcher: 👍 I'm gonna change these interface names to encourage us to use this one more often.

robfletcher: See #491 (I won't merge until after this goes in)

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroup.kt
```diff
@@ -1,33 +1,22 @@
-/*
```
emjburns: Any reason you deleted this header?

asher: The file and class `SecurityGroup` was renamed to `SecurityGroupSpec` which still has the header. This is a brand new file. I'm under the impression we no longer add Netflix copyright headers to new files?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroup.kt
asher: Great point, will change. I was concerned about these having their own ids.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroup.kt
emjburns: Oh. 🤷‍♀ I thought we did! No idea.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroup.kt
robfletcher: We should possibly decouple `Monikered` from `ResourceSpec` (but if so let's do it as its own change)

PR #480, opened 2019-10-01
fix occasional test breakage due to Moniker

robfletcher:


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/model/Moniker.kt
```diff
@@ -38,10 +38,13 @@ data class Moniker(
...
-      stack == null && detail == null -> "$app-v$sequence"
-      detail == null && sequence != null -> "$app-$stack-v$sequence"
-      else -> "$app-${stack.orEmpty()}-$detail-v${sequence?.padStart(3, '0')}"
+      sequence == null -> name
+      stack == null && detail == null -> "$app-v$sequenceString"
+      detail == null -> "$app-$stack-v$sequenceString"
+      else -> "$app-${stack.orEmpty()}-$detail-v$sequenceString"
     }
+
+  private val sequenceString = sequence?.toString()?.padStart(3, '0')
```
cfieber: should mod 1000 this I think

robfletcher: or reject any out of range number?

PR #482, opened 2019-10-01
anonymized a bunch of test data

robfletcher:


cfieber:
@robzienert made a thing that I think would work for this:

cfieber:
specifically this thinger has some AWS specific randomizers:
https://github.com/spinnaker/kork/blob/master/kork-test/src/main/java/com/netflix/spinnaker/kork/test/mimicker/producers/aws/AwsProducer.java

robfletcher:
Future enhancement

PR #483, opened 2019-10-01
fix(artifact): status as strings, remove RELEASE

emjburns:
Seemed like calling this endpoint was broken if we pass the status as enums. Also, the rocket endpoint doesn't like bad statuses.

This was causing keel to not find new versions when registering a new image.


emjburns:
I'm not exactly sure. Serialization something?

PR #485, opened 2019-10-02
Assign AZs to clusters automatically if not specified

### LINKED ISSUE #386: Default availability zones in a cluster to "all"
Users should not have to specify all availability zones explicitly. The default should be to balance across all AZs.

robfletcher:
Fixes #386


robfletcher:
I think `CloudDriverCache` could do with some love, but outside of this PR.

Issue #442, opened 2019-09-13
Notifications


emjburns:
Resources should have some way to notify 
- when the resource changes (start/end of actuation?)
- maybe when there's an error
- maybe when there's a diff detected

We could default to emailing the service account (at netflix, this is usually the team's DL).
We could provide notification config for each resource.
We could provide notification config as an overall "resource" or "config", tie it to an app, and have some options on this to control which events we're notifying for as well as who is being notified.

emjburns:
The rocket team has started with this config in the `rocket.yml` file:

```yml
notifications:
  stash:
    enabled: true|false
  slack:
  - channel: my-team-channel
    rocketBranch: master
    conditions:
    - failure
  - channel: my-other-channel
    rocketBranch: pull/\d+
    conditions:
    - failure
    - success
    - aborted
  email:
  - to:
    - myteam-critical@netflix.com
    rocketBranch: master
    conditions:
    - failure
  - to:
    - myteam-noncritical@netflix.com
    - devnull@netflix.com
    rocketBranch: pull/\d+
    conditions:
    - failure
    - success
    - aborted
```

emjburns:
Preliminary idea: an environment can have notifications (same for every resource that supports them). Config is a list of notifications at the environment level.

```yaml
notifications:
- type: slack
  address: #my-slack-channel
  frequency: VERBOSE / NORMAL / QUIET
- type: email
  address: my-email@blah.com
  frequency: QUIET
```
where:
VERBOSE: starting, complete, failed
NORMAL: complete, failed
QUIET: failed

robfletcher:
Seems legit

emjburns:
That's my question as well.

PR #486, opened 2019-10-02
feat(notifications): add env notifications for infra tasks

emjburns:
Closes https://github.com/spinnaker/keel/issues/442 (Details about format in the issue)

Environments can now have notifications, they are applied at the environment level (examples in the ticket). For now we support either slack or email, and you can't define your own custom message (`:rainbow: Managed update succeeded`). 

The environment level notification applies to every resource who cares (load balancers, security groups, clusters). You can't override it or shut it off for a specific resource (for now).

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/NotificationConfig.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/NotificationConfig.kt
@@ -1,0 +1,39 @@
+/*
+ *
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.api
+
+// todo eb: this does not allow you to customize the notification message, but we can add that later.
+data class NotificationConfig(
+  val type: NotificationType,
+  val address: String, // either slack channel or email address
+  val frequency: NotificationFrequency
+)
+
+enum class NotificationFrequency {
+  VERBOSE, // notification on task starting, completing, failing
+  NORMAL, // notification on task completing or failing
+  QUIET // notification only for failure
+}
+
+enum class NotificationType {
+  SLACK, EMAIL;
+
+  override fun toString(): String {
+    return name.toLowerCase()
+  }
+}
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
@@ -2,8 +2,8 @@ package com.netflix.spinnaker.keel.ec2.resource
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.keel.api.Resource
-import com.netflix.spinnaker.keel.api.ResourceKind
 import com.netflix.spinnaker.keel.api.ResourceId
+import com.netflix.spinnaker.keel.api.ResourceKind
 import com.netflix.spinnaker.keel.api.SPINNAKER_API_V1
 import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancer
 import com.netflix.spinnaker.keel.api.ec2.LoadBalancerType
@@ -31,6 +31,7 @@ class ApplicationLoadBalancerHandler(
   private val cloudDriverService: CloudDriverService,
   private val cloudDriverCache: CloudDriverCache,
   private val orcaService: OrcaService,
+  private val environmentResolver: EnvironmentResolver,
   override val objectMapper: ObjectMapper,
   override val normalizers: List<ResourceNormalizer<*>>
 ) : ResourceHandler<ApplicationLoadBalancer> {
@@ -58,6 +59,8 @@ class ApplicationLoadBalancerHandler(
     val description = "$action ${resource.kind} load balancer ${resource.spec.moniker.name} in " +
       "${resource.spec.location.accountName}/${resource.spec.location.region}"
 
+    val notifications = environmentResolver.getNotificationsFor(resource.id)
+
     val taskRef =
       resource.spec.let { spec ->
 
@@ -69,7 +72,7 @@ class ApplicationLoadBalancerHandler(
               spec.moniker.app,
               description,
               listOf(spec.toUpsertJob()),
-              OrchestrationTrigger(resource.id.toString())
+              OrchestrationTrigger(correlationId = resource.id.toString(), notifications = notifications)
             )
           )
       }
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
@@ -85,6 +85,8 @@ class ApplicationLoadBalancerHandler(
     val description = "Delete load balancer ${resource.spec.moniker.name} in " +
       "${resource.spec.location.accountName}/${resource.spec.location.region}"
 
+    val notifications = environmentResolver.getNotificationsFor(resource.id)
+
     val taskRef =
       resource.spec.let { spec ->
         orcaService
@@ -95,7 +97,7 @@ class ApplicationLoadBalancerHandler(
               spec.moniker.app,
               description,
               listOf(spec.toDeleteJob()),
-              OrchestrationTrigger(resource.id.toString())
+              OrchestrationTrigger(correlationId = resource.id.toString(), notifications = emptyList())
             )
           )
       }
--- a/keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandlerTest.kt
+++ b/keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandlerTest.kt
@@ -16,6 +16,7 @@ import com.netflix.spinnaker.keel.ec2.normalizers.ApplicationLoadBalancerNormali
 import com.netflix.spinnaker.keel.model.OrchestrationRequest
 import com.netflix.spinnaker.keel.orca.OrcaService
 import com.netflix.spinnaker.keel.orca.TaskRefResponse
+import com.netflix.spinnaker.keel.persistence.memory.InMemoryDeliveryConfigRepository
 import com.netflix.spinnaker.keel.plugin.ResourceNormalizer
 import com.netflix.spinnaker.keel.serialization.configuredYamlMapper
 import com.netflix.spinnaker.keel.test.resource
@@ -32,6 +33,7 @@ import kotlinx.coroutines.runBlocking
 import strikt.api.expectThat
 import strikt.assertions.get
 import strikt.assertions.isEqualTo
+import java.time.Clock
 import java.util.UUID
 
 @Suppress("UNCHECKED_CAST")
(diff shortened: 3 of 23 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/NotificationConfig.kt
```diff
@@ -0,0 +1,39 @@
...
+package com.netflix.spinnaker.keel.api
+
+// todo eb: this does not allow you to customize the notification message, but we can add that later.
+data class NotificationConfig(
+  val type: NotificationType,
+  val address: String, // either slack channel or email address
+  val frequency: NotificationFrequency
+)
+
+enum class NotificationFrequency {
```
emjburns: I don't know if this is the right strategy, but I do know that I don't want users to have to configure notifications for "when a pipeline starts" because I want to hide these things from them.

Maybe it's better to just have a list of events that you can notify on, like `ACUTATION_STARTING`, `ACUTATION_COMPLETE`, `ACUTATION_FAILED`? However, if we call out explicit states, we can never add more or less. I sort of like the idea of just specifying the level and then figuring out what fits in there... 

Thoughts?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
```diff
@@ -58,6 +59,8 @@ class ApplicationLoadBalancerHandler(
     val description = "$action ${resource.kind} load balancer ${resource.spec.moniker.name} in " +
       "${resource.spec.location.accountName}/${resource.spec.location.region}"
 
+    val notifications = environmentResolver.getNotificationsFor(resource.id)
```
emjburns: I thought it was nicer to put things that dealt with fetching environments and re-shaping the contents in one place where all the handlers could use it.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/model/EchoNotification.kt
```diff
@@ -0,0 +1,77 @@
...
+
+import com.netflix.spinnaker.keel.api.NotificationConfig
+import com.netflix.spinnaker.keel.api.NotificationFrequency
+import com.netflix.spinnaker.keel.api.NotificationFrequency.NORMAL
+import com.netflix.spinnaker.keel.api.NotificationFrequency.QUIET
+import com.netflix.spinnaker.keel.api.NotificationFrequency.VERBOSE
+import com.netflix.spinnaker.keel.model.NotificationEvent.ORCHESTRATION_COMPLETE
+import com.netflix.spinnaker.keel.model.NotificationEvent.ORCHESTRATION_FAILED
+import com.netflix.spinnaker.keel.model.NotificationEvent.ORCHESTRATION_STARTING
+
```
emjburns: All these things are in service of giving `echo` the exact map payload that it so desperately wants.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -115,14 +117,15 @@ class SqlDeliveryConfigRepository(
...
               .from(ENVIRONMENT)
               .where(ENVIRONMENT.DELIVERY_CONFIG_UID.eq(uid))
-              .fetch { (environmentUid, name, constraintsJson) ->
+              .fetch { (environmentUid, name, constraintsJson, notificationsJson) ->
                 Environment(
                   name = name,
                   resources = resourcesForEnvironment(environmentUid),
-                  constraints = mapper.readValue(constraintsJson)
+                  constraints = mapper.readValue(constraintsJson),
+                  notifications = mapper.readValue(notificationsJson ?: "[]")
```
emjburns: You can't [give a default value to a `text` field](https://bugs.mysql.com/bug.php?id=21532), so I'm handling the null case for this field here so it's easy to roll this out.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/NotificationConfig.kt
cfieber: could add `DEFAULT` to the enum and make frequency in `NotificationConfig` a `Map<NotificationFrequency, Boolean>` as to whether to notify on a particular state (maybe it should be renamed to `NotificationStatus` or something in that case)

so I could have:
```
frequency:
  DEFAULT: true
  ACTUATION_STARTING: false
```

And I would get notified on complete/failed. And I'd also get notified on any future state we added until I explicitly opted out of it.

Actually come to think of it maybe the default should always be true, and we have a list of states to explicitly suppress..?

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandlerTest.kt
```diff
@@ -32,13 +33,16 @@ import kotlinx.coroutines.runBlocking
...
 import strikt.assertions.isEqualTo
+import java.time.Clock
 import java.util.UUID
 
 @Suppress("UNCHECKED_CAST")
 internal class ApplicationLoadBalancerHandlerTests : JUnit5Minutests {
   private val cloudDriverService = mockk<CloudDriverService>()
   private val cloudDriverCache = mockk<CloudDriverCache>()
   private val orcaService = mockk<OrcaService>()
+  private val clock = Clock.systemDefaultZone()
```
cfieber: I don't see this being used anywhere?

cfieber: herp de derp

cfieber: use your eyeballs @cfieber

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
cfieber: good call

keel-sql/src/main/resources/db/changelog/20191002-add-env-notifications.yml
```diff
@@ -0,0 +1,17 @@
...
+      author: emjburns
+      changes:
+        - addColumn:
+            tableName: environment
+            columns:
+              - column:
+                  name: notifications
+                  type: text
+                  constraints:
+                    nullable: true
```
cfieber: I think we can have a future migration that populates this column with `"[]"` where null and adds a nullable false once we are past the point of having to roll back to prior to the existence of this column.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/NotificationConfig.kt
erikmunson: I don't really have much of an opinion on the `DEFAULT` thing, or map vs. not map. Feels like the need for specifying exact types of notification events will come up later, but could easily see adding another field (parallel to `frequency`) that would let you turn on/off event types.

robfletcher: I'm not sure I understand why it means we can't add things in the future. We can always run migrations or have defaults for event types we add in future. I'm concerned that describing it in the sense of levels might be too obscure for people or we'll get bugged for "I want NORMAL, except…". But I think this is a great start.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
```diff
@@ -69,7 +72,7 @@ class ApplicationLoadBalancerHandler(
               spec.moniker.app,
               description,
               listOf(spec.toUpsertJob()),
-              OrchestrationTrigger(resource.id.toString())
+              OrchestrationTrigger(correlationId = resource.id.toString(), notifications = notifications)
```
robfletcher: I wonder if we can do this in a centralized place. I worry we'll forget to do notifications for a particular resource type or something stupid. (Future change, just thinking aloud).

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/model/EchoNotification.kt
```diff
@@ -0,0 +1,77 @@
...
+  val address: String,
+  val `when`: List<String>,
+  val level: String = "pipeline",
+  val message: Map<String, NotificationMessage>
+)
+
+enum class NotificationEvent {
+  ORCHESTRATION_STARTING {
+    override fun text(): String = "orchestration.starting"
+    override fun notificationMessage(): NotificationMessage = NotificationMessage("$RAINBOW Managed update starting")
```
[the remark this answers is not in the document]
robfletcher: I definitely think it would make sense to provide context around _what_ is being updated. Ideally a link too. A standard method on `ResourceSpec` to give a human-readable description of what it is?

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/model/EchoNotification.kt
```diff
@@ -0,0 +1,77 @@
...
+  },
+  ORCHESTRATION_COMPLETE {
+    override fun text(): String = "orchestration.complete"
+    override fun notificationMessage(): NotificationMessage = NotificationMessage("$RAINBOW Managed update succeeded")
+  },
+  ORCHESTRATION_FAILED {
+    override fun text(): String = "orchestration.failed"
+    override fun notificationMessage(): NotificationMessage = NotificationMessage("$RAINBOW Managed update failed")
```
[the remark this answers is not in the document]
emjburns: LOL. I was just thinking of using the same messaging to tie it together "this is a pipeline that you didn't know about, but oh it has the rainbow, it must be a managed thing".

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/model/EchoNotification.kt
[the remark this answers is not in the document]
emjburns: This will already provide a link to the orchestration(task) in the ui - same way as normal pipeline notification works. FYI this is the tiny notification that pops up in a slack message.

Screen Shot 2019-10-03 at 1 45 09 PM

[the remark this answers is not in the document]
emjburns: That's basically what it looks like, and the task description is a clickable link to the task in the UI. Given that information, what more should we include? (this is easy to update in the future)

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
emjburns: Hmm good point. I could just make it required to pass in, that way you have to think about it? that would probably solve the issue enough to help ward off future mistakes.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/NotificationConfig.kt
emjburns: I just mean not easily - Like, if you have to specify "starting"/"complete"/"failed"/etc events it's hard to keep everyone in "recommenced" settings. 

I think, let's tackle this in the future with maybe something like `overrides` as a parallel field to frequency, where someone could add or subtract notifications to their hearts content.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
emjburns: Oo nice call. Found a place where I missed one (security group create).

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/model/EchoNotification.kt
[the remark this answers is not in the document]
emjburns: I'll switch to the thundercloud for failed.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
robfletcher: Let's think about it in a future change.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/model/EchoNotification.kt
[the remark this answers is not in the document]
robfletcher: Looks good to me

PR #487, opened 2019-10-02
feat(multiregion): support for region-specific ids/events from multi-region specs

asher:


keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -195,8 +238,8 @@ class ResourceTagger(
       account
     }
 
-    val (region, resourceId) = if (suffix.contains(":")) {
-      suffix.split(":", limit = 2).toList()
+    val (region, resourceId) = if (toString().count { it == ':' } == 4) {
```
asher: This fixes a recently introduced bug where tags are improperly generated except when region is `*`

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -83,25 +83,53 @@ class ResourceTagger(
 
   @EventListener(ResourceCreated::class)
   fun onCreateEvent(event: ResourceCreated) {
-    if (event.resourceId.shouldTag()) {
+    if (!event.resourceId.shouldTag()) {
+      return
+    }
+    /**
+     * If a MultiRegion resource provides a set of regional resourceIds, entity tags will be created
+     * against those ids only, not the
```
erikmunson: is there a missing part of this comment sentence?

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -83,25 +83,53 @@ class ResourceTagger(
...
+     */
+    if (event.resourceIds.isEmpty()) {
       log.debug("Persisting tag desired for resource {} because it exists now", event.resourceId)
       val spec = event.resourceId.generateKeelTagSpec()
       persistTagState(spec)
     }
+
+    event.resourceIds.forEach { id ->
```
emjburns: I know that if the above condition is true (`resourceIds.isEmpty()`) then this will not run, but it still feels weird to me that it's not in an `else {` block, or there's not a return in the above `if`. Sort of makes me do a double take in a confusing way.

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -83,25 +83,53 @@ class ResourceTagger(
...
       val entityRef = event.resourceId.toEntityRef()
       val spec = KeelTagSpec(
         keelId = event.resourceId,
         entityRef = entityRef,
-        tagState = TagNotDesired(startTime = clock.millis())
+        tagState = TagNotDesired(startTime = now)
       )
       persistTagState(spec)
     }
+
```
emjburns: same comment as above for here

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
asher: 😱

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -83,25 +83,53 @@ class ResourceTagger(
...
+    if (!event.resourceId.shouldTag()) {
+      return
+    }
+    /**
+     * If a MultiRegion resource provides a set of regional resourceIds, entity tags will be created
+     * against those ids only, not the
+     */
+    if (event.resourceIds.isEmpty()) {
       log.debug("Persisting tag desired for resource {} because it exists now", event.resourceId)
       val spec = event.resourceId.generateKeelTagSpec()
```
emjburns: Does this need to have its own id passed into it now? Or maybe I'm confused...

emjburns:
This is overall excellent. 

Does it correctly remove the tags if the resource gets deleted? (i.e the `onDeleteEvent()`). I think it will just try and remove the parent tag vs all the child tags

keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
asher: It doesn't, the unparameterized `generateKeelTagSpec()` is still present (now on line 192) and refers to its own id. It just doesn't appear in the diff.

asher:
@emjburns excellent catch re: delete handling, fixed.

PR #489, opened 2019-10-03
AZs assigned by a normalizer rather than in class init

robfletcher:


emjburns:
I definitely like this better than what we were doing before. 

We talked offline about how we could get these normalizers (or something like them) to run every time a resources is evaluated, which I think would definitely be better. In that case we'd have to make sure we had some way of presenting this to the user - what did we do for you? what does your spec look like? what does the 'full version' look like?

Issue #488, opened 2019-10-03
Expanded resource metadata in application summary endpoint


erikmunson:
We currently have [an endpoint](https://github.com/spinnaker/keel/blob/master/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt) whose purpose is to serve up summary information about a specific application as it relates to Keel. The object that endpoint returns has a couple fields today: `hasManagedResources` which is a boolean, and `resources` which is an optionally included list of all the resource IDs that map to an app.

In the UI, we're now reaching a point where we need to get fresh (i.e. near-real-time) and slightly more detailed resource metadata for all the resources in an application, which we can then join together on the client with data from clouddriver. We've gotten pretty far relying on entity tags for indicating infrastructure is managed in the UI, but I don't believe they will serve us well as requirements grow and evolve (especially in the real-time cases). I propose we expand the existing application summary endpoint to include a new, **summarized** resource description that can be used by the UI without needing to download and deal with the entirety of resource definitions or make large numbers of fan-out API calls. Currently thinking something like:

```javascript
{
  hasManagedResources: true,
  resources: [
    // These would be some kind of ResourceSummary type
    {
      id: 'keel:resource:id',
      kind: 'cluster',
      status: 'HAPPY',
      moniker: {
        app: 'emburnstest',
        stack: 'prestaging',
        detail: 'foo'
      },
      locations: {
        // accountName can be handled in a generic way, but implicitly allows us
        // to find which cloud provider a resource is attached to
        accountName: 'prod',
        // Summarized regions, rather than full cloud provider-specific
        // metadata about regions/subnets/zones/etc.
        regions: ['us-west-2', 'us-east-1']
      }
    }
  ]
}
```
I'd want to stick to a few overall goals here:
- **Provide enough identifying metadata to do client side joins on clouddriver infrastructure**: This is basically kind + account + region + moniker info
- **Avoid cloud provider-specific identifiers if possible**: Across the rest of our APIs, we generally treat the concepts of `account` and `region` as being generic and applicable to any cloud provider. Many providers have their own interesting specifics (i.e. zones, subnets, etc.), but there is a lot of code in Deck (and I presume other services) that understands account + region at a core level and can then choose to delegate provider-specific concerns beneath that out to different modules. It'd be great to be able to do that here as well (as a relevant example, entity tags will apply themselves to infra with a matching account + region + moniker, regardless of provider)
[...]

PR #494, opened 2019-10-04
feat(api): provide resource summary info by app

emjburns:
For the UI views for managed delivery we need an endpoint (by app) to poll for status and info. 

This implementation is expensive - we recalculate the status on every call. There's a todo in there to fix that. 

`GET https://localhost:8087/application/emburnstest?includeDetails=true` 
```json
{
  "hasManagedResources": true,
  "resources": [
    {
      "id": "ec2:cluster:test:emburnstest-md-wow",
      "kind": "cluster",
      "status": "ERROR",
      "moniker": {
        "app": "emburnstest",
        "stack": "md",
        "detail": "wow"
      },
      "locations": {
        "accountName": "test",
        "regions": [
          "us-west-1"
        ]
      }
    },
    {
      "id": "ec2:security-group:test:emburnstest-managed",
      "kind": "security-group",
      "status": "HAPPY",
      "moniker": {
        "app": "emburnstest",
        "stack": "managed"
      },
      "locations": {
        "accountName": "test",
        "regions": [
          "us-west-1"
        ]
      }
    }
  ]
}
```

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
@@ -1,0 +1,42 @@
+/*
+ *
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.api
+
+/**
+ * An object which is located in an account and region
+ */
+interface Locatable {
+  val locations: Locations<out RegionSpec>
+}
+
+open class Locations<T : RegionSpec>(
+  val accountName: String,
+  val regions: Set<T>
+) {
+  override fun equals(other: Any?): Boolean {
+    return other is Locations<*> &&
+      accountName == other.accountName &&
+      regions == other.regions
+  }
+
+  override fun hashCode(): Int {
+    var result = accountName.hashCode()
+    result = 31 * result + regions.hashCode()
+    return result
+  }
+}
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
@@ -24,19 +24,7 @@ interface Locatable {
   val locations: Locations<out RegionSpec>
 }
 
-open class Locations<T : RegionSpec>(
+data class Locations<T : RegionSpec>(
   val accountName: String,
   val regions: Set<T>
-) {
-  override fun equals(other: Any?): Boolean {
-    return other is Locations<*> &&
-      accountName == other.accountName &&
-      regions == other.regions
-  }
-
-  override fun hashCode(): Int {
-    var result = accountName.hashCode()
-    result = 31 * result + regions.hashCode()
-    return result
-  }
-}
+)
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/model/SimpleRegionSpec.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/model/SimpleRegionSpec.kt
@@ -1,0 +1,41 @@
+/*
+ *
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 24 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.api
+
```
emjburns: This interface is for resources that can be located by account and region.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+interface Locatable {
+  val locations: Locations<out RegionSpec>
+}
+
+open class Locations<T : RegionSpec>(
+  val accountName: String,
+  val regions: Set<T>
+) {
+  override fun equals(other: Any?): Boolean {
```
emjburns: This can't be a data class because things override it, and equals comparison is wonky if we don't implement an equals function.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/model/SimpleRegionSpec.kt
```diff
@@ -0,0 +1,41 @@
...
+package com.netflix.spinnaker.keel.model
+
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
+import com.fasterxml.jackson.databind.annotation.JsonSerialize
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer
+import com.fasterxml.jackson.databind.ser.std.ToStringSerializer
+import com.netflix.spinnaker.keel.api.RegionSpec
+
```
emjburns: Simple region representation parses to a string (you can see that in the example response body above.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/model/SubnetAwareRegionSpec.kt
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.model
+
+import com.netflix.spinnaker.keel.api.RegionSpec
+
```
emjburns: This is used for clusters, and in the future load balancers. This extra info is not returned in the summary though, because we don't care at that level.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -150,6 +160,26 @@ interface ResourceRepository : PeriodicallyCheckedRepository<ResourceHeader> {
     return true
   }
 
+  fun Resource<*>.toResourceSummary() =
+    ResourceSummary(
+      id = id,
+      kind = kind,
+      status = getStatus(id), // todo eb: this will become expensive
```
emjburns: This makes another database call to calculated the status. When this becomes a problem (or soon) I think I'll just create a table for the status and every time we write a history event for a resource we then update the status.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -133,20 +136,6 @@ data class ClusterSpec(
     overrides
   )
 
-  data class Locations(
```
emjburns: These are removed and replaced with the general form.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
robfletcher: Nothing extends this. Things extend `RegionSpec` but there's no different implementations of `Locations`. I think it _can_ be a data class.

emjburns: Changing! Great point. This is much nicer.

emjburns:
Closes https://github.com/spinnaker/keel/issues/488

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
emjburns: We can't just read a single status, mainly to detect flapping (10 events which only contain 'delta detected' then 'actuation launched', in pairs). Is that the best way to determine this? Probably not. But that's what it's doing now.

emjburns: We could probably optimize by first fetching one event, then if needed fetching 10. I'll discuss this with you offline and make any changes in a separate PR.

PR #497, opened 2019-10-07
Make ImageResolver a standard Resolver

### LINKED ISSUE #461: Make `ImageIdProvider` work with multi-region clusters
We should probably use image name rather than id or if that's really what people want to do, require that it be a single-region cluster.

### LINKED ISSUE #496: Convert existing ImageResolver to conform to new Resolver interface

robfletcher:
Fixes #496
Fixes #461

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/config/EC2Config.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/config/EC2Config.kt
@@ -81,7 +81,6 @@ class EC2Config {
       cloudDriverService,
       cloudDriverCache,
       orcaService,
-      imageResolver,
       environmentResolver,
       clock,
       publisher,
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/config/EC2Config.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/config/EC2Config.kt
@@ -19,18 +19,14 @@ package com.netflix.spinnaker.config
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.keel.clouddriver.CloudDriverCache
 import com.netflix.spinnaker.keel.clouddriver.CloudDriverService
-import com.netflix.spinnaker.keel.clouddriver.ImageService
 import com.netflix.spinnaker.keel.ec2.resource.ApplicationLoadBalancerHandler
 import com.netflix.spinnaker.keel.ec2.resource.ClassicLoadBalancerHandler
 import com.netflix.spinnaker.keel.ec2.resource.ClusterHandler
 import com.netflix.spinnaker.keel.ec2.resource.EnvironmentResolver
-import com.netflix.spinnaker.keel.ec2.resource.ImageResolver
 import com.netflix.spinnaker.keel.ec2.resource.SecurityGroupHandler
 import com.netflix.spinnaker.keel.orca.OrcaService
-import com.netflix.spinnaker.keel.persistence.ArtifactRepository
 import com.netflix.spinnaker.keel.persistence.DeliveryConfigRepository
 import com.netflix.spinnaker.keel.plugin.Resolver
-import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.context.annotation.Bean
@@ -40,22 +36,6 @@ import java.time.Clock
 @Configuration
 @ConditionalOnProperty("keel.plugins.ec2.enabled")
 class EC2Config {
-
-  @Bean
-  fun imageResolver(
-    dynamicConfigService: DynamicConfigService,
-    cloudDriverService: CloudDriverService,
-    deliveryConfigRepository: DeliveryConfigRepository,
-    artifactRepository: ArtifactRepository,
-    imageService: ImageService
-  ): ImageResolver =
-    ImageResolver(
-      dynamicConfigService,
-      cloudDriverService,
-      deliveryConfigRepository,
-      artifactRepository,
-      imageService
-    )
 
   @Bean
   fun environmentResolver(
@@ -70,7 +50,6 @@ class EC2Config {
     cloudDriverService: CloudDriverService,
     cloudDriverCache: CloudDriverCache,
     orcaService: OrcaService,
-    imageResolver: ImageResolver,
     environmentResolver: EnvironmentResolver,
     clock: Clock,
     objectMapper: ObjectMapper,
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
@@ -6,7 +6,6 @@ import com.fasterxml.jackson.annotation.JsonUnwrapped
 import com.netflix.spinnaker.keel.api.Locatable
 import com.netflix.spinnaker.keel.api.Locations
 import com.netflix.spinnaker.keel.api.MultiRegion
-import com.netflix.spinnaker.keel.ec2.resource.ResolvedImages
 import com.netflix.spinnaker.keel.model.Moniker
 import com.netflix.spinnaker.keel.model.SubnetAwareRegionSpec
 import java.time.Duration
@@ -14,7 +13,7 @@ import java.time.Duration
 /**
  * Transforms a [ClusterSpec] into a concrete model of server group desired states.
  */
-fun ClusterSpec.resolve(resolvedImages: ResolvedImages): Set<ServerGroup> =
+fun ClusterSpec.resolve(): Set<ServerGroup> =
   locations.regions.map {
     ServerGroup(
       name = moniker.name,
@@ -24,11 +23,7 @@ fun ClusterSpec.resolve(resolvedImages: ResolvedImages): Set<ServerGroup> =
         it.subnet,
         it.availabilityZones
       ),
-      launchConfiguration = resolveLaunchConfiguration(
-        it.region,
-        resolvedImages.appVersion,
-        checkNotNull(resolvedImages.imagesByRegion[it.region]) { "No image resolved for ${it.region}" }
-      ),
+      launchConfiguration = resolveLaunchConfiguration(it),
       capacity = resolveCapacity(it.region),
       dependencies = resolveDependencies(it.region),
       health = resolveHealth(it.region),
@@ -38,23 +33,26 @@ fun ClusterSpec.resolve(resolvedImages: ResolvedImages): Set<ServerGroup> =
   }
     .toSet()
 
-private fun ClusterSpec.resolveLaunchConfiguration(region: String, appVersion: String, imageId: String): LaunchConfiguration =
-  LaunchConfiguration(
-    appVersion = appVersion,
-    imageId = imageId,
-    instanceType = checkNotNull(overrides[region]?.launchConfiguration?.instanceType
+private fun ClusterSpec.resolveLaunchConfiguration(region: SubnetAwareRegionSpec): LaunchConfiguration {
+  val image = checkNotNull(overrides[region.region]?.launchConfiguration?.image
+    ?: defaults.launchConfiguration?.image) { "No image resolved / specified for ${region.region}" }
+  return LaunchConfiguration(
+    appVersion = image.appVersion,
+    imageId = image.id,
+    instanceType = checkNotNull(overrides[region.region]?.launchConfiguration?.instanceType
       ?: defaults.launchConfiguration?.instanceType),
-    ebsOptimized = checkNotNull(overrides[region]?.launchConfiguration?.ebsOptimized
+    ebsOptimized = checkNotNull(overrides[region.region]?.launchConfiguration?.ebsOptimized
       ?: defaults.launchConfiguration?.ebsOptimized),
-    iamRole = checkNotNull(overrides[region]?.launchConfiguration?.iamRole
+    iamRole = checkNotNull(overrides[region.region]?.launchConfiguration?.iamRole
       ?: defaults.launchConfiguration?.iamRole),
-    keyPair = checkNotNull(overrides[region]?.launchConfiguration?.keyPair
+    keyPair = checkNotNull(overrides[region.region]?.launchConfiguration?.keyPair
       ?: defaults.launchConfiguration?.keyPair),
-    instanceMonitoring = overrides[region]?.launchConfiguration?.instanceMonitoring
+    instanceMonitoring = overrides[region.region]?.launchConfiguration?.instanceMonitoring
       ?: defaults.launchConfiguration?.instanceMonitoring ?: false,
-    ramdiskId = overrides[region]?.launchConfiguration?.ramdiskId
+    ramdiskId = overrides[region.region]?.launchConfiguration?.ramdiskId
       ?: defaults.launchConfiguration?.ramdiskId
   )
+}
 
 internal fun ClusterSpec.resolveCapacity(region: String) =
   overrides[region]?.capacity ?: defaults.capacity ?: Capacity(1, 1, 1)
@@ -87,7 +85,7 @@ private fun ClusterSpec.resolveHealth(region: String): Health {
 
 data class ClusterSpec(
   override val moniker: Moniker,
-  val imageProvider: ImageProvider,
+  val imageProvider: ImageProvider? = null,
   override val locations: Locations<SubnetAwareRegionSpec>,
   private val _defaults: ServerGroupSpec,
   val overrides: Map<String, ServerGroupSpec> = emptyMap()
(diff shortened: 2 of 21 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -4,14 +4,13 @@ import com.fasterxml.jackson.annotation.JsonCreator
...
 import com.netflix.spinnaker.keel.api.MultiRegion
-import com.netflix.spinnaker.keel.ec2.resource.ResolvedImages
 import com.netflix.spinnaker.keel.model.Moniker
 import java.time.Duration
 
 /**
  * Transforms a [ClusterSpec] into a concrete model of server group desired states.
  */
-fun ClusterSpec.resolve(resolvedImages: ResolvedImages): Set<ServerGroup> =
+fun ClusterSpec.resolve(): Set<ServerGroup> =
```
robfletcher: By this time `ImageResolver` should have applied specific AMI ids in each region

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -84,7 +81,7 @@ private fun ClusterSpec.resolveHealth(region: String): Health {
 
 data class ClusterSpec(
   override val moniker: Moniker,
-  val imageProvider: ImageProvider,
+  val imageProvider: ImageProvider? = null,
```
robfletcher: Optional if you’re specifying images in each region directly

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -165,6 +166,11 @@ data class ClusterSpec(
     val ramdiskId: String? = null
   )
 
+  data class VirtualMachineImage(
+    val id: String,
+    val appVersion: String
+  )
```
robfletcher: Supplied directly in submitted resource or resolved by `ImageResolver`

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ImageProvider.kt
```diff
@@ -29,14 +29,6 @@ import com.netflix.spinnaker.keel.ec2.jackson.ImageProviderDeserializer
 @JsonDeserialize(using = ImageProviderDeserializer::class)
 sealed class ImageProvider
 
-/**
- * Provides the ami id for the image
- */
-@JsonDeserialize(using = JsonDeserializer.None::class)
-data class IdImageProvider(
-  val imageId: String
-) : ImageProvider()
```
robfletcher: No longer used since this is not compatible with multi region clusters. Instead image id can be specified directly in each region if this is what you want.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/normalizers/ImageResolver.kt
```diff
@@ -1,75 +1,59 @@
...
 
+@Component
 class ImageResolver(
   private val dynamicConfigService: DynamicConfigService,
-  private val cloudDriverService: CloudDriverService,
   private val deliveryConfigRepository: DeliveryConfigRepository,
   private val artifactRepository: ArtifactRepository,
   private val imageService: ImageService
-) {
+) : Resolver<ClusterSpec> {
```
robfletcher: Image resolution now works like any other resolver

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/config/EC2Config.kt
```diff
@@ -41,22 +37,6 @@ import java.time.Clock
...
-    artifactRepository: ArtifactRepository,
-    imageService: ImageService
-  ): ImageResolver =
-    ImageResolver(
-      dynamicConfigService,
-      cloudDriverService,
-      deliveryConfigRepository,
-      artifactRepository,
-      imageService
-    )
```
robfletcher: Just declaring this using `@Component` now

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -35,23 +30,25 @@ fun ClusterSpec.resolve(resolvedImages: ResolvedImages): Set<ServerGroup> =
   }
     .toSet()
 
-private fun ClusterSpec.resolveLaunchConfiguration(region: String, appVersion: String, imageId: String): LaunchConfiguration =
-  LaunchConfiguration(
-    appVersion = appVersion,
-    imageId = imageId,
-    instanceType = checkNotNull(overrides[region]?.launchConfiguration?.instanceType
+private fun ClusterSpec.resolveLaunchConfiguration(region: ClusterSpec.ClusterRegion): LaunchConfiguration {
+  val image = checkNotNull(region.image) { "No image resolved / specified for ${region.region}" }
```
robfletcher: Exception will get thrown here if image was neither specified or resolved

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -138,13 +135,17 @@ data class ClusterSpec(
...
   data class ClusterRegion(
     val region: String,
     val subnet: String,
     /**
      * If empty this implies the cluster should use all availability zones.
      */
-    val availabilityZones: Set<String> = emptySet()
+    val availabilityZones: Set<String> = emptySet(),
+    val image: VirtualMachineImage? = null
   )
```
robfletcher: `ClusterRegion` now has an optional `image` property. The 99% case will have this getting resolved using the `ImageProvider`, but if you really want to you could stick a particular AMI id on here.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -1,75 +1,55 @@
...
 
+@Component
 class ImageResolver(
   private val dynamicConfigService: DynamicConfigService,
-  private val cloudDriverService: CloudDriverService,
   private val deliveryConfigRepository: DeliveryConfigRepository,
   private val artifactRepository: ArtifactRepository,
   private val imageService: ImageService
-) {
+) : Resolver<ClusterSpec> {
```
robfletcher: `ImageResolver` is now a standard `Resolver`. It automatically gets wired into the `ClusterHandler` and contributes to desired state resolution in the same way as any other resolver.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -68,8 +67,7 @@ class ClusterHandler(
 
   override suspend fun toResolvedType(resource: Resource<ClusterSpec>): Map<String, ServerGroup> =
     with(resource.spec) {
-      val resolvedImages = imageResolver.resolveImageId(resource)
-      resolve(resolvedImages).byRegion()
+      resolve().byRegion()
```
robfletcher: `ClusterHandler` no longer treats `ImageResolver` as a special case - it's just another resolver.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpecTests.kt
```diff
@@ -183,29 +112,75 @@ private data class ResolvedSpecFixture(
...
+        ),
+        ClusterRegion(
+          region = "us-west-2",
+          subnet = "internal (vpc0)",
+          availabilityZones = setOf("us-west-2a", "us-west-2b", "us-west-2c"),
+          image = VirtualMachineImage(
+            appVersion = appVersion,
+            id = usWestImageId
+          )
+        )
```
robfletcher: When testing the code that converts `ClusterSpec` into `Map<String, ServerGroup>` we don't have to worry about cases for different `ImageProvider` types as whatever type is getting used, the AMI id would have been resolved by now.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolverTests.kt
```diff
@@ -147,34 +149,18 @@ internal class ImageResolverTests : JUnit5Minutests {
...
-        val resolved = resolve()
-        expect {
-          that(resolved.appVersion).isEqualTo("fnord-$version2")
-          that(resolved.imagesByRegion).isEqualTo(mapOf(imageRegion to imageProvider.imageId))
-        }
+      test("returns the original spec unchanged") {
+        expectThat(resolve())
+          .propertiesAreEqualTo(resource)
       }
     }
```
robfletcher: `ClusterSpec.imageProvider` being `null` is now a thing. `ImageResolver` should just return the unmodified spec in that case.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -38,23 +33,26 @@ fun ClusterSpec.resolve(resolvedImages: ResolvedImages): Set<ServerGroup> =
...
     .toSet()
 
-private fun ClusterSpec.resolveLaunchConfiguration(region: String, appVersion: String, imageId: String): LaunchConfiguration =
-  LaunchConfiguration(
-    appVersion = appVersion,
-    imageId = imageId,
-    instanceType = checkNotNull(overrides[region]?.launchConfiguration?.instanceType
+private fun ClusterSpec.resolveLaunchConfiguration(region: SubnetAwareRegionSpec): LaunchConfiguration {
+  val image = checkNotNull(overrides[region.region]?.launchConfiguration?.image
+    ?: defaults.launchConfiguration?.image) { "No image resolved / specified for ${region.region}" }
```
luispollo: Can the default for launch configuration image be null? Am I reading this right?

robfletcher: Yes, because most specs will not want to specify an image explicitly. They'll get it from an artifact.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -1,75 +1,58 @@
...
-    val images = cloudDriverService.namedImages(
-      serviceAccount = DEFAULT_SERVICE_ACCOUNT,
-      imageName = imageProvider.imageId,
-      account = null,
-      region = null
-    )
-    check(images.size == 1) {
-      "Expected a single image for AMI id ${imageProvider.imageId} but found ${images.size}"
+  override fun invoke(resource: Resource<ClusterSpec>): Resource<ClusterSpec> {
+    val imageProvider = resource.spec.imageProvider ?: return resource
```
luispollo: Newbie Kotlin idiom question: if I read this right, here we just return the `resource` as is (i.e. no "resolving") if `resource.spec.imageProvider` is `null`, right? The `else` block if we were writing this in another language would generally have a logging statement in this case, is my guess. Do we just return quietly here because this case doesn't matter much, because it's just the idiom in Kotlin, or something else? 😃

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -1,75 +1,58 @@
...
+    val image = runBlocking {
```
luispollo: Curious about the use of `runBlocking` here... The docs say "it is designed to bridge regular blocking code to libraries that are written in suspending style, to be used in main functions and in tests". Are the `resolve*` function below written "in suspending style"?

luispollo: Never mind -- I need to do my homework and read more on this.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
robfletcher: If the resource has no `imageProvider` this resolver has no work to do so it just returns the resource unchanged. I'm not sure what logging would add here.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
robfletcher: yeah, the various calls to CloudDriver are suspending functions.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -117,22 +90,35 @@ class ImageResolver(
...
+    overrides.putAll(spec.overrides)
+    spec.locations.regions.map { it.region }.forEach { region ->
+      overrides[region] = overrides[region].withVirtualMachineImage(VirtualMachineImage(imageIdByRegion.getValue(region), image.appVersion))
+    }
+
+    return copy(spec = spec.copy(overrides = overrides))
+  }
+
+  private fun ServerGroupSpec?.withVirtualMachineImage(image: VirtualMachineImage) =
+    (this ?: ServerGroupSpec()).run {
```
luispollo: Sorry, another newbie question: can you help me read this? If `this` is `null` then run the block?! When would that ever happen?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
luispollo: Got it.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
robfletcher: Yes, the receiver of the function is declared as nullable, so `this` can be `null` here. It's a pretty neat feature of Kotlin that you can do that. If the `?` were omitted from the function's receiver then the compiler would not allow this since it knows `this` could _not_ be `null`.

robfletcher: Basically, in this specific case it's saying "use either `this` or a default new instance". This is because image ids are per region but it's probably unusual for a user to be overriding anything else in the server group spec on a per-region basis, so the `ServerGroupSpec` for any given region is likely not defined.

emjburns: I would be in favor of changing this function so it's a bit clearer, even if it's longer. The way this is written it's hard for me to understand as well.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -38,23 +33,26 @@ fun ClusterSpec.resolve(resolvedImages: ResolvedImages): Set<ServerGroup> =
...
+  return LaunchConfiguration(
+    appVersion = image.appVersion,
+    imageId = image.id,
+    instanceType = checkNotNull(overrides[region.region]?.launchConfiguration?.instanceType
       ?: defaults.launchConfiguration?.instanceType),
-    ebsOptimized = checkNotNull(overrides[region]?.launchConfiguration?.ebsOptimized
+    ebsOptimized = checkNotNull(overrides[region.region]?.launchConfiguration?.ebsOptimized
```
gal-yardeni: what does `region.region` means?

robfletcher: It's a little awkward that the `ClusterRegion` object contains various properties, one of which is the region name. It used to be just a String before this change but now has other properties such as AZs, vpc subnet…

PR #500, opened 2019-10-08
fix(resolver): log more context when calculating desired fails

asher:


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -93,7 +93,7 @@ class ResourceActuator(
         }
       }
     } catch (e: Exception) {
-      log.error("Resource check for {} failed due to \"{}\"", id, e.message)
+      log.error("Resource check for {} failed due to \"{}\" {}", id, e.message, e.cause?.message ?: "")
```
robfletcher: Ha, I was just in the middle of making a very similar change

robfletcher:
This is a good start, but I want to take it further so the event stored in the resource history also contains the appropriate cause message

PR #501, opened 2019-10-08
try to capture some notion of cause in ResourceCheckError

robfletcher:


robfletcher:
@asher building on #500 this should give us a fighting chance of being able to see what went wrong from the resource history API endpoint

PR #503, opened 2019-10-10
feat(examples): yaml that will break if you change the spec

emjburns:
I've created examples of valid yaml files that will be objectmappered to their corresponding java classes every time tests are run. This should help us keep a current folder of yaml examples that is less likely to break.


robfletcher:
Nice, this will be super useful.

PR #507, opened 2019-10-14
feat(status): include resource valid events in history

emjburns:
Right now if we have a "resource check error" we persist an error state. If we resolve this state but the outcome is that the resource is valid we stay in an "error" state (the last history event is an error). This PR changes the history a bit so that we persist one `ResourceValid` status if the resource is valid.

The event gets sent [here](https://github.com/spinnaker/keel/blob/684f9e546a5879cdce7c0cbfe615686df58200ba/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt#L88-L92)

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
@@ -82,12 +82,6 @@ sealed class ResourceEvent {
     get() = ids?.map { ResourceId(it) }?.toSet() ?: emptySet()
 
   /**
-   * Should the event be recorded in a resource's history?
-   */
-  @JsonIgnore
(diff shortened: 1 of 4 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -81,12 +81,6 @@ sealed class ResourceEvent {
     @JsonIgnore
     get() = ids?.map { ResourceId(it) }?.toSet() ?: emptySet()
 
-  /**
-   * Should the event be recorded in a resource's history?
-   */
-  @JsonIgnore
-  open val ignoreInHistory: Boolean = false
```
emjburns: `ResourceValid` was the only event that used this.

PR #511, opened 2019-10-15
Rationalize vpc in specs

### LINKED ISSUE #504: Rationalize specification of VPC and subnet in resource specs
Specification of locations should include…

* `vpcName` e.g. "vpc0" for clusters / security groups / load balancers
* `subnetPurpose` e.g. "internal" for clusters / load balancers

We can then determine the VPC id (which we need to send to Orca) from those.

robfletcher:
I was bothered by the way the specification of VPC was inconsistent between resource types. Would love to know if this stab at making it more sensible is legit.

Potentially fixes #504


keel-api/src/test/resources/examples/alb-example.yml
```diff
@@ -4,14 +4,14 @@ kind: application-load-balancer
...
 spec:
-  vpcName: vpc0
   subnetType: internal (vpc0)
   moniker:
     app: keeldemo
     stack: example
     detail: albec2v1
   locations:
     accountName: test
+    vpcName: vpc0
```
robfletcher: Generally moving `vpcName` from the root of the spec into the `locations`

keel-api/src/test/resources/examples/cluster-example.yml
```diff
@@ -14,6 +14,7 @@ spec:
       type: DEB
   locations:
     accountName: test
+    vpcName: vpc0
```
robfletcher: I ended up adding this to cluster for consistency. Not sure if that's desirable

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Location.kt
```diff
@@ -20,6 +20,7 @@ package com.netflix.spinnaker.keel.api.ec2
 data class Location(
   val accountName: String,
   val region: String,
-  val subnet: String?, // TODO: is this actually optional?
+  val vpcName: String,
+  val subnet: String,
```
robfletcher: VPC and subnet now required

keel-api/src/test/resources/examples/alb-example.yml
```diff
@@ -4,14 +4,13 @@ kind: application-load-balancer
...
     regions:
     - region: us-east-1
       subnet: internal (vpc0)
```
robfletcher: Generally moving `vpcName` from the root of the spec into the `locations` and `subnet` to each individual region.

keel-api/src/test/resources/examples/cluster-example.yml
asher: Can this replace the need for `subnet` in the regions map?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Location.kt
asher: This seems wrong, since a `Subnet` always includes a `vpcId`.

keel-api/src/test/resources/examples/cluster-example.yml
robfletcher: No, I think a VPC has many subnets as I understand it

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Location.kt
robfletcher: An alternative approach I considered was to look up the VPC based on the subnet when needed. Because we have subnets with no "purpose" property (which is what this `subnet` property maps to) I wasn't sure that would cover edge cases. Then again, I guess this doesn't either.

keel-api/src/test/resources/examples/cluster-example.yml
asher: A VPC can have many subnets but I don't think that prevents us from attempting to provide an opinionated default. In the absence of either a subnet or vpc, we could attempt to default to a subnet of "internal (vpc0)" which will correlate to a vpcId if it exists. Given just a `vpcName`, we could attempt to find a subnet name matching `"internal ($vpcName)"`.

I think it makes more sense not to require `vpcName` or `subnet`, with both derived by default from a subnet with the purpose of `internal (vpc0)` if present. If you need to deploy elsewhere, only specifying a different subnet should always be sufficient. There are subnets without a listed purpose but I don't think those are properly supported by clouddriver anyways, they need to be tagged first.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Location.kt
asher: I'm 95% sure we can ignore subnets without a purpose for now.

keel-api/src/test/resources/examples/cluster-example.yml
robfletcher: That all makes sense. I guess my only question is whether that reflects a Netflix-only opinion of network configuration. Can/should we assume that's also valid for OSS users? I guess we could make those defaults configurable down the road.

asher: I think Spinnaker is opinionated enough that a "purpose" is required for use in OSS, but not about its contents. What do you think of Spring configuration properties for potential defaults like this that aren't easily resolved by "all the things available", like AZ's?

robfletcher: Yes, that's where I was heading. Although, I feel like I'd like to do that as a separate piece of work.

luispollo: BTW, I couldn't find any information in the OSS docs about how VPCs and subnets are supposed to be setup, what tags are expected to exist on those resources for Spinnaker to find them (including the naming conventions above), etc. The closest thing I found was [this](https://www.spinnaker.io/setup/install/providers/aws/aws-ec2/) but it only discusses setting up the accounts.

keel-api/src/test/resources/examples/alb-example.yml
luispollo: Do we foresee a case where a user would deploy the same thing in subnets with different purposes in different regions (say `internal (vpc0)` in `us-east-1` and `external (vpc0)` in `us-west-2`)? Seems unlikely...

luispollo: In fact, we already do that for load balancers in the UI with the guided setup:
image

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Location.kt
erikmunson: how does this play with security groups not needing a subnet?

robfletcher: They can just specify the VPC name or have it defaulted

keel-api/src/test/resources/examples/alb-example.yml
robfletcher: You mean `subnet` could be moved up from the individual region to sit next to `vpcName`? What do you think @asher ?

asher: This seems right. The resolved subnet might differ across both regions and individual AZ's, but all should map back to the same named purpose.

erikmunson: don't have an informed opinion on whether subnet should be regional or not in these specs, but worth thinking about whether abstracting up to `internal` vs. `external` like we do in Luis's UI shots should really be at this level or at another level that can automatically resolve all the appropriate subnets for each region. Those two threads of conversation might be separate.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
```diff
@@ -26,5 +26,13 @@ interface Locatable<T : RegionSpec> : ResourceSpec {
...
+  /**
+   * If not specified here, this should be derived from [subnet] or use a default VPC name.
+   */
+  val vpcName: String?,
+  /**
+   * If not specified here, this should be derived from a default subnet purpose using [vpcName].
+   */
+  val subnet: String?,
   val regions: Set<T>
 )
```
robfletcher: VPC name and subnet (purpose) are now both here. That means

1. every region must use a VPC with the same name.
2. every region must use a subnet with the same purpose.
3. annoyingly, the `subnet` property is present even for `SecurityGroupSpec` where it will always be `null` (or at least ignored).

keel-api/src/test/resources/examples/alb-example.yml
luispollo: nit: See, @asher even said it himself above: what we're naming in the spec is basically the purpose of the subnet. 😃 Anyway, I don't have an issue with keeping the subnet names in the spec (as opposed to the more abstract "purpose"), but in that case I'd suggest at least renaming the field to `subnetName` for consistency, or removing the `Name` suffix from `vpcName`.

keel-api/src/test/resources/examples/cluster-example.yml
luispollo: For future reference (thanks @asher!): https://docs.armory.io/spinnaker-install-admin-guides/aws-subnets/#configuring-subnets

keel-api/src/test/resources/examples/alb-example.yml
luispollo: P.S. @asher Just enlightened me that the value we expect in that `subnet` field is actually the value of the `purpose` metadata tag attached to the resource. So my last point above is moot. Wondering if it would be best to name the property `subnetPurpose` for clarity, though...

robfletcher: If you hit the CloudDriver endpoint you'll see that subnets don't have a "name" as such. The closest thing is the "purpose". But yeah, I'd be down with dropping `Name` from `vpcName`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
```diff
@@ -20,11 +20,31 @@ package com.netflix.spinnaker.keel.api
...
+   */
+  val subnet: String?,
+  override val regions: Set<RegionSpec>
+) : Locations<RegionSpec>
+
+data class SimpleLocations(
+  override val accountName: String,
+  override val vpcName: String?,
+  override val regions: Set<String>
+) : Locations<String>
```
robfletcher: Instead of one type of `Locations` and 2 types of `RegionSpec` there are now 2 types of `Locations`, one of which just has regions as a `Set<String>` (so no more custom Jackson handling).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
```diff
@@ -20,11 +20,32 @@ package com.netflix.spinnaker.keel.api
...
+   */
+  val subnet: String?,
+  override val regions: Set<SubnetAwareRegionSpec>
+) : Locations<SubnetAwareRegionSpec>
+
+data class SimpleLocations(
+  override val accountName: String,
+  override val vpcName: String?,
+  override val regions: Set<SimpleRegionSpec>
+) : Locations<SimpleRegionSpec>
```
robfletcher: Instead of one type of `Locations` and 2 types of `RegionSpec` there are now 2 types of `Locations`. This leads to more consistent YAML between resource types that specify zones and those that do not.

keel-api/src/test/resources/examples/security-group-example.yml
```diff
@@ -10,13 +10,13 @@ spec:
     detail: ec2v1
   locations:
     accountName: test
+    vpcName: vpc0
     regions:
-      - us-west-2
-      - us-east-1
-  vpcName: vpc0
+    - name: us-west-2
+    - name: us-east-1
```
robfletcher: This is a change to how regions are represented in security groups. But… it means they look the same as in other resources when those other resources are using default AZs. I think this is a good thing.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/RegionSpec.kt
```diff
@@ -18,5 +18,17 @@
 package com.netflix.spinnaker.keel.api
 
 interface RegionSpec {
-  val region: String
+  val name: String
```
robfletcher: Changed this because it makes the YAML and code representations make more sense.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
```diff
@@ -18,13 +18,34 @@
 package com.netflix.spinnaker.keel.api
 
 /**
- * An object which is located in an account and region
+ * An resource spec which is located in an account and one or more regions.
```
luispollo: nit: "*A* resource spec..."

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/NetworkResolver.kt
```diff
@@ -0,0 +1,87 @@
...
+import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec
+import com.netflix.spinnaker.keel.api.ec2.ClassicLoadBalancerSpec
+import com.netflix.spinnaker.keel.api.ec2.ClusterSpec
+import com.netflix.spinnaker.keel.ec2.SPINNAKER_EC2_API_V1
+import com.netflix.spinnaker.keel.plugin.Resolver
+import org.springframework.stereotype.Component
+
+abstract class NetworkResolver<T : Locatable<*>> : Resolver<T> {
+
+  protected fun SubnetAwareLocations.withResolvedNetwork(): SubnetAwareLocations {
```
luispollo: Just curious: is it a common pattern to have a class whose sole purpose is to declare extensions for a different class like you did here? It wasn't initially clear to me that `NetworkResolver` types were being used anywhere in the code until I went back and noticed the `withResolvedNetwork` methods and the overrides to `invoke` again. 
Also, it looks like all 3 sub-types of `NetworkResolver` have the exact same override for `invoke` and the only difference is they have a different `supportedKind` but I couldn't spot where that was used either. I guess in general I was just wondering if we could have just added the code to resolve locations to the base `ResourceSpec` class (although I do realize not all resources are "locatable", but maybe in the `Locatable` interface then?) to simplify things.

robfletcher: It's because `Resolver<T>` extends the functional type `(Resource<T>) -> Resource<T>` which specifies the `invoke` method. That's an operator function. If an object has `invoke` you can use it like a function: e.g. `myThing("foo")`.

PR #513, opened 2019-10-15
feat(diff): ad hoc delivery config diffing

emjburns:
Ad-hoc diffing for delivery configs. I return a list of diffs for each environment.


robfletcher:
Oh neat. Good work.

PR #515, opened 2019-10-16
vpcName -> vpc, accountName -> account

robfletcher:
Following on from #511 this simplifies configuration a little and makes naming more consistent.

### DIFF
--- a/keel-api/src/test/resources/examples/alb-example.yml
+++ b/keel-api/src/test/resources/examples/alb-example.yml
@@ -9,8 +9,8 @@ spec:
     stack: example
     detail: albec2v1
   locations:
-    accountName: test
-    vpcName: vpc0
+    account: test
+    vpc: vpc0
     subnet: internal (vpc0)
     regions:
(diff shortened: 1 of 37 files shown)


keel-api/src/test/resources/examples/alb-example.yml
```diff
@@ -9,8 +9,8 @@ spec:
     stack: example
     detail: albec2v1
   locations:
-    accountName: test
-    vpcName: vpc0
+    account: test
```
emjburns: Accounts have two properties: name and id. is removing `Name` confusing? I always get confused when we just have `account` in various clouddriver code - sometimes it's name, sometimes it's id.

robfletcher: In Keel's API we never use ids. I agree in CloudDriver where it's ambiguous. Keel's API should never expect users to know ids.

emjburns: Okay

PR #516, opened 2019-10-16
feat(artifact): pull version on demand if none found

emjburns:
Right now if you try and calculate the desired state for an image it'll fail if the artifact isn't registered. Sure, we register the artifact for you, but that only takes affect the next time.

This PR changes the behavior slightly so that if the artifact isn't registered or it is but we don't have any versions for it we fetch the versions from igor. If we still can't find a version we throw an exception. 

This makes it possible to create an ad-hoc diff for images and clusters.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
@@ -27,8 +27,10 @@ import com.netflix.spinnaker.keel.model.OrchestrationRequest
 import com.netflix.spinnaker.keel.model.OrchestrationTrigger
 import com.netflix.spinnaker.keel.orca.OrcaService
 import com.netflix.spinnaker.keel.persistence.ArtifactRepository
-import com.netflix.spinnaker.keel.plugin.ResourceHandler
+import com.netflix.spinnaker.keel.persistence.NoSuchArtifactException
 import com.netflix.spinnaker.keel.plugin.Resolver
+import com.netflix.spinnaker.keel.plugin.ResourceHandler
+import kotlinx.coroutines.runBlocking
 import org.springframework.context.ApplicationEventPublisher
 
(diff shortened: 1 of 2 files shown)


keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -76,10 +72,34 @@ class ImageHandler(
...
-  private fun DeliveryArtifact.findLatestVersion(statuses: List<ArtifactStatus>): String =
-    artifactRepository
-      .versions(this, statuses)
-      .firstOrNull() ?: throw NoKnownArtifactVersions(this)
+  /**
+   * First checks our repo, and if a version isn't found checks igor.
+   */
+  private fun DeliveryArtifact.findLatestVersion(statuses: List<ArtifactStatus>): String {
+    try {
+      val knowVersion = artifactRepository
```
robfletcher: Should that be `knownVersion`?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -76,10 +72,34 @@ class ImageHandler(
...
+      }
+    } catch (e: NoSuchArtifactException) {
+      if (!artifactRepository.isRegistered(name, type)) {
+        // we clearly care about this artifact, let's register it.
+        publisher.publishEvent(ArtifactRegisteredEvent(this, statuses))
+      }
+    }
+
+    // even though the artifact isn't registered we should grab the latest version to use
+    return runBlocking {
```
robfletcher: So in this case we just return the latest version?

emjburns: yep, same as what we would get from the artifact repository if it had been listening to the versions.

PR #522, opened 2019-10-17
feat(resources/export): export deployed security groups as keel crds

asher:


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
```diff
@@ -60,7 +60,7 @@ class MemoryCloudDriverCache(
     .expireAfterWrite(1, HOURS)
     .build<String, Subnet>()
 
-  private fun credentialBy(name: String): Credential =
+  override fun credentialBy(name: String): Credential =
```
asher: Making this a public part of the `CloudDriverCache` interface to assist with determining the locations configured with an account.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt
```diff
@@ -156,6 +157,15 @@ abstract class ResourceHandler<S : ResourceSpec, R : Any>(
    */
   open suspend fun delete(resource: Resource<S>): List<Task> = TODO("Not implemented")
 
+  /**
+   * Generate a Keel SubmittedResource CRD from currently existing resources.
```
emjburns: Can we say `definition` here instead of CRD?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -134,11 +140,111 @@ class SecurityGroupHandler(
...
+      } catch (e: ResourceNotFound) {
+        null
+      }
+    }
+      .filterValues { it != null }
+
+    val securityGroups =
+      coroutineScope {
+        summaries.map { (region, summary) ->
+          async {
```
emjburns: If you're using `async` here, do you want to wait for everything to complete at all?

emjburns: Ah I see the `await` at the end of this, never mind. I'm still new to these fancy things.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExportController.kt
```diff
@@ -0,0 +1,78 @@
...
+    produces = [MediaType.APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  fun get(
+    @PathVariable("cloudProvider") cloudProvider: String,
+    @PathVariable("account") account: String,
+    @PathVariable("type") type: String,
+    @PathVariable("name") name: String,
+    @RequestParam("serviceAccount") serviceAccount: String
+  ): SubmittedResource<*> {
+    val apiVersion = SPINNAKER_API_V1.subApi(cloudProviderOverrides[cloudProvider] ?: cloudProvider)
```
robfletcher: I’m wondering if we should have a way of registering these API versions

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -134,11 +140,111 @@ class SecurityGroupHandler(
...
+      )
+    )
+  }
+
   private fun ResourceDiff<Map<String, SecurityGroup>>.toIndividualDiffs() =
     desired.map { (region, desire) ->
       ResourceDiff(desire, current?.getOrDefault(region, null))
     }
 
+  private fun SecurityGroupSpec.generateOverrides(
```
emjburns: oh nice!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExportController.kt
```diff
@@ -0,0 +1,78 @@
...
+    val kind = CaseFormat.LOWER_CAMEL.to(CaseFormat.LOWER_HYPHEN, type)
```
robfletcher: Why do we not just expect the kebab case in the URL?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Exportable.kt
```diff
@@ -0,0 +1,11 @@
+package com.netflix.spinnaker.keel.api
+
+import com.netflix.spinnaker.keel.model.Moniker
+
+data class Exportable(
+  val account: String,
```
robfletcher: Is `account` an EC2-specific concept? Or at least one that may not be common to all cloud providers?

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt
```diff
@@ -173,7 +183,10 @@ fun Collection<ResourceHandler<*, *>>.supporting(
   apiVersion: ApiVersion,
   kind: String
 ): ResourceHandler<*, *> =
-  find { it.apiVersion == apiVersion && it.supportedKind.first.singular == kind }
+  find {
+    it.apiVersion == apiVersion &&
+      (it.supportedKind.first.singular == kind || it.supportedKind.first.plural == kind)
```
robfletcher: Curious why this change is necessary

asher: Clouddriver types are generally plural but I don't know if this is true for all cloud providers.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Exportable.kt
asher: No, its lower level and should be required by all cloudProviders. If you deploy to 3 k8s clusters, each cluster should appear as a unique account.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExportController.kt
asher: This REST method attempts to be compatible with how the rest of Spinnaker views resources. We could expect deck or gate to convert `securityGroups` to `security-group` (and forgo the plural kind match you asked about below) or `security-groups` but I don't know if that sort of conversion will always be correct or if we should burden callers with knowledge of how clouddriver types map to keel kinds.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExportController.kt
asher: +1 (though not currently sure what it'd look like)

PR #525, opened 2019-10-18
Use name instead of requiring people to specify moniker

### LINKED ISSUE #524: Replace `moniker` property with `name` in all resource specs
Then add… 
```kotlin
val moniker: Moniker by lazy { parseMoniker(name) }
```

robfletcher:
Fixes #524

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Monikered.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Monikered.kt
@@ -21,6 +21,8 @@ package com.netflix.spinnaker.keel.api
  * A resource that has a moniker-compliant name.
  */
 interface Monikered : ResourceSpec {
+  val name: String
+
   val moniker: Moniker
 
   /**
--- a/keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/ApplicationLoadBalancerSpecTests.kt
+++ b/keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/ApplicationLoadBalancerSpecTests.kt
@@ -21,10 +21,7 @@ internal object ApplicationLoadBalancerSpecTests : JUnit5Minutests {
         Fixture(
           yaml = """
             |---
-            |moniker:
-            |  app: testapp
-            |  stack: managedogge
-            |  detail: wow
+            |name: testapp-managedogge-wow
             |locations:
             |  account: test
             |  vpc: vpc0
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ApplicationLoadBalancerSpec.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ApplicationLoadBalancerSpec.kt
@@ -3,16 +3,16 @@ package com.netflix.spinnaker.keel.api.ec2
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.annotation.JsonInclude.Include.NON_EMPTY
-import com.netflix.spinnaker.keel.api.Moniker
 import com.netflix.spinnaker.keel.api.SubnetAwareLocations
 import com.netflix.spinnaker.keel.api.ec2.LoadBalancerType.APPLICATION
 import com.netflix.spinnaker.keel.clouddriver.model.ApplicationLoadBalancerModel.Action
 import com.netflix.spinnaker.keel.clouddriver.model.ApplicationLoadBalancerModel.Rule
 import com.netflix.spinnaker.keel.clouddriver.model.ApplicationLoadBalancerModel.TargetGroupAttributes
+import com.netflix.spinnaker.keel.core.parseMoniker
 import java.time.Duration
 
 data class ApplicationLoadBalancerSpec(
-  override val moniker: Moniker,
+  override val name: String,
   override val locations: SubnetAwareLocations,
   override val internal: Boolean = true,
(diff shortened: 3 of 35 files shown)


keel-api/src/test/resources/examples/alb-example.yml
```diff
@@ -4,10 +4,7 @@ kind: application-load-balancer
 metadata:
   serviceAccount: my-email@spinnaker.io
 spec:
-  moniker:
-    app: keeldemo
-    stack: example
-    detail: albec2v1
+  name: keeldemo-example-albec2v1
```
robfletcher: Like… this is just easier

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Monikered.kt
```diff
@@ -19,12 +19,16 @@ package com.netflix.spinnaker.keel.api
...
+import com.netflix.spinnaker.keel.model.parseMoniker
 
 /**
  * An object which has a moniker
  */
 interface Monikered : ResourceSpec {
+  val name: String
+
   val moniker: Moniker
+    @JsonIgnore get() = parseMoniker(name)
```
robfletcher: The `Monikered` interface now requires a name it can parse into a `Moniker`.

asher: Nice.

emjburns:
I think this is definitely an improvement. But, `name` should be in metadata and be able to be parsed into a moniker

cfieber:
I feel like having the naming components separated into a Moniker object might play nicer with kustomize-ing things - likely the app component won't change and the stack and or detail will as you start overriding definitions.

Not sure if that is better or worse though than just having to supply the whole name in every override?

robfletcher:
> I think this is definitely an improvement. But, `name` should be in metadata and be able to be parsed into a moniker

I think you're confusing the resource's id with the name (e.g. the name of the load balancer, security group, or cluster you'd see in Deck).

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/ApplicationLoadBalancerSpecTests.kt
```diff
@@ -21,10 +21,7 @@ internal object ApplicationLoadBalancerSpecTests : JUnit5Minutests {
         Fixture(
           yaml = """
             |---
-            |moniker:
-            |  app: testapp
-            |  stack: managedogge
-            |  detail: wow
+            |name: testapp-managedogge-wow
```
gal-yardeni: so much better :)

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Monikered.kt
```diff
@@ -21,6 +21,8 @@ package com.netflix.spinnaker.keel.api
  * A resource that has a moniker-compliant name.
  */
 interface Monikered : ResourceSpec {
+  val name: String
```
emjburns: I think it would be valuable to have a validation function that took the app name, and made sure the name started with the app name. I worry about this making it really easy to have resources that break our spinnaker assumptions (by accident)

robfletcher: Although I agree, we don't do that now with the `moniker` property

robfletcher: Are there situations where it's valid to be managing a resource that's outside of the application?

emjburns: Hmm, good point that we don't validate now. I'm not sure - I think the resources we support right now need to be part of valid application. TBD on whether that will change in the future.

emjburns:
Is this backwards compatible? What's your plan to roll it out?

robfletcher:
> Is this backwards compatible? What's your plan to roll it out?

Nope. I was planning to add a migrator and then PR people's repos.

erikmunson:
I believe I communicated some of my overall feedback on this direction last year on other mediums (Slack, etc.), but will leave a note here for posterity. I don't really agree that this is better UX, and I'm not confident it makes anything easier or better for end users. Couple of thoughts:

- To echo @cfieber's original concern, while it's nice in theory to have a `name` field that folks can use with Kustomize or similar it also (I think??) makes things significantly more awkward when overriding things like stacks or details on resources — IMO it feels much more natural to override a resource's `stack` field specifically, and leave everything else untouched

- I fundamentally do not agree with the premise that combining `app` + `stack` + `detail` into a single field is easier to read, write, or work with for end users. While a unified `name` field is likely very convenient for those that are already deeply familiar with Spinnaker's naming structure and know it by heart, to the less deeply initiated I believe it's a significantly worse experience. Offering a _seemingly_ unstructured `name` field which then later (when you try to submit or validate it) blows up with errors about a very specific structure seems like bad UX, and in fact it is for that reason that all throughout Deck we model the name as its component parts with separate form fields for app + stack + detail

With all that said, this change is not a hill I want or intend to die on, and it's somewhat reversible if we change our minds later. My only request is that we are open and willing to get real user feedback going forward and don't hesitate to switch to something else if we see signs this syntax isn't working as well as we had hoped. These things are always fuzzy and subjective, and while I certainly have my own strong opinions I have no grounds to claim my view is an 'authoritative' one.

robfletcher:
I think you make some very good points, Erik. I'll give it some thought. I'm totally happy to _not_ go in this direction if it makes sense to not to.

emjburns:
I agree with Erik. I'm not opposed to changing things, but I especially agree that it's good if you're familiar with Spinnaker but potentially hellish if you're new. I'd prefer to not do this now and wait until we get feedback to update this.

PR #528, opened 2019-10-18
merged network and AZ resolvers as they are interdependent

robfletcher:


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/NetworkResolver.kt
```diff
@@ -5,24 +5,42 @@ import com.netflix.spinnaker.keel.api.Locatable
...
+            availabilityZones = cloudDriverCache.resolveAvailabilityZones(
+              account = account,
+              subnet = resolvedSubnet,
+              region = region
+            )
+          )
+        } else {
+          region
+        }
+      }.toSet()
```
robfletcher: Also resolving AZs now

PR #532, opened 2019-10-20
require vpc on locations but default to vpc0

robfletcher:
This fixes the situation where we can't resolve current state of a security group because its `vpc` property is `null`. We need to map `vpc` to a VPC id so we can send that to CloudDriver to fetch the current state. We don't currently apply resolvers to the spec passed to `ResourceHandler.desired` so the property can be `null` if the spec relies on the default applied by `NetworkResolver`. This is an interim fix until we figure out the proper approach.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
@@ -30,22 +30,27 @@ interface Locations<T : RegionSpec> {
    * If not specified here, this should be derived from the [SubnetAwareLocations.subnet] (if
    * present) or use a default VPC name.
    */
-  val vpc: String?
+  val vpc: String
   val regions: Set<T>
 }
 
 data class SubnetAwareLocations(
   override val account: String,
-  override val vpc: String?,
   /**
    * If not specified here, this should be derived from a default subnet purpose using [vpc].
    */
   val subnet: String?,
+  // TODO: this is not ideal as we'd like this default to be configurable
+  override val vpc: String = defaultVPC(subnet),
   override val regions: Set<SubnetAwareRegionSpec>
 ) : Locations<SubnetAwareRegionSpec>
 
 data class SimpleLocations(
   override val account: String,
-  override val vpc: String?,
+  // TODO: this is not ideal as we'd like this default to be configurable
+  override val vpc: String = "vpc0",
   override val regions: Set<SimpleRegionSpec>
 ) : Locations<SimpleRegionSpec>
+
+fun defaultVPC(subnet: String?) =
+  subnet?.let { Regex("""^.+\((.+)\)$""").find(it)?.groupValues?.get(1) } ?: "vpc0"
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/NetworkResolver.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/NetworkResolver.kt
@@ -3,30 +3,24 @@ package com.netflix.spinnaker.keel.ec2.resolvers
 import com.netflix.spinnaker.keel.api.ApiVersion
 import com.netflix.spinnaker.keel.api.Locatable
 import com.netflix.spinnaker.keel.api.Resource
-import com.netflix.spinnaker.keel.api.SimpleLocations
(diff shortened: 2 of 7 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
```diff
@@ -30,22 +30,27 @@ interface Locations<T : RegionSpec> {
...
 
 data class SubnetAwareLocations(
   override val account: String,
-  override val vpc: String?,
   /**
    * If not specified here, this should be derived from a default subnet purpose using [vpc].
    */
   val subnet: String?,
+  // TODO: this is not ideal as we'd like this default to be configurable
+  override val vpc: String = defaultVPC(subnet),
```
robfletcher: when there is a subnet we can default VPC from the subnet

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
```diff
@@ -30,22 +30,27 @@ interface Locations<T : RegionSpec> {
...
   override val regions: Set<SubnetAwareRegionSpec>
 ) : Locations<SubnetAwareRegionSpec>
 
 data class SimpleLocations(
   override val account: String,
-  override val vpc: String?,
+  // TODO: this is not ideal as we'd like this default to be configurable
+  override val vpc: String = "vpc0",
```
robfletcher: with no subnet we can use a simpler default

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/NetworkResolverTests.kt
```diff
@@ -44,15 +45,16 @@ import strikt.assertions.hasSize
 import strikt.assertions.isEqualTo
 import java.time.Duration
 
-internal abstract class NetworkResolverTests<T : Locatable<*>> : JUnit5Minutests {
+internal abstract class NetworkResolverTests<L : Locations<*>, T : Locatable<L>> : JUnit5Minutests {
```
robfletcher: Re-worked the tests a bit so they handle the built-in default

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/NetworkResolver.kt
```diff
@@ -44,32 +38,12 @@ abstract class NetworkResolver<T : Locatable<*>>(
...
-    resource.run {
-      copy(
-        spec = spec.run {
-          copy(
-            locations = locations.withResolvedNetwork()
-          )
-        }
-      )
-    }
-}
```
robfletcher: Don't need a `NetworkResolver` implementation for `SecurityGroup` as the only thing it did was default the `vpc`.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/NetworkResolverTests.kt
```diff
@@ -44,17 +41,43 @@ import strikt.assertions.hasSize
 import strikt.assertions.isEqualTo
 import java.time.Duration
 
-internal abstract class NetworkResolverTests<T : Locatable<*>> : JUnit5Minutests {
+internal abstract class NetworkResolverTests<T : Locatable<SubnetAwareLocations>> : JUnit5Minutests {
```
robfletcher: These tests can be a bit simpler too as they no longer need to deal with resources that use `SimpleLocations`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
```diff
@@ -30,22 +30,27 @@ interface Locations<T : RegionSpec> {
...
   override val regions: Set<SimpleRegionSpec>
 ) : Locations<SimpleRegionSpec>
+
+fun defaultVPC(subnet: String?) =
+  subnet?.let { Regex("""^.+\((.+)\)$""").find(it)?.groupValues?.get(1) } ?: "vpc0"
```
robfletcher: Get `vpc` from the `subnet` property, e.g. `"external (vpc1)"`. If the regex doesn't match we still fall back to the standard default.

PR #536, opened 2019-10-21
feat(resources/export): export for ec2 load balancers

asher:
- The `type` sent to `ExportController` should be "${loadBalancerType}$type", i.e. "classicLoadBalancers" or "applicationLoadBalancers" -> `/export/aws/{account}/applicationLoadBalancers/{name}?serviceAccount=foo@spinnaker.io`

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
@@ -1,10 +1,15 @@
 package com.netflix.spinnaker.keel.ec2.resource
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.keel.api.Exportable
 import com.netflix.spinnaker.keel.api.Resource
 import com.netflix.spinnaker.keel.api.ResourceId
 import com.netflix.spinnaker.keel.api.ResourceKind
+import com.netflix.spinnaker.keel.api.SubmittedResource
+import com.netflix.spinnaker.keel.api.SubnetAwareLocations
+import com.netflix.spinnaker.keel.api.SubnetAwareRegionSpec
 import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancer
+import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerOverride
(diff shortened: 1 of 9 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClassicLoadBalancerHandler.kt
```diff
@@ -201,6 +276,35 @@ class ClassicLoadBalancerHandler(
...
+    regionalClbs: Map<String, ClassicLoadBalancer>
+  ) =
+    regionalClbs.forEach { (region, clb) ->
+      val dependenciesDiff = ResourceDiff(clb.dependencies, dependencies).hasChanges()
+      val listenersDiff = ResourceDiff(clb.listeners, listeners).hasChanges()
+      val healthCheckDiff = ResourceDiff(clb.healthCheck, healthCheck).hasChanges()
+
+      if (dependenciesDiff || listenersDiff || healthCheckDiff) {
+        (overrides as MutableMap)[region] = ClassicLoadBalancerOverride(
+          dependencies = if (dependenciesDiff) {
```
emjburns: can you use `dependenciesDiff ?: null` here to shorten?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
```diff
@@ -112,27 +118,99 @@ class ApplicationLoadBalancerHandler(
...
+      moniker = base.moniker,
+      locations = SubnetAwareLocations(
+        account = exportable.account,
+        vpc = base.location.vpc,
+        subnet = base.location.subnet,
+        regions = albs.map { (region, alb) ->
+          SubnetAwareRegionSpec(
+            name = region,
+            availabilityZones =
+            if (alb.location.availabilityZones.containsAll(
```
emjburns: Do you think it's worth pulling out this calculation to before we're creating the `spec`? I think I like resolving most values first (especially if there's more complicated logic than like `substring()`.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClassicLoadBalancerHandler.kt
asher: `dependenciesDiff` != `clb.dependencies`

Issue #527, opened 2019-10-18
kind name shouldn't have to be unique across all api versions


emjburns:
This function: https://github.com/spinnaker/keel/blob/c09ce2cf6ee0d6f3d42137308b65ca0b4fa250c6/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/Main.kt#L86-L94
makes it so that we're registering types by kind -> spec class. We should make this something like apiversion/kind -> spec class so that we can have a "cluster" for each cloud provider.

You shouldn't have to study every other object type in order to pick a unique name.

robfletcher:
Yeah, sorry. I was lazy.

PR #537, opened 2019-10-21
Swagger API documentation

robfletcher:
- [ ] add `"io.swagger:swagger-annotations:1.5.20"` to Kork
- [ ] add `"org.reflections:reflections:+"` to Kork

Addresses #230

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/config/SwaggerConfiguration.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/config/SwaggerConfiguration.kt
@@ -2,9 +2,11 @@ package com.netflix.spinnaker.config
 
 import com.fasterxml.classmate.ResolvedType
 import com.fasterxml.classmate.TypeResolver
-import com.netflix.spinnaker.keel.api.ApiVersion
+import com.fasterxml.jackson.databind.annotation.JsonSerialize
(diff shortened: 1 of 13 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/config/SwaggerConfiguration.kt
```diff
@@ -0,0 +1,62 @@
...
+
+@Configuration
+@EnableSwagger2
+class SwaggerConfiguration {
+  @Bean
+  fun api(resolver: TypeResolver, handlers: List<ResourceHandler<*, *>>): Docket =
+    Docket(SWAGGER_2)
+      .apply {
+        handlers
+          .map { it.supportedKind.second }
```
emjburns: I think we should map by api version + supportedKind (so we don't end up with this issue again: https://github.com/spinnaker/keel/issues/527)

keel-api/src/main/resources/swagger-api.yml
```diff
@@ -0,0 +1,33 @@
+swagger.api:
```
luispollo: So we couldn't use Swagger code annotations to specify the info that's in this file?

robfletcher: I'd rather have it in this file.

robfletcher:
Superseded by #811

PR #540, opened 2019-10-22
Titus actuation support

emjburns:
Changes in titus spec to flatten structure. Actuation support needed for managing titus resources. Tests!

Still todo:
- notification support for environments
- an "image provider" instead of specifying a digest
- have kind be `cluster` instead of `titus-cluster`


robfletcher:
I like the flatter structure

PR #543, opened 2019-10-23
fix(bake): rebake if region missing

emjburns:
If we have an image diff where the image exists in one region but not any of the others a simple "bake" doesn't fix it. Instead, we bake every minute for the rest of time.

This change adds a force rebake to try and remedy this situation (context parameter checked in orca [here](https://github.com/spinnaker/orca/blob/master/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy#L140)).


emjburns:
Opened an issue to track a re-evaluation of the image name logic: https://github.com/spinnaker/keel/issues/544.

Discussed offline, Going to proceed with this interim "enhancement" in the mean time.

PR #550, opened 2019-10-24
Kind not required to be globally unique

### LINKED ISSUE #527: kind name shouldn't have to be unique across all api versions

robfletcher:
The size of this PR is not reflective of how much of a pain in the arse it was to implement.

Fixes #527

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/Main.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/Main.kt
@@ -86,11 +86,11 @@ class KeelApplication {
   fun registerResourceSpecSubtypes() {
     plugins
       .filterIsInstance<ResourceHandler<*, *>>()
-      .map { it.supportedKind }
-      .forEach { (kind, type) ->
+      .map { Triple(it.apiVersion, it.supportedKind.first.singular, it.supportedKind.second) }
+      .forEach { (apiVersion, kind, type) ->
         objectMappers.forEach { objectMapper ->
-          log.info("Registering ResourceSpec sub-type {}: {}", kind.singular, type.simpleName)
-          objectMapper.registerSubtypes(NamedType(type, kind.singular))
+          log.info("Registering ResourceSpec sub-type {}/{}: {}", apiVersion, kind, type.simpleName)
+          objectMapper.registerSubtypes(NamedType(type, "$apiVersion/$kind"))
         }
       }
   }
--- a/keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/ConvertExampleFilesTest.kt
+++ b/keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/ConvertExampleFilesTest.kt
@@ -30,6 +30,7 @@ import com.netflix.spinnaker.keel.serialization.configuredYamlMapper
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
 import strikt.api.expectCatching
+import strikt.assertions.isA
 import strikt.assertions.succeeded
 
 class ConvertExampleFilesTest : JUnit5Minutests {
@@ -43,8 +44,11 @@ class ConvertExampleFilesTest : JUnit5Minutests {
 
       test("yaml can be parsed") {
         expectCatching {
-          mapper.readValue<SubmittedResource<ClusterSpec>>(file)
-        }.succeeded()
+          mapper.readValue<SubmittedResource<*>>(file)
+        }
+          .succeeded()
+          .get { spec }
+          .isA<ClusterSpec>()
       }
     }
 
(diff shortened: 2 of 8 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/Main.kt
```diff
@@ -86,11 +86,11 @@ class KeelApplication {
...
       .filterIsInstance<ResourceHandler<*, *>>()
-      .map { it.supportedKind }
-      .forEach { (kind, type) ->
+      .map { Triple(it.apiVersion, it.supportedKind.first.singular, it.supportedKind.second) }
+      .forEach { (apiVersion, kind, type) ->
         objectMappers.forEach { objectMapper ->
-          log.info("Registering ResourceSpec sub-type {}: {}", kind.singular, type.simpleName)
-          objectMapper.registerSubtypes(NamedType(type, kind.singular))
+          log.info("Registering ResourceSpec sub-type {}/{}: {}", apiVersion, kind, type.simpleName)
+          objectMapper.registerSubtypes(NamedType(type, "$apiVersion/$kind"))
```
robfletcher: The type id for each resource type is now compound. It has to be a string though.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/ConvertExampleFilesTest.kt
```diff
@@ -38,62 +42,77 @@ class ConvertExampleFilesTest : JUnit5Minutests {
...
 
       test("yaml can be parsed") {
         expectCatching {
-          mapper.readValue<SubmittedResource<ClusterSpec>>(file)
-        }.succeeded()
+          mapper.readValue<SubmittedResource<*>>(file)
+        }
+          .succeeded()
+          .get { spec }
+          .isA<ClusterSpec>()
```
robfletcher: I made this change to all the tests in this class so that it also tests we can identify the type correctly. If we give Jackson the generic type in the `readValue` call it doesn't exercise our type id logic.

keel-api/src/test/resources/examples/titus-cluster-example.yml
```diff
@@ -1,6 +1,6 @@
 ---
 apiVersion: titus.spinnaker.netflix.com/v1
-kind: titus-cluster
+kind: cluster
```
robfletcher: 🎉

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
```diff
@@ -63,15 +62,11 @@ data class Resource<T : ResourceSpec>(
...
 data class SubmittedResource<T : ResourceSpec>(
   val metadata: Map<String, Any?>,
   val apiVersion: ApiVersion,
   val kind: String,
-  @JsonTypeInfo(
-    use = Id.NAME,
-    include = As.EXTERNAL_PROPERTY,
-    property = "kind"
-  )
   val spec: T
```
robfletcher: Now use a custom deserializer instead of trying to use `@JsonTypeInfo`. There's no way to get that to use a compound properly without it also appearing as a single value in the JSON.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/SubmittedResourceDeserializer.kt
```diff
@@ -0,0 +1,54 @@
...
+      val kind = root.get("kind").textValue()
+      val metadata = mapper.convertValue<Map<String, Any?>>(root.get("metadata"))
+      val specType = resolveResourceSpecType(apiVersion, kind)
+      val spec = mapper.convertValue(root.get("spec"), specType) as ResourceSpec
+
+      try {
+        return SubmittedResource(metadata, apiVersion, kind, spec)
+      } catch (e: Exception) {
+        throw instantiationException(SubmittedResource::class.java, e)
+      }
```
robfletcher: This ensures we get standard Jackson errors if initialization goes wrong. That means we get the correct HTTP response codes if there's invalid content (e.g. `serviceAccount` is missing from metadata).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/SubmittedResourceDeserializer.kt
```diff
@@ -0,0 +1,61 @@
...
+      val kind = root.get("kind").textValue()
+      val metadata = mapper.convertValue<Map<String, Any?>>(root.get("metadata"))
+      val specType = resolveResourceSpecType(apiVersion, kind)
+      val spec = mapper.convertValue(root.get("spec"), specType) as ResourceSpec
+
+      try {
+        return SubmittedResource(metadata, apiVersion, kind, spec)
+      } catch (e: Exception) {
+        throw instantiationException<SubmittedResource<*>>(e)
+      }
```
robfletcher: This ensures we get standard Jackson errors if initialization goes wrong. That means we get the correct HTTP response codes if there's invalid content (e.g. `serviceAccount` is missing from metadata).

emjburns:
This is awesome, thank you!

PR #556, opened 2019-10-25
feat(docker): react to and store docker images

emjburns:
Add a new artifact type (`DOCKER_IMAGE`), and save those too.

Also, adding the `/artifact/register` endpoint back for easy testing.

### DIFF
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
@@ -3,6 +3,8 @@ package com.netflix.spinnaker.keel.artifact
 import com.netflix.spinnaker.igor.ArtifactService
 import com.netflix.spinnaker.keel.api.ArtifactStatus
 import com.netflix.spinnaker.keel.api.ArtifactType
+import com.netflix.spinnaker.keel.api.ArtifactType.DEB
+import com.netflix.spinnaker.keel.api.ArtifactType.DOCKER_IMAGE
 import com.netflix.spinnaker.keel.api.DeliveryArtifact
 import com.netflix.spinnaker.keel.events.ArtifactEvent
 import com.netflix.spinnaker.keel.events.ArtifactRegisteredEvent
@@ -30,8 +32,19 @@ class ArtifactListener(
       .forEach { korkArtifact ->
         val artifact = korkArtifact.toDeliveryArtifact()
         if (artifactRepository.isRegistered(artifact.name, artifact.type)) {
-          val version = "${korkArtifact.name}-${korkArtifact.version}"
-          val status = artifactStatus(korkArtifact)
+          val version: String
+          val status: ArtifactStatus
+          when {
+            korkArtifact.isDeb() -> {
+              version = "${korkArtifact.name}-${korkArtifact.version}"
+              status = debStatus(korkArtifact)
+            }
+            korkArtifact.isDockerImage() -> {
+              version = korkArtifact.name + ":" + korkArtifact.version
+              status = ArtifactStatus.FINAL // todo eb: should we default? should we re-think status? should status be null?
+            }
+            else -> throw UnsupportedArtifactTypeException(korkArtifact.type)
+          }
           log.info("Registering version {} ({}) of {} {}", version, status, artifact.name, artifact.type)
           artifactRepository.store(artifact, version, status)
             .also { wasAdded ->
@@ -53,22 +66,22 @@ class ArtifactListener(
       artifactRepository.register(artifact)
     }
 
-    if (artifactRepository.versions(artifact).isEmpty()) {
-      storeLatestVersion(artifact, event.statuses)
+    if (artifact.type == DEB && artifactRepository.versions(artifact).isEmpty()) {
+      storeLatestDebVersion(artifact, event.statuses)
     }
   }
 
   /**
    * Grab the latest version which matches the statuses we care about, so the artifact is relevant.
    */
-  protected fun storeLatestVersion(artifact: DeliveryArtifact, statuses: List<ArtifactStatus>) =
+  protected fun storeLatestDebVersion(artifact: DeliveryArtifact, statuses: List<ArtifactStatus>) =
     runBlocking {
       artifactService
         .getVersions(artifact.name, statuses.map { it.toString() })
         .firstOrNull()
         ?.let { firstVersion ->
           val version = "${artifact.name}-$firstVersion"
-          val status = artifactStatus(artifactService.getArtifact(artifact.name, firstVersion))
+          val status = debStatus(artifactService.getArtifact(artifact.name, firstVersion))
           log.debug("Storing latest version {} ({}) for registered artifact {}", version, status, artifact)
           artifactRepository.store(artifact, version, status)
         }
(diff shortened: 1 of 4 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
```diff
@@ -6,7 +6,7 @@ data class DeliveryArtifact(
 )
 
 enum class ArtifactType {
-  DEB
+  DEB, DOCKER_IMAGE
```
erikmunson: could i convince you to make this `DOCKER` for YAML definition purposes? seems to fit more nicely with the `DEB` naming. Either way, I'll continue on living somehow 😄

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -30,8 +32,19 @@ class ArtifactListener(
...
-          val status = artifactStatus(korkArtifact)
+          val version: String
+          val status: ArtifactStatus
+          when {
+            korkArtifact.isDeb() -> {
+              version = "${korkArtifact.name}-${korkArtifact.version}"
+              status = debStatus(korkArtifact)
+            }
+            korkArtifact.isDockerImage() -> {
+              version = korkArtifact.name + ":" + korkArtifact.version
```
luispollo: nit: Any reason not to use the string template like on line 39?

erikmunson:
THIS IS EXCITING

exciting

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -53,22 +66,22 @@ class ArtifactListener(
       artifactRepository.register(artifact)
     }
 
-    if (artifactRepository.versions(artifact).isEmpty()) {
-      storeLatestVersion(artifact, event.statuses)
+    if (artifact.type == DEB && artifactRepository.versions(artifact).isEmpty()) {
```
luispollo: nit: Any reason to use one check for type (`korkArtifact.isDeb()`, etc.) above and a different check here? I prefer the check style here. We could replace the checks above with this style as well for consistency.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/UnsupportedArtifactTypeException.kt
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.keel.artifact
+
+import com.netflix.spinnaker.keel.api.ArtifactStatus
+
+class UnsupportedArtifactTypeException(
+  val type: String
+) : RuntimeException("Artifact type $type not in supported types: ${ArtifactStatus.values()}")
```
luispollo: This should be `ArtifactType.values()`, no?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -78,14 +91,20 @@ class ArtifactListener(
...
     return ArtifactStatus.valueOf(status)
   }
 
   private val artifactTypeNames by lazy { ArtifactType.values().map(ArtifactType::name) }
 
+  private fun Artifact.isDeb(): Boolean =
+    type.equals(DEB.toString(), true)
+
+  private fun Artifact.isDockerImage(): Boolean =
+    type.equals(DOCKER_IMAGE.toString(), true)
```
robfletcher: I don’t understand this syntax. Isn’t it just `type == DOCKER_IMAGE`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
robfletcher: Strong agree

emjburns: Sure :D

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: Updating, good catch.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/UnsupportedArtifactTypeException.kt
emjburns: Yes! Good catch.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: They're on two different objects - a `[korkArtifact](https://github.com/spinnaker/kork/blob/master/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java)` is the generic artifact payload, where everything important is a string. This artifact is a `DeliveryArtifact` (or something like that) where the type field is an enum. So, we need two different checks unfortunately.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: This is on a `[korkArtifact](https://github.com/spinnaker/kork/blob/master/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java)`, not our artifact. So the type field is a string.

robfletcher: Is that trailing `true` a default or something?

emjburns: It's ignore case = true

robfletcher: ah

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
luispollo: I know, but on line 33 you already have the `DeliveryArtifact` object and so all the checks inside the `when` block could be using it, which would make for the same type of check as you have here...

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -53,22 +66,22 @@ class ArtifactListener(
...
+    if (artifact.type == DEB && artifactRepository.versions(artifact).isEmpty()) {
+      storeLatestDebVersion(artifact, event.statuses)
     }
   }
 
   /**
    * Grab the latest version which matches the statuses we care about, so the artifact is relevant.
    */
-  protected fun storeLatestVersion(artifact: DeliveryArtifact, statuses: List<ArtifactStatus>) =
+  protected fun storeLatestDebVersion(artifact: DeliveryArtifact, statuses: List<ArtifactStatus>) =
```
asher: Doesn't need to change in this PR, but instead of deb vs. docker, we should think about artifact formats in terms of those that need to be baked, vs. those that are directly deployable. We'll probably need to onboard RPM based use-cases in the near future, and possibly Chocolatey too. It's also interesting to consider rosco's support for baking various types of k8s manifests. In the near term, this may just mean thinking about things in terms of `Artifact.needsBake()` instead of `Artifact.isDeb()`.

luispollo: @asher Side-question: for the K8S manifests support in rosco, what's the output of "baking" them? Just curious.

emjburns: Oh that's a great point. I'll think about that going forward.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: Ah good call! I'll just use that 🤦‍♂

PR #559, opened 2019-10-29
fix(export): better mapping between clouddriver types and keel kinds

asher:
- also return a 404 instead of 500 when attempting to export resources
that cannot be found.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExportController.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExportController.kt
@@ -1,20 +1,24 @@
 package com.netflix.spinnaker.keel.rest
 
-import com.google.common.base.CaseFormat
 import com.netflix.spinnaker.keel.api.Exportable
 import com.netflix.spinnaker.keel.api.SPINNAKER_API_V1
 import com.netflix.spinnaker.keel.api.SubmittedResource
 import com.netflix.spinnaker.keel.clouddriver.CloudDriverCache
+import com.netflix.spinnaker.keel.clouddriver.ResourceNotFound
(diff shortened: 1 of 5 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExportController.kt
```diff
@@ -55,7 +67,12 @@ class ExportController(
     @RequestParam("serviceAccount") serviceAccount: String
   ): SubmittedResource<*> {
     val apiVersion = SPINNAKER_API_V1.subApi(cloudProviderOverrides[cloudProvider] ?: cloudProvider)
-    val kind = CaseFormat.LOWER_CAMEL.to(CaseFormat.LOWER_HYPHEN, type)
+    val singularType = if (type.endsWith(char = 's', ignoreCase = true)) {
```
robfletcher: Would it be less sketchy to just include plurals in the `typeToKind`?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExportController.kt
```diff
@@ -75,4 +92,10 @@ class ExportController(
       handler.export(exportable)
     }
   }
+
+  @ExceptionHandler(ResourceNotFound::class)
+  @ResponseStatus(HttpStatus.NOT_FOUND)
```
robfletcher: ❤️

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExportController.kt
asher: Aye, that def looks less sketch

Issue #557, opened 2019-10-25
Non-screaming names for artifact types in YAML definitions


erikmunson:
When defining artifacts on resources, we use the values that keel defines on its own enum of artifact types which are in `SCREAMING_CAMEL_CASE`, e.g.

```yaml
imageProvider:
    deliveryArtifact:
      name: "grpc-lab-emunson-server"
      type: "DEB" or "DOCKER_IMAGE"
```

This looks a bit out of place in the YAML definitions themselves, and ideally I'd love to see us move to something less screamy, e.g.

```yaml
imageProvider:
    deliveryArtifact:
      name: "grpc-lab-emunson-server"
      type: "deb" or "docker"
```

This should probably be done as a backwards-compatible change where we just coerce differently-cased values to the enum value, so we can migrate existing definitions without hassle.

luispollo:
@erikmunson I'll take a look at this.

luispollo:
@erikmunson I'm hoping this can be closed now that #560 has been merged, but I can't make changes to issues yet, so pinging you to close if you agree.

erikmunson:
As long as you've tested out the YAML and it works as desired, I'm happy saying it's fixed!

luispollo:
Yes, as per the example here: https://github.com/spinnaker/keel/pull/560#issuecomment-548153240

Thanks for closing it!

PR #560, opened 2019-10-29
feat(serialization): Use friendly names for artifact types

luispollo:
Addresses #557 by introducing "friendly names" for artifact types (i.e. lower-case names as opposed to the default enum value names). Given that all current friendly names match the upper-case enum names, this could probably be even simpler, but when I started we had a conversion between `"docker"` and `DOCKER_TYPE` and so I figured it's OK to have a place to add that kind of logic if we ever have other types that don't match the friendly name. The `@JsonCreator` factory method also allows for backwards-compatibility with existing specs still using the upper-as names.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
@@ -1,12 +1,23 @@
 package com.netflix.spinnaker.keel.api
 
+import com.fasterxml.jackson.annotation.JsonCreator
+import com.fasterxml.jackson.annotation.JsonValue
+
 data class DeliveryArtifact(
   val name: String,
   val type: ArtifactType = ArtifactType.DEB
 )
 
-enum class ArtifactType {
-  DEB, DOCKER
+enum class ArtifactType(@JsonValue val friendlyName: String) {
+  DEB("deb"),
+  DOCKER("docker");
+
+  companion object {
+    @JsonCreator @JvmStatic
+    fun fromFriendlyName(friendlyName: String): ArtifactType? {
+      return valueOf(friendlyName.toUpperCase())
+    }
+  }
 }
 
 // todo eb: add RELEASE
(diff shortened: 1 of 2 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
```diff
@@ -1,12 +1,26 @@
...
 data class DeliveryArtifact(
   val name: String,
   val type: ArtifactType = ArtifactType.DEB
 )
 
-enum class ArtifactType {
-  DEB
+enum class ArtifactType(@JsonValue val friendlyName: String) {
+  DEB("deb"),
+  DOCKER_IMAGE("docker");
```
luispollo: This will conflict with @emjburns' change in #556. I'll wait for her to merge and then resolve the conflict.

luispollo:
@emjburns @erikmunson Please review.

erikmunson:
this is great! thanks luis. FWIW, it appears Emily switched the enum value to `DOCKER` after Rob and I bugged her about it, so you may be able to omit some of the mapping stuff here (maybe? I don't know these things): https://github.com/spinnaker/keel/pull/556/files#diff-2cd4b503a4f236b6c490d3fcccb3481dR9

luispollo:
Hey @erikmunsonm, yes, I noticed that after I made my change. I'll remove the line to convert from `"docker"` to `DOCKER_IMAGE` as soon as @emjburns merges her PR.

luispollo:
@emjburns @erikmunson This is now rebased after Emily's PR was merged. @robfletcher If you want to take a look as well, please do!

luispollo:
@emjburns Right, it will work with both formats. I added a test for that as well.
@erikmunson Yes, JSON output will be using the friendly format. For example:
```yaml
---
metadata:
  serviceAccount: "keel@spinnaker.io"
kind: "cluster"
spec:
  moniker:
    app: "keeldemo"
    stack: "main"
  imageProvider:
    deliveryArtifact:
      name: "keeldemo"
      type: "deb"
  locations:
    account: "test"
    subnet: "internal (vpc0)"
    vpc: "vpc0"
    regions:
    - name: "us-west-2"
  launchConfiguration:
    instanceType: "m5.large"
    ebsOptimized: true
    iamRole: "keeldemoInstanceProfile"
    keyPair: "nf-test-keypair-a"
    instanceMonitoring: false
  capacity:
    min: 1
    max: 1
    desired: 1
  dependencies:
    loadBalancerNames:
    - "keeldemo-main-clb"
    securityGroupNames:
    - "keeldemo"
    - "nf-datacenter"
    - "nf-infrastructure"
```

luispollo:
@emjburns If you're OK with it, could you please add the `ready to merge` label?

PR #564, opened 2019-10-30
fix(bake): ignore image where region or tags are missing

emjburns:
When we search for an image we may get a list back where some images don't have a region, and some have a region but don't have tags in that region.

This is a problem because we assume an image has the same name in each region we're deploying to and we use the tags to figure out what package version the image is running.

This PR allows us to find the latest image which has all the regions we care about and also has tags for each region.

Thoughts on whether this will help with the amount of re-baking we're doing?


robfletcher:
This still seems harder than just not requiring the same AMI name in all regions

emjburns:
Agreed, but that will be explored in a different change set

PR #567, opened 2019-10-30
fix(bake): include account in find packages

emjburns:
Accidentally didn't use account in the query to find an image by package name. That didn't work. Trying again with the right api calls here.

Also removing `titus-cluster` from tag-able things because I forgot to earlier

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
@@ -98,10 +98,10 @@ class ImageService(
    * Each ami must have tags.
    */
   suspend fun getLatestNamedImageWithAllRegionsForAppVersion(appVersion: AppVersion, account: String, regions: List<String>): NamedImage? =
-    cloudDriverService.images(
+    cloudDriverService.namedImages(
       serviceAccount = DEFAULT_SERVICE_ACCOUNT,
-      provider = "aws",
-      name = appVersion.toImageName().replace("~", "_")
+      imageName = appVersion.toImageName().replace("~", "_"),
+      account = account
     )
       .sortedWith(NamedImageComparator)
       .findLast {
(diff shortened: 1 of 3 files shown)


keel-tagging-plugin/src/main/kotlin/com/netflix/spinnaker/keel/tagging/ResourceTagger.kt
```diff
@@ -77,7 +77,6 @@ class ResourceTagger(
 
   private val taggableResources = listOf(
     "cluster",
-    "titus-cluster", // todo eb: fix to just be cluster
```
luispollo: Unrelated change?

emjburns: Yes, but I snuck it in. Bad practice.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -98,10 +98,10 @@ class ImageService(
    * Each ami must have tags.
    */
   suspend fun getLatestNamedImageWithAllRegionsForAppVersion(appVersion: AppVersion, account: String, regions: List<String>): NamedImage? =
-    cloudDriverService.images(
+    cloudDriverService.namedImages(
       serviceAccount = DEFAULT_SERVICE_ACCOUNT,
-      provider = "aws",
-      name = appVersion.toImageName().replace("~", "_")
+      imageName = appVersion.toImageName().replace("~", "_"),
+      account = account
```
asher: Could this end up being the account we'll want to deploy to instead of the account containing the image?

emjburns: We pass in "test" here (when used in the ImageHandler). Same with the use in the ImageResolver. So I don't think so.

PR #568, opened 2019-10-31
feat(docs): Update README with new intro section,  link to blog post

erikmunson:


README.md
```diff
@@ -1,15 +1,16 @@
 _**IMPORTANT:** This service is currently under development and is not ready for production._
 
 ---
+# High Level Project Overview
+**Keel is the service where much of our work on the new Managed Delivery iniative is happening. For a high level overview of the Managed Delivery project and its goals, [check out our introductory post on the Spinnaker blog](https://blog.spinnaker.io/managed-delivery-evolving-continuous-delivery-at-netflix-eb74877fb33c).**
```
luispollo: I love this, but maybe don't make **the entire text bold**? 😃 
I think the links call enough attention already, and making everything bold takes your attention away from the core of the page which is the architecture stuff just below. Just a thought.

erikmunson: Not sure I agree with the sentiment (people are _far_ worse at deciding what to focus on than they think they are), but the feedback is heard and I just committed an un-bolding.

Issue #561, opened 2019-10-29
Apply PAUSED status to resources which a veto currently applies to


erikmunson:
We have a list of lovely resource statuses, which we're about to start surfacing in the UI. Another status we should start calculating is PAUSED — which should be chosen for a resource whenever it falls under the scope of an active veto plugin. Right now that means `ApplicationVeto`, but ideally we'd just check whether _any_ veto applies and return PAUSED if so.

erikmunson:
For whoever implements this, here's how we currently check for whether a veto applies to a given resource as a way to get started: https://github.com/spinnaker/keel/pull/348/files#diff-909791cad394c56c8595f2e446444cf1R22-R43

tl;dr give VetoEnforcer a resource name and check `allowed`.

luispollo:
@erikmunson I'll look at this.

erikmunson:
Closed via #572

PR #572, opened 2019-11-01
feat(resource-status): Return PAUSED status for vetoed resources

luispollo:
Addresses #561 

~After discussing with @emjburns, I implemented an override of the resource status in `ApplicationController`, which is what the UI calls to get resource summaries, so that it was possible to use the auto-wired `VetoEnforcer` bean without introducing new cross-module dependencies (the `ResourceRepository` interface, which is where status is currently calculated for resources, is in `keel-core`, and `VetoEnforcer` is in `keel-veto`). This works, but has the undesired side-effect that the `PAUSED` status as defined by a presence of any veto on a resource is only visible from the `/application` API. For example, if you call `/resources` you'll still get whatever the "real" status of the resource is.~

@asher @robfletcher If you have any ideas on how to do this in a different/better way, I'd be happy to change things around.

@erikmunson FYI.

Update: based on feedback below, I revisited the implementation to follow the existing pattern of emitting `ResourceEvent`s from `ResourceActuator`, and introduced new events for when actuation is paused/resumed. I kept the tests for `ApplicationController` (which are now unrelated) because we didn't have any before. I extracted the logic to parse the app name from a `ResourceId` into a getter so it can be used consistently. Hope this will make more sense now.

~P.S. I think I may have uncovered an issue with different assumptions we make about the `ResourceId` format in different places (see `FIXME` comment in `ApplicationVeto`).~

### DIFF
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -45,8 +45,7 @@ class ResourceActuator(
       log.debug("Skipping actuation for resource {} because it was vetoed: {}", id, response.message)
       publisher.publishEvent(ResourceCheckSkipped(apiVersion, kind, id))
 
-      val lastEvent = resourceRepository.eventHistory(id, limit = 1).first()
-      if (lastEvent !is ResourceActuationPaused) {
+      if (resourceRepository.lastEvent(id) !is ResourceActuationPaused) {
         log.info("Actuation for resource {} paused due to veto: {}", id, response.message)
         publisher.publishEvent(ResourceActuationPaused(apiVersion, kind, id.value, id.applicationName, response.message,
           clock.instant()))
@@ -64,8 +63,7 @@ class ResourceActuator(
 
     log.debug("Checking resource {}", id)
 
-    val lastEvent = resourceRepository.eventHistory(id, limit = 1).first()
-    if (lastEvent is ResourceActuationPaused) {
+    if (resourceRepository.lastEvent(id) is ResourceActuationPaused) {
       log.info("Actuation for resource {} resuming", id)
       publisher.publishEvent(ResourceActuationResumed(apiVersion, kind, id.value, id.applicationName, clock.instant()))
     }
@@ -99,7 +97,7 @@ class ResourceActuator(
         else -> {
           log.info("Resource {} is valid", id)
           // TODO: not sure this logic belongs here
-          val lastEvent = resourceRepository.eventHistory(resource.id, limit = 1).first()
+          val lastEvent = resourceRepository.lastEvent(id)
           if (lastEvent is ResourceDeltaDetected || lastEvent is ResourceActuationLaunched) {
             publisher.publishEvent(ResourceDeltaResolved(resource, clock))
           } else {
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -7,6 +7,8 @@ import com.netflix.spinnaker.keel.api.ResourceSpec
 import com.netflix.spinnaker.keel.api.id
 import com.netflix.spinnaker.keel.diff.ResourceDiff
 import com.netflix.spinnaker.keel.events.ResourceActuationLaunched
+import com.netflix.spinnaker.keel.events.ResourceActuationPaused
+import com.netflix.spinnaker.keel.events.ResourceActuationResumed
 import com.netflix.spinnaker.keel.events.ResourceCheckError
 import com.netflix.spinnaker.keel.events.ResourceDeltaDetected
 import com.netflix.spinnaker.keel.events.ResourceDeltaResolved
@@ -42,6 +44,13 @@ class ResourceActuator(
     if (!response.allowed) {
       log.debug("Skipping actuation for resource {} because it was vetoed: {}", id, response.message)
       publisher.publishEvent(ResourceCheckSkipped(apiVersion, kind, id))
+
+      val lastEvent = resourceRepository.eventHistory(id, limit = 1).first()
+      if (lastEvent !is ResourceActuationPaused) {
+        log.info("Actuation for resource {} paused due to veto: {}", id, response.message)
+        publisher.publishEvent(ResourceActuationPaused(apiVersion, kind, id.value, id.applicationName, response.message,
+          clock.instant()))
+      }
       return
     }
 
@@ -54,6 +63,12 @@ class ResourceActuator(
     }
 
     log.debug("Checking resource {}", id)
+
+    val lastEvent = resourceRepository.eventHistory(id, limit = 1).first()
+    if (lastEvent is ResourceActuationPaused) {
+      log.info("Actuation for resource {} resuming", id)
+      publisher.publishEvent(ResourceActuationResumed(apiVersion, kind, id.value, id.applicationName, clock.instant()))
+    }
 
     val resource = resourceRepository.get(id)
 
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -44,12 +44,8 @@ class ResourceActuator(
     if (!response.allowed) {
       log.debug("Skipping actuation for resource {} because it was vetoed: {}", id, response.message)
       publisher.publishEvent(ResourceCheckSkipped(apiVersion, kind, id))
-
-      if (resourceRepository.lastEvent(id) !is ResourceActuationPaused) {
-        log.info("Actuation for resource {} paused due to veto: {}", id, response.message)
-        publisher.publishEvent(ResourceActuationPaused(apiVersion, kind, id.value, id.applicationName, response.message,
-          clock.instant()))
-      }
+      publisher.publishEvent(ResourceActuationPaused(apiVersion, kind, id.value, id.application, response.message,
+        clock.instant()))
       return
     }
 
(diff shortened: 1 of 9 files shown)


emjburns:
@luispollo we could add the same sort of logic into the `ResourceController` to check if a resource has been "paused", and apply the same override. But, I'm not sure if that's the best way.

Re: `ResourceId`. Yes, we don't have a format for it that's defined. Are you seeing any problems currently, or are you just wary that we have to be careful not to create problems for ourselves?

I think we chose to have a semi-parseable string given the vague feeling of "it's easy to look at, serialize, we've done things for redis like this before, and right now there aren't many resource types". I agree that it makes working with the IDs a little more challenging though.

robfletcher:
Why don't we just trigger a new `ResourceEvent` type when we pause management or resume management?

keel-veto/src/main/kotlin/com/netflix/spinnaker/keel/veto/application/ApplicationVeto.kt
```diff
@@ -34,6 +34,8 @@ class ApplicationVeto(
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
   override fun check(id: ResourceId): VetoResponse {
+    // FIXME (lpollo): The following expects a ResourceId to have a specific format, but it's not defined/enforced
+    //  anywhere. We should add a validation and perhaps token extraction methods to ResourceId to be used consistently.
     val appName = id.toString().split(":").last().split("-").first()
```
robfletcher: Any resource should have the application name in its metadata. Seems like it would make sense to get it from there.

luispollo: So would you suggest to change this method to take a `Resource` instead of a `ResourceId`?

emjburns:
@robfletcher that's a great idea! That way, the status of the resource will be reflected easily.

erikmunson:
Oh hey I love that idea — then you could actually follow the chain of events in the history view too. Event names could be something like `ResourceActuationPaused` and `ResourceActuationResumed`?

keel-veto/src/main/kotlin/com/netflix/spinnaker/keel/veto/application/ApplicationVeto.kt
robfletcher: Either that or it will need to load the full resource in here. Not sure which is more appropriate.

luispollo:
@gal-yardeni FYI

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceId.kt
```diff
@@ -24,4 +24,8 @@ import com.netflix.spinnaker.keel.serialization.ResourceIdDeserializer
 @JsonDeserialize(using = ResourceIdDeserializer::class)
 data class ResourceId(val value: String) {
   override fun toString(): String = value
+
+  // Resource names (the last token in a ResourceId) follow the Moniker format (<app>-<stack>-<detail>)
+  val applicationName: String
```
emjburns: you can just call this `application`

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -55,6 +63,11 @@ class ResourceActuator(
 
     log.debug("Checking resource {}", id)
 
+    if (resourceRepository.lastEvent(id) is ResourceActuationPaused) {
```
emjburns: Should we care about `ResourceActuationResumed`? I am leaning no - here's my logic.

We're only updating this status every time we check the resources, so if we emit a `ResourceActuationResumed` event here we will immediately below emit another event with the state of the resource. Because of that, `ResourceActuationResumed` will never be the last status for more than a couple of seconds. Additionally, we don't get any status information from `ResourceActuationResumed` - we don't know if the resources is happy, has a diff, etc. 

Is there any case I'm not thinking of where it's helpful to know that a resource was paused but has just switched?

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -42,6 +44,12 @@ class ResourceActuator(
     if (!response.allowed) {
       log.debug("Skipping actuation for resource {} because it was vetoed: {}", id, response.message)
       publisher.publishEvent(ResourceCheckSkipped(apiVersion, kind, id))
+
+      if (resourceRepository.lastEvent(id) !is ResourceActuationPaused) {
+        log.info("Actuation for resource {} paused due to veto: {}", id, response.message)
```
emjburns: I think that we could just publish this event every time, and then set 
`IgnoreRepeatedHistory` on the event (used [here](https://github.com/spinnaker/keel/blob/master/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt#L149). That way, we don't have to always read the event history in this file. 

Also, the log message here is redundant - it says the same thing as the one on line 45.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -100,6 +103,14 @@ interface ResourceRepository : PeriodicallyCheckedRepository<ResourceHeader> {
    */
   fun eventHistory(id: ResourceId, limit: Int = DEFAULT_MAX_EVENTS): List<ResourceEvent>
 
+  /**
+   * Retrieves the last event from the history of state change events for the resource represented by [id] or null if
+   * none found.
+   *
+   * @param id the resource id.
+   */
+  fun lastEvent(id: ResourceId): ResourceEvent? = eventHistory(id, 1).firstOrNull()
```
emjburns: Super useful 👍

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: My thinking was that it might be useful to surface that event on the UI so you have a complete picture of when actuation was paused and then resumed. You’re right that we emit other events immediately after, but it might not be immediately obvious that they happened as a result on removing the veto/resuming.

Having said that, I’m only becoming familiar with the event history and how it’s used, so if it’s overkill I’d be happy to remove it.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: Oh, that’s useful! I didn’t know about it.

About the log message, it looks redundant, but since the check occurs on a recurring schedule and I’m only emitting the event the first time the veto kicks in after actuation had been in progress, I thought it was relevant to log. Of course, if we change this to always emit the event with the flag then it’s unnecessary. I’ll make that change.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: @erikmunson Do we consume the event history directly from the UI today, or only resource status?

erikmunson: As of this moment we’re only linking out to the history endpoint for a resource as raw JSON, but it is available to people and the next thing I’ll be working on is consuming it for a UI view.

Either way (JSON or UI), 100% agreed that having a resume event in the activity feed itself has value to human readers that are trying to understand the narrative timeline of what happened. I don’t see a reason to think about it when the target audience is just code/machines, but it should be present in the history we expose publicly.

emjburns: Fair points! Let's leave it in then.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: Added `ignoreRepeatedHistory = true` for `ResourceActuationPaused` and `ResourceActuationResumed`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceId.kt
luispollo: Done!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -123,6 +134,7 @@ interface ResourceRepository : PeriodicallyCheckedRepository<ResourceHeader> {
       history.isActuating() -> ACTUATING
       history.isError() -> ERROR
       history.isCreated() -> CREATED
+      history.isPaused() -> PAUSED
```
emjburns: What state do you want the UI to show if the resource was paused, but has just been resumed? Right now the state is `UNKNOWN` for this case. Is that appropriate?

erikmunson: my 2 cents: I'd like to avoid `UNKNOWN` wherever possible. It's scary and people won't like it.

maybe bad idea: what if we had a `RESUMED` status that allowed the UI to show a message like "Management was paused, but you just resumed it. Spinnaker will check the the current status and might take action to resolve any drift from the declarative configuration"?

emjburns: I agree, and I think a `RESUMED` status is fine for now.

luispollo: Hmm, but to your point earlier @emjburns, that status will only last a few seconds since another actuation-related event will be emitted immediately after. This is why I did not add a new status. Am I missing something?

erikmunson: While that's true, I think the idea is philosophically what I mentioned before: `UNKNOWN` is a very scary thing to see, so when we run into a case where it's possible to avoid an `UNKNOWN` status showing up in the first place I'd advocate we avoid the risk of a scary status even if it's a small risk.

Probably not going to cause huge problems in this specific case either way, but I think it's wise to build up the habit/muscle of keeping UNKNOWN states as rare as possible.

luispollo: Fair enough. Let me add that status.

PR #584, opened 2019-11-06
feat(deb): scheduled fetch latest artifact

emjburns:
Scheduled method to fetch the latest artifact version in case we
- we miss events
- can't handle the event we received so we throw it out
- are running locally and want to test out artifact version things

I have some open questions, namely:
- should this be a distributed mechanism so that we're not doing the fetching on each instance every 3 minutes?
- should I launch a bunch of async tasks with coroutines for the fetch for each artifact, instead of going one by one in a loop?

### DIFF
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
@@ -9,12 +9,14 @@ import com.netflix.spinnaker.keel.api.DeliveryArtifact
 import com.netflix.spinnaker.keel.events.ArtifactEvent
 import com.netflix.spinnaker.keel.events.ArtifactRegisteredEvent
 import com.netflix.spinnaker.keel.persistence.ArtifactRepository
+import com.netflix.spinnaker.keel.persistence.VERSION_COMPARATOR
 import com.netflix.spinnaker.keel.telemetry.ArtifactVersionUpdated
 import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import kotlinx.coroutines.runBlocking
 import org.slf4j.LoggerFactory
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.context.event.EventListener
+import org.springframework.scheduling.annotation.Scheduled
 import org.springframework.stereotype.Component
 
 @Component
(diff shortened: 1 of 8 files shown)


keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -71,6 +73,36 @@ class ArtifactListener(
...
+  /**
+   * For each registered debian artifact, get the last version, and persist if it's newer than what we have.
+   */
+  // todo eb: this doesn't need to run every couple of minutes on each instance...
+  // todo eb: should we fetch more than one version?
+  @Scheduled(fixedDelayString = "\${keel.artifact-refresh.frequency:PT3M}")
+  fun syncDebArtifactVersions() =
+    runBlocking {
+      artifactRepository.getAll(DEB).forEach { artifact ->
+        val lastRecordedVersion: String? = artifactRepository.versions(artifact).firstOrNull()
```
robfletcher: Instead of doing the whole thing inside of a `runBlocking` scope you could put `launch` inside the loop which will create a new coroutine for each artifact. We might want to consider backing off or scheduling checks for each artifact individually once we start handling a lot, though.

luispollo:
> * should this be a distributed mechanism so that we're not doing the fetching on each instance every 3 minutes?

That reminded me of a question that occurred to me recently and I forgot to ask: do we have a distributed lock or other coordination mechanism in place today across instances for the actuation loop so that we don't have multiple instances looking at the same resources? Sorry for the tangent.

emjburns:
@luispollo yes we do! the method is [here](https://github.com/spinnaker/keel/blob/master/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt#L128), and the coordination mechanism is with sql/the timestamp/etc.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: Ok. I'll do that, bump the scheduled method way down, and add an admin endpoint to trigger it. That seems reasonable.

luispollo:
@emjburns Could you please elaborate a bit on how this helps with testing artifacts stuff locally? Is it because instead of waiting for events to come in from echo/igor (which you'd never get on your machine unless you're also running them locally) you can trigger retrieval of artifact info with that new `/sync` API?

emjburns:
@luispollo that's correct - because we won't get new events it's pretty impossible to test that flow (a flow which we use quite heavily). When you run locally now I've configured it so that keel will refresh the artifacts every minute. That will allow you to actually test promoting a new artifact!

PR #585, opened 2019-11-06
chore(tags): remove entity tagging

emjburns:
Entity tags were a hacky way of showing that a resource was managed. Now that we have a beautiful non-hacky way (https://github.com/spinnaker/deck/pull/7590) we can remove the tagging nonsense!

🗑 Goodbye tags, you do not spark joy 🌈


erikmunson:
joy

asher:
I think this means we no longer need `MultiRegion.regionalIds` or anything referencing `regionalIds` too, I added that so that entity tags could be scoped to regional ec2 resources maintained via a multi-region spec. They may come in handy later but not sure.

robfletcher:
http://giphygifs.s3.amazonaws.com/media/EReLGTYVBnyXS/giphy.gif

emjburns:
@asher good call! I'll remove that as well.

PR #586, opened 2019-11-06
feat(environments): allowed-times constraint

asher:


keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
```diff
@@ -184,7 +189,18 @@ internal class DeliveryConfigControllerTests : JUnit5Minutests {
...
+        |          "type": "depends-on",
         |          "environment": "test"
+        |        },
+        |        {
+        |          "type": "time-window",
+        |          "windows": [
+        |            {
+        |              "hours": "6-18",
+        |              "days": "Monday-Friday",
+        |              "tz": "America/Los_Angeles"
```
luispollo: Should the YAML payload have the `tz` field as well?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
```diff
@@ -1,14 +1,36 @@
...
   val environment: String
-) : Constraint()
+) : Constraint("depends-on")
+
+/**
+ * A constraint that requires the current time to fall within an allowed window
+ */
+data class TimeWindowConstraint(
+  val windows: List<TimeWindow>,
+  val tz: String?
```
luispollo: If `tz` is optional, what's the default?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
```diff
@@ -1,14 +1,36 @@
...
 
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize
+import com.fasterxml.jackson.annotation.JsonSubTypes
+import com.fasterxml.jackson.annotation.JsonSubTypes.Type
+import com.fasterxml.jackson.annotation.JsonTypeInfo
 
-@JsonDeserialize(`as` = DependsOnConstraint::class)
-sealed class Constraint
+@JsonTypeInfo(
+  include = JsonTypeInfo.As.PROPERTY,
```
robfletcher: This should be `EXISTING_PROPERTY`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
```diff
@@ -1,14 +1,36 @@
...
+  use = JsonTypeInfo.Id.NAME,
+  property = "type")
+@JsonSubTypes(
+  Type(value = DependsOnConstraint::class, name = "depends-on"),
+  Type(value = TimeWindowConstraint::class, name = "time-window")
```
robfletcher: You should be able to omit the `name` since it's a constant in the types

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
```diff
@@ -184,7 +189,18 @@ internal class DeliveryConfigControllerTests : JUnit5Minutests {
...
         |        {
+        |          "type": "depends-on",
         |          "environment": "test"
+        |        },
+        |        {
+        |          "type": "time-window",
+        |          "windows": [
+        |            {
+        |              "hours": "6-18",
+        |              "days": "Monday-Friday",
```
robfletcher: With this format can I specify something like "Monday, Wednesday and Friday"?

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
robfletcher: Yeah, we discussed this previously. It's necessary otherwise the server is going to interpret those hours as UTC

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
```diff
@@ -6,6 +6,24 @@ import com.netflix.spinnaker.keel.api.DeliveryConfig
 
 interface ConstraintEvaluator<T : Constraint> {
 
+  companion object {
+    inline fun <reified T> getConstraintForEnvironment(
```
luispollo: Kotlin newbie question: does this need to be `reified` if you're not checking/using the type of T in the method?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
```diff
@@ -0,0 +1,203 @@
...
+import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
+import org.springframework.stereotype.Component
+import java.text.ParsePosition
+import java.time.Clock
+import java.time.DayOfWeek
+import java.time.format.DateTimeFormatter
+import java.time.format.TextStyle
+import java.util.Locale
+import java.util.TimeZone
+import java.util.TimeZone.getTimeZone
```
robfletcher: Shouldn't we use `java.time.ZoneId` here?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
robfletcher: Correct. Good spot.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
```diff
@@ -0,0 +1,203 @@
...
+
+/**
```
luispollo: Thanks for taking the time to document the class so well. ❤️

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
```diff
@@ -0,0 +1,203 @@
...
+ *
+ * A timezone can be set in the constraint and a system-wide default can by set via
+ * the dynamic property "default.time-zone" which defaults to UTC.
+ *
+ * Example env constraint:
+ *
+ * constraints:
+ *  - type: time-window
+ *    windows:
+ *      - days: Monday,Tuesday-Friday
```
erikmunson: Up in a test file I saw [all lowercase for this value](https://github.com/spinnaker/keel/pull/586/files#diff-cb20a8e71cc1d6a198f43860b58d5696R149) — looks like in this evaluator we're coercing any capitalization into lowercase?

I don't have a problem with that necessarily, but might be good to shoot for a standardized way of doing examples.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
```diff
@@ -0,0 +1,203 @@
...
+ *        hours: 11-16
+ *        tz: America/Los_Angeles
+ *      - days: Wed
+ *        hours: 12-14
```
erikmunson: Any opinions on 24hr time vs `12pm-2pm`? Latter seems a little more readable, but not sure whether it's "better" from an overall perspective.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
```diff
@@ -0,0 +1,203 @@
...
+
+    constraint.windows.forEach {
+      val hours = parseHours(it.hours, deliveryConfig, targetEnvironment)
+      val hoursMatch = hours.isEmpty() || hours.contains(now.hour)
+
+      val today = now.dayOfWeek
+        .toString()
+        .toLowerCase()
+      val todayShort = now.dayOfWeek
+        .getDisplayName(TextStyle.SHORT, Locale.getDefault())
```
luispollo: Shouldn't this be `Locale.US` for consistency with the formatters up above?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
```diff
@@ -0,0 +1,203 @@
...
+
+  private fun String.isDay(): Boolean = daysOfWeek.contains(this)
+
+  private fun String.isDayRange(): Boolean {
+    val days = this.split("-")
+    if (days.size != 2) {
+      return false
+    }
+
+    return daysOfWeek.contains(days[0]) && daysOfWeek.contains(days[1])
```
luispollo: What would happen here if you specify something wacky like `thu-mon`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
```diff
@@ -0,0 +1,203 @@
...
+     */
+    val days = this.split("-").map { it.capitalize() }
+    val day1 = fullDayFormatter.parseUnresolved(days[0], ParsePosition(0))
+      ?: shortDayFormatter.parseUnresolved(days[0], ParsePosition(0))
+      ?: throw InvalidConstraintException("Failed parsing day range $this")
+    val day2 = fullDayFormatter.parseUnresolved(days[1], ParsePosition(0))
+      ?: shortDayFormatter.parseUnresolved(days[1], ParsePosition(0))
+      ?: throw InvalidConstraintException("Failed parsing day range $this")
+
+    val intRange = IntRange(DayOfWeek.from(day1).value, DayOfWeek.from(day2).value)
```
luispollo: What would happen here if you specify something backwards like `thu-mon`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
robfletcher: next you'll want us to use Fahrenheit

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
asher: No, it's an optional field.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
robfletcher: Or is that just 3 separate windows?

asher: "and" isn't valid but you could specify "Monday, Wednesday, Friday"

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
luispollo: Kotlin studies paying off! 😄

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
erikmunson: Kelvin, obviously.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
erikmunson: you can do a list of ranges/individual days: https://github.com/spinnaker/keel/pull/586/files#diff-cb20a8e71cc1d6a198f43860b58d5696R149

erikmunson: @asher jinx.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
asher: I think it's the other way around, they should be Locale.default. Thanks for pointing out the discrepancy.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluatorTests.kt
```diff
@@ -0,0 +1,168 @@
...
+  data class Fixture(
+    val clock: Clock,
+    val constraint: TimeWindowConstraint
+  ) {
+    val artifact = DeliveryArtifact("fnord", ArtifactType.DEB)
+    val constrainedEnvironment = Environment(
+      name = "prod",
+      constraints = setOf(constraint)
+    )
+    private val previousEnvironment = Environment(
```
luispollo: Philosophical question for the group: I thought `time-window` would be a type of constraint/requirement that would gate deployment into an environment regardless of a promotion flow between environments (I remember discussing this with Michael for the MD blog post), but the function in the evaluator interface is called `canPromote`. Can this be used with a single environment?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
asher: Probably not what the user would want.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
asher: Am and pm, how very quaint.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
luispollo: Should we throw an `InvalidConstraintException` in that case, then?

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
asher: Also note the dynamic config to set a default TZ for windows other than UTC. This will be set to "America/Los_Angeles " so our users will only have to specify a TZ if they want something else.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
luispollo: Should we throw an `InvalidConstraintException` in that case, then?

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
```diff
@@ -141,7 +141,12 @@ internal class DeliveryConfigControllerTests : JUnit5Minutests {
         |      application: someapp
         |- name: prod
         |  constraints:
-        |  - environment: test
+        |  - type: depends-on
+        |    environment: test
+        |  - type: time-window
+        |    windows:
+        |      - hours: 6-18
+        |      - days: mon-fri
```
emjburns: Should this yaml be:
```
windows:
- hours: 6-18
  days: mon-fri
```
(no dash by days) to mean that those go together as one window?

erikmunson:
Another thought: the execution window UI for pipelines has three shortcut links that auto-select days for you: `All`, `Weekends`, and `Weekdays`.

Do any of those three seem like they'd be useful additions in the `days` field?

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
robfletcher: Ooh yes, otherwise it will be 2 separate windows

asher: good catch, I'm so bad at yaml!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
emjburns: I'm delighted with 24hour

asher:
@erikmunson I'll add the weekdays/weekends shortcuts for days, good idea. (All is the default when days are omitted)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
asher: Or what if I add an additional window here that uses lower case? I want to show the flexibility in how days can be specified, but not sure how best to do that.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluatorTests.kt
emjburns: I think you should be able to control deployment into any environment with a `time-window` constraint. That applies if you have only 1 environment and you only want new builds deployed there on, let's say, weekdays.

I think that each constraint that is on an environment is evaluated, and they are && together. So if every constraint returns `true` to `canPromote()` then we would promote the artifact.

luispollo: Thanks, Emily. I was just wondering if for the tests for this particular constraint (unlike for `depends-on`, which obviously needs 2 environments), we should then have a single environment in the fixtures.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
erikmunson: Alright alright, if everybody prefers 24h then let’s go with that. I’m coming from the perspective that for really _any_ UI we and other engineers use day to day, times get mapped to 12h because that’s typically easier to visually parse at a glance. I generally see these YAML configs as more similar in their UX needs to UIs vs. actual programming code.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluatorTests.kt
asher: I'll delete "staging" from the fixture if its causing confusion.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
erikmunson: yeah maybe? I dunno. I suspect all these constraints will need really thorough docs to be successful. It's a detail and probably not super important tbh.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/TimeWindowConstraintEvaluator.kt
asher: Updated the behavior so wrap around hours and days work as one would reasonably expect.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
```diff
@@ -1,14 +1,36 @@
...
+ */
+data class TimeWindowConstraint(
+  val windows: List<TimeWindow>,
+  val tz: String? = null
+) : Constraint("allowed-times")
+
+data class TimeWindow(
+  val days: String? = null,
+  val hours: String? = null
+)
```
emjburns: I think defining these here as strings means we get no validation when you submit a delivery-config. Do you think we could do some sort of validation like we do [here](https://github.com/spinnaker/keel/blob/master/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt#L30), where that validation is pulled from the `AllowedTimesConstraintEvaluator`? 

I'd like to know when I submit that I've spelled something wrong vs an error later :)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/InvalidConstraintException.kt
```diff
@@ -0,0 +1,7 @@
+package com.netflix.spinnaker.keel.exceptions
+
+import java.lang.RuntimeException
+
+class InvalidConstraintException(
+  message: String
```
emjburns: Probably should add constraint name here, so we know where the error is coming from?
```
class InvalidConstraintException(
  constraintName: String,
  message: String
) : RuntimeException("$constraintName: $message")
``` or something?

File: keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/AllowedTimesConstraintEvaluatorTests.kt [hunk 18]
```diff
@@ -0,0 +1,236 @@
...
+        )
+      }
+
+      test("InvalidConstraintException exception") {
+        expectCatching {
+          subject.canPromote(artifact, "1.1", manifest, environment.name)
+        }
+          .failed()
+          .isA<InvalidConstraintException>()
+      }
```
emjburns: I'd like a test for when timezone is not set

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/AllowedTimesConstraintEvaluator.kt
```diff
@@ -0,0 +1,228 @@
...
+ *  - type: allowed-times
+ *    windows:
+ *      - days: Monday,Tuesday-Friday
+ *        hours: 11-16
+ *        tz: America/Los_Angeles
+ *      - days: Wed
+ *        hours: 12-14
+ *        tz: America/Los_Angeles
+ */
+@Component
```
emjburns: This class is super clear. Thanks for adding great class docs, useful helper functions, keeping the core logic short and easy to reason about, and pulling the complex parsing out into smaller functions. 

I struggled with this kind of calculation in Swabbie, and it's delightful to see something that reads well and makes sense here!

Issue #455, opened 2019-09-18
Create endpoint to delete everything for an application


robfletcher:
It basically needs to do this:

```sql
delete from resource where application = '<whatever>';
delete from delivery_config where application = '<whatever>';
delete from resource_event where not exists (select uid from resource r where r.uid = uid);
delete from resource_last_checked where not exists (select uid from resource r where r.uid = resource_uid);
delete from environment where not exists (select uid from delivery_config d where d.uid = delivery_config_uid);
delete from environment_artifact_versions where not exists (select uid from environment e where e.uid = environment_uid);
delete from environment_resource where not exists (select uid from environment e where e.uid = environment_uid);
delete from delivery_config_artifact where not exists (select uid from delivery_config d where d.uid = delivery_config_uid);
delete from delivery_config_last_checked where not exists (select uid from delivery_config d where d.uid = delivery_config_uid);```

luispollo:
@robfletcher I think this can be closed now.

PR #587, opened 2019-11-07
feat(deleteApp): Delete all data for an application

gal-yardeni:
Deleting all data for a given application (this endpoint will not be exposed, for admin use only).

Usage:
DELETE /admin/applications/< applicationName >

Issue: https://github.com/spinnaker/keel/issues/455


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
```diff
@@ -0,0 +1,29 @@
...
+@RequestMapping(path = ["/admin"])
+class AdminController(
+  private val deliveryConfigRepository: DeliveryConfigRepository,
+  private val resourceRepository: ResourceRepository
+) {
+  private val log by lazy { getLogger(javaClass) }
+
+  @DeleteMapping(
+    path = ["/{application}"]
+  )
```
emjburns: This should probably have `@ResponseStatus(ACCEPTED)` here, and not return anything.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
```diff
@@ -0,0 +1,29 @@
...
+  fun deleteApplicationData(
+    @PathVariable("application") application: String
+  ) {
+    log.debug("Deleting all data for application: $application")
+    resourceRepository.deleteByApplication(application)
+    return deliveryConfigRepository.deleteByApplication(application)
```
emjburns: No need to return this if your function has no return object

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryDeliveryConfigRepository.kt
```diff
@@ -15,9 +15,23 @@ import java.time.Instant.EPOCH
...
+    configs
+      .values
+      .filter { it.application == application }
+      .map { it.application }
+      .singleOrNull()
+      ?.also {
+        configs.remove(it)
+        lastCheckTimes.remove(it)
+      }
+      ?: throw NoSuchElementException(application)
```
emjburns: Does it matter if the application doesn't exist? You're trying to make it not exist.

I'm always on the fence about whether a 'delete' call should fail if the object doesn't exist. It's informative, but annoying if it fails because it's already in the state you want. So just asking for your thoughts on this

gal-yardeni: Yep, I don't know whether we should return 404 or not in case the application does not exist, any thoughts?

gal-yardeni:
@luispollo tagging you also as I can't assign reviewers :(

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryDeliveryConfigRepository.kt
asher: I think it's appropriate to serve a 404 in this case. Does retrofit already convert this exception type into an actual 404 response?

robfletcher: > I think it's appropriate to serve a 404 in this case. Does retrofit already convert this exception type into an actual 404 response?

Could always write a test to see

gal-yardeni: Thanks for the feedback - will change it to return 404.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -33,6 +35,35 @@ open class SqlResourceRepository(
   private val objectMapper: ObjectMapper
 ) : ResourceRepository {
 
+  override fun deleteByApplication(application: String) {
```
gal-yardeni: @robfletcher Please lmk if there's a better way of writing this - I wanted to first check if there are any applications that exist with `application` as the name, and throw an exception if not prior to actually deleting those records.

gal-yardeni: Nevermind :) just figured it out.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -36,6 +38,52 @@ class SqlDeliveryConfigRepository(
 
   private val mapper = configuredObjectMapper()
 
+  override fun deleteByApplication(application: String) {
+    jooq.deleteFrom(DELIVERY_CONFIG)
```
luispollo: Shouldn’t all these deletes be executed in a transaction?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryDeliveryConfigRepository.kt
luispollo: Should this be `NoSuchApplication`?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -33,6 +35,32 @@ open class SqlResourceRepository(
...
+
+  jooq.deleteFrom(RESOURCE)
```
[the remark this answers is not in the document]
emjburns: spotless should  auto-format indentation for everything - sometimes it picks weird things (that might look weird) but it has rules and always makes everything the same.

luispollo:
P.S. Questions to the group, for my education. My understanding is that all API calls are authenticated with a client certificate, which is crucial for APIs like this. How/where is that configured in each of the services? And what about authorization? I understand we use Fiat, but I haven't yet seen an example of calling it to authorize a call -- is that because we're not, or is there some magic happening under the covers?

emjburns:
@luispollo I'll answer that offline.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
gal-yardeni: I used the same pattern as other the delete methods - I'll check if I've missed something.

asher: This particular query pattern isn't a good fit for a transaction, we don't want to elongate the period of time within which some of these tables will locked to inserts.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -36,6 +38,52 @@ class SqlDeliveryConfigRepository(
...
+      .where(DELIVERY_CONFIG.APPLICATION.eq(application))
+      .execute()
+
+    jooq.deleteFrom(ENVIRONMENT)
+      .where(
+        notExists(jooq.select(DELIVERY_CONFIG.UID)
```
asher: Avoid using a sub-query for deletes, the lock implications aren't ideal. It would be preferable to first `select uid from delivery_config where application..`,  also select relevant environment ids, and delete everything based on those.

gal-yardeni:
Thanks for reviewing! Please let me know if I missed something, or any other ideas for improvement will be great (@asher . @emjburns ).

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -36,6 +37,66 @@ class SqlDeliveryConfigRepository(
...
+        .execute()
+    }
+
+    environmentUIDs.forEach { uid ->
+      jooq.deleteFrom(ENVIRONMENT_ARTIFACT_VERSIONS)
+        .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(uid))
+        .execute()
+    }
+
+    environmentUIDs.forEach { uid ->
```
asher: Minor nit, but I think this section would be a tad more readable if it were grouped into one each of `deliveryConfigUIDs.forEach{ }` and `environmentUIDs.forEach{ }`

gal-yardeni: That was a great suggestion, thanks! and fixed :)

PR #591, opened 2019-11-07
don't treat missing AMI as a fatal resolution error

robfletcher:
This creates a new base exception type we can use for any "transient" failures in resolving desired state. In this case I'm using it to treat the inability to resolve an AMI for the desired app version as a fatal error as the presumption is it's currently baking (or about to bake).

### DIFF
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -2,6 +2,7 @@ package com.netflix.spinnaker.keel.actuation
 
 import com.netflix.spinnaker.keel.api.ApiVersion
 import com.netflix.spinnaker.keel.api.Resource
+import com.netflix.spinnaker.keel.api.ResourceDependencyNotFound
 import com.netflix.spinnaker.keel.api.ResourceId
 import com.netflix.spinnaker.keel.api.ResourceSpec
 import com.netflix.spinnaker.keel.api.id
@@ -9,6 +10,7 @@ import com.netflix.spinnaker.keel.diff.ResourceDiff
 import com.netflix.spinnaker.keel.events.ResourceActuationLaunched
 import com.netflix.spinnaker.keel.events.ResourceActuationPaused
 import com.netflix.spinnaker.keel.events.ResourceActuationResumed
+import com.netflix.spinnaker.keel.events.ResourceCheckDependencyMissing
 import com.netflix.spinnaker.keel.events.ResourceCheckError
 import com.netflix.spinnaker.keel.events.ResourceDeltaDetected
 import com.netflix.spinnaker.keel.events.ResourceDeltaResolved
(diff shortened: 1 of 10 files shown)


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -101,6 +103,9 @@ class ResourceActuator(
           }
         }
       }
+    } catch (e: ResourceDependencyNotFound) {
+      log.warn("Resource check for {} failed (hopefully temporarily) due to {}", id, e.message)
+      publisher.publishEvent(ResourceCheckDependencyMissing(resource, e, clock))
```
luispollo: I read this as "the resource check dependency is missing". Maybe rename to `ResourceCheckMissingDependency`?

robfletcher:
> I like! Should this show up as a status in the ui, or just be invisible?

It will be considered `DIFF` status

robfletcher:
> Minor naming suggestion (but actually I see you named `ResourceCheckDependencyMissing` following a pattern we use in `ResourceMissing`, so take it with a grain of salt). Otherwise LGTM.

I might change it to something broader. Give me a couple minutes to think it over.

robfletcher:
@luispollo I updated class names

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -101,6 +103,9 @@ class ResourceActuator(
           }
         }
       }
+    } catch (e: ResourceCurrentlyUnresolvable) {
+      log.warn("Resource check for {} failed (hopefully temporarily) due to {}", id, e.message)
+      publisher.publishEvent(ResourceCheckPending(resource, e, clock))
```
emjburns: Can you add a friendly message here, so that the stack trace doesn't show up in the history view?

emjburns:
`ResourceCheckPending` is sort of confusing to me. As a user, my first reaction is "what the heck is it waiting on?". If a human readable reason shows up in the history I think that's fine, but a little less clear than the dependency missing language.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
robfletcher: It won't. It's only actually using the message from `e` inside the event constructor

robfletcher:
> `ResourceCheckPending` is sort of confusing to me. As a user, my first reaction is "what the heck is it waiting on?". If a human readable reason shows up in the history I think that's fine, but a little less clear than the dependency missing language.

I wanted it to also make sense if we want to use it for something other than dependencies. I'll give it some more thought.

robfletcher:
@emjburns would `ResourceCheckIncomplete` make more sense? It should have a decent message because it's taking the one from the underlying exception, which in this case comes from `ImageResolver`

emjburns:
ehh, I think `ResourceCheckPending` is okay as long as it has a message. If we find that we only use it for missing dependencies we can update it in the future, but I do like that it's more generic...

luispollo:
FWIW, I thought the name you picked for the base exception (`ResourceCurrentlyUnresolvable`) was perfect. Is there a reason why we couldn't use the same name for the event? It would go along much more nicely with the other event names in `isDiff()` IMHO.

robfletcher:
> FWIW, I thought the name you picked for the base exception (`ResourceCurrentlyUnresolvable`) was perfect. Is there a reason why we couldn't use the same name for the event? It would go along much more nicely with the other event names in `isDiff()` IMHO.

`ResourceCheckUnresolvable`?

PR #592, opened 2019-11-07
fix(notifications): allow inside voice case

emjburns:
Allow notification config in lowercase instead of requiring uppercase.
```json
      "notifications": [
        {
          "type": "slack",
          "address": "#emburns-slack-test",
          "frequency": "verbose"
        }
      ]
```


robfletcher:
Feels like we could solve this by making enums deserialize in a case-insensitive way rather than having to stick lowercase properties on every enum we use.

luispollo:
@robfletcher I thought about that as well, but wanted to make the parsing backwards-compatible with the upper-case versions, which is achieved with that `@JsonCreator` factory method. I think we could achieve the same thing with a custom (de-)serializer, but it seemed to be more code, not 100% sure.

luispollo:
@emjburns FWIW, if we decide to change approaches, this is the tutorial that was most helpful for me when working on this: https://www.baeldung.com/jackson-serialize-enums

emjburns:
Found a better way!

Issue #570, opened 2019-10-31
Export Titus Resource


emjburns:
Implement the "export a current resource" functionality for Titus

luispollo:
@gal-yardeni I have an interest in picking this up as well, as I just spent some time working on improvements for the EC2 cluster export but this would give me an opportunity to implement a small feature end-to-end. Please let me know if you got something else you're working on, and I'd be happy to take it.

gal-yardeni:
@luispollo Honestly I kind of want to dig into it, but let's talk offline!

PR #593, opened 2019-11-08
fix(export): Omit fields with default values in cluster spec

### LINKED ISSUE #581: Use more of the currently implemented defaults on cluster exports
The cluster export endpoint does a lovely job defaulting many of the fields we currently have good values for, but there are some fields that remain in exported specs even though we support defaults and most people use them:

- `ebsOptimized`, `iamRole`, `keyPair`, and `instanceMonitoring` inside `launchConfiguration`
- The `vpc` and `subnet` fields (I believe these are currently not defaulted via export because the spec defaulting mechanism is different than other fields)
- The entire `health` block — e.g. if I export a cluster I always get:
```
    cooldown: "PT10S"
    warmup: "PT10M"
    healthCheckType: "EC2"
    terminationPolicies:
    - "Default"
```

luispollo:
Closes #581 (maybe). See discussion below for implementation details.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
@@ -30,7 +30,7 @@ interface Locations<T : RegionSpec> {
    * If not specified here, this should be derived from the [SubnetAwareLocations.subnet] (if
    * present) or use a default VPC name.
    */
-  val vpc: String
+  val vpc: String?
   val regions: Set<T>
 }
 
@@ -41,16 +41,35 @@ data class SubnetAwareLocations(
    */
   val subnet: String?,
   // TODO: this is not ideal as we'd like this default to be configurable
-  override val vpc: String = defaultVPC(subnet),
+  override val vpc: String? = defaultVPC(subnet),
   override val regions: Set<SubnetAwareRegionSpec>
-) : Locations<SubnetAwareRegionSpec>
+) : Locations<SubnetAwareRegionSpec> {
+  fun withDefaultsOmitted() =
+    copy(
+      vpc = if (vpc == LocationConstants.DEFAULT_VPC_NAME) {
+        null
+      } else {
+        vpc
+      },
+      subnet = if (subnet == LocationConstants.DEFAULT_SUBNET_PURPOSE.format(vpc)) {
+        null
+      } else {
+        subnet
+      }
+    )
+}
 
 data class SimpleLocations(
   override val account: String,
   // TODO: this is not ideal as we'd like this default to be configurable
-  override val vpc: String = "vpc0",
+  override val vpc: String? = LocationConstants.DEFAULT_VPC_NAME,
   override val regions: Set<SimpleRegionSpec>
 ) : Locations<SimpleRegionSpec>
 
 fun defaultVPC(subnet: String?) =
   subnet?.let { Regex("""^.+\((.+)\)$""").find(it)?.groupValues?.get(1) } ?: "vpc0"
+
+object LocationConstants {
+  const val DEFAULT_VPC_NAME = "vpc0"
+  const val DEFAULT_SUBNET_PURPOSE = "internal (%s)"
+}
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
@@ -25,7 +25,7 @@ fun ClusterSpec.resolve(): Set<ServerGroup> =
       location = Location(
         account = locations.account,
         region = it.name,
-        vpc = locations.vpc,
+        vpc = locations.vpc ?: error("No vpc supplied or resolved"),
         subnet = locations.subnet ?: error("No subnet purpose supplied or resolved"),
         availabilityZones = it.availabilityZones
       ),
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
@@ -50,9 +50,11 @@ private fun ClusterSpec.resolveLaunchConfiguration(region: SubnetAwareRegionSpec
     ebsOptimized = checkNotNull(overrides[region.name]?.launchConfiguration?.ebsOptimized
       ?: defaults.launchConfiguration?.ebsOptimized),
     iamRole = checkNotNull(overrides[region.name]?.launchConfiguration?.iamRole
-      ?: defaults.launchConfiguration?.iamRole),
+      ?: defaults.launchConfiguration?.iamRole
+      ?: LaunchConfiguration.defaultIamRoleForApp(moniker.app)),
     keyPair = checkNotNull(overrides[region.name]?.launchConfiguration?.keyPair
-      ?: defaults.launchConfiguration?.keyPair),
+      ?: defaults.launchConfiguration?.keyPair
+      ?: LaunchConfiguration.defaultKeyPairForAccount(locations.account)),
     instanceMonitoring = overrides[region.name]?.launchConfiguration?.instanceMonitoring
       ?: defaults.launchConfiguration?.instanceMonitoring ?: false,
     ramdiskId = overrides[region.name]?.launchConfiguration?.ramdiskId
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
@@ -51,10 +51,10 @@ private fun ClusterSpec.resolveLaunchConfiguration(region: SubnetAwareRegionSpec
       ?: defaults.launchConfiguration?.ebsOptimized),
     iamRole = checkNotNull(overrides[region.name]?.launchConfiguration?.iamRole
       ?: defaults.launchConfiguration?.iamRole
-      ?: LaunchConfiguration.defaultIamRoleForApp(moniker.app)),
+      ?: LaunchConfiguration.defaultIamRoleFor(moniker.app)),
     keyPair = checkNotNull(overrides[region.name]?.launchConfiguration?.keyPair
       ?: defaults.launchConfiguration?.keyPair
-      ?: LaunchConfiguration.defaultKeyPairForAccount(locations.account)),
+      ?: LaunchConfiguration.defaultKeyPairFor(locations.account, region.name)),
     instanceMonitoring = overrides[region.name]?.launchConfiguration?.instanceMonitoring
       ?: defaults.launchConfiguration?.instanceMonitoring ?: false,
     ramdiskId = overrides[region.name]?.launchConfiguration?.ramdiskId
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
@@ -30,37 +30,4 @@ data class Health(
   // Note: the default for this in Deck is currently setOf(TerminationPolicy.Default), but we were advised by Netflix
   // SRE to change the default to OldestInstance
   val terminationPolicies: Set<TerminationPolicy> = setOf(TerminationPolicy.OldestInstance)
-) {
-  fun toClusterHealthSpec(omitDefaults: Boolean = false) =
-    ClusterSpec.HealthSpec(
-      if (omitDefaults && cooldown == defaults.cooldown) {
-        null
-      } else {
-        cooldown
-      },
-      if (omitDefaults && warmup == defaults.warmup) {
-        null
-      } else {
-        warmup
-      },
-      if (omitDefaults && healthCheckType == defaults.healthCheckType) {
-        null
-      } else {
-        healthCheckType
-      },
-      if (omitDefaults && enabledMetrics == defaults.enabledMetrics) {
-        null
-      } else {
-        enabledMetrics
-      },
-      if (omitDefaults && terminationPolicies == defaults.terminationPolicies) {
-        null
-      } else {
-        terminationPolicies
-      }
-    )
-
-  companion object {
-    val defaults by lazy { Health() }
-  }
-}
+)
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
@@ -27,7 +27,7 @@ data class Health(
   val warmup: Duration = Duration.ofSeconds(600),
   val healthCheckType: HealthCheckType = HealthCheckType.EC2,
   val enabledMetrics: Set<Metric> = emptySet(),
-  val terminationPolicies: Set<TerminationPolicy> = setOf(TerminationPolicy.Default)
+  val terminationPolicies: Set<TerminationPolicy> = setOf(TerminationPolicy.OldestInstance)
 ) {
   fun toClusterHealthSpec() =
     ClusterSpec.HealthSpec(cooldown, warmup, healthCheckType, enabledMetrics, terminationPolicies)
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
@@ -27,8 +27,21 @@ data class Health(
   val warmup: Duration = Duration.ofSeconds(600),
   val healthCheckType: HealthCheckType = HealthCheckType.EC2,
   val enabledMetrics: Set<Metric> = emptySet(),
-  val terminationPolicies: Set<TerminationPolicy> = setOf(TerminationPolicy.OldestInstance)
+  val terminationPolicies: Set<TerminationPolicy> = setOf(TerminationPolicy.Default)
 ) {
   fun toClusterHealthSpec() =
     ClusterSpec.HealthSpec(cooldown, warmup, healthCheckType, enabledMetrics, terminationPolicies)
+
+  fun toClusterHealthSpecWithoutDefaults() =
+    ClusterSpec.HealthSpec(
+      if (cooldown == defaults.cooldown) { null } else { cooldown },
+      if (warmup == defaults.warmup) { null } else { warmup },
+      if (healthCheckType == defaults.healthCheckType) { null } else { healthCheckType },
+      if (enabledMetrics == defaults.enabledMetrics) { null } else { enabledMetrics },
+      if (terminationPolicies == defaults.terminationPolicies) { null } else { terminationPolicies }
+      )
+
+  companion object {
+    val defaults = Health()
+  }
 }
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
@@ -27,6 +27,8 @@ data class Health(
   val warmup: Duration = Duration.ofSeconds(600),
   val healthCheckType: HealthCheckType = HealthCheckType.EC2,
   val enabledMetrics: Set<Metric> = emptySet(),
+  // Note: the default for this in Deck is currently setOf(TerminationPolicy.Default), but we were advised by Netflix
+  // SRE to change the default to OldestInstance
   val terminationPolicies: Set<TerminationPolicy> = setOf(TerminationPolicy.OldestInstance)
 ) {
   fun toClusterHealthSpec() =
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
@@ -34,12 +34,32 @@ data class Health(
 
   fun toClusterHealthSpecWithoutDefaults() =
     ClusterSpec.HealthSpec(
-      if (cooldown == defaults.cooldown) { null } else { cooldown },
-      if (warmup == defaults.warmup) { null } else { warmup },
-      if (healthCheckType == defaults.healthCheckType) { null } else { healthCheckType },
-      if (enabledMetrics == defaults.enabledMetrics) { null } else { enabledMetrics },
-      if (terminationPolicies == defaults.terminationPolicies) { null } else { terminationPolicies }
-      )
+      if (cooldown == defaults.cooldown) {
+        null
+      } else {
+        cooldown
+      },
+      if (warmup == defaults.warmup) {
+        null
+      } else {
+        warmup
+      },
+      if (healthCheckType == defaults.healthCheckType) {
+        null
+      } else {
+        healthCheckType
+      },
+      if (enabledMetrics == defaults.enabledMetrics) {
+        null
+      } else {
+        enabledMetrics
+      },
+      if (terminationPolicies == defaults.terminationPolicies) {
+        null
+      } else {
+        terminationPolicies
+      }
+    )
 
   companion object {
     val defaults = Health()
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
@@ -31,32 +31,29 @@ data class Health(
   // SRE to change the default to OldestInstance
   val terminationPolicies: Set<TerminationPolicy> = setOf(TerminationPolicy.OldestInstance)
 ) {
-  fun toClusterHealthSpec() =
-    ClusterSpec.HealthSpec(cooldown, warmup, healthCheckType, enabledMetrics, terminationPolicies)
-
-  fun toClusterHealthSpecWithoutDefaults() =
+  fun toClusterHealthSpec(omitDefaults: Boolean = false) =
     ClusterSpec.HealthSpec(
-      if (cooldown == defaults.cooldown) {
+      if (omitDefaults && cooldown == defaults.cooldown) {
         null
       } else {
         cooldown
       },
-      if (warmup == defaults.warmup) {
+      if (omitDefaults && warmup == defaults.warmup) {
         null
       } else {
         warmup
       },
-      if (healthCheckType == defaults.healthCheckType) {
+      if (omitDefaults && healthCheckType == defaults.healthCheckType) {
         null
       } else {
         healthCheckType
       },
-      if (enabledMetrics == defaults.enabledMetrics) {
+      if (omitDefaults && enabledMetrics == defaults.enabledMetrics) {
         null
       } else {
         enabledMetrics
       },
-      if (terminationPolicies == defaults.terminationPolicies) {
+      if (omitDefaults && terminationPolicies == defaults.terminationPolicies) {
         null
       } else {
         terminationPolicies
@@ -64,6 +61,6 @@ data class Health(
     )
 
   companion object {
-    val defaults = Health()
+    val defaults by lazy { Health() }
   }
 }
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
@@ -27,34 +27,8 @@ data class LaunchConfiguration(
   val instanceMonitoring: Boolean = false,
   val ramdiskId: String? = null
 ) {
-  fun toLaunchConfigurationSpec(account: String, application: String, omitDefaults: Boolean = false) =
-    ClusterSpec.LaunchConfigurationSpec(
-      instanceType = instanceType,
-      ebsOptimized = if (omitDefaults && !ebsOptimized) {
-        null
-      } else {
-        ebsOptimized
-      },
-      iamRole = if (omitDefaults && iamRole == defaultIamRoleForApp(application)) {
-        null
-      } else {
-        iamRole
-      },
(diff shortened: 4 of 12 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
```diff
@@ -27,8 +27,21 @@ data class Health(
   val warmup: Duration = Duration.ofSeconds(600),
   val healthCheckType: HealthCheckType = HealthCheckType.EC2,
   val enabledMetrics: Set<Metric> = emptySet(),
-  val terminationPolicies: Set<TerminationPolicy> = setOf(TerminationPolicy.OldestInstance)
+  val terminationPolicies: Set<TerminationPolicy> = setOf(TerminationPolicy.Default)
```
luispollo: I changed this because `Default` is what is the default value in the UI.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandlerTests.kt
```diff
@@ -287,6 +306,51 @@ internal class ClusterHandlerTests : JUnit5Minutests {
...
+        }
+      }
+
+      context("other handling of default properties in cluster export") {
+        before {
+          coEvery { cloudDriverService.activeServerGroup("us-east-1") } returns activeServerGroupResponseEast
+          coEvery { cloudDriverService.activeServerGroup("us-west-2") } returns activeServerGroupResponseWest.withNonDefaultHealthProps()
+        }
+
+        test("export omits properties with default values from complex fields") {
```
luispollo: @robfletcher I tried to add this test under the `derivedContext` up above, but was unable to override the mocked return for `cloudDriverService.activeServerGroup("us-west-2")` that you see here. Please let me know if there's a better way to do this.

luispollo:
@robfletcher @emjburns @asher Please review. This is a WIP -- I need to add handling of the defaults for other fields but wanted to check that I'm going down the right path.

P.S. I'm still unsure whether this is a good idea from the user's perspective... I think it will be confusing to have only a subset of the properties on the export when the others have default values.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
asher: Please revert this, changing the spinnaker default ec2 termination policy to `OldestInstance` was a request from our SRE org that is also reflected in pipelines created by newt / project generators. It's a miss that deck hasn't been updated.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
```diff
@@ -27,8 +27,21 @@ data class Health(
...
     ClusterSpec.HealthSpec(cooldown, warmup, healthCheckType, enabledMetrics, terminationPolicies)
+
+  fun toClusterHealthSpecWithoutDefaults() =
+    ClusterSpec.HealthSpec(
+      if (cooldown == defaults.cooldown) { null } else { cooldown },
+      if (warmup == defaults.warmup) { null } else { warmup },
+      if (healthCheckType == defaults.healthCheckType) { null } else { healthCheckType },
+      if (enabledMetrics == defaults.enabledMetrics) { null } else { enabledMetrics },
+      if (terminationPolicies == defaults.terminationPolicies) { null } else { terminationPolicies }
+      )
```
asher: super minor nit but the indent looks off here, I'm surprised spotless didn't reformat

emjburns: Hmm this same thing happened on another PR by Gal - I wonder if something is funky with spotless?

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandlerTests.kt
```diff
@@ -287,6 +306,51 @@ internal class ClusterHandlerTests : JUnit5Minutests {
...
+            export(exportable)
+          }
+        }
+
+        test("has the expected basic properties") {
+          expectThat(kind)
+            .isEqualTo("cluster")
+          expectThat(apiVersion)
+            .isEqualTo(SPINNAKER_EC2_API_V1)
+          expectThat(spec.regionalIds)
```
asher: @emjburns removed the MultiRegion type and regionalIds along with entity tags. Rebasing and removing this assertion will prob clear up the test failure.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
emjburns: Probably worth adding a code comment too! I didn't know/forgot that this was a request.

luispollo: Ah! Thanks for the heads-up! So are we going to change the default in the UI as well?
image

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
luispollo: Yeah, I was thinking the same thing... I had it catch a couple of indentation issues before, so I dunno...

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandlerTests.kt
luispollo: Yep, I had rebased and didn't notice it broke the test. Fixed.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
luispollo: BTW, if I hit Code > Reformat code in IntelliJ, will that use the google format plugin we're using, and would that match the rules in spotless?

emjburns: You have to set that up. If you run `./gradlew sA` on the command line it'll reformat it. This might help you set up in [intellij](https://stackoverflow.com/questions/42979700/how-to-configure-google-java-code-formatter-in-intellij-idea-17)

emjburns:
Can you post an example of what this looks like for a cluster which has nothing different, and then one health thing different? (Making sure that any sensitive info is changed)

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
asher: I hope so, feel like touching deck? :)

asher: Keep in mind that managed delivery in some ways is a vehicle to free us from past decisions. Whenever you see keel diverge from older norms, it certainly may be in error but it's just as likely intended to make something better.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
luispollo: It was already setup/enabled for me, but it doesn't look like it reformats the code on-the-fly. I had to manually click on Code > Reformat code, FYI.

luispollo:
> Can you post an example of what this looks like for a cluster which has nothing different, and then one health thing different? (Making sure that any sensitive info is changed)

Yes, was about to!

Example where all the health fields have default values (`health` is omitted entirely -- this was existing behavior):
```yaml
metadata:
  serviceAccount: "<redacted>"
kind: "cluster"
spec:
  moniker:
    app: "keeldemo"
    stack: "main"
  imageProvider:
    deliveryArtifact:
      name: "keeldemo"
      type: "deb"
  locations:
    account: "<redacted>"
    subnet: "internal (vpc0)"
    vpc: "vpc0"
    regions:
    - name: "us-west-2"
  launchConfiguration:
    instanceType: "m5.large"
    ebsOptimized: true
    iamRole: "<redacted>"
    keyPair: "<redacted>"
    instanceMonitoring: false
  capacity:
    min: 1
    max: 1
    desired: 1
  dependencies:
    loadBalancerNames:
    - "keeldemo-main-clb"
    securityGroupNames:
    - "<redacted>"
    - "<redacted>"
    - "<redacted>"
```
Example where only one of the fields under `health` has a non-default value (same cluster  @erikmunson used as an example in #581):
```yaml
metadata:
  serviceAccount: "<redacted>"
kind: "cluster"
spec:
  moniker:
    app: "<redacted>"
    stack: "<redacted>"
  imageProvider:
    deliveryArtifact:
      name: "<redacted>"
      type: "deb"
  locations:
    account: "test"
    subnet: "internal (vpc0)"
    vpc: "vpc0"
    regions:
    - name: "us-east-1"
  launchConfiguration:
    instanceType: "m5.xlarge"
    ebsOptimized: false
    iamRole: "<redacted>"
    keyPair: "<redacted>"
    instanceMonitoring: false
  capacity:
    min: 3
    max: 3
    desired: 3
  dependencies:
    securityGroupNames:
    - "<redacted>"
    - "<redacted>"
    - "<redacted>"
  health:
    terminationPolicies:
    - "Default"
```
It looks like there's some consensus we should address defaulting `iamRole` and `keyPairName` as part of this PR, so I'll post other examples when I have those in place.

luispollo:
I added defaults for `iamRole` and `keyPair`, as discussed, but not using a `Resolver`. I felt like the primary use case for resolvers is in the opposite direction, i.e. converting specs to concrete representations of resource, and so the end result is actually _more_ details in the output, not less (which was the point of this PR).

As for `vpc` and `subnet`, I chose to refactor some of the code around defaults to make it more usable outside resolvers.

Please let me know if this is an acceptable implementation.

luispollo:
Sample YAML after all the changes:
```yaml
---
metadata:
  serviceAccount: "delivery-engineering@netflix.com"
kind: "cluster"
spec:
  moniker:
    app: "<redacted>"
    stack: "<redacted>"
  imageProvider:
    deliveryArtifact:
      name: "<redacted>"
      type: "deb"
  locations:
    account: "<redacted>"
    regions:
    - name: "us-east-1"
  launchConfiguration:
    instanceType: "m5.xlarge"
  capacity:
    min: 3
    max: 3
    desired: 3
  dependencies:
    securityGroupNames:
    - "<redacted>"
    - "<redacted>"
    - "<redacted>"
  health:
    terminationPolicies:
    - "Default"
```

@erikmunson Does that look better?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
erikmunson: I'm gonna get this updated in the UI if possible, **but** — what do people think about auto-converting `Default` to `OldestInstance` when we see it in an existing server group we're exporting? Every pipeline created through the UI up until this moment will have been setup with `Default`, so anybody we have export their cluster will end up with `Default` in their YAML config and won't get any benefit whatsoever from us changing the default at the resource level.

erikmunson:
❤️ SO MUCH better. Thanks Luis

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
```diff
@@ -21,9 +21,40 @@ data class LaunchConfiguration(
...
+      instanceMonitoring = if (omitDefaults && !instanceMonitoring) {
+        null
+      } else {
+        instanceMonitoring
+      },
+      ramdiskId = ramdiskId
+    )
+
+  companion object {
+    fun defaultKeyPairForAccount(account: String) = "nf-$account-keypair-a"
```
erikmunson: is there a way we can reach out to clouddriver to ask it for the default keypair for an account rather than assuming it's going to look like this? IIRC clouddriver is aware of the appropriate keypair to use.

emjburns: The [keyPair controller](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/KeyPairController.groovy) is what you should use!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
```diff
@@ -30,7 +30,7 @@ interface Locations<T : RegionSpec> {
    * If not specified here, this should be derived from the [SubnetAwareLocations.subnet] (if
    * present) or use a default VPC name.
    */
-  val vpc: String
+  val vpc: String?
```
emjburns: Why have you made vpc nullable?

luispollo: Because it has a default and should be possible to omit from the spec, which was one of the specific requests from @erikmunson in the issue description (see #581). This is similar to other fields (like `subnet` in the implementation classes) which are already nullable. If it's not nullable, then I can't omit that value when exporting (or at least I didn't know of/find another way with the current Jackson config, or even modifying it).

asher: See https://github.com/spinnaker/keel/commit/bcf575aec24370c8b1dd78fcfa2e8b726daad87c#diff-570d68b946f65312a79f34d24cc10247 - vpc was only just changed from nullable to required.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
luispollo: @emjburns That API just returns the entire list of known key pairs. How do I tell which one is the default for an account? The JSON for each key pair has just these fields:
```json
  {
    "account": "test",
    "region": "us-west-2",
    "keyName": "<redacted>",
    "keyFingerprint": "<redacted>"
  }
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
luispollo: Hmm, I see. Thanks for the heads-up @asher. So what threw me off here is that, unlike other objects that show up in a cluster spec (say, a `ClusterSpec.LaunchConfigurationSpec` for the `launchConfiguration` field), the `locations` field just uses an implementation of `Locations`, which are the same classes used to represent the concrete resources after resolution...

What would you and @emjburns suggest here? Should I add a `ClusterSpec.LocationsSpec` so I can make those fields nullable? -- that seems like a much bigger change, though, I'm a bit reluctant. Or is there a better way of skipping those fields in the JSON output if they match certain values (custom deserializer?).

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
emjburns: Aha! Good call. It looks like the [credentials controller](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/CredentialsController.groovy#L54) will return this information (`GET /credentials/ACCOUNT`) as the value `defaultKeyPair`. It uses some templating for the region, so I think what we should do is cache the default keypairs for all accounts and regions so we can easily look this up.

luispollo: Just adding a note here based on peer review with @emjburns: it looks like we actually have two different naming conventions for key pair names. One is `nf-$account-keypair-a` like I defined above, which we assume takes effect whenever there's not a `defaultKeyPair` specified for that account in the CloudDriver config; the other is `nf-keypair-${account}-${region}`, which many of the accounts in our particular config follow.

In the former case, since the name will match the convention regardless of region, it would be obvious to omit from the export if it also matches what CloudDriver reports as the default key pair name. In the latter, it's a little trickier, but we discussed possibly adding a new type of `Resolver` to query the CloudDriver cache and have some logic to accommodate for both those patterns. However, I'm still unsure how to adequately use the resolver pattern in the export use case since, as I pointed out before, resolvers will actually _apply_ defaults and _add_ detail, whereas the point of this PR is to _remove_ detail from the reverse-engineered specs...

Please share your thoughts.

asher: It looks like deck does nothing but run `const defaultKeyPair = acct.defaultKeyPair.replace('{{region}}', command.region)` to handle the `{{region}}` expansion for some of the `defaultKeyPair` values returned by: https://clouddriver/credentials?expand=true. It seems reasonable to directly replicate that behavior and build a `defaultKeypairForAccount(account: String): String` method somewhere that utilizes a local in-memory cache of the account credential data. I wouldn't try to fit it into the `Resolver` interface for the reasons you gave. Could just be a member of `MemoryCloudDriverCache`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
luispollo: @asher @emjburns Just checking back on this one. What would you advise on this chicken-and-egg problem with using the same data class for concrete resource definitions and for specs (where we want to make fields nullable for the export)?

asher: A `LocationSpec` where everything is nullable seems reasonable to me. @robfletcher any thoughts?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
```diff
@@ -27,34 +27,8 @@ data class LaunchConfiguration(
...
-        null
-      } else {
-        instanceMonitoring
-      },
-      ramdiskId = ramdiskId
-    )
-
   companion object {
-    fun defaultKeyPairForAccount(account: String) = "nf-$account-keypair-a"
+    fun defaultKeyPairForAccount(account: String) = "nf-keypair-$account-{{region}}"
```
luispollo: My understanding is that the "latest and greatest" convention used by newer accounts, hence the change here. This will apply during resolution only.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
luispollo: @emjburns @asher I pushed a commit refactoring the way the default key pair was handled during export to use clouddriver data as discussed + other changes for consistency. Let me know if it looks decent.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
emjburns: We shouldn't have to care about this in keel. Clouddriver will return the default keypair for the account. 
We need to handle expanding the `{{region}}` templating in the right place in keel.

asher: I'm confused as to why this is still present given `clouddriverCache.defaultKeyPairForAccount()`.

luispollo: @emjburns @asher These are used in this block of `ClusterSpec.resolveLaunchConfiguration()`:
https://github.com/spinnaker/keel/blob/a456b51eb6eb9fb35796a1c86601b87de22dc40b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt#L52-L57

I added them because I noticed that `iamRole` and `keyPair` were defaulting to `null` during resolution, and I thought we actually wanted to enforce those conventions when creating resources... If that's not the case, I'll just remove both functions and those references.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
luispollo: @erikmunson My vote goes against doing that as it would change what is currently expected behavior by the app teams of their infrastructure, and under their radar. For these cases, perhaps we could think about education campaigns or informational/warning messages that are integrated into the MD UI to let users know there's a better default, so they can make the change when they're ready, or even opt-in to an MD "auto-resolve"? Something like a "compliance view" of sorts? I have at least one tool in mind that I could show you that implements that concept.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
robfletcher: I'd vote for deferring any more complex handling of keypairs to a further PR that addresses that issue in isolation.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
robfletcher: Yeah, I guess so, although given we have multiple implementations of `Locations` already, are we going to need to double up the spec types as well?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -50,9 +50,11 @@ private fun ClusterSpec.resolveLaunchConfiguration(region: SubnetAwareRegionSpec
     ebsOptimized = checkNotNull(overrides[region.name]?.launchConfiguration?.ebsOptimized
       ?: defaults.launchConfiguration?.ebsOptimized),
     iamRole = checkNotNull(overrides[region.name]?.launchConfiguration?.iamRole
-      ?: defaults.launchConfiguration?.iamRole),
+      ?: defaults.launchConfiguration?.iamRole
+      ?: LaunchConfiguration.defaultIamRoleFor(moniker.app)),
     keyPair = checkNotNull(overrides[region.name]?.launchConfiguration?.keyPair
-      ?: defaults.launchConfiguration?.keyPair),
+      ?: defaults.launchConfiguration?.keyPair
+      ?: LaunchConfiguration.defaultKeyPairFor(locations.account, region.name)),
```
robfletcher: This is great that we can free people from having to specify this

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
```diff
@@ -21,9 +21,16 @@ data class LaunchConfiguration(
...
   val instanceMonitoring: Boolean = false,
   val ramdiskId: String? = null
-)
+) {
+  companion object {
+    // TODO (lpollo): make these configurable?
+    fun defaultKeyPairTemplateFor(account: String) = "nf-keypair-$account-{{region}}"
+    fun defaultKeyPairFor(account: String, region: String) = "nf-keypair-$account-$region"
+    fun defaultIamRoleFor(application: String) = "${application}InstanceProfile"
+  }
```
robfletcher: It certainly would make sense to make this configurable but let's consider it out of scope of this change and add an issue to the backlog

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -526,6 +578,9 @@ class ClusterHandler(
       "spinnaker:stack",
       "spinnaker:details"
     )
+
+    private val REGION_PLACEHOLDER = """\{\{region\}\}""".toRegex()
+    private val REGION_PATTERN = """([a-z]{2}(-gov)?)-([a-z]+)-\d"""
```
robfletcher: Why is one of these done with `toRegex` and not the other? Also I don't understand the first one.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
luispollo: I'm considering this resolved based on @robfletcher's feedback below: https://github.com/spinnaker/keel/pull/593#discussion_r346449171

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
luispollo: Right, which is why I was reluctant to do that in this PR. Unless there are major objections, and given that truly `vpc` is intended to be optional and that we currently use those classes the way we do, I'd propose we keep it nullable for now and revisit this in a separate PR.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
luispollo: https://github.com/spinnaker/keel/issues/604

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: These are used in this block:
https://github.com/spinnaker/keel/blob/3044fd1348b93f012e1e678247a322bad59fd202/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt#L515-L529
If the default key pair configured in clouddriver includes the region placeholder, then we replace that with the pattern to create a new regex on-the-fly and compare against that particular server group's current config.

For example, if the default is `"nf-keypair-123456789012-{{region}}"` and the server group we're looking has a launch config with a key pair like `"nf-keypair-123456789012-us-west-2"`, it will match and we omit the key pair from the export.

Make sense?

P.S. I just realized I could/should have used `REGION_PLACEHOLDER` in that statement with the `replace()`. 😛

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/Health.kt
luispollo: @erikmunson Based on this thread and offline discussion, I added a separate issue to handle this: https://github.com/spinnaker/keel/issues/605

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
robfletcher: If you're searching for the literal string `{{region}}` it also feels like a regex might be overkill ;-)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
robfletcher: Yeah, it's a tough call because it genuinely is nullable in the spec but mandatory in the hydrated resource model. Thus far we've created separate classes when that is the case.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: Yeah, I had more stuff in there initially then ended up with just the region placeholder. Will make the change.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
luispollo: Not in this case. 😄

luispollo: @robfletcher I started down the path of creating a new data class for `SubnetAwareLocationsSpec`, but it's gonna have to implement the `Locations` interface so that `ClusterSpec` can still be `Locatable`, and changing that type in `ClusterSpec` has a ripple effect that touches a ton of places in the code. Plus, I would still need to make the `vpc` field in `Locations` nullable since, well, it can be `null` in specs...

Can we agree on keeping it nullable across the existing interfaces/classes for now and revisit this in a separate PR?

luispollo:
@emjburns @robfletcher @asher Other than the open question on nullable `vpc` fields [above](https://github.com/spinnaker/keel/pull/593#discussion_r344826335), I don't know what else is left to push this over the finish line. If you think it's good enough for now, please add the "ready to merge" label.

luispollo:
@gal-yardeni FYI since there's some touchpoints from this code with #570.

luispollo:
@asher @emjburns @robfletcher Now that I have approver status, I'm tempted to get this merged, _mwahaha_! 😈 Unless someone objects. Going once...

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
asher: I'd still like to see the netflixism that is `defaultKeyPairFor` removed from keel.

asher: (Can happen in a follow-up PR though)

luispollo: I added https://github.com/spinnaker/keel/issues/604 to follow-up.

PR #594, opened 2019-11-08
feat(environments): db schema for stateful constraints

asher:


asher:
Looking for early schema feedback before I start implementing a stateful constraint. This yaml currently produce the following table schema:
```
CREATE TABLE `environment_artifact_constraint` (
  `environment_uid` char(26) COLLATE utf8mb4_unicode_ci NOT NULL,
  `artifact_version` varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL,
  `constraint_type` varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL,
  `status` enum('PENDING','PASS','FAIL','OVERRIDE_PASS','OVERRIDE_FAIL') COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT 'PENDING',
  `judged_by` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
  `judged_at` timestamp NULL DEFAULT NULL,
  `comment` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
  `json` longtext COLLATE utf8mb4_unicode_ci,
  PRIMARY KEY (`environment_uid`,`artifact_version`,`constraint_type`),
  KEY `environment_artifact_constraint_judged_at_idx` (`judged_at`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
```diff
@@ -0,0 +1,71 @@
...
+                  type: timestamp
+                  constraints:
+                    nullable: true
+              - column:
+                  name: comment
+                  type: varchar(255)
+                  constraints:
+                    nullable: true
+              - column:
+                  name: json
```
asher: Intended for storing `constraint_type` specific attributes. A manual judgement constraint likely wouldn't use this at all but a canary constraint may store metatdata about a canary run such as kayenta id, score, report link, etc.

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
```diff
@@ -0,0 +1,77 @@
...
+                  type: longtext
+                  constraints:
+                    nullable: true
+
+  - changeSet:
+      id: mysql-environment-artifact-constraint-status-enum
```
emjburns: Why do you create this column as `int` then modify to enum?

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
```diff
@@ -0,0 +1,77 @@
...
+                  type: varchar(255)
+                  constraints:
+                    nullable: false
+              - column:
+                  name: constraint_type
+                  type: varchar(255)
+                  constraints:
+                    nullable: false
+              - column:
+                  name: created_at
```
emjburns: What is this column for?

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
emjburns: Should this be `details` or something like that vs labeled `json`?

emjburns:
Can you give an example of how this would be used? From reading the schema, I'm guessing that this is something like an list of information that gets stored each time a constraint is evaluated, and then when we make the promote decision we're looking at whether an artifact has passed all the required constraints?

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
emjburns: It seems more like "evaluated at" if it's recording whether or not a constraint is satisfied?

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
asher: I don't love `json` but noticed there's a column of that name on the `resource_event` table. The `resource` table has a `metatdata` longtext column for serialized blobs too though, how about if I change this to that?

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
asher: Liquibase lacks first class enum support, hence the explicit sql statement. Enums are internally stored as ints which is why I used that type as a placeholder.

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
emjburns: oh, i'd say just leave it as `json` then like the `resource_event` table. I like that more than `metadata` because I feel like `metadata` has a specific meaning.

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
emjburns: ahh good to know!

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
asher: It's just a row insertion timestamp, since this table doesn't use a ULID primary key, so that timeouts can be calculated on constraints with external judgement dependencies. A `PENDING` manual-judgement constraint on an environment will eventually surface in the UI as a notification or request for user interaction. If it's ignored, it needs to time out after a while, i.e. if `judged_at == null && created_at < now() - 2days`, we may want to treat it as `FAILED`.

emjburns: Gotcha! That makes sense.

asher:
@emjburns re: usage, I'm initially imagining this as applicable to stateful constraints that can't immediately be computed vs. all constraints. Imagine a canary constraint that deploys a canary cluster within an environment, then evaluates with kayenta for 6 hours. The first eval of the environment/artifact kicks off a canary, writing a row to the new table. Subsequent evals would be able to use that data to see that a canary is in progress, check kayenta, and do nothing if its continuing or pass/fail the constraint if appropriate. I don't think our existing constraints would use this initially but it could be leveraged to allow for manual overrides of constraints such as `allowed-times`.

emjburns:
@asher I think that makes a lot of sense and I think it's a good first pass at what we'd need to do stateful constraints. Thanks for the explanation!

PR #601, opened 2019-11-13
fix(bake): correctly compare package name

emjburns:
The one where an an accidental comparison of `packageName` to itself results in incorrect baking.


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -122,9 +126,11 @@ class ImageService(
     )
       .sortedWith(NamedImageComparator)
       .findLast {
-        AppVersion.parseName(it.appVersion).run {
-          packageName == packageName
```
emjburns: This line was the culprit.

emjburns: I re-wrote this comparison to make it a little clearer (no longer using `.run`)

PR #609, opened 2019-11-19
feat(constraints): initial stateful constraint support

asher:
- with a manual-judgement reference implementation

### DIFF
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
@@ -1,7 +1,9 @@
 package com.netflix.spinnaker.keel.actuation
 
+import com.netflix.spinnaker.keel.api.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.api.StatefulConstraint
 import com.netflix.spinnaker.keel.constraints.ConstraintEvaluator
 import com.netflix.spinnaker.keel.persistence.ArtifactRepository
 import com.netflix.spinnaker.keel.telemetry.ArtifactVersionApproved
@@ -15,7 +17,16 @@ class EnvironmentPromotionChecker(
   private val constraints: List<ConstraintEvaluator<*>>,
   private val publisher: ApplicationEventPublisher
 ) {
+  private val statefulEvaluators: List<ConstraintEvaluator<*>> = constraints
+    .filter { it.constraintType is StatefulConstraint }
+  private val statelessEvaluators = constraints - statefulEvaluators
 
+  /**
+   * TODO (Critical): The EC2 ClusterSpec enables artifact filtering by ArtifactStatus. This is
+   *  currently handled in ImageResolver after artifacts are "approved" for an environment.
+   *  Instead, this needs to happen here or earlier. Otherwise we may deploy canary clusters or
+   *  request manual judgements for artifacts that should never be allowed in a given environemnt.
+   */
   suspend fun checkEnvironments(deliveryConfig: DeliveryConfig) {
     deliveryConfig
       .artifacts
@@ -29,9 +40,13 @@ class EnvironmentPromotionChecker(
               versions.firstOrNull()
             } else {
               versions.firstOrNull { v ->
-                constraints.all { constraintEvaluator ->
-                  !environment.hasSupportedConstraint(constraintEvaluator) || constraintEvaluator.canPromote(artifact, v, deliveryConfig, environment.name)
-                }
+                /**
+                 * Only check stateful evaluators if all stateless evaluators pass. We don't
+                 * want to request judgement or deploy a canary for artifacts that aren't
+                 * deployed to a required environment or outside of an allowed time.
+                 */
+                checkConstraints(statelessEvaluators, artifact, deliveryConfig, v, environment) &&
+                  checkConstraints(statefulEvaluators, artifact, deliveryConfig, v, environment)
               }
             }
 
@@ -65,6 +80,23 @@ class EnvironmentPromotionChecker(
           }
         }
       }
+  }
+
+  private fun checkConstraints(
+    evaluators: List<ConstraintEvaluator<*>>,
+    artifact: DeliveryArtifact,
+    deliveryConfig: DeliveryConfig,
+    version: String,
+    environment: Environment
+  ): Boolean {
+    return if (evaluators.isEmpty()) {
+      true
+    } else {
+      evaluators.all { evaluator ->
+        !environment.hasSupportedConstraint(evaluator) ||
+          evaluator.canPromote(artifact, version, deliveryConfig, environment)
+      }
+    }
   }
 
   private fun Environment.hasSupportedConstraint(constraintEvaluator: ConstraintEvaluator<*>) =
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
@@ -1,0 +1,36 @@
+package com.netflix.spinnaker.keel.api
+
+import java.time.Duration
+import java.time.Instant
+
+data class ConstraintState(
+  val deliveryConfigName: String,
+  val environmentName: String,
+  val artifactVersion: String,
+  val constraintType: String,
+  val status: ConstraintStatus,
+  val createdAt: Instant = Instant.now(),
+  val judgedBy: String? = null,
+  val judgedAt: Instant? = null,
+  val comment: String? = null,
+  // TODO: reconsider type when implementing a constraint that uses attributes
+  val attributes: Map<String, Any?> = emptyMap()
+) {
+  fun canPromote() = status.passes()
+
+  fun failed() = status.failed()
+
+  fun timedOut(timeout: Duration, now: Instant) =
+    createdAt.plus(timeout).isBefore(now)
+}
+
+enum class ConstraintStatus(private val passed: Boolean, private val failed: Boolean) {
+  PENDING(false, false),
+  PASS(true, false),
+  FAIL(false, true),
+  OVERRIDE_PASS(true, false),
+  OVERRIDE_FAIL(false, true);
+
+  fun passes() = passed
+  fun failed() = failed
+}
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
@@ -3,6 +3,7 @@ package com.netflix.spinnaker.keel.constraints
 import com.netflix.spinnaker.keel.api.Constraint
 import com.netflix.spinnaker.keel.api.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.DeliveryConfig
+import com.netflix.spinnaker.keel.api.Environment
 
 interface ConstraintEvaluator<T : Constraint> {
 
@@ -31,6 +32,6 @@ interface ConstraintEvaluator<T : Constraint> {
     artifact: DeliveryArtifact,
     version: String,
     deliveryConfig: DeliveryConfig,
-    targetEnvironment: String
+    targetEnvironment: Environment
   ): Boolean
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
@@ -3,6 +3,7 @@ package com.netflix.spinnaker.keel.constraints
 import com.netflix.spinnaker.keel.api.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.DependsOnConstraint
+import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.constraints.ConstraintEvaluator.Companion.getConstraintForEnvironment
 import com.netflix.spinnaker.keel.persistence.ArtifactRepository
 import org.springframework.stereotype.Component
@@ -18,9 +19,9 @@ class DependsOnConstraintEvaluator(
     artifact: DeliveryArtifact,
     version: String,
     deliveryConfig: DeliveryConfig,
-    targetEnvironment: String
+    targetEnvironment: Environment
   ): Boolean {
-    val constraint = getConstraintForEnvironment(deliveryConfig, targetEnvironment, constraintType)
+    val constraint = getConstraintForEnvironment(deliveryConfig, targetEnvironment.name, constraintType)
 
     val requiredEnvironment = deliveryConfig
       .environments
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluator.kt
@@ -1,0 +1,67 @@
+package com.netflix.spinnaker.keel.constraints
+
+import com.netflix.spinnaker.keel.api.ConstraintState
+import com.netflix.spinnaker.keel.api.ConstraintStatus
+import com.netflix.spinnaker.keel.api.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.api.ManualJudgementConstraint
+import com.netflix.spinnaker.keel.constraints.ConstraintEvaluator.Companion.getConstraintForEnvironment
+import com.netflix.spinnaker.keel.persistence.DeliveryConfigRepository
+import org.springframework.stereotype.Component
+import java.time.Clock
+
+@Component
+class ManualJudgementConstraintEvaluator(
+  private val deliveryConfigRepository: DeliveryConfigRepository,
+  private val clock: Clock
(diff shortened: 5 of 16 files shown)


asher:
Up next after this is approved: REST controller methods to interact with constraint state (i.e. to load pending manual judgements, provide a manual judgement, override a failed canary constraint, etc.)

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -12,10 +16,21 @@ import org.springframework.stereotype.Component
...
+    .filterIsInstance<StatefulConstraintEvaluator<*>>()
+  private val statelessEvaluators: List<ConstraintEvaluator<*>> = constraints
+    .filter { !statefulEvaluators.contains(it) }
 
+  /**
+   * TODO (Critical): The EC2 ClusterSpec enables artifact filtering by ArtifactStatus. This is
+   *  currently handled in ImageResolver after artifacts are "approved" for an environment.
+   *  Instead, this needs to happen here or earlier. Otherwise we may deploy canary clusters or
+   *  request manual judgements for artifacts that should never be allowed in a given environemnt.
+   */
```
asher: We need to refactor how ArtifactStatus constraints are defined and evaluated. This likely means establishing a rule of only one ArtifactStatus restriction per environment and may include making it or a top-level `Environment` property.

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
```diff
@@ -75,3 +75,11 @@ databaseChangeLog:
...
               - column:
                   name: judged_at
+
+  - changeSet:
+      id: mysql-environment-artifact-constraint-status-not-enum
+      author: asher
+      changes:
+        - sql:
+            dbms: mysql
+            sql: ALTER TABLE `environment_artifact_constraint` MODIFY COLUMN `status` VARCHAR(64)
```
asher: Jooq not properly supporting enums should have been a signal that jooq auto-class generation wouldn't support them either.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
```diff
@@ -0,0 +1,34 @@
...
+  val judgedAt: Instant? = null,
+  val comment: String? = null,
+  val json: String? = null
+) {
+  fun canPromote() = status.passes()
+
+  fun failed() = status.failed()
+
+  fun timedOut(timeout: Duration, now: Instant) =
+    createdAt.plus(timeout).isBefore(now)
```
luispollo: `isAfter(now)`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
```diff
@@ -0,0 +1,34 @@
...
+}
+
+enum class ConstraintStatus(private val passed: Boolean, private val failed: Boolean) {
+  PENDING(false, false),
+  PASS(true, false),
+  FAIL(false, true),
+  OVERRIDE_PASS(true, false),
+  OVERRIDE_FAIL(false, true);
```
luispollo: I find these a little confusing as they have the same values as `PASS` and `FAIL`, respectively. What's the indented use?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
asher: if `createdAt.plus(timeout)` is after `now`, the constraint hasn't timed out.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
asher: Canary constraint fails, status -> FAIL
Human manually overrides the canary failure, status -> OVERRIDE_PASS

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Environment.kt
```diff
@@ -1,8 +1,36 @@
 package com.netflix.spinnaker.keel.api
 
+import com.fasterxml.jackson.annotation.JsonIgnore
+
 data class Environment(
   val name: String,
+  @JsonIgnore
+  var id: String? = null,
```
luispollo: Couldn't we default this to `randomUID().toString()` instead to avoid making it nullable? When loading from the repo it will be non-null, otherwise it seems we're assuming it's non-null everywhere anyway...

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
```diff
@@ -31,6 +33,9 @@ interface ConstraintEvaluator<T : Constraint> {
     artifact: DeliveryArtifact,
     version: String,
     deliveryConfig: DeliveryConfig,
-    targetEnvironment: String
+    targetEnvironment: Environment,
+    priorState: ConstraintState? = null
```
luispollo: Shouldn't this be defined in `StatefulConstraintEvaluator` instead, since it's about storing state?

asher: I think the current `canPromote` signature needs a refactor but I'd rather not have a different interface for constraint evaluation depending on type, at least not yet.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Environment.kt
asher: That would break things. The way we're assigning uuids on objects such as `Environment` but hiding them from users makes it difficult to utilize them directly (in places where using a uuid directly is more efficient.)

Let me explore mapping `ConstraintState` to an environment via `DeliveryConfig.name` + `Environment.name` instead of always using environment.uuid, would enable reverting the addition of `id` on `Environment`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Environment.kt
```diff
@@ -1,7 +1,10 @@
 package com.netflix.spinnaker.keel.api
 
+import com.fasterxml.jackson.annotation.JsonIgnore
+
 data class Environment(
   val name: String,
+  @JsonIgnore
```
luispollo: This line needs to be removed too, right?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
luispollo: IIUC, the only use for this is in `EnvironmentPromotionChecker`, where we get _current_ state for the constraint immediately before passing it in to `canPromote`. Maybe we just add `currentState` with a getter in `StatefulConstraintEvaluator` which you can call from the `canPromote` override in sub-classes? That would avoid any references to state in the base interface. Just a thought.

luispollo: Sorry, not a getter, a function like `currentStateFor(<args>)`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
```diff
@@ -31,6 +33,9 @@ interface ConstraintEvaluator<T : Constraint> {
...
   ): Boolean
 }
+
+interface StatefulConstraintEvaluator<T : Constraint> : ConstraintEvaluator<T>
```
robfletcher: Does this really need to be a separate interface?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
```diff
@@ -18,9 +20,10 @@ class DependsOnConstraintEvaluator(
     artifact: DeliveryArtifact,
     version: String,
     deliveryConfig: DeliveryConfig,
-    targetEnvironment: String
+    targetEnvironment: Environment,
+    priorState: ConstraintState?
   ): Boolean {
-    val constraint = getConstraintForEnvironment(deliveryConfig, targetEnvironment, constraintType)
+    val constraint = getConstraintForEnvironment(deliveryConfig, targetEnvironment.name, constraintType)
```
robfletcher: My feeling on simplifying this is that we should have a way to get `DeliveryConfig` from `Environment` (i.e. `@JsonBackReference` or whatever that thing is), and then just pass the `Environment` object in here. Outside of this change, tho.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluator.kt
```diff
@@ -0,0 +1,62 @@
...
+  override fun canPromote(
+    artifact: DeliveryArtifact,
+    version: String,
+    deliveryConfig: DeliveryConfig,
+    targetEnvironment: Environment,
+    priorState: ConstraintState?
+  ): Boolean {
+    val constraint = getConstraintForEnvironment(deliveryConfig, targetEnvironment.name, constraintType)
+
+    val state = priorState ?: ConstraintState(
```
robfletcher: I presume we’ll eventually want to send a notification if `priorState == null` so that users know to do something?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -105,13 +109,16 @@ class SqlDeliveryConfigRepository(
...
+        val environmentUID = (
+          jooq
+            .select(ENVIRONMENT.UID)
+            .from(ENVIRONMENT)
+            .where(ENVIRONMENT.DELIVERY_CONFIG_UID.eq(uid))
+            .and(ENVIRONMENT.NAME.eq(environment.name))
+            .fetchOne(ENVIRONMENT.UID)
+            ?: randomUID().toString()
+          )
+          .also {
```
robfletcher: Doesn’t this change mean we’ll attempt to insert duplicate environment ids?

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
robfletcher: What was the column type previously? We should never store enums by ordinal!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
```diff
@@ -0,0 +1,35 @@
...
+  val deliveryConfigName: String,
+  val environmentName: String,
+  val artifactVersion: String,
+  val constraintType: String,
+  val status: ConstraintStatus,
+  val createdAt: Instant = Instant.now(),
+  val judgedBy: String? = null,
+  val judgedAt: Instant? = null,
+  val comment: String? = null,
+  val json: String? = null
```
robfletcher: What’s this field for? Would we be better off with a generic type if the state requires additional data?

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -67,8 +86,34 @@ class EnvironmentPromotionChecker(
...
+    deliveryConfig: DeliveryConfig,
+    version: String,
+    environment: Environment
+  ): Boolean {
+    if (evaluators.isEmpty()) {
+      return true
+    } else {
+      return evaluators.all { evaluator ->
+        val constraint = environment.constraintForEvaluator(evaluator)
+        if (constraint == null) {
```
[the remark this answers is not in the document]
luispollo: Sorry to intrude, but I wanted to see IIUC: constraint evaluators are just `@Components` wired into that list. An environment may or may not have a constraint matching each of those evaluators, hence it can be null. Right? 😃

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
asher: Nope, will remove. I made it so that I can filter evaluators on stateful vs. stateless to enforce evaluation order, but looks like I can do that fine via the original interface via: 
``` 
private val statefulEvaluators: List<ConstraintEvaluator<*>> = constraints
    .filter { it.constraintType is StatefulConstraint}
```

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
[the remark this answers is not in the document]
asher: @luispollo is right, most environments aren't going to utilize every constraint type, so `constraintForEvaluator` will only return a constraint a minority of times. When it doesn't, it means the evaluator in question doesn't get a chance to veto, so it counts as an automatic `true` for `evaluators.all {}` instead of having its `canPromote()` called.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluator.kt
luispollo: That makes sense to me as well.

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
asher: It was originally a native sql ENUM.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -29,9 +42,13 @@ class EnvironmentPromotionChecker(
               versions.firstOrNull()
             } else {
               versions.firstOrNull { v ->
-                constraints.all { constraintEvaluator ->
-                  !environment.hasSupportedConstraint(constraintEvaluator) || constraintEvaluator.canPromote(artifact, v, deliveryConfig, environment.name)
-                }
+                /**
+                 * Only check stateful evaluators if all stateless evaluators pass. We don't
```
luispollo: This actually brings up a question on the ordering of constraints for me. Since you can't tell a `List` from `Set` in YAML, I suspect users might assume order matters when specifying constraints for an environment (in which case it would not be OK to make this assumption here), even though they're modeled as a `Set` internally. Should we perhaps clarify that order does _not_ matter in the documentation?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
asher: No, this is fixing a preexisting bug that meant updates to environment constraints and notifications would never be persisted.  Can't duplicate environment ids, it's a primary key, but we want the insert to always run to get the `onDuplicateKeyUpdate()`.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
asher: Sure, please do update documentation :)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
luispollo: @asher I still think `priorState` should be removed from the interface. The `canPromote` method has all the info it needs without it to call `deliveryConfigRepository.getConstraintState(deliveryConfig.name, targetEnvironment.name, version, constraint.type)` to get the current constraint state. I also feel like `EnvironmentPromotionChecker` should not be concerned with the underlying implementation of `ConstraintEvaluators`, including whether they store state or not.

asher: Sure, we can just duplicate code in evaluators if you think it's the more elegant route.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluator.kt
asher: The `.also` block this comment is above already contains:
```
          // TODO: Emit an event here, requesting a manual judgement
```
which I think is clear.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
luispollo: Is the `allowed-times` constraint rolled out to main yet, BTW? I can add an example of that while I'm at it.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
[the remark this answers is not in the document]
robfletcher: Oh, I see.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
luispollo: Like you said earlier, there's really not a super elegant solution, but that would be my vote, yes.

PR #613, opened 2019-11-21
feat(export): Titus cluster export

### LINKED ISSUE #570: Export Titus Resource

gal-yardeni:
Closes #570.

Other changes of note in this PR:
- Removes the top-level `container` field in `TitusClusterSpec` which was redundant with the `@JsonUnwrapped` `defaults.container` one and caused that field to show up twice in JSON/YAML

### DIFF
--- a/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
+++ b/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
@@ -35,6 +35,7 @@ import com.netflix.spinnaker.keel.api.titus.exceptions.TitusAccountConfiguration
 import com.netflix.spinnaker.keel.clouddriver.CloudDriverCache
 import com.netflix.spinnaker.keel.clouddriver.CloudDriverService
 import com.netflix.spinnaker.keel.clouddriver.ResourceNotFound
+import com.netflix.spinnaker.keel.clouddriver.model.Resources
 import com.netflix.spinnaker.keel.clouddriver.model.TitusActiveServerGroup
 import com.netflix.spinnaker.keel.diff.ResourceDiff
 import com.netflix.spinnaker.keel.events.Task
@@ -143,30 +144,17 @@ class TitusClusterHandler(
     val spec = TitusClusterSpec(
       moniker = exportable.moniker,
       locations = locations,
-      container = serverGroups.values.first().container,
-      // containr
-
-//      imageProvider = if (base.buildInfo?.packageName != null) {
-//        ArtifactImageProvider(
-//          deliveryArtifact = DeliveryArtifact(name = base.buildInfo.packageName!!))
-//      } else {
-//        null
-//      },
-
-      _defaults = TitusServerGroupSpec(
-        capacity = base.capacity,
-        dependencies = base.dependencies,
-        tags = base.tags
-      ),
+      container = base.container,
+      _defaults = base.exportSpec(),
       overrides = mutableMapOf()
     )
 
-//    spec.generateOverrides(
-//      exportable.account,
-//      exportable.moniker.app,
-//      serverGroups
-//        .filter { it.value.location.region != base.location.region }
-//    )
+    spec.generateOverrides(
+      exportable.account,
+      exportable.moniker.app,
+      serverGroups
+        .filter { it.value.location.region != base.location.region }
+    )
 
     return SubmittedResource(
       apiVersion = supportedKind.apiVersion,
@@ -271,6 +259,14 @@ class TitusClusterHandler(
         ResourceDiff(desired, current?.get(region))
       }
 
+  private fun TitusClusterSpec.generateOverrides(account: String, application: String, serverGroups: Map<String, TitusServerGroup>) =
+    serverGroups.forEach { (region, serverGroup) ->
+      if (ResourceDiff(serverGroup, defaults).hasChanges()) {
+        // FIXME (lpollo): return diff as TitusServerGroupSpec including only different values, other fields null
+        (overrides as MutableMap)[region] = serverGroup.exportSpec()
+      }
+    }
+
   private suspend fun CloudDriverService.getServerGroups(resource: Resource<TitusClusterSpec>): Iterable<TitusServerGroup> =
     getServerGroups(resource.spec.locations.account,
       resource.spec.moniker,
@@ -364,4 +360,28 @@ class TitusClusterHandler(
 
   private fun Instant.iso() =
     atZone(ZoneId.systemDefault()).format(DateTimeFormatter.ISO_DATE_TIME)
+
+  private fun TitusServerGroup.exportSpec() =
+    TitusServerGroupSpec(
+      capacity = capacity,
+      capacityGroup = capacityGroup,
+      constraints = constraints,
+      container = container,
+      dependencies = dependencies,
+      entryPoint = entryPoint,
+      env = env,
+      iamProfile = iamProfile,
+      migrationPolicy = migrationPolicy,
+      resources = resources.exportSpec(),
+      tags = tags
+    )
+
+  private fun Resources.exportSpec() =
+    ResourcesSpec(
+      cpu = cpu,
+      disk = disk,
+      gpu = gpu,
+      memory = memory,
+      networkMbps = networkMbps
+    )
 }
--- a/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
+++ b/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
@@ -121,6 +121,7 @@ class TitusClusterHandler(
     }
 
   override suspend fun export(exportable: Exportable): SubmittedResource<TitusClusterSpec> {
+    // TODO[GY] : remove unchanged/default fields from response (across all export responses)
     val serverGroups = cloudDriverService.getServerGroups(
       exportable.account,
       exportable.moniker,
@@ -261,8 +262,6 @@ class TitusClusterHandler(
       val workingSpec = serverGroup.exportSpec()
       val diff = ResourceDiff(workingSpec, defaults)
       if (diff.hasChanges()) {
-        log.debug(diff.toDebug())
-        // man, I wish kotlin had **kwargs... :-P
         (overrides as MutableMap)[region] = TitusServerGroupSpec(
           capacity = if ("capacity" in diff.affectedRootPropertyNames) {
             workingSpec.capacity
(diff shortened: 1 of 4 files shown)


gal-yardeni:
The export titus currently produce the below as a response:
`{
  "metadata": {
    "serviceAccount": "delivery-engineering@netflix.com"
  },
  "apiVersion": "titus.spinnaker.netflix.com/v1",
  "kind": "cluster",
  "spec": {
    "moniker": {
      "app": "serverlablpollo"
    },
    "deployWith": {
      "strategy": "red-black",
      "rollbackOnFailure": true,
      "resizePreviousToZero": false,
      "maxServerGroups": 2,
      "delayBeforeDisable": "PT0S",
      "delayBeforeScaleDown": "PT0S"
    },
    "locations": {
      "account": "titustestvpc",
      "vpc": "vpc0",
      "regions": [
        {
          "name": "us-east-1"
        },
        {
          "name": "us-west-2"
        },
        {
          "name": "eu-west-1"
        }
      ]
    },
    "overrides": {
      "us-west-2": {
        "capacity": {
          "min": 2,
          "max": 2,
          "desired": 2
        }
      }
    },
    "capacity": {
      "min": 1,
      "max": 1,
      "desired": 1
    },
    "capacityGroup": "serverlablpollo",
    "constraints": {
      "hard": {},
      "soft": {}
    },
    "container": {
      "organization": "emburns",
      "image": "spin-titus-demo",
      "digest": "sha256:3b919225c332aa46aeba3eda7c9a6be6b3bd87f09c3bebed1863a3c8dd4e28ca"
    },
    "dependencies": {
      "securityGroupNames": [
        "nf-infrastructure",
        "serverlablpollo",
        "nf-datacenter"
      ]
    },
    "entryPoint": "",
    "env": {},
    "iamProfile": "serverlablpolloInstanceProfile",
    "migrationPolicy": {
      "type": "systemDefault"
    },
    "resources": {
      "cpu": 1,
      "disk": 10000,
      "gpu": 0,
      "memory": 512,
      "networkMbps": 128
    },
    "tags": {}
  }
}`

Will open a different PR for both Titus and ec2 export result to be less verbose and show only "interesting" data to the user.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -113,6 +120,48 @@ class TitusClusterHandler(
...
+    if (serverGroups.isEmpty()) {
+      throw ResourceNotFound("Could not find cluster: ${exportable.moniker.name} " +
+        "in account: ${exportable.account} for export")
+    }
+
+    val base = serverGroups.values.first()
+
+    val locations = SimpleLocations(
+      account = exportable.account,
+      regions = (exportable.regions.map {
```
emjburns: This field (`exportable.regions`) is actually a list of all the regions that aren't deprecated in the account I think. Right now, when you export you get a spec with all the regions defined even if you export a resource in only one region. I think that `regions` here should be `serverGroups.keys` so that we get the actual regions.

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
```diff
@@ -141,6 +140,14 @@ class TitusClusterHandlerTests : JUnit5Minutests {
     spec = spec
   )
 
+  val exportable = Exportable(
+    account = spec.locations.account,
+    serviceAccount = "keel@spinnaker",
+    moniker = spec.moniker,
+    regions = spec.locations.regions.map { it.name }.toSet(),
```
emjburns: this is actually all regions in the account

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
```diff
@@ -433,6 +440,52 @@ class TitusClusterHandlerTests : JUnit5Minutests {
...
+
+        test("export omits properties with default values from complex fields") {
+          val exported = runBlocking {
+            export(exportable)
+          }
+          expectThat(exported.spec.defaults.container)
+            .isNotNull()
+
+          expectThat(exported.spec.defaults.iamProfile)
+            .isNotNull()
```
emjburns: If this test is testing that we omit things, should we be checking if things are not null?

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
```diff
@@ -433,6 +440,52 @@ class TitusClusterHandlerTests : JUnit5Minutests {
             .containsDistinctElements()
         }
       }
+      context("the cluster has active server groups") {
+        before {
+          coEvery { cloudDriverService.titusActiveServerGroup("us-east-1") } returns activeServerGroupResponseEast
+          coEvery { cloudDriverService.titusActiveServerGroup("us-west-2") } returns activeServerGroupResponseWest
+        }
+
+        derivedContext<SubmittedResource<TitusClusterSpec>>("exported titus cluster spec") {
```
emjburns: I have a poor understanding of how fixtures work. Just a question, why is a derived context needed here?

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
asher: The mocked clouddriver response contains `iamProfile = moniker.app + "InstanceProfile"` which the export should treat as a default and exclude. The same logic to check if the default iam role is used for ec2 (https://github.com/spinnaker/keel/blob/master/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt#L492) also applies to titus. Then, this test can be flipped to an `isNull` check as @emjburns noted.

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
luispollo: A `derivedContext` allows you to transform the fixture from the parent context into another type which sometimes makes it easier to test specific aspects of objects (like in this case the submitted resource).

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
gal-yardeni: This is actually the same as       `regions = setOf(SimpleRegionSpec("us-east-1"), SimpleRegionSpec("us-west-2"))` , we are just refering to this map here.

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
gal-yardeni: @asher you are right, as @emjburns suggested, we will create a different PR to exclude the default fields.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
luispollo: Fixed.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -208,17 +259,94 @@ class TitusClusterHandler(
         ResourceDiff(desired, current?.get(region))
       }
 
+  private fun TitusClusterSpec.generateOverrides(serverGroups: Map<String, TitusServerGroup>) =
+    serverGroups.forEach { (region, serverGroup) ->
+      val workingSpec = serverGroup.exportSpec()
+      val diff = ResourceDiff(workingSpec, defaults)
+      if (diff.hasChanges()) {
+        (overrides as MutableMap)[region] = TitusServerGroupSpec(
```
luispollo: I wish kotlin had `**kwargs`... 😃

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
luispollo: @emjburns An example I found in the docs that helped me understand: https://github.com/dmcg/minutest/blob/master/core/src/test/kotlin/dev/minutest/examples/DerivedContextExampleTests.kt

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -113,6 +120,50 @@ class TitusClusterHandler(
...
+      _defaults = base.exportSpec(),
+      overrides = mutableMapOf()
+    )
+
+    spec.generateOverrides(
+      serverGroups
+        .filter { it.value.location.region != base.location.region }
+    )
+
+    // TODO: omit defaults from exported spec
```
emjburns: same todo as above?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
emjburns: Since everything on `TitusServerGroupSpec` has a default value of null, you could do something like:
```kotlin
var newSpec = TitusServerGroupSpec()
if ("capacity" in diff.affectedRootPropertyNames) {
  newSpec = newSpec.copy(capacity = workingSpec.capacity)
}
// continued for each property
```
That would remove all the ` else { null }` bits which take up a lot of space, and I think simplify how this reads.

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
emjburns: Okay, makes sense! Thanks.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
gal-yardeni: @emjburns great suggestion! will make the fix.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
gal-yardeni: Removed.

PR #614, opened 2019-11-21
Misc tidying

robfletcher:
* made `Constraint` not a sealed class, as we want contributors to be able to create new ones
* turned `EnvironmentResolver` into `TaskLauncher` to pull out the common code for Orca interaction
* fixed some cruft


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
```diff
@@ -89,30 +87,15 @@ class ApplicationLoadBalancerHandler(
...
-                  OrchestrationTrigger(
-                    correlationId = "${resource.id}:${desired.location.region}",
-                    notifications = notifications)
-                )
-              )
-              .let {
-                log.info("Started task ${it.ref} to $description")
-                Task(id = it.taskId, name = description)
-              }
+            taskLauncher.submitJobToOrca(
```
luispollo: Much cleaner, thanks for doing this! ❤️

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/TaskLauncher.kt
```diff
@@ -0,0 +1,74 @@
...
+ */
+class TaskLauncher(
+  private val orcaService: OrcaService,
+  private val deliveryConfigRepository: DeliveryConfigRepository
+) {
+  suspend fun submitJobToOrca(
+    resource: Resource<*>,
+    description: String,
+    correlationId: String,
+    job: Map<String, Any?>
```
luispollo: Should we make this a `vararg` and rename the function to `submitJobsToOrca` in case we ever want to pass in more than one job?

robfletcher: Yeah. I was thinking the same, but I figure we can do it any time.

PR #615, opened 2019-11-21
fix(resolution): Gracefully handle missing appversion tag

### LINKED ISSUE #555: Handle missing appversion tag on server group
If the AMI deployed to a server group is missing the `appversion` tag `ClusterHandler` will fail to resolve the current state. Instead we should resolved to a model with `appversion: null`, in which case the resource check will re-deploy.

luispollo:
Fixes #555 by allowing the app version field in the launch configuration to be null instead of throwing an exception.

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
@@ -19,7 +19,7 @@ package com.netflix.spinnaker.keel.api.ec2
 
 data class LaunchConfiguration(
   val imageId: String,
-  val appVersion: String,
+  val appVersion: String? = null,
   val instanceType: String,
   val ebsOptimized: Boolean = false,
   val iamRole: String,
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
@@ -19,7 +19,7 @@ package com.netflix.spinnaker.keel.api.ec2
 
 data class LaunchConfiguration(
   val imageId: String,
-  val appVersion: String? = null,
+  val appVersion: String?,
   val instanceType: String,
   val ebsOptimized: Boolean = false,
   val iamRole: String,
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -379,7 +379,7 @@ class ClusterHandler(
         if (them.distinctBy { it.launchConfiguration.appVersion }.size == 1) {
           publisher.publishEvent(ArtifactVersionDeployed(
             resourceId = resource.id,
-            artifactVersion = them.first().launchConfiguration.appVersion
+            artifactVersion = them.first().launchConfiguration.appVersion!!
           ))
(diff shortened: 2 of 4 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -395,7 +395,7 @@ class ClusterHandler(
         if (them.distinctBy { it.launchConfiguration.appVersion }.size == 1) {
           publisher.publishEvent(ArtifactVersionDeployed(
             resourceId = resource.id,
-            artifactVersion = them.first().launchConfiguration.appVersion
+            artifactVersion = them.first().launchConfiguration.appVersion!!
```
gal-yardeni: Isn't '!!' throws 'KotlinNullPointerException'? Do we want it here?

luispollo: Not exactly. It converts a nullable type (say `String?`) to the non-nullable (`String`) counter-part, and only throws the KNPE if the expression is null. You generally only use when you're sure the value won't be null but the compiler can't infer it. I need to add a check here though to make sure it's not.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
```diff
@@ -19,7 +19,7 @@ package com.netflix.spinnaker.keel.api.ec2
 
 data class LaunchConfiguration(
   val imageId: String,
-  val appVersion: String,
+  val appVersion: String? = null,
```
robfletcher: Do we need the default value?

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandlerTests.kt
```diff
@@ -395,6 +397,59 @@ internal class ClusterHandlerTests : JUnit5Minutests {
...
+        runBlocking {
+          upsert(resource, diff)
+        }
+
+        val slot = slot<OrchestrationRequest>()
+        coVerify { orcaService.orchestrate("keel@spinnaker", capture(slot)) }
+
+        expectThat(slot.captured.job.first()) {
+          get("type").isEqualTo("createServerGroup")
+          // TODO (lpollo): since we rely on the tag when looking up server groups, should we add the tag when creating them?
```
luispollo: @robfletcher Do you have an opinion on this?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
luispollo: I guess not.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandlerTests.kt
luispollo: @robfletcher FWIW, I checked clouddriver and can see where the tag is being used/parsed, but not created/added, so this remains a mystery to me...

robfletcher: Yeah, I'm honestly not 100% sure. I _believe_ it's in CloudDriver somewhere.

luispollo: OK. Removing that `TODO` then.

PR #627, opened 2019-11-26
fix(resolution): Fix EC2 cluster resolution for key pairs to match export

luispollo:
This PR addresses an issue reported internally while submitting a cluster resource from an export. The logic to detect/omit defaults for the key pair name was not matching what was used in resource resolution.

TO-DO:
- [x] Add logic to compare default key pair from clouddriver with overrides and see if they make sense together
- [x] Add tests

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
@@ -56,8 +56,7 @@ private fun ClusterSpec.resolveLaunchConfiguration(region: SubnetAwareRegionSpec
       ?: defaults.launchConfiguration?.iamRole
       ?: LaunchConfiguration.defaultIamRoleFor(moniker.app)),
     keyPair = checkNotNull(overrides[region.name]?.launchConfiguration?.keyPair
-      ?: defaults.launchConfiguration?.keyPair
-      ?: LaunchConfiguration.defaultKeyPairFor(locations.account, region.name)),
+      ?: defaults.launchConfiguration?.keyPair),
     instanceMonitoring = overrides[region.name]?.launchConfiguration?.instanceMonitoring
       ?: defaults.launchConfiguration?.instanceMonitoring ?: false,
     ramdiskId = overrides[region.name]?.launchConfiguration?.ramdiskId
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
@@ -51,14 +51,16 @@ private fun ClusterSpec.resolveLaunchConfiguration(region: SubnetAwareRegionSpec
     instanceType = checkNotNull(overrides[region.name]?.launchConfiguration?.instanceType
       ?: defaults.launchConfiguration?.instanceType),
     ebsOptimized = checkNotNull(overrides[region.name]?.launchConfiguration?.ebsOptimized
-      ?: defaults.launchConfiguration?.ebsOptimized),
+      ?: defaults.launchConfiguration?.ebsOptimized
+      ?: LaunchConfiguration.DEFAULT_EBS_OPTIMIZED),
     iamRole = checkNotNull(overrides[region.name]?.launchConfiguration?.iamRole
       ?: defaults.launchConfiguration?.iamRole
       ?: LaunchConfiguration.defaultIamRoleFor(moniker.app)),
     keyPair = checkNotNull(overrides[region.name]?.launchConfiguration?.keyPair
       ?: defaults.launchConfiguration?.keyPair),
     instanceMonitoring = overrides[region.name]?.launchConfiguration?.instanceMonitoring
-      ?: defaults.launchConfiguration?.instanceMonitoring ?: false,
+      ?: defaults.launchConfiguration?.instanceMonitoring
+      ?: LaunchConfiguration.DEFAULT_INSTANCE_MONITORING,
     ramdiskId = overrides[region.name]?.launchConfiguration?.ramdiskId
       ?: defaults.launchConfiguration?.ramdiskId
   )
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
@@ -22,13 +22,15 @@ data class LaunchConfiguration(
   val appVersion: String?,
   val baseImageVersion: String?,
   val instanceType: String,
-  val ebsOptimized: Boolean = false,
+  val ebsOptimized: Boolean = DEFAULT_EBS_OPTIMIZED,
   val iamRole: String,
   val keyPair: String,
-  val instanceMonitoring: Boolean = false,
+  val instanceMonitoring: Boolean = DEFAULT_INSTANCE_MONITORING,
   val ramdiskId: String? = null
 ) {
   companion object {
+    const val DEFAULT_EBS_OPTIMIZED = false
+    const val DEFAULT_INSTANCE_MONITORING = false
     // TODO (lpollo): make these configurable?
     fun defaultKeyPairFor(account: String, region: String) = "nf-keypair-$account-$region"
     fun defaultIamRoleFor(application: String) = "${application}InstanceProfile"
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
@@ -30,7 +30,6 @@ data class LaunchConfiguration(
 ) {
   companion object {
     // TODO (lpollo): make these configurable?
-    fun defaultKeyPairTemplateFor(account: String) = "nf-keypair-$account-{{region}}"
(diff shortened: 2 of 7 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
```diff
@@ -30,7 +30,6 @@ data class LaunchConfiguration(
 ) {
   companion object {
     // TODO (lpollo): make these configurable?
-    fun defaultKeyPairTemplateFor(account: String) = "nf-keypair-$account-{{region}}"
     fun defaultKeyPairFor(account: String, region: String) = "nf-keypair-$account-$region"
```
asher: This should be removed now.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/LaunchConfigurationResolver.kt
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+  private fun ClusterSpec.LaunchConfigurationSpec.withResolvedKeyPair(account: String): ClusterSpec.LaunchConfigurationSpec {
+    return if (keyPair == null) {
+      val defaultKeyPair = cloudDriverCache.defaultKeyPairForAccount(account)
+      if (!defaultKeyPair.contains("{{region}}")) {
+        copy(keyPair = defaultKeyPair)
+      } else {
+        // TODO: if it has a placeholder, then instead of setting the default we should set the key pair in the overrides
+        //  for each region with the placeholder replaced, so long as the user has not already specified something.
```
asher: I think our external cluster spec should support `{{region}}` expansion so that keypair overrides aren't always necessary when this regional keypair scheme is followed in an account.

luispollo: I thought about that as well, but assumed that setting the corresponding regional overrides in the resolved spec was the (one?) way to achieve that. Should I do that in the `ClusterSpec` `resolve*` methods instead?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -56,8 +56,7 @@ private fun ClusterSpec.resolveLaunchConfiguration(region: SubnetAwareRegionSpec
       ?: defaults.launchConfiguration?.iamRole
       ?: LaunchConfiguration.defaultIamRoleFor(moniker.app)),
     keyPair = checkNotNull(overrides[region.name]?.launchConfiguration?.keyPair
-      ?: defaults.launchConfiguration?.keyPair
-      ?: LaunchConfiguration.defaultKeyPairFor(locations.account, region.name)),
+      ?: defaults.launchConfiguration?.keyPair),
```
luispollo: This does not assume the hard-coded default key pair format now, and instead lets `LaunchConfigurationResolver` figure that out based on information reported by clouddriver.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -56,8 +56,7 @@ private fun ClusterSpec.resolveLaunchConfiguration(region: SubnetAwareRegionSpec
       ?: defaults.launchConfiguration?.iamRole
       ?: LaunchConfiguration.defaultIamRoleFor(moniker.app)),
     keyPair = checkNotNull(overrides[region.name]?.launchConfiguration?.keyPair
-      ?: defaults.launchConfiguration?.keyPair
-      ?: LaunchConfiguration.defaultKeyPairFor(locations.account, region.name)),
```
luispollo: This does not assume the hard-coded default key pair format now, and instead lets `LaunchConfigurationResolver` figure that out based on information reported by clouddriver.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/LaunchConfiguration.kt
luispollo: Done.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/LaunchConfigurationResolver.kt
asher: I think what's here now looks good!

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/LaunchConfigurationResolver.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.stereotype.Component
+
+@Component
+class LaunchConfigurationResolver(private val cloudDriverCache: CloudDriverCache) : Resolver<ClusterSpec> {
+  override val apiVersion: ApiVersion = SPINNAKER_EC2_API_V1
+  override val supportedKind: String = "cluster"
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  override fun invoke(resource: Resource<ClusterSpec>): Resource<ClusterSpec> =
+    resource.run {
```
asher: We should return early and short circuit calling `ClusterSpec.withResolvedKeyPairs` if a keypair is already specified in the submitted spec.

luispollo: The challenge here is that there are multiple places where key pairs can appear... Do you suggest checking all places (`defaults` and/or `overrides` for all regions?).

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -51,14 +51,16 @@ private fun ClusterSpec.resolveLaunchConfiguration(region: SubnetAwareRegionSpec
     instanceType = checkNotNull(overrides[region.name]?.launchConfiguration?.instanceType
       ?: defaults.launchConfiguration?.instanceType),
     ebsOptimized = checkNotNull(overrides[region.name]?.launchConfiguration?.ebsOptimized
-      ?: defaults.launchConfiguration?.ebsOptimized),
+      ?: defaults.launchConfiguration?.ebsOptimized
+      ?: LaunchConfiguration.DEFAULT_EBS_OPTIMIZED),
```
luispollo: This fixes an unrelated issue I caught when attempting to use the diff API on an exported spec. Since we're nulling out fields with defaults now, a spec with a missing `ebsOptimized` field is a valid one.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -51,15 +51,16 @@ private fun ClusterSpec.resolveLaunchConfiguration(region: SubnetAwareRegionSpec
...
     iamRole = checkNotNull(overrides[region.name]?.launchConfiguration?.iamRole
       ?: defaults.launchConfiguration?.iamRole
       ?: LaunchConfiguration.defaultIamRoleFor(moniker.app)),
```
luispollo: Side-question: we're still using the hard-coded, Netflix-specific IAM role template as a default when nothing is specified in the spec. Is there such a thing as "default IAM role name" for an account in clouddriver?

luispollo: Follow-up: it doesn't look like it. There's actually a template configured for the default IAM role in the clouddriver config file, but it's not exposed anywhere in the API that I could find.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/LaunchConfigurationResolver.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+  private fun ClusterSpec.withResolvedKeyPairs(account: String): ClusterSpec {
+    val defaultKeyPair = cloudDriverCache.defaultKeyPairForAccount(account)
+    assert(defaultKeyPair != null) { "Default key pair missing for account $account in clouddriver config." }
+
+    var defaultLaunchConfig: ClusterSpec.LaunchConfigurationSpec? = null
+    val overrideLaunchConfigs = mutableMapOf<String, ClusterSpec.LaunchConfigurationSpec>()
+
+    if (defaultKeyPair.contains(REGION_PLACEHOLDER)) {
```
gal-yardeni: Can you rename `REGION_PLACEHOLDER` to `REGION`? I found it a bit confusing..

luispollo: Hmm, but that's what it is, a placeholder for the region...

gal-yardeni: Not sure I follow - we can take it offline :)

luispollo: The definition of that constant is:
```kotlin
    private const val REGION_PLACEHOLDER = "{{region}}"
```

Issue #629, opened 2019-11-27
We should not check a resource every minute if it's unhealthy


emjburns:
When a resource is unhappy (meaning we keep taking action and it keeps failing to rectify a diff) we should slow our roll. We should back off and instead check only once ever X minutes (10? 30? configurable) so that if something changes we will catch it, but we're not constantly submitting tasks.

PR #630, opened 2019-11-27
feat(veto): slow down checking for diffs when unhappy

emjburns:
Closes https://github.com/spinnaker/keel/issues/629 (which I definitely just opened but still!).

This pr introduces an "unhappy" veto that will let us not check an unhappy resource for a configurable amount of time. Some small changes to support that:
- support for sending different events based on which veto said no
- vetos accept either the whole resource, or id + application, because we can't reliably determine application from just id (like for images)
- changes around tests to support ^
- a new resource status: `vetoed` (not used yet)

The status of a resource that is vetoed by the "application veto" (pause management) will be `paused`, while the status of a resource that has been vetoed by the "unhappy veto" will remain `unhappy` to continue to convey that to the user.

_edit_

Updated this PR to 
- introduce basic diff fingerprinting
- move vetoing to after we calculate the diff of a resource

Diff fingerprinting means that we immediately check a resource again if the diff has changed so that we can react quickly to someone making updates to a resource.

Moving vetoing to after the diff has been calculated means that we now always make the calls to clouddriver to calculate a diff _even_ if the resource is paused or will be vetoed. Essentially, vetoing stops the *action* of resolving the diff, not checking the resource in general. This is necessary for the unhappy veto to recheck as soon as the diff changes (which is really a much nicer experience).

### DIFF
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -8,8 +8,9 @@ import com.netflix.spinnaker.keel.diff.ResourceDiff
 import com.netflix.spinnaker.keel.events.ResourceActuationLaunched
 import com.netflix.spinnaker.keel.events.ResourceActuationPaused
 import com.netflix.spinnaker.keel.events.ResourceActuationResumed
-import com.netflix.spinnaker.keel.events.ResourceCheckUnresolvable
+import com.netflix.spinnaker.keel.events.ResourceActuationVetoed
 import com.netflix.spinnaker.keel.events.ResourceCheckError
+import com.netflix.spinnaker.keel.events.ResourceCheckUnresolvable
 import com.netflix.spinnaker.keel.events.ResourceDeltaDetected
 import com.netflix.spinnaker.keel.events.ResourceDeltaResolved
 import com.netflix.spinnaker.keel.events.ResourceMissing
@@ -22,6 +23,7 @@ import com.netflix.spinnaker.keel.plugin.ResourceHandler
 import com.netflix.spinnaker.keel.plugin.supporting
 import com.netflix.spinnaker.keel.telemetry.ResourceCheckSkipped
 import com.netflix.spinnaker.keel.veto.VetoEnforcer
+import com.netflix.spinnaker.keel.veto.VetoResponse
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import org.slf4j.LoggerFactory
@@ -41,18 +43,11 @@ class ResourceActuator(
 
   suspend fun <T : ResourceSpec> checkResource(resource: Resource<T>) {
     val id = resource.id
-    val response = vetoEnforcer.canCheck(id)
+    val response = vetoEnforcer.canCheck(resource)
     if (!response.allowed) {
       log.debug("Skipping actuation for resource {} because it was vetoed: {}", id, response.message)
       publisher.publishEvent(ResourceCheckSkipped(resource.apiVersion, resource.kind, id))
-      publisher.publishEvent(
-        ResourceActuationPaused(
-          resource.apiVersion,
-          resource.kind,
-          id.value,
-          id.application,
-          response.message,
-          clock.instant()))
+      publishVetoedEvent(response, resource)
       return
     }
 
@@ -72,8 +67,8 @@ class ResourceActuator(
         ResourceActuationResumed(
           resource.apiVersion,
           resource.kind,
-          id.value,
-          id.application,
+          resource.id.value,
+          resource.spec.application,
           clock.instant()))
     }
 
@@ -140,6 +135,33 @@ class ResourceActuator(
         }
       }
       desired.await() to current.await()
+    }
+
+  /**
+   * We want a specific status for specific types of vetos. This function publishes the
+   * right event based on which veto said no.
+   */
+  private fun publishVetoedEvent(response: VetoResponse, resource: Resource<*>) =
+    when {
+      response.vetoName == "ApplicationVeto" -> publisher.publishEvent(
+        ResourceActuationPaused(
+          resource.apiVersion,
+          resource.kind,
+          resource.id.value,
+          resource.spec.application,
+          response.message,
+          clock.instant()))
+      response.vetoName == "UnhappyVeto" -> {
+        // don't publish an event, we want the status to stay as "unhappy" for clarity
+      }
+      else -> publisher.publishEvent(
+        ResourceActuationVetoed(
+          resource.apiVersion,
+          resource.kind,
+          resource.id.value,
+          resource.spec.application,
+          response.message,
+          clock.instant()))
     }
 
   // These extensions get round the fact tht we don't know the spec type of the resource from
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -47,21 +47,25 @@ class ResourceActuator(
     val id = resource.id
     val plugin = handlers.supporting(resource.apiVersion, resource.kind)
 
+    // todo eb: extract "application veto" as its own concept, and check before generating the diff.
+
     if (plugin.actuationInProgress(resource)) {
       log.debug("Actuation for resource {} is already running, skipping checks", id)
-      publisher.publishEvent(ResourceCheckSkipped(resource.apiVersion, resource.kind, id))
+      publisher.publishEvent(ResourceCheckSkipped(resource.apiVersion, resource.kind, id, "ActuationInProgress"))
       return
     }
 
     try {
       val (desired, current) = plugin.resolve(resource)
       val diff = ResourceDiff(desired, current)
-      diffFingerprintRepository.store(id, diff)
+      if (diff.hasChanges()) {
+        diffFingerprintRepository.store(id, diff)
+      }
 
       val response = vetoEnforcer.canCheck(resource)
       if (!response.allowed) {
         log.debug("Skipping actuation for resource {} because it was vetoed: {}", id, response.message)
-        publisher.publishEvent(ResourceCheckSkipped(resource.apiVersion, resource.kind, id))
+        publisher.publishEvent(ResourceCheckSkipped(resource.apiVersion, resource.kind, id, response.vetoName))
         publishVetoedEvent(response, resource)
         return
       }
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -16,6 +16,7 @@ import com.netflix.spinnaker.keel.events.ResourceDeltaResolved
 import com.netflix.spinnaker.keel.events.ResourceMissing
 import com.netflix.spinnaker.keel.events.ResourceValid
 import com.netflix.spinnaker.keel.events.Task
+import com.netflix.spinnaker.keel.persistence.DiffFingerprintRepository
 import com.netflix.spinnaker.keel.persistence.ResourceRepository
 import com.netflix.spinnaker.keel.plugin.CannotResolveCurrentState
 import com.netflix.spinnaker.keel.plugin.CannotResolveDesiredState
@@ -34,6 +35,7 @@ import java.time.Clock
 @Component
 class ResourceActuator(
   private val resourceRepository: ResourceRepository,
+  private val diffFingerprintRepository: DiffFingerprintRepository,
   private val handlers: List<ResourceHandler<*, *>>,
   private val vetoEnforcer: VetoEnforcer,
   private val publisher: ApplicationEventPublisher,
@@ -43,14 +45,6 @@ class ResourceActuator(
 
   suspend fun <T : ResourceSpec> checkResource(resource: Resource<T>) {
     val id = resource.id
-    val response = vetoEnforcer.canCheck(resource)
-    if (!response.allowed) {
-      log.debug("Skipping actuation for resource {} because it was vetoed: {}", id, response.message)
-      publisher.publishEvent(ResourceCheckSkipped(resource.apiVersion, resource.kind, id))
-      publishVetoedEvent(response, resource)
-      return
-    }
-
     val plugin = handlers.supporting(resource.apiVersion, resource.kind)
 
     if (plugin.actuationInProgress(resource)) {
@@ -59,22 +53,31 @@ class ResourceActuator(
       return
     }
 
-    log.debug("Checking resource {}", id)
-
-    if (resourceRepository.lastEvent(id) is ResourceActuationPaused) {
-      log.info("Actuation for resource {} resuming", id)
-      publisher.publishEvent(
-        ResourceActuationResumed(
-          resource.apiVersion,
-          resource.kind,
-          resource.id.value,
-          resource.spec.application,
-          clock.instant()))
-    }
-
     try {
       val (desired, current) = plugin.resolve(resource)
       val diff = ResourceDiff(desired, current)
+      diffFingerprintRepository.store(id, diff)
+
+      val response = vetoEnforcer.canCheck(resource)
+      if (!response.allowed) {
+        log.debug("Skipping actuation for resource {} because it was vetoed: {}", id, response.message)
+        publisher.publishEvent(ResourceCheckSkipped(resource.apiVersion, resource.kind, id))
+        publishVetoedEvent(response, resource)
+        return
+      }
+
+      log.debug("Checking resource {}", id)
+
+      if (resourceRepository.lastEvent(id) is ResourceActuationPaused) {
+        log.info("Actuation for resource {} resuming", id)
+        publisher.publishEvent(
+          ResourceActuationResumed(
+            resource.apiVersion,
+            resource.kind,
+            resource.id.value,
+            resource.spec.application,
+            clock.instant()))
+      }
 
       when {
         current == null -> {
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
@@ -2,23 +2,16 @@ package com.netflix.spinnaker.keel.telemetry
 
 import com.netflix.spinnaker.keel.api.ApiVersion
 import com.netflix.spinnaker.keel.api.ArtifactType
-import com.netflix.spinnaker.keel.api.Resource
 import com.netflix.spinnaker.keel.api.ResourceId
-import com.netflix.spinnaker.keel.api.id
 
 sealed class TelemetryEvent
 
 data class ResourceCheckSkipped(
   val apiVersion: ApiVersion,
   val kind: String,
-  val id: ResourceId
-) : TelemetryEvent() {
-  constructor(resource: Resource<*>) : this(
-    resource.apiVersion,
-    resource.kind,
-    resource.id
-  )
-}
+  val id: ResourceId,
+  val skipper: String = "unknown"
+) : TelemetryEvent()
 
 data class ArtifactVersionUpdated(
   val name: String,
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepositoryTests.kt
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.time.MutableClock
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
 import io.mockk.mockk
+import strikt.api.expect
 import strikt.api.expectThat
 import strikt.assertions.containsExactlyInAnyOrder
 import strikt.assertions.hasSize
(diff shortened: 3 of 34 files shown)


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -142,6 +137,33 @@ class ResourceActuator(
...
+          resource.kind,
+          resource.id.value,
+          resource.spec.application,
+          response.message,
+          clock.instant()))
+      response.vetoName == "UnhappyVeto" -> {
+        // don't publish an event, we want the status to stay as "unhappy" for clarity
+      }
+      else -> publisher.publishEvent(
+        ResourceActuationVetoed(
```
emjburns: This isn't used yet (we only have two vetos), but will certainly be used in the future. Laying the plumbing now.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -172,6 +174,10 @@ interface ResourceRepository : PeriodicallyCheckedRepository<Resource<out Resour
     return first() is ResourceActuationPaused
   }
 
+  private fun List<ResourceEvent>.isVetoed(): Boolean {
```
emjburns: FYI @erikmunson this will never be hit right now (so no resource will have a status of `vetoed`). This is for when we add other vetos.

keel-veto/src/main/kotlin/com/netflix/spinnaker/keel/veto/Veto.kt
```diff
@@ -57,9 +59,21 @@ interface Veto {
    * What's currently being vetoed
    */
   fun currentRejections(): List<String>
+
+  /**
+   * What's currently being vetoed for an app
+   */
+  fun currentRejectionsByApp(application: String): List<ResourceId>
```
emjburns: I added this method to support ui views in the future. It seemed important that we could easily understand what vetos are vetoing which resources within an app.

keel-veto/src/test/kotlin/com/netflix/spinnaker/keel/veto/ApplicationVetoTests.kt
```diff
@@ -115,53 +83,5 @@ class ApplicationVetoTests : JUnit5Minutests {
           .hasSize(0)
       }
     }
-
```
emjburns: These weren't needed because I made sure we always pass the app name into the veto.

keel-veto/src/main/kotlin/com/netflix/spinnaker/keel/veto/VetoController.kt
```diff
@@ -66,6 +66,15 @@ class VetoController(
     return veto.currentRejections()
   }
 
+  @GetMapping(
```
emjburns: a new endpoint to get the resource ids that are currently being vetoed for each application.

gal-yardeni:
@emjburns question about the name "unhappy veto" - does it mean that the resource that was previously managed (and then vetoed) is really "unhappy" (having an error or something wrong with it)? or is it just a way to convey to the user that his resources are not being managed?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceId.kt
```diff
@@ -25,7 +25,12 @@ import com.netflix.spinnaker.keel.serialization.ResourceIdDeserializer
...
-  // Resource names (the last token in a ResourceId) follow the Moniker format (<app>-<stack>-<detail>)
-  val application: String
-    get() = value.split(":").last().split("-").first()
+  // Resource names (the last token in a ResourceId) follow the Moniker format (<app>-<stack>-<detail>),
+  // unless it's a resource which doesn't support this, then we return null (like an image)
+  val application: String?
+    get() = if (value.startsWith("bakery:")) {
+      null
+    } else {
+      value.split(":").last().split("-").first()
```
gal-yardeni: Is it possible that the resource Id will not be in this format (<app>-<stack>-<detail>) or doesn't starts with "bakery:"? If so, can we get an exception here?

asher:
I think flapping vetoes should be tied to an artifact version for image and cluster resources instead of a resourceId. Vetoing by resourceId alone seems like a fit for load balancers and firewall rules though.

emjburns:
@asher great point. Ok if I make a todo and an issue for that, and split it into a separate PR? Or, do you think that's foundational enough to address here?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceId.kt
emjburns: I think the hard thing is there's no way to know if it doesn't fit the form except when you're implementing the resource. Right now, it's either it fits or it's the image. In the future, this could be different. I know this isn't the best way to do things, but I can't think of another way that keeps the convenience of this method (except removing the method and defining it only in resources that have a last piece of the id that's frigga named). Got any ideas?

asher:
@emjburns re: tying a FlappyBirdVeto to an artifact version, I think it's worth some initial thought just from the standpoint of the db schema modeled in this PR. Doesn't actually have to happen now but there's an opportunity to dodge future liquibase refactoring which is always nice.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -64,22 +53,31 @@ class ResourceActuator(
...
-          id.application,
-          clock.instant()))
-    }
-
     try {
       val (desired, current) = plugin.resolve(resource)
       val diff = ResourceDiff(desired, current)
+      diffFingerprintRepository.store(id, diff)
+
+      val response = vetoEnforcer.canCheck(resource)
```
emjburns: Moving veto checking to after the diff

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -64,22 +53,31 @@ class ResourceActuator(
...
-          resource.kind,
-          id.value,
-          id.application,
-          clock.instant()))
-    }
-
     try {
       val (desired, current) = plugin.resolve(resource)
       val diff = ResourceDiff(desired, current)
+      diffFingerprintRepository.store(id, diff)
```
emjburns: Note: if there's no diff, this stores a hash of "{}". Is that desired behavior? Or, if there's no diff, we could just clear the record of the diff...

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DiffFingerprintRepository.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import com.netflix.spinnaker.keel.api.ResourceId
+import com.netflix.spinnaker.keel.diff.ResourceDiff
+import java.security.MessageDigest
+import javax.xml.bind.DatatypeConverter
+
+/**
+ * Stores a hash of the diff
+ */
+interface DiffFingerprintRepository {
```
emjburns: Using this repository to see how many times we've seen the exact same diff.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -41,21 +45,6 @@ class ResourceActuator(
 
   suspend fun <T : ResourceSpec> checkResource(resource: Resource<T>) {
     val id = resource.id
-    val response = vetoEnforcer.canCheck(id)
```
asher: This covered application level vetoes, right? I think that particular veto type should be special cased and still eval'd here, so it can short circuit diff generation.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DiffFingerprintRepository.kt
asher: Is there a reason this isn't a part of `ResourceRepository`? It is entirely `resourceId` based.

keel-sql/src/main/resources/db/changelog/20191202-create-diff-fingerprint-table.yml
```diff
@@ -0,0 +1,48 @@
...
+      author: emjburns
+      changes:
+        - addPrimaryKey:
+            tableName: diff_fingerprint
+            constraintName: diff_fingerprint_pk
+            columnNames: resource_id
+      rollback:
+        - dropPrimaryKey:
+            constraintName: diff_fingerprint_pk
+            tableName: resource_id
```
asher: tableName should be `diff_fingerprint`

keel-sql/src/main/resources/db/changelog/20191202-create-diff-fingerprint-table.yml
```diff
@@ -0,0 +1,48 @@
...
+                    nullable: false
+              - column:
+                  name: count
+                  type: int
+                  defaultValue: 1
+                  constraints:
+                    nullable: false
+              - column:
+                  name: hash
+                  type: longtext
```
asher: This can be fixed length, based on the hash algo used. For sha1, I think `char(40)` is appropriate.

keel-veto/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
```diff
@@ -0,0 +1,95 @@
...
+  private val diffFingerprintRepository: DiffFingerprintRepository,
+  private val unhappyVetoRepository: UnhappyVetoRepository
+) : Veto {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  @Value("veto.unhappy.waiting-time")
+  private var waitingTime: String = "PT10M"
+
+  private val MAX_DIFF_COUNT = 10
```
asher: 10 seems like a lot for things like large prod server groups. I can imagine situations wherein we'd appreciate if this and `waitingTime` were dynamic properties.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: Do you think it even belongs in a veto? I'm happy making it another type of thing.
Another option is we could have two places where we evaluate vetos, and have each veto choose where it's evaluated...

keel-veto/src/main/kotlin/com/netflix/spinnaker/keel/veto/Veto.kt
asher: I can see this rolled into the `ApplicationController` response as-is. An initial first-pass in deck (cc: @erikmunson) could just be changing the color of :managed: badge when vetoed (no idea if that's low effort or even viable, just brainstorming.)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDiffFingerprintRepository.kt
```diff
@@ -0,0 +1,76 @@
...
+    val hash = diff.generateHash()
+    jooq
+      .select(DIFF_FINGERPRINT.COUNT, DIFF_FINGERPRINT.FIRST_DETECTION_TIME)
+      .from(DIFF_FINGERPRINT)
+      .where(DIFF_FINGERPRINT.RESOURCE_ID.eq(resourceId.toString()))
+      .and(DIFF_FINGERPRINT.HASH.eq(hash))
+      .forUpdate()
+      .fetchOne()
+      ?.let { (count, firstDetectionTime) ->
+        jooq.insertInto(DIFF_FINGERPRINT)
```
asher: This only runs if the `select for update` returns a row matching resourceId and our current fingerprint, so this should just be an `update` clause.

asher: I also think a transaction is required for the `select for update` to work as intended. So long as we're confident that only one keel instance will eval a resource at a time, it shouldn't matter if we do a locking read though.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DiffFingerprintRepository.kt
emjburns: Discussed offline - for now, we're going to keep this separate because it's tracking info about the resource vs core resource things.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: Undesired behavior. Changing so that if there's no diff, we don't store anything (but the old record remains). I can't think of any problems with this right now since flapping = unhappy status + same diff more than 10 times in a row

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: Adding a todo to fix this - I don't want to make this PR any bigger.

keel-veto/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
emjburns: Good point. I'll make them optionally dynamic.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDiffFingerprintRepository.kt
emjburns: Fixed (hopefully)

keel-veto/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
emjburns: Ha also I meant for this to be 5.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
robfletcher: At the same time it might make sense to be able to provide users the context that there is a diff in their resource even though we're not going to resolve it.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
```diff
@@ -2,23 +2,16 @@ package com.netflix.spinnaker.keel.telemetry
...
-  val id: ResourceId
-) : TelemetryEvent() {
-  constructor(resource: Resource<*>) : this(
-    resource.apiVersion,
-    resource.kind,
-    resource.id
-  )
-}
+  val id: ResourceId,
+  val skipper: String = "unknown"
```
robfletcher: What's this name all about?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceId.kt
robfletcher: It looks to me like you removed the only place this was used `ApplicationVeto`. The two remaining uses are in `ResourceActuator` where it has access to the full resource & therefore could just use `resource.spec.application` instead.

I guess what I'm saying is, I think we can remove this property.

keel-sql/src/main/resources/db/changelog/20191202-create-diff-fingerprint-table.yml
```diff
@@ -0,0 +1,48 @@
...
+            tableName: diff_fingerprint
+            columns:
+              - column:
+                  name: resource_id
+                  type: varchar(255)
+                  constraints:
+                    nullable: false
+              - column:
+                  name: first_detection_time
+                  type: bigint
```
robfletcher: Is that the right data type for an `Instant`?

emjburns: It's a long (`clock.instant().toEpochMilli()`)

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
emjburns: Who skipped the check. I added this so that we can use this metric to track the unhappy resources because they'll all have been skipped by the same thing.

keel-sql/src/main/resources/db/changelog/20191202-create-diff-fingerprint-table.yml
robfletcher: Why not just use a timestamp, then you don't have to use `toEpochMilli`? Same as e.g. `resource_event` table.

emjburns: Because @asher said that all the other services store time as the long value. Personally I find using `toEpochMilli()` is so much easier than reasoning about time and timestamps and timestamp format.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: At this moment I don't think that's that useful, especially because we don't have a way to surface that anywhere

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceId.kt
emjburns: Good call, removed and put a note on it to avoid parsing application from it (because I have a feeling that'll come up again...)

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -41,45 +45,43 @@ class ResourceActuator(
...
-          id.value,
-          id.application,
-          response.message,
-          clock.instant()))
-      return
-    }
-
     val plugin = handlers.supporting(resource.apiVersion, resource.kind)
 
+    // todo eb: extract "application veto" as its own concept, and check before generating the diff.
```
asher: 👍

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepositoryTests.kt
```diff
@@ -0,0 +1,134 @@
...
+      test("9 minutes later we should still skip") {
+        clock.incrementBy(Duration.ofMinutes(9))
+        val vetoStatus = subject.getVetoStatus(resourceId)
+        expect {
+          that(vetoStatus.shouldSkip).isEqualTo(true)
+          that(vetoStatus.shouldRecheck).isEqualTo(false)
+        }
+      }
+
+      test("11 minutes later don't skip, instead recheck") {
```
asher: These test are great!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDiffFingerprintRepository.kt
```diff
@@ -0,0 +1,77 @@
...
+  private val jooq: DSLContext,
+  private val clock: Clock
+) : DiffFingerprintRepository {
+  override fun store(resourceId: ResourceId, diff: ResourceDiff<*>) {
+    val hash = diff.generateHash()
+    jooq
+      .select(DIFF_FINGERPRINT.COUNT, DIFF_FINGERPRINT.FIRST_DETECTION_TIME, DIFF_FINGERPRINT.HASH)
+      .from(DIFF_FINGERPRINT)
+      .where(DIFF_FINGERPRINT.RESOURCE_ID.eq(resourceId.toString()))
+      .forUpdate()
```
asher: Should either make the select + update a transaction, or just remove the `forUpdate` (no point in acquiring a lock on the row if it's immediately released)

keel-sql/src/main/resources/db/changelog/20191202-create-diff-fingerprint-table.yml
asher: I would ❤️ if we converted all timestamp persistence to epoch milli's so we don't have to consider TZ's except where we want to and maintain millisecond precision (I seem to remember that mysql timestamp's are second precision only but mot 💯 on that.)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDiffFingerprintRepository.kt
emjburns: I'll remove the `forUpdate`.

keel-sql/src/main/resources/db/changelog/20191202-create-diff-fingerprint-table.yml
robfletcher: Hmm, yeah, I was forgetting the hoops we have to jump thru to convert a tz-agnostic `Instant` into a `LocalDateTime` to get it into the table. Pretty sure we wouldn't have that issue with Postgres… grumble, grumble.

robfletcher: I guess my main / only issue with using epoch millis is that it's harder to interpret raw SQL query results.

PR #632, opened 2019-11-27
feat(constraints): crud support for state

asher:
- Includes all current constraint states in the `ApplicationController` method polled by deck, when the application has a delivery-config and details are expanded.
- REST methods to read constraint history and update status.
- Persistence refactor, including adding a `current_constraint` table to serve the `ApplicationController` access pattern.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
@@ -17,6 +17,7 @@
  */
 package com.netflix.spinnaker.keel.rest
 
+import com.netflix.spinnaker.keel.persistence.DeliveryConfigRepository
 import com.netflix.spinnaker.keel.persistence.ResourceRepository
 import org.slf4j.LoggerFactory
 import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
@@ -29,7 +30,8 @@ import org.springframework.web.bind.annotation.RestController
 @RestController
 @RequestMapping(path = ["/application"])
 class ApplicationController(
-  private val resourceRepository: ResourceRepository
+  private val resourceRepository: ResourceRepository,
+  private val deliveryConfigRepository: DeliveryConfigRepository
 ) {
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
@@ -41,14 +43,25 @@ class ApplicationController(
     @PathVariable("application") application: String,
     @RequestParam("includeDetails", required = false, defaultValue = "false") includeDetails: Boolean
   ): Map<String, Any> {
+    val hasDeliveryConfig = deliveryConfigRepository.hasDeliveryConfig(application)
+
     if (includeDetails) {
       val resources = resourceRepository.getSummaryByApplication(application)
+      val constraintStates = if (hasDeliveryConfig) {
+        deliveryConfigRepository.constraintStateFor(application)
+      } else {
+        emptyList()
+      }
 
       return mapOf(
         "hasManagedResources" to resources.isNotEmpty(),
-        "resources" to resources
+        "resources" to resources,
+        "hasDeliveryConfig" to hasDeliveryConfig,
+        "currentEnvironmentConstraints" to constraintStates
       )
     }
-    return mapOf("hasManagedResources" to resourceRepository.hasManagedResources(application))
+    return mapOf(
+      "hasManagedResources" to resourceRepository.hasManagedResources(application),
+      "hasDeliveryConfig" to hasDeliveryConfig)
   }
 }
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
@@ -1,20 +1,30 @@
 package com.netflix.spinnaker.keel.rest
 
 import com.netflix.spinnaker.keel.actuation.ResourcePersister
+import com.netflix.spinnaker.keel.api.ConstraintState
+import com.netflix.spinnaker.keel.api.ConstraintStatus
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.SubmittedDeliveryConfig
 import com.netflix.spinnaker.keel.diff.AdHocDiffer
 import com.netflix.spinnaker.keel.diff.EnvironmentDiff
+import com.netflix.spinnaker.keel.exceptions.InvalidConstraintException
 import com.netflix.spinnaker.keel.persistence.DeliveryConfigRepository
+import com.netflix.spinnaker.keel.persistence.ResourceRepository.Companion.DEFAULT_MAX_EVENTS
 import com.netflix.spinnaker.keel.yaml.APPLICATION_YAML_VALUE
 import org.slf4j.LoggerFactory
+import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
+import org.springframework.web.bind.annotation.ExceptionHandler
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestHeader
 import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
+import java.time.Instant
 
 @RestController
 @RequestMapping(path = ["/delivery-configs"])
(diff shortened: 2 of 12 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -45,5 +55,51 @@ class DeliveryConfigController(
...
+  //  In this case, we should ensure the user has an appropriate role based on a service account
+  //  provided to keel. Service account may need to move to the environment or delivery config level.
+  fun updateConstraintStatus(
+    @RequestHeader("X-SPINNAKER-USER") user: String,
+    @PathVariable("name") name: String,
+    @PathVariable("environment") environment: String,
+    @PathVariable("type") type: String,
+    @RequestParam("artifactVersion") artifactVersion: String,
+    @RequestParam("status") status: String,
+    @RequestParam("comment") comment: String?
```
asher: @erikmunson I'm not sure if `/delivery-configs/{name}/environment/{environment}/constraint-state/{type}` plus the required request params makes for the best interface, let me know if you have any feedback on the new routes.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -45,5 +55,51 @@ class DeliveryConfigController(
...
+    @RequestParam("limit") limit: Int?
+  ): List<ConstraintState> =
+    deliveryConfigRepository.constraintStateFor(name, environment, limit ?: DEFAULT_MAX_EVENTS)
+
+  @PostMapping(
+    path = ["/{name}/environment/{environment}/constraint-state/{type}"],
+    consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
+    produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  // TODO: Keel needs fiat integration and should use kork's AuthenticatedRequest class.
```
emjburns: This integration already exists and is used in the `ResourceController`: ([here](https://github.com/spinnaker/keel/blob/master/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt#L79) is an example of it in action)

emjburns: Unless you're talking about something else? I agree that service account probably needs to be able to be pulled to a higher level

asher: I missed that, thanks! I'm going to leave a TODO here until we move service account to a higher level, but will update its text.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
```diff
@@ -7,25 +7,91 @@ import com.netflix.spinnaker.keel.api.ResourceId
...
+   * related to a application retrieved by its name.
+   *
+   * @param application the application name
+   *
+   * @return A list of the most recent [ConstraintState]'s by environment per type or an
+   * empty list if none exist.
+   */
+  fun constraintStateFor(application: String): List<ConstraintState>
+
+  /**
```
gal-yardeni: Thanks for adding documentations!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
gal-yardeni: I thought of suggesting to have a separate controller for the constraints I guess it doesn't make much sense since it relies on the delivery config name. Maybe just `constraint` instead of `constraint-state`?

asher: +1 to changing to route to `constraint`, will do. If we end up with an `EnvironmentController`, I think constraints will belong there, fwiw.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -43,12 +45,22 @@ class ApplicationController(
...
+        deliveryConfigRepository.constraintStateFor(application)
+      } else {
+        emptyList()
+      }
 
       return mapOf(
         "hasManagedResources" to resources.isNotEmpty(),
-        "resources" to resources
+        "resources" to resources,
+        "hasDeliveryConfig" to hasDeliveryConfig,
```
luispollo: Just for context, what's the idea behind the `hasXyz` properties in the output here? For `hasManagedResources` at least, it seems to be redundant with `resources`?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -43,12 +45,22 @@ class ApplicationController(
...
+        "currentEnvironmentConstraints" to constraintStates
       )
     }
-    return mapOf("hasManagedResources" to resourceRepository.hasManagedResources(application))
+    return mapOf(
+      "hasManagedResources" to resourceRepository.hasManagedResources(application),
+      "hasDeliveryConfig" to deliveryConfigRepository.hasDeliveryConfig(application))
```
luispollo: Same as lines 48 and (I'm guessing) 56? Should we move the the corresponding `val` declarations to before the `if` statement to avoid repetition?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -45,5 +55,50 @@ class DeliveryConfigController(
   fun diff(@RequestBody deliveryConfig: SubmittedDeliveryConfig): List<EnvironmentDiff> =
     adHocDiffer.calculate(deliveryConfig)
 
+  @GetMapping(
+    path = ["/{name}/environment/{environment}/constraint"],
```
luispollo: Since this returns a list of items, I think the path should be `constraints`.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
```diff
@@ -257,8 +266,63 @@ internal class DeliveryConfigControllerTests : JUnit5Minutests {
                 ?.constraints
             ).isNotNull()
           }
+
+          context("no environments with stateful constraints have been evaluated") {
+            test("no constraint state") {
+              getProdConstraintStates(contentType)
+                .andExpect(status().isOk)
+                .andExpect(content().string(containsString("[]")))
```
luispollo: Should we parsing the content into a map and checking the specific fields instead of comparing strings? I'm genuinely asking to understand what pattern we're using (if any).

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -45,5 +55,50 @@ class DeliveryConfigController(
...
+  ) {
+    val currentState = deliveryConfigRepository.getConstraintState(name, environment, artifactVersion, type)
+      ?: throw InvalidConstraintException("$name/$environment/$type/$artifactVersion", "constraint not found")
+
+    deliveryConfigRepository.storeConstraintState(
+      currentState.copy(
+        status = ConstraintStatus.valueOf(status),
+        comment = comment ?: currentState.comment,
+        judgedAt = Instant.now(),
+        judgedBy = user
```
luispollo: I think I missed this before, but isn't this specific to the `manual-judgement` constraint type?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
luispollo: ➕

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
```diff
@@ -93,3 +93,73 @@ databaseChangeLog:
...
+        - addPrimaryKey:
+            tableName: current_constraint
+            constraintName: current_constraint_pk
+            columnNames: application, environment_uid, type
+        - createIndex:
+            tableName: current_constraint
+            indexName: current_constraint_uid_idx
+            columns:
+              - column:
+                  name: constraint_uid
```
luispollo: Shouldn't these changes go into a new YAML file?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
asher: No, these are intended to apply to all stateful constraints.

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
asher: Why?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
asher: Good call, updated

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
luispollo: Hmm... Can you please give me an example of judgement on a stateful constraint besides manual judgement? Maybe it's just the naming of these properties that is throwing me off.

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
luispollo: I'm not familiar with the framework we're using for DB migrations, but I was just commenting based on prior experience with other such tools where a "golden rule" of migrations was that, once committed into source control, you never changed the original migration script since different developers might have different versions of the code checked out. It's not a big deal in this case since you're just adding to the file, but my question was, again, more around the patterns we're following for my own education.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
asher: I think this is really a `deck` question for @erikmunson but let's discuss outside of the context of this PR.

keel-sql/src/main/resources/db/changelog/20191108-environment-artifact-constraint.yml
asher: We should consider each `changeSet` within a migration file to be immutable (liquibase tracks the md5 hash of individual changeSets, not files) but don't have rules established beyond that. It would be nice to collapse the schema for public consumption when there's lots of refactoring between beginning and releasing a feature but we historically haven't done that solely due to it complicating the release process within netflix. When modifying a table that is already in use and defined by one or more existing liquibase files, I tend to create a new file but I think there's room for judgement there too.

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
asher: Updated to selectively deserialize depending on the contentType. In this case, I wanted an assertion that would apply for yaml and json response bodies without being concerned for which was which but if you're looking for patterns to follow, that wasn't a good one.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
asher: Canary was the example discussed in the original PR.

asher: 1) Constraint evaluators always produce a boolean judgement. 
2) Automated judgements can always be overridden by an operator.
3) Automated constraint types shouldn't necessarily be tied to an implementation. A `canary` constraint type that is judged by "Kayenta" today may have been judged by "ACA" yesterday without requiring changes to a service owner's delivery config.

luispollo: Makes sense. I think for me it was just the term overlap with `manual-judgement` that made it sound specific for me (as opposed to "approval" or "promotion" as used in other constraint-related interfaces). Maybe some room for renaming, but I can't come up with a better name at the moment.

PR #639, opened 2019-12-04
Removed references to Mahe

### LINKED ISSUE #628: Mahe should not be referenced by public Keel project
Mahe (Netflix's implementation of dynamic properties) should not be directly referenced in the public Keel repository as it's not available to OSS users. There are also assumptions being made about how base AMIs work in `keel-bakery`. We should pull and interface out of `BaseImageCache` and move the concrete implementation to Netflix's private repo.

robfletcher:
Mahe is not a public Spinnaker service. We are currently using it to fetch the specific AMIs for base images in `keel-bakery-plugin`. This is not going to work for anyone outside of Netflix.

I discussed this with @asher and it should be possible to add detail to the response from Rosco's bake options endpoint that include the specific AMIs for each available base image. That would be a sensible default implementation for open source.

We could probably add the same to the Netflix bakery's equivalent endpoint and remove the Mahe-based implementation altogether, but for now I'm moving it to keel-nflx.

Fixes #628


robfletcher:
Note, as it stands it's no longer possible to run the Keel application with the bakery plugin enabled unless you use `keel-nflx` as there's no public implementation of `BaseImageCache`. Thinking about that now. Might just add a configuration based implementation.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/DefaultBaseImageCache.kt
```diff
@@ -0,0 +1,16 @@
...
+  private val baseImages: Map<String, Map<String, String>>
+) : BaseImageCache {
+  override fun getBaseImage(os: String, label: BaseLabel) =
+    baseImages[os]?.get(label.name.toLowerCase()) ?: throw UnknownBaseImage(os, label)
+}
+
+@ConfigurationProperties("bakery.base-images")
+class BaseImageCacheProperties {
+  var baseImages: Map<String, Map<String, String>> = emptyMap()
+}
```
robfletcher: Default implementation of `BaseImageCache` requires you to set the images up in configuration. This is what Rosco does as well so it should be adequate for now.

PR #649, opened 2019-12-07
fix(actuation) Fix stack overflow in ResourceActuator

luispollo:
Fixes #610.

The root cause for this error was twofold:
1. An `UnknownBaseImage` exception was being thrown here since there were no base images in the config file and the `@ConfigurationProperties` responsible from loading that data into the config properties class had one path too many in the prefix:
  https://github.com/spinnaker/keel/blob/78b9e61498f3a27078cf97f53badab517c92568c/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/DefaultBaseImageCache.kt#L10

2. The `resolve` extension function defined in `ResourceActuator` was using a [`coroutineScope`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/coroutine-scope.html) to run the current and desired state evaluations in parallel, which, according to the documentation, has the behavior that "when any child coroutine in this scope fails, this scope fails and all the rest of the children are cancelled". This is why the HTTP requests to clouddriver were being stopped mid-way (socket closed exception) -- when the `UnknownBaseImage` exception was thrown in one child coroutine, the other was forcefully cancelled.
   https://github.com/spinnaker/keel/blob/18d60daf007182401da91dceb08298163b53b999/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt#L126-L145

I've replaced `coroutineScope` with [`supervisorScope`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/supervisor-scope.html), where "a failure of a child does not cause this scope to fail and does not affect its other children".

After fixing the coroutine scope issue, I now get this in the logs:
```
[...]

PR #650, opened 2019-12-07
fix(baking): Fix slow bakes due to missing data in rocket cache

luispollo:
The root cause for the delay of ~5 minutes between detecting an image diff and being able to retrieve the artifact details from igor was that the rocket cache consumed by igor takes a while to update. It turns out that we don't need any more detail to start a bake task than what we already had in the resource diff and the spec, so long as we use a simple naming convention for full package names, so this fixes that issue by removing the call to igor.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
@@ -30,6 +30,7 @@ import com.netflix.spinnaker.keel.persistence.NoSuchArtifactException
 import com.netflix.spinnaker.keel.plugin.Resolver
 import com.netflix.spinnaker.keel.plugin.ResourceHandler
 import com.netflix.spinnaker.keel.plugin.SupportedKind
+import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import kotlinx.coroutines.runBlocking
 import org.springframework.context.ApplicationEventPublisher
 
@@ -104,7 +105,16 @@ class ImageHandler(
     val appVersion = AppVersion.parseName(resourceDiff.desired.appVersion)
     val packageName = appVersion.packageName
     val version = resourceDiff.desired.appVersion.substringAfter("$packageName-")
-    val artifact = igorService.getArtifact(packageName, version)
+    val artifact = Artifact.builder()
+      .type("DEB")
+      .customKind(false)
+      .name(resource.spec.artifactName)
+      .version(version)
+      .location("rocket")
+      .reference("/${packageName}_${version}_all.deb")
+      .metadata(mapOf())
+      .provenance("n/a")
+      .build()
 
     log.info("baking new image for {}", resource.spec.artifactName)
     val description = "Bake ${resourceDiff.desired.appVersion}"
--- a/keel-bakery-plugin/src/test/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandlerTests.kt
+++ b/keel-bakery-plugin/src/test/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandlerTests.kt
@@ -22,7 +22,6 @@ import com.netflix.spinnaker.keel.orca.TaskRefResponse
 import com.netflix.spinnaker.keel.persistence.memory.InMemoryArtifactRepository
 import com.netflix.spinnaker.keel.serialization.configuredObjectMapper
 import com.netflix.spinnaker.keel.test.resource
-import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
 import io.mockk.coEvery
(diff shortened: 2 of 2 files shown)


keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -104,7 +105,16 @@ class ImageHandler(
...
     val packageName = appVersion.packageName
     val version = resourceDiff.desired.appVersion.substringAfter("$packageName-")
-    val artifact = igorService.getArtifact(packageName, version)
+    val artifact = Artifact.builder()
+      .type("DEB")
+      .customKind(false)
+      .name(resource.spec.artifactName)
+      .version(version)
+      .location("rocket")
+      .reference("/${packageName}_${version}_all.deb")
```
erikmunson: Is this .deb file naming pattern something that’s conventional enough we can rely on/expect non-Netflix builds to do it? I ask because I have literally no idea how the Debian ecosystem works beyond what I know from Netflix.

luispollo: That’s a good question. I asked a similar one in chat today. The short answer is “I don’t know”, but I got feedback that the `_all` suffix in the package name is standard at Netflix. Also, I hard-coded `rocket` there for the `location` field because, in my understanding, that’s the only artifact source we’re supporting today, which is very Netflix-specific. I’d love for @emjburns to chime in though.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -124,7 +124,7 @@ class ResourceActuator(
   }
 
   private suspend fun ResourceHandler<*, *>.resolve(resource: Resource<out ResourceSpec>): Pair<Any, Any?> =
-    coroutineScope {
+    supervisorScope {
```
robfletcher: What does this do?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/DefaultBaseImageCache.kt
```diff
@@ -10,7 +10,7 @@ class DefaultBaseImageCache(
     baseImages[os]?.get(label.name.toLowerCase()) ?: throw UnknownBaseImage(os, label)
 }
 
-@ConfigurationProperties("keel.plugins.bakery.base-images")
+@ConfigurationProperties(prefix = "keel.plugins.bakery")
```
robfletcher: D'oh. Thanks for the fix. However, we shouldn't actually be using this inside of Netflix.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: Did you read the PR description? 😊

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/DefaultBaseImageCache.kt
luispollo: Hmm, ok. Let’s talk offline. We’re definitely calling methods that rely on this config.

luispollo: P.S. This is from #649 too.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
emjburns: The debian stuff we're doing right now is very netflix specific. That's a fine assumption for now I think. I wouldn't worry at this moment about generalizing any debian logic - we don't know of anyone else who wants to use it.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/DefaultBaseImageCache.kt
emjburns: If you run oss keel, you'll need that. If you run -nflx keel, we will get the base amis.

keel-bakery-plugin/src/test/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandlerTests.kt
```diff
@@ -281,7 +261,7 @@ internal class ImageHandlerTests : JUnit5Minutests {
         }
 
         expectThat(request.captured.job.first())
-          .hasEntry("package", "keel_0.160.0-h62.02c0fbf_all.deb")
+          .hasEntry("package", "keel_0.161.0-h63.24d0843_all.deb")
```
emjburns: Why did the version change here?

luispollo: Because I removed `before` block with the mock for `igorService.getArtifact` above, since `upsert` is no longer calling it. That mock had an inconsistency between the version passed in and the package name set in the artifact, which is why this test was passing in the first place. 😃

emjburns: great! Thanks!

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: Rebased.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/DefaultBaseImageCache.kt
luispollo: Rebased.

luispollo: Rebased.

luispollo:
@emjburns Are you OK with this change overall?

Issue #637, opened 2019-12-03
Pipeline constraint


asher:
A constraint that executes a pipeline and only passes if the pipeline completes successfully.

PR #651, opened 2019-12-07
feat(constraints): pipeline execution constraint

asher:
An environment promotion constraint to gate promotions on the successful execution
 of a Spinnaker pipeline. The target pipeline is launched via orca the first time
 a pipeline constraint is evaluated for a given artifact version and [Environment].

 The state of that pipeline execution is then assessed on each subsequent evaluation
 and can optionally be retried on failure.

 Example env constraint:
```
 constraints:
   - type: pipeline
     serviceAccount: keel@spinnaker (OPTIONAL, defaults to DEFAULT_SERVICE_ACCOUNT)
     pipelineId: (the id visible when editing a pipeline, not its name)
     retries: (OPTIONAL, the number of times to retry a failure, defaults to 0)
     parameters: (OPTIONAL, Map of trigger parameters to pass along, `user` and `type` are reserved)
```

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
@@ -17,7 +17,8 @@ import java.time.zone.ZoneRulesException
 @JsonSubTypes(
   Type(value = DependsOnConstraint::class, name = "depends-on"),
   Type(value = TimeWindowConstraint::class, name = "allowed-times"),
-  Type(value = ManualJudgementConstraint::class, name = "manual-judgement")
+  Type(value = ManualJudgementConstraint::class, name = "manual-judgement"),
+  Type(value = PipelineConstraint::class, name = "pipeline")
 )
 abstract class Constraint(val type: String)
 
@@ -59,6 +60,14 @@ data class TimeWindowConstraint(
 data class ManualJudgementConstraint(
   val timeout: Duration = Duration.ofDays(7)
 ) : StatefulConstraint("manual-judgement")
+
+data class PipelineConstraint(
+  val serviceAccount: String = "anonymous",
+  val timeout: Duration = Duration.ofHours(2),
+  val pipelineId: String,
+  val retries: Int = 0,
+  val parameters: Map<String, Any?> = emptyMap()
+) : StatefulConstraint("pipeline")
 
 data class TimeWindow(
   val days: String? = null,
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
@@ -1,5 +1,8 @@
 package com.netflix.spinnaker.keel.api
 
+import com.fasterxml.jackson.annotation.JsonSubTypes
+import com.fasterxml.jackson.annotation.JsonSubTypes.Type
+import com.fasterxml.jackson.annotation.JsonTypeInfo
 import java.time.Duration
 import java.time.Instant
 
@@ -13,10 +16,9 @@ data class ConstraintState(
   val judgedBy: String? = null,
   val judgedAt: Instant? = null,
   val comment: String? = null,
-  // TODO: reconsider type when implementing a constraint that uses attributes
-  val attributes: Map<String, Any?> = emptyMap()
+  val attributes: ConstraintStateAttributes? = null
 ) {
-  fun canPromote() = status.passes()
+  fun passed() = status.passes()
 
   fun failed() = status.failed()
 
@@ -34,3 +36,19 @@ enum class ConstraintStatus(private val passed: Boolean, private val failed: Boo
   fun passes() = passed
   fun failed() = failed
 }
+
+@JsonTypeInfo(
+  include = JsonTypeInfo.As.EXISTING_PROPERTY,
+  use = JsonTypeInfo.Id.NAME,
+  property = "type")
+@JsonSubTypes(
+  Type(value = PipelineConstraintStateAttributes::class, name = "pipeline")
+)
+abstract class ConstraintStateAttributes(val type: String)
+
+data class PipelineConstraintStateAttributes(
+  val executionId: String? = null,
+  val attempt: Int,
+  val latestAttempt: Instant,
+  val lastExecutionStatus: String? = null
+) : ConstraintStateAttributes("pipeline")
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
@@ -1,9 +1,13 @@
 package com.netflix.spinnaker.keel.constraints
 
 import com.netflix.spinnaker.keel.api.Constraint
+import com.netflix.spinnaker.keel.api.ConstraintState
+import com.netflix.spinnaker.keel.api.ConstraintStatus
 import com.netflix.spinnaker.keel.api.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.constraints.ConstraintEvaluator.Companion.getConstraintForEnvironment
+import com.netflix.spinnaker.keel.persistence.DeliveryConfigRepository
 
 interface ConstraintEvaluator<T : Constraint> {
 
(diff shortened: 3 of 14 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
```diff
@@ -35,3 +39,49 @@ interface ConstraintEvaluator<T : Constraint> {
     targetEnvironment: Environment
   ): Boolean
 }
+
+abstract class StatefulConstraintEvaluator<T : Constraint> : ConstraintEvaluator<T> {
```
asher: @luispollo I think this is more in line with what you suggested earlier when I was toying with overloading the `ConstraintEvaluator` interface with stateful bits?

keel-orca/src/test/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluatorTests.kt
```diff
@@ -0,0 +1,171 @@
...
+    val trigger = slot<Map<String, Any?>>()
+    val subject = PipelineConstraintEvaluator(orcaService, deliveryConfigRepository, clock)
+  }
+
+  fun tests() = rootContext<Fixture> {
+    fixture {
+      Fixture(
+        constraint = PipelineConstraint(
+          pipelineId = randomUID().toString(),
+          parameters = mapOf("youSayManaged" to "weSayDelivery")
```
luispollo: 😄❤️

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
```diff
@@ -13,10 +16,9 @@ data class ConstraintState(
   val judgedBy: String? = null,
   val judgedAt: Instant? = null,
   val comment: String? = null,
-  // TODO: reconsider type when implementing a constraint that uses attributes
-  val attributes: Map<String, Any?> = emptyMap()
+  val attributes: ConstraintStateAttributes? = null
 ) {
-  fun canPromote() = status.passes()
+  fun passed() = status.passes()
```
robfletcher: ❤️ Better

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaService.kt
```diff
@@ -45,7 +57,7 @@ data class TaskRefResponse(
   val taskId by lazy { ref.substringAfterLast("/") }
 }
 
-data class TaskDetailResponse(
+data class ExecutionDetailResponse(
```
robfletcher: I like this name better too

keel-orca/src/test/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluatorTests.kt
```diff
@@ -0,0 +1,171 @@
...
+        expectThat(state).isNotNull()
+
+        val attributes = state!!.attributes as PipelineConstraintStateAttributes
+        with(attributes) {
+          expectThat(executionId).isEqualTo(fixture.executionId)
+          expectThat(attempt).isEqualTo(1)
+          expectThat(latestAttempt).isLessThanOrEqualTo(clock.instant())
+          expectThat(lastExecutionStatus).isNull()
+        }
+      }
```
robfletcher: I'd love to get your thoughts on how Strikt could help make this more expressive. You _can_ do it with out the `!!` but it's kinda syntactically obscure.

```kotlin
val state = deliveryConfigRepository.getConstraintState(manifest.name, environment.name, version, type)
expectThat(state)
  .isNotNull() // now the subject is known to be non-null
  .isA<PipelineConstraintStateAttributes>() // now the subject is known to be an instance of this
  .and { 
    get { executionId }.isEqualTo(fixture.executionId)
    get { attempt }.isEqualTo(1)
    get { latestAttempt }.isLessThanOrEqualTo(clock.instant())
    get { lastExecutionStatus }.isNull()
  }
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
```diff
@@ -60,6 +61,14 @@ data class ManualJudgementConstraint(
   val timeout: Duration = Duration.ofDays(7)
 ) : StatefulConstraint("manual-judgement")
 
+data class PipelineConstraint(
+  val serviceAccount: String = DEFAULT_SERVICE_ACCOUNT,
```
emjburns: I think this needs to come from the delivery config. The default service account has all the permissions, and therefore this would be a huge escalation of privilege. As a first step, can we just get any resource in the environment and take the service account from that resource instead of using the default one?

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluator.kt
```diff
@@ -0,0 +1,221 @@
...
+
+    if (shouldTrigger(constraint, attributes)) {
+      try {
+        val executionId = runBlocking {
+          startPipeline(targetEnvironment, constraint)
+        }
+
+        attributes = PipelineConstraintStateAttributes(
+          executionId = executionId,
+          attempt = attributes?.attempt ?: 0 + 1,
```
emjburns: Is `0 + 1` doing something other than returning 1?

keel-orca/src/test/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluatorTests.kt
emjburns: <3

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluator.kt
luispollo: I interpreted this as `(attributes?.attempt ?: 0) + 1`, but maybe not?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
asher: I'd rather we move the declaration of a service account to the root level of the delivery config. I don't think it should be declared on resources; perhaps we shouldn't even allow resources to be submitted outside of a delivery config.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluator.kt
asher: This is supposed to mean what @luispollo interpreted it as but turns out it actually does need the parenthesis, will fix.

keel-orca/src/test/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluatorTests.kt
asher: This is great, thank you! I still have a lot to learn about strikt, will update around this pattern and keep it in mind for the future.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
emjburns: Let's have the "how much do we support resources separate from delivery configs" offline.

Can require specification of the service account and not have it use the default in the meantime? Or, have the pipeline run as anonymous as the default? I just don't want there to be such an easy way to get all the privileges for all the accounts.

asher: How about for this PR, I remove `serviceAccount` from the constraint definition, leaving pipelines to only be runnable under the keel `DEFAULT_SERVICE_ACCOUNT`, with a TODO to use the service account defined on the delivery config once that's a requirement?

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluator.kt
```diff
@@ -0,0 +1,221 @@
...
+          latestAttempt = clock.instant()
+        )
+      } catch (e: Exception) {
+        log.warn("Failed triggering pipeline ${constraint.pipelineId} for " +
+          "${deliveryConfig.application}:$deliveryConfig/$targetEnvironment")
+
+        attributes = PipelineConstraintStateAttributes(
+          executionId = null,
+          attempt = attributes?.attempt ?: 0 + 1,
+          latestAttempt = clock.instant()
```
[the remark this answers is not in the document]
asher: This sounds great but I don't think there's a reliable way to extract exceptions or error messages from an orca execution today, which is an ongoing struggle for the UI as well.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
```diff
@@ -35,3 +39,49 @@ interface ConstraintEvaluator<T : Constraint> {
...
+          deliveryConfigRepository.storeConstraintState(it)
+          // TODO: Emit an event here?
+        }
+
+    when {
+      state.failed() -> return false
+      state.passed() -> return true
+    }
+
+    return canPromote(artifact, version, deliveryConfig, targetEnvironment, constraint, state)
```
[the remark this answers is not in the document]
asher:
```
    return when {
      state.failed() -> false
      state.passed() -> true
      else -> canPromote(artifact, version, deliveryConfig, targetEnvironment, constraint, state)
    }
```
looks much better, thanks!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
asher: I assumed keel's `DEFAULT_SERVICE_ACCOUNT` was heavily write restricted, and that keel was only able to actuate in restricted accounts via user specified service accounts. Sound like that isn't the case which we may want to discuss further offline but I'll make this default to "anonymous" for now.

asher:
Closes #637

PR #645, opened 2019-12-05
feat(logging): Add X-SPINNAKER-RESOURCE-ID to logs for traceability

luispollo:
This is a proof-of-concept to instrument keel with a discrete field in the logs that can be used to aid in tracing the processing of resources and in troubleshooting. I recommend viewing the diff with the "ignore white space" option enabled.

As part of this exercise, I investigated/learned the following:
- Logging setup of Spinnaker services at Netflix
   -  Linux `init.d` scripts redirect console output emitted by the logger to a file
   - A SumoLogic collector process monitors the log file and ships logs to SumoLogic
   - We use logback as the SLF4J logging backend, configured to output log messages to the console in a JSON format that is easily understood/parsed by SumoLogic. This means our logs outside of development environments are already in a structured format. Here's the relevant SLF4J config:
     ```xml
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
       <encoder class="net.logstash.logback.encoder.LogstashEncoder">
         <provider class="net.logstash.logback.composite.loggingevent.ArgumentsJsonProvider"/>
       </encoder>
     </appender>
      ```
    - The fact that we're using a logstash encoder and provider class above is misleading, since we're no longer shipping logs to logstash. But, as I said, the same JSON format generated by those classes is understood by SumoLogic

- Behavior of SLF4J [MDC](http://www.slf4j.org/manual.html#mdc) (Mapped Diagnostic Context) support with the above configuration
  - It turns out any key/value pairs added to the MDC map are automatically added as discrete fields in the generated JSON log output. This means that the only thing we need to do to add more metadata to our logs is to add that data to the MDC context.
  - MDC implementations in the logging frameworks rely on thread-local objects and hence do not work out-of-the-box with Kotlin coroutines.
  - Official SLF4J MDC support for Kotlin coroutines is offered in this module: https://github.com/Kotlin/kotlinx.coroutines/tree/master/integration/kotlinx-coroutines-slf4j
  - The challenge is finding the right places to instrument with the provided coroutine context

With the above in place, and the changes in this PR, you get the following logging output when the resource check loop kicks in (note the presence of the `X-SPINNAKER-RESOURCE-ID` field in the JSON, and that the thread names are different but the MDC context is preserved):
```json
[...]

PR #652, opened 2019-12-09
feat(logging): Add X-SPINNAKER-RESOURCE-ID to logs for traceability

luispollo:
As a follow-up to #645, this adds an `X-SPINNAKER-RESOURCE-ID` discrete field to the keel logs when the output is in JSON format, which can be helpful in narrowing down logs when troubleshooting.

Instrumentation is currently limited to only the resource checking loop in `ResourceActuator` and the resource export flow initiated by `ExportController`, but the helper methods defined in `TracingSupport` should be reusable elsewhere.

Perhaps of particular note in this PR, the resource ID is not propagated to calls executed in thread pools, such as those from the `okhttp3` HTTP client. It should be technically possible to wire those threads with MDC, but I feel like that should be left for a future PR.

### DIFF
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -15,7 +15,7 @@ import com.netflix.spinnaker.keel.events.ResourceMissing
 import com.netflix.spinnaker.keel.events.ResourceValid
 import com.netflix.spinnaker.keel.events.Task
 import com.netflix.spinnaker.keel.pause.ResourcePauser
-import com.netflix.spinnaker.keel.logging.TracingSupport.Companion.withResourceTracingContext
+import com.netflix.spinnaker.keel.logging.TracingSupport.Companion.withTracingContext
 import com.netflix.spinnaker.keel.persistence.DiffFingerprintRepository
 import com.netflix.spinnaker.keel.persistence.ResourceRepository
 import com.netflix.spinnaker.keel.plugin.CannotResolveCurrentState
@@ -45,20 +45,20 @@ class ResourceActuator(
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
   suspend fun <T : ResourceSpec> checkResource(resource: Resource<T>) {
-    withResourceTracingContext(resource) {
+    withTracingContext(resource) {
       val id = resource.id
       val plugin = handlers.supporting(resource.apiVersion, resource.kind)
 
       if (resourcePauser.isPaused(resource)) {
         log.debug("Actuation for resource {} is paused, skipping checks", id)
         publisher.publishEvent(ResourceCheckSkipped(resource.apiVersion, resource.kind, id, "ActuationPaused"))
-        return@withResourceTracingContext
+        return@withTracingContext
       }
 
       if (plugin.actuationInProgress(resource)) {
         log.debug("Actuation for resource {} is already running, skipping checks", id)
         publisher.publishEvent(ResourceCheckSkipped(resource.apiVersion, resource.kind, id, "ActuationInProgress"))
-        return@withResourceTracingContext
+        return@withTracingContext
       }
 
       try {
@@ -73,7 +73,7 @@ class ResourceActuator(
           log.debug("Skipping actuation for resource {} because it was vetoed: {}", id, response.message)
           publisher.publishEvent(ResourceCheckSkipped(resource.apiVersion, resource.kind, id, response.vetoName))
           publishVetoedEvent(response, resource)
-          return@withResourceTracingContext
+          return@withTracingContext
         }
 
         log.debug("Checking resource {}", id)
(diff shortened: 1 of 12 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Exportable.kt
```diff
@@ -3,9 +3,13 @@ package com.netflix.spinnaker.keel.api
...
+  val cloudProvider: String,
   val account: String,
   val serviceAccount: String,
   val moniker: Moniker,
   val regions: Set<String>,
   val kind: String
-)
+) {
+  fun toResourceId() =
+    "$cloudProvider:$kind:$account:${moniker.name}".let(::ResourceId)
```
luispollo: Please help correct this format if it's wrong.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/logging/TracingSupport.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.slf4j.MDC
+import kotlinx.coroutines.slf4j.MDCContext
+import kotlinx.coroutines.withContext
+
+/**
+ * Support for tracing resources in log statements via MDC in coroutines.
+ */
+class TracingSupport {
+  companion object {
+    const val X_SPINNAKER_RESOURCE_ID = "X-SPINNAKER-RESOURCE-ID"
```
marchello2000: Not sure if `kork` is used in `keel`... if it is: https://github.com/spinnaker/kork/blob/master/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java#L66

luispollo: It is! Technically, this is not a header though, and not necessarily restricted to web request flows as exemplified by the scheduled resource checks flow instrumented in this PR. Would you still recommend it?

gal-yardeni:
Thanks for adding this, I think it's important to have a mechanism to see the tracing of a specific resource. One request - maybe the name withResourceTracingContext is too long/confusing? maybe something like logResource / traceResource etc.?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/logging/TracingSupport.kt
marchello2000: i guess if it's not a header or mdc probably leave as is.

asher: We probably do want to propagate it as a request header, keel can make plenty of downstream calls to other spinnaker µservices in the context of evaluating a resource.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -44,82 +45,83 @@ class ResourceActuator(
...
-      publisher.publishEvent(ResourceCheckSkipped(resource.apiVersion, resource.kind, id, "ActuationInProgress"))
-      return
-    }
-
-    try {
-      val (desired, current) = plugin.resolve(resource)
-      val diff = ResourceDiff(desired, current)
-      if (diff.hasChanges()) {
-        diffFingerprintRepository.store(id, diff)
+    withResourceTracingContext(resource) {
```
asher: I'm concerned that `MDC.remove(X_SPINNAKER_RESOURCE_ID)` doesn't appear in this PR. Other beans may run on a thread that formerly ran `ResourceActuator`, so it's good practice to clean up thread locals. A pattern that has become common in other services is to add to MDC via a try/finally block, with cleanup in the finally.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/logging/TracingSupport.kt
luispollo: Sure, we can look into that, but in a separate PR, if you don't mind. 😊

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: That's good feedback, @asher, thanks! Originally I was playing around with just adding `MDCContex()` to coroutine builders like `async` and `launch` in specific places, which had the behavior of propagating current MDC into the coroutine and restoring the original context outside that scope, but it seems like with `withContext` that's not the case. Let me take yet another look at the library to make sure I got this right (maybe even add some tests).

luispollo:
> Thanks for adding this, I think it's important to have a mechanism to see the tracing of a specific resource. One request - maybe the name withResourceTracingContext is too long/confusing? maybe something like logResource / traceResource etc.?

Thanks, @gal-yardeni. I dont like that it's such a long name, either, but I wanted to make sure it was apparent this function was creating another coroutine context to run the block with. Maybe `withTracingContext` as a compromise?

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: @asher you were spot on, thanks! I added a `try/catch` block to remove the tracing field as suggested. Please review.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/logging/TracingSupport.kt
asher: 👍

Issue #619, opened 2019-11-22
Move sorting info to artifact


emjburns:
docker artifacts need to contain information about how to sort tags so that we can get the "latest" artifact. Right now this info is w/in a cluster in the container specification. It shouldn't be.

PR #654, opened 2019-12-10
Artifact version generalization

emjburns:
Closes https://github.com/spinnaker/keel/issues/619.

This is a giant PR, and I'm sorry. I didn't have the foresight to make it into smaller PRs. 

Here are the changes:
- create a `DockerArtifact` and a `DebianArtifact`
- store version information in the artifact repository
- compare versions with the appropriate comparator whenever you ask for a list of versions
- flip the debian comparator so it's in decending order, which is what we want everywhere
- allow you to re-register artifacts so that you can change the sort order
- export titus resources with a container strategy
- sync versions of docker containers like we do for debs, so it's easy to run locally
- default docker status to null, and update version filtering so that we don't pass in the list of statuses when we don't want to filter by version (this bit us before, so fixing it now)

Here are the problems:
- re-registering artifacts means that two people could be battling over the same artifact. we need to also persist what delivery config (or cluster?) the artifact goes with

I'll mark the specific places where each of these happens. Thanks for taking the time to review this <3


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
```diff
@@ -132,10 +132,8 @@ class ResourcePersister(
     }
 
   private fun DeliveryArtifact.register() {
-    if (!artifactRepository.isRegistered(name, type)) {
```
emjburns: Always re-register to allow changes to docker versioning strategy.

keel-actuator/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionCheckerTests.kt
```diff
@@ -45,10 +44,7 @@ internal class EnvironmentPromotionCheckerTests : JUnit5Minutests {
       publisher
     )
 
-    val artifact = DeliveryArtifact(
-      name = "fnord",
-      type = DEB
-    )
+    val artifact = DebianArtifact(name = "fnord")
```
emjburns: I had to change every place we refer to `DeliveryArtifact`, so there are a lot of these types of changes. You can ignore all of them because there was no functionality change.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -77,7 +77,7 @@ class ArtifactController(
     @PathVariable name: String,
     @PathVariable type: ArtifactType
   ): List<String> =
-    artifactRepository.versions(DeliveryArtifact(name, type))
+    artifactRepository.versions(name, type)
```
emjburns: We need to be able to query by artifact name and type, because it's not always valid to construct the artifact. How do we know how it's sorted? In the repository, we query for the full artifact to get this information.

keel-api/src/test/resources/examples/cluster-example.yml
```diff
@@ -11,7 +11,7 @@ spec:
   imageProvider:
     deliveryArtifact:
       name: keeldemo
-      type: DEB
+      type: deb
```
emjburns: Look @erikmunson no more shouting, ARE YOU HAPPY!?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -30,25 +30,25 @@ class ArtifactListener(
...
           val version: String
-          val status: ArtifactStatus
+          var status: ArtifactStatus? = null
           when (artifact.type) {
             DEB -> {
               version = "${korkArtifact.name}-${korkArtifact.version}"
               status = debStatus(korkArtifact)
             }
             DOCKER -> {
               version = "${korkArtifact.name}:${korkArtifact.version}"
```
emjburns: It felt better to make status `null` for docker.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
```diff
@@ -206,7 +204,8 @@ internal class ArtifactListenerTests : JUnit5Minutests {
     .build()
 
   data class SyncArtifactsFixture(
-    val artifact: DeliveryArtifact,
```
emjburns: I added tests for syncing docker image versions as well here, because that's a big change in this PR.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -70,7 +70,7 @@ class ImageService(
   suspend fun getLatestNamedImage(packageName: String, account: String, region: String? = null): NamedImage? =
     cloudDriverService.namedImages(DEFAULT_SERVICE_ACCOUNT, packageName, account, region)
       .sortedWith(NamedImageComparator)
-      .lastOrNull {
```
emjburns: I changed the NamedImageComparator (and also the DebianComparator) to default to descending order. We always want the "latest", and it seemed easier to always have that first instead of finding the last of a list, or reversing the list. All the docker comparators are also sorted in descending order.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/NamedImage.kt
```diff
@@ -14,9 +14,9 @@ data class NamedImage(
 
 object NamedImageComparator : Comparator<NamedImage> {
   override fun compare(a: NamedImage, b: NamedImage): Int {
-    val byAppVersion = VERSION_COMPARATOR.compare(a.appVersion, b.appVersion)
+    val byAppVersion = DEBIAN_VERSION_COMPARATOR.compare(a.appVersion, b.appVersion)
     return if (byAppVersion == 0) {
-      a.creationDate.compareTo(b.creationDate)
+      b.creationDate.compareTo(a.creationDate)
```
emjburns: The swap to descending..dun dun dunnnn

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
```diff
@@ -17,3 +21,37 @@ enum class ArtifactType {
...
+  property = "type"
+)
+@JsonSubTypes(
+  Type(DebianArtifact::class),
+  Type(DockerArtifact::class)
+)
+interface DeliveryArtifact {
+  val name: String
+  val type: ArtifactType
+  val versioningStrategy: VersioningStrategy
```
emjburns: The common things for an artifact are: name, type, and how versions are sorted.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/VersioningStrategy.kt
```diff
@@ -0,0 +1,148 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
+import com.netflix.frigga.ami.AppVersion
+import com.netflix.rocket.semver.shaded.DebianVersionComparator
+import com.netflix.spinnaker.keel.api.SortType.INCREASING
+import com.netflix.spinnaker.keel.api.SortType.SEMVER
+import com.netflix.spinnaker.keel.serialization.PropertyNamePolymorphicDeserializer
+import net.swiftzer.semver.SemVer
+import org.slf4j.LoggerFactory
+import org.springframework.util.comparator.NullSafeComparator
+
```
emjburns: I moved all the docker sorting logic to be in proper comparators, so that it was easy to sort lists with them.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ArtifactEvent.kt
```diff
@@ -9,12 +8,8 @@ data class ArtifactEvent(
   val details: Map<String, Any>?
 )
 
-/**
- * This event optionally takes [statuses] so that we can fetch the most relevant first artifact.
- */
```
emjburns: This seemed more complicated than it was worth.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -1,18 +1,23 @@
...
 interface ArtifactRepository {
 
+  /**
+   * Creates or updates a registered artifact
+   */
+  // todo eb: allow the same artifact with different versioning strategies to be registered
+  // so that two users aren't battling
+  // maybe require an artifact to be associated with a delivery config?
+  // maybe have an artifact also have a name to reference?
   fun register(artifact: DeliveryArtifact)
```
emjburns: ^ This is the big registration of artifacts change. I think we need artifacts to have one more piece of information to tie them to a delivery config so that we can support two people caring about the same image but choosing different versioning strategies. That's a change for another PR, though.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -91,28 +101,6 @@ interface ArtifactRepository {
   )
 }
 
-val VERSION_COMPARATOR: Comparator<String> = object : Comparator<String> {
```
emjburns: This has been moved to where the other comparators are

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
```diff
@@ -6,25 +6,36 @@ import com.netflix.spinnaker.keel.api.DeliveryArtifact
...
 
 class InMemoryArtifactRepository : ArtifactRepository {
   private val artifacts = mutableMapOf<DeliveryArtifact, MutableList<ArtifactVersionAndStatus>>()
-  private val approvedVersions = mutableMapOf<Triple<DeliveryArtifact, DeliveryConfig, String>, MutableList<String>>()
-  private val deployedVersions = mutableMapOf<Triple<DeliveryArtifact, DeliveryConfig, String>, MutableList<String>>()
+  private val approvedVersions = mutableMapOf<Key, MutableList<String>>()
+  private val deployedVersions = mutableMapOf<Key, MutableList<String>>()
   private val log: Logger by lazy { LoggerFactory.getLogger(javaClass) }
 
+  private data class Key(
```
emjburns: I find `Triple` confusing to work with, so I created a `Key` class.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -64,26 +97,28 @@ class SqlArtifactRepository(
...
         .from(DELIVERY_ARTIFACT, DELIVERY_ARTIFACT_VERSION)
         .where(DELIVERY_ARTIFACT.UID.eq(DELIVERY_ARTIFACT_VERSION.DELIVERY_ARTIFACT_UID))
         .and(DELIVERY_ARTIFACT.NAME.eq(artifact.name))
         .and(DELIVERY_ARTIFACT.TYPE.eq(artifact.type.name))
-        .and(DELIVERY_ARTIFACT_VERSION.STATUS.`in`(*status.toTypedArray()))
+        .apply { if (statuses.isNotEmpty()) and(DELIVERY_ARTIFACT_VERSION.STATUS.`in`(*statuses.map { it.toString() }.toTypedArray())) }
         .fetch()
         .getValues(DELIVERY_ARTIFACT_VERSION.VERSION)
-        .sortAppVersion()
+        .sortedWith(artifact.versioningStrategy.comparator)
```
emjburns: ⭐️ 🕵️‍♀️  This is the actual functionality I was going for with this PR: an artifact has sorting defined on it, and that sorting is used to correctly return the versions, and the "latest" version.

keel-api/src/test/resources/examples/cluster-example.yml
erikmunson: THIS IS THE GREATEST DAY OF MY LIFE. I AM VERY HAPPY.

THANKS FOR ASKING.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
luispollo: Maybe only re-register if the version strategy has actually changed, then? Or at least only publish the event when that is the case?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -86,49 +82,65 @@ class ArtifactListener(
...
               }
               else -> false
             }
 
             if (hasNew) {
               log.debug("Artifact {} has a missing version {}, persisting..", artifact, latestVersion)
-              val version = artifactService.getArtifact(artifact.name, latestVersion)
-              artifactRepository.store(artifact, latestAppVersion, debStatus(version))
+              val status = when (artifact.type) {
+                DEB -> debStatus(artifactService.getArtifact(artifact.name, latestVersion.removePrefix("${artifact.name}-")))
```
luispollo: Any reason not to use the class types in the `when` statement like you did elsewhere?

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
emjburns: That seems like an unneeded optimization - what would it save us?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -86,49 +82,65 @@ class ArtifactListener(
...
+                // todo eb: is there a better way to think of docker status?
```
luispollo: I'm not sure about status for Docker artifacts specifically, but as a side-comment, when reviewing this I was a bit confused with the fact that we name that type `ArtifactStatus`(which is somewhat open to interpretation, I think). Looking at the values of the enum, perhaps we could instead name it something like `ReleaseStatus` instead?

Going back to the question on Docker images, do we know if there's any life-cycle management of image releases like there is for Debian packages (meaning, there might be tags on the images with that release status, for instance)?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: Kotlin can't detect that there are only two valid types of artifacts from using the class type (i think because `DeliveryArtifact` isn't sealed), so I'd have to create a third `else` condition and then handle that. Seems simpler to do it this way.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
luispollo: It would avoid polluting the event log with duplicate events? Or is this event marked with the automatic non-dupe thing?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -86,49 +82,65 @@ class ArtifactListener(
...
+              }
+              artifactRepository.store(artifact, latestVersion, status)
             }
           }
         }
       }
     }
 
+  private fun getLatestStoredVersion(artifact: DeliveryArtifact): String? =
+    artifactRepository.versions(artifact).sortedWith(artifact.versioningStrategy.comparator).firstOrNull()
```
luispollo: This is neat! 🤘

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -140,7 +140,7 @@ class ImageService(
       .filter {
         AppVersion.parseName(it.appVersion).packageName == packageName
       }
-      .lastOrNull { namedImage ->
+      .firstOrNull { namedImage ->
```
luispollo: I hope you caught all of them. 😉

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
```diff
@@ -17,3 +21,37 @@ enum class ArtifactType {
...
+}
+
+@JsonTypeName("deb")
+data class DebianArtifact(
+  override val name: String,
+  val statuses: List<ArtifactStatus> = emptyList(),
```
luispollo: Side-comment: this got me thinking that these classes don't really represent specific artifacts, per se, but rather a specification for how to _find_ artifacts matching certain attributes. I don't know that it would be worth renaming them, but I find myself confused at times thinking these actually represent specific "snapshot-in-time" versions of an artifact.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/VersioningStrategy.kt
```diff
@@ -0,0 +1,148 @@
...
+  companion object {
+    private val log by lazy { LoggerFactory.getLogger(this::class.java) }
+
+    fun parseWithRegex(input: String, strategy: TagVersionStrategy, customRegex: String?): String? {
+      val regex = Regex(customRegex ?: strategy.regex)
+      val result = regex.find(input) ?: return null
+      return when (result.groupValues.size) {
+        2 -> result.groupValues[1]
+        1 -> {
+          log.warn("Regex (${customRegex ?: strategy.regex}) produced zero capture groups on tag $input")
```
luispollo: Could it be that there was just no capturing groups in the custom regex and so group 0 (entire match) should be returned here?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/VersioningStrategy.kt
```diff
@@ -0,0 +1,148 @@
...
+          null
+        }
+        else -> {
+          log.warn("Regex (${customRegex
+            ?: strategy.regex}) produced more than one capture group on tag $input, using the first group")
```
luispollo: This seems like a guess. Throw an exception for unsupported regex format instead?

keel-sql/src/main/kotlin/com/netflix/spinnaker/exceptions/DataParsingException.kt
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.exceptions
+
+import com.netflix.spinnaker.keel.api.ArtifactType
+
+class DataParsingException(
```
luispollo: Since this is specific to parsing artifacts, maybe `ArtifactParsingException`?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -24,21 +26,52 @@ import java.time.Instant
...
+      .onDuplicateKeyUpdate()
+      .set(DELIVERY_ARTIFACT.DETAILS, artifact.toJson())
       .execute()
-      .also { count ->
-        if (count == 0) log.warn("Duplicate artifact registered: {}", artifact)
-      }
   }
 
-  override fun store(artifact: DeliveryArtifact, version: String, status: ArtifactStatus): Boolean {
+  private fun DeliveryArtifact.toJson() =
```
luispollo: Maybe `detailsAsJson()` since this only returns the details, not all properties of the artifact?

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
```diff
@@ -139,7 +135,7 @@ internal class DeliveryConfigControllerTests : JUnit5Minutests {
         |application: keel
         |artifacts:
         |- name: keel
-        |  type: DEB
+        |  type: deb
```
robfletcher: Is this gonna break anyone's config?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -66,17 +66,13 @@ class ArtifactListener(
   @EventListener(ArtifactRegisteredEvent::class)
   fun onArtifactRegisteredEvent(event: ArtifactRegisteredEvent) {
     val artifact = event.artifact
-    if (artifactRepository.isRegistered(artifact.name, artifact.type)) {
-      log.debug("Artifact {} is already registered", artifact)
-    } else {
-      log.debug("Registering artifact {}", artifact)
-      artifactRepository.register(artifact)
-    }
+    artifactRepository.register(artifact)
```
robfletcher: Is this call just a no-op if the artifact is registered already?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -162,10 +171,9 @@ class ArtifactListener(
     return ArtifactStatus.valueOf(status)
   }
 
-  private val artifactTypeNames by lazy { ArtifactType.values().map(ArtifactType::name) }
+  private fun Artifact.type() = ArtifactType.valueOf(type.toUpperCase())
```
robfletcher: Why a `fun` rather than a `val`?

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
robfletcher: Kotlin adds a `reversed()` function to `Comparator` so you could always do ```.sortedWith(NamedImageComparator.reversed())```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
robfletcher: Yeah, that's better

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.netflix.spinnaker.keel.api.ArtifactType.DOCKER
+import com.netflix.spinnaker.keel.api.DebianArtifact
+import com.netflix.spinnaker.keel.api.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.DockerArtifact
+import com.netflix.spinnaker.keel.serialization.configuredObjectMapper
+
+/**
+ * A helper class to construct the proper artifact type from the serialized json.
+ */
+object ArtifactUtils {
```
robfletcher: You don't need to embed these functions in an object. Functions can be first-class package members in Kotlin.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
emjburns: It's not recorded in the event log

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
emjburns: Oh, ya maybe probably. I'll hand update their configs.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: It's an update call if the artifact is registered so that we can handle changes in docker sorting. If this gets sql load intensive we can optimize this.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: Release status would definitely be better, and we can change that in another PR.

There isn't any built-in lifecycle management with docker images that I know of. I think at netflix we might add some sort of indication, but I don't want to support that now or worry about that now.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: `artifactTypeNames` just moved down to line 176. This is an additional function.

robfletcher: Yeah, but `type` seems weird as a function to me, rather than a property

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
emjburns: It was really hard for me to keep track of that mentally - especially with apis that return descending order by default (like the rocket versions api). I kept getting confused - when do we have to reverse? What's the default sort order here? So I thought this might be a little bit easier.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
emjburns: I think I did! I did a big search :D :D

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
robfletcher: Got it.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
emjburns: Oh, yes, you're definitely right about these drifting to represent all the information about how we find an artifact. Given that they're a "delivery artifact" i think it's okay to have that slightly different definition (not a snapshot-in-time artifact). Do you think a class level comment would help explain this?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/VersioningStrategy.kt
emjburns: I don't think so. For example, what if the capture group for docker is meant to capture the job number out of a tag like this `master-h5.5a52206` (regex: `^master-h(\d+).*$` will capture `5` from `h5`). Also present in the list of tags is `latest`. That regex will not produce a capture group on `latest`, so we want to exclude it from the sort.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/VersioningStrategy.kt
emjburns: It's definitely a guess. That seems like a better idea.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
emjburns: Oh nifty I forgot about that!

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: Oh, the artifact in kork already has a property `type`, so this function is to parse the `ArtifactType` from that string value.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
luispollo: I think it would!

keel-api/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
luispollo: Actually we made these case-insensitive for backwards compatibility.

Issue #633, opened 2019-12-02
Support for kork-plugins


robzienert:
I took a brief dive through Keel to figure out what would be needed to set it up for plugins. From the quick perusal, there's a lot that needs to be done for MVP. I'm writing this issue up to first set context of restrictions, and then have outlined what I've identified so far as needing to change to support plugins for Vetos and Constraints. I haven't gotten far enough to evaluate what it'd take for Resources, but I assume it wouldn't be far off.

This is meant as a collection of thoughts. I'd love to get a conversation going and land on something most are happy-ish with.

# Context

Today, `kork-plugins` only supports local plugins, which are loaded into a child `ClassLoader`. The implication of this is that dependencies that are shared between the parent `ClassLoader` and the plugin must be the same (down to the exact version), lest runtime errors occur when trying to invoke objects that share an interface or class, but versions differ. 

The implication of this is that Spinnaker plugin contracts must have as few dependencies as possible to limit the chance of plugin breakages between service version changes.

This means that the plugin API module must: 1) Be defined in Java, and 2) have little to no dependencies outside of `kork-plugins-api`. By reducing the dependencies the plugins API, there will be less opportunity for plugins to fail between Keel version changes: Imagine Keel changing the Jackson version it uses - all plugins would immediately fail if they were using Jackson from a prior Keel version.

I explicitly pick on defining in Java because Kotlin is itself a dependency; one that we would be foisting upon any plugin developer. Plugin developers may write their plugins in Kotlin, so existing plugins (e.g. EC2, Titus) in Keel need not be refactored to Java.

The work I'm proposing is what I'd recommend as minimally viable, to introduce plugins with as little impact to existing code today, while not sacrificing a good plugin experience. Since we don't know exactly what a "good plugin experience" is yet, it means we need to be judicious to the point of strict in what we allow to leak into the plugin contract.

# Work

- [x] Rename `keel-api` to `keel-web`. This is to fall inline with other services in the ecosystem, where plugin contracts are defined in `{service}-api`. We need less organizational fragmentation across our services, but I'm open to naming `keel-web` to something else if a compelling argument can be made for changing the other service modules similarly.
- [ ] Refactor the classes and interfaces defined within `keel-core` (`com.netflix.spinnaker.keel.api`) to Java and move to `keel-api` module. Jackson cannot come along for the ride, so `keel-core` may need to have some glue code written to shim. See `appendix1`
[...]

PR #664, opened 2019-12-11
refactor(web): Rename keel-api module to keel-web

robzienert:
Small change to make the Keel project more closely adhere to the rest of the Spinnaker project organizationally. Other services use `{service}-api` module to define plugin extension points.

See https://github.com/spinnaker/keel/issues/633


keel-core-test/README.md
```diff
@@ -1,3 +1,3 @@
-# keel-api-tck
+# keel-core-tck
```
robfletcher: ha, that's not even right. Apparently I did call this `keel-api` at some stage

PR #666, opened 2019-12-12
wip(api): Refactoring core models for plugins into java

robzienert:
This is far from being complete, but I wanted to put _something_ out there. I'm targeting `Resolver` as the scope for this initial change.

Things that need to be done:

- [ ] Documentation of all public interfaces and methods.
- [ ] Removal of Kotlin classes that still exist in `keel-core`.
- [ ] Kotlin DSL methods for compatibility where `val` interfaces were changed to methods.
- [ ] Shimming of Jackson usage in `keel-core`, since Jackson can't come for the ride in `keel-api`
- [ ] ... Probably other stuff, too.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,7 +30,7 @@ allprojects {
 }
 
 subprojects {
-  if (name != "keel-bom") {
+  if (name != "keel-bom" && name != "keel-api") {
     apply(plugin = "nebula.kotlin")
 
     repositories {
--- a/keel-actuator/keel-actuator.gradle.kts
+++ b/keel-actuator/keel-actuator.gradle.kts
@@ -4,6 +4,7 @@ plugins {
 }
 
 dependencies {
+  api(project(":keel-api"))
   api(project(":keel-core"))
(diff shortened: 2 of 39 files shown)


build.gradle.kts
```diff
@@ -30,7 +30,7 @@ allprojects {
 }
 
 subprojects {
-  if (name != "keel-bom") {
+  if (name != "keel-bom" && name != "keel-api") {
```
robfletcher: Why do you want to exclude this from the BOM?

robzienert: That's not what this is doing. This is opting the `keel-api` module out of including kotlin and all of these dependencies from it.

emjburns:
@robzienert can I close this? @robfletcher is working on refactoring `keel-api`, and we discussed that converting to java is not strictly necessary.

robfletcher:
I'm still potentially going to try to convert things to Java. It's just not phase 1.

emjburns:
Yes, that's why I want to close this. It won't be helpful after the refactoring you're doing I think.

Issue #643, opened 2019-12-05
APIs to support Environments UI


erikmunson:
Keel knows all kinds of things about resources, environments, and artifacts through polling and events. In order to create an [Environments UI](https://gallery.io/projects/MCHbtQVoQ2HCZcM1NWkzJjWA/files/MCEJu8Y2hyDScR6DpT6FAz4yBgK8GqbAOJA), though, we need to derive from that distributed set of information a human-friendly summary that can answer questions like:
- What environments do I have, and what resources live inside them?
- What artifact version is each environment/resource running?
- What new artifact versions are flowing through my environments?
- What new artifact versions are going to flow through my environments later?
- What constraints, if any, are in a pending state or need my attention? (e.g. 'there's a smoke test running right now' and 'awaiting manual judgment')
- What is the progress of rolling a given artifact version out across environments and resources?

After some brainstorming, it feels like we need two views on this:
1) artifact versions, with metadata about where the version came from (build, source control, etc.) and a summary of its status across environments
2) environments, and resources they contain, with a summary of the artifact versions that are relevant at that time

By combining those two lenses on artifacts + environments together, we should be able to craft a view like the one in our designs.

An open concern/issue: we originally envisioned a way of rolling out changes to resource definitions or similar and bundling together artifact version + definition change into a 'state bundle'. Right now, we don't have any concept of versioning/rolling out definitions, so to avoid trying to engineer a thing we don't understand we're just going to make the thing that used to be a 'state bundle' and artifact for the first POC. Hopefully we can get more clarity on that later, thoughts are (of course) welcome.

erikmunson:
Quick example of some brainstorming around the second view into data described (environments/resources) — imagine it's available at a path like `/application/{appName}/environments`:

```js
// sorted by the order things will flow through, based on
// depends-on constraint? e.g. prod depends-on test so
// test goes first in the list
[
  {
    name: "test",
    // Ideally this would be a merged list of all the
    // locations from all the resources in an environment,
    // which would allow us to rollup summaries about
    // environments when all resources are in
    // the same account, etc.
    locations: [{
      // probably don't need subnet, vpc
      accountName: "awstest",
      regions: [ "us-west-2" ]
    }],
    resources: [
      // We'll also want to attach artifact version data
      // onto the ResourceSummary type that comes back in
      // application/summary
      "ec2:classic-load-balancer:test:keeldemo-main-clb",
      "ec2:cluster:test:keeldemo-main",
[...]

PR #672, opened 2019-12-13
Endpoint to export artifact version status per environment

robfletcher:
As discussed in #643.


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -167,10 +179,26 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
...
           }
         }
+
+        test("the artifact version is pending in the environment") {
+          val environment1Versions = subject
+            .versionsByEnvironment(manifest)
+            .first { it.name == environment1.name }
+            .artifacts
+            .first { it.name == artifact1.name && it.type == artifact1.type }
+            .versions
```
luispollo: These lines repeat across all new tests... Would be nice to extract them into a function, maybe?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
```diff
@@ -138,6 +146,36 @@ class InMemoryArtifactRepository : ArtifactRepository {
     }
   }
 
+  override fun versionsByEnvironment(deliveryConfig: DeliveryConfig): List<EnvironmentArtifactsSummary> =
+    deliveryConfig
```
emjburns: I find this hard to parse, with the super nested structure. Any way we can simplify to make it easier to understand what's going on?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/EnvironmentArtifactsSummary.kt
```diff
@@ -0,0 +1,16 @@
...
+)
+
+data class ArtifactVersions(
+  val name: String,
+  val type: ArtifactType,
+  val versions: Map<PromotionStatus, List<String>>
+)
+
+enum class PromotionStatus {
+  PREVIOUS, CURRENT, DEPLOYING, PENDING
```
emjburns: I think you might need another state here. 

There are potentially many artifacts that were approved for an environment but not deployed to it. How are you representing that state?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -215,6 +225,113 @@ class SqlArtifactRepository(
       .execute()
   }
 
+  override fun versionsByEnvironment(deliveryConfig: DeliveryConfig): List<EnvironmentArtifactsSummary> {
+    return deliveryConfig.environments.map { environment ->
+      val artifactVersions = deliveryConfig.artifacts.map { artifact ->
+        //          select e.name, eav.artifact_version, eav.approved_at, eav.deployed_at
```
emjburns: Remove this code?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EnvironmentController.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping(path = ["/environments"])
+class EnvironmentController(
+  private val deliveryConfigRepository: DeliveryConfigRepository
+) {
+
+  @GetMapping(
```
emjburns: do you need or want a similar api at the delivery config level?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -215,6 +225,113 @@ class SqlArtifactRepository(
...
+//          and da.name = 'keel'
+//          and da.type = 'DEB'
+//          and e.delivery_config_uid = dc.uid
+//          and dc.name = 'keel-manifest'
+//          and not exists (
+//          select 1 from environment_artifact_versions eav
+//          where eav.artifact_version = dav.version
+//          and eav.environment_uid = e.uid
+//        )
+        val rawResults = jooq
```
emjburns: This query is so large and confusing. Could you put a function description that describes the steps you're taking to build this information?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -215,6 +225,113 @@ class SqlArtifactRepository(
...
+              .and(ENVIRONMENT.NAME.eq(environment.name))
+              .andNotExists(
+                selectOne()
+                  .from(ENVIRONMENT_ARTIFACT_VERSIONS)
+                  .where(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.eq(DELIVERY_ARTIFACT_VERSION.VERSION))
+                  .and(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(ENVIRONMENT.UID))
+              )
+          )
+          .fetch()
+        val turnt = rawResults.map { (version, approvedAt, deployedAt) ->
```
emjburns: what does `turnt` mean?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/EnvironmentArtifactsSummary.kt
robfletcher: Yeah, right now we're not and they show up as `DEPLOYING` which is a problem I only noticed when I pointed it at real data.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
robfletcher: D'oh

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
robfletcher: It means I forgot to inline some variables

robfletcher:
I think the query will likely change. We may even want to explicitly store data to more concretely back this API. The important thing right now is that it behaves correctly.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
robfletcher: I made it an extension on the fixture

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -215,6 +221,84 @@ class SqlArtifactRepository(
...
+            deploying = deploying?.version,
+            pending = pending.map { it.version },
+            previous = previous.map { it.version }
+          )
+        )
+      }
+      EnvironmentArtifactsSummary(environment.name, artifactVersions)
+    }
+  }
+
```
robfletcher: Hopefully more readable now

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -215,6 +221,84 @@ class SqlArtifactRepository(
       .execute()
   }
 
+  override fun versionsByEnvironment(deliveryConfig: DeliveryConfig): List<EnvironmentArtifactsSummary> {
```
asher: Can we change the method signature to support limiting? For `versionsInEnvironment`, it would probably be ok to `order by ENVIRONMENT_ARTIFACT_VERSIONS.APPROVED_AT desc` and take the topN.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -215,6 +221,84 @@ class SqlArtifactRepository(
...
+        val deploying = versions.filter { it.approvedAt != null && it.deployedAt == null }.maxBy { checkNotNull(it.approvedAt) }
+        val pending = versions.filter { it.approvedAt == null }
+        ArtifactVersions(
+          name = artifact.name,
+          type = artifact.type,
+          versions = ArtifactVersionStatus(
+            current = current?.version,
+            deploying = deploying?.version,
+            pending = pending.map { it.version }.sortedWith(artifact.versioningStrategy.comparator),
+            previous = previous.map { it.version }.sortedWith(artifact.versioningStrategy.comparator)
```
robfletcher: @emjburns good job with that `versioningStrategy`. I was worried sorting this stuff was going to be a PITA.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
robfletcher: Could we just default to a limit? I worry that it seems misleading to specify a limit in the signature that we may blow because it's the product of multiple queries. I don't think there's any reliable way to sort the 2nd query in SQL.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -215,6 +221,84 @@ class SqlArtifactRepository(
       .execute()
   }
 
+  override fun versionsByEnvironment(deliveryConfig: DeliveryConfig): List<EnvironmentArtifactsSummary> {
+    return deliveryConfig.environments.map { environment ->
```
asher: It looks like the response is entirely built from the `ENVIRONMENT_ARTIFACT_VERSIONS` table, so another approach could be to: 

1) get all envIds for the delivery config
2) select * from ENVIRONMENT_ARTIFACT_VERSIONS where ENVIRONMENT_UID in (....) and I think `versionsInEnvironment` and `pendingVersions` can both be built from that. I've found having jooq fetch into a `ResultSet` is helpful when deriving disparate things from a single query. 

I think it would still make sense to have a limit parameter on the method in this case, even though it would have to be implemented in memory, if only to limit the cost to browsers parsing the response.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -215,6 +221,84 @@ class SqlArtifactRepository(
...
+          )
+        val versions = versionsInEnvironment
+          .unionAll(pendingVersions)
+          .fetch()
+          .map { (version, approvedAt, deployedAt) ->
+            ArtifactVersionResult(version, approvedAt, deployedAt)
+          }
+        val deployed = versions.filter { it.deployedAt != null }
+        val current = deployed.maxBy { checkNotNull(it.deployedAt) }
+        val previous = deployed.filterNot { it == current }
```
asher: I don't think this is going to be right. If `deployed` was sorted by `deployedAt` in descending order, then we can `val previous = deployed.first { it != current }` but as written it looks like it might be an arbitrary assignment.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
robfletcher: It has to union that with a read from `DELIVERY_ARTIFACT_VERSION` where `NOT EXISTS` in `ENVIRONMENT_ARTIFACT_VERSIONS` because versions only appear on that latter table once they get approved.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
robfletcher: shit yes. Fixing

robfletcher: Urgh, why does Kotlin not have a combination of `sortedBy` and `sortedWith`. This is going to suck.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -215,6 +221,84 @@ class SqlArtifactRepository(
...
+            DELIVERY_CONFIG
+          )
+          .where(DELIVERY_ARTIFACT.UID.eq(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID))
+          .and(DELIVERY_ARTIFACT.NAME.eq(artifact.name))
+          .and(DELIVERY_ARTIFACT.TYPE.eq(artifact.type.name))
+          .and(ENVIRONMENT.UID.eq(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID))
+          .and(ENVIRONMENT.DELIVERY_CONFIG_UID.eq(DELIVERY_CONFIG.UID))
+          .and(DELIVERY_CONFIG.NAME.eq(deliveryConfig.name))
+          .and(ENVIRONMENT.NAME.eq(environment.name))
+        val pendingVersions = jooq
```
asher: This needs to take artifact status into account since environments can be concerned with different sets of statuses within a delivery config. The current implementation would list `CANDIDATE` artifacts as pending on an environment that only accepts `RELEASE`.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EnvironmentController.kt
robfletcher: I don't know. Do you @erikmunson?

erikmunson: I will 100% need one later when we move into the multi-app world, but for now in this moment I really don't know how to make use of that properly in deck so this is sufficient. There are so many fundamental assumptions we have about the app you're in _only_ ever showing things from the one app that I'm going to avoid figuring it out until we know more about what our desired UX is.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: Thanks!

robfletcher:
@asher 2aad114 contains filtering by artifact status. I don't see a sane way to do it in the query. Trying to rustle together a test as well (I know, I know)

robfletcher:
Ok dabe1c1 contains filtering by status with test coverage

Issue #660, opened 2019-12-10
Remove the ability to have resources without a delivery config


emjburns:
It's hard to support two different ways of defining resources. 

A specific example is clusters with artifact support. We don't want to duplicate the artifact definition in the cluster resource (`ArtifactImageProvider`) if you already should have defined the artifact in the top section of the delivery config. We want to allow the cluster to reference top level artifacts. This is hard to do if we support clusters without delivery configs.

*We've decided to experiment with only allowing resources as part of a delivery config.*

To support this we need to figure out the experience for getting started and managing just one resource. 
- Do we create a mostly empty delivery config when a user exports a resource? 
- Can we wrap individual submissions of resources in a delivery config behind the scenes? 
- Can we start out by only doing this with clusters, and then move to other resources?

PR #678, opened 2019-12-17
chore(resource): Deprecate submitting individual resources

emjburns:
Deprecating individual resource creation/deletion because of https://github.com/spinnaker/keel/issues/660


emjburns:
I'm just doing a phased approach, I'll remove them entirely once I make sure everything is working.

PR #689, opened 2019-12-18
fix(constraints): allowed-times hour bug, constraint eval ordering

asher:


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/AllowedTimesConstraintEvaluator.kt
```diff
@@ -34,10 +34,9 @@ import org.springframework.stereotype.Component
  *    windows:
  *      - days: Monday,Tuesday-Friday
  *        hours: 11-16
- *        tz: America/Los_Angeles
  *      - days: Wed
  *        hours: 12-14
- *        tz: America/Los_Angeles
+ *    tz: America/Los_Angeles
```
emjburns: rogue spotless?

asher: rogue original comment :) tz is really set for the constraint instead of on individual windows

PR #691, opened 2019-12-18
fix(titus): enforce titus resource min and max

emjburns:
also fixing a moniker test which was giving me some typing trouble when compiling via the command line


emjburns:
@erikmunson totally agree!

PR #695, opened 2019-12-30
refactor(export): Refactor export to simplify omitting defaults and creating overrides

### LINKED ISSUE #640: Revisit resource spec modeling w.r.t. defaults and exporting
Relevant slack thread:

Erik Munson:managed: 4:43 PM
potentially :hot_pepper: take I would like people's thoughts on: if we add a new piece of functionality to a resource spec, as part of the work to add that functionality three things should be included:
- Actual functionality on the spec
- The ability to default the spec, using good judgment case-by-case to decide how 'smart' a default should be
- Proper handling of defaults/other cases when exporting that resource
I say this because I've observed that as we add things, the cognitive load and coordination overhead of trying to do 1, 2, and 3 separately when we notice they haven't been done is somewhat of a drag on our focus, and as more folks adopt MD the thrash of each new feature going through 1 -> 2 -> 3 separately will likely be even more jarring/confusing.
:heavy_plus_sign:
1

Luis Pollo 4:45 PM
makes sense to me… I think part of the challenge is that the code to handle those things is also scattered across different classes, and there’s not really a framework to easily implement them
:heavy_plus_sign:
1

I haven’t fully given up on coming up with a more generic solution to define/omit defaults
But haven’t had time to think it through yet
Erik Munson:managed: 4:46 PM
i agree re: things being scattered — i've been wondering if making an effort to address all three concerns in the same cognitive cycle would help us feel and identify that pain to make it better vs. the current state where each piece of the pain is spread around to a different person, week, cognitive cycle etc.
Rob Fletcher:voidchicken: 4:46 PM
I was talking to Emily about exactly this earlier today
Erik Munson:managed: 4:47 PM
lovely. perhaps less controversial than i anticipated :slightly_smiling_face:
Luis Pollo 4:47 PM
any conclusions @fletch?
Rob Fletcher:voidchicken: 4:47 PM
no
just that I was worried it would be easy to forget about the export bit
but the export is vital functionality
Luis Pollo 4:50 PM
let me give it some thought as I pick up #625 reported by @erikmunson

luispollo:
This is my attempt at simplifying the process of omitting defaults and creating overrides when exporting resource specs. 

I'd originally envisioned being able to fully automate this process by comparing a baseline "dummy" spec with all resolvable fields resolved, but after running into several blockers (e.g. the fact that `Resolver` interfaces are defined in terms of `Resource` and not `ResourceSpec` and don't provide any feedback as to what they changed, the fact that the `resolve()` methods in `ResourceSpec` types operate on concrete resolved types and not specs, the fact that `ResourceSpec` types use a mix of spec-like data classes with all properties nullable and concrete types with no nullable properties, etc.) I gave up.

Instead, I've tried to turn 🍋into lemonade by simplifying the process to generate the spec objects for the following use cases:
1. When generating the `defaults` block in `ClusterSpec` and `TitusClusterSpec` from the corresponding server group and related objects returned from clouddriver, we had a lot of code to essentially translate from one type to another, and to omit defaults (static and dynamic) from the spec, generally by using lots of `if/else` conditions around constructor parameters, or using `copy` to modify the spec, or both. My approach was to use a bit of reflection and rely on certain assumptions (namely the availability of a primary constructor in the spec classes which receive all properties, and that property names in the clouddriver types and the spec types are the same) to automate the process of converting types. In addition, I relied on the diff library we were already using to simplify the process of omitting defaults.

2. When generating the `overrides` block in the specs, again it was convenient to just compare a spec for a region with the `defaults` object from step 1, basically using the same mechanism.

This still requires implementors to manually code in defaults/resolvables in certain places for comparison, but hopefully creates enough of a pattern to make the overall process easier. Going back to the blockers I found initially, I feel like our domain model for "concrete" resource representations and specs is too loose, which likely led to the mixed type usage I referenced above, but fixing that would be a massive and risky change set that is way beyond the scope of this PR.

Closes #640 (kinda).

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
@@ -10,6 +10,7 @@ import com.fasterxml.jackson.databind.SerializationFeature.WRITE_DATE_KEYS_AS_TI
 import com.fasterxml.jackson.databind.SerializationFeature.WRITE_DURATIONS_AS_TIMESTAMPS
 import com.fasterxml.jackson.databind.module.SimpleModule
 import com.fasterxml.jackson.databind.ser.std.ToStringSerializer
+import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator.Feature.USE_NATIVE_TYPE_ID
 import com.fasterxml.jackson.dataformat.yaml.YAMLMapper
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
 import com.fasterxml.jackson.module.kotlin.registerKotlinModule
@@ -27,7 +28,9 @@ fun configuredObjectMapper(): ObjectMapper =
  * Factory method for [YAMLMapper]s configured how we like 'em.
  */
 fun configuredYamlMapper(): YAMLMapper =
-  YAMLMapper().configureMe()
+  YAMLMapper()
+    .configureMe()
+    .disable(USE_NATIVE_TYPE_ID)
 
 private fun <T : ObjectMapper> T.configureMe(): T =
   apply {
--- a/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/api/ResourceMetadataTests.kt
+++ b/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/api/ResourceMetadataTests.kt
@@ -67,6 +67,6 @@ internal class ResourceMetadataTests : JUnit5Minutests {
         expectThat(mapper.readValue<Map<String, Any?>>(this))
           .hasEntry("namespace", "default")
       }
-    }
+      }
   }
 }
--- a/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/api/ResourceMetadataTests.kt
+++ b/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/api/ResourceMetadataTests.kt
@@ -67,6 +67,6 @@ internal class ResourceMetadataTests : JUnit5Minutests {
         expectThat(mapper.readValue<Map<String, Any?>>(this))
           .hasEntry("namespace", "default")
       }
-      }
+    }
   }
 }
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -183,19 +183,7 @@ class ClusterHandler(
         null
       },
       locations = locations,
-      _defaults = ClusterSpec.ServerGroupSpec(
-        launchConfiguration = base.launchConfiguration
-          .exportSpec(exportable.account, exportable.moniker.app),
-        capacity = base.capacity,
-        dependencies = base.dependencies,
-        health = base.health.exportSpec(),
-        scaling = if (!base.scaling.hasScalingPolicies()) {
-          null
-        } else {
-          base.scaling
-        },
-        tags = base.tags
-      ),
+      _defaults = base.exportSpec(exportable.account, exportable.moniker.app),
       overrides = mutableMapOf()
     )
 
@@ -746,12 +734,31 @@ class ClusterHandler(
   /**
    * Translates a ServerGroup object to a ClusterSpec.ServerGroupSpec with default values omitted for export.
    */
-  private fun ServerGroup.exportSpec(account: String, application: String) =
-    ClusterSpec.ServerGroupSpec(
-      launchConfiguration = launchConfiguration.exportSpec(account, application),
-      health = health.exportSpec(),
-      dependencies = dependencies
+  private fun ServerGroup.exportSpec(account: String, application: String): ClusterSpec.ServerGroupSpec {
+    val defaults = ClusterSpec.ServerGroupSpec(
+      capacity = Capacity(1, 1, 1),
+      dependencies = ClusterDependencies(),
+      health = Health().exportSpec(),
+      scaling = Scaling(),
+      tags = emptyMap()
     )
+
+    val thisSpec = ClusterSpec.ServerGroupSpec(
+      launchConfiguration = launchConfiguration.exportSpec(account, application),
+      capacity = capacity,
+      dependencies = dependencies,
+      health = health.exportSpec(),
+      scaling = if (!scaling.hasScalingPolicies()) {
+        null
+      } else {
+        scaling
+      },
+      tags = tags
+    )
+
+    // it's safe to assume a non-null result here because not all properties have defaults
+    return buildSpecFromDiff(defaults, thisSpec)!!
+  }
 
   /**
    * Translates a LaunchConfiguration object to a ClusterSpec.LaunchConfigurationSpec with default values omitted for export.
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -53,6 +53,8 @@ import com.netflix.spinnaker.keel.plugin.Resolver
 import com.netflix.spinnaker.keel.plugin.ResourceHandler
 import com.netflix.spinnaker.keel.plugin.SupportedKind
 import com.netflix.spinnaker.keel.plugin.TaskLauncher
+import com.netflix.spinnaker.keel.plugin.buildSpecFromDiff
+import com.netflix.spinnaker.keel.plugin.convert
 import com.netflix.spinnaker.keel.retrofit.isNotFound
 import java.time.Clock
 import java.time.Duration
@@ -164,8 +166,6 @@ class ClusterHandler(
       .toSet()
 
     val base = serverGroups.values.first()
-    val healthDiff = ResourceDiff(base.health, Health())
-    val modifiedHealth = healthDiff.hasChanges()
 
     val locations = SubnetAwareLocations(
       account = exportable.account,
@@ -188,11 +188,7 @@ class ClusterHandler(
           .exportSpec(exportable.account, exportable.moniker.app),
         capacity = base.capacity,
         dependencies = base.dependencies,
-        health = if (modifiedHealth) {
-          base.health.exportSpec()
-        } else {
-          null
-        },
+        health = base.health.exportSpec(),
         scaling = if (!base.scaling.hasScalingPolicies()) {
           null
         } else {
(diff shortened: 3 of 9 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
```diff
@@ -27,7 +28,9 @@ fun configuredObjectMapper(): ObjectMapper =
  * Factory method for [YAMLMapper]s configured how we like 'em.
  */
 fun configuredYamlMapper(): YAMLMapper =
-  YAMLMapper().configureMe()
+  YAMLMapper()
+    .configureMe()
+    .disable(USE_NATIVE_TYPE_ID)
```
luispollo: This is somewhat unrelated. It fixes this:
```yaml
  deployWith: !<red-black>
```
...into this:
```yaml
  deployWith:
    strategy: "red-black"
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/api/ResourceMetadataTests.kt
```diff
@@ -67,6 +67,6 @@ internal class ResourceMetadataTests : JUnit5Minutests {
         expectThat(mapper.readValue<Map<String, Any?>>(this))
           .hasEntry("namespace", "default")
       }
-    }
+      }
```
luispollo: Oops. Will fix.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -789,79 +732,54 @@ class ClusterHandler(
...
-      } else {
-        if (keyPair == defaultKeyPair) {
-          null
-        } else {
-          keyPair
-        }
-      },
-      instanceMonitoring = if (!instanceMonitoring) {
+  private fun ServerGroup.exportSpec(account: String, application: String): ClusterSpec.ServerGroupSpec {
+    val defaults = ClusterSpec.ServerGroupSpec(
```
luispollo: Unfortunately, all the spec types generally have all nullable fields initalized with `null`, as opposed to default values as commonly found in some of the clouddriver models. There's a `resolve()` method (not part of any interface) generally implemented in the top-level object like `ClusterSpec` which actually resolves several of these defaults, but in terms of the clouddriver models, which also doesn't help here. We could definitely refactor that code to be more reusable, but for now I think this is a good-enough compromise.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceSpecExportHelper.kt
```diff
@@ -0,0 +1,83 @@
...
+    .map { param ->
+      val prop = T::class.memberProperties.find { prop -> prop.name == param.name }
+      param to prop?.get(working)
+    }
+    .toMap()
+
+  return if (params.values.all { it == null }) {
+    null
+  } else {
+    ctor.callBy(params)
```
luispollo: It all boils down to this call, basically. This is using the primary constructor of the spec type with a parameter map that includes only properties that were different (plus/minus other properties depending on `allowedProperties` and `forcedProperties`).

luispollo:
P.S. Another thing I struggled with is that I think the spec models for defaults and overrides is kinda broken with regards to optionality, IMHO. In overrides, we want all properties to be nullable by definition so they can be omitted, whereas in the defaults section several properties are mandatory, and yet we're using the same data classes to represent both under the covers, which generally forces us to make even mandatory properties nullable, which in turn defeats the purpose of using Kotlin nullability to represent optionals. 

Also, as I mentioned above, we're currently not following a strict standard on how to pick the types to represent data that goes in specs -- some use dedicated data classes with all properties defaulting to `null` (e.g. `ClusterSpec.HealthSpec`), others  reuse the same data classes used to represent concrete resources returned from clouddriver (e.g. `Scaling`, `Capacity`, etc.) where properties are _not_ nullable. I don't love that we have this duplication of models just for the sake of optionality in the first place, and maybe require even a third set to represent overrides per the above, but I don't really have a great suggestion on how to solve that at the moment.

In any case, I think that solving these problems is beyond the scope of this PR, but wanted to throw that out there for comments.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -440,8 +406,22 @@ class TitusClusterHandler(
   private fun Instant.iso() =
     atZone(ZoneId.systemDefault()).format(DateTimeFormatter.ISO_DATE_TIME)
 
-  private fun TitusServerGroup.exportSpec() =
-    TitusServerGroupSpec(
+  private fun TitusServerGroup.exportSpec(application: String): TitusServerGroupSpec {
```
luispollo: This adds support for omitting defaults from Titus cluster exports, which was partly where the conversation on refactoring the export code emerged from.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
robfletcher: We either implement it like that (null defaults) because we need you to specify it _either_ cluster-wide _or_ in the region specific overrides. I don't see a way to do that with the type system alone. If we assign defaults on object construction we would not be able to say later whether that default was something intentionally set by the user, or just a default.

For example, should the default values found in region specific overrides trump the non-default value found cluster wide? What if the user had actually specified the default value in one region, but now _all_ the regions have the default because of the way the constructor works?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
robfletcher: Where was that a problem? The manifest we use for keel itself already uses the latter format.

robfletcher:
I like the general approach of using the diff tool to figure out the delta

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
luispollo: In the export API output.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: Yeah, I know... it’s complicated, which is why I gave up on refactoring the types. The only thing that occurred to me is that we could have yet another set of data classes for overrides so we could maintain proper non-nullability in the class that represents defaults, but it seemed like a lot of work for little benefit.

The other problem is that, beyond hard-coded defaults in the class constructors, we have “dynamic defaults” calculated by resolvers — those are the hardest ones to account for in the export flow as far as omitting fields with default values. I cheated a bit with things like the key pair stuff because that resolver is in our code base, but the moment someone adds a resolver as an extension, there’s no way I could think of to also dynamically figure out what to omit. Not with the current APIs anyway.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -789,79 +732,54 @@ class ClusterHandler(
...
       } else {
-        instanceMonitoring
+        scaling
       },
-      ramdiskId = ramdiskId
+      tags = tags
     )
+
+    // it's safe to assume a non-null result here because not all properties have defaults
+    return buildSpecFromDiff(defaults, thisSpec)!!
```
emjburns: I dislike using `!!`. Can we instead throw an exception if it's null? like `return buildSpecFromDiff(defaults, thisSpec) ?: throw OhNoException()`

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceSpecExportHelper.kt
emjburns: woah

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceSpecExportHelper.kt
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+/**
+ * Given an input object of an arbitrary type I, and an output type O whose structure resembles that of the input
+ * object's class, convert the input object to the output type by building a map of constructor parameters for the
+ * output type matching properties with the same name in the input object. This is similar to Jackson's
+ * ObjectMapper.convertValue(from, to), but should be cheaper.
```
emjburns: Why does it matter if it's cheaper? Is there some serious performance concern for these small objects? 

I'm asking because we use `ObjectMapper.convertValue(from, to)` a lot, and it's easy to understand and reason about. To me this is less easy to understand.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -456,12 +436,13 @@ class TitusClusterHandler(
...
-  private fun Resources.exportSpec() =
-    ResourcesSpec(
-      cpu = cpu,
-      disk = disk,
-      gpu = gpu,
-      memory = memory,
-      networkMbps = networkMbps
-    )
+    // it's safe to assume a non-null result here because not all properties have defaults
+    return buildSpecFromDiff(defaults, thisSpec)!!
```
emjburns: Same comment re: `!!` and throwing an exception

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterSpec.kt
```diff
@@ -94,7 +94,7 @@ data class TitusClusterSpec(
 }
 
 data class TitusServerGroupSpec(
-  val container: Container,
+  val container: Container? = null,
```
emjburns: Container can't be null, ever. Why the change?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
robfletcher: or `return checkNotNull(buildSpecFromDiff(defaults, thisSpec))`

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceSpecExportHelper.kt
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Given a base (a.k.a. "left") and a working (a.k.a. "right") objects of type T to compare, create a "diff" of those
+ * objects and use only the properties whose values are different between the two to create an instance of a
+ * ResourceSpec-like class of type S with the corresponding values from the working object.
+ *
+ * If all the diff properties have a null value in the working object, returns null.
+ */
+inline fun <reified T : Any, reified S : Any> buildSpecFromDiff(
+  base: T,
```
asher: I really like the approach taken here, well done.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterSpec.kt
asher: I think so the class can be used for overrides, in which case all properties need to be nullable?

luispollo:
> Container can't be null so I don't want it to be nullable

I understand your concern, but this is a side-effect is the way we’ve modeled specs, even outside of exports. Please see the comments above — hopefully they make sense.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterSpec.kt
emjburns: 👍 alright

PR #696, opened 2020-01-06
Persist promotion status for artifact versions in environments

robfletcher:


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -275,40 +309,33 @@ class SqlArtifactRepository(
...
-            ArtifactVersionResult(version, approvedAt, deployedAt)
+          .filter { (_, artifactStatus, _) ->
+            artifact !is DebianArtifact || artifact.statuses.isEmpty() || ArtifactStatus.valueOf(artifactStatus) in artifact.statuses
           }
-          .sortedWith(compareBy(artifact.versioningStrategy.comparator) { it.version })
-        val deployed = versions.filter { it.deployedAt != null }
-        val current = deployed.maxBy { checkNotNull(it.deployedAt) }
-        val previous = deployed.filterNot { it == current }
-        val deploying = versions.filter { it.approvedAt != null && it.deployedAt == null }.maxBy { checkNotNull(it.approvedAt) }
-        val pending = versions.filter { it.approvedAt == null }
```
robfletcher: I was really trying to eliminate this craziness

keel-sql/src/main/resources/db/changelog/20191218-add-status-to-environment-artifact-versions.yml
```diff
@@ -0,0 +1,17 @@
...
+            columns:
+              - column:
+                  name: status
+                  type: varchar(255)
+                  constraints:
+                    nullable: false
+      rollback:
+        - dropColumn:
+            tableName: environment_artifact_versions
+            columnName: status
```
robfletcher: That's what this is

emjburns: Is this "promotion" status or "artifact" status?

robfletcher:
@emjburns The problem with adding a denormalized `release_status` field to `environment_artifact_versions` is that I then would have to do a join every time I inserted to that table in order to avoid doing an additional join in one query.

PR #703, opened 2020-01-07
feat(constraint): kayenta canary constraint with ec2 implementation

asher:
An environment promotion constraint that deploys control (baseline) and experiment (canary)
 clusters which are evaluated by Kayenta using a constraint-provided canary config.

 When the constraint specifies a multi-region canary, the Kayenta ACA must pass in all
 regions for promotion to be allowed. Kayenta false positives can be overridden to
 allow promotion via the `DeliveryConfigController.updateConstraintStatus` REST method.

 Example constraint:
 ```
 constraints:
  - type: canary
    serviceAccount: "keel@keel.io" // Will soon move to the Delivery Config root
    canaryConfigId: b11b739d-7b1d-47c7-8430-d8ddffafb645 // Will later support named configs
    beginAnalysisAfter: PT3M
    canaryAnalysisInterval: PT10M
    lifetime: PT30M
    marginalScore: 50
    passScore: 90
    capacity: 1
    source:
      account: test
      cloudProvider: aws
      cluster: app-test-web
    regions:
      - us-east-1
      - us-west-2
 ```

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
@@ -3,6 +3,7 @@ package com.netflix.spinnaker.keel.clouddriver.model
 import com.fasterxml.jackson.annotation.JsonAlias
 import com.fasterxml.jackson.annotation.JsonCreator
 import com.netflix.spinnaker.keel.api.Capacity
+import com.netflix.spinnaker.keel.clouddriver.CloudDriverCache
 import com.netflix.spinnaker.keel.model.Moniker
 
 // todo eb: this should be more general so that it works for all server groups, not just ec2
@@ -24,6 +25,13 @@ data class ActiveServerGroup(
   val moniker: Moniker,
   val buildInfo: BuildInfo? = null
 )
+
+fun ActiveServerGroup.subnet(cloudDriverCache: CloudDriverCache): String =
+  asg.vpczoneIdentifier.substringBefore(",").let { subnetId ->
+    cloudDriverCache
+      .subnetBy(subnetId)
+      .purpose ?: error("Subnet $subnetId has no purpose!")
+  }
 
 data class ActiveServerGroupImage(
   val imageId: String,
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
@@ -18,7 +18,8 @@ import java.time.zone.ZoneRulesException
   Type(value = DependsOnConstraint::class, name = "depends-on"),
   Type(value = TimeWindowConstraint::class, name = "allowed-times"),
   Type(value = ManualJudgementConstraint::class, name = "manual-judgement"),
-  Type(value = PipelineConstraint::class, name = "pipeline")
+  Type(value = PipelineConstraint::class, name = "pipeline"),
+  Type(value = CanaryConstraint::class, name = "canary")
 )
 abstract class Constraint(val type: String)
 
@@ -68,6 +69,34 @@ data class PipelineConstraint(
   val retries: Int = 0,
   val parameters: Map<String, Any?> = emptyMap()
 ) : StatefulConstraint("pipeline")
+
+// TODO: doesn't support sliding look-back, custom metric scope, or extended scope parameters (will default to asg)
+data class CanaryConstraint(
+  val serviceAccount: String = "anonymous",
+  val timeout: Duration = Duration.ofHours(24),
+  // TODO: resolve config name -> id via kayenta/canaryConfig?application=application
+  val canaryConfigId: String,
+  val beginAnalysisAfter: Duration = Duration.ofMinutes(10),
+  val canaryAnalysisInterval: Duration = Duration.ofMinutes(30),
+  val cleanupDelay: Duration = Duration.ZERO,
+  val lifetime: Duration,
+  // TODO: this default is only for testing; we need a configurable way to resolve constraint defaults -af
+  val metricsAccountName: String = "atlas-global.prod",
+  val storageAccount: String? = null,
+  val marginalScore: Int,
+  val passScore: Int,
+  val source: CanarySource,
+  val regions: Set<String>,
+  val capacity: Int,
+  // If true, the failure of a canary in one region triggers the immediate cancellation of other regions
+  val failureCancelsRunningRegions: Boolean = true
+) : StatefulConstraint("canary")
+
+data class CanarySource(
+  val account: String,
+  val cloudProvider: String,
+  val cluster: String
+)
 
 data class TimeWindow(
   val days: String? = null,
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintDeployHandler.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintDeployHandler.kt
@@ -19,7 +19,7 @@ interface CanaryConstraintDeployHandler {
    *
    * @return A map where values are [Task]'s pointing to regional orca canary tasks, keyed by region
    */
-  fun deployCanary(
+  suspend fun deployCanary(
     constraint: CanaryConstraint,
     version: String,
     deliveryConfig: DeliveryConfig,
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintDeployHandler.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintDeployHandler.kt
@@ -1,0 +1,29 @@
+package com.netflix.spinnaker.keel.constraints
+
+import com.netflix.spinnaker.keel.api.CanaryConstraint
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.events.Task
+
+interface CanaryConstraintDeployHandler {
+  val supportedClouds: Set<String>
+
+  /**
+   * Deploy and analyze a canary using cloud provider specific logic.
+   *
+   * @param constraint The canary constraint configuration
+   * @param version Artifact to be canaried
+   * @param deliveryConfig The [DeliveryConfig] containing the environment gated by this canary
+   * @param targetEnvironment The [Environment] gated by this canary
+   * @param regions Regions that need a canary deployment, may be a subset of regions defined in [constraint]
+   *
+   * @return A map where values are [Task]'s pointing to regional orca canary tasks, keyed by region
+   */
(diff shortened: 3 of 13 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -25,6 +26,13 @@ data class ActiveServerGroup(
   val buildInfo: BuildInfo? = null
 )
 
+fun ActiveServerGroup.subnet(cloudDriverCache: CloudDriverCache): String =
+  asg.vpczoneIdentifier.substringBefore(",").let { subnetId ->
+    cloudDriverCache
+      .subnetBy(subnetId)
+      .purpose ?: error("Subnet $subnetId has no purpose!")
```
emjburns: 😆 what an accidentally funny error message

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/constraints/Ec2CanaryConstraintDeployHandler.kt
```diff
@@ -0,0 +1,166 @@
...
+    targetEnvironment: Environment,
+    regions: Set<String>
+  ): Map<String, Task> {
+
+    val judge = "canary:${deliveryConfig.application}:${targetEnvironment.name}:${constraint.canaryConfigId}"
+
+    val image = runBlocking {
+      imageService.getLatestNamedImageWithAllRegionsForAppVersion(
+        appVersion = AppVersion.parseName(version.replace("~", "_")),
+        account = "test", // TODO: update when we have "default image account" configuration
```
emjburns: I think we have this as a dynamic property, example in the `ImageResolver`

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/constraints/Ec2CanaryConstraintDeployHandler.kt
```diff
@@ -0,0 +1,166 @@
...
+        )
+      }
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    return regionalJobs.map { (region, task) ->
+      val description = "Canary $version for ${deliveryConfig.application}/environment " +
+        "${targetEnvironment.name} in $region"
+
+      region to
```
emjburns: is this the start of returning a map? `region to` is such weird syntax, it would be a nice visual cue to me if you added a `return` even though it's not required by the compiler.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintEvaluator.kt
```diff
@@ -0,0 +1,332 @@
...
+    return false
+  }
+
+  @Suppress("UNCHECKED_CAST")
+  private fun canaryStatus(attributes: CanaryConstraintAttributes?): Set<CanaryStatus> {
+    if (attributes?.executions.isNullOrEmpty()) {
+      return emptySet()
+    }
+
+    return attributes!!.executions
```
emjburns: can't this still be null here? because the if check above only checks if executions is null or empty, not if the overall object is null

keel-orca/src/test/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintEvaluatorTests.kt
```diff
@@ -0,0 +1,415 @@
...
+import strikt.assertions.all
+import strikt.assertions.contains
+import strikt.assertions.hasSize
+import strikt.assertions.isA
+import strikt.assertions.isEqualTo
+import strikt.assertions.isFalse
+import strikt.assertions.isNotNull
+import strikt.assertions.isTrue
+
+internal class CanaryConstraintEvaluatorTests : JUnit5Minutests {
```
emjburns: Great tests!

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintEvaluator.kt
robfletcher: I think the compiler is not quite grasping the implications of `isNullOrEmpty`. Is that a built in function or one of ours?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintDeployHandler.kt
```diff
@@ -0,0 +1,29 @@
...
+   * @param version Artifact to be canaried
+   * @param deliveryConfig The [DeliveryConfig] containing the environment gated by this canary
+   * @param targetEnvironment The [Environment] gated by this canary
+   * @param regions Regions that need a canary deployment, may be a subset of regions defined in [constraint]
+   *
+   * @return A map where values are [Task]'s pointing to regional orca canary tasks, keyed by region
+   */
+  fun deployCanary(
+    constraint: CanaryConstraint,
+    version: String,
```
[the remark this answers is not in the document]
emjburns: I'm not going to make that change immediately, so no rush to change here. I can do a refactor later.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -772,13 +773,6 @@ class ClusterHandler(
         cloudDriverCache.securityGroupByName(location.account, location.region, it).id
       }
 
-  private val ActiveServerGroup.subnet: String
-    get() = asg.vpczoneIdentifier.substringBefore(",").let { subnetId ->
-      cloudDriverCache
-        .subnetBy(subnetId)
-        .purpose ?: throw IllegalStateException("Subnet $subnetId has no purpose!")
-    }
```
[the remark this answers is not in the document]
asher: I wanted to be able to call `ActiveServerGroup.subnet` from the canary handler without duplicating code but I can back it out. What would you think of a
`fun subnetBy(serverGroup: ActiveServerGroup): String?` method on `cloudDriverCache`?

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintEvaluator.kt
asher: Emily called it! `isNullOrEmpty` just wasn't applicable to `attributes`. After changing to `if (attributes == null || attributes.executions.isNullOrEmpty())`, the compiler casts `attributes` to the non-nullable type.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/constraints/Ec2CanaryConstraintDeployHandler.kt
asher: Yeah, `collection.map { foo to bar }` returns a List<Pair> which is what `toMap` operates on. Where are you suggesting adding a return? `collection.map { return foo to bar }` doesn't appear to be valid. I can also see if refactoring around `regionalJobs.keys.associateWith { }` is more readable.

emjburns: Discussed offline - this syntax really through me for a loop, totally missed that this was inside the `return` & `.map {`. I'd prefer if the `try {` was extracted to a variable (`task`) and then this would become `region to task`. Fine either way though.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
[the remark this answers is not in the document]
robfletcher: Oh, I didn't spot the re-use

keel-orca/src/test/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintEvaluatorTests.kt
luispollo: +1! Your tests are always a treat to review and make the implementation itself much easier to understand. 👏🏻

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/constraints/Ec2CanaryConstraintDeployHandler.kt
```diff
@@ -0,0 +1,166 @@
...
+        try {
+          runBlocking {
+            taskLauncher.submitJobToOrca(
```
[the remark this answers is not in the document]
asher: I suspect because naming things and consistency are both difficult :) I think "task" is a better fit based on pre-keel naming conventions (i.e. keel calls the /task REST endpoint in orca, Tasks tab in deck). "Job" is used within orca in multiple places and multiple meanings, but not for an orca orchestration. I get tripped up by the juxtaposition of the `TaskLauncher` class containing "task" but its only methods use "job". I'm into the idea of refactoring for naming consistency.

asher:
@robfletcher heads up that since your review, I made `CanaryConstraintDeployHandler.deployCanary()` a suspend function and was able to eliminate all of the `runBlocking` calls in `Ec2CanaryConstraintDeployHandler`. There's still a few uses in this PR but I think we need to make `ConstraintEvaluator.canPromote()` suspend based to effectively eliminate the rest. I think that's well worth doing in a future PR, provided it doesn't overly complicate the ability to implement constraint plugins in Java down the line.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
```diff
@@ -69,6 +70,34 @@ data class PipelineConstraint(
...
+data class CanaryConstraint(
+  val serviceAccount: String = "anonymous",
+  val timeout: Duration = Duration.ofHours(24),
+  // TODO: resolve config name -> id via kayenta/canaryConfig?application=application
+  val canaryConfigId: String,
+  val beginAnalysisAfter: Duration = Duration.ofMinutes(10),
+  val canaryAnalysisInterval: Duration = Duration.ofMinutes(30),
+  val cleanupDelay: Duration = Duration.ZERO,
+  val lifetime: Duration,
+  // TODO: this default is only for testing; we need a configurable way to resolve constraint defaults -af
```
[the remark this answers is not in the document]
asher: I decided it was ok to leave as a todo for now since it can be set as needed in the constraint yml. I'll add a `ConfigurationProperties` class to redefine these defaults on a per-keel-install basis as a fast follow. The current hard-coded default (atlas-global.prod) will avoid the need to specify `metricsAccountName` for most (but not all) use-cases at Netflix. Every company that runs keel will likely want to define their own default based on their chosen telemetry platform and its configuration.

Issue #655, opened 2019-12-10
Tie artifact to a delivery config


emjburns:
Artifacts now store more information than their name and type. Because of this, you need to be able to update that information.

For example, docker artifacts also have a sort type which is one of five options. You need to be able to update that if your tagging strategy changes. Also, two people should be able to care about the same artifact but choose a different versioning strategy.

So, we need to tie the artifact to where it's being used. Maybe this means adding the delivery config name, or the cluster name, or something like that.

PR #704, opened 2020-01-08
feat(deliveryConfig): require delivery config, tie artifact to it

emjburns:
(closes https://github.com/spinnaker/keel/issues/655) This is a big PR with a major change: resources are required to be associated with a delivery config.
This PR also introduces artifact references for both aws and titus. That looks like this:

```json
"artifacts": [
    {
      "name": "emburns/spin-titus-demo",
      "type": "docker",
      "tagVersionStrategy": "branch-job-commit-by-job",
      "reference": "my-docker-artifact"
    },
   {
      "name": "deb-sample-app-server",
      "type": "deb",
      "reference": "my-deb"
    }
  ]
```
and a ec2 cluster in the same delivery config could define:
```json
{
          "apiVersion": "ec2.spinnaker.netflix.com/v1",
          "kind": "cluster",
          "spec": {
...
            "imageProvider": {
              "reference": "my-deb"
            }
...
          }
```
and a titus cluster in the same delivery config could define:
```json
{
	  "apiVersion": "titus.spinnaker.netflix.com/v1",
	  "kind": "cluster",
          "spec": {
...
            "container": {
              "reference": "my-docker-artifact"
            }
...
          }
```

### DIFF
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
@@ -89,7 +89,7 @@ class ArtifactListener(
    * For each registered debian artifact, get the last version, and persist if it's newer than what we have.
    */
   // todo eb: should we fetch more than one version?
-  @Scheduled(initialDelay = 60000, fixedDelayString = "\${keel.artifact-refresh.frequency:PT6H}")
+  @Scheduled(fixedDelayString = "\${keel.artifact-refresh.frequency:PT6H}")
   fun syncArtifactVersions() =
     runBlocking {
       artifactRepository.getAll().forEach { artifact ->
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
@@ -12,6 +12,7 @@ import com.netflix.spinnaker.keel.clouddriver.CloudDriverService
 import com.netflix.spinnaker.keel.events.ArtifactEvent
 import com.netflix.spinnaker.keel.events.ArtifactRegisteredEvent
 import com.netflix.spinnaker.keel.events.ArtifactSyncEvent
+import com.netflix.spinnaker.keel.exceptions.UnsupportedArtifactTypeException
 import com.netflix.spinnaker.keel.persistence.ArtifactRepository
 import com.netflix.spinnaker.keel.telemetry.ArtifactVersionUpdated
 import com.netflix.spinnaker.kork.artifacts.model.Artifact
@@ -37,26 +38,25 @@ class ArtifactListener(
     event
       .artifacts
       .filter { it.type.toUpperCase() in artifactTypeNames }
-      .forEach { korkArtifact ->
-        if (artifactRepository.isRegistered(korkArtifact.name, korkArtifact.type())) {
-          val artifact = artifactRepository.get(korkArtifact.name, korkArtifact.type())
+      .forEach { artifact ->
+        if (artifactRepository.isRegistered(artifact.name, artifact.type())) {
           val version: String
           var status: ArtifactStatus? = null
-          when (artifact) {
-            is DebianArtifact -> {
-              version = "${korkArtifact.name}-${korkArtifact.version}"
-              status = debStatus(korkArtifact)
+          when (artifact.type()) {
+            DEB -> {
+              version = "${artifact.name}-${artifact.version}"
+              status = debStatus(artifact)
             }
-            is DockerArtifact -> {
-              version = korkArtifact.version
+            DOCKER -> {
+              version = artifact.version
             }
-            else -> throw UnsupportedArtifactTypeException(korkArtifact.type)
+            else -> throw UnsupportedArtifactTypeException(artifact.type)
           }
           log.info("Registering version {} ({}) of {} {}", version, status, artifact.name, artifact.type)
-          artifactRepository.store(artifact, version, status)
+          artifactRepository.store(artifact.name, artifact.type(), version, status)
             .also { wasAdded ->
               if (wasAdded) {
-                publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.type))
+                publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.type()))
               }
             }
         }
@@ -116,7 +116,7 @@ class ArtifactListener(
                 // todo eb: is there a better way to think of docker status?
                 DOCKER -> null
               }
-              artifactRepository.store(artifact, latestVersion, status)
+              artifactRepository.store(artifact.name, artifact.type, latestVersion, status)
             }
           }
         }
@@ -145,7 +145,7 @@ class ArtifactListener(
           val version = "${artifact.name}-$firstVersion"
           val status = debStatus(artifactService.getArtifact(artifact.name, firstVersion))
           log.debug("Storing latest version {} ({}) for registered artifact {}", version, status, artifact)
-          artifactRepository.store(artifact, version, status)
+          artifactRepository.store(artifact.name, artifact.type, version, status)
         }
     }
 
@@ -157,7 +157,7 @@ class ArtifactListener(
       getLatestDockerTag(artifact)
         ?.let { firstVersion ->
           log.debug("Storing latest version {} for registered artifact {}", firstVersion, artifact)
-          artifactRepository.store(artifact, firstVersion, null)
+          artifactRepository.store(artifact.name, artifact.type, firstVersion, null)
         }
     }
 
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
@@ -1,5 +1,8 @@
 package com.netflix.spinnaker.keel.api
 
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.fasterxml.jackson.annotation.JsonProperty.Access
 import com.fasterxml.jackson.annotation.JsonSubTypes
 import com.fasterxml.jackson.annotation.JsonSubTypes.Type
 import com.fasterxml.jackson.annotation.JsonTypeInfo
@@ -42,7 +45,7 @@ interface DeliveryArtifact {
 @JsonTypeName("deb")
 data class DebianArtifact(
   override val name: String,
-  override val deliveryConfigName: String? = null,
+  @JsonProperty(access = Access.WRITE_ONLY) override val deliveryConfigName: String? = null,
   override val reference: String = name,
   val statuses: List<ArtifactStatus> = emptyList(),
   override val versioningStrategy: VersioningStrategy = DebianSemVerVersioningStrategy
@@ -53,16 +56,17 @@ data class DebianArtifact(
 @JsonTypeName("docker")
 data class DockerArtifact(
   override val name: String,
-  override val deliveryConfigName: String? = null,
+  @JsonProperty(access = Access.WRITE_ONLY) override val deliveryConfigName: String? = null,
   override val reference: String = name,
   val tagVersionStrategy: TagVersionStrategy = SEMVER_TAG,
   val captureGroupRegex: String? = null,
-  override val versioningStrategy: VersioningStrategy = DockerVersioningStrategy(tagVersionStrategy, captureGroupRegex)
+  @JsonProperty(access = Access.WRITE_ONLY) override val versioningStrategy: VersioningStrategy = DockerVersioningStrategy(tagVersionStrategy, captureGroupRegex)
 ) : DeliveryArtifact {
   override val type = DOCKER
 
   val organization
-    get() = name.split("/").first()
+    @JsonIgnore get() = name.split("/").first()
+
   val image
-    get() = name.split("/").last()
+    @JsonIgnore get() = name.split("/").last()
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
@@ -35,11 +35,15 @@ interface DeliveryArtifact {
   val name: String
   val type: ArtifactType
   val versioningStrategy: VersioningStrategy
+  val reference: String // friendly reference to use within a delivery config
+  val deliveryConfigName: String? // the delivery config this artifact is a part of
 }
 
 @JsonTypeName("deb")
 data class DebianArtifact(
   override val name: String,
+  override val deliveryConfigName: String? = null,
+  override val reference: String = name,
   val statuses: List<ArtifactStatus> = emptyList(),
   override val versioningStrategy: VersioningStrategy = DebianSemVerVersioningStrategy
 ) : DeliveryArtifact {
@@ -49,9 +53,16 @@ data class DebianArtifact(
 @JsonTypeName("docker")
 data class DockerArtifact(
   override val name: String,
+  override val deliveryConfigName: String? = null,
+  override val reference: String = name,
   val tagVersionStrategy: TagVersionStrategy = SEMVER_TAG,
   val captureGroupRegex: String? = null,
   override val versioningStrategy: VersioningStrategy = DockerVersioningStrategy(tagVersionStrategy, captureGroupRegex)
 ) : DeliveryArtifact {
   override val type = DOCKER
+
+  val organization
+    get() = name.split("/").first()
+  val image
+    get() = name.split("/").last()
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
@@ -11,10 +11,11 @@ interface ArtifactRepository {
   /**
    * Creates or updates a registered artifact
    */
-  // todo eb: https://github.com/spinnaker/keel/issues/655
   fun register(artifact: DeliveryArtifact)
 
-  fun get(name: String, type: ArtifactType): DeliveryArtifact
+  fun get(name: String, type: ArtifactType, deliveryConfigName: String): List<DeliveryArtifact>
+
+  fun get(name: String, type: ArtifactType, reference: String, deliveryConfigName: String): DeliveryArtifact
 
   fun isRegistered(name: String, type: ArtifactType): Boolean
 
@@ -24,20 +25,24 @@ interface ArtifactRepository {
    * @return `true` if a new version is persisted, `false` if the specified version was already
    * known (in which case this method is a no-op).
    */
+  fun store(name: String, type: ArtifactType, version: String, status: ArtifactStatus?): Boolean
+
   fun store(artifact: DeliveryArtifact, version: String, status: ArtifactStatus?): Boolean
 
   /**
    * @returns the versions we have for an artifact, optionally filtering by status if provided
    */
   fun versions(
-    name: String,
-    type: ArtifactType,
+    artifact: DeliveryArtifact,
     statuses: List<ArtifactStatus> = emptyList()
   ): List<String>
 
+  /**
+   * Lists all versions, unsorted and regardless of status
+   */
   fun versions(
-    artifact: DeliveryArtifact,
-    statuses: List<ArtifactStatus> = enumValues<ArtifactStatus>().toList()
+    name: String,
+    type: ArtifactType
   ): List<String>
 
   /**
@@ -117,6 +122,11 @@ interface ArtifactRepository {
 class NoSuchArtifactException(name: String, type: ArtifactType) :
   RuntimeException("No $type artifact named $name is registered") {
   constructor(artifact: DeliveryArtifact) : this(artifact.name, artifact.type)
+}
+
+class ArtifactNotFoundException(name: String, type: ArtifactType, reference: String, deliveryConfig: String?) :
+  RuntimeException("No $type artifact named $name with reference $reference in delivery config $deliveryConfig is registered") {
+  constructor(artifact: DeliveryArtifact) : this(artifact.name, artifact.type, artifact.reference, artifact.deliveryConfigName)
 }
 
 class ArtifactAlreadyRegistered(name: String, type: ArtifactType) :
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
@@ -23,15 +23,11 @@ interface DeliveryConfigRepository : PeriodicallyCheckedRepository<DeliveryConfi
 
   /**
    * Retrieve the [Environment] a resource belongs to, by the resource [id].
-   *
-   * @return An [Environment] or `null` if the resource is not managed via an [Environment]
    */
   fun environmentFor(resourceId: ResourceId): Environment
 
   /**
    * Retrieve the [DeliveryConfig] a resource belongs to (the parent of its environment).
-   *
-   * @return A [DeliveryConfig] or `null` if the resource is not managed via a [DeliveryConfig].]
    */
   fun deliveryConfigFor(resourceId: ResourceId): DeliveryConfig
 
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
@@ -1,5 +1,6 @@
 package com.netflix.spinnaker.keel.persistence
 
+import com.netflix.spinnaker.keel.api.ArtifactType
 import com.netflix.spinnaker.keel.api.ConstraintState
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
@@ -21,24 +22,18 @@ interface DeliveryConfigRepository : PeriodicallyCheckedRepository<DeliveryConfi
   fun get(name: String): DeliveryConfig
 
   /**
-   * Check if the given application has a [DeliveryConfig].
-   * @return Boolean
-   */
-  fun hasDeliveryConfig(application: String): Boolean
-
-  /**
    * Retrieve the [Environment] a resource belongs to, by the resource [id].
    *
    * @return An [Environment] or `null` if the resource is not managed via an [Environment]
    */
-  fun environmentFor(resourceId: ResourceId): Environment?
+  fun environmentFor(resourceId: ResourceId): Environment
 
   /**
    * Retrieve the [DeliveryConfig] a resource belongs to (the parent of its environment).
    *
    * @return A [DeliveryConfig] or `null` if the resource is not managed via a [DeliveryConfig].]
    */
-  fun deliveryConfigFor(resourceId: ResourceId): DeliveryConfig?
+  fun deliveryConfigFor(resourceId: ResourceId): DeliveryConfig
 
   /**
    * @return All [DeliveryConfig] instances associated with [application], or an empty collection if
@@ -107,3 +102,8 @@ interface DeliveryConfigRepository : PeriodicallyCheckedRepository<DeliveryConfi
 
 sealed class NoSuchDeliveryConfigException(message: String) : RuntimeException(message)
 class NoSuchDeliveryConfigName(name: String) : NoSuchDeliveryConfigException("No delivery config named $name exists in the repository")
+
+class NoMatchingArtifactException(deliveryConfigName: String, type: ArtifactType, reference: String) :
+  RuntimeException("No artifact with reference $reference and type $type found in delivery config $deliveryConfigName")
+
+class OrphanedResourceException(id: ResourceId) : RuntimeException("Resource $id exists without being a part of a delivery config")
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
@@ -12,44 +12,95 @@ import com.netflix.spinnaker.keel.api.PromotionStatus
 import com.netflix.spinnaker.keel.api.PromotionStatus.CURRENT
 import com.netflix.spinnaker.keel.api.PromotionStatus.DEPLOYING
 import com.netflix.spinnaker.keel.api.PromotionStatus.PREVIOUS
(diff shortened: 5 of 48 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
```diff
@@ -35,11 +35,15 @@ interface DeliveryArtifact {
   val name: String
```
emjburns: Artifacts now have two additional fields: `reference` and `deliveryConfigName`. Reference is a friendly name that you define to refer to your artifact later. It defaults to the "name".

We auto-populate delivery config name when you submit your delivery config. I thought about making a submitted and an actual type for the artifacts, but doing it this way allows us to more easily roll this out (existing artifacts don't have a delivery config name, so we auto populate a default, which we will have to then migrate).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
```diff
@@ -12,44 +12,95 @@ import com.netflix.spinnaker.keel.api.PromotionStatus
...
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
 class InMemoryArtifactRepository : ArtifactRepository {
-  private val artifacts = mutableMapOf<DeliveryArtifact, MutableList<ArtifactVersionAndStatus>>()
-  private val approvedVersions = mutableMapOf<Key, MutableList<String>>()
-  private val deployedVersions = mutableMapOf<Key, MutableList<String>>()
-  private val statusByEnvironment = mutableMapOf<Key, MutableMap<String, PromotionStatus>>()
+  // we want to store versions by name and type, not each artifact, so that we only store them once
+  private val versions = mutableMapOf<VersionsKey, MutableList<ArtifactVersionAndStatus>>()
```
emjburns: This is a change across both artifact repositories. I'm storing the versions of an artifact keyed off of "name" and "type". This separates the versions from the delivery artifact. I did this so that several people could care about different facets of an artifact - say, only `RELEASE` builds for a debian, and a specific versioning strategy for docker images. In the sql implementation this is the new `artifact_versions` table.

We store information about the delivery artifact separately, so that we can look up the correct filtering and sorting when we need them.

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/ContainerProvider.kt
```diff
@@ -21,25 +21,29 @@ import com.fasterxml.jackson.databind.JsonDeserializer
...
-) : Container(organization, image)
+) : ContainerProvider() {
+  fun repository() = "$organization/$image"
+}
 
 @JsonDeserialize(using = JsonDeserializer.None::class)
-data class ContainerWithVersionedTag(
-  override val organization: String,
-  override val image: String,
+data class VersionedTagProvider(
```
emjburns: A backwards compatible container provider so that existing delivery configs don't break

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/DockerImageResolver.kt
```diff
@@ -53,37 +55,47 @@ abstract class DockerImageResolver<T : ResourceSpec>(
...
+    val tag: String = findTagGivenDeliveryConfig(deliveryConfig, environment, artifact)
 
-    val newContainer = getContainer(account, container, tag)
+    val newContainer = getContainer(account, artifact, tag)
     return updateContainerInSpec(resource, newContainer)
   }
 
+  fun getArtifact(container: ContainerProvider, deliveryConfig: DeliveryConfig): DockerArtifact =
+    when (container) {
+      is ReferenceProvider -> {
```
emjburns: This is the bit that looks up the actual artifact based on a reference in your `container` section of your cluster spec

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ImageProvider.kt
```diff
@@ -41,6 +41,14 @@ data class ArtifactImageProvider(
   val artifactStatuses: List<ArtifactStatus> = emptyList() // treated as "all statuses" by ImageResolver
 ) : ImageProvider()
 
+/**
+ * Provides image id by referencing an artifact defined in the delivery config
+ */
```
emjburns: This is the bit that allows you to reference artifacts for ec2 cluster

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -44,46 +47,46 @@ class ImageResolver(
...
+        is ReferenceArtifactImageProvider -> resolveFromReference(resource, imageProvider)
+        // todo eb: artifact provider is here for backwards compatibility. Remove?
+        is ArtifactImageProvider -> resolveFromArtifact(resource, imageProvider.deliveryArtifact as DebianArtifact)
         is JenkinsImageProvider -> resolveFromJenkinsJob(resource, imageProvider)
       }
     }
     return resource.withVirtualMachineImages(image)
   }
 
+  private suspend fun resolveFromReference(
```
emjburns: this is the bit that looks up the correct artifact given a reference in the ec2 cluster spec

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolverTests.kt
```diff
@@ -158,38 +157,18 @@ internal class ImageResolverTests : JUnit5Minutests {
...
     derivedContext<Fixture<ArtifactImageProvider>>("an image derived from an artifact") {
+      val artifact = DebianArtifact(name = "fnord", deliveryConfigName = "my-manifest", statuses = listOf(RELEASE))
       fixture {
         Fixture(
-          ArtifactImageProvider(DebianArtifact("fnord"), listOf(FINAL))
+            ArtifactImageProvider(artifact, listOf(RELEASE))
         )
       }
 
-      context("the resource is not in an environment") {
```
emjburns: No longer a relevant case, resources must be in an env

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/TaskLauncherTests.kt
```diff
@@ -65,25 +64,13 @@ class TaskLauncherTests : JUnit5Minutests {
       } returns TaskRefResponse("/tasks/${randomUID()}")
     }
 
-    context("no environments") {
```
emjburns: No longer a valid case

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -46,8 +46,11 @@ class SqlArtifactRepository(
       .set(DELIVERY_ARTIFACT.UID, randomUID().toString())
       .set(DELIVERY_ARTIFACT.NAME, artifact.name)
       .set(DELIVERY_ARTIFACT.TYPE, artifact.type.name)
```
emjburns: It's probably worth giving this file a close look because I'm not a sql expert. There's lots of changes in the sql artifact repository because we're now storing versions for artifacts separate from the "delivery artifact".

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -225,9 +226,9 @@ class SqlDeliveryConfigRepository(
           constraints = mapper.readValue(constraintsJson),
           notifications = mapper.readValue(notificationsJson ?: "[]")
         )
-      }
+      } ?: throw OrphanedResourceException(resourceId)
```
emjburns: It's no longer valid to not have a delivery config for a resource, hence the new exception throwing.

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
```diff
@@ -0,0 +1,91 @@
...
+      id: tie-artifacts-to-delivery-configs
+      author: emjburns
+      changes:
+        - addColumn:
+            tableName: delivery_artifact
+            columns:
+              - column:
+                  name: delivery_config_name
+                  type: varchar(255)
+                  value: "keel"
```
emjburns: Default values here which are def garbage at first, but it should be easy enough to go back and correct them by hand.

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
```diff
@@ -0,0 +1,91 @@
...
+                  name: type
+      rollback:
+        - dropPrimaryKey:
+            constraintName: artifact_versions_pk
+            tableName: artifact_versions
+        - dropIndex:
+            indexName: artifact_versions_artifact_idx
+            tableName: artifact_versions
+
+#  - changeSet:
```
emjburns: Should I drop the old artifact versions table here? or should I leave it unused and drop it in another PR?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -172,7 +172,8 @@ class TitusClusterHandler(
     )
   }
 
-  fun generateContainer(container: ContainerWithDigest, account: String): Container {
+  // todo eb: this should generate a reference...
```
emjburns: This is a todo for another PR: essentially, fix cluster export to work with artifact references etc.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -61,24 +59,19 @@ class ApplicationController(
...
-      } else {
-        emptyList()
-      }
+      val constraintStates = deliveryConfigRepository.constraintStateFor(application)
 
       return mapOf(
         "applicationPaused" to resourcePauser.applicationIsPaused(application),
         "hasManagedResources" to resources.isNotEmpty(),
         "resources" to resources,
-        "hasDeliveryConfig" to hasDeliveryConfig,
```
emjburns: Everything always has a delivery config, so we don't need to include this field.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
```diff
@@ -94,17 +94,6 @@ class ResourceController(
     resourcePauser.resumeResource(id)
   }
 
-  @Deprecated("Individual resource creation is deprecated. Please use a delivery config.")
```
emjburns: Removal of the resource create/delete endpoints since it's for real no longer supported

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
```diff
@@ -64,77 +60,14 @@ internal class ResourceControllerTests {
   }
 
   @Test
-  fun `can create a resource as YAML`() {
```
emjburns: No longer a valid operation

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -37,26 +38,25 @@ class ArtifactListener(
...
-            else -> throw UnsupportedArtifactTypeException(korkArtifact.type)
+            else -> throw UnsupportedArtifactTypeException(artifact.type)
           }
           log.info("Registering version {} ({}) of {} {}", version, status, artifact.name, artifact.type)
-          artifactRepository.store(artifact, version, status)
+          artifactRepository.store(artifact.name, artifact.type(), version, status)
             .also { wasAdded ->
               if (wasAdded) {
-                publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.type))
+                publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.type()))
```
robfletcher: I don't understand why this has changed from a property to a function call. It seems to still be referenced as a property elsewhere

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
robfletcher: Do we also exclude `deliveryConfigName` from exported JSON/YAML?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -11,10 +11,11 @@ interface ArtifactRepository {
   /**
    * Creates or updates a registered artifact
    */
-  // todo eb: https://github.com/spinnaker/keel/issues/655
   fun register(artifact: DeliveryArtifact)
 
-  fun get(name: String, type: ArtifactType): DeliveryArtifact
+  fun get(name: String, type: ArtifactType, deliveryConfigName: String): List<DeliveryArtifact>
```
robfletcher: What's the rationale for now returning a list? This should be unique, surely?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
```diff
@@ -20,25 +21,19 @@ interface DeliveryConfigRepository : PeriodicallyCheckedRepository<DeliveryConfi
...
-   */
-  fun hasDeliveryConfig(application: String): Boolean
-
   /**
    * Retrieve the [Environment] a resource belongs to, by the resource [id].
    *
    * @return An [Environment] or `null` if the resource is not managed via an [Environment]
    */
-  fun environmentFor(resourceId: ResourceId): Environment?
+  fun environmentFor(resourceId: ResourceId): Environment
```
robfletcher: Looks like that KDoc needs updating since it can't return `null` any more. Does it throw an exception?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
```diff
@@ -20,25 +21,19 @@ interface DeliveryConfigRepository : PeriodicallyCheckedRepository<DeliveryConfi
...
 
   /**
    * Retrieve the [DeliveryConfig] a resource belongs to (the parent of its environment).
    *
    * @return A [DeliveryConfig] or `null` if the resource is not managed via a [DeliveryConfig].]
    */
-  fun deliveryConfigFor(resourceId: ResourceId): DeliveryConfig?
+  fun deliveryConfigFor(resourceId: ResourceId): DeliveryConfig
```
robfletcher: Same as ^^^

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -44,46 +47,46 @@ class ImageResolver(
     val imageProvider = resource.spec.imageProvider ?: return resource
     val image = runBlocking {
       when (imageProvider) {
-        is ArtifactImageProvider -> resolveFromArtifact(resource, imageProvider)
+        is ReferenceArtifactImageProvider -> resolveFromReference(resource, imageProvider)
+        // todo eb: artifact provider is here for backwards compatibility. Remove?
+        is ArtifactImageProvider -> resolveFromArtifact(resource, imageProvider.deliveryArtifact as DebianArtifact)
```
robfletcher: Would be good to get rid of this if we can

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClassicLoadBalancerHandlerTests.kt
```diff
@@ -63,17 +63,20 @@ internal class ClassicLoadBalancerHandlerTests : JUnit5Minutests {
...
-  private val clock = Clock.systemDefaultZone()
+  private val deliveryConfigRepository: InMemoryDeliveryConfigRepository = mockk() {
+    // we're just using this to get notifications
+    every { environmentFor(any()) } returns Environment("test")
+  }
   private val taskLauncher = TaskLauncher(
-    orcaService,
-    InMemoryDeliveryConfigRepository(clock)
+      orcaService,
+      deliveryConfigRepository
```
robfletcher: What's going on with all the indentation changes in this file? Doesn't look right to me.

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
robfletcher: I'd vote for dropping it. Do we need to migrate data across? We should be able to do that with liquibase.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
```diff
@@ -24,8 +24,8 @@ import org.springframework.test.web.servlet.result.MockMvcResultMatchers.status
 
 @ExtendWith(SpringExtension::class)
 @SpringBootTest(
-  classes = [KeelApplication::class, MockEurekaConfiguration::class],
-  webEnvironment = MOCK
+    classes = [KeelApplication::class, MockEurekaConfiguration::class],
+    webEnvironment = MOCK
```
robfletcher: more weird indentation shenanigoats in this file, doesn't look like there are any actual changes

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
asher: What happens when an artifact is referenced by multiple delivery configs?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -238,7 +239,7 @@ class SqlDeliveryConfigRepository(
       .and(ENVIRONMENT.DELIVERY_CONFIG_UID.eq(DELIVERY_CONFIG.UID))
       .fetchOne { (name) ->
         get(name)
-      }
+      } ?: throw OrphanedResourceException(resourceId)
```
asher: Not for this PR, but it looks like this exception won't trigger our flap detection veto and it would be nice to stop evaluating orphaned resources early on.

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
asher: Not sure if this makes sense over a nullable column without a default value.

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
```diff
@@ -0,0 +1,91 @@
+databaseChangeLog:
+  - changeSet:
+      id: tie-artifacts-to-delivery-configs
+      author: emjburns
+      changes:
+        - addColumn:
+            tableName: delivery_artifact
+            columns:
+              - column:
+                  name: delivery_config_name
```
asher: If 2 apps/delivery configs use the same artifact, does that currently result in 2 rows in `delivery_artifact`? I'm wondering if this should have a one-to-many relationship with `delivery_config`.

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
asher: Any concerns regarding migrating data to the new table?

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
```diff
@@ -0,0 +1,91 @@
...
+            columnName: delivery_config_name
+        - dropColumn:
+            tableName: delivery_artifact
+            columnName: reference
+  - changeSet:
+      id: create-new-artifacts-table
+      author: emjburns
+      changes:
+        - createTable:
+            tableName: artifact_versions
```
robfletcher: Does this table not need a reference to the `delivery_artifact` table?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
emjburns: @robfletcher there's not great/any support for exporting an artifact, yet. If you mean within the json when you ask for the delivery config, I definitely don't remove it. That would probably be a good improvement.

@asher it'll be a separate "artifact" in our database - that way, two configs can care about, say, the same package but different release statuses. The "delivery artifact" becomes more a way to sort and filter versions. The actual full list of versions is stored with just artifact name and type.

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
emjburns: I toyed with that too. I'm not sure what is best - but, we really don't want to allow this to be null. An artifact should be associated with a delivery config. So I felt like the initial weirdness was an ok thing for where we are in adoption.

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
emjburns: I don't think we need to migrate data - the `ArtifactListener` will fetch the latest artifact version for each artifact every ~6 hours, and I don't think loosing the "historical" artifact record matters at all.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
robfletcher: That is what I meant, yes

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
robfletcher: Should have a not null constraint then

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
emjburns: You might have two artifacts of the same name and type, but different sorting or release status. For example, you may care about only a snapshot of a deb for a feature branch env but for the main environment you care about release status.

I think I'd model that as two artifacts:
```
artifacts:
  - name: keel
     type: deb
     statuses: 
     - snapshot
     reference: feature-stack-artifact
  - name: keel
     type: deb
     statuses: 
     - release
     reference: main-artifact
```

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
emjburns: Ya I agree. We have to do it after migrating all current delivery configs though.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClassicLoadBalancerHandlerTests.kt
emjburns: I don't know! My spotless just did that :( do you have any idea how to fix it?

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
robfletcher: Oh, I think I get it. `delivery_artifact` represents a subscription to an artifact by a particular `delivery_config` and this is a common list of versions. Is that right? If so do we need a reference from `delivery_artifact` to `delivery_config`?

robfletcher: Ah, just saw @asher made the same comment above

emjburns: Nope, because we're storing versions only by name and type. Decoupling this allows us to keep track of versions independently of a delivery artifact. This means probably a little more storage of these versions, but is really a requirement when there are two artifacts that sort the versions in different ways. I can explain my thinking more offline if you'd like.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: This object (`artifact`) used to be a `DeliveryArtifact`, and that has the correct type as a property. Now, it's a `KorkArtifact`, and we need to parse the correct `ArtifactType` using the function call, because a `KorkArtifact` also has a `type` field on it, which is text.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
emjburns: I'll update. Yes it does: an `OrphanedResourceException`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
emjburns: I'll update. Yes it does: an `OrphanedResourceException`

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
emjburns: Yes! @robfletcher you are right.

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
emjburns: The information in `delivery_artifact` is specific to one delivery config. Two or more delivery configs could care about an artifact in the same way, but I don't think that's a super common case and if it happens, it's not a big deal to duplicate that sorting data. I kind of feel like it would be a lot more complicated if we had a one-to-many relationship, and I don't see it having any big advantages

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
robfletcher: So that `type()` function is defined in Kork?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
emjburns: @robfletcher made serialization better so that delivery config name doesn't show

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClassicLoadBalancerHandlerTests.kt
robfletcher: That's weird. I just ran `gradle spotlessApply` and it didn't change anything. I wonder how we have different settings.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: No it's defined in `ArtifactListener`: https://github.com/spinnaker/keel/blob/master/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt#L174

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
emjburns: Dropping the table

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
robfletcher: is nice

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
asher: Is keel a valid delivery config name? I feel like it should be a reserved value that's a little more explicit (i.e. "_pending" if there's validation that delivery config names have to start with a-zA-Z0-9)

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
asher: In that case, the `delivery_artifact_name_type_idx` index on `delivery_artifact` (name, type) needs to lose its unique constraint.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClassicLoadBalancerHandlerTests.kt
emjburns: intellij, you fool! Fixed.

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
emjburns: Great call! I'll do that.

emjburns: @asher done.

keel-sql/src/main/resources/db/changelog/20200106-tie-artifacts-to-delivery-configs.yml
emjburns: Hmm. This will only apply for existing values, are you worried that it's not explicit enough and easy to miss that it's garbage data?

PR #705, opened 2020-01-08
default the value for promotion status

robfletcher:
Fix for #696 where I added a new column but didn't set a default value.

### DIFF
--- a/keel-sql/src/main/resources/db/changelog/20191218-add-status-to-environment-artifact-versions.yml
+++ b/keel-sql/src/main/resources/db/changelog/20191218-add-status-to-environment-artifact-versions.yml
@@ -9,6 +9,7 @@ databaseChangeLog:
               - column:
                   name: promotion_status
                   type: varchar(255)
+                  value: PENDING
                   constraints:
                     nullable: false
       rollback:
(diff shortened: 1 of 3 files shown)


keel-sql/src/main/resources/db/changelog/20191218-add-status-to-environment-artifact-versions.yml
```diff
@@ -9,6 +9,7 @@ databaseChangeLog:
               - column:
                   name: promotion_status
                   type: varchar(255)
+                  value: PENDING
```
robfletcher: Set the value for all entries (gets updated later)

keel-sql/src/main/resources/db/changelog/20200107-default-promotion-status.yml
```diff
@@ -0,0 +1,19 @@
...
+                name: promotion_status
+                value: APPROVED
+          where: approved_at is not null
+      - upate:
+          tableName: delivery_artifact_version
+          columns:
+            - column:
+                name: promotion_status
+                value: PREVIOUS
+          where: deployed_at is not null
```
robfletcher: We'll need a push of each artifact to happen to populate `CURRENT` values but this will make them vaguely sensible at least

PR #708, opened 2020-01-09
fix(artifact): fixing uniqueness

emjburns:
Turns out we obviously needed a unique constraint for artifacts. Making a delivery artifact unique on name, type, delivery config, and reference. This means that you can care about the same exact artifact in more than one way (i.e. snapshot status for feature branch env, release status for prod flow) as long as you give it a different "reference".

### DIFF
--- a/keel-sql/src/main/resources/db/changelog/20200109-fix-delivery-artifact-uniqueness.yml
+++ b/keel-sql/src/main/resources/db/changelog/20200109-fix-delivery-artifact-uniqueness.yml
@@ -1,0 +1,34 @@
+databaseChangeLog:
+  - changeSet:
+      id: shorten-artifact-type
+      author: emjburns
+      changes:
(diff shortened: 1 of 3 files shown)


keel-sql/src/main/resources/db/changelog/20200109-fix-delivery-artifact-uniqueness.yml
```diff
@@ -0,0 +1,14 @@
...
+  - changeSet:
+      id: fix-delivery-artifact-uniqueness
+      author: emjburns
+      changes:
+        - addColumn:
+            tableName: delivery_artifact
+            columns:
+              - column:
+                  name: fingerprint
+                  type: varchar(255)
```
asher: Since sha1 digests are of fixed length, it would be more efficient to make this a `char` of that exact length.

PR #709, opened 2020-01-10
feat(Orca agent): Adding an agent for orca to query tasks statuses

### LINKED ISSUE #471: Propagate errors from Orca tasks to resource history
When an Orca task fails with a clear reason message, e.g. "No suitable subnet was found for internal subnet purpose 'internal (vpc0)'!" we should make that available via the resource history API.

gal-yardeni:
Closes #471 
1. Creating a scheduled agent mechanism - a new agent_lock table which will hold all scheduled agents and their lock time. For now, OrcaAgent will be the only scheduled agent.
2. Adding a new table called "task_tracking" which will contain all the active tasks from orca.
3. Whenever a new task gets created in orca, the information will be stored at this table.
4. Every minute (or other configurable time), the ocra agent will iterate over the tasks in that table and will query orca for their statuses. 
5. If the task is in a completed state (meaning, failed or succeeded), the agent will send a new event that will be logged into the resource history table. In case of a failure, we will fetch the error message from orca and log it as well. Once the agent will finish processing the terminal tasks, they will be removed from the table.


keel-sql/src/main/resources/db/changelog/20200109-create-task-tracking-table.yml
```diff
@@ -0,0 +1,43 @@
...
+        - dropTable:
+            tableName: task_tracking
+  - changeSet:
+      id: create-task_tracking-constraint-indicies
+      author: gyardeni
+      changes:
+        - addPrimaryKey:
+            tableName: task_tracking
+            constraintName: task_tracking_pk
+            columnNames: task_id, resource_id
```
asher: The PK can just be `task_id` as they are globally unique and cannot be reused.

keel-sql/src/main/resources/db/changelog/20200109-create-task-tracking-table.yml
```diff
@@ -0,0 +1,43 @@
...
+                  constraints:
+                    nullable: false
+              - column:
+                  name: task_name
+                  type: longtext
+                  value: "{}"
+                  constraints:
+                    nullable: false
+              - column:
+                  name: resource_id
```
asher: Not sure if `resource_id` should be a part of this table, we have to environment constraints that launch orca tasks so they aren't necessarily tied to a single resource.

keel-sql/src/main/resources/db/changelog/20200109-create-task-tracking-table.yml
```diff
@@ -0,0 +1,43 @@
+databaseChangeLog:
+  - changeSet:
+      id: create-task-tracking-table
+      author: gyardeni
+      changes:
+        - createTable:
+            tableName: task_tracking
```
asher: This table doesn't seem like data that needs to be stored in keel forever, in which case it would be good to add a timestamp field to help with purging old records.

luispollo:
> 3\. Every minute, a new ocra agent will iterate over the tasks in that table and will query orca for their statuses. If the tasks is in a terminal state (meaning, failed or succeeded), the agent will send a new event that will be logged into the resource history table.

IIRC, `TERMINAL` is only for failed tasks. `SUCCEEDED` is the status for success I think.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -147,6 +149,8 @@ class ImageHandler(
         )
       )
     )
+    // TODO: check this
+    publisher.publishEvent(TaskCreatedEvent(TaskRecord(taskId = taskRef.taskId, taskName = description, resourceId = resource.id.toString())))
```
luispollo: Isn't this redundant with the event published in `TaskLauncher`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/TaskTrackingRepository.kt
```diff
@@ -0,0 +1,14 @@
+package com.netflix.spinnaker.keel.persistence
+
+interface TaskTrackingRepository {
+
+  fun store(task: TaskRecord)
+  fun getTasks(): Set<TaskRecord>
```
luispollo: Probably add a `limit` parameter to this like we do for most other `get` repository methods?

keel-orca/src/test/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgentTests.kt
```diff
@@ -0,0 +1,45 @@
...
+
+internal class OrcaTaskMonitorAgentTests : JUnit5Minutests {
+
+  data class OrcaTaskMonitorAgentFixture(
+    val event: TaskCreatedEvent,
+    val repository: TaskTrackingRepository = mockk(relaxUnitFun = true),
+    val orcaService: OrcaService = mockk(relaxUnitFun = true),
+    val publisher: ApplicationEventPublisher = mockk(relaxUnitFun = true)
+  ) {
+    val listner: OrcaTaskMonitorAgent = OrcaTaskMonitorAgent(repository, orcaService, publisher)
```
luispollo:
```suggestion
    val listener: OrcaTaskMonitorAgent = OrcaTaskMonitorAgent(repository, orcaService, publisher)
```

keel-sql/src/main/resources/db/changelog/20200109-create-task-tracking-table.yml
luispollo: Maybe a generic field like `source` that we form with a prefix like `resource:` or `environment:` plus the identifier? I'm guessing this is mostly for logging/debugging purposes since the important bit is really just the `task_id`?

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.slf4j.LoggerFactory
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.context.event.EventListener
+import org.springframework.stereotype.Component
+
+// This agent will run every minute, will fetch all the records from task_tracking table,
+// and will publish the corresponding event [taskSucceeded / taskFailed] to the resource history log.
+
+@Component
+class OrcaTaskMonitorAgent(
```
luispollo:
```suggestion
class OrcaTaskMonitoringAgent(
```

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/TaskLauncher.kt
```diff
@@ -93,6 +97,8 @@ class TaskLauncher(
       )
       .let {
         log.info("Started task {} to upsert {}", it.ref, subject)
+        // TODO: check this
+        publisher.publishEvent(TaskCreatedEvent(TaskRecord(taskId = it.taskId, taskName = description, resourceId = subject)))
```
luispollo: Do we gain anything by generating an event here as opposed to just calling the `taskTrackingRepository` directly to save the record?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/TaskTrackingRepository.kt
luispollo: Actually, come to think of it, if you go ahead with a separate monitoring loop to look at orca tasks as you have proposed, I think you would need some form of distributed coordination across the keel instances so that they’re not looking at the same tasks and/or interfering with each other. When we first discussed this, my initial thought was that we could perhaps hook the retrieval of orca task info into the existing resource monitoring loop because, if an orca task fails, in theory the resource would never resolve to the desired state, in which case we’d eventually give up and generate a failure event — at which point we could try to obtain more info about the related tasks, as opposed to just blindly polling orca all the time. The disadvantage is that it might take much longer to detect the need to ask orca for details. 

At first it seemed cleaner to have a separate loop/schedule, but I’m afraid it might end up adding a lot more complexity.

Thoughts?

keel-sql/src/main/resources/db/changelog/20200109-create-task-tracking-table.yml
gal-yardeni: @asher I want to add logic that once a task is finished, it'll be removed from this table, so this table will only hold unfinished tasks. Do you think we should add the timestamp in addition?

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/TaskLauncher.kt
gal-yardeni: Yes - we want to have all the taskTrackingRepository logic in one place, which is in the OrcaAgent class.

gal-yardeni: It's a pattern I've observed used widely in keel and other projects.

keel-sql/src/main/resources/db/changelog/20200109-create-task-tracking-table.yml
gal-yardeni: After rethinking this and talking to Emily, will add the timestamp to have the ability to purge or handle records that are not getting handled.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
gal-yardeni: Will refactor this to use task launcher instead.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/TaskTrackingRepository.kt
gal-yardeni: Thanks for your comments @luispollo ! It makes complete sense to add here some mechanisms for distributed systems. I will refactor the repository to use PeriodicallyCheckedRepository.

keel-sql/src/main/resources/db/changelog/20200109-create-task-tracking-table.yml
gal-yardeni: Thanks for the feedback - to keep this data as generic as possible for future uses as well (we might want to include errors from ocra for constraints as well), will rename the `resource_id` to `subject`, like the actual field in the taskLauncher.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/TaskTrackingRepository.kt
gal-yardeni: @asher would love to get your feedback on the following:
For other classes that implement `PeriodicallyCheckedRepository`, there are 2 tables - the actual data table and the XXX_last_checked table. Do you think it makes sense to add another table to hold the task_last_checked here? Or maybe just use a different mechanism to find the correct tasks?

asher: I think a different mechanism would be preferable since tasks should be short lived. I think it would make sense to hold a lock on running `OrcaTaskMonitorAgent` so that only one keel instance is checking tasks at a time, but not at the individual task level.

I wrote a simple sql based agent lock implementation for this purpose in orca: https://github.com/spinnaker/orca/blob/master/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/notifications/SqlNotificationClusterLock.kt

The same pattern would work well here. That class could optionally be moved to `kork` with names made more generic to avoid reimplementing.

gal-yardeni:
@asher I refactor the code based on your recommendation - let me know if other things need to be fixed before moving forward with the actual orca agent implementation. 
Regarding the handlers who are calling 'taskLauncher.submitJobToOcra` method without a subject -> there is an overloading `submitJobToOcra` method within the taskLauncher that fetched the resource_id from the resource that is sent, and assing it as a subject, so there it looks like there is no call to orca that gets an empty subject field.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlAgentLockRepository.kt
```diff
@@ -0,0 +1,72 @@
...
+  private val jooq: DSLContext,
+  private val clock: Clock,
+  private val retryProperties: RetryProperties,
+  private val agents: List<ScheduledAgent>
+) : AgentLockRepository {
+
+  @Scheduled(fixedDelayString = "\${keel.scheduled.agent.frequency:PT1m}")
+  fun invoke() {
+    agents.forEach {
+      val lockAcquired = tryAcquireLock(it.javaClass.simpleName, TimeUnit.MINUTES.toSeconds(5))
```
asher: Since lock time has the effect of setting execution frequency, I think it should be a property on `ScheduledAgent` that can be overridden by implementing classes.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
```diff
@@ -0,0 +1,88 @@
...
+    coroutineScope {
+      taskTrackingRepository.getTasks()
+        .associate {
+          it.subject to
+            async {
+              orcaService.getOrchestrationExecution(it.id)
+            }
+        }
+        .mapValues { it.value.await() }
+        .map { (resourceId, taskDetails) ->
```
asher: I think the map lambda would be more readable without the nested `if` statements, could do something like:
```suggestion
        .filterValues { it.status.isComplete() }
        .map { (resourceId, taskDetails) ->
           val succeeded = taskDetails.isSuccess()
           when (succeeded) {
             true -> publisher.publishEvent(
               ResourceActuationSucceeded(
                   resourceRepository.get(ResourceId(resourceId)), clock))
             false -> publisher.publishEvent(
               ResourceActuationFailed(
                 resourceRepository.get(ResourceId(resourceId)), "", clock))  
```

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  // 1. Get active tasks from task tracking table
+  // 2. get tasks status from orca
+  // 3. publish event
+  override suspend fun invokeAgent() {
+    coroutineScope {
+      taskTrackingRepository.getTasks()
+        .associate {
+          it.subject to
```
asher: `it.subject` is assumed to be a `resourceId` below but I think it'll need to be parsed and its type determined. The calls to TaskLauncher haven't been updated yet,  but I think we want to support subject values like: `resource:$id` or `constraint:$id`

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
gal-yardeni: Thanks! I was thinking about how to get rid of that `if` statement.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/TaskLauncher.kt
```diff
@@ -75,7 +82,31 @@ class TaskLauncher(
...
+  ): Task =
+    submitJobToOrca(
+      serviceAccount = serviceAccount,
+      application = application,
+      notifications = notifications,
+      subject = subject,
+      description = description,
+      correlationId = correlationId,
+      stages = stages,
+      type = SubjectType.CONSTRAIN,
```
gal-yardeni: @emjburns following our conversation, I added a `type` property. Since I didn't want to change the signature of the current calls, I added another overloading method. Let me know what you think!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/SubjectType.kt
```diff
@@ -0,0 +1,6 @@
+package com.netflix.spinnaker.keel.api
+
+// currently we have only resource and costarain types
+enum class SubjectType {
+  RESOURCE, CONSTRAIN;
```
emjburns:
```suggestion
  RESOURCE, CONSTRAINT;
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -289,6 +291,54 @@ data class ResourceActuationResumed(
...
+data class ResourceTaskFailed(
+  override val apiVersion: ApiVersion,
+  override val kind: String,
+  override val id: String,
+  override val application: String,
+  val reason: String?,
+  override val timestamp: Instant
+) : ResourceEvent() {
+  @JsonIgnore
+  override val ignoreRepeatedInHistory = true
```
emjburns: Do we want to ignore repeat events? I think the method that checks this only cares about the type of message, not the contents. So I'd say we probably want to have `ignoreRepeatedInHistory = false`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -289,6 +291,54 @@ data class ResourceActuationResumed(
...
+ */
+data class ResourceTaskSucceeded(
+  override val apiVersion: ApiVersion,
+  override val kind: String,
+  override val id: String,
+  override val application: String,
+  override val timestamp: Instant
+) : ResourceEvent() {
+  @JsonIgnore
+  override val ignoreRepeatedInHistory = true
```
emjburns: same with this one

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
gal-yardeni: Right, changed it. thanks!

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
```diff
@@ -0,0 +1,90 @@
...
+  fun onTaskEvent(event: TaskCreatedEvent) {
+    taskTrackingRepository.store(event.taskRecord)
+  }
+
+  // 1. Get active tasks from task tracking table
+  // 2. For each task, call orca and ask for details
+  // 3. For each completed task, will emit an event for success/failure
+  override suspend fun invokeAgent() {
+    coroutineScope {
+      taskTrackingRepository.getTasks()
```
asher: This doesn't need to be addressed in this PR, but we should be sure to limit the number of tasks we'll concurrently request from orca here. Since `getTasks()` returns a list, can use `List.chunked()` to chunk the task list into smaller lists, which we'd use to put a ceiling on the number of concurrent orca requests we'd `await` for.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
```diff
@@ -0,0 +1,90 @@
...
+          if (resourceId.startsWith(SubjectType.RESOURCE.toString())) {
+            val id = resourceId.substringAfter(":")
+            try {
+              when (taskDetails.status.isSuccess()) {
+                true -> publisher.publishEvent(
+                  ResourceTaskSucceeded(
+                    resourceRepository.get(ResourceId(id)), clock))
+                false -> publisher.publishEvent(
+                  ResourceTaskFailed(
+                    resourceRepository.get(ResourceId(id)), taskDetails.execution.stages.getFailureMessage(), clock))
```
asher: It would be good to have atlas metrics that reflect the rate of task failed/succeeded events, tagged by the application (I think the cardinality of `resourceId` is too high.)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlAgentLockRepository.kt
```diff
@@ -0,0 +1,109 @@
...
+            AGENT_LOCK.EXPIRY.lt(now.toEpochMilli())
+          )
+          .execute()
+      }
+    }
+
+    return changed == 1
+  }
+
+  private fun <T> withRetry(action: () -> T): T {
```
emjburns: I made this a [common function](https://github.com/spinnaker/keel/blob/master/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlRetry.kt) - you can just wire in that class to here, and use it like [this](https://github.com/spinnaker/keel/blob/master/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt#L51)

gal-yardeni: Will address it in a separate pr, thanks for the reminder!

gal-yardeni:
> @gal-yardeni Late review: this was great and should be very helpful to our users. I just missed seeing a test for the actual behavior of `OrcaMonitorAgent.invokeAgent()`.

Will be added in my next PR. thanks!

PR #710, opened 2020-01-10
feat(constraint): option to pass a multi-region canary that succeeds in a subset of regions

asher:


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Constraint.kt
```diff
@@ -89,8 +90,22 @@ data class CanaryConstraint(
   val regions: Set<String>,
   val capacity: Int,
   // If true, the failure of a canary in one region triggers the immediate cancellation of other regions
-  val failureCancelsRunningRegions: Boolean = true
-) : StatefulConstraint("canary")
+  val failureCancelsRunningRegions: Boolean = true,
+  // If set to >0, a multi-region canary only has to pass in this number of regions for the constraint to pass
+  val passIfSucceedsInNRegions: Int = 0
```
asher: Don't love this property name but not thinking of good alternates at the moment.

luispollo: `successfulRegionsThreshold`?

luispollo: `minSuccessfulRegions`?

PR #711, opened 2020-01-10
fix(artifact): make fingerprint unique

emjburns:
Now that the artifact fingerprint is calculated, make it unique. 

~@asher I also want to make the fingerprint column non-null. Do I use [addNotNullConstraint](https://www.liquibase.org/documentation/changes/add_not_null_constraint.html) or change data type? Or is there a better way to do this?~

Removed `computeFingerprints` and added the not-null constraint.


asher:
@emjburns `addNotNullConstraint` looks good, without a `defaultNullValue`. Should this PR also disable `computeFingerprints`?

emjburns:
@asher yep, added the not null and removed the migration function. I think this is ready to go now.

PR #715, opened 2020-01-11
feat(debug): a way to add debugging context to resources

emjburns:
I've been frustrated when I've had to debug resource actuation decisions and I haven't been able to get a lot of information about why something happened. For example, why does my image keep rebaking? 

I thought it might be useful to create a way to attach debug messages to resources - sort of like the way we have context on clouddriver tasks which tell the steps taken. I didn't want to just log messages, because then you have to rely on remembering what to search for. So, I created a "debug log" for each resource. 

To use it, all you have to do is emit an event with the resource id (a `DebugEvent`). These will get collected and surfaced from the endpoint `resources/{id}/debug`.

The idea is that this will mostly be used for us to figure out what is going wrong, but it could be available for a user to get more information about why we did something. This will provide a lot more detail than the current event log - and that's good! I think they should stay separate.

A sample:

```json

[{
    "resourceId": "bakery:image:deb-sample-app-server",
    "context": "Finding current image",
    "message": "current image is Image(baseAmiVersion=nflx-base-5x, appVersion=deb-sample-app-server-0.69.0, regions=[us-east-1, us-west-2])",
    "timestamp": 1578765130641
  },
  {
    "resourceId": "bakery:image:deb-sample-app-server",
    "context": "Finding latest qualifying named image for deb-sample-app-server",
    "message": "rejected images={deb-sample-app-server-0.73.0-h303.1=[image is only in regions [us-west-2], not all required regions ([us-east-1, us-west-2])], deb-sample-app-server-0.73.0-h303.1=[image is only in regions [us-east-1, us-west-1], not all required regions ([us-east-1, us-west-2])]}",
    "timestamp": 1578765130640
  },
  {
    "resourceId": "bakery:image:deb-sample-app-server",
    "context": "Finding latest qualifying named image for deb-sample-app-server",
    "message": "selected image=NamedImage(imageName=deb-sample-app-server-0.69.0-h299.b ... blah blah)",
    "timestamp": 1578765130640
  },
  {
    "resourceId": "bakery:image:deb-sample-app-server",
    "context": "Resolving type",
    "message": "baseimage=xenialbase-x",
    "timestamp": 1578765130534
  },
  {
    "resourceId": "bakery:image:deb-sample-app-server",
    "context": "Finding latest version of deb-sample-app-server",
    "message": "latest known version = deb-sample-app-server-0.73.0-h303.1",
    "timestamp": 1578765129929
  }]
```

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
@@ -19,16 +19,20 @@ package com.netflix.spinnaker.keel.clouddriver
 
 import com.netflix.frigga.ami.AppVersion
 import com.netflix.spinnaker.keel.api.DEFAULT_SERVICE_ACCOUNT
+import com.netflix.spinnaker.keel.api.ResourceId
 import com.netflix.spinnaker.keel.clouddriver.model.Image
 import com.netflix.spinnaker.keel.clouddriver.model.NamedImage
 import com.netflix.spinnaker.keel.clouddriver.model.NamedImageComparator
 import com.netflix.spinnaker.keel.clouddriver.model.appVersion
 import com.netflix.spinnaker.keel.clouddriver.model.hasAppVersion
+import com.netflix.spinnaker.keel.events.DebugEvent
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import org.springframework.context.ApplicationEventPublisher
 
 class ImageService(
-  private val cloudDriverService: CloudDriverService
+  private val cloudDriverService: CloudDriverService,
+  private val publisher: ApplicationEventPublisher
 ) {
   val log: Logger by lazy { LoggerFactory.getLogger(javaClass) }
 
@@ -39,8 +43,8 @@ class ImageService(
    * If possible, return the latest image that's present in all regions and has tags.
    * If that doesn't exist, just return the latest image.
    */
-  suspend fun getLatestImageWithAllRegions(artifactName: String, account: String, regions: List<String>): Image? =
-    getLatestNamedImageWithAllRegions(artifactName, account, regions)?.toImage(artifactName)
+  suspend fun getLatestImageWithAllRegions(resourceId: ResourceId, artifactName: String, account: String, regions: List<String>): Image? =
+    getLatestNamedImageWithAllRegions(resourceId, artifactName, account, regions)?.toImage(artifactName)
       ?: getLatestImage(artifactName, account)
 
   private fun NamedImage.toImage(artifactName: String): Image? =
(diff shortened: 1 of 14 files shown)


gal-yardeni:
I love it! Thanks for adding it, I think it's great to have an easier way to figure out the lifecycle of a specific resource without digging into the logs and/or db.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -123,21 +127,46 @@ class ImageService(
...
+          errors += "[image does not have tags for all regions ($regions): existing tags ${it.tagsByImageId}]"
+        }
+        if (errors == "") {
+          true
+        } else {
+          eliminatedImages[it.imageName] = errors
+          false
+        }
       }
+    publisher.publishEvent(DebugEvent(resourceId, "Finding latest qualifying named image for $packageName", "rejected images=$eliminatedImages"))
```
gal-yardeni: Why are there 2 events published here?

gal-yardeni:
Comment about the structure:
Is there a way to add if the message is info/debug/error? from just looking at the response for `resources/{id}/debug`, it's a bit hard to understand what went wrong (you need to understand all the errors/syntax that we're using)

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
emjburns: I want to log all the images that were rejected and why, as well as the image that was chosen

luispollo:
I love this idea, thanks for working on it. However, this is precisely the problem I was trying to solve with #652. As we discussed at the time, the reason why it was still hard to debug after adding that was there was just not a log of logging happening...

The bulk of your PR is doing exactly that: instrumenting the code to generate meaningful log events. I'd be curious to see if you just added a SLF4J logging statement to `DebugEventListener.onDebugEvent` if they would get the corresponding `X-SPINNAKER-RESOURCE-ID` attached to them, my point being that, either way, you'll need to know the resource ID to search either type of log, so if we could just leverage the existing logging infrastructure I'd favor that as opposed to a custom solution. Speaking of which, one immediate concern that occurs to me that we don't even think about when using standard logging is with the size of those logs... Here, IIUC, there's nothing to prevent the table from growing indefinitely at the moment.

I'd be more than happy to collaborate to improve the tracing support stuff so that it fulfills our needs. I hope this doesn't sound like I'm trying to compete for the best solution, I just in general like to avoid implementing custom solutions when there's a (somewhat?) standard solution available.

asher:
Writing two rows to the DB every time an cluster or image resource is evaluated doesn't seem like a good idea. Doing so only when a diff is detected might be appropriate but I have concerns about the current behavior which seems better suited to logging via slf4j.

keel-sql/src/main/resources/db/changelog/20201011-resource-debug-log.yml
```diff
@@ -0,0 +1,60 @@
+databaseChangeLog:
+  - changeSet:
+      id: resource-debug-log
+      author: emjburns
+      changes:
+        - createTable:
+            tableName: resource_debug
```
asher: A table without a primary or unique key is a table from which rows cannot be deleted without blocking inserts.

emjburns:
@asher I'm hesitant to add this information to the logs because that can't be surfaced to a user. Let's all chat offline about how we want to surface information, and then depending on what we decide we'll either close this PR or I'll brainstorm new ways to make this not write to the db so much

emjburns:
We discussed offline, and it feels best to add this information to logs at first. When we know what questions customers are asking we can figure out how to surface that information.

PR #716, opened 2020-01-13
bring actuationInProgress method signature in line

robfletcher:
I noticed this one method signature didn't align with the generic type on the handler class itself.

### DIFF
--- a/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -186,5 +186,11 @@ class ResourceActuator(
     resourceDiff: ResourceDiff<*>
   ): List<Task> =
     update(resource as Resource<S>, resourceDiff as ResourceDiff<R>)
+
+  @Suppress("UNCHECKED_CAST")
+  private suspend fun <S : ResourceSpec> ResourceHandler<S, *>.actuationInProgress(
+    resource: Resource<*>
+  ): Boolean =
+    actuationInProgress(resource as Resource<S>)
   // end type coercing extensions
 }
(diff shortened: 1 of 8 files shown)


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -186,5 +186,11 @@ class ResourceActuator(
     resourceDiff: ResourceDiff<*>
   ): List<Task> =
     update(resource as Resource<S>, resourceDiff as ResourceDiff<R>)
+
+  @Suppress("UNCHECKED_CAST")
+  private suspend fun <S : ResourceSpec> ResourceHandler<S, *>.actuationInProgress(
+    resource: Resource<*>
+  ): Boolean =
+    actuationInProgress(resource as Resource<S>)
```
robfletcher: cast in one place instead of in every handler

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt
```diff
@@ -164,10 +164,10 @@ abstract class ResourceHandler<S : ResourceSpec, R : Any>(
   }
 
   /**
-   * @return `true` if this plugin is still busy running a previous actuation for the resource
-   * associated with [id], `false` otherwise.
+   * @return `true` if this plugin is still busy running a previous actuation for [resource],
+   * `false` otherwise.
    */
-  open suspend fun <T : ResourceSpec> actuationInProgress(resource: Resource<T>): Boolean = false
+  open suspend fun actuationInProgress(resource: Resource<S>): Boolean = false
```
robfletcher: Looks like this method was previously changed from a resource id parameter to the full resource, but the generic type was not kept consistent with the handler's existing generic type.

Issue #690, opened 2019-12-18
ArtifactStatus filtering needs to happen before Env Constraint checks


asher:
From `EnvironmentPromotionChecker`:

TODO (Critical): The EC2 ClusterSpec enables artifact filtering by ArtifactStatus. This is
currently handled in ImageResolver after artifacts are "approved" for an environment.
Instead, this needs to happen here or earlier. Otherwise we may deploy canary clusters or
request manual judgements for artifacts that should never be allowed in a given environment.

PR #717, opened 2020-01-14
fix(artifacts): don't approve a version with an invalid status

emjburns:
Fixes https://github.com/spinnaker/keel/issues/690

When we ask for versions of an artifact, we now only get valid versions based on the artifact status information.

This means we can't approve a version for deployment that has an invalid status.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
@@ -40,9 +40,11 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
   data class Fixture<T : ArtifactRepository>(
     val subject: T
   ) {
-    val artifact1 = DebianArtifact("foo", deliveryConfigName = "my-manifest", statuses = listOf(SNAPSHOT))
-    val artifact2 = DebianArtifact("bar", deliveryConfigName = "my-manifest", statuses = listOf(SNAPSHOT))
-    val artifact3 = DockerArtifact("baz", deliveryConfigName = "my-manifest")
+    // the artifact built off a feature branch
+    val artifact1 = DebianArtifact("keeldemo", deliveryConfigName = "my-manifest", reference = "candidate", statuses = listOf(SNAPSHOT))
+    // the artifact built off of master
+    val artifact2 = DebianArtifact("keeldemo", deliveryConfigName = "my-manifest", reference = "master", statuses = listOf(RELEASE))
+    val artifact3 = DockerArtifact("docker", deliveryConfigName = "my-manifest", reference = "docker-artifact")
     val environment1 = Environment("test")
     val environment2 = Environment("staging")
     val manifest = DeliveryConfig(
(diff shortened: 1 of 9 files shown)


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -408,22 +416,6 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
       }
     }
 
-    context("getting latest artifact approved in env respects status") {
```
emjburns: This test is no longer valid.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -40,9 +40,11 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
   data class Fixture<T : ArtifactRepository>(
     val subject: T
   ) {
```
emjburns: I updated the tests in here a little bit for clarity and to reflect the slight changes in functionality.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
```diff
@@ -232,7 +229,10 @@ class InMemoryArtifactRepository : ArtifactRepository {
                 current = statuses.filterValues { it == CURRENT }.keys.firstOrNull(),
                 deploying = statuses.filterValues { it == DEPLOYING }.keys.firstOrNull(),
                 pending = versions[VersionsKey(artifact.name, artifact.type)]
```
emjburns: Just formatting here. My intellij is spunky...

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolverTests.kt
```diff
@@ -217,19 +216,6 @@ internal class ImageResolverTests : JUnit5Minutests {
           }
         }
 
-        context("only an artifact with the wrong status has been approved for the environment") {
```
emjburns: No longer a valid case.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -178,21 +176,16 @@ class SqlArtifactRepository(
...
   ): String? {
     val environment = deliveryConfig.environmentNamed(targetEnvironment)
     val envUid = deliveryConfig.getUidFor(environment)
     val artifactId = artifact.uid
     val versions: List<String> = jooq
       .select(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
-      .from(ENVIRONMENT_ARTIFACT_VERSIONS, ARTIFACT_VERSIONS)
+      .from(ENVIRONMENT_ARTIFACT_VERSIONS)
       .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(envUid))
       .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifactId))
```
asher: shouldn't this also have something like: 

`.and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.eq(PromotionStatus.APPROVED.toString()))`

emjburns: I think yes! I'm going to dig into the logic here to understand why it wasn't like that before, and return with a fix or an explanation.

emjburns: Ok, so we don't write to this table unless the artifact is approved. Basically any `PromotionStatus` means it's been approved for the environment. So we don't need that.

PR #718, opened 2020-01-14
Optionally specify locations at the environment level, inherited by all resources

robfletcher:
Still to-do:

- [ ] Fix resources that use `locations` in their `id`
- [ ] Add optional `locations` to `SubmittedEnvironment` and wire it up

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
@@ -4,7 +4,14 @@ import com.netflix.spinnaker.keel.api.ArtifactType
 import com.netflix.spinnaker.keel.api.ConstraintState
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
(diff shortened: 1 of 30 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
```diff
@@ -96,6 +103,30 @@ interface DeliveryConfigRepository : PeriodicallyCheckedRepository<DeliveryConfi
...
+ * Resolve [Locations] either from the spec or from [resource]'s [Environment]. One or the other
+ * must be non-null or an exception is thrown.
+ */
+fun <S : Locatable<SubnetAwareLocations>> DeliveryConfigRepository.resolveLocations(resource: Resource<S>): SubnetAwareLocations {
+  val specLocations = resource.spec.locations
+  val envLocations = environmentFor(resource.id).locations
+  return checkNotNull(specLocations ?: envLocations) {
+    "Locations must be specified either in the resource spec, or its environment"
+  }
+}
```
robfletcher: Standard extension methods used to resolve locations from `resource` or the `Environment` it belongs to.

I feel like I could probably do something clever with generic types and make this just one method.

robfletcher:
Closing as #730 is a much less intrusive approach

PR #723, opened 2020-01-14
feat(constraints): Add interactive notifications for manual judgment

### LINKED ISSUE #733: Implement interactive notifications for manual judgement constraint
Currently the only way to approve/reject a manual constraint is by calling the API. The goal of this issue is to implement a more user-friendly alternative by using Slack notifications that allow the user to push a button to indicate approval.

luispollo:
Closes #733.

Adds interactive notifications for the manual judgment constraint, so that users can approve/reject primarily from chat tools (currently only Slack support is available in `echo`).

Here's what it looks like in Slack:
Manual Jugment

This is currently disabled by default as we figure out how to handle Slack notification callbacks in the Netflix deployment.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
```diff
@@ -16,7 +17,9 @@ data class ConstraintState(
   val judgedBy: String? = null,
   val judgedAt: Instant? = null,
   val comment: String? = null,
-  val attributes: ConstraintStateAttributes? = null
+  val attributes: ConstraintStateAttributes? = null,
+  @JsonIgnore
+  var uid: String? = null
```
luispollo: I had to expose the `uid` here because I needed something short that I could include in the callback ID and that was unique.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
```diff
@@ -62,11 +66,10 @@ abstract class StatefulConstraintEvaluator<T : Constraint> : ConstraintEvaluator
...
         type = constraint.type,
         status = ConstraintStatus.PENDING
-      )
-        .also {
-          deliveryConfigRepository.storeConstraintState(it)
-          // TODO: Emit an event here?
-        }
+      ).also {
+        deliveryConfigRepository.storeConstraintState(it)
+        eventPublisher.publishEvent(ConstraintStateChanged(targetEnvironment, constraint, null, it))
```
luispollo: Somewhat unrelated, but this may help with publishing a user-visible resource history event when constraints approve/reject an artifact. Cc: @asher 

P.S. It increasingly feels to me that we'll need other event types beyond "resource events" (events associated with artifacts, environments, etc.) that would be useful to expose in the UI.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/constraints/Ec2CanaryConstraintDeployHandler.kt
```diff
@@ -16,7 +16,7 @@ import com.netflix.spinnaker.keel.constraints.toStageBase
 import com.netflix.spinnaker.keel.ec2.resolvers.ImageResolver
 import com.netflix.spinnaker.keel.events.Task
 import com.netflix.spinnaker.keel.model.parseMoniker
-import com.netflix.spinnaker.keel.model.toEchoNotification
+import com.netflix.spinnaker.keel.model.toOrcaNotification
```
luispollo: I renamed this because it is actually an orca-specific format to represent notifications, and not what the echo API uses (orca does the translation).

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ManualJudgementNotifier.kt
```diff
@@ -0,0 +1,73 @@
...
+                    value = ConstraintStatus.OVERRIDE_PASS.name
+                  ),
+                  EchoNotification.ButtonAction(
+                    name = "manual-judgement",
+                    label = "Reject",
+                    value = ConstraintStatus.OVERRIDE_FAIL.name
+                  )
+                ),
+                color = "#fcba03"
+              )
```
luispollo: The magic pretty much boils down to this. This is what tells echo to include the buttons in the Slack message and provides callback properties that allow echo to call us back with the response.

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/model/EchoNotification.kt
```diff
@@ -0,0 +1,101 @@
...
+data class EchoNotification(
+  val notificationType: Type,
+  val to: List<String>,
+  val cc: List<String>? = null,
+  val templateGroup: String? = null,
+  val severity: Severity,
+  val source: Source? = null,
+  val interactiveActions: InteractiveActions?,
+  val additionalContext: Map<String, Any?>? = null
+) {
```
luispollo: This is the actual format of an echo notification.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/model/OrcaNotification.kt
```diff
@@ -26,7 +26,8 @@ import com.netflix.spinnaker.keel.model.NotificationEvent.ORCHESTRATION_COMPLETE
 import com.netflix.spinnaker.keel.model.NotificationEvent.ORCHESTRATION_FAILED
 import com.netflix.spinnaker.keel.model.NotificationEvent.ORCHESTRATION_STARTING
 
-data class EchoNotification(
+// This gets translated into an echo notification format in orca
+data class OrcaNotification(
```
luispollo: I renamed this because it is actually an orca-specific format to represent notifications, and not what the echo API uses (orca does the translation).

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -347,6 +349,52 @@ class SqlDeliveryConfigRepository(
       }
   }
 
+  override fun getConstraintStateForNotification(notificationId: String): ConstraintState? {
```
luispollo: I started with this name because I planned to add a `notification_id` column to the table, when I realized I could just use the `uid`. Will rename.

keel-sql/src/main/resources/db/changelog/20200112-add-notification-id-to-constraint-state.yml
```diff
@@ -0,0 +1,30 @@
+#databaseChangeLog:
```
luispollo: Will remove.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/InteractiveNotificationCallbackController.kt
```diff
@@ -0,0 +1,45 @@
...
+  @PostMapping(
+    consumes = [MediaType.APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
+    produces = [MediaType.APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  // TODO: This should be validated against write access to a service account. Service accounts should
+  //  become a top-level property of either delivery configs or environments.
+  // TODO(lfp): Related to the above, we'll need an additional authentication method for interactive constraint
+  //  approval outside of the Spinnaker UI, e.g. in Slack, since X-SPINNAKER-USER will be extracted from the Slack
+  //  message and not provided by the UI. My plan is to include an OTP in the callback URL.
+  fun handleInteractionCallback(
```
luispollo: This is where the second half of the magic happens. This is the callback we get from echo with the user's response.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
asher: 💯

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/model/EchoNotification.kt
```diff
@@ -0,0 +1,101 @@
...
+    val messageId: String,
+    val user: String
+  )
+}
+
+//  class ButtonAction extends InteractiveAction {
+//    String label
+//    String value
+//
+// //    ButtonAction(String name, String label, String value) {
```
asher: Just a reminder to clean up before merging.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
asher: Minor nit, since `notificationId == ENVIRONMENT_ARTIFACT_CONSTRAINT.UID`, might want to name this more generically, such as `getConstraintStateById`.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/InteractiveNotificationCallbackController.kt
```diff
@@ -0,0 +1,45 @@
...
+) {
+  @PostMapping(
+    consumes = [MediaType.APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
+    produces = [MediaType.APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  // TODO: This should be validated against write access to a service account. Service accounts should
+  //  become a top-level property of either delivery configs or environments.
+  // TODO(lfp): Related to the above, we'll need an additional authentication method for interactive constraint
+  //  approval outside of the Spinnaker UI, e.g. in Slack, since X-SPINNAKER-USER will be extracted from the Slack
+  //  message and not provided by the UI. My plan is to include an OTP in the callback URL.
```
asher: A related TODO for the future, we need to validate that the account specified by `X-SPINNAKER-USER` (or future methods) actually has permissions to the environment the constraint is on.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: Yeah, absolutely. Will rename the method and the argument! 👍

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/InteractiveNotificationCallbackController.kt
luispollo: On the need for an OTP here, I was discussing with @anotherchrisberry and @ajordens earlier today and, given that we enforce logins for Slack through Google, and that Google enforces  MFA for our accounts, perhaps we could leave the OTP bit as a `TODO` for later? Thoughts?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryDeliveryConfigRepository.kt
```diff
@@ -91,6 +91,10 @@ class InMemoryDeliveryConfigRepository(
   ): ConstraintState? =
     constraints["$deliveryConfigName:$environmentName:$artifactVersion:$type"]
 
+  override fun getConstraintStateById(uid: String): ConstraintState? {
+    TODO("not implemented") // To change body of created functions use File | Settings | File Templates.
```
emjburns: nit: remove this comment please :)

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ManualJudgementNotifier.kt
```diff
@@ -0,0 +1,73 @@
...
+  fun constraintStateChanged(event: ConstraintStateChanged) {
+    // if this is the first time the constraint was evaluated, send a notification
+    // so the user can react via other interfaces outside the UI (e.g. e-mail, Slack)
+    if (event.constraint is ManualJudgementConstraint &&
+      event.previousState == null &&
+      event.currentState.status == ConstraintStatus.PENDING) {
+      event.environment.notifications.map {
+        // TODO: run in parallel
+        runBlocking {
+          echoService.sendNotification(
```
emjburns: This is a lot of configuration for the notification wrapped up in one function. Do you think it's worth pulling it out into a separate function? Something like `fun Notification.toEchoNotification(event: ConstraintStateChanged)` or something? 

I always like to split up the creation of the object when the initialization has logic in it.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/InteractiveNotificationCallbackController.kt
```diff
@@ -0,0 +1,51 @@
...
+  private val deliveryConfigRepository: DeliveryConfigRepository
+) {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  @PostMapping(
+    consumes = [MediaType.APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
+    produces = [MediaType.APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  // TODO: This should be validated against write access to a service account. Service accounts should
+  //  become a top-level property of either delivery configs or environments.
```
emjburns: All the pre-work has been done for this! We could make that change at any time.
Would probably also make sense to add an api version for delivery configs around the same time, so we can version them?

luispollo: Yeah, I'm totally in favor of adding an API "header" section to the delivery config! On the service account access check, I seem to recall you mentioned we already had an example of that somewhere?

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ManualJudgementNotifier.kt
luispollo: Sounds good. Will make that change.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryDeliveryConfigRepository.kt
luispollo: Sure. I actually need to implement the method, possibly it in tests.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/InteractiveNotificationCallbackController.kt
emjburns: Yep: it's in the `ResourceController` [here](https://github.com/spinnaker/keel/blob/master/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt#L102)

asher: I think it would be better to resolve this TODO and its companion in `DeliveryConfigController` after `serviceAccount` is moved to the delivery config level.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
emjburns: should the type here be UID?

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ManualJudgementNotifier.kt
```diff
@@ -0,0 +1,76 @@
...
+      event.environment.notifications.map {
+        // TODO: run in parallel
+        runBlocking {
+          echoService.sendNotification(event.toEchoNotification(it))
+        }
+      }
+    }
+  }
+
+  private fun ConstraintStateChanged.toEchoNotification(config: NotificationConfig): EchoNotification {
```
emjburns: We got feedback around the content of this message yesterday - are you planning to update this for this PR, or leave it for a future PR?

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/TaskLauncher.kt
```diff
@@ -96,11 +96,11 @@ class TaskLauncher(
         Task(id = it.taskId, name = description)
       }
 
-  private val Resource<*>.notifications: List<EchoNotification>
+  private val Resource<*>.notifications: List<OrcaNotification>
     get() = deliveryConfigRepository
       .environmentFor(id)
-      .notifications
-      .map { it.toEchoNotification() }
+      ?.notifications
```
emjburns: why is this `?` now? I don't think that environment for `id` should ever be null - resources must be in an environment. Is the signature wrong on that method? (i'm wondering if I missed it when making delivery configs required)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -288,6 +288,8 @@ class SqlDeliveryConfigRepository(
             .onDuplicateKeyUpdate()
             .set(CURRENT_CONSTRAINT.CONSTRAINT_UID, MySQLDSL.values(CURRENT_CONSTRAINT.CONSTRAINT_UID))
             .execute()
+
+          state.uid = uid
```
emjburns: does this do anything? I don't see state being used again in this method

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/InteractiveNotificationCallbackController.kt
emjburns: not suggesting the change should be done in this PR, I am just excited that the pre-work is done for this todo!

luispollo: @asher I agree it should be checked in `DeliveryConfigController`, but since this is a separate REST endpoint altogether, don't we need to add that check here as well? In any case, I'd live to leave it as a `TODO` for a future PR. 😄

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ManualJudgementNotifier.kt
luispollo: Future, pretty please! Would like to get a first version of this deployed so more people can play around with it. The message should be easy to change later.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ConstraintState.kt
luispollo: Ah, good catch. I didn't notice that type existed. Will take a look.

emjburns:
LGTM, some small comments here and there

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/InteractiveNotificationCallbackController.kt
asher: Yes, we need to do the check here. I think the TODO here was actually added after I made a comment to that effect in an earlier thread. Just meant to add that they should be resolved together but not in this PR.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/TaskLauncher.kt
luispollo: Honestly I don't know why I added those. Removing.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: Yes, it updates the `uid` in the original object passed in, so it can be used by the caller. I use that to generate a unique callback ID to send to echo, which later gets sent back to keel and allows us to retrieve the corresponding `ConstraintState` to update with the user's approval/rejection.

luispollo:
@emjburns https://github.com/spinnaker/echo/pull/744 has been approved and merged. Any last minute requests or can I get a ✅ here too?

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ManualJudgementNotifier.kt
```diff
@@ -54,23 +60,27 @@ class ManualJudgementNotifier(
...
       ),
-      interactiveActions = EchoNotification.InteractiveActions(
-        callbackServiceId = "keel",
-        callbackMessageId = currentState.uid?.toString() ?: error("ConstraintState.uid not present"),
-        actions = listOf(
-          EchoNotification.ButtonAction(
-            name = "manual-judgement",
-            label = "Approve",
-            value = ConstraintStatus.OVERRIDE_PASS.name
+      interactiveActions = if (dynamicConfigService.isEnabled(INTERACTIVE_NOTIFICATIONS_ENABLED, false)) {
```
luispollo: I made this configurable and set to `false` by default so we can get this merged while we figure out what to do about the public API to handle Slack notification callbacks.

emjburns: Any reason you did this in a fast property vs a config value? For this kind of thing I would prefer controls in both, with a preference for the property to be read from a config value (because it's more similar to how we do other things right now).

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -453,6 +457,52 @@ class SqlDeliveryConfigRepository(
     }
   }
 
+  override fun getConstraintStateById(uid: UID): ConstraintState? {
+    return jooq
```
emjburns: This should have the new sql retries

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ManualJudgementNotifier.kt
luispollo: Sorry, I don't know what I was thinking... For some reason I thought `dynamicConfigService` defaulted to reading config from the config files. And I can't even say it was the first config property I added. Will fix! 🤦‍♂

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: Will do!

PR #724, opened 2020-01-15
chore(deb): add debug logs

emjburns:
- Adding some debugging for images.
- Removing a no longer relevant todo.
- Fixing the formatting for debians when they are returned from the api.

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
@@ -123,21 +123,48 @@ class ImageService(
    * Returns the latest image that is present in all regions.
    * Each ami must have tags.
    */
-  suspend fun getLatestNamedImageWithAllRegions(packageName: String, account: String, regions: List<String>): NamedImage? =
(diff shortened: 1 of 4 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -123,21 +123,45 @@ class ImageService(
...
+    val image = filteredImages
       .find {
+        var errors = ""
         val curAppVersion = AppVersion.parseName(it.appVersion)
-        curAppVersion.packageName == packageName &&
-          it.accounts.contains(account) &&
-          it.amis.keys.containsAll(regions) &&
-          tagsExistForAllAmis(it.tagsByImageId)
+        if (curAppVersion.packageName != packageName) {
+          errors += "[package name ${curAppVersion.packageName} does not match required package]"
```
luispollo: Would it be best perhaps to make this a list and add each message as an item and then `.join("\n")` at the end so that they're not all in one line?

emjburns: 😆 probably. I'll do that

Issue #712, opened 2020-01-10
make titus work with "depends-on" constraint


emjburns:
Right now, for titus clusters, we never emit the "deployed" event: https://github.com/spinnaker/keel/blob/master/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt#L351

This means that a version is never marked deployed, and we don't promote it

PR #728, opened 2020-01-15
feat(titus): depends-on constratint support

emjburns:
Closes https://github.com/spinnaker/keel/issues/712.

While I was fixing the bug I created for ec2 reference artifact promotion I also implemented titus support.

### DIFF
--- a/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/image/CurrentlyDeployedDockerImageApprover.kt
+++ b/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/image/CurrentlyDeployedDockerImageApprover.kt
@@ -29,8 +29,7 @@ class CurrentlyDeployedDockerImageApprover(
       val deliveryConfig = deliveryConfigRepository.deliveryConfigFor(resourceId)
       val env = deliveryConfigRepository.environmentFor(resourceId)
 
-      if (resource.spec is TitusClusterSpec) {
-        val spec = resource.spec as TitusClusterSpec // needed because kotlin can't cast it automatically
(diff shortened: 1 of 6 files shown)


keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/image/CurrentlyDeployedDockerImageApprover.kt
```diff
@@ -0,0 +1,65 @@
...
+  @EventListener(ArtifactVersionDeployed::class)
+  fun onArtifactVersionDeployed(event: ArtifactVersionDeployed) =
+    runBlocking {
+      val resourceId = event.resourceId
+      val resource = resourceRepository.get(resourceId)
+      val deliveryConfig = deliveryConfigRepository.deliveryConfigFor(resourceId)
+      val env = deliveryConfigRepository.environmentFor(resourceId)
+
+      if (resource.spec is TitusClusterSpec) {
+        val spec = resource.spec as TitusClusterSpec // needed because kotlin can't cast it automatically
```
robfletcher: If you swapped the statements it would be able to, i.e.

```
val spec = resource.spec
if (spec is TitusClusterSpec) {
```

It can't do it this way because `spec` is defined on an interface so the compiler can't tell what kind of trickery might be involved in returning that value and whether it's guaranteed to be the same every time.

Another way is to use the `as?` operator that will return `null` if the type cast fails:

```
(resource.spec as? TitusClusterSpec)?.let { spec ->
```

emjburns: Nice! I'll do that in the ec2 version as well.

PR #730, opened 2020-01-15
Optionally specify locations at the environment level

### LINKED ISSUE #445: Resources should be able to inherit account/regions from their environment
i.e. omit `account` and/or `region` fields from resources contained in an environment and they get defaulted to what is set in the environment. In the case of region, that would mean multiple resources get created, one for each region.

It's possible this would require us to create an EC2 specific `Environment` implementation as account / region may not be applicable to all cloud providers.

robfletcher:
This allows all resources in an environment to omit their `locations` property in favor of the one set at the environment level. It relies on some Jackson JSON parsing trickery. The nice part is that the `locations` property on the resource spec does _not_ need to be nullable, and we don't need to keep looking up the environment in order to fetch the locations -- instead it's all handled when parsing the submitted delivery config.

Closes #445

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryConfig.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryConfig.kt
@@ -1,6 +1,8 @@
 package com.netflix.spinnaker.keel.api
 
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
 import com.netflix.spinnaker.keel.persistence.NoMatchingArtifactException
+import com.netflix.spinnaker.keel.serialization.SubmittedEnvironmentDeserializer
 
 const val DEFAULT_SERVICE_ACCOUNT = "keel@spinnaker.io"
 
@@ -18,11 +20,16 @@ data class SubmittedDeliveryConfig(
   val environments: Set<SubmittedEnvironment> = emptySet()
 )
 
+@JsonDeserialize(using = SubmittedEnvironmentDeserializer::class)
 data class SubmittedEnvironment(
   val name: String,
   val resources: Set<SubmittedResource<*>>,
   val constraints: Set<Constraint> = emptySet(),
-  val notifications: Set<NotificationConfig> = emptySet()
+  val notifications: Set<NotificationConfig> = emptySet(),
+  /**
+   * Optional locations that are propagated to any [resources] where they are not specified.
+   */
+  val locations: SubnetAwareLocations? = null
 )
 
 val DeliveryConfig.resources: Set<Resource<*>>
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
@@ -17,10 +17,13 @@
  */
 package com.netflix.spinnaker.keel.api
 
+import com.fasterxml.jackson.annotation.JacksonInject
+
 /**
  * A resource spec which is located in an account and one or more regions.
  */
 interface Locatable<T : Locations<*>> : ResourceSpec {
+  @get:JacksonInject("locations")
   val locations: T
 }
 
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/DeserializationContexts.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/DeserializationContexts.kt
@@ -1,0 +1,12 @@
+package com.netflix.spinnaker.keel.serialization
+
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.ObjectMapper
+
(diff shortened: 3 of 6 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryConfig.kt
```diff
@@ -16,11 +19,16 @@ data class SubmittedDeliveryConfig(
...
 data class SubmittedEnvironment(
   val name: String,
   val resources: Set<SubmittedResource<*>>,
   val constraints: Set<Constraint> = emptySet(),
-  val notifications: Set<NotificationConfig> = emptySet()
+  val notifications: Set<NotificationConfig> = emptySet(),
+  /**
+   * Optional locations that are propagated to any [resources] where they are not specified.
+   */
+  val locations: SubnetAwareLocations? = null
```
robfletcher: New property added on `SubmittedEnvironment`. Note there isn't a corresponding one on `Environment`, we just throw it away after applying to the resources.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/Locatable.kt
```diff
@@ -17,10 +17,13 @@
...
 
+import com.fasterxml.jackson.annotation.JacksonInject
+
 /**
  * A resource spec which is located in an account and one or more regions.
  */
 interface Locatable<T : Locations<*>> : ResourceSpec {
+  @get:JacksonInject("locations")
   val locations: T
 }
```
robfletcher: Adding this `@JacksonInject` annotation means that if there is no corresponding data in the JSON being parsed, Jackson will look in the deserialization context for "injected" values using this key.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/DeserializationContexts.kt
```diff
@@ -0,0 +1,12 @@
...
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.ObjectMapper
+
+internal val DeserializationContext.mapper: ObjectMapper
+  // Such an intuitive API. ObjectMapper is the only thing that implements ObjectCodec but useful
+  // methods like convertValue are not on that interface so we need to do a dirty type cast.
+  get() = parser.codec as ObjectMapper
+
+internal inline fun <reified T> DeserializationContext.instantiationException(cause: Throwable) =
+  instantiationException(T::class.java, cause)
```
robfletcher: These just moved out of another file because I'm now using them in 2 places.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/SubmittedEnvironmentDeserializer.kt
```diff
@@ -0,0 +1,62 @@
...
+      val resources = try {
+        mapper.copy().let { nestedMapper ->
+          nestedMapper.injectableValues = InjectableValues.Std(mapOf("locations" to locations))
+          nestedMapper
+            .convertValue<Set<SubmittedResource<*>>?>(root.path("resources"))
+            ?: emptySet()
+        }
+      } catch (e: IllegalArgumentException) {
+        throw JsonMappingException.wrapWithPath(e, root, "resources")
+      }
```
robfletcher: Here's the interesting bit. I read the `locations` property (if there is one) on the `SubmittedEnvironment` then create a nested `ObjectMapper` with that injected value and read the set of `SubmittedResource`s. The `locations` are only in scope while I'm doing that and don't leak out into the wider deserialization context.

emjburns:
I think this is a crazy and cool idea. LGTM

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/SubmittedEnvironmentDeserializer.kt
robfletcher: I've tidied the code up a bit now so there are no so many try/catch blocks everywhere. Intent remains identical.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/serialization/SubmittedEnvironmentDeserializerTests.kt
```diff
@@ -0,0 +1,175 @@
...
+          |  subnet: internal (vpc0)
+          |  vpc: vpc0
+          |  regions:
+          |  - name: us-east-1
+          |  - name: us-west-2
+          """.trimMargin()
+        )
+      }
+
+      test("locations corresponds to what was specified on the resource") {
```
luispollo: “...to what was specified on the parent environment”?

robfletcher: Good catch. Bit of copy 'n paste there.

Issue #444, opened 2019-09-13
handle a change to delivery config where a resource is removed or renamed


cfieber:
If a resource is removed from an environment (or edited such that it lands with a new resource id, e.g. changed the stack name), we should remove the declarative config from that resource from keel

currently we end up with the original resource still managed, but orphaned from its delivery config.

Issue #702, opened 2020-01-07
Delete Environments from delivery configs


emjburns:
You should be able to delete an environment from a delivery config

Issue #679, opened 2019-12-17
Delete Delivery configs


emjburns:
We need a way to delete delivery configs.

Does deleting one delete all the configs for the resources inside it? Probably.
Does it actually delete the infra? Probably not.

PR #731, opened 2020-01-17
feat(deliveryConfig): delete and update configs

emjburns:
Closes https://github.com/spinnaker/keel/issues/444, https://github.com/spinnaker/keel/issues/679, and https://github.com/spinnaker/keel/issues/702.

Adding functionality to delete a delivery config.
Adding functionality to edit a delivery config : for example, rename environments, remove resources, remove artifacts, delete environments.
Switched the delete-by-application endpoint over to just deleting all the delivery configs in an app, because we can no longer have lone resources.

I created a class `Cleaner` to handle the removing because it has to be orchestrated between three repositories. 

Feedback desired: is this approach okay? Are there areas which could be improved?


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/Cleaner.kt
```diff
@@ -0,0 +1,96 @@
...
+@Component
+class Cleaner(
+  private val deliveryConfigRepository: DeliveryConfigRepository,
+  private val artifactRepository: ArtifactRepository,
+  private val resourceRepository: ResourceRepository,
+  private val clock: Clock,
+  private val publisher: ApplicationEventPublisher
+) {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
```
emjburns: Deleting and removing things from a delivery config happens here.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
```diff
@@ -33,7 +32,9 @@ class AdminController(
     @PathVariable("application") application: String
   ) {
     log.debug("Deleting all data for application: $application")
-    deleteApplication(application)
+    deliveryConfigRepository.getByApplication(application).forEach { config ->
```
emjburns: updated our admin controller to just delete all delivery configs

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
```diff
@@ -1,95 +0,0 @@
-package com.netflix.spinnaker.keel.rest
```
emjburns: Removed because we're no longer using the same methods on the admin controller, so this isn't relevant. There are tests for the `Cleaner` methods, so no need to update the tests here.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -129,6 +131,21 @@ class SqlArtifactRepository(
       .execute() == 1
   }
 
+  override fun delete(artifact: DeliveryArtifact) {
```
emjburns: should this be in a `jooq.transaction`?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -88,6 +90,51 @@ class SqlDeliveryConfigRepository(
     return deliveryConfigUIDs.size
   }
 
+  override fun delete(name: String) {
```
emjburns: should this be in a `jooq.transaction`?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CleanerTests.kt
```diff
@@ -0,0 +1,158 @@
...
+import io.mockk.mockk
+import java.time.Clock
+import org.springframework.context.ApplicationEventPublisher
+import strikt.api.expect
+import strikt.api.expectThat
+import strikt.api.expectThrows
+import strikt.assertions.isEmpty
+import strikt.assertions.isEqualTo
+
+abstract class CleanerTests<T : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository> :
```
luispollo: Might as well go with a `D` here for the `DeliveryConfigRepository` type for consistency... 😄

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CleanerTests.kt
```diff
@@ -0,0 +1,158 @@
...
+        before {
+          val updatedConfig = deliveryConfig.copy(
+            artifacts = setOf(newArtifact),
+            environments = setOf(firstEnv.copy(resources = setOf(secondResource)))
+          )
+          persist(updatedConfig)
+          subject.removeResources(old = deliveryConfig, new = updatedConfig)
+        }
+
+        test("no longer present resources are removed") {
```
luispollo: Add the opposite test to check that the new resources were _not_ inadvertently removed?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/Cleaner.kt
```diff
@@ -0,0 +1,96 @@
...
+import com.netflix.spinnaker.keel.events.ResourceDeleted
+import org.slf4j.LoggerFactory
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.stereotype.Component
+import org.springframework.transaction.annotation.Propagation.REQUIRED
+import org.springframework.transaction.annotation.Transactional
+import java.time.Clock
+
+/**
+ * A class used to orchestrate the deletion of delivery configs
```
luispollo: "...and associated resources"?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/Cleaner.kt
```diff
@@ -0,0 +1,96 @@
...
+  private val resourceRepository: ResourceRepository,
+  private val clock: Clock,
+  private val publisher: ApplicationEventPublisher
+) {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  /**
+   * Deletes a delivery config and everything in it.
+   */
+  @Transactional(propagation = REQUIRED)
```
luispollo: Does this automagically take care of running all the methods that touch the DB in the same transaction?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/Cleaner.kt
```diff
@@ -0,0 +1,96 @@
...
+            log.debug("Updating config ${new.name}: removing resource ${resource.id} in environment ${environment.name}")
+            deliveryConfigRepository.deleteResourceFromEnv(
+              deliveryConfigName = old.name, environmentName = environment.name, resourceId = resource.id
+            )
+            delete(resource.id)
+          }
+        }
+      }
+
+    old.environments.forEach { environment ->
```
luispollo: Why another loop for `old.environments` here?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/Cleaner.kt
```diff
@@ -0,0 +1,96 @@
...
+      }
+    }
+  }
+
+  fun delete(id: ResourceId): Resource<out ResourceSpec> =
+    resourceRepository
+      .get<ResourceSpec>(id)
+      .also {
+        resourceRepository.delete(id)
+        publisher.publishEvent(ResourceDeleted(it, clock))
```
luispollo: This is good. I think we're already missing events for operations on environments and delivery configs. Should we stretch this PR a bit and add a couple, at least for deletion events?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
```diff
@@ -45,6 +45,23 @@ interface DeliveryConfigRepository : PeriodicallyCheckedRepository<DeliveryConfi
    */
   fun deleteByApplication(application: String): Int
 
+  /**
+   * Deletes a delivery config, and the environments within in it.
```
luispollo: Hmm... I think it would be more intuitive if it was the other way around. I realize this would require `DeliveryConfigRepository` to hold references to `ArtifactRepository` and `ResourceRepository` to do this, but I wondered if that's considered an anti-pattern for some reason? From a logical modeling standpoint, it would seem much easier to wrap my brain around than having the separate `Cleaner` class coordinating this kind of thing.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: I think so!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: I think so. It's hard enough that we have to write all this jooq/sql code to cascade deletes and retrievals of object trees (which I feel is pretty prone to implementation mistakes), so I think we definitely need to wrap things in transactions when operating on multiple tables.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -88,6 +90,51 @@ class SqlDeliveryConfigRepository(
...
+      .from(DELIVERY_CONFIG)
+      .where(DELIVERY_CONFIG.NAME.eq(name))
+      .fetchOne(DELIVERY_CONFIG.UID)
+
+    jooq
+      .deleteFrom(DELIVERY_CONFIG)
+      .where(DELIVERY_CONFIG.NAME.eq(name))
+      .execute()
+    jooq
+      .deleteFrom(DELIVERY_CONFIG_LAST_CHECKED)
```
luispollo: Side-question: why do we have a separate table for this in the first place?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -88,6 +90,51 @@ class SqlDeliveryConfigRepository(
...
+      .where(ENVIRONMENT_RESOURCE.ENVIRONMENT_UID.eq(envUid(deliveryConfigName, environmentName)))
+      .and(ENVIRONMENT_RESOURCE.RESOURCE_UID.eq(resourceId.uid))
+      .execute()
+  }
+
+  override fun deleteEnvironment(deliveryConfigName: String, environmentName: String) {
+    val deliveryConfigUid = getDeliveryConfigId(deliveryConfigName)
+    jooq
+      .deleteFrom(ENVIRONMENT_ARTIFACT_CONSTRAINT)
+      .where(ENVIRONMENT_ARTIFACT_CONSTRAINT.ENVIRONMENT_UID.eq(envUid(deliveryConfigName, environmentName)))
```
luispollo: Could move `envUid(deliveryConfigName, environmentName)` above the jooq methods so you don't call it repeatedly.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -121,7 +173,13 @@ class SqlDeliveryConfigRepository(
...
         jooq.insertInto(DELIVERY_CONFIG_ARTIFACT)
           .set(DELIVERY_CONFIG_ARTIFACT.DELIVERY_CONFIG_UID, uid)
-          .set(DELIVERY_CONFIG_ARTIFACT.ARTIFACT_UID, jooq.select(DELIVERY_ARTIFACT.UID).from(DELIVERY_ARTIFACT).where(DELIVERY_ARTIFACT.NAME.eq(artifact.name)))
+          .set(DELIVERY_CONFIG_ARTIFACT.ARTIFACT_UID, jooq
+            .select(DELIVERY_ARTIFACT.UID)
+            .from(DELIVERY_ARTIFACT)
+            .where(DELIVERY_ARTIFACT.NAME.eq(artifact.name))
+            .and(DELIVERY_ARTIFACT.TYPE.eq(artifact.type.name))
+            .and(DELIVERY_ARTIFACT.DELIVERY_CONFIG_NAME.eq(artifact.deliveryConfigName))
+            .and(DELIVERY_ARTIFACT.REFERENCE.eq(artifact.reference)))
```
luispollo: Good catch! Thanks for adding this! Just one question: I thought not all artifacts had a `reference`?

emjburns: All artifacts have a reference, it's a requirement. It defaults to the name.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
emjburns: because that information changes a lot more frequently that the delivery configs, and it is a different type of information (cc @robfletcher ?)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/Cleaner.kt
emjburns: Oops, should have moved this logic up

emjburns:
@luispollo re your comment on a `Cleaner` class vs implementing this in the `DeliveryConfigRepository` (discussed offline, summarizing here): deletion needs to coordinate between 3 repositories (a design discussion for the future), and I chose to create the cleaner class so that all the logic was easily testable and easy to rip out. It's hidden within the `ResourcePersister`. I couldn't just modify the resource persister because I wanted to create test that would use both the in memory implementations and the sql implementations, and those weren't visible from the module the `ResourcePersister` is in.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
asher: yeah, the current implementation is more efficient than continually updating all rows on `resource` and `delivery_config`.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
asher: Agree, since envUid() is a db subquery, it's not free to repeat.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: Got it, thanks.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/Cleaner.kt
luispollo: This might be a useful reference, BTW: https://stackoverflow.com/a/39998943

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
asher: Instead of one repository calling into another, I actually see no reason not to read/write directly to any table from from `DeliveryConfigRepository`. The thing I find useful about our current repository model is in how they enable periodic polling with a distributed lock per item polled. I think it makes sense if that's all a repository was, not an owner of tables.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
asher: Its very important that we add retry logic for every sql write as we implement multi-table transactions like this. They will result in deadlocks under the right sort of concurrency - that is normal and ok, but keel can't currently handle them at all.

emjburns: A PR for sql retries: https://github.com/spinnaker/keel/pull/741

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/Cleaner.kt
emjburns: Yes, it does. I've added some transactions to the other methods too

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/Cleaner.kt
emjburns: Let's not stretch this PR - no reason to.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
emjburns: Created a jira to track that larger change, which definitely needs to happen: https://github.com/spinnaker/keel/issues/740

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -105,6 +107,77 @@ class SqlDeliveryConfigRepository(
...
+
+  override fun deleteEnvironment(deliveryConfigName: String, environmentName: String) {
+    val deliveryConfigUid = sqlRetry.withRetry(READ) {
+      jooq.select(DELIVERY_CONFIG.UID)
+        .from(DELIVERY_CONFIG)
+        .where(DELIVERY_CONFIG.NAME.eq(deliveryConfigName))
+        .fetchOne(DELIVERY_CONFIG.UID)
+    }
+    val envUid = sqlRetry.withRetry(READ) {
+      jooq.select(ENVIRONMENT.UID)
```
asher:
```suggestion
      envUid(deliveryConfigName, environmentName)
        .fetchOne(ENVIRONMENT.UID)
```

PR #732, opened 2020-01-17
chore(constraints): refactor updateConstraintStatus rest controller

asher:


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -93,21 +93,21 @@ class DeliveryConfigController(
     @RequestHeader("X-SPINNAKER-USER") user: String,
     @PathVariable("name") name: String,
     @PathVariable("environment") environment: String,
```
emjburns: do we need the delivery config name in here? env names arent guaranteed to be unique

emjburns: nvm im blind

PR #734, opened 2020-01-18
feat(ec2): flexible region staggering for clusters

asher:
This enables flexible staggering of region deployments for multi-region ec2 cluster resources. An ec2 cluster that deploys to regions [`A`, `B`, `C`] can declare a stagger as part of its deployment strategy, such as:
```
        stagger:
        - region: A
          pauseTime: PT1H
          hours: 10-15
        - region: B
          hours: 14-18
        - region: C
          hours: 20-02
```
In this scenario, keel will still submit an orca task per-region within the same actuation pass, just like an unstaggered cluster. Region `A` will deploy immediately if we're within the hours of 10am - 3pm in the orca "user" timezone (configurable but defaults to `America/Los_Angeles`, not the system tz) or will wait until 10am if not. After the cluster is done deploying in `A`, the task will wait for 1 hour. After the wait, the `B` deploy will start if within the hours of 2-6pm. It does not use `pauseTime`, since the `C` window does not begin until 8pm. Neither `pauseTime` or `hours` is required or mutually exclusive, but every stagger must use at least one. If only a subset of regions are in the stagger list, remaining regions will deploy in parallel after the completion of the final stagger.

Staggering is only applied for regional diffs that deploy a new serverGroup. Diffs that can be applied in place, such as scaling policies or simple capacity changes are currently applied immediately. 

This feature requires an Orca that is >= commit b966cc9.


robfletcher:
If a new artifact is released at 7pm would it deploy to `C` first?

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaExecutionUtils.kt
```diff
@@ -0,0 +1,59 @@
+package com.netflix.spinnaker.keel.orca
+
+import java.time.Duration
+
+class OrcaExecutionUtils {
+  companion object {
```
robfletcher: We could just put these functions at package level rather than needing a utils class

robfletcher:
Something just occurred to me: in the region staggering scenario would a user expect infrastructure changes and new versions of an artifact to be staggered in the same way? Do we have any desire to discriminate between those things?

asher:
> If a new artifact is released at 7pm would it deploy to `C` first?

No, ordering is currently enforced via the `dependsOnExecution` stage based on the order of the `stagger` list, so in this case the first server group change would occur the following day at 10am.

asher:
> Something just occurred to me: in the region staggering scenario would a user expect infrastructure changes and new versions of an artifact to be staggered in the same way? Do we have any desire to discriminate between those things?

Probably not under normal circumstances for our currently implemented infrastructure resource types. Steady-state infrastructure changes seem typically like "add an ingress to a firewall" or "create a new load balancer for a new service shard" which avoids all of the risk inherent in running new application code for the first time. There will probably come a time though when we'll want to stagger more than clusters.

robfletcher:
> Probably not under normal circumstances for our currently implemented infrastructure resource types. Steady-state infrastructure changes seem typically like "add an ingress to a firewall" or "create a new load balancer for a new service shard" which avoids all of the risk inherent in running new application code for the first time. There will probably come a time though when we'll want to stagger more than clusters.

I was actually thinking of the reverse scenario were you might want to roll out capacity changes or something instantly while staggering the release of new artifact versions.

asher:
> I was actually thinking of the reverse scenario were you might want to roll out capacity changes or something instantly while staggering the release of new artifact versions.

That's actually how this PR behaves, only changes that create a new server group are staggered. Changes that can be made in-place (capacity, suspend processes, or scaling policies) which can't deploy a new artifact are still applied immediately.

robfletcher:
> That's actually how this PR behaves, only changes that create a new server group are staggered. Changes that can be made in-place (capacity, suspend processes, or scaling policies) which can't deploy a new artifact are still applied immediately.

Aha, I missed the implication of that in the code. But something like an instance type change that requires a clone ASG operation would still stagger, right?

asher:
> Aha, I missed the implication of that in the code. But something like an instance type change that requires a clone ASG operation would still stagger, right?

Yeah, an instance type change would still stagger here.

PR #737, opened 2020-01-21
Gradle 6.1

robfletcher:


robfletcher:
I'll revisit later, needs changes to spinnaker-gradle-project

PR #738, opened 2020-01-21
Fix ability to inherit non subnet aware locations from environment

robfletcher:


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/SubmittedEnvironmentDeserializer.kt
```diff
@@ -27,7 +30,7 @@ class SubmittedEnvironmentDeserializer : StdNodeBasedDeserializer<SubmittedEnvir
       val notifications: Set<NotificationConfig> = convert(root, "notifications") ?: emptySet()
       val locations: SubnetAwareLocations? = convert(root, "locations")
       val resources: Set<SubmittedResource<*>> = copy().run {
-        injectableValues = InjectableValues.Std(mapOf("locations" to locations))
+        injectableValues = InjectableLocations(locations)
```
robfletcher: Fixed by creating my own implementation of this `InjectableValues` thing that can detect the desired type

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/SubmittedEnvironmentDeserializer.kt
```diff
@@ -44,3 +47,21 @@ class SubmittedEnvironmentDeserializer : StdNodeBasedDeserializer<SubmittedEnvir
...
+    valueId: Any,
+    context: DeserializationContext,
+    forProperty: BeanProperty,
+    beanInstance: Any?
+  ): Any? {
+    val value = super.findInjectableValue(valueId, context, forProperty, beanInstance) as? SubnetAwareLocations
+    return when {
+      value == null -> null
+      forProperty.type.isTypeOrSubTypeOf(SimpleLocations::class.java) -> value.toSimpleLocations()
+      else -> value
```
robfletcher: Converts the `locations` value if necessary

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/SubmittedEnvironmentDeserializer.kt
asher: Do we lose any flexibility here for injecting additional (non-location) values in the same way?

robfletcher: Yeah, we'd have to amend this as & when we do. If that's the case we may want an extension that allows for an optional conversion based on context.

PR #742, opened 2020-01-22
feat(configs): add api version to delivery configs

emjburns:
Adding api version to delivery configs so that we can make versioned changes to delivery configs. Not required for submission, we will default the value to v1.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryConfig.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryConfig.kt
@@ -11,7 +11,8 @@ data class DeliveryConfig(
   val application: String,
   val serviceAccount: String,
   val artifacts: Set<DeliveryArtifact> = emptySet(),
-  val environments: Set<Environment> = emptySet()
+  val environments: Set<Environment> = emptySet(),
+  val apiVersion: ApiVersion = ApiVersion("delivery.config.spinnaker.netflix.com", "v1")
 )
 
 data class SubmittedDeliveryConfig(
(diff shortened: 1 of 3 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryConfig.kt
```diff
@@ -10,7 +10,8 @@ data class DeliveryConfig(
   val name: String,
   val application: String,
   val artifacts: Set<DeliveryArtifact> = emptySet(),
-  val environments: Set<Environment> = emptySet()
+  val environments: Set<Environment> = emptySet(),
+  val apiVersion: ApiVersion = ApiVersion("delivery.config.spinnaker.netflix.com", "v1")
```
gal-yardeni: question - is this value should be hardcoded? what should happen if we'll change to v2?

emjburns: That's just the default value. What that means is that the existing configs won't have to update - they'll automatically get the first version. But when we change in the future we will have some way of indicating it (this field) and processing it differently (not built yet).

asher: What's the likely mechanism of supporting old versions in a backwards compatible way? Should the DeliveryConfig class name itself be versioned and of a generic type?

emjburns: Great question - I'm not sure yet. I was tentatively thinking that there would be something that would convert the old config into the new version, save it as the new version, and return that back to the submitter.

PR #743, opened 2020-01-22
feat(titus): cluster region staggering

asher:


keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -112,31 +116,165 @@ class TitusClusterHandler(
...
+            description = "Upsert server group ${diff.desired.moniker.name} in " +
+              "${diff.desired.location.account}/${diff.desired.location.region}",
+            correlationId = "${resource.id}:${diff.desired.location.region}",
+            job = job
+          )
+        }
+      }
+    }
+
+  private suspend fun upsertStaggered(
```
emjburns: Do you think any of this could be generalized or shared between any resource that wants to do a staggered rollout? This looks similar to the code in the ec2 cluster handler.

asher: I'd love suggestions or ideas!

emjburns: What I'm thinking is sort of along the lines of what I did for the Docker image resolver ([here](https://github.com/spinnaker/keel/blob/master/keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/DockerImageResolver.kt)). 

So, maybe we have a base class in `keel-plugin` that helps orchestrate things. Each handler that wanted to use it could implement some functions, like a function to get `isStaggered` from their own spec type.

Then there could be a function that takes the `create` diffs and the `modify` diffs, submits things to orca, and returns the task list?

asher: Interesting, I'll close this PR.

PR #756, opened 2020-01-27
use appropriate HTTP response code when delivery config is unparseable

robfletcher:


robfletcher:
> Since you're already taking a look at this - can we always send the error response back with the application json header so that it works with the gate error handling?

That's a non-trivial change. We need to come up with a standard way of representing error responses. I'm happy to take that on but it feels like it's outside the remit of this change which is to address a very specific issue.

PR #758, opened 2020-01-27
refactor(plugins): Decouple jackson from core models

robzienert:
A little experiment around decoupling Jackson annotations from the core models. This would allow us to, for the most part, move existing models into a plugin API without bringing Jackson along for the ride!

Uses [Jackson Mixin Annotations](https://github.com/FasterXML/jackson-docs/wiki/JacksonMixInAnnotations), which is just an abstract class that you define with all of the constructors & methods that you need to put annotations on. Once the mixin class is created, an association is made between the mixin and the actual target class, voila. Nice thing is that it supports all annotations that Jackson knows about, so I believe things should "just work".

There's a little more boilerplate with this, but it's definitely less interruptive than other options I've found so far.

I see two initial drawbacks on this:

1. Boilerplate. For any class you want to do customization of serialization for, you'll need a mixin class.
1. Discoverability. This is a problem no matter the solution, I think, but with decoupling serde from the actual model, the codebase may be more difficult to reason about for new developers.

I went down the path of writing a helper to reduce boilerplate via the `@MixinFor` annotation which would automate the mapping of mixins to their target class, but I couldn't get it to work for a quick POC. What exists here works, however.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/mixins/ApiVersionMixin.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/mixins/ApiVersionMixin.kt
@@ -1,11 +1,11 @@
 /*
- * Copyright 2018 Netflix, Inc.
+ * Copyright 2020 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *    http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/mixins/ApiVersionMixin.kt
```diff
@@ -0,0 +1,37 @@
...
+@MixinFor(ApiVersion::class)
+@JsonSerialize(using = ToStringSerializer::class)
+abstract class ApiVersionMixin(
+  val group: String,
+  val version: String
+) {
+
+  @JsonCreator
+  constructor(value: String) :
+    this(value.substringBefore("/"), value.substringAfter("/"))
```
robfletcher: This confuses me. We're constructing the mixin here, but this looks like constructing the actual type. What's going on?

robzienert: I have no idea how the internals work, but Jackson is somehow capable of defining constructors on the mixin and applying it to the actual type, like you're guessing. Not sure how it actually gets the logic of this constructor applied to the type, but it does work. 🤷‍♂

robfletcher:
Closing this as I've run with something very similar in the recent changes to the model classes exposed via `keel-api`

PR #765, opened 2020-01-28
Start of better REST error handling

robfletcher:


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -0,0 +1,61 @@
...
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+}
+
+/**
+ * Error details returned as JSON/YAML.
+ */
+data class ApiError(val message: String?) {
+  constructor(ex: Throwable) : this(ex.cause?.message ?: ex.message)
+}
```
robfletcher: My intention is that we could add further fields to this type to do a better job of providing specific detail for certain types of error. This is a start, tho.

PR #766, opened 2020-01-28
feat(environments): ability to pin artifact versions to environments

asher:


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -221,18 +242,27 @@ class SqlArtifactRepository(
...
+     * the pinned version. Otherwise return the most recently approved version.
+     */
+    return sqlRetry.withRetry(READ) {
+      jooq.select(ENVIRONMENT_ARTIFACT_PIN.ARTIFACT_VERSION)
+        .from(ENVIRONMENT_ARTIFACT_PIN)
+        .where(
+          ENVIRONMENT_ARTIFACT_PIN.ENVIRONMENT_UID.eq(envUid),
+          ENVIRONMENT_ARTIFACT_PIN.ARTIFACT_UID.eq(artifactId))
+        .fetchOne(ENVIRONMENT_ARTIFACT_PIN.ARTIFACT_VERSION)
+        ?: jooq
```
asher: This change to `latestVersionApprovedIn` ensures cluster resources use a pinned artifact version, if present, when calculating desired state. This is applicable for `DockerImageResolver` and the EC2 `ImageResolver`, with the exception of Jenkins Artifacts which are only supported by the EC2 `ImageResolver`.  I haven't thought through pinning for jenkins artifacts yet.

keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -30,17 +35,23 @@ class EnvironmentPromotionChecker(
...
-                /**
-                 * Only check stateful evaluators if all stateless evaluators pass. We don't
-                 * want to request judgement or deploy a canary for artifacts that aren't
-                 * deployed to a required environment or outside of an allowed time.
-                 */
-                checkConstraints(statelessEvaluators, artifact, deliveryConfig, v, environment) &&
-                  checkConstraints(statefulEvaluators, artifact, deliveryConfig, v, environment)
+            val version = when {
+              pinnedEnvs.hasPinFor(environment.name, artifact) -> {
+                pinnedEnvs.versionFor(environment.name, artifact)
```
asher: Constraint evaluators are bypassed for pinned artifact versions.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -433,6 +463,107 @@ class SqlArtifactRepository(
...
+      val artifact = get(deliveryConfigName, reference, ArtifactType.valueOf(type.toUpperCase()))
+
+      sqlRetry.withRetry(WRITE) {
+        jooq.insertInto(ENVIRONMENT_ARTIFACT_PIN)
+          .set(ENVIRONMENT_ARTIFACT_PIN.ENVIRONMENT_UID, deliveryConfig.getUidFor(environment))
+          .set(ENVIRONMENT_ARTIFACT_PIN.ARTIFACT_UID, artifact.uid)
+          .set(ENVIRONMENT_ARTIFACT_PIN.ARTIFACT_VERSION, version)
+          .set(ENVIRONMENT_ARTIFACT_PIN.PINNED_AT, clock.millis())
+          .set(ENVIRONMENT_ARTIFACT_PIN.PINNED_BY, pinnedBy ?: "anonymous")
+          .set(ENVIRONMENT_ARTIFACT_PIN.COMMENT, comment)
```
emjburns: noob sql question: do we need to handle duplicate keys here, and update the time/pinned by / comment?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -433,6 +463,107 @@ class SqlArtifactRepository(
...
+    sqlRetry.withRetry(WRITE) {
+      jooq.select(ENVIRONMENT_ARTIFACT_PIN.ENVIRONMENT_UID, ENVIRONMENT_ARTIFACT_PIN.ARTIFACT_UID)
+        .from(ENVIRONMENT_ARTIFACT_PIN)
+        .innerJoin(ENVIRONMENT)
+        .on(ENVIRONMENT_ARTIFACT_PIN.ENVIRONMENT_UID.eq(ENVIRONMENT.UID))
+        .where(
+          ENVIRONMENT.NAME.eq(targetEnvironment),
+          ENVIRONMENT.DELIVERY_CONFIG_UID.eq(deliveryConfig.uid))
+        .fetch { (envUid, artUid) ->
+          deletePin(envUid, artUid)
```
emjburns: noob sql question: do we need transactions here, or is this already going to happen together? or no need to worry about that since it's a read then a write?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
asher: Sure do! Great catch.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
asher: There's no advantage to including the select in a txn (no impact on behavior, we run with a `READ_COMMITTED`, not `REPEATABLE_READ` txn isolation level) and I want to avoid a transaction around the deletes in the scenario where an environment has multiple artifacts pinned.

keel-sql/src/main/resources/db/changelog/20200127-environment-artifact-pin-table.yml
```diff
@@ -0,0 +1,56 @@
...
+                  constraints:
+                    nullable: false
+              - column:
+                  name: artifact_version
+                  type: varchar(255)
+                  constraints:
+                    nullable: false
+              - column:
+                  name: pinned_at
+                  type: bigint
```
robfletcher: would this be better as a `timestamp` like other columns in our schema?

asher: The `unhappy_veto` table introduced millisecond epoch timestamps and there was some discussion around time types then. We should have consistency but I'd rather see all of the `timestamp` columns move to epoch like other spinnaker services. I think it's problematic that the `timezone` type is zoned (and the mysql running on our laptops is likely running in local time while our aurora instances are in UTC) and we're better off avoiding use of sql datetime functions that that the `timezone` type enables. Let's discuss offline though, if we don't want to go thru the pain of migrating existing columns, it does make sense to change the type here.

robfletcher: Yeah, let's tackle it as a separate change if we already have inconsistency.

PR #768, opened 2020-01-30
Refactoring to support Spinnaker plugin model

robfletcher:
- [ ] `keel-api` should contain everything a plugin implementor would need
  - [ ] `DeliveryConfig` and `Environment`
  - [x] `Resource` and `ResourceSpec`
  - [x] `ResourceHandler`
  - [x] `Resolver`
  - [ ] `TaskLauncher` without direct dependency on `keel-orca`, i.e. just as an interface
- [ ] `keel-api` should have no dependencies on anything beyond Kotlin (for now)
- [ ] Everything in `keel-actuator` and `keel-plugin` should either be moved to `keel-api` or `keel-core`
- [ ] `keel-bakery-plugin`, `keel-ec2-plugin` and `keel-titus-plugin` should only depend on `keel-api` (this is likely to involve the most work)
- [ ] Fix up package names in `keel-api` so they don't conflict with other modules

The main things to consider are keeping Jackson, Java-object-diff, and any other libraries (I think those are the main ones that are likely to need some refactoring to remove from the crucial places) out of `keel-api`. 
We can revisit whether it's worth the effort to make `keel-api` pure Java (in order to remove the Kotlin dependency) later.

I think the `Submitted*` types don't need to be in `keel-api` which simplifies things a little as they are most likely to use Jackson annotations.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/diff/ResourceDiff.kt
```diff
@@ -0,0 +1,13 @@
...
+  val desired: T
+  val current: T?
+  val affectedRootPropertyTypes: List<Class<*>>
+  val affectedRootPropertyNames: Set<String>
+  fun hasChanges(): Boolean
+  fun toDeltaJson(): Map<String, Any?>
+  fun toUpdateJson(): Map<String, Any?>
+  fun toDebug(): String
+  fun T?.toMap(): Map<String, Any?>?
+}
```
robfletcher: Interface that is independent of java-object-diff

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/resource/ImageHandler.kt
```diff
@@ -40,12 +39,11 @@ class ImageHandler(
   private val imageService: ImageService,
   private val publisher: ApplicationEventPublisher,
   private val taskLauncher: TaskLauncher,
-  objectMapper: ObjectMapper,
```
robfletcher: No need to wire ObjectMapper into resource handlers any more

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceHandler.kt
```diff
@@ -155,11 +128,9 @@ abstract class ResourceHandler<S : ResourceSpec, R : Any>(
   open suspend fun delete(resource: Resource<S>): List<Task> = TODO("Not implemented")
 
   /**
-   * Generate a Keel SubmittedResource definition from currently existing resources.
+   * Generate a spec from currently existing resources.
    */
-  open suspend fun export(
-    exportable: Exportable
-  ): SubmittedResource<*> {
+  open suspend fun export(exportable: Exportable): S {
```
robfletcher: Made `export` just return the spec as it wasn't doing anything interesting with `SubmittedResource` (all implementations were identical. I'm hoping to keep that class out of `keel-api`.

robfletcher:
Opened #769 to track this across multiple PRs

keel-api/keel-api.gradle.kts
```diff
@@ -0,0 +1,5 @@
+plugins {
+  `java-library`
+}
+
+// DO NOT ADD ANY DEPENDENCIES HERE!
```
emjburns: Could you also add a brief explanation of why so that this comment is more likely to be respected? :)

robfletcher: Done

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryConfig.kt
```diff
@@ -12,7 +12,7 @@ data class DeliveryConfig(
   val serviceAccount: String,
   val artifacts: Set<DeliveryArtifact> = emptySet(),
   val environments: Set<Environment> = emptySet(),
-  val apiVersion: ApiVersion = ApiVersion("delivery.config.spinnaker.netflix.com", "v1")
```
gal-yardeni: Question - why were `ApiVersion` and `ResoucreId` objects and not Strings from the first place?

robfletcher: The intention was to restrict the format really, and to not confuse them with other values

PR #779, opened 2020-02-05
Moved DeliveryConfig and related stuff to keel-api

robfletcher:


keel-actuator/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
```diff
@@ -161,11 +160,10 @@ class ResourcePersister(
   }
 
   private fun Set<DeliveryArtifact>.transform(deliveryConfigName: String) =
-    this.map { artifact ->
+    map { artifact ->
       when (artifact) {
         is DockerArtifact -> artifact.copy(deliveryConfigName = deliveryConfigName)
         is DebianArtifact -> artifact.copy(deliveryConfigName = deliveryConfigName)
-        else -> throw UnsupportedArtifactTypeException(artifact.type.value())
```
robfletcher: Got rid of the need for this by making `DeliveryArtifact` a sealed class

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -171,7 +168,7 @@ class ArtifactListener(
     return ArtifactStatus.valueOf(status)
   }
 
-  private fun Artifact.type() = ArtifactType.valueOf(type.toUpperCase())
+  private fun Artifact.type() = ArtifactType.valueOf(type.toLowerCase())
```
robfletcher: It's slightly worrying how often we do this

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
```diff
@@ -33,8 +33,8 @@ internal class ArtifactListenerTests : JUnit5Minutests {
     .provenance("https://my.jenkins.master/jobs/fnord-release/58")
     .build()
 
-  val deb = DebianArtifact(name = "fnord", deliveryConfigName = "fnord-config")
-  val docker = DockerArtifact(name = "fnord/myimage", tagVersionStrategy = BRANCH_JOB_COMMIT_BY_JOB, deliveryConfigName = "fnord-config")
+  val debianArtifact = DebianArtifact(name = "fnord", deliveryConfigName = "fnord-config")
+  val dockerArtifact = DockerArtifact(name = "fnord/myimage", tagVersionStrategy = BRANCH_JOB_COMMIT_BY_JOB, deliveryConfigName = "fnord-config")
```
robfletcher: Names conflicted with renamed enums

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts.kt
```diff
@@ -17,46 +17,26 @@
...
-object DebianSemVerVersioningStrategy : VersioningStrategy(DEBIAN_VERSION_COMPARATOR) {
-  override fun toString(): String =
-    javaClass.simpleName
-
-  override fun equals(other: Any?): Boolean {
-    return other is DebianSemVerVersioningStrategy
-  }
-}
+val DockerArtifact.organization
+  get() = name.split("/").first()
```
robfletcher: Made this an extension to keep the `DockerArtifact` class simple (and easier to convert to Java)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts.kt
```diff
@@ -17,46 +17,26 @@
...
-    "${javaClass.simpleName}[strategy=$strategy, captureGroupRegex=$captureGroupRegex]}"
-}
+val DockerArtifact.image
+  get() = name.split("/").last()
+
+val VersioningStrategy.comparator: Comparator<String>
+  get() = when (this) {
+    is DebianSemVerVersioningStrategy -> DEBIAN_VERSION_COMPARATOR
+    is DockerVersioningStrategy -> TagComparator(strategy, captureGroupRegex)
+  }
```
robfletcher: The implementations of the comparators for `VersioningStrategy` rely on some 3rd party libs so it made sense to make this an extension and just move the model portion to `keel-api`. Also means it no longer requires `@JsonIgnore`.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifactTests.kt
```diff
@@ -0,0 +1,75 @@
...
+import com.fasterxml.jackson.module.kotlin.readValue
+import com.netflix.spinnaker.keel.serialization.configuredObjectMapper
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import strikt.api.expectCatching
+import strikt.api.expectThat
+import strikt.assertions.containsKey
+import strikt.assertions.succeeded
+
+internal class DeliveryArtifactTests : JUnit5Minutests {
```
robfletcher: Wanted to test that the non-annotation based Jackson stuff worked

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: There's a kork `artifact` and the artifact we use. Both have a type. We only do this (I think) when we're working with or converting the kork artifact.

robfletcher: I sure hope there aren't any times we do it that aren't covered by a test

emjburns: 😬 me too.

robfletcher:
> Awesome! I don't understand all the jackson bits, but it looks good.

I'll try to clarify that with the next PR

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
```diff
@@ -0,0 +1,38 @@
+package com.netflix.spinnaker.keel.api
+
+enum class ArtifactType {
+  deb, docker
```
luispollo: Was changing this to `UPPER` to `lower` to simplify (de-)serialization, or some other reason? Feels like it breaks naming conventions...

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/TagVersioningStrategy.kt
```diff
@@ -17,24 +17,18 @@
...
-  @JsonProperty("semver-job-commit-by-job")
-  SEMVER_JOB_COMMIT_BY_JOB("""^v.*-h(\d+).*$""", INCREASING), // popular netflix strategy
-  @JsonProperty("semver-job-commit-by-semver")
-  SEMVER_JOB_COMMIT_BY_SEMVER("""^v(.*)-h\d+.*$""", SEMVER); // popular netflix strategy
+enum class TagVersionStrategy(val regex: String, val sortType: SortType, val friendlyName: String) {
+  INCREASING_TAG("""(^.*$)""", INCREASING, "increasing-tag"),
+  SEMVER_TAG("""(^.*$)""", SEMVER, "semver-tag"),
+  BRANCH_JOB_COMMIT_BY_JOB("""^master-h(\d+).*$""", INCREASING, "branch-job-commit-by-job"), // popular netflix strategy
+  SEMVER_JOB_COMMIT_BY_JOB("""^v.*-h(\d+).*$""", INCREASING, "semver-job-commit-by-job"), // popular netflix strategy
+  SEMVER_JOB_COMMIT_BY_SEMVER("""^v(.*)-h\d+.*$""", SEMVER, "semver-job-commit-by-semver"); // popular netflix strategy
```
luispollo: @emjburns Side-comment: as much as I guess they're self-explanatory, these names feel too long and easy to get wrong if you're manually creating a spec. Maybe there's an opportunity to simplify a bit? I'd offer suggestions but I'm unfamiliar with the details -- maybe we could add some examples in the comments?

emjburns: Here are the docs about it, with examples: https://www.spinnaker.io/guides/user/managed-delivery/artifacts/#docker-images. I prefer explicit over short in this case.

luispollo: Agreed. After reading the descriptions again, I understand what the names mean. It's unfortunate that the `-by-<field>` at the end is ambiguous, but I can't think of better alternatives. 👍

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
robfletcher: Exactly that. I'm not totally happy with doing this but I'm not sure if there's another way to avoid having to use `@JsonValue`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/TagVersioningStrategy.kt
robfletcher: Yeah, I had a similar reaction to these. They didn't feel super meaningful to me.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifact.kt
robfletcher: I'll see if I can figure another way to do it and revert that part of the change.

PR #783, opened 2020-02-06
Plugin readiness

robfletcher:
I screwed up and accidentally pushed the first couple of commits to this onto the end of my previous PR when trying to rebase it.

I've moved all the Jackson code relating to handling types from `keel-api` into a new `keel-api-jackson` module. This is so it's easier to understand and we've got a good home for anything further we need to add there.


keel-api-jackson/keel-api-jackson.gradle.kts
```diff
@@ -26,6 +26,9 @@ dependencies {
   implementation("org.springframework:spring-context")
   implementation("org.springframework.boot:spring-boot-autoconfigure")
 
+  // this is really just so we can access `configuredObjectMapper`
+  testImplementation(project(":keel-core"))
```
robfletcher: I guess the question I have is whether it's worth moving _all_ the Jackson configuration to this new module, or whether there's value in having it specifically dedicated to supporting the API types. 

Any thoughts on that?

emjburns: I have no idea.

Issue #762, opened 2020-01-28
Export should put only differences in overrides, not things that are the same between regions 


emjburns:
Also, you're allowed to leave keypair blank (and you probably should). We shouldn't export that.

For example:

Actual:
```yaml
- apiVersion: ec2.spinnaker.netflix.com/v1
    kind: cluster
    metadata: {}
    spec:
      capacity:
        desired: 2
        max: 2
        min: 2
...
      launchConfiguration:
        ebsOptimized: true
        instanceType: r4.xlarge # why are you here instance type?
        keyPair: nf-keypair-111-us-east-1 # this shouldn't be shown even though technically it's different...
      overrides:
        eu-west-1:
          launchConfiguration:
            ebsOptimized: true
            instanceType: r4.xlarge
            keyPair: nf-keypair-111-eu-west-1
        us-west-2:
          capacity:
            desired: 3
            max: 3
            min: 3
          launchConfiguration:
            ebsOptimized: true
            instanceType: r4.xlarge
            keyPair: nf-keypair-111-us-west-2
```

expected:
```yaml
  - apiVersion: ec2.spinnaker.netflix.com/v1
    kind: cluster
    metadata: {}
    spec:
      capacity:
        desired: 2
        max: 2
        min: 2
...
      launchConfiguration:
        ebsOptimized: true
        instanceType: r4.xlarge
      overrides:
        us-west-2:
          capacity:
            desired: 3
            max: 3
            min: 3
```

spinnakerbot:
This issue hasn't been updated in 52 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

PR #784, opened 2020-02-07
fix(export): export less & more correct

emjburns:
Fixes https://github.com/spinnaker/keel/issues/762

Fixing some problems reported with exporting. 

- simplifying some logic for converting objects into spec objects (use object mapper now)
- choose the server group with the largest size to be the "base" (because that's probably the main one)
- allow you to override just capacity(min/max) if you have scaling policies also
- pull the export tests out to their own files for titus and ec2 clusters
- update the tests a bit to add new cases and condense some of the `strikt` logic


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -888,20 +896,20 @@ class ClusterHandler(
...
-      health = Health().exportSpec(),
+      health = Health().toSpecWithoutDefaults(),
       scaling = Scaling(),
       tags = emptyMap()
     )
 
-    val thisSpec = ClusterSpec.ServerGroupSpec(
-      launchConfiguration = launchConfiguration.exportSpec(account, application),
+    val thisSpec = ServerGroupSpec(
+      launchConfiguration = launchConfiguration.exportSpec(account, location.region, application, launchConfiguration.instanceType),
```
emjburns: we need to pass in the instance type here, otherwise we incorrectly think it is overridden.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/TestUtils.kt
```diff
@@ -0,0 +1,105 @@
...
+import com.netflix.spinnaker.keel.clouddriver.model.SecurityGroupSummary
+import com.netflix.spinnaker.keel.clouddriver.model.Subnet
+import com.netflix.spinnaker.keel.clouddriver.model.SuspendedProcess
+import com.netflix.spinnaker.keel.clouddriver.model.Tag
+import com.netflix.spinnaker.keel.clouddriver.model.TargetTrackingConfiguration
+import com.netflix.spinnaker.keel.ec2.CLOUD_PROVIDER
+import com.netflix.spinnaker.keel.model.parseMoniker
+import org.apache.commons.lang3.RandomStringUtils
+
+fun ServerGroup.toCloudDriverResponse(
```
emjburns: Pulling this out to be used by both export and cluster handler tests. No change in functionality.

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TestUtils.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.netflix.spinnaker.keel.api.titus.cluster.moniker
+import com.netflix.spinnaker.keel.clouddriver.model.Placement
+import com.netflix.spinnaker.keel.clouddriver.model.SecurityGroupSummary
+import com.netflix.spinnaker.keel.clouddriver.model.ServiceJobProcesses
+import com.netflix.spinnaker.keel.clouddriver.model.TitusActiveServerGroup
+import com.netflix.spinnaker.keel.clouddriver.model.TitusActiveServerGroupImage
+import com.netflix.spinnaker.keel.model.parseMoniker
+import org.apache.commons.lang3.RandomStringUtils
+
+fun TitusServerGroup.toClouddriverResponse(
```
emjburns: Pulling this out to be used by both export and cluster handler tests. No change in functionality.

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
```diff
@@ -470,103 +433,6 @@ class TitusClusterHandlerTests : JUnit5Minutests {
             .containsDistinctElements()
         }
       }
-      context("export without overrides") {
```
emjburns: Pulled to their own file, and update.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandlerTests.kt
```diff
@@ -444,72 +349,6 @@ internal class ClusterHandlerTests : JUnit5Minutests {
           verify { publisher.publishEvent(ofType<ArtifactVersionDeployed>()) }
         }
       }
-
```
emjburns: Pulled to their own file, and updated.

keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceSpecExportHelper.kt
```diff
@@ -60,24 +60,3 @@ inline fun <reified T : Any, reified S : Any> buildSpecFromDiff(
     ctor.callBy(params)
   }
 }
-
```
emjburns: Removed in favor of using the object mapper

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
```diff
@@ -174,34 +165,6 @@ class TitusClusterHandlerTests : JUnit5Minutests {
     )
   )
 
-  private fun TitusServerGroup.toClouddriverResponse(
```
emjburns: Pulled to TestUtils

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -337,8 +346,7 @@ class ClusterHandler(
     val spec = ClusterSpec(
       moniker = exportable.moniker,
       imageProvider = if (base.buildInfo?.packageName != null) {
-        ArtifactImageProvider(
-          deliveryArtifact = DebianArtifact(name = base.buildInfo.packageName!!))
+        ReferenceArtifactImageProvider(reference = base.buildInfo.packageName.toString())
```
luispollo: This seems unrelated?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/ExportError.kt
```diff
@@ -0,0 +1,3 @@
+package com.netflix.spinnaker.keel.exceptions
+
+class ExportError(message: String) : RuntimeException(message)
```
robfletcher: Is this translated into an HTTP error on the export API?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
emjburns: oh it totally is. But I wanted to do it. I can back it out if you care

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/ExportError.kt
emjburns: Do you think this should be a `500`? this is basically just used because kotlins null safety didn't catch that I throw an exception when there's an empty list, but then finding the first in the list can still be null. I don't think we will ever throw it, and if we don't I think it's a bug on our side.

robfletcher: In that case 500 seems right. I was more just wondering if this exception might start getting used for other things in future.

emjburns: I don't think it will, but if it does I'll add the exception handling by the controllers.

gal-yardeni:
LGTM! I think it makes sense to choose the server group with the largest size to be the "base" as opposed to just choosing the first one.

emjburns:
I screwed up this PR. New one here: https://github.com/spinnaker/keel/pull/795

Issue #769, opened 2020-01-30
Initial plugin readiness


robfletcher:
Part of #633 

I want to break this up into multiple PRs so this issue will track them.

## To-do

- [ ] `keel-api` should contain everything a plugin implementor would need
  - [x] `DeliveryConfig` and `Environment`
  - [x] `Resource` and `ResourceSpec`
  - [x] `ResourceHandler`
  - [x] `Resolver`
  - [x] `TaskLauncher` without direct dependency on `keel-orca`, i.e. just as an interface
  - [ ] An equivalent abstraction for CloudDriver without relying directly on Retrofit
  - [x] `Constraint` (migrating sub-classes to be registered like resource types) and `ConstraintEvaluator`
- [x] `keel-api` should have no dependencies on anything beyond Kotlin (for now)
- [x] Everything in `keel-actuator` and `keel-plugin` should either be moved to `keel-api` or `keel-core`
- [ ] `keel-bakery-plugin`, `keel-ec2-plugin` and `keel-titus-plugin` should only depend on `keel-api` (this is likely to involve the most work)
- [x] Fix up package names in `keel-api` so they don't conflict with other modules

## Considerations / Assumptions

The main things to consider are keeping Jackson, Java-object-diff, and any other libraries (I think those are the main ones that are likely to need some refactoring to remove from the crucial places) out of `keel-api`. 
We can revisit whether it's worth the effort to make `keel-api` pure Java (in order to remove the Kotlin dependency) later.

I think the `Submitted*` types don't need to be in `keel-api` which simplifies things a little as they are most likely to use Jackson annotations.

PR #788, opened 2020-02-07
moved some more things to keel-api, merged some other modules

robfletcher:
Part of  #769

I merged `keel-actuator`, `keel-plugin`, and `keel-veto` into `keel-core`. Since core is the engine for running Keel it didn't make sense to me to have those things separate any more.

I also migrated `TaskLauncher`, `Monikered`, and all the `Locations` stuff to `keel-api` with the eventual goal of having `keel-bakery-plugin`, `keel-ec2-plugin`, and `keel-titus-plugin` not having to depend on `keel-core`.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/plugin/TaskLauncher.kt
```diff
@@ -30,12 +29,12 @@ interface TaskLauncher {
...
     user: String,
     application: String,
-    notifications: List<OrcaNotification>,
+    notifications: Collection<NotificationConfig>,
     subject: String,
     description: String,
     correlationId: String,
     stages: List<Map<String, Any?>>,
-    artifacts: List<Artifact> = emptyList()
+    artifacts: List<Map<String, Any?>> = emptyList()
```
robfletcher: I don't want to have `keel-api` depend on `kork-artifacts`

keel-api-jackson/src/main/kotlin/com/netflix/spinnaker/keel/json/mixins/LocatableMixin.kt
```diff
@@ -0,0 +1,9 @@
+package com.netflix.spinnaker.keel.json.mixins
+
+import com.fasterxml.jackson.annotation.JacksonInject
+import com.netflix.spinnaker.keel.api.Locations
+
+interface LocatableMixin<T : Locations<*>> {
```
luispollo: Side-comment: I understand the reason, but having to do all this in code as opposed to using annotations feels like a step backwards... 🙁 What's the issue with annotations and plugins?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Monikered.kt
```diff
@@ -17,14 +17,15 @@
  */
 package com.netflix.spinnaker.keel.api
 
-import com.fasterxml.jackson.annotation.JsonIgnore
-
 /**
- * An object which has a moniker
+ * An resource that has a moniker-compliant name.
```
luispollo:
```suggestion
 * A resource that has a moniker-compliant name.
```

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/plugin/TaskLauncher.kt
```diff
@@ -30,12 +29,12 @@ interface TaskLauncher {
   suspend fun submitJob(
     user: String,
     application: String,
-    notifications: List<OrcaNotification>,
+    notifications: Collection<NotificationConfig>,
```
luispollo: Shouldn't this be a `Set` instead?

robfletcher: Do you mean because we want to guarantee uniqueness of elements?

luispollo: Yes. I would expect we wouldn't allow duplicate/redundant notification configs, in this case.

Issue #748, opened 2020-01-23
Split up resource handlers


robfletcher:
Currently resource handlers violate the single responsibility principle. They are responsible for:

1. getting the current state of a resource
2. computing the desired state of a resource
3. actuating changes to a resource based on a diff
4. exporting a resource

I think it would make sense to split those responsibilities up. In particular I think the tests suffer from having to cover too much functionality (although this is partly due to the way we've structured tests around components rather than behavior, boo). Also there's potential to re-use the same code for 3 for many different resource types.

emjburns:
I think this makes sense. I'm in favor of anything that can help us reuse code across handlers. In general I'd like to make lots of things abstract and have individual implementations be responsible for things like grabbing specific fields from within their spec and updating spec pieces. I did this for the [docker image resolver](https://github.com/spinnaker/keel/blob/master/keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/DockerImageResolver.kt).

spinnakerbot:
This issue hasn't been updated in 63 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

PR #792, opened 2020-02-10
Package organization

robfletcher:
Part of #769 

This is just re-organizing package locations so that `keel-api` and `keel-core` do not both contain a `com.netflix.spinnaker.keel.api` package.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/plugin/KeelPlugin.kt
```diff
@@ -1,21 +0,0 @@
...
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-package com.netflix.spinnaker.keel.plugin
-
-interface KeelPlugin {
-  val name: String
-    get() = javaClass.simpleName
-}
```
robfletcher: I think this is literally the only actual change here. The only sub-type of this interface was `ResourceHandler` so I thought it made sense to get rid of it.

luispollo: Hmm... What about #748?

robfletcher: I think whether the result of that would require a common top-level interface is an open question

PR #805, opened 2020-02-12
fix(config): add validation that resource names are unique

emjburns:
We currently don't validate that you can't have the same resource in more than one env. This PR adds validation based on resource id. 

If the resource is present in more than one env, we throw an exception with a detailed error message and delete the delivery config. We do it in this order because we only calculate the resource id once it's been normalized and saved.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
@@ -14,6 +14,7 @@ import com.netflix.spinnaker.keel.core.api.SubmittedDeliveryConfig
 import com.netflix.spinnaker.keel.core.api.SubmittedResource
 import com.netflix.spinnaker.keel.core.api.id
 import com.netflix.spinnaker.keel.core.api.normalize
+import com.netflix.spinnaker.keel.core.api.resources
 import com.netflix.spinnaker.keel.diff.DefaultResourceDiff
 import com.netflix.spinnaker.keel.events.ArtifactRegisteredEvent
 import com.netflix.spinnaker.keel.events.ResourceCreated
@@ -77,8 +78,10 @@ class ResourcePersister(
     try {
       validate(new)
     } catch (e: ValidationException) {
-      log.warn("Validation of ${new.name} failed, deleting delivery config")
-      deleteDeliveryConfig(new.name)
+      log.warn("Validation of ${new.name} failed, deleting already persisted resources")
+      new.resources.forEach { resource ->
+        resourceRepository.delete(resource.id)
+      }
       throw e
     }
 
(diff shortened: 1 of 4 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
```diff
@@ -71,6 +73,15 @@ class ResourcePersister(
         )
       }
     )
+
+    try {
+      validate(new)
+    } catch (e: ValidationException) {
+      log.warn("Validation of ${new.name} failed, deleting delivery config")
+      deleteDeliveryConfig(new.name)
```
robfletcher: Why do we need to delete it? We didn't save it yet.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
```diff
@@ -81,6 +92,29 @@ class ResourcePersister(
     return new
   }
 
+  /**
+   * Validates that resources have unique ids, throws an exception if invalid
+   */
+  private fun validate(config: DeliveryConfig) {
+    val resources = config.environments.map { it.resources }.flatten().map { it.id }
```
gal-yardeni: can a map have duplicated keys in this case?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
emjburns: Ah you're right, we only need to delete the resources. Good catch.

robfletcher: I would propose we don't persist the resources until after this validation check

emjburns: That's definitely better. I'll do that

emjburns: Fixed. Wow much better. So wise.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourcePersister.kt
```diff
@@ -60,32 +60,60 @@ class ResourcePersister(
...
+    val distinct = resources.distinct()
+
+    if (resources.size != distinct.size) {
+      val duplicates = resources.groupingBy { it }.eachCount().filter { it.value > 1 }.keys.toList()
+      val envToResources: Map<String, MutableList<String>> = config.environments
+        .map { env -> env.name to env.resources.map { it.id }.toMutableList() }.toMap()
+      val envsAndDuplicateResources = envToResources
+        .filterValues { rs: List<String> ->
+          // remove all the resources we don't care about from this mapping
+          rs.filter { it in duplicates }
```
luispollo: The result of this expression is not being assigned to anything, so my guess is that the line below doesn't have the intended effect...

Shouldn't the two statements be combined as follows?
```kotlin
rs.filter { it in duplicates }.isNotEmpty()
```

emjburns: 🤦‍♂ yes!

PR #806, opened 2020-02-13
fix(resourceEvents): adding statuses to ResourceTaskSucceeded and ResourceTaskFailed

gal-yardeni:
Adding `ACTUATING` status to both ResourceTaskSucceeded and ResourceTaskFailed events (events triggered based on statuses of orca tasks), as it currently implemented as informative tasks (meaning just provide info to the user, and we're not deciding upon them if there is an actual error).


gal-yardeni:
closing this PR as it got messed up. opened https://github.com/spinnaker/keel/pull/808 instead.

PR #814, opened 2020-02-19
feat(environments): automated blacklisting of versions from environments

asher:
Given a scenario where an approved artifact version only comes up healthy in one region of a multi-region cluster, we want to ensure the cluster and its containing environment are automatically restored to a good and consistent state. Prior to this PR, the cluster would enter an unhappy state after 5 deploy attempts with the cluster left running inconsistent versions across regions. Other clusters in the environment with the same artifact reference continue to resolve desired state to the problematic artifact version.

With this PR, after 2 failed attempts deploying a cluster, the artifact version is vetoed from the environment. All clusters in the environment referencing the artifact should then resolve to the previous version. In the scenario where a multi-region cluster successfully deployed in only one region, that region is rolled back in the next actuation loop.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
@@ -44,9 +44,9 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
     val subject: T
   ) {
     // the artifact built off a feature branch
-    val artifact1 = DebianArtifact("keeldemo", deliveryConfigName = "my-manifest", reference = "candidate", statuses = listOf(SNAPSHOT))
+    val artifact1 = DebianArtifact("keeldemo", deliveryConfigName = "my-manifest", reference = "candidate", statuses = setOf(SNAPSHOT))
     // the artifact built off of master
-    val artifact2 = DebianArtifact("keeldemo", deliveryConfigName = "my-manifest", reference = "master", statuses = listOf(RELEASE))
+    val artifact2 = DebianArtifact("keeldemo", deliveryConfigName = "my-manifest", reference = "master", statuses = setOf(RELEASE))
     val artifact3 = DockerArtifact("docker", deliveryConfigName = "my-manifest", reference = "docker-artifact")
     val environment1 = Environment("test")
     val environment2 = Environment("staging")
@@ -108,7 +108,7 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
       .artifacts
       .first {
         if (artifact is DebianArtifact) {
-          it.name == artifact.name && it.type == artifact.type && artifact.statuses == it.statuses.toList()
+          it.name == artifact.name && it.type == artifact.type && artifact.statuses == it.statuses
         } else {
           it.name == artifact.name && it.type == artifact.type
         }
@@ -210,7 +210,7 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
         }
 
         test("querying for all returns all") {
-          val artifactWithAll = artifact1.copy(statuses = emptyList())
+          val artifactWithAll = artifact1.copy(statuses = emptySet())
           expectThat(subject.versions(artifactWithAll)).containsExactly(version5, version4, version3, version2, version1)
         }
 
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
@@ -12,6 +12,7 @@ import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.artifacts.DockerArtifact
 import com.netflix.spinnaker.keel.core.api.ArtifactVersionStatus
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactPin
+import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVetoes
 import com.netflix.spinnaker.time.MutableClock
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
@@ -105,7 +106,13 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
       .versionsByEnvironment(manifest)
       .first { it.name == environment.name }
       .artifacts
-      .first { it.name == artifact.name && it.type == artifact.type }
+      .first {
+        if (artifact is DebianArtifact) {
+          it.name == artifact.name && it.type == artifact.type && artifact.statuses == it.statuses.toList()
+        } else {
+          it.name == artifact.name && it.type == artifact.type
+        }
+      }
       .versions
   }
 
@@ -431,6 +438,12 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
             .isEqualTo(version4)
             .isEqualTo(pin1.version)
         }
+
+        test("pinned version cannot be vetoed") {
+          subject.pinEnvironment(manifest, pin1)
+          expectThat(subject.markAsVetoedIn(manifest, artifact2, pin1.version!!, pin1.targetEnvironment))
+            .isFalse()
+        }
       }
     }
 
@@ -463,6 +476,53 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
           that(subject.getAll().size).isEqualTo(3)
           that(subject.getAll(docker).size).isEqualTo(1)
           that(subject.getAll(deb).size).isEqualTo(2)
+        }
+      }
+    }
+
+    context("the latest version is vetoed") {
+      before {
+        subject.flush()
+        persist()
+        subject.approveVersionFor(manifest, artifact2, version4, environment2.name)
+        subject.approveVersionFor(manifest, artifact2, version5, environment2.name)
+        subject.markAsVetoedIn(manifest, artifact2, version5, environment2.name)
+      }
+
+      test("latestVersionApprovedIn reflects the veto") {
+        expectThat(subject.latestVersionApprovedIn(manifest, artifact2, environment2.name))
+          .isEqualTo(version4)
+      }
+
+      test("vetoedEnvironmentVersions reflects the veto") {
+        expectThat(subject.vetoedEnvironmentVersions(manifest))
+          .isEqualTo(
+            listOf(
+              EnvironmentArtifactVetoes(
+                deliveryConfigName = manifest.name,
+                targetEnvironment = environment2.name,
+                artifact = artifact2,
+                versions = mutableSetOf(version5)
+              )
+            )
+          )
+      }
+
+      test("version status reflects the veto") {
+        expectThat(versionsIn(environment2, artifact2)) {
+          get(ArtifactVersionStatus::vetoed).containsExactly(version5)
+          get(ArtifactVersionStatus::approved).containsExactly(version4)
+        }
+      }
+
+      test("unveto the vetoed version") {
+        subject.deleteVeto(manifest, artifact2, version5, environment2.name)
+
+        expectThat(subject.latestVersionApprovedIn(manifest, artifact2, environment2.name))
+          .isEqualTo(version5)
+        expectThat(versionsIn(environment2, artifact2)) {
+          get(ArtifactVersionStatus::vetoed).isEmpty()
+          get(ArtifactVersionStatus::approved).containsExactlyInAnyOrder(version4, version5)
         }
       }
     }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
@@ -5,6 +5,7 @@ import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.constraints.ConstraintEvaluator
 import com.netflix.spinnaker.keel.constraints.StatefulConstraintEvaluator
+import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVetoes
 import com.netflix.spinnaker.keel.core.api.PinnedEnvironment
 import com.netflix.spinnaker.keel.persistence.ArtifactRepository
 import com.netflix.spinnaker.keel.telemetry.ArtifactVersionApproved
@@ -26,6 +27,9 @@ class EnvironmentPromotionChecker(
     val pinnedEnvs: Map<String, PinnedEnvironment> = artifactRepository
       .pinnedEnvironments(deliveryConfig)
       .associateBy { envPinKey(it.targetEnvironment, it.artifact) }
+    val vetoedArtifacts: Map<String, EnvironmentArtifactVetoes> = artifactRepository
+      .vetoedEnvironmentVersions(deliveryConfig)
+      .associateBy { envPinKey(it.targetEnvironment, it.artifact) }
 
     deliveryConfig
       .artifacts
@@ -35,12 +39,18 @@ class EnvironmentPromotionChecker(
           log.warn("No versions for ${artifact.type} artifact ${artifact.name} are known")
         } else {
           deliveryConfig.environments.forEach { environment ->
+            val vetoedVersions: Set<String> =
+              (vetoedArtifacts[envPinKey(environment.name, artifact)]?.versions)
+                ?: emptySet()
+
             val version = when {
               pinnedEnvs.hasPinFor(environment.name, artifact) -> {
                 pinnedEnvs.versionFor(environment.name, artifact)
               }
               environment.constraints.isEmpty() -> {
-                versions.firstOrNull()
+                versions.firstOrNull { v ->
+                  !vetoedVersions.contains(v)
+                }
               }
               else -> {
                 versions.firstOrNull { v ->
@@ -49,7 +59,8 @@ class EnvironmentPromotionChecker(
                    * want to request judgement or deploy a canary for artifacts that aren't
                    * deployed to a required environment or outside of an allowed time.
                    */
-                  checkConstraints(statelessEvaluators, artifact, deliveryConfig, v, environment) &&
+                  !vetoedVersions.contains(v) &&
+                    checkConstraints(statelessEvaluators, artifact, deliveryConfig, v, environment) &&
                     checkConstraints(statefulEvaluators, artifact, deliveryConfig, v, environment)
                 }
               }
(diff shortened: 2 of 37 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -105,8 +112,17 @@ data class ClusterSpec(
...
   @JsonInclude(NON_EMPTY)
-  val overrides: Map<String, ServerGroupSpec> = emptyMap()
-) : Monikered, Locatable<SubnetAwareLocations> {
+  val overrides: Map<String, ServerGroupSpec> = emptyMap(),
+  @JsonIgnore
+  override val deliveryArtifact: DeliveryArtifact? = null,
+  @JsonIgnore
+  override val artifactVersion: String? = null,
+  @JsonIgnore
+  override val maxDiffCount: Int? = 2,
```
asher: This changes the number of times keel will attempt to actuate `ClusterSpec` for a given diff fingerprint from 5 to 2.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -105,8 +112,17 @@ data class ClusterSpec(
...
+  @JsonIgnore
+  // Once clusters go unhappy, only retry when the diff changes, or if manually unvetoed
+  override val unhappyWaitTime: Duration? = Duration.ZERO
```
asher: This means that unhappy clusters remain paused indefinitely until either the diff changes, or a user intervenes.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -123,13 +133,16 @@ class ImageResolver(
...
+            image.namedImage.baseImageVersion
           )
         )
     }
 
-    return copy(spec = spec.copy(overrides = overrides))
+    return copy(spec = spec.copy(
+      overrides = overrides,
+      deliveryArtifact = image.artifact,
+      artifactVersion = image.version))
```
asher: This is how `ArtifactVersioned` properties are applied to `ClusterSpec`.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -378,6 +384,201 @@ class SqlArtifactRepository(
...
+                ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.eq(version)
+              )
+              .execute()
+
+            /**
+             * If there's a previously deployed version in [targetEnvironment], set `promotion_reference`
+             * to the version that's currently being vetoed. If that version also fails to fully deploy,
+             * this is used to short-circuit further automated vetoes. We want to avoid a cloud provider
+             * or other issue unrelated to an artifact version triggering continual automated rollbacks
+             * thru all previously deployed versions.
```
asher: Calling attention to this comment explaining the use of `ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS`, a new column added by this PR and used to prevent recursively vetoing prior artifact versions.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -105,7 +106,13 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
       .versionsByEnvironment(manifest)
       .first { it.name == environment.name }
       .artifacts
-      .first { it.name == artifact.name && it.type == artifact.type }
+      .first {
+        if (artifact is DebianArtifact) {
+          it.name == artifact.name && it.type == artifact.type && artifact.statuses == it.statuses.toList()
```
luispollo: Should the statuses  comparison be ignoring order? Or perhaps the fields should be sets instead of collections or lists?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -26,6 +27,9 @@ class EnvironmentPromotionChecker(
     val pinnedEnvs: Map<String, PinnedEnvironment> = artifactRepository
       .pinnedEnvironments(deliveryConfig)
       .associateBy { envPinKey(it.targetEnvironment, it.artifact) }
+    val vetoedArtifacts: Map<String, EnvironmentArtifactVetoes> = artifactRepository
+      .vetoedEnvironmentVersions(deliveryConfig)
+      .associateBy { envPinKey(it.targetEnvironment, it.artifact) }
```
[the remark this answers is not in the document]
luispollo: Do we foresee any other usages of `EnvironmentArtifactVetoes` beyond keeping multi-region clusters consistent and which might persist for long periods, in which case storing the whole collection in memory here might be of concern?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactVersioned.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.keel.api
+
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+
+interface ArtifactVersioned {
```
luispollo: Perhaps a better name for this would be something like `ArtifactVersionProvider` or `ArtifactVersionAware`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -71,6 +78,45 @@ class ResourceActuator(
...
+           * [ArtifactVersioned] is a special [resource] sub-type. When a veto response sets
+           * [VetoResponse.vetoArtifact] and the resource under evaluation is of type
+           * [ArtifactVersioned], blacklist the desired artifact version from the environment
+           * containing [resource]. This ensures that the environment will be fully restored to
+           * a prior good-state.
+           */
+          if (response.vetoArtifact && resource.spec is ArtifactVersioned) {
+            try {
+              val (version, artifact) = when (desired) {
+                is Map<*, *> -> {
```
luispollo: Why the stars here if we're casting to a specific type below?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -71,6 +78,45 @@ class ResourceActuator(
...
+                artifactRepository.markAsVetoedIn(
+                  deliveryConfig = deliveryConfig,
+                  artifact = artifact,
+                  version = version,
+                  targetEnvironment = environment)
+                // TODO: emit event + metric
+              }
+            } catch (e: Exception) {
+              log.warn("Failed to veto presumed bad artifact version for ${resource.id}", e)
+              // TODO: emit metric
```
luispollo: Beyond metrics, I feel like this should cause a notification as it will have user impact.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
```diff
@@ -18,3 +18,18 @@ data class PinnedEnvironment(
...
+
+data class EnvironmentArtifactVeto(
+  val deliveryConfigName: String,
+  val targetEnvironment: String,
+  val reference: String,
+  val type: String,
+  val version: String
+)
+
+data class EnvironmentArtifactVetoes(
```
luispollo: Any reason why we can't just use `List<EnvironmentArtifactVeto>` instead of this class?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactsSummary.kt
```diff
@@ -10,12 +11,15 @@ data class EnvironmentArtifactsSummary(
 data class ArtifactVersions(
   val name: String,
   val type: ArtifactType,
+  val statuses: Collection<ArtifactStatus>,
```
luispollo: Not a `Set`? Also in `DebianArtifact` I see it's a `List`, which seems like it should also be a `Set` as there's no point in allowing redundant status on the collection.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/UnhappyControl.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.keel.api
+
+import java.time.Duration
+
+interface UnhappyControl {
```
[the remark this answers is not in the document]
emjburns: ooo nice

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
```diff
@@ -90,15 +93,37 @@ class UnhappyVeto(
...
+          String::class.java,
+          "veto.unhappy.waiting-time",
+          configuredWaitingTime))
+
+  private fun unhappyMessage(resource: Resource<*>): String {
+    val maxDiffs = maxDiffCount(resource)
+    val waitingTime = waitingTime(resource)
+    if (waitingTime == Duration.ZERO) {
+      return "Resource is unhappy and our $maxDiffs actions have not fixed it. " +
+        "Resource will remain paused until the diff changes or manually unpaused."
```
luispollo:
```suggestion
        "Resource will remain paused until the diff changes or the resource is manually unpaused."
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactVersioned.kt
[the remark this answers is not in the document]
emjburns: What do you think about changing this class name to `VersionedArtifact`? ArtifactVersioned keeps tripping me up as a I read it.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -41,6 +42,12 @@ class ImageResolver(
   override val supportedKind: String = "cluster"
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
+  data class VersionedNameImage(
```
luispollo:
```suggestion
  data class VersionedNamedImage(
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -378,6 +384,200 @@ class SqlArtifactRepository(
...
+        ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_REFERENCE)
+      .from(ENVIRONMENT_ARTIFACT_VERSIONS)
+      .where(
+        ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(envUid),
+        ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artUid),
+        ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.eq(version))
+      .fetchOne { (status, reference) ->
+        if (reference != null && !force) {
+          log.warn(
+            "Not vetoing artifact version as it appears to have already been an automated rollback target: " +
```
luispollo: Is this true regardless of the value of `status`?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -582,6 +814,13 @@ class SqlArtifactRepository(
       .where(ENVIRONMENT.NAME.eq(environment.name))
       .and(ENVIRONMENT.DELIVERY_CONFIG_UID.eq(uid))
 
+  private fun DeliveryConfig.getUidStringFor(environment: Environment): String =
```
luispollo: Side-comment: I find these methods named `SomeClass.getSomethingFor(another: AnotherClass)` hard to read. I'd suggest adding the name of the thing we're getting the property fo r, as in `DeliveryConfig.getUidStringForEnvironment(environment)`, in this case. I know it makes for more verbose code, but also more readable IMHO.

Beyond the scope of this PR, though, as they're all over the place.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -107,6 +108,42 @@ class ArtifactController(
     artifactRepository.deletePin(deliveryConfig, targetEnvironment)
   }
 
+  @PostMapping(
+    path = ["/veto"]
+  )
+  @ResponseStatus(ACCEPTED)
+  fun veto(
+    @RequestHeader("X-SPINNAKER-USER") user: String,
+    @RequestBody veto: EnvironmentArtifactVeto
```
luispollo: Since this is the only place we use this class (the one named in the singular), is it worth having it or could we just use `EnvironmentArtifactVetoes` and the UI/user will pass an array with a single version if needed?

I'm only asking because I got a bit confused when I saw the two classes looking almost identical in the model.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactVersioned.kt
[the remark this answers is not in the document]
luispollo: @emjburns I had the same feeling. Suggested a few other names here: https://github.com/spinnaker/keel/pull/814#discussion_r381423206

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -71,6 +78,45 @@ class ResourceActuator(
...
+           * [VetoResponse.vetoArtifact] and the resource under evaluation is of type
+           * [ArtifactVersioned], blacklist the desired artifact version from the environment
+           * containing [resource]. This ensures that the environment will be fully restored to
+           * a prior good-state.
+           */
+          if (response.vetoArtifact && resource.spec is ArtifactVersioned) {
+            try {
+              val (version, artifact) = when (desired) {
+                is Map<*, *> -> {
+                  if (desired.size > 0) {
```
emjburns: Is this what mulit-region looks like?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
asher: Great point, will fix.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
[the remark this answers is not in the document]
asher: `EnvironmentPromotionChecker` is already loading every artifact version into memory further down. There are scalability issues in this class, but I don't think loading in a separate list of just vetoed versions (should always be a small subset of the whole) is meaningful or divergent from existing behavior. Not sure if I'm groking the other part of your question,`EnvironmentArtifactVetoes` aren't persisted objects and keel currently has zero logic to clean up anything regarding ancient artifact versions which is indeed bad. `EnvironmentPromotionChecker` contains operations that grow linearly in cost with every artifact release. That needs to be addressed but this PR doesn't introduce new cases. I think that subject would be better discussed in an issue regarding artifact cleanup and query cost.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
asher: It seemed required, please feel free to suggest a change.

asher: If you're specifically asking why it isn't `is Map<String, ArtifactVersioned>`, that doesn't work.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
asher: Yes.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
asher: Not this PR.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactVersioned.kt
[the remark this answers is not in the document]
asher: Ok.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
asher: Yep.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
asher: The internal class has `DeliveryArtifact` on it, the external class has the (reference, type) properties by which users refer to a `DeliveryArtifact`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactsSummary.kt
asher: Good call

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: I guess my point was to add that to the `TODO` comment if it made sense. 😊

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
luispollo: Ah, Ok! I missed that.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: What do you mean? Inellij can show you the type hints if the actual type is not present.

asher: Discovering intellj's command-p feature while typing a method (i.e.: `foo(` <- hit cmd-p here)) after a year of use was a revelation for me in case it's new to you as well.

PR #815, opened 2020-02-19
Misc improvements to OpenAPI rendering

robfletcher:
- marked all properties as `nullable:  true/false` based on Kotlin data type.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/docs/Description.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/docs/Description.kt
@@ -1,0 +1,7 @@
+package com.netflix.spinnaker.keel.api.docs
+
+import kotlin.annotation.AnnotationTarget.CLASS
+import kotlin.annotation.AnnotationTarget.PROPERTY
+
+@Target(CLASS, PROPERTY)
+annotation class Description(val value: String)
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/docs/Optional.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/docs/Optional.kt
@@ -1,0 +1,6 @@
+package com.netflix.spinnaker.keel.api.docs
+
+import kotlin.annotation.AnnotationTarget.PROPERTY
+
+@Target(PROPERTY)
+annotation class Optional
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/DeliveryArtifactModelConverter.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/DeliveryArtifactModelConverter.kt
@@ -1,0 +1,14 @@
+package com.netflix.spinnaker.keel.apidocs
+
+import com.netflix.spinnaker.keel.api.artifacts.DebianArtifact
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.artifacts.DockerArtifact
+import org.springframework.stereotype.Component
+
+@Component
+class DeliveryArtifactModelConverter : SubtypesModelConverter<DeliveryArtifact>(DeliveryArtifact::class.java) {
+  override val subTypes = listOf(
+    DebianArtifact::class.java,
+    DockerArtifact::class.java
+  )
+}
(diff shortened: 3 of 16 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/ExternalSchemaModelConverter.kt
```diff
@@ -0,0 +1,82 @@
...
+import io.swagger.v3.core.converter.AnnotatedType
+import io.swagger.v3.core.converter.ModelConverter
+import io.swagger.v3.core.converter.ModelConverterContext
+import io.swagger.v3.oas.models.media.ComposedSchema
+import io.swagger.v3.oas.models.media.Schema
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+
+@Component
+class ExternalSchemaModelConverter : BaseModelConverter() {
```
robfletcher: This can load a `Schema` object from `src/main/resources` then merge it with the schema that's been auto-generated. Useful for adding external documentation or overriding specific things.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -43,7 +43,11 @@ class DeliveryConfigController(
     consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
     produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
   )
-  fun upsert(@RequestBody deliveryConfig: SubmittedDeliveryConfig): DeliveryConfig =
+  fun upsert(
+    @RequestBody
+    @io.swagger.v3.oas.annotations.parameters.RequestBody(description = "The delivery config. If its `name` matches an existing delivery config the operation is an update, otherwise a new delivery config is created.")
+    deliveryConfig: SubmittedDeliveryConfig
+  ): DeliveryConfig =
```
robfletcher: Example of using swagger annotations on a controller. Unfortunate that here it has the same name as the Spring one.

robfletcher: I guess we could create our own annotation that combines the two

robfletcher: Actually using our own annotations in general might be a good way round the problem of not wanting to pull dependencies into `keel-api`

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/DeliveryArtifactModelConverter.kt
```diff
@@ -0,0 +1,14 @@
+package com.netflix.spinnaker.keel.apidocs
+
+import com.netflix.spinnaker.keel.api.artifacts.DebianArtifact
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.artifacts.DockerArtifact
+import org.springframework.stereotype.Component
+
+@Component
+class DeliveryArtifactModelConverter : SubtypesModelConverter<DeliveryArtifact>(DeliveryArtifact::class.java) {
```
luispollo: I thought I saw this in the previous PR? 🤔

robfletcher: nope, just resource specs and constraints

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/docs/Description.kt
```diff
@@ -0,0 +1,7 @@
+package com.netflix.spinnaker.keel.api.docs
+
+import kotlin.annotation.AnnotationTarget.CLASS
+import kotlin.annotation.AnnotationTarget.PROPERTY
+
+@Target(CLASS, PROPERTY)
+annotation class Description(val value: String)
```
robfletcher: Our own annotation we can use to add a description to models in the API without relying on Swagger dependency

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/docs/Optional.kt
```diff
@@ -0,0 +1,6 @@
+package com.netflix.spinnaker.keel.api.docs
+
+import kotlin.annotation.AnnotationTarget.PROPERTY
+
+@Target(PROPERTY)
+annotation class Optional
```
robfletcher: Allows us to force override properties that look required but we handle in a special way

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -14,6 +14,7 @@ import com.netflix.spinnaker.keel.persistence.DeliveryConfigRepository
 import com.netflix.spinnaker.keel.persistence.ResourceRepository.Companion.DEFAULT_MAX_EVENTS
 import com.netflix.spinnaker.keel.yaml.APPLICATION_YAML_VALUE
 import io.swagger.v3.oas.annotations.Operation
+import io.swagger.v3.oas.annotations.parameters.RequestBody as SwaggerRequestBody
```
robfletcher: On controllers we can use standard Swagger annotations. Examples are only really useful there anyway so I didn't add anything to the custom annotations in the end.

Issue #818, opened 2020-02-20
Validate delivery config in ImportDeliveryConfig stage


luispollo:
Ideally, we would add a call from deck to keel to validate a delivery config, given info about the source control repo, so the user knows if there are problems even before the pipeline run. However, given that the stage is very fast to run, the first step will probably be to add the call to orca and output validation errors to the stage execution details.

PR #821, opened 2020-02-21
feat(api): Validation API and better parsing errors

### LINKED ISSUE #367: More precise/actionable exceptions when spec validation fails
When submitting a resource spec that is valid YAML, but _not a valid resource spec_, Keel sends back a pretty generic/unhelpful exception ([`Submitted resource with an incorrect structure, cannot convert to ${resource.kind} ${resource.spec}"`](https://github.com/spinnaker/keel/blob/8476be7592e36149a8f61dbb5ded3f2cb1a0d9af/keel-plugin/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResolvableResourceHandler.kt#L60)). For example, I ran into this message because the `image` resource kind requires an `application` field which I had omitted. This is compounded by the fact that on the gate side that exception gets tossed in the trash in favor of a generic retrofit exception object.

While I don't have an excellent technical solution to propose, I'm thinking roughly:

1. Exception messages that are capable of pointing to specific, actionable validation failures (e.g. `Missing field 'application'` or basic type info like `must be a string`)
2. The plumbing to get those validation failures in front of users at the gate level
3. Ideally, a way of doing item 1 that makes it easy to get basic validation for new/existing resources without a ton of extra work (e.g. how [K8s CRDs use OpenAPI](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#specifying-a-structural-schema) to do validation in a standard/tooling-friendly way)

luispollo:
Introduces the following:
1. Additional logic in `ExceptionHandler` when dealing with YAML/JSON parsing errors by leveraging details exposed by Jackson and translating them into a well-defined format that should be easy to parse in the UI/CLI.
2. A `/delivery-configs/validate` API that is basically a no-op but which, by delegating to Jackson to parse the request body as a `SubmittedDeliveryConfig`, takes advantage of the functionality in (1) to implement a very simple validation API.

Example response:
```yaml
message: "Cannot construct instance of `com.netflix.spinnaker.keel.core.api.SubmittedResource`,\
  \ problem: Cannot deserialize value of type `com.netflix.spinnaker.keel.test.DummyEnum`\
  \ from String \"wrong\": not one of the values accepted for Enum class: [VALUE]\n\
  \ at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: com.netflix.spinnaker.keel.test.DummyResourceSpec[\"\
  enumData\"])\n at [Source: UNKNOWN; line: -1, column: -1] (through reference chain:\
  \ java.util.HashSet[0]) (through reference chain: com.netflix.spinnaker.keel.core.api.SubmittedDeliveryConfig[\"\
  environments\"]->java.util.HashSet[0]->com.fasterxml.jackson.databind.node.ObjectNode[\"\
  resources\"])"
details:
  error: "invalid_type"
  message: "Cannot deserialize value of type `com.netflix.spinnaker.keel.test.DummyEnum`\
    \ from String \"wrong\": not one of the values accepted for Enum class: [VALUE]\n\
    \ at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: com.netflix.spinnaker.keel.test.DummyResourceSpec[\"\
    enumData\"])"
  location:
    line: -1
    column: -1
  path:
    field: "environments"
  - type: "array"
    index: 0
  - type: "com.fasterxml.jackson.databind.node.ObjectNode"
    field: "resources"
  - type: "array"
    index: 0
  - type: "test.spinnaker.netflix.com/v1/whatever"
    field: "enumData"
  pathExpression: ".environments[0].resources[0].enumData"
```

Closes #367 
Enablement for #818

### DIFF
--- a/keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/Resources.kt
+++ b/keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/Resources.kt
@@ -15,6 +15,7 @@ import com.netflix.spinnaker.keel.api.plugins.SupportedKind
 import com.netflix.spinnaker.keel.core.api.SubmittedResource
 import de.danielbechler.diff.inclusion.Inclusion.EXCLUDED
 import de.danielbechler.diff.introspection.ObjectDiffProperty
+import java.time.Duration
 import java.util.UUID
 
 const val TEST_API = "test.$SPINNAKER_API_V1"
@@ -137,7 +138,11 @@ data class DummyResourceSpec(
   override val id: String = randomString(),
   val data: String = randomString(),
   override val application: String = "fnord"
-) : ResourceSpec
+) : ResourceSpec {
+  val intData: Int = 1234
+  val boolData: Boolean = true
+  val timeData: Duration = Duration.ofMinutes(5)
+}
 
 data class DummyLocatableResourceSpec(
   override val id: String = randomString(),
--- a/keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/Resources.kt
+++ b/keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/Resources.kt
@@ -134,6 +134,8 @@ fun <T : ResourceSpec> submittedResource(
     spec = spec
   )
 
+enum class DummyEnum { VALUE }
+
 data class DummyResourceSpec(
   override val id: String = randomString(),
   val data: String = randomString(),
@@ -142,6 +144,7 @@ data class DummyResourceSpec(
   val intData: Int = 1234
   val boolData: Boolean = true
   val timeData: Duration = Duration.ofMinutes(5)
+  val enumData: DummyEnum = DummyEnum.VALUE
 }
 
 data class DummyLocatableResourceSpec(
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
@@ -87,8 +87,8 @@ class ExceptionHandler(
         else -> when (rootCause) {
           null -> ParsingError.OTHER
           is MissingKotlinParameterException -> ParsingError.MISSING_PROPERTY
-          is IllegalStateException -> ParsingError.ILLEGAL_VALUE
-          is IllegalArgumentException -> ParsingError.ILLEGAL_VALUE
+          is IllegalStateException -> ParsingError.INVALID_VALUE
+          is IllegalArgumentException -> ParsingError.INVALID_VALUE
           is MismatchedInputException -> ParsingError.INVALID_TYPE
           is InvalidTypeIdException -> ParsingError.INVALID_TYPE
           is InvalidFormatException -> ParsingError.INVALID_FORMAT
@@ -162,9 +162,9 @@ data class ApiError(
 
 enum class ParsingError {
   MISSING_PROPERTY,
-  ILLEGAL_VALUE,
   INVALID_TYPE,
   INVALID_FORMAT,
+  INVALID_VALUE, // used when we can't determine if the problem is type or format
   OTHER;
 
   @JsonValue
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
@@ -88,10 +88,10 @@ class ExceptionHandler(
           null -> ParsingError.OTHER
           is MissingKotlinParameterException -> ParsingError.MISSING_PROPERTY
           is IllegalStateException -> ParsingError.ILLEGAL_VALUE
-          is InvalidFormatException -> ParsingError.ILLEGAL_VALUE
           is IllegalArgumentException -> ParsingError.ILLEGAL_VALUE
           is MismatchedInputException -> ParsingError.INVALID_TYPE
           is InvalidTypeIdException -> ParsingError.INVALID_TYPE
+          is InvalidFormatException -> ParsingError.INVALID_FORMAT
           is DateTimeParseException -> ParsingError.INVALID_FORMAT
           else -> ParsingError.OTHER
         }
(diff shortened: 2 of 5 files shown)


keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/Resources.kt
```diff
@@ -115,7 +116,11 @@ data class DummyResourceSpec(
   override val id: String = randomString(),
   val data: String = randomString(),
   override val application: String = "fnord"
-) : ResourceSpec
+) : ResourceSpec {
+  val intData: Int = 1234
+  val boolData: Boolean = true
+  val timeData: Duration = Duration.ofMinutes(5)
```
luispollo: Added these so I could use the dummy spec in the validation tests.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -58,12 +77,121 @@ class ExceptionHandler {
...
+        is MissingKotlinParameterException -> ParsingError.MISSING_PROPERTY
+        else -> when (rootCause) {
+          null -> ParsingError.OTHER
+          is MissingKotlinParameterException -> ParsingError.MISSING_PROPERTY
+          is IllegalStateException -> ParsingError.ILLEGAL_VALUE
+          is InvalidFormatException -> ParsingError.ILLEGAL_VALUE
+          is IllegalArgumentException -> ParsingError.ILLEGAL_VALUE
+          is MismatchedInputException -> ParsingError.INVALID_TYPE
+          is InvalidTypeIdException -> ParsingError.INVALID_TYPE
+          is DateTimeParseException -> ParsingError.INVALID_FORMAT
```
luispollo: I realize the mapping here is not perfect, but it was tricky because exception types are not consistently used (e.g. we have validation that throws `IllegalStateException` for missing required fields), so I did the best I could. Please do suggest changes if you can think of better mappings.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -58,12 +77,121 @@ class ExceptionHandler {
...
+  private fun JsonMappingException.findRootCause(): Pair<Throwable, List<JsonMappingException.Reference>> {
+    if (this.cause == null) {
+      return Pair(this, this.path)
+    }
+    var exception: Throwable = this
+    var paths: MutableList<JsonMappingException.Reference> = this.path.toMutableList()
+    while (exception.cause != null && exception.cause != exception) {
+      exception = exception.cause!!
+      if (exception is JsonMappingException) {
+        paths.addAll(exception.path)
```
luispollo: This builds a flat list of all the paths leading to the leaf node/property with the problem.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -58,12 +77,121 @@ class ExceptionHandler {
...
+  val message: String?,
+  val details: ApiErrorDetails?
+) {
+  constructor(ex: Throwable, details: ApiErrorDetails? = null) :
+    this(ex.cause?.message ?: ex.message, details)
+}
+
+enum class ParsingError {
+  MISSING_PROPERTY,
+  ILLEGAL_VALUE,
```
gal-yardeni: What does `illegal` means? maybe it should be `invalid` or `unsupported`?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandlerTests.kt
```diff
@@ -0,0 +1,275 @@
...
+import com.netflix.spinnaker.keel.test.TEST_API
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import strikt.api.expect
+import strikt.assertions.getValue
+import strikt.assertions.isA
+import strikt.assertions.isEqualTo
+import strikt.assertions.size
+
+class ExceptionHandlerTests : JUnit5Minutests {
```
gal-yardeni: Love these tests!

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
luispollo: I guess I was thinking of `IllegalArgumentException` and similar exceptions. I don't mind changing it to `INVALID_VALUE`.

emjburns:
@coryb I'd love your feedback on the response from this validation endpoint. Is this data useful? Is this format something that's easy for you to parse?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandlerTests.kt
```diff
@@ -0,0 +1,309 @@
...
+              get { error }.isEqualTo(ParsingError.INVALID_TYPE)
+              get { path }.size.isEqualTo(3)
+              get { pathExpression }.isEqualTo(".environments[0].constraints")
+            }
+          }
+        }
+      }
+    }
+
+    context("resource missing field checked by validation") {
```
asher: This context name makes a good point, can you add a comment somewhere here or in a github issue documenting the gaps in getting the new `/delivery-configs/validate` endpoint to fully validate resources?

luispollo: Thanks for the feedback, @asher. I feel like most of this will go away when @robfletcher is done with the JSON schema stuff. I expect we'll replace my poor-man's validation with a check against the schema. Is your point that, even then, there might be validations that are _not_ part of the model definition and might need to be checked on the side?

asher: Complete resource validation has to happen after resolvers are applied but with that in place, json schema validation should be sufficient.

luispollo: OK. I already had a `TODO` in there to replace the `ApiError` validation with a check using the schema.

luispollo:
@coryb Please don't be discouraged from commenting because the PR is merged. I wanted to try to use this response myself in the Import Delivery Config stage in Spinnaker, which will help figure out if the format is helpful, but would love another data point.

Issue #740, opened 2020-01-21
Rethink repository structure in code


emjburns:
We currently have three main repositories (artifact, resource, and delivery config). These grew organically as we added delivery configs. Now that we no longer support resources w/o delivery configs we should reconsider how things are structured. Since you can never have a resource without a delivery config you shouldn't have to wire in both and do the orchestration yourself in various places.

I propose constructing a general repository that contains the existing repositories and orchestrates saving, updating, and deleting. This is mostly a kotlin-level change in how we interact with the storage layer in `keel`.

luispollo:
Would this also involve removing the `ResourcePersister` and `Cleaner` classes?

emjburns:
🤷‍♀ yes probably.

luispollo:
As requested, reporting a SQL-related error that we might be able to solve as part of this refactoring:
```
com.netflix.spinnaker.keel.plugin.CannotResolveDesiredState: Unable to resolve desired state of titus:cluster:titustestvpc:mddemotitus due to: Cursor returned more than one result
   at com.netflix.spinnaker.keel.actuation.ResourceActuator$resolve$2$desired$1.invokeSuspend(ResourceActuator.kt:130)
   at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
   at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
   at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:561)
   at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:727)
   at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:667)
   at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:655)
Caused by: org.jooq.exception.TooManyRowsException: Cursor returned more than one result
   at org.jooq_3.12.3.MYSQL.debug(Unknown Source)
   at org.jooq.impl.Tools.fetchOne(Tools.java:2087)
   at org.jooq.impl.AbstractResultQuery.fetchOne(AbstractResultQuery.java:653)
   at org.jooq.impl.AbstractResultQuery.fetchOne(AbstractResultQuery.java:658)
   at org.jooq.impl.SelectImpl.fetchOne(SelectImpl.java:2892)
[...]

PR #824, opened 2020-02-24
Unify repositories

emjburns:
Fixes https://github.com/spinnaker/keel/issues/740.

This is a stab at creating one repository to rule them all. This first PR is no change in logic, just moving things around.

The `CombinedRepository` contains a `DeliveryConfigRepository`, an `ArtifactRepository`, and a `ResourceRepository`. The methods on the combined repository mostly proxy through to the individual methods on each repo. In some cases, they also emit events. In the future, there will be metrics around the calls. 

The `ResourcePersister` and `ResourceCleaner` are gone. That functionality is encapsulated in the `CombinedRepository`. Their tests have moved to the `CombinedRepositoryTests`.

I'm working on figuring out why the transaction tests are failing, but I'm putting this PR up to get feedback this approach. There are no functional changes in this PR. Most of the changes are imports in many files.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -0,0 +1,386 @@
...
+ */
+@Component
+class CombinedRepository(
+  val deliveryConfigRepository: DeliveryConfigRepository,
+  val artifactRepository: ArtifactRepository,
+  val resourceRepository: ResourceRepository,
+  val clock: Clock,
+  val publisher: ApplicationEventPublisher
+) {
+
```
emjburns: This is the big change: one repository to rule them all! As a first step, how does this seem? I know there's a lot of proxying of methods, but it gives us a great place to emit events (like `ResourceCreated`, etc) & capture metrics.

It also gives us a place to change the implementation & repository structure in the future so that we can have better support for transactions and correctness there.

It also makes it easier to create/use tests: you don't have to figure out which repos you need to wire in, you just wire in this one. I've also created a test util to create a mock repo or an in memory repo

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  private fun Set<DeliveryArtifact>.transform(deliveryConfigName: String) =
+    map { artifact ->
+      when (artifact) {
+        is DockerArtifact -> artifact.copy(deliveryConfigName = deliveryConfigName)
+        is DebianArtifact -> artifact.copy(deliveryConfigName = deliveryConfigName)
+      }
+    }.toSet()
+
```
emjburns: There's a problem in this function `upsertDeliveryConfig(deliveryConfig: DeliveryConfig)` - we're not actually rolling back when there's a problem (or, the test is doing the wrong thing now that we have different imports). Would love some eyes on this.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -0,0 +1,344 @@
...
+import strikt.assertions.failed
+import strikt.assertions.first
+import strikt.assertions.hasSize
+import strikt.assertions.isA
+import strikt.assertions.isEmpty
+import strikt.assertions.isEqualTo
+import strikt.assertions.isTrue
+import strikt.assertions.succeeded
+
+abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository> :
```
robfletcher: Does it make sense to combine all these tests? I wonder if we could have kept them as separate test classes that each tested a particular aspect of the combined repo.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/CheckSchedulerTests.kt
```diff
@@ -21,6 +23,7 @@ internal object CheckSchedulerTests : JUnit5Minutests {
 
   private val resourceRepository = mockk<ResourceRepository>()
   private val deliveryConfigRepository = mockk<DeliveryConfigRepository>()
+  private val combinedRepository: CombinedRepository = combinedInMemoryRepository(deliveryConfigRepository = deliveryConfigRepository, resourceRepository = resourceRepository)
```
robfletcher: It seems weird that the test has to rely on all these things. It would make more sense to just mock the combined repository and ignore the underlying implementation.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
emjburns: I wanted to test both sql and non-sql for create & delete flows, so I thought it made sense to pull them both here. Would you prefer I split them out? If so, can you suggest what categories you'd like? I agree that small test files are easier to reason about.

emjburns: There actually aren't that many test in there, just create + verify that we start checking things + delete + validation. And, they all care about all three repositories. So I'm inclined to leave them in the same file for now, and split them up if they get unwieldy.

robfletcher: Okay, maybe I misunderstood what I was seeing. It looked like you merged all the tests for the other repositories

emjburns: Oh no not at all. Those are all separate. This is just the muli-repo create + delete tests. Let me add a comment clarifying that.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintEvaluator.kt
```diff
@@ -57,7 +57,7 @@ import org.springframework.stereotype.Component
 class CanaryConstraintEvaluator(
   private val handlers: List<CanaryConstraintDeployHandler>,
   private val orcaService: OrcaService,
-  override val deliveryConfigRepository: DeliveryConfigRepository,
+  override val combinedRepository: CombinedRepository,
```
asher: I'd ❤️ to see this variable just called `repository` throughout every file but nbd.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -0,0 +1,387 @@
...
+ * This paves the way for re-thinking how we interact with sql/storing of resources,
+ * and updating our internal repository structure to allow storing delivery configs to more easily also
+ * store artifacts and resources.
+ *
+ * This also gives us an easy place to emit telemetry and events around the usage of methods.
+ *
+ * TODO eb: refactor repository interaction so transactionality is easier.
+ */
+@Component
+class CombinedRepository(
```
asher: Consider making a `KeelRepository` interface covering all of the public methods here, and use that type to inject `CombinedRepository` everywhere needed.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintEvaluator.kt
emjburns: I was wondering if that would be confusing, but since both you and Gal suggested it I'll make the change!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
emjburns: Done!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
```diff
@@ -0,0 +1,228 @@
...
+  val publisher: ApplicationEventPublisher
+  val log: Logger
+
+  @Transactional(propagation = Propagation.REQUIRED)
+  fun upsertDeliveryConfig(submittedDeliveryConfig: SubmittedDeliveryConfig): DeliveryConfig
+
+  /**
+   * Validates that resources have unique ids, throws an exception if invalid
+   */
+  fun validate(config: DeliveryConfig) {
```
asher: This seems out of place on the interface, it mixes up model concerns with storage. It looks like it doesn't depend on other repository methods unless I'm missing something, so maybe it could be a top-level extension function in DeliveryConfig.kt or somewhere adjacent.

Otherwise LGTM, great work!

emjburns: Ah good point. I'll fix!

PR #830, opened 2020-02-25
fix(veto): fix long overflow impacting clusters

asher:


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepository.kt
```diff
@@ -65,7 +65,7 @@ abstract class UnhappyVetoRepository(
 
   fun calculateExpirationTime(wait: Duration): Instant =
     when (wait) {
-      Duration.ZERO -> Instant.MAX
+      Duration.ZERO -> Instant.ofEpochMilli(Long.MAX_VALUE)
```
asher: It surprisingly turns out that `Instant.MAX.toEpochMilli()` throws:
```
java.lang.ArithmeticException: long overflow
        at java.lang.Math.multiplyExact(Math.java:892)
        at java.time.Instant.toEpochMilli(Instant.java:1237)
```

luispollo: If we want this to mean "forever" as described, should we maybe parse as seconds instead of millis?

PR #840, opened 2020-02-28
fix(constraints): process old pending constraints, fix concurrency bug

### LINKED ISSUE #858: Adding pipeline constraint runs unnecessary pipeline executions
Adding a pipeline constraint on prod (which is a few versions behind), and that suddenly created a bunch of pipeline executions, I think this is because it’s suddenly running the check for all possible newer versions for prod.  Oddly a few moments later they almost all were suddenly cancelled.

### LINKED ISSUE #859: Canary constraint runs on irrelevant versions
Canary started on the latest version to deploy, it passed, but then it started on the next previous version (there was a gap in releases between my test and prod cluster).

### LINKED ISSUE #860: Repeated deployment of the same artifact
When I added my staging environment cluster (after removing the constraint in bug #2), it rolled out the same change multiple times (see the two SGs in the prod cluster).

image

asher:
Fixes #858 
Fixes #859 
Fixes #860

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
@@ -2,12 +2,15 @@ package com.netflix.spinnaker.keel.actuation
 
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.api.anyStateful
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.constraints.ConstraintEvaluator
+import com.netflix.spinnaker.keel.constraints.ConstraintStatus.PENDING
 import com.netflix.spinnaker.keel.constraints.StatefulConstraintEvaluator
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVetoes
 import com.netflix.spinnaker.keel.core.api.PinnedEnvironment
-import com.netflix.spinnaker.keel.persistence.ArtifactRepository
+import com.netflix.spinnaker.keel.core.comparator
+import com.netflix.spinnaker.keel.persistence.KeelRepository
 import com.netflix.spinnaker.keel.telemetry.ArtifactVersionApproved
 import org.slf4j.LoggerFactory
 import org.springframework.context.ApplicationEventPublisher
@@ -15,7 +18,7 @@ import org.springframework.stereotype.Component
 
 @Component
 class EnvironmentPromotionChecker(
-  private val artifactRepository: ArtifactRepository,
+  private val repository: KeelRepository,
   private val constraints: List<ConstraintEvaluator<*>>,
   private val publisher: ApplicationEventPublisher
 ) {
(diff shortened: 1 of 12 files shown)


asher:
This PR attempts to fix the following environment constraint bugs:

- If an environment with at least one `StatefulConstraint` has no approved versions and keel knows of many versions for the related `DeliveryArtifact`, constraint evaluators are invoked nearly concurrently for every approved version.
  - New behavior: once a version is found with pending constraints, version iteration is short-circuited in `EnvironmentPromotionChecker`. 
- `PENDING` constraints would be left pending (unless manually overridden) indefinitely if a newer artifact version passes first.
  - New behavior: `PENDING` constraints for an environment are always checked during promotion actuation so they can be timed out or approved.
- If a `manual-judgement` is pending concurrently for v1 and v2, and a user approves v2, then v1, v1 is never actually approved for the environment.
  - New behavior: Whenever a `ConstraintState` is stored with a passing status, keel checks if all other stateful constraints have passed. If so, it is queued for approval, gated behind stateless constraints passing at dequeue time. If they don't pass (i.e. there's an `allowed-times` constraint and we're not in the allowed window), the version remains queued for approval.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionCheckerTests.kt
```diff
@@ -319,7 +357,129 @@ internal class EnvironmentPromotionCheckerTests : JUnit5Minutests {
...
+          every { statefulEvaluator.canPromote(artifact, "2.0", deliveryConfig, environment) } returns false
+
+          every {
+            repository.constraintStateFor("my-manifest", "staging", "2.0")
+          } returns listOf(pendingManualJudgement)
+
+          runBlocking { subject.checkEnvironments(deliveryConfig) }
+        }
+
+        test("stateful constraints are only evaluated for the most recent version") {
```
asher: This covers the "adding a new environment (thus no approved versions) with a pipeline constraint attempts to launch a concurrent pipeline for every known artifact version" bug.

asher:
cc: @lorin

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryDeliveryConfigRepository.kt
```diff
@@ -94,13 +95,22 @@ class InMemoryDeliveryConfigRepository(
...
       .add(stateId)
+
+    val allPass = constraints
+      .filterKeys { it.startsWith(idPrefix) }
+      .values
+      .all { it.status.passes() }
+
+    if (allPass) {
+      queueAllConstraintsApproved(state.deliveryConfigName, state.environmentName, state.artifactVersion)
+    }
```
asher: This covers scenarios such as submitting a manual-judgement for an older artifact version when there's already a newer version approved. It also ensures that manually overriding old failed judgement to passed results in artifact approval.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -266,8 +266,16 @@ class SqlArtifactRepository(
...
           .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.ne(VETOED.name))
           .orderBy(ENVIRONMENT_ARTIFACT_VERSIONS.APPROVED_AT.desc())
-          .limit(1)
-          .fetchOne(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
+          .limit(20)
+          .fetch(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
+          .let {
+            when (it.isNotEmpty()) {
+              true -> it
+                .sortedWith(artifact.versioningStrategy.comparator)
```
asher: This changes `latestVersionApprovedIn` from returning the most recently approved version (which may not be the newest version, approval isn't necessarily linear!) to the most recent version within the set of the most recently approved 20 versions, according to the artifacts versioning strategy.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -53,51 +76,105 @@ class EnvironmentPromotionChecker(
...
+             * so they can be timed out, failed, or approved. If a newer version was selected above,
+             * it is approved last in an attempt to retain ENVIRONMENT_ARTIFACT_VERSIONS.APPROVED_AT
+             * ordering.
+             */
+            var approvedPending = false
+            if (!hasPin) {
+              pendingVersionsToCheck
+                .sortedWith(artifact.versioningStrategy.comparator.reversed()) // oldest first
+                .forEach {
+                  if (checkConstraints(statefulEvaluators, artifact, deliveryConfig, it, environment)) {
```
luispollo: Why do we skip checking the stateless constraints here?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionCheckerTests.kt
```diff
@@ -319,7 +357,129 @@ internal class EnvironmentPromotionCheckerTests : JUnit5Minutests {
...
+      context("a new artifact passes stateful constraints while older versions are pending") {
+        deriveFixture {
+          copy(environment = Environment(
+            name = "staging",
+            constraints = setOf(DependsOnConstraint("test"), ManualJudgementConstraint())
+          ))
+        }
+
+        before {
+          // TODO: sucks that this is necessary but when using deriveFixture you get a different mockk
```
luispollo: The docs say you can use `modifyFixture` with `parentFixture` to modify the original fixture, which should include the mocks if that's where they're defined.

asher: Nice!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
asher: Great catch, the comment bellow about why stateless constraints are re-checked for `queuedForApproval` versions does apply here as well. Will fix.

PR #841, opened 2020-02-28
feat(repository): disallow duplicate artifact refs

### LINKED ISSUE #761: Ensure that artifact reference is unique across submitted delivery configs
In order to ensure that artifact `reference` is unique within a delivery config, we should have some constructor validation on the delivery config to ensure that we don't have duplicate references.

We could either allow unique by reference, or unique by reference and type. Probably reference is better.

lorin:
Enforce that all artifacts in a delivery configuration have a unique reference.

Fixes #761.

If a user attempts to submit a delivery configuration that violates this, the response payload looks like this:

```
---
message: "Multiple artifacts are using the same reference: {lhochstein/lorintesttitus=my-artifact,\
  \ emburns/spin-titus-demo=my-artifact}. Please ensure each artifact has a unique\
  \ reference."
```

~(As an aside, I don't think keel should return a 500 error on input that fails validation, but that's the current behavior).~ (*My mistake: returning 500 was not the previous behavior on a validation error*)

It will also log a similar error message at the ERROR level:

```
Multiple artifacts are using the same reference: {lhochstein/lorintesttitus=my-artifact, emburns/spin-titus-demo=my-artifact}. Please ensure each artifact has a unique reference.
```

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
@@ -16,6 +16,7 @@ import com.netflix.spinnaker.keel.core.api.resources
 import com.netflix.spinnaker.keel.events.ArtifactRegisteredEvent
 import com.netflix.spinnaker.keel.events.ResourceCreated
 import com.netflix.spinnaker.keel.events.ResourceUpdated
+import com.netflix.spinnaker.keel.exceptions.DuplicateArtifactReferenceException
 import com.netflix.spinnaker.keel.exceptions.DuplicateResourceIdException
 import com.netflix.spinnaker.keel.resources.ResourceTypeIdentifier
 import com.netflix.spinnaker.keel.test.DummyResourceSpec
@@ -304,7 +305,7 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
 
     context("persisting delivery config manifests") {
 
-      context("a delivery config with non-unique resource ids errors while persisting ") {
+      context("a delivery config with non-unique resource ids errors while persisting") {
         val submittedConfig = SubmittedDeliveryConfig(
           name = configName,
           application = "keel",
@@ -340,6 +341,43 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
             subject.upsertDeliveryConfig(submittedConfig)
           }.failed()
             .isA<DuplicateResourceIdException>()
+
+          expectThat(subject.allResourceNames().size).isEqualTo(0)
+        }
+      }
+
+      context("a delivery config with non-unique artifact references errors while persisting") {
+        // Two different artifacts with the same reference
+        val artifacts = setOf(
+          DockerArtifact(name = "org/thing-1", deliveryConfigName = configName, reference = "thing"),
+          DockerArtifact(name = "org/thing-2", deliveryConfigName = configName, reference = "thing")
+        )
+
+        val submittedConfig = SubmittedDeliveryConfig(
+          name = configName,
+          application = "keel",
+          serviceAccount = "keel@spinnaker",
+          artifacts = artifacts,
+          environments = setOf(
+            SubmittedEnvironment(
+              name = "test",
+              resources = setOf(
+                SubmittedResource(
+                  metadata = mapOf("serviceAccount" to "keel@spinnaker"),
+                  apiVersion = "test.$SPINNAKER_API_V1",
+                  kind = "whatever",
+                  spec = DummyResourceSpec(data = "o hai")
+                )
+              ),
+              constraints = emptySet()
+            )
+          )
+        )
+        test("an error is thrown and config is deleted") {
+          expectCatching {
+            subject.upsertDeliveryConfig(submittedConfig)
+          }.failed()
+            .isA<DuplicateArtifactReferenceException>()
 
           expectThat(subject.allResourceNames().size).isEqualTo(0)
         }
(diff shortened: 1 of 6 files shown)


lorin:
@luispollo @asher

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -156,25 +157,53 @@ class CombinedRepository(
...
+   * - resources have unique ids
+   * - artifacts have unique references
+   *
+   * Throws an exception if config fails any checks
    */
   private fun validate(config: DeliveryConfig) {
+
+    // helper function to get duplicates in a list
+    val duplicates: (List<String>) -> List<String> =
+      { ls -> ls.groupingBy { it }.eachCount().filter { it.value > 1 }.keys.toList() }
```
asher: I had trouble groking this at first glance. I think it's more readable as a `fun` with vertical spacing like:
```suggestion
    fun duplicates(resourceIds: List<String>) =
    	resourceIds.groupingBy { it }
        .eachCount()
        .filter { it.value > 1 }
        .keys
        .toList()
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -156,25 +157,53 @@ class CombinedRepository(
...
+    /**
+     * check: artifacts have unique references
+     */
+    val refs = config.artifacts.map { it.reference }
+    val duplicateRefs = duplicates(refs)
+
+    if (duplicateRefs.isNotEmpty()) {
+      val duplicatesArtifactNameToRef: Map<String, String> = config.artifacts
+        .filter { duplicateRefs.contains(it.reference) }
+        .map { art -> art.name to art.reference }.toMap()
```
asher: Instead of combining `.map` and `toMap()`, you can use `.associate { art -> art.name to art.reference }`

lorin: TIL!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
lorin: Addressed in 1bea51c.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
lorin: Addressed in 1bea51c.

lorin:
Rebased against master

luispollo:
> (As an aside, I don't think keel should return a 500 error on input that fails validation, but that's the current behavior).

I agree. In general we've bee using 400 in that case. Take a look at the `ExceptionHandler` class -- it converts known exception types into the desired status code/response. You might need to add another exception type to the 400 case.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -16,6 +16,7 @@ import com.netflix.spinnaker.keel.core.api.resources
 import com.netflix.spinnaker.keel.events.ArtifactRegisteredEvent
 import com.netflix.spinnaker.keel.events.ResourceCreated
 import com.netflix.spinnaker.keel.events.ResourceUpdated
+import com.netflix.spinnaker.keel.exceptions.DuplicateArtifactReferenceException
```
luispollo: As I mentioned in the other comment, if you add this exception to this handler, then it will cause the API to return a 400:
https://github.com/spinnaker/keel/blob/4c5a4852401eee558709ac987703ab2e178ad771/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt#L40-L42

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/DuplicateArtifactReferenceException.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.keel.exceptions
+
+class DuplicateArtifactReferenceException(
+  private val artifactNameToRef: Map<String, String>
+) : ValidationException(
+  "Multiple artifacts are using the same string as a reference: $artifactNameToRef. " +
```
luispollo:
```suggestion
  "Multiple artifacts are using the same reference: $artifactNameToRef. " +
```

luispollo: Just simplifying a bit ☝️

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/DuplicateArtifactReferenceException.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.keel.exceptions
+
+class DuplicateArtifactReferenceException(
```
luispollo: As I mentioned in the other comment, if you add this exception (or perhaps the parent `ValidationException`) to the handler below, then it will cause the API to return a 400:
https://github.com/spinnaker/keel/blob/4c5a4852401eee558709ac987703ab2e178ad771/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt#L40-L42

lorin: Addressed in 8f805d44.

lorin:
rebase against master

luispollo:
P.S. It would be nice to add a test case to `ExceptionHandlerTests` to cover the new handling of `ValidationException` in another PR. Sorry I didn't catch this earlier.

PR #842, opened 2020-02-29
fix(events): Fix paused events for resources due to pausing application

### LINKED ISSUE #634: Define application or environment-level events
We currently have event types inheriting from `ResourceEvent` and `ArtifactEvent` but also need at least one higher level event types. I.e. constraints are evaluated at the environment level; a `PendingConstraintEvent` could be any of `ApplicationEvent`, `DeliveryConfigEvent`, or `EnvironmentEvent`. Perhaps we only actually need an `ApplicationEvent` type however?

### LINKED ISSUE #837: Find an alternative approach to adding ResourceActuationPaused events to history
Right now Keel [prepends a ResourceActuationPaused event onto the history API response when a resource is paused](https://github.com/spinnaker/keel/blob/b8d0711bb2cb4b1ea50c5c59a313cfadaf874429/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt#L40-L50). This gets _super confusing_ because it appears as if it happens immediately at the moment you chose to view the history. When I first saw this I got distracted trying to figure out what on earth was going on, because as far as I knew it had been paused for a while. It also means that we completely lose/mixup the value of the history events, which is to give you a sense of when things happened in relation to each other. Can we look up the timestamp of when things were paused, iterate through the list of events to find the right spot and splice it in where a user would expect in the event list? Here’s how it actually looks in practice (this app was paused more than a week ago):

Screen Shot 2020-02-28 at 11 47 07 AM

Regardless of implementation, I'd like to see ResourceActuationPaused events look like ResourceActuationResumed — sitting in the appropriate spot in the list with a stable timestamp that doesn't change between API calls. This also makes it incredibly difficult to do polling and fetch new data in the UI, because it means we can't rely on previously fetch events being immutable.

luispollo:
Introduces the following changes:
- Generalizes the event model so it can be used for other "scopes" beyond resources
- Modifies the database structure to repurpose the existing `resource_event` table into a generic `event` table as per the above
- Introduces a new abstract `ApplicationEvent` base class for application-level events
- Introduces new `ApplicationActuationPaused` and `ApplicationActuationResumed` events
- Modifies the logic in the `EventController` to add a matching and "immutable" `ResourceActuationPaused` event for every `ApplicationActuationPaused` in the event history in the right position of the timestamp-ordered list

Sample table records:
```
scope        uid                          type                           timestamp
RESOURCE     01E26DWCV4STWP39WQ06RPA8AJ   "ResourceValid"                2020-02-29 02:24:57
RESOURCE     01E26DWCV4STWP39WQ06RPA8AJ   "ResourceActuationResumed"     2020-02-29 02:24:40
APPLICATION  serverlablpollo              "ApplicationActuationResumed"  2020-02-29 02:24:37
APPLICATION  serverlablpollo              "ApplicationActuationPaused"   2020-02-29 02:23:57
RESOURCE     01E26DWCV4STWP39WQ06RPA8AJ   "ResourceValid"                2020-02-29 01:48:00
RESOURCE     01E26DWCV4STWP39WQ06RPA8AJ   "ResourceActuationResumed"     2020-02-29 01:47:52
APPLICATION  serverlablpollo              "ApplicationActuationResumed"  2020-02-29 01:47:48
APPLICATION  serverlablpollo              "ApplicationActuationPaused"   2020-02-29 01:39:04
RESOURCE     01E26DWCV4STWP39WQ06RPA8AJ   "ResourceValid"                2020-02-28 17:35:42
RESOURCE     01E26DWCV4STWP39WQ06RPA8AJ   "ResourceCreated"              2020-02-28 17:35:37
```
Matching API response:
```yaml
---
- type: "ResourceValid"
  apiVersion: "titus.spinnaker.netflix.com/v1"
  kind: "cluster"
  id: "titus:cluster:titustestvpc:serverlablpollo"
  application: "serverlablpollo"
  timestamp: "2020-02-29T02:24:56.684Z"
  scope: "RESOURCE"
- type: "ResourceActuationResumed"
  apiVersion: "titus.spinnaker.netflix.com/v1"
  kind: "cluster"
  id: "titus:cluster:titustestvpc:serverlablpollo"
  application: "serverlablpollo"
  timestamp: "2020-02-29T02:24:39.906Z"
  scope: "RESOURCE"
- type: "ResourceActuationPaused"
  apiVersion: "titus.spinnaker.netflix.com/v1"
  kind: "cluster"
  id: "titus:cluster:titustestvpc:serverlablpollo"
  application: "serverlablpollo"
  reason: "Resource actuation paused at the application level"
  timestamp: "2020-02-29T02:23:57.265Z"
  scope: "RESOURCE"
- type: "ResourceValid"
  apiVersion: "titus.spinnaker.netflix.com/v1"
  kind: "cluster"
  id: "titus:cluster:titustestvpc:serverlablpollo"
  application: "serverlablpollo"
  timestamp: "2020-02-29T01:47:59.870Z"
  scope: "RESOURCE"
- type: "ResourceActuationResumed"
  apiVersion: "titus.spinnaker.netflix.com/v1"
  kind: "cluster"
  id: "titus:cluster:titustestvpc:serverlablpollo"
  application: "serverlablpollo"
  timestamp: "2020-02-29T01:47:51.579Z"
  scope: "RESOURCE"
- type: "ResourceActuationPaused"
  apiVersion: "titus.spinnaker.netflix.com/v1"
  kind: "cluster"
  id: "titus:cluster:titustestvpc:serverlablpollo"
  application: "serverlablpollo"
  reason: "Resource actuation paused at the application level"
  timestamp: "2020-02-29T01:39:03.582Z"
  scope: "RESOURCE"
- type: "ResourceValid"
  apiVersion: "titus.spinnaker.netflix.com/v1"
  kind: "cluster"
  id: "titus:cluster:titustestvpc:serverlablpollo"
  application: "serverlablpollo"
  timestamp: "2020-02-28T17:35:42.055Z"
  scope: "RESOURCE"
- type: "ResourceCreated"
  apiVersion: "titus.spinnaker.netflix.com/v1"
  kind: "cluster"
  id: "titus:cluster:titustestvpc:serverlablpollo"
  application: "serverlablpollo"
  timestamp: "2020-02-28T17:35:37.426Z"
  scope: "RESOURCE"
```

Closes #837 
Closes #634

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
@@ -270,34 +270,6 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
         }
       }
 
-      context("deleting resource by application name") {
-        before {
-          subject.store(resource)
-          subject.deleteByApplication(resource.application)
-        }
-
-        test("the resource is no longer returned when listing all resources") {
-          subject.allResources(callback)
-          verify(exactly = 0) { callback(any()) }
-        }
-
-        test("the resource can no longer be retrieved by name") {
-          expectThrows<NoSuchResourceException> {
-            subject.get<DummyResourceSpec>(resource.id)
-          }
-        }
-
-        test("events for the resource are also deleted") {
-          expectThrows<NoSuchResourceException> {
-            subject.eventHistory(resource.id)
-          }
-        }
-
-        test("deleting a non-existent application throws an exception") {
-          expectThat(subject.deleteByApplication(resource.application)).isEqualTo(0)
-        }
-      }
-
       context("fetching event history for the resource") {
         before {
           repeat(3) {
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
@@ -24,6 +24,7 @@ import com.netflix.spinnaker.keel.events.ResourceDeltaDetected
 import com.netflix.spinnaker.keel.events.ResourceDeltaResolved
 import com.netflix.spinnaker.keel.events.ResourceEvent
 import com.netflix.spinnaker.keel.events.ResourceUpdated
+import com.netflix.spinnaker.keel.events.ResourceValid
 import com.netflix.spinnaker.keel.test.DummyResourceSpec
 import com.netflix.spinnaker.keel.test.locatableResource
 import com.netflix.spinnaker.keel.test.resource
@@ -201,28 +202,27 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
           context("updating the state again") {
             before {
               tick()
-              // TODO: ensure persisting a map with actual data
-              subject.appendHistory(ResourceDeltaDetected(resource, emptyMap(), clock))
+              subject.appendHistory(ResourceValid(resource, clock))
             }
 
             test("the event is included in the history") {
               expectThat(subject.eventHistory(resource.id))
                 .hasSize(3)
                 .first()
-                .isA<ResourceDeltaDetected>()
+                .isA<ResourceValid>()
             }
 
             context("a subsequent identical event that should be ignored") {
               before {
                 tick()
-                subject.appendHistory(ResourceDeltaDetected(resource, emptyMap(), clock))
+                subject.appendHistory(ResourceValid(resource, clock))
               }
 
               test("the event is not included in the history") {
                 expectThat(subject.eventHistory(resource.id))
                   .and {
-                    first().isA<ResourceDeltaDetected>()
-                    second().not().isA<ResourceDeltaDetected>()
+                    first().isA<ResourceValid>()
+                    second().not().isA<ResourceValid>()
                   }
               }
             }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
@@ -1,3 +1,60 @@
 package com.netflix.spinnaker.keel.events
 
-class ApplicationEvent
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonSubTypes
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import java.time.Instant
+
+@JsonTypeInfo(
+  use = JsonTypeInfo.Id.NAME,
+  property = "type",
+  include = JsonTypeInfo.As.PROPERTY
+)
+@JsonSubTypes(
+  JsonSubTypes.Type(value = ApplicationActuationPaused::class, name = "ApplicationActuationPaused"),
+  JsonSubTypes.Type(value = ApplicationActuationResumed::class, name = "ApplicationActuationResumed")
+)
+sealed class ApplicationEvent : PersistentEvent() {
+  override val scope = Scope.APPLICATION
+
+  @JsonIgnore
+  open val ignoreRepeatedInHistory: Boolean = false
+}
+
+/**
+ * Actuation at the application level has been paused.
+ *
+ * @property reason The reason why actuation was paused.
+ */
+data class ApplicationActuationPaused(
+  override val application: String,
+  val reason: String?,
+  override val timestamp: Instant
+) : ApplicationEvent() {
+  @JsonIgnore
+  override val ignoreRepeatedInHistory = true
+
+  constructor(application: String, reason: String? = null) : this(
+    application,
+    reason,
+    clock.instant()
+  )
+}
+
+/**
+ * Actuation at the application level has resumed.
+ */
+data class ApplicationActuationResumed(
+  override val application: String,
+  val reason: String?,
+  override val timestamp: Instant
+) : ApplicationEvent() {
+  @JsonIgnore
+  override val ignoreRepeatedInHistory = true
+
+  constructor(application: String, reason: String? = null) : this(
+    application,
+    reason,
+    clock.instant()
+  )
+}
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
@@ -3,6 +3,7 @@ package com.netflix.spinnaker.keel.events
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonSubTypes
 import com.fasterxml.jackson.annotation.JsonTypeInfo
+import java.time.Clock
 import java.time.Instant
 
 @JsonTypeInfo(
@@ -15,10 +16,11 @@ import java.time.Instant
   JsonSubTypes.Type(value = ApplicationActuationResumed::class, name = "ApplicationActuationResumed")
 )
 sealed class ApplicationEvent : PersistentEvent() {
-  @JsonIgnore override val scope = Scope.APPLICATION
-
   @JsonIgnore
-  open val ignoreRepeatedInHistory: Boolean = false
+  override val scope = Scope.APPLICATION
+
+  override val uid: String
+    get() = application
 }
 
 /**
@@ -34,7 +36,7 @@ data class ApplicationActuationPaused(
   @JsonIgnore
   override val ignoreRepeatedInHistory = true
 
-  constructor(application: String, reason: String? = null) : this(
+  constructor(application: String, reason: String? = null, clock: Clock = Companion.clock) : this(
     application,
     reason,
     clock.instant()
@@ -49,10 +51,9 @@ data class ApplicationActuationResumed(
   val reason: String?,
   override val timestamp: Instant
 ) : ApplicationEvent() {
-  @JsonIgnore
-  override val ignoreRepeatedInHistory = true
+  @JsonIgnore override val ignoreRepeatedInHistory = true
 
-  constructor(application: String, reason: String? = null) : this(
+  constructor(application: String, reason: String? = null, clock: Clock = Companion.clock) : this(
     application,
     reason,
     clock.instant()
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
@@ -1,0 +1,3 @@
+package com.netflix.spinnaker.keel.events
+
+class ApplicationEvent
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
@@ -15,7 +15,7 @@ import java.time.Instant
   JsonSubTypes.Type(value = ApplicationActuationResumed::class, name = "ApplicationActuationResumed")
 )
 sealed class ApplicationEvent : PersistentEvent() {
-  override val scope = Scope.APPLICATION
+  @JsonIgnore override val scope = Scope.APPLICATION
 
   @JsonIgnore
   open val ignoreRepeatedInHistory: Boolean = false
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
@@ -15,7 +15,7 @@ import java.time.Instant
   JsonSubTypes.Type(value = ApplicationActuationPaused::class, name = "ApplicationActuationPaused"),
   JsonSubTypes.Type(value = ApplicationActuationResumed::class, name = "ApplicationActuationResumed")
 )
-sealed class KeelApplicationEvent : PersistentEvent() {
+sealed class ApplicationEvent : PersistentEvent() {
   @JsonIgnore
   override val scope = Scope.APPLICATION
 
@@ -32,7 +32,7 @@ data class ApplicationActuationPaused(
   override val application: String,
   val reason: String?,
   override val timestamp: Instant
-) : KeelApplicationEvent() {
+) : ApplicationEvent() {
   @JsonIgnore
   override val ignoreRepeatedInHistory = true
 
@@ -50,7 +50,7 @@ data class ApplicationActuationResumed(
   override val application: String,
   val reason: String?,
   override val timestamp: Instant
-) : KeelApplicationEvent() {
+) : ApplicationEvent() {
   @JsonIgnore override val ignoreRepeatedInHistory = true
 
   constructor(application: String, reason: String? = null, clock: Clock = Companion.clock) : this(
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/KeelApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/KeelApplicationEvent.kt
@@ -15,7 +15,7 @@ import java.time.Instant
   JsonSubTypes.Type(value = ApplicationActuationPaused::class, name = "ApplicationActuationPaused"),
   JsonSubTypes.Type(value = ApplicationActuationResumed::class, name = "ApplicationActuationResumed")
 )
-sealed class ApplicationEvent : PersistentEvent() {
+sealed class KeelApplicationEvent : PersistentEvent() {
   @JsonIgnore
   override val scope = Scope.APPLICATION
 
@@ -32,7 +32,7 @@ data class ApplicationActuationPaused(
   override val application: String,
   val reason: String?,
   override val timestamp: Instant
-) : ApplicationEvent() {
+) : KeelApplicationEvent() {
   @JsonIgnore
   override val ignoreRepeatedInHistory = true
 
@@ -50,7 +50,7 @@ data class ApplicationActuationResumed(
   override val application: String,
   val reason: String?,
   override val timestamp: Instant
-) : ApplicationEvent() {
+) : KeelApplicationEvent() {
   @JsonIgnore override val ignoreRepeatedInHistory = true
 
   constructor(application: String, reason: String? = null, clock: Clock = Companion.clock) : this(
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/KeelApplicationEventListener.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/KeelApplicationEventListener.kt
@@ -5,9 +5,9 @@ import org.springframework.context.event.EventListener
 import org.springframework.stereotype.Component
 
 @Component
-class ApplicationEventListener(private val resourceRepository: ResourceRepository) {
-    @EventListener(ApplicationEvent::class)
-    fun onResourceEvent(event: ApplicationEvent) {
+class KeelApplicationEventListener(private val resourceRepository: ResourceRepository) {
+    @EventListener(KeelApplicationEvent::class)
+    fun onResourceEvent(event: KeelApplicationEvent) {
       resourceRepository.appendHistory(event)
     }
   }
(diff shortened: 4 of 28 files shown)


luispollo:
@asher Not sure why this fails in the build:
```
com.netflix.spinnaker.keel.tx.DeliveryConfigTransactionTests > minutests()[3][2] FAILED
    org.opentest4j.AssertionFailedError: ▼ Expect that true:
      ✗ is false
        found true
```
All the tests pass on my machine... 🤔 Any ideas?

keel-sql/src/main/resources/db/changelog/20200228-multi-scope-event-table.yml
```diff
@@ -0,0 +1,32 @@
...
+            oldTableName: resource_event
+        - modifyDataType:
+            tableName: event
+            columnName: uid
+            newDataType: varchar(255)
+        - addColumn:
+            tableName: event
+            columns:
+              - column:
+                  name: scope
```
luispollo: @asher Should we add an index on `scope, uid`?

luispollo:
@lorin If you want to review... (I think someone with admin permissions needs to add you to the org -- @robzienert @ajordens?)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
```diff
@@ -0,0 +1,21 @@
+package com.netflix.spinnaker.keel.events
+
+import java.time.Clock
+import java.time.Instant
+
+abstract class PersistentEvent {
```
luispollo: This is the new base class for all persistent events. It defines the common properties I expect to see in all such events regardless of the sub-class.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -228,14 +220,23 @@ data class ResourceActuationPaused(
...
+  constructor(resource: Resource<*>, reason: String? = null) : this(
     resource.apiVersion,
     resource.kind,
     resource.id,
     resource.application,
     reason,
     clock.instant()
   )
+
+  constructor(resource: Resource<*>, reason: String? = null, timestamp: Instant) : this(
```
luispollo: Added to allow the synthetic events of this type to be added with a specific timestamp in `EventController`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -31,7 +34,7 @@ import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Component
 
 @Component
-class ResourcePauser(
+class ActuationPauser(
```
luispollo: Renamed this since it handles pausing of things other than resources.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -108,6 +110,22 @@ interface ResourceRepository : PeriodicallyCheckedRepository<Resource<out Resour
    */
   fun deleteByApplication(application: String): Int
 
+  /**
+   * Retrieves the history of persisted events for [application].
+   *
+   * @param application the name of the application.
+   * @param limit the maximum number of events to return.
+   */
+  fun applicationEventHistory(application: String, limit: Int = DEFAULT_MAX_EVENTS): List<ApplicationEvent>
```
luispollo: We don't have an `ApplicationRepository` since we don't store application metadata, so I decided to add these here for now. Let me know if there's a better place.

keel-sql/keel-sql.gradle.kts
```diff
@@ -4,6 +4,7 @@ plugins {
   `java-library`
   id("kotlin-spring")
   id("ch.ayedo.jooqmodelator") version "3.6.0"
+  id("org.liquibase.gradle") version "2.0.2"
```
luispollo: I added this so we can run liquibase migration stuff from the command line.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -199,15 +204,49 @@ open class SqlResourceRepository(
...
+  }
+
+  override fun applicationEventHistory(application: String, downTo: Instant): List<ApplicationEvent> {
+    return sqlRetry.withRetry(READ) {
+      jooq
+        .select(EVENT.JSON)
+        .from(EVENT)
+        .where(EVENT.SCOPE.eq(Scope.APPLICATION.name))
+        .and(EVENT.UID.eq(application))
+        .and(EVENT.TIMESTAMP.lessOrEqual(LocalDateTime.ofInstant(downTo, ZoneOffset.UTC)))
```
luispollo: Apparently Jackson is converting all timestamps to UTC so JSON is stored that way. Also, for reasons I've been unable to explain, MySQL is _also_ converting timestamps to UTC, even when the timezone it's configured to use is different, so I hard-coded UTC here for that reason.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
```diff
@@ -33,22 +34,32 @@ class EventController(
...
+      if (lastBeforeAppPaused == null) {
+        log.warn("Unable to find a resource event just before application paused event at ${appPaused.timestamp}")
+      } else {
+        events.add(
+          events.indexOf(lastBeforeAppPaused),
+          ResourceActuationPaused(resource, "Resource actuation paused at the application level",
+            appPaused.timestamp)
+        )
+      }
+    }
```
luispollo: This is the gist of the actual fix: we can't just add a singe `ResourceActuationPaused` to the list because they may have been many `ApplicationActuationPaused` events in history, so we add a synthetic event matching each one of those to the list in the right position based on timestamp.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -201,28 +202,27 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
           context("updating the state again") {
             before {
               tick()
-              // TODO: ensure persisting a map with actual data
-              subject.appendHistory(ResourceDeltaDetected(resource, emptyMap(), clock))
+              subject.appendHistory(ResourceValid(resource, clock))
```
luispollo: This was unrelated, but `ResourceDeltaDetected` has `ignoreRepeatedInHistory` set to `false`, so I changed this to an event that had it set to `true`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
```diff
@@ -0,0 +1,61 @@
...
+@JsonTypeInfo(
+  use = JsonTypeInfo.Id.NAME,
+  property = "type",
+  include = JsonTypeInfo.As.PROPERTY
+)
+@JsonSubTypes(
+  JsonSubTypes.Type(value = ApplicationActuationPaused::class, name = "ApplicationActuationPaused"),
+  JsonSubTypes.Type(value = ApplicationActuationResumed::class, name = "ApplicationActuationResumed")
+)
+sealed class ApplicationEvent : PersistentEvent() {
```
robfletcher: One thing that occurs to me is that this name will conflict with a pretty common Spring class. I don't think we use it so it's not a huge deal, but we might want to reconsider.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
```diff
@@ -47,7 +50,7 @@ class InMemoryResourceRepository(
       .singleOrNull()
       ?.also {
         resources.remove(it)
-        events.remove(it)
+        resourceEvents.remove(it)
```
robfletcher: Should we remove application level events too?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
robfletcher: Timestamps are type `java.time.Instant` which is always UTC

keel-sql/src/main/resources/db/changelog/20200228-multi-scope-event-table.yml
robfletcher: Yes. Probably one on timestamp too since we seem to always order by that

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
robfletcher: Would it make sense to do this in `ActuationPauser` rather than in the controller?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -108,6 +110,22 @@ interface ResourceRepository : PeriodicallyCheckedRepository<Resource<out Resour
...
+
+  /**
+   * Retrieves the history of persisted events for [application].
+   *
+   * @param application the name of the application.
+   * @param downTo the time of the oldest event to return.
+   */
+  fun applicationEventHistory(application: String, downTo: Instant): List<ApplicationEvent>
```
lorin: My instinct is to add a `limit` parameter to all API calls that return a list of objects, just like there is for the previous call.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -129,6 +147,12 @@ interface ResourceRepository : PeriodicallyCheckedRepository<Resource<out Resour
    */
   fun appendHistory(event: ResourceEvent)
 
+  /**
+   * Records an event associated with an application.
+   * TODO: adding this here as there's no ApplicationRepository or EventRepository, but might want to move it.
+   */
+  fun appendHistory(event: ApplicationEvent)
```
lorin: Is there an advantage to having separate `appendHistory` calls for the different subclasses of `PersistentEvent` versus just having one method for the base class:

```kt
fun appendHistory(event: PersistentEvent)
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
```diff
@@ -97,22 +100,46 @@ class InMemoryResourceRepository(
...
+    appendHistory(applicationEvents, event)
+  }
+
+  private fun <T : PersistentEvent> appendHistory(eventList: MutableMap<String, MutableList<T>>, event: T) {
+    eventList.computeIfAbsent(event.uid) {
       mutableListOf()
     }
       .let {
-        if (!event.ignoreRepeatedInHistory || event.javaClass != it.firstOrNull()?.javaClass) {
+        val lastEvent = it.firstOrNull()
```
lorin: I found this a little confusing because the variable is named `lastEvent`, but the method called is `firstOrNull`. Does "last" here mean "most recent?"

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
luispollo: Actually it did bit me already. `KeelApplicationEvent`?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -219,15 +258,42 @@ open class SqlResourceRepository(
...
+        .limit(1)
+        .fetchOne()
+        ?.let { (json) ->
+          objectMapper.readValue<ApplicationEvent>(json)
+        }
+
+      if (event.javaClass == previousEvent?.javaClass) return
+    }
+
+    jooq
```
lorin: General question about code with multiple database queries: do we wrap them in transactions?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
luispollo: I thought about that, but wasn't 100% sure we weren't using application references elsewhere (meaning, those events might still be relevant even if all the resources were removed). Also, the only place this method is called is from `deleteByApplication`, which in turn is only called by `CombinedRepository.deleteResourcesByApplication`, which, in turn... is not called from anywhere.

I think this may have been a side-effect of the change Emily made from delete-by-application to be delete-by-delivery-config. We might be able to remove this method altogether...

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
luispollo: Good question. Or maybe just an extension on the `List<ResourceEvent>`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
```diff
@@ -0,0 +1,21 @@
...
+  abstract val scope: Scope
+  abstract val application: String
```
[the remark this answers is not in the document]
luispollo: I couldn't think of any, since everything we do in keel is either directly or indirectly associated with a Spinnaker application.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
luispollo: I actually tried to merge them, but the issue I ran into is that we are dynamically determining the value to insert in the `uid` column for `ResourceEvents` by doing a select into the `resources` table, which prevents adding events for resources that don't exist in the database (which is a good thing since we don't have FKs anywhere). We could add the `uid` to insert as a parameter, but I think this would complicate life unnecessarily for callers.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
luispollo: Good point. It means "most recent". The reason I call `firstOrNull` is that the lists are always ordered in descending order.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: We've been doing that more and more and filling in the gaps. I can add that.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
```diff
@@ -34,25 +36,10 @@ class InMemoryResourceRepository(
 ) : ResourceRepository {
 
   private val resources = mutableMapOf<String, Resource<*>>()
-  private val events = mutableMapOf<String, MutableList<ResourceEvent>>()
+  private val resourceEvents = mutableMapOf<String, MutableList<ResourceEvent>>()
+  private val applicationEvents = mutableMapOf<String, MutableList<KeelApplicationEvent>>()
   private val lastCheckTimes = mutableMapOf<String, Instant>()
 
-  override fun deleteByApplication(application: String): Int {
```
luispollo: This was not used anywhere...

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -35,37 +40,6 @@ open class SqlResourceRepository(
   private val sqlRetry: SqlRetry
 ) : ResourceRepository {
 
-  override fun deleteByApplication(application: String): Int {
```
luispollo: This was not used anywhere...

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
luispollo: I decided not to add this now because the whole point of this variant of the method was not to be bound by an arbitrary limit and instead make sure we look into the application event history in the exact same time window as the resource events we're returning. Hopefully, since those _are_ subject to a limit, the number of application events returned will not be a problem.

luispollo:
@robfletcher Hopefully this is in decent shape now, if you want to take a final look.

gal-yardeni:
btw, here's the issue for this: https://github.com/spinnaker/keel/issues/634

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/KeelApplicationEvent.kt
```diff
@@ -0,0 +1,61 @@
...
+@JsonTypeInfo(
+  use = JsonTypeInfo.Id.NAME,
+  property = "type",
+  include = JsonTypeInfo.As.PROPERTY
+)
+@JsonSubTypes(
+  JsonSubTypes.Type(value = ApplicationActuationPaused::class, name = "ApplicationActuationPaused"),
+  JsonSubTypes.Type(value = ApplicationActuationResumed::class, name = "ApplicationActuationResumed")
+)
+sealed class KeelApplicationEvent : PersistentEvent() {
```
gal-yardeni: Can this be renamed to `ApplicationEvent`? makes it a bit more generic maybe?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/KeelApplicationEventListener.kt
```diff
@@ -0,0 +1,13 @@
+package com.netflix.spinnaker.keel.events
+
+import com.netflix.spinnaker.keel.persistence.ResourceRepository
+import org.springframework.context.event.EventListener
+import org.springframework.stereotype.Component
+
+@Component
+class KeelApplicationEventListener(private val resourceRepository: ResourceRepository) {
```
gal-yardeni: Same as the comment above

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
gal-yardeni: Again I find the name to be a bit confusing... maybe it's just me :) but can't it be just `event`? curios to hear other opinions.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -258,8 +260,11 @@ class CombinedRepository(
   override fun deleteResource(id: String) =
     resourceRepository.delete(id)
 
-  override fun deleteResourcesByApplication(application: String): Int =
-    resourceRepository.deleteByApplication(application)
+  override fun applicationEventHistory(application: String, limit: Int): List<KeelApplicationEvent> =
+    resourceRepository.applicationEventHistory(application, limit)
+
+  override fun applicationEventHistory(application: String, downTo: Instant): List<KeelApplicationEvent> =
```
gal-yardeni: what does `downTo` mean?

luispollo: "Starting at the most recent event, **down to** this earlier timestamp, give me the application event history".

gal-yardeni:
Look good to me and makes total sense. Small comments about naming :)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
gal-yardeni: where is this comment from?

luispollo: Sorry, I'm not following... Do you mean you'd like this added as a function comment somewhere? I can add that. Otherwise, do you have a suggestion for a better name for `downTo`, perhaps?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/KeelApplicationEvent.kt
luispollo: That's what I named it originally, but then... https://github.com/spinnaker/keel/pull/842#discussion_r387319892 😃

PR #845, opened 2020-03-03
refactor(specs): Merge apiVersion and kind fields

robfletcher:
This does away with the independent `apiVersion` field and instead includes the API and version as part of the `kind`, which also now has a proper type to enforce the format.

Kinds are formatted `group/kind@version`, for example:

 - `ec2/security-group@v1`
 - `titus/cluster@v1`

etc.

The real driver for this is so that I can use `kind` as a discriminator in the Open API schema. However, the fact that this enables us to not do a bunch of custom deserialization is also nice.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
```diff
@@ -23,21 +22,25 @@ data class Resource<T : ResourceSpec>(
...
     return result
   }
 }
 
+/**
+ * Creates a resource id in the correct format.
+ */
+fun assignId(kind: String, spec: ResourceSpec) =
+  "${kind.substringBefore("/")}:${kind.substringAfterLast("/")}:${spec.id}"
+
```
robfletcher: Created a function to standardize id format as ids with `/` would break URLs.

keel-sql/src/main/resources/db/changelog/20200302-drop-api-version.yml
```diff
@@ -0,0 +1,17 @@
...
+    changes:
+    - sql:
+        sql: >
+          update resource
+          set kind = concat_ws(
+            '/',
+            substring(api_version, 1, locate('.', api_version) - 1),
+            substring(api_version, locate('/', api_version) + 1),
+            kind
+          )
```
robfletcher: This updates kinds in the database to the new format

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
luispollo: Shouldn't we just do this in the constructor instead?

luispollo: Also, won't colons be confusing with resource IDs since those use colons as well? Maybe `my.api@v1:my:resource` or just `my.api.v1:my:resource`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/SubmittedResource.kt
```diff
@@ -15,34 +15,33 @@
...
   @Description("The kind of resource `spec` represents.")
   val kind: String,
   @Description("The specification of the resource")
+  @JsonTypeInfo(use = Id.NAME, include = As.EXTERNAL_PROPERTY, property = "kind")
   val spec: T
 )
 
 val <T : ResourceSpec> SubmittedResource<T>.id: String
-  get() = "${apiVersion.substringBefore(".")}:$kind:${spec.id}"
+  get() = assignId(kind, spec)
```
luispollo: Hmm, I see where you're using this now. Forget my other comment, but maybe rename the function to `normalizeId` or `generateId`? We're not really assigning anything here.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryListenerTests.kt
```diff
@@ -26,7 +25,6 @@ internal class TelemetryListenerTests : JUnit5Minutests {
   private val registry = spyk<Registry>(NoopRegistry())
   private val counter = mockk<Counter>(relaxUnitFun = true)
   private val event = ResourceValid(
-    apiVersion = "ec2.$SPINNAKER_API_V1",
     kind = "cluster",
```
luispollo:
```suggestion
    kind = "ec2.$SPINNAKER_API_VI/cluster",
```

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ec2.kt
```diff
@@ -15,7 +15,5 @@
  */
 package com.netflix.spinnaker.keel.ec2
 
-import com.netflix.spinnaker.keel.SPINNAKER_API_V1
-
-const val SPINNAKER_EC2_API_V1 = "ec2.$SPINNAKER_API_V1"
+const val SPINNAKER_EC2_API_V1 = "ec2/v1"
```
luispollo: So this previously resulted in `ec2.spinnaker.netflix.com/v1`. Do we really want to get rid of the "FQDN" there?

keel-sql/src/main/resources/db/changelog/20200302-drop-api-version.yml
```diff
@@ -0,0 +1,17 @@
+databaseChangeLog:
+- changeSet:
+    id: drop-api-version
+    author: fletch
+    changes:
+    - sql:
+        sql: >
+          update resource
+          set kind = concat_ws(
+            '/',
```
luispollo: I thought the point was to switch from slashes to colons because of URIs? 🤔

keel-sql/src/main/resources/db/changelog/20200302-drop-api-version.yml
```diff
@@ -0,0 +1,17 @@
...
+            substring(api_version, 1, locate('.', api_version) - 1),
```
luispollo: If you had `apiVersion = 'ec2.spinnaker.netflix.com/v1'` and `kind = 'cluster'`, this would translate to `ec2/v1/cluster`, right? Again, I wasn't clear on whether we want to drop the rest of the API name...

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
robfletcher: this is resource ids

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/SubmittedResource.kt
robfletcher: yeah, that is better

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ec2.kt
robfletcher: I felt like it didn't add any value

keel-sql/src/main/resources/db/changelog/20200302-drop-api-version.yml
robfletcher: colons in the id. I deliberately tried to make it so the ids don't need to change. The kinds need to contain the context that was previously in api_version

keel-sql/src/main/resources/db/changelog/20200302-drop-api-version.yml
robfletcher: Correct

keel-sql/src/main/resources/db/changelog/20200302-drop-api-version.yml
robfletcher: Another factor is that I don't want the version from the `api_version` getting into the id. The reason is that id should uniquely identify a resource so that we don't get conflicts. If you have resources with ids `ec2/v1/cluster:prod:keel` and `ec2/v2/cluster:prod:keel` that _is_ a conflict but we wouldn't detect it because `v1` != `v2`.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ec2.kt
luispollo: Fair enough. No issues.

keel-sql/src/main/resources/db/changelog/20200302-drop-api-version.yml
luispollo: Makes sense. I misunderstood what you were trying to do. Thanks for clarifying.

robfletcher:
> Looks good to me. Are you planning to update people's YAMLs manually? I have a list if that helps.

Yeah, I'll send PRs. This includes a migration script for data.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceKind.kt
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  companion object {
+    @JvmStatic
+    fun parseKind(value: String) =
+      Regex("([\\w.-]+)/([\\w.-]+)@v(.+)").matchEntire(value)?.let { matchResult ->
+        ResourceKind(matchResult.groupValues[1], matchResult.groupValues[2], matchResult.groupValues[3])
+      } ?: error("$value is not a valid resource kind")
+  }
+}
```
robfletcher: Okay, here's an actual type that represents the `kind` of the resource. The format is enforced by this, but it serializes/deserializes as a string.

This means we don't accidentally create tests with invalid kinds, among other things.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/SubmittedResource.kt
```diff
@@ -15,34 +15,34 @@
...
-@JsonDeserialize(using = SubmittedResourceDeserializer::class)
 @Description("A resource as submitted to the Managed Delivery API.")
 data class SubmittedResource<T : ResourceSpec>(
   @Description("Optional metadata about the resource.")
   val metadata: Map<String, Any?> = emptyMap(),
-  @Description("The API version of the resource kind.")
-  val apiVersion: String,
   @Description("The kind of resource `spec` represents.")
-  val kind: String,
+  val kind: ResourceKind,
```
robfletcher: The `kind` properties here and on `SubmittedResource` change to the new type

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
```diff
@@ -23,21 +21,25 @@ data class Resource<T : ResourceSpec>(
...
     result = 31 * result + spec.hashCode()
     return result
   }
 }
 
+/**
+ * Creates a resource id in the correct format.
+ */
+fun generateId(kind: ResourceKind, spec: ResourceSpec) =
+  "${kind.group}:${kind.kind}:${spec.id}"
```
robfletcher: When it comes time to assign an id to a resource I'm using the group (e.g. `ec2`) and the un-qualified kind (e.g. `cluster`). I do _not_ want the version to be part of the id as we should consider a `ec2/cluster@v1` and an `ec2/cluster@v2` with the same moniker as a conflict (because they're specifying the same actual cluster in the cloud).

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ApiVersion.kt
```diff
@@ -0,0 +1,5 @@
+package com.netflix.spinnaker.keel.api
+
+data class ApiVersion(val group: String, val version: String = "1") {
+  fun qualify(kind: String) = ResourceKind(group, kind, version)
+}
```
robfletcher: I recreated `ApiVersion` so we have a way of defining a constant that we can easily use to create `ResourceKind` instances within that API version.

keel-api-jackson/src/main/kotlin/com/netflix/spinnaker/keel/json/mixins/ResourceKindMixin.kt
```diff
@@ -0,0 +1,10 @@
+package com.netflix.spinnaker.keel.json.mixins
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
+import com.fasterxml.jackson.databind.annotation.JsonSerialize
+import com.fasterxml.jackson.databind.ser.std.ToStringSerializer
+import com.netflix.spinnaker.keel.json.ResourceKindDeserializer
+
+@JsonSerialize(using = ToStringSerializer::class)
+@JsonDeserialize(using = ResourceKindDeserializer::class)
+interface ResourceKindMixin
```
robfletcher: This ensures `ResourceKind` is just handled as a string in JSON/YAML

keel-web/src/test/resources/examples/cluster-example.yml
```diff
@@ -1,6 +1,5 @@
 ---
-apiVersion: ec2.spinnaker.netflix.com/v1
-kind: cluster
+kind: ec2/cluster@v1
```
robfletcher: An example of how the YAML will differ

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceKind.kt
```diff
@@ -0,0 +1,29 @@
+package com.netflix.spinnaker.keel.api
+
+data class ResourceKind(
+  val group: String,
```
gal-yardeni: I find the names to be a bit confusing - what are the valid values for `group` and `kind`?

robfletcher: They can be anything, but an example would be group=ec2, kind=cluster

robfletcher: I'm happy to change those names if it would be clearer

gal-yardeni: After talking with Luis, he pointed out that this is similar to how k8s is defining resources, so it's probably better to keep it as is for simplicity. Just wondering if we are planning to document it somehow but I guess it all going to be documented once we'll have the open API schemas explaining it?

robfletcher: Also, users will not be interacting with those values separately. They'll always use the string form like `ec2/cluster@v1` (see the `toString` of that class)

keel-sql/src/main/resources/db/changelog/20200302-drop-api-version.yml
```diff
@@ -0,0 +1,16 @@
+databaseChangeLog:
+- changeSet:
+    id: drop-api-version
+    author: fletch
+    changes:
```
lorin: This migration doesn't have a rollback in it. What's the convention for this project for specifying rollbacks?

robfletcher: I usually add one but I noticed the previous couple of migrations didn’t have them

Issue #676, opened 2019-12-17
Limit delivery configs to one-per-app


asher:
We currently allow many delivery configs per application. This might be a great feature but until there's tooling/UI support for managing, including the ability to prevent and identify conflict across delivery configs, it opens up potential for a class of problems that could be easy to make but difficult to identify.

Let's disable the ability to submit >1 delivery config per application in a way that is easily backed out. One method would be to add a unique index on `delivery_config.application`.

emjburns:
+ 1, I agree.

emjburns:
Thinking about this, this would make it a bit harder to test delivery configs.

emjburns:
oops.

spinnakerbot:
This issue hasn't been updated in 99 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

PR #848, opened 2020-03-05
feat(api): API  improvements for environments/artifacts UI

### LINKED ISSUE #643: APIs to support Environments UI

luispollo:
This PR addresses additions and changes to the `/application` API required by `deck` in the new Environments UI, including:
- Refactoring the `/application` endpoint to support returning multiple views into the environment, resource and artifact data
- Moving the existing "environments summary" API previously under `/delivery-configs` over to the `/application` endpoint
- Adding support for an artifact-centric view into the data

Sample response:
```yaml
---
applicationPaused: false
hasManagedResources: true
resources:
- id: titus:cluster:titustestvpc:serverlablpollo
  kind: titus/cluster@v1
  status: ACTUATING
  moniker:
    app: serverlablpollo
  locations:
    account: titustestvpc
    vpc: vpc0
    regions:
    - name: eu-west-1
    - name: us-east-1
    - name: us-west-2
environments:
- artifacts:
  - name: spkr/mddemo-titus
    type: docker
    statuses: []
    versions:
      current: latest
      pending: []
      approved: []
      previous:
      - oldest
      vetoed: []
  name: test
  resources:
  - titus:cluster:titustestvpc:serverlablpollo
artifacts:
- name: spkr/mddemo-titus
  type: docker
  versions:
  - version: latest
    environments:
    - name: test
      state: current
      deployedAt: '2020-03-04T13:20:03Z'
  - version: oldest
    environments:
    - name: test
      state: previous
      deployedAt: '2020-03-03T13:20:03Z'
      replacedAt: '2020-03-04T13:20:03Z'
      replacedBy: latest
```

More comments on the implementation inline.

Closes #643

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -1,0 +1,31 @@
+package com.netflix.spinnaker.keel.core.api
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
+import java.time.Instant
+
+/**
+ * Summarized data about a specific artifact, mostly for use by the UI.
+ */
+data class ArtifactSummary(
+  val name: String,
+  val type: ArtifactType,
+  val versions: Set<ArtifactVersionSummary>?
+)
+
+data class ArtifactVersionSummary(
+  val version: String,
+  val environments: Set<ArtifactSummaryInEnvironment>
+)
+
+data class ArtifactSummaryInEnvironment(
+  @JsonProperty("name")
+  val environment: String,
+  @JsonIgnore
+  val version: String,
+  val state: String,
+  val deployedAt: Instant? = null,
+  val replacedAt: Instant? = null,
+  val replacedBy: String? = null
+)
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -11,7 +11,7 @@ import java.time.Instant
 data class ArtifactSummary(
   val name: String,
   val type: ArtifactType,
-  val versions: Set<ArtifactVersionSummary>?
+  val versions: Set<ArtifactVersionSummary> = emptySet()
 )
 
 data class ArtifactVersionSummary(
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
@@ -775,34 +775,62 @@ class SqlArtifactRepository(
     }
   }
 
-  override fun getCurrentVersionDeployedIn(
+  override fun getArtifactSummaryInEnvironment(
     environmentName: String,
     artifactName: String,
-    artifactType: ArtifactType
+    artifactType: ArtifactType,
+    version: String
   ): ArtifactSummaryInEnvironment? {
     return sqlRetry.withRetry(READ) {
       jooq
-        .select(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION, ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT)
+        .select(
+          ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID,
+          ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID,
+          ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION,
+          ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT,
+          ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS
+        )
         .from(ENVIRONMENT_ARTIFACT_VERSIONS)
         .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(
-          select(ENVIRONMENT.UID).from(ENVIRONMENT).where(ENVIRONMENT.NAME.eq(environmentName))
+          select(ENVIRONMENT.UID).from(ENVIRONMENT).where(ENVIRONMENT.NAME.eq(environmentName)))
         )
         .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(
-          select(DELIVERY_ARTIFACT.UID).from(DELIVERY_ARTIFACT)
+          select(DELIVERY_ARTIFACT.UID)
+            .from(DELIVERY_ARTIFACT)
             .where(DELIVERY_ARTIFACT.NAME.eq(artifactName))
-            .and(DELIVERY_ARTIFACT.TYPE.eq(artifactType.name))
-          )
-        ))
-        .and(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT.isNotNull)
-        .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.eq(CURRENT.name))
+            .and(DELIVERY_ARTIFACT.TYPE.eq(artifactType.name)))
+        )
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.eq(version))
         .orderBy(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT.desc())
         .limit(1)
-        .fetchOne { (version, deployedAt) ->
+        .fetchOne { (environmentUid, artifactUid, version, deployedAt, promotionStatus) ->
+          val (replacedBy, replacedAt) = when (promotionStatus) {
+            CURRENT.name, PREVIOUS.name -> {
+              jooq
+                .select(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION, ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT)
+                .from(ENVIRONMENT_ARTIFACT_VERSIONS)
+                .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(environmentUid))
+                .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifactUid))
+                .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.ne(version))
+                .and(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT.isNotNull)
+                .and(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT.greaterThan(deployedAt))
+                .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.ne(VETOED.name))
+                .orderBy(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT.asc())
+                .limit(1)
+                .fetchOne { (replacedBy, replacedAt) ->
+                  Pair(replacedBy, replacedAt.toInstant(ZoneOffset.UTC))
+                } ?: Pair(null, null)
+            }
+            else -> Pair(null, null)
+          }
+
           ArtifactSummaryInEnvironment(
             environment = environmentName,
             version = version,
-            state = "current",
-            deployedAt = deployedAt.toInstant(ZoneOffset.UTC)
+            state = promotionStatus,
+            deployedAt = deployedAt.toInstant(ZoneOffset.UTC),
+            replacedAt = replacedAt,
+            replacedBy = replacedBy
           )
         }
     }
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
@@ -831,7 +831,7 @@ class SqlArtifactRepository(
           ArtifactSummaryInEnvironment(
             environment = environmentName,
             version = version,
-            state = promotionStatus,
+            state = promotionStatus.toLowerCase(),
             deployedAt = deployedAt.toInstant(ZoneOffset.UTC),
             replacedAt = replacedAt,
             replacedBy = replacedBy
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
@@ -8,6 +8,7 @@ import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
 import com.netflix.spinnaker.keel.api.artifacts.DebianArtifact
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.artifacts.DockerArtifact
+import com.netflix.spinnaker.keel.core.api.ArtifactSummaryInEnvironment
 import com.netflix.spinnaker.keel.core.api.ArtifactVersionStatus
 import com.netflix.spinnaker.keel.core.api.ArtifactVersions
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactPin
@@ -39,6 +40,7 @@ import com.netflix.spinnaker.keel.sql.RetryCategory.WRITE
 import java.security.MessageDigest
 import java.time.Clock
 import java.time.Instant
+import java.time.ZoneOffset
 import javax.xml.bind.DatatypeConverter
 import org.jooq.DSLContext
 import org.jooq.Record1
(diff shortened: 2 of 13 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -66,44 +69,64 @@ class ActuationPauser(
...
+        oldestAppResumedEvent != null && oldestAppResumedEvent.timestamp.isAfter(oldestResourceEvent.timestamp) &&
+        // ...and either there's no ApplicationActuationPaused event or the oldest one is later the oldest resume (i.e. a different pause)
+        (oldestAppPausedEvent == null || oldestAppResumedEvent.timestamp.isBefore(oldestAppPausedEvent.timestamp))) {
+        // ...then add a synthetic ResouceActuationPaused event with the timestamp of resource creation
+        events.add(
+          events.indexOf(oldestResourceEvent),
+          ResourceActuationPaused(resource, "Resource actuation paused at the application level",
+            oldestResourceEvent.timestamp)
+        )
+      }
```
luispollo: This now handles the case where a resource is added to the application **after** the application is paused, which I missed in #842. It's is kinda of annoying to track that based on the application events as we can't just query the history of those events without limit, so I'm basing the query off of the oldest resource event, and inferring whether or not the app was paused before resource creation as explained in the comments.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -44,32 +44,24 @@ class ApplicationController(
...
+      "hasManagedResources" to applicationService.hasManagedResources(application)
+    ).also { results ->
+      if (includeDetails) {
+        entities.forEach { entity ->
+          results[entity] = when (entity) {
+            "resources" -> applicationService.getResourceSummariesFor(application)
+            "environments" -> applicationService.getEnvironmentSummariesFor(application)
+            "artifacts" -> applicationService.getArtifactSummariesFor(application)
+            else -> throw InvalidRequestException("Unknown entity type: $entity")
+          }
```
luispollo: This allows for flexible requests from the UI -- you can specify one or more "entity types" to return in the response. Defaults to `"resource"` as that was the original behavior.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -44,32 +44,24 @@ class ApplicationController(
...
         }
       }
-      val constraintStates = repository.constraintStateFor(application)
-
-      return mapOf(
-        "applicationPaused" to actuationPauser.applicationIsPaused(application),
-        "hasManagedResources" to resources.isNotEmpty(),
-        "resources" to resources,
-        "currentEnvironmentConstraints" to constraintStates
```
luispollo: I removed this per discussions with @erikmunson as constraints will show up elsewhere in the response later.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -85,17 +84,6 @@ class DeliveryConfigController(
     //  in [ExceptionHandler])
     mapOf("status" to "valid")
 
-  @GetMapping(
```
luispollo: This is now `GET /application/{app}?includeDetails=true&entities=artifacts`.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -0,0 +1,169 @@
...
+      }
+      ?: emptyList()
+
+  /**
+   * Returns a list of [ArtifactSummary] for the specified application by traversing the list of [EnvironmentSummary]
+   * for the application and reindexing the data so that it matches the right format.
+   *
+   * This function assumes there's a single delivery config associated with the application.
+   */
+  fun getArtifactSummariesFor(application: String): List<ArtifactSummary> {
```
luispollo: This function is fairly long and involved, but I did not want (frankly, I did not feel skilled enough) to write a SQL query like @robfletcher wrote to get the environment summaries above. Also, 99% of the data was already present in the environment summaries, so I took the approach of transforming the data to fit the format here. I added comments throughout the function -- hopefully they help.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -0,0 +1,169 @@
...
+          versions = versionSummariesByArtifact[artifact.key]
+        )
+      }
+    }
+
+    return artifactSummaries
+  }
+
+  private fun getFirstDeliveryConfigFor(application: String): DeliveryConfig? =
+    repository.getDeliveryConfigsByApplication(application).also {
```
luispollo: Again, upon discussion, we agreed to assume a single delivery config per application for now.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -773,6 +775,71 @@ class SqlArtifactRepository(
...
+            else -> Pair(null, null)
+          }
+
+          ArtifactSummaryInEnvironment(
+            environment = environmentName,
+            version = version,
+            state = promotionStatus.toLowerCase(),
+            deployedAt = deployedAt.toInstant(ZoneOffset.UTC),
+            replacedAt = replacedAt,
+            replacedBy = replacedBy
```
robfletcher: Why do you only do UTC conversion on one of the timestamps?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
luispollo: Well, that's kinda the point here... Since we didn't want to store `ResourceActuationPaused` events for every resource in the application when it is paused, we added this mechanism to inject fake ones when the API is called. Having said that, whether we should just inject the original `ApplicationActuationPaused` events in the history instead of changing the type, I'm not sure... I guess we could, but it would mix events with two different scopes (app and resource) in an endpoint that was intended to return only resource events.

@erikmunson any thoughts on this from the UI side?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: Good question. I'm still learning my way around `LocalDateTime` to/from `Instant` conversions. Should this be using the component's `clock` instead?

luispollo: I think that's the right thing to do...
https://github.com/spinnaker/keel/blob/35e30ffa607c27b3998b722b3ea963cdb93bd458/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt#L518

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
luispollo: Here's a diagram to explain this algorithm if it helps...
image

lorin: This diagram is really helpful.

I'm personally wary of adding the first _fake_ paused event because I worry that an operator will draw the wrong conclusion about when the pause was initiated.

I wonder if the UI itself could provide some sort if visual feedback that the app was paused before the resource was created without needing to create a synthetic event? Since we can assume that a resume without a matching pause implies that there was a pause that happened before the create, we could indicate this somehow in the UI view without creating a pause event that didn't actually happen at that time.

luispollo: That’s an excellent observation and question. I have to agree that if we could avoid that synthetic event with a weird timestamp, it would be best. On the other hand, the other alternative we had discussed was to actually emit such an event upon resource creation of the parent application was paused, which would effectively look exactly the same as depicted here.

@erikmunson, what are your thoughts?

luispollo: I've decided to move this to a separate PR (#853) to see if we can get the API changes needed for the UI merged sooner.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
robfletcher: It depends. Do we want the timestamp stored in the database to be Pacific Time? If so we should inject the system default clock and use that. If we want them to be UTC then this would be correct.

Personally I find it quite annoying to have to do time zone math when I'm trying to debug an issue but I'm open to the argument that UTC is less ambiguous.

lorin: Can we switch from `TIMESTAMP` to `DATETIME` in the database schema? This would encode the timezone in the data in the database.

luispollo: It appears to me that we're storing everything in UTC in the database, despite the fact that we explicitly convert `Instant` to `LocalDateTime` using the current clock.

For example, this code:
https://github.com/spinnaker/keel/blob/d5646dda5821cd927555f911335c2e0ad5f27489/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt#L284

Results in this record being added to my database a few minutes ago:
```
01E2KCK7DP3HJQJAQDTQX74Y1N	2020-03-09 19:01:04	{"type":"ResourceActuationLaunched","kind":"titus/cluster@v1","id":"titus:cluster:titustestvpc:serverlablpollo","application":"serverlablpollo","plugin":"TitusClusterHandler","tasks":[{"id":"01E30AR0X70N8RZNR4RED8VH05","name":"Upsert server group serverlablpollo in titustestvpc/eu-west-1"},{"id":"01E30AR10QHV54MYTBV3ZG07Z2","name":"Upsert server group serverlablpollo in titustestvpc/us-east-1"},{"id":"01E30AR0V3446ER87CECW56801","name":"Upsert server group serverlablpollo in titustestvpc/us-west-2"}],"timestamp":"2020-03-09T19:01:03.702Z","ignoreRepeatedInHistory":false,"uid":"titus:cluster:titustestvpc:serverlablpollo"}	RESOURCE
```

What am I missing?

luispollo: @robfletcher Sorry, I totally misread your initial question. I'm actually converting both! Lines 825 and 835.

robfletcher: What bothers me is that this seems inconsistent with `DeliveryConfigRepository` and `ResourceRepository`. We should standardize how we treat timestamps across the entire codebase

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
+import java.time.Instant
+
+/**
+ * Summarized data about a specific artifact, mostly for use by the UI.
+ */
+data class ArtifactSummary(
+  val name: String,
+  val type: ArtifactType,
+  val versions: Set<ArtifactVersionSummary>?
```
emjburns: should probably not be null, and if there are no versions it should be an empty set

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -0,0 +1,169 @@
...
+import org.springframework.stereotype.Component
+
+/**
+ * Service object that offers high-level APIs for application-related operations.
+ */
+@Component
+class ApplicationService(
+  private val pauser: ActuationPauser,
+  private val repository: KeelRepository,
+  private val artifactRepository: ArtifactRepository
```
emjburns: isn't the artifact repository in the keel repository?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -0,0 +1,169 @@
...
+    resources = resources.map { summary ->
+      if (pauser.resourceIsPaused(summary.id)) {
+        // we only update the status if the individual resource is paused,
+        // because the application pause is reflected in the response as a top level key.
+        summary.copy(status = ResourceStatus.PAUSED)
+      } else {
+        summary
+      }
+    }
+    // val constraintStates = repository.constraintStateFor(application)
```
emjburns: why the commented out line here?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
robfletcher: What's the difference between `KeelRepository` and `CombinedRepository`? Is the former the new name for the latter or are they different things?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -0,0 +1,169 @@
...
+
+    return resources
+  }
+
+  fun getEnvironmentSummariesFor(application: String): List<EnvironmentSummary> =
+    getFirstDeliveryConfigFor(application)
```
emjburns: is the "first" delivery config deterministic? seems like we're making some assumptions here and i want to make sure those hold

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: In general, I prefer a longer function comment at the top explaining the steps, then well-named variables (as you've done) to make it clear to read. For example,
```kotlin
// map of version summaries by artifact
    val versionSummariesByArtifact = ...
```
is a comment that doesn't add much IMO, because the variable has a great name.

luispollo: Yeah, good point. I'll reorganize the comments.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: I agree, but can we also agree that perhaps that's worth a PR of its own? 😃

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: `KeelRepository` is the interface, `CombinedRepository` the (only) implementation.

@emjburns It is, but I added new methods which are not used anywhere else but here so it seemed unnecessary to expose them in `KeelRepository`. I can add them if you think it makes sense.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: We had a `constraintsSomething` field at the top level of the controller response earlier, which is where I think this came from IIRC, but upon discussing with @erikmunson we agreed that constraints will show up elsewhere. AFAIK the old field was always returning empty anyway.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: Not really, but what criteria do we have that I could use to pick the "right" one? Should we be failing these calls if there's more than one?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
alanmquach: Would this be sent as `entities=resources,environments,artifacts`? I believe that's how we are sending lists of things elsewhere

luispollo: No... This would be sent like `entities=resources&entities=environments&entities=artifacts` which is how multi-value HTTP parameters are mapped in Spring (or any HTTP request library, I believe). I can change that to do the parsing of the single parameter if you prefer, I just thought I'd follow the standard...

luispollo: Actually, never mind. It looks like Spring handles both formats automatically! 👍

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: P.S. One of the reasons I added this new `ApplicationService` class is that I feel this kind of coordination among multiple repositories would be best encapsulated in service objects like these. Thoughts?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
alanmquach: Great! Thank you for confirming.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: https://github.com/spinnaker/keel/issues/862

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: P.S. https://github.com/spinnaker/keel/issues/676

PR #853, opened 2020-03-09
fix(events): Fix issue with synthetic resource paused events in API

luispollo:
Fixes a left-over issue from #842 whereby a resource created after the parent application was paused would not have a corresponding `ResourceActuationPaused` event in the beginning of its event history.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
@@ -24,12 +24,14 @@ import com.netflix.spinnaker.keel.events.ApplicationActuationPaused
 import com.netflix.spinnaker.keel.events.ApplicationActuationResumed
 import com.netflix.spinnaker.keel.events.ResourceActuationPaused
 import com.netflix.spinnaker.keel.events.ResourceActuationResumed
+import com.netflix.spinnaker.keel.events.ResourceCreated
 import com.netflix.spinnaker.keel.events.ResourceEvent
 import com.netflix.spinnaker.keel.persistence.PausedRepository
 import com.netflix.spinnaker.keel.persistence.PausedRepository.Scope
 import com.netflix.spinnaker.keel.persistence.PausedRepository.Scope.APPLICATION
 import com.netflix.spinnaker.keel.persistence.PausedRepository.Scope.RESOURCE
 import com.netflix.spinnaker.keel.persistence.ResourceRepository
+import java.time.Clock
 import org.slf4j.LoggerFactory
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Component
@@ -38,7 +40,8 @@ import org.springframework.stereotype.Component
 class ActuationPauser(
   val resourceRepository: ResourceRepository,
   val pausedRepository: PausedRepository,
-  val publisher: ApplicationEventPublisher
+  val publisher: ApplicationEventPublisher,
+  val clock: Clock
 ) {
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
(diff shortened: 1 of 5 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -66,44 +69,64 @@ class ActuationPauser(
...
+        oldestAppResumedEvent != null && oldestAppResumedEvent.timestamp.isAfter(oldestResourceEvent.timestamp) &&
+        // ...and either there's no ApplicationActuationPaused event or the oldest one is later the oldest resume (i.e. a different pause)
+        (oldestAppPausedEvent == null || oldestAppResumedEvent.timestamp.isBefore(oldestAppPausedEvent.timestamp))) {
+        // ...then add a synthetic ResouceActuationPaused event with the timestamp of resource creation
+        events.add(
+          events.indexOf(oldestResourceEvent),
+          ResourceActuationPaused(resource, "Resource actuation paused at the application level",
+            oldestResourceEvent.timestamp)
+        )
+      }
```
luispollo: This now handles the case where a resource is added to the application **after** the application is paused, which I missed in #842. It's is kinda of annoying to track that based on the application events as we can't just query the history of those events without limit, so I'm basing the query off of the oldest resource event, and inferring whether or not the app was paused before resource creation as explained in the comments.

luispollo: Here's a diagram to explain this algorithm if it helps...
image

erikmunson: Continuing on [the conversation in the previous PR's thread](https://github.com/spinnaker/keel/pull/848#discussion_r389049346) I have a couple of thoughts here:

1. On the broader topic of injecting events, I'm not highly opinionated about whether it's a `ApplicationActuationPaused` or `ResourceActuationPaused` event type, though I think in the near-term future where we don't have any awareness or nice UI treatments about which things are application vs. resource scoped it's maybe slightly preferable to stick with resource events. Separate from the type of event, I do still feel we need to do this yucky injection stuff in general (i.e. what was implemented in #842) because without it we omit critical operational context from the history.

2. On the specific question of how to handle this case (dealing with newly-created resources), I agree with @lorin that ideally we should find an approach which doesn't require falsifying history (i.e. generating a totally fake/synthetic event vs. just taking a "real" one from the application and injecting it at the correct time). I already have a to-do on my list for the history modal (will likely happen sometime in the next 6 weeks but not in the immediate-term) to put up a big banner/notice at the top if you're looking at a resource whose status is `PAUSED`. So in theory, we could just leave out the logic for new resources covered in this PR and let the UI make clear what's going on (things are paused) _as long as the resource still comes back with a status of `PAUSED`_. [Currently that's not the case because we're using the presence of a `ResourceActuationPaused` event as a cue to return that status](https://github.com/spinnaker/keel/blob/d5646dda5821cd927555f911335c2e0ad5f27489/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt#L198).

So tl;dr I think we could get away with not doing this PR if we managed to make resources return the right status without requiring an event in the history directly — I just don't know how much that's technically desirable (i.e. is it too expensive to lookup paused state every time you want to calculate resource status? etc. etc.)

robfletcher: I'm inclined to agree. I very much feel that the more complex we make this behavior the harder it is to understand (especially to 6-months-from-now-me) and the more error prone it becomes.

My suggestion would be to include application level events in the resource history. That way we'd see the "resume" event rather than having to synthesize resource level events at all.

luispollo: OK, then. Let me close this and open a different PR. My plan is to:
- Fix the issue with returning paused state for a resource based solely on resource events
- Replace the use of synthetic resource paused/resumed events with just the application counterparts.

Thanks for the feedback!

PR #865, opened 2020-03-10
fix(events): Revisit reflecting application paused/resumed events on resources

luispollo:
This is an alternative take on #842 and #853. Instead of injecting synthetic resource events in the history, we inject the original application-level paused/resumed events as appropriate.

Sample API response:
```yaml
- type: "ApplicationActuationPaused"
  application: "fnord"
  reason: "Application fnord paused"
  timestamp: "2020-03-11T06:27:03.453Z"
  scope: "application"
  uid: "fnord"
- type: "ApplicationActuationResumed"
  application: "fnord"
  reason: "Application fnord resumed"
  timestamp: "2020-03-11T06:17:03.453Z"
  scope: "application"
  uid: "fnord"
- type: "ApplicationActuationPaused"
  application: "fnord"
  reason: "Application fnord paused"
  timestamp: "2020-03-11T06:07:03.453Z"
  scope: "application"
  uid: "fnord"
- type: "ResourceValid"
  kind: "test/whatever@v1"
  id: "test:whatever:33323037"
  application: "fnord"
  timestamp: "2020-03-11T05:57:03.453Z"
  scope: "resource"
  uid: "test:whatever:33323037"
- type: "ResourceCreated"
  kind: "test/whatever@v1"
  id: "test:whatever:33323037"
  application: "fnord"
  timestamp: "2020-03-11T05:47:03.453Z"
  ignoreRepeatedInHistory: false
  scope: "resource"
  uid: "test:whatever:33323037"
```

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
@@ -35,9 +35,9 @@ data class ApplicationActuationPaused(
   @JsonIgnore
   override val ignoreRepeatedInHistory = true
 
-  constructor(application: String, reason: String? = null, clock: Clock = Companion.clock) : this(
+  constructor(application: String, clock: Clock = Companion.clock) : this(
     application,
-    reason,
+    "Application $application paused",
     clock.instant()
   )
 }
@@ -52,9 +52,9 @@ data class ApplicationActuationResumed(
 ) : ApplicationEvent() {
   @JsonIgnore override val ignoreRepeatedInHistory = true
 
-  constructor(application: String, reason: String? = null, clock: Clock = Companion.clock) : this(
+  constructor(application: String, clock: Clock = Companion.clock) : this(
     application,
-    reason,
+    "Application $application resumed",
     clock.instant()
   )
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
@@ -15,8 +15,7 @@ import java.time.Instant
   JsonSubTypes.Type(value = ApplicationActuationPaused::class, name = "ApplicationActuationPaused"),
   JsonSubTypes.Type(value = ApplicationActuationResumed::class, name = "ApplicationActuationResumed")
 )
-sealed class ApplicationEvent : PersistentEvent() {
-  @JsonIgnore
+abstract class ApplicationEvent : PersistentEvent() {
   override val scope = Scope.APPLICATION
 
   override val uid: String
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
@@ -29,7 +29,6 @@ abstract class ApplicationEvent : PersistentEvent() {
  */
 data class ApplicationActuationPaused(
   override val application: String,
-  val reason: String?,
   override val timestamp: Instant
 ) : ApplicationEvent() {
   @JsonIgnore
@@ -37,7 +36,6 @@ data class ApplicationActuationPaused(
 
   constructor(application: String, clock: Clock = Companion.clock) : this(
     application,
-    "Application $application paused",
     clock.instant()
   )
 }
@@ -47,14 +45,12 @@ data class ApplicationActuationPaused(
  */
 data class ApplicationActuationResumed(
   override val application: String,
-  val reason: String?,
   override val timestamp: Instant
 ) : ApplicationEvent() {
   @JsonIgnore override val ignoreRepeatedInHistory = true
 
   constructor(application: String, clock: Clock = Companion.clock) : this(
     application,
-    "Application $application resumed",
     clock.instant()
   )
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
@@ -1,8 +1,20 @@
 package com.netflix.spinnaker.keel.events
 
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.fasterxml.jackson.annotation.JsonSubTypes
+import com.fasterxml.jackson.annotation.JsonTypeInfo
 import java.time.Clock
 import java.time.Instant
 
+@JsonTypeInfo(
+  use = JsonTypeInfo.Id.NAME,
+  property = "type",
+  include = JsonTypeInfo.As.PROPERTY
+)
+@JsonSubTypes(
+  JsonSubTypes.Type(value = ApplicationEvent::class, name = "application"),
+  JsonSubTypes.Type(value = ResourceEvent::class, name = "resource")
+)
 abstract class PersistentEvent {
   abstract val scope: Scope
   abstract val application: String
@@ -15,7 +27,7 @@ abstract class PersistentEvent {
   }
 
   enum class Scope {
-    RESOURCE,
-    APPLICATION
+    @JsonProperty("resource") RESOURCE,
+    @JsonProperty("application") APPLICATION
   }
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
@@ -1,5 +1,6 @@
 package com.netflix.spinnaker.keel.events
 
+import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.annotation.JsonSubTypes
 import com.fasterxml.jackson.annotation.JsonTypeInfo
@@ -20,6 +21,7 @@ abstract class PersistentEvent {
   abstract val application: String
   abstract val uid: String // The unique ID of the thing associated with the scope. Defined in sub-classes.
   abstract val timestamp: Instant
+  @JsonIgnore
   open val ignoreRepeatedInHistory: Boolean = false
 
   companion object {
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
@@ -69,19 +69,19 @@ class ActuationPauser(
   fun pauseApplication(application: String) {
     log.info("Pausing application $application")
     pausedRepository.pauseApplication(application)
-    publisher.publishEvent(ApplicationActuationPaused(application, "Application $application paused", clock))
+    publisher.publishEvent(ApplicationActuationPaused(application, clock))
   }
 
   fun resumeApplication(application: String) {
     log.info("Resuming application $application")
     pausedRepository.resumeApplication(application)
-    publisher.publishEvent(ApplicationActuationResumed(application, "Application $application resumed", clock))
+    publisher.publishEvent(ApplicationActuationResumed(application, clock))
   }
 
   fun pauseResource(id: String) {
     log.info("Pausing resource $id")
     pausedRepository.pauseResource(id)
-    publisher.publishEvent(ResourceActuationPaused(resourceRepository.get(id), null, clock.instant()))
+    publisher.publishEvent(ResourceActuationPaused(resourceRepository.get(id), clock))
   }
 
   fun resumeResource(id: String) {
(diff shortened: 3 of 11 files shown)


erikmunson:
nice.

tiny tactical request for my own convenience on the UI — can you also include the `scope` thing you were modeling on these events in previous PRs? e.g. the resource events get `scope: 'resource'`, app events get `scope: 'application'`?

makes it much easier for me to express the types on the client without handwaving some optional fields around.

luispollo:
> nice.
> 
> tiny tactical request for my own convenience on the UI — can you also include the `scope` thing you were modeling on these events in previous PRs? e.g. the resource events get `scope: 'resource'`, app events get `scope: 'application'`?
> 
> makes it much easier for me to express the types on the client without handwaving some optional fields around.

Ah, yes! I thought of adding it and forgot. Now it makes sense to have it in the JSON.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -66,59 +69,55 @@ class ActuationPauser(
...
-      // We do this dynamically here so that it applies to all resources in the app, even those added _after_ the
-      // application was paused.
-      val appPausedEvents = resourceRepository
-        .applicationEventHistory(resource.application, events.last().timestamp)
-        .filterIsInstance<ApplicationActuationPaused>()
+  /**
+   * Adds [ApplicationActuationPaused] and [ApplicationActuationResumed] events to the resource event history so that
+   * it reflects actuation pauses/resumes at the application level.
+   */
+  fun addApplicationActuationEvents(originalEvents: List<ResourceEvent>, resource: Resource<*>) =
```
luispollo: Little picture if it helps...
image

robfletcher:
Not started reviewing yet, but I did notice the other day that the `ignoreRepeatedInHistory` is appearing in JSON/YAML output which seems… unnecessary.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
```diff
@@ -15,7 +27,7 @@ abstract class PersistentEvent {
   }
 
   enum class Scope {
-    RESOURCE,
-    APPLICATION
+    @JsonProperty("resource") RESOURCE,
+    @JsonProperty("application") APPLICATION
```
robfletcher: What does this do? Those don't look like properties to me, how is that annotation relevant?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
```diff
@@ -1,8 +1,20 @@
...
+import com.fasterxml.jackson.annotation.JsonSubTypes
+import com.fasterxml.jackson.annotation.JsonTypeInfo
 import java.time.Clock
 import java.time.Instant
 
+@JsonTypeInfo(
+  use = JsonTypeInfo.Id.NAME,
+  property = "type",
+  include = JsonTypeInfo.As.PROPERTY
+)
```
robfletcher: isn't the discriminator actually the `scope` property? (in which case it would also be `include = As.EXISTING_PROPERTY`)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -66,59 +69,55 @@ class ActuationPauser(
...
-        publisher.publishEvent(ResourceActuationResumed(it))
-      }
+    publisher.publishEvent(ApplicationActuationResumed(application, "Application $application resumed", clock))
   }
 
   fun pauseResource(id: String) {
     log.info("Pausing resource $id")
     pausedRepository.pauseResource(id)
-    publisher.publishEvent(ResourceActuationPaused(resourceRepository.get(id)))
+    publisher.publishEvent(ResourceActuationPaused(resourceRepository.get(id), null, clock.instant()))
```
robfletcher: What's the `null` there?

It would also be nice to be consistent about how the exception constructors use `Clock` vs `Instant`. The calls just above to `ApplicationActuationPaused` are passing in the clock, and here it's passing an instant. I don't have a strong preference (lean toward the latter) but it would be good if they always did the same thing.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -66,59 +69,55 @@ class ActuationPauser(
   fun pauseApplication(application: String) {
     log.info("Pausing application $application")
     pausedRepository.pauseApplication(application)
-    publisher.publishEvent(ApplicationActuationPaused(application))
+    publisher.publishEvent(ApplicationActuationPaused(application, "Application $application paused", clock))
```
robfletcher: Seems like that message could be generated inside the constructor rather than having to be passed in since it looks like it's always going to be the same

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -66,59 +69,55 @@ class ActuationPauser(
...
+          .filter { it is ApplicationActuationPaused || it is ApplicationActuationResumed }
 
-      appPausedEvents.forEach { appPaused ->
-        val lastBeforeAppPaused = events.firstOrNull { event ->
-          event.timestamp.isBefore(appPaused.timestamp)
-        }
+        relevantAppEvents.forEach { appEvent ->
+          val lastBeforeAppEvent = events.firstOrNull { event ->
+            event.timestamp.isBefore(appEvent.timestamp)
+          }
```
robfletcher: Couldn't we just apply `sortedBy { it.timestamp }` after we have the list rather than having to work out where to insert things?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
robfletcher: For my understanding… is this happening on-the-fly when we generate the list of events for the history API response?

Would it make sense to just make this a responsibility of the repository method that gets the events? It could do this with a single join query.

luispollo: It is. I guess we could change the query (no join required, it's the same `event` table), but I have a feeling that we might still go back and change this behavior (yet again), so maybe not worth it?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
luispollo: For enum values, this just uses the value you pass in as the enum value. In other words, just a way to make them lowercase here.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
luispollo: It is, but it turns out that is not what `property` is used for by Jackson. It's just used to name the JSON property that will include the sub-type `name` as defined below. I was confused by this too until I read the docs again.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
luispollo: Good catch, will fix.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
luispollo: The `null` is for `reason`. It was already `null` previously but the new constructor required it. As for clock vs. instant, I agree. The reason this case was "special" was that I was once creating synthetic `ResourceActuationPaused` events from an `ApplicationActuationPaused` by copying the timestamp. I can undo this now that we're just using the original application events.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
luispollo: Er... That would make total sense, wouldn't it? 🤦‍♂ Will change.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
luispollo: Having said that, I did not read up on, or try, `As.EXISTING_PROPERTY`. Let me give it a try.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
luispollo: From the docs:
> Starting with Jackson 2.6 this annotation may also be used to change serialization of Enum like so:
>
> ```java
> public enum MyEnum {
>     @JsonProperty("theFirstValue") THE_FIRST_VALUE,
>     @JsonProperty("another_value") ANOTHER_VALUE;
> }
> ```
> as an alternative to using `JsonValue` annotation.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
luispollo: I tried `AS.EXISTING_PROPERTY` with `scope` but that didn't work. For some reason it only works if I use `"type"` as the name of the property, which is the same name used in the `@JsonTypeInfo` annotation in the 2nd-level base classes. 🤷‍♂  

The reason I even added this here was so that Jackson could rebuild the class hierarchy (e.g. `PersistentEvent` -> `ResourceEvent` -> `ResourceCreated`), even though those base classes (which are, in fact, abstract) are never (de-)serialized themselves.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
robfletcher: yeah. I'm happy that it's on-the-fly. Let's go with it for a bit and we can consider optimizing later.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
robfletcher: That's great! We should change all the other places we're doing lowercasing the other way. I'll submit a PR

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
robfletcher: Aren't we going to end up with both `type` and `scope` in the JSON holding the same value?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
```diff
@@ -36,9 +35,9 @@ data class ApplicationActuationPaused(
   @JsonIgnore
   override val ignoreRepeatedInHistory = true
 
-  constructor(application: String, reason: String? = null, clock: Clock = Companion.clock) : this(
+  constructor(application: String, clock: Clock = Companion.clock) : this(
     application,
-    reason,
+    "Application $application paused",
```
lorin: What's the role of the reason field?

The comment on line 28 describes the reason field as `The reason why actuation was paused`, but the string being passed `"Application $application paused"` doesn't seem to provide additional information.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
lorin: These diagrams are _really_ helpful, thanks for taking the time to generate them.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
luispollo: No, because `type` is "overwritten" by the concrete sub-classes (which are the only ones that can actually be created and hence serialized). The output looks like the sample in the PR description.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
luispollo: Yeah, fair point. I think I can just remove them.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
robfletcher: boy I was confused there. Looking at the sample it's clear they're different things

PR #871, opened 2020-03-10
fix(api): Make application API backwards-compatible again

luispollo:
What the title says...

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
@@ -44,12 +44,19 @@ class ApplicationController(
   )
   fun get(
     @PathVariable("application") application: String,
-    @RequestParam("entities", required = false, defaultValue = "") entities: List<String>
+    @RequestParam("includeDetails", required = false, defaultValue = "false") includeDetails: Boolean,
+    @RequestParam("entities", required = false, defaultValue = "") entities: MutableList<String>
   ): Map<String, Any> {
     return mutableMapOf<String, Any>(
       "applicationPaused" to actuationPauser.applicationIsPaused(application),
-      "hasManagedResources" to applicationService.hasManagedResources(application)
+      "hasManagedResources" to applicationService.hasManagedResources(application),
+      "currentEnvironmentConstraints" to applicationService.getConstraintStatesFor(application)
+
     ).also { results ->
+      // for backwards-compatibility
+      if (includeDetails && !entities.contains("resources")) {
+        entities.add("resources")
+      }
       entities.forEach { entity ->
         results[entity] = when (entity) {
           "resources" -> applicationService.getResourceSummariesFor(application)
--- a/keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
+++ b/keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
@@ -137,7 +137,8 @@ internal class ApplicationControllerTests : JUnit5Minutests {
             """
               {
                 "applicationPaused":false,
-                "hasManagedResources":true
+                "hasManagedResources":true,
+                "currentEnvironmentConstraints":[]
               }
             """.trimIndent()
           ))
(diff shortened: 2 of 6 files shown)


keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
```diff
@@ -44,6 +45,7 @@ import strikt.assertions.containsExactly
   webEnvironment = MOCK
 )
 @AutoConfigureMockMvc
+@DirtiesContext
```
robfletcher: I don't understand how this can leak context when it's dropping all the repositories, unless there's a bug in one of the `dropAll` implementations

robfletcher: `InMemoryDeliveryConfigRepository` doesn't drop content from its `applicationConstraintMapper` or `lastCheckTimes` properties for example

robfletcher: and there are a bunch of things in `InMemoryArtifactRepository` that aren't cleared properly

luispollo: Yeah, I remember fixing one of those in a previous PR. I'll take a look, but it was worth understanding that context is indeed shared by default and how to force it not to be.

robfletcher: It's worth knowing, but this annotation will have an impact on test execution time as it will require the Spring context to be re-created. I'd rather not get into the habit of using it.

luispollo: I agree. Having said that, I've done my best to clear all in-memory representations within the repositories and, upon removing the annotation, that particular test is still failing (whereas all tests pass on my machine). Any other ideas?

luispollo: BTW, just a thought: most test classes clear the in-memory repos in an `after` block that applies to all tests. Which means that, after the last test in that class runs, nothing does. What should be the pattern to make sure we clean up after finishing each set of tests?

luispollo: Actually, the failing test should be using SQL, not in-memory repos...

```
> Task :keel-sql:test

com.netflix.spinnaker.keel.tx.DeliveryConfigTransactionTests > minutests()[3][2] FAILED
    org.opentest4j.AssertionFailedError: ▼ Expect that true:
      ✗ is false
        found true
```

@asher any thoughts on why this might pass locally and fail here?

luispollo: This is nuts. It now passed without any related changes. 🤷‍♂

luispollo:
@robfletcher Anything else you'd like changed?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -44,12 +44,19 @@ class ApplicationController(
...
     @PathVariable("application") application: String,
-    @RequestParam("entities", required = false, defaultValue = "") entities: List<String>
+    @RequestParam("includeDetails", required = false, defaultValue = "false") includeDetails: Boolean,
+    @RequestParam("entities", required = false, defaultValue = "") entities: MutableList<String>
   ): Map<String, Any> {
     return mutableMapOf<String, Any>(
       "applicationPaused" to actuationPauser.applicationIsPaused(application),
-      "hasManagedResources" to applicationService.hasManagedResources(application)
+      "hasManagedResources" to applicationService.hasManagedResources(application),
+      "currentEnvironmentConstraints" to applicationService.getConstraintStatesFor(application)
```
asher: Feel free to move `currentEnvironmentConstraints` into the `also` block so that it's only on the "legacy" `includeDetails` version of this route.

luispollo: It's OK -- @erikmunson and I hadn't concluded on where the right place to include constraint information in the new format would be, so let's leave it here for now.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
robfletcher: Well, that's the nature of non-deterministic interference

PR #874, opened 2020-03-12
feat(build): Enable composite builds

luispollo:
Attempts to fix issues building keel together with the Netflix internal repo.

The issue I ran into is that one of our internal plugins (nebula) has a transitive dependency on a library that's used by one of the plugins in OSS keel, but with a newer version that breaks the plugin with this exception:
```
Caused by: java.lang.NoSuchMethodError: net.jodah.failsafe.RetryPolicy.withDelay(JLjava/util/concurrent/TimeUnit;)Lnet/jodah/failsafe/RetryPolicy;
        at ch.ayedo.jooqmodelator.core.FlywayDependentHealthChecker.<init>(HealthCheck.kt:35)
        at ch.ayedo.jooqmodelator.core.HealthChecker$Companion.getDefault(HealthCheck.kt:18)
        at ch.ayedo.jooqmodelator.core.Modelator.waitForDatabase(Modelator.kt:54)
        at ch.ayedo.jooqmodelator.core.Modelator.access$waitForDatabase(Modelator.kt:8)
        at ch.ayedo.jooqmodelator.core.Modelator$generate$$inlined$use$lambda$1.invoke(Modelator.kt:46)
        at ch.ayedo.jooqmodelator.core.Modelator$generate$$inlined$use$lambda$1.invoke(Modelator.kt:8)
        at ch.ayedo.jooqmodelator.core.DockerClientExtensionsKt.useContainer(DockerClientExtensions.kt:13)
        at ch.ayedo.jooqmodelator.core.Modelator.generate(Modelator.kt:45)
        at ch.ayedo.jooqmodelator.gradle.JooqModelatorTask.generateMetamodel(JooqModelatorTask.kt:75)
        ...
```

The resolution issue can be easily seen running a gradle scan during the build (we were supposed to be using `net.jodah:failsafe` version `1.1.1` but end up using `2.0.1`):
image

I've tried a number of different things following the gradle documentation, as noted in the comments in `keel-sql.gradle.kts`, but none of them fixed the problem. I looked into the plugin's source code and the only thing that caught my attention was [this line](https://github.com/ayedo/jooq-modelator/blob/988d4d72951a701415a9a86363e34c5123fb4142/src/main/kotlin/ch/ayedo/jooqmodelator/gradle/JooqModelatorTask.kt#L71):
```kotlin
  val classLoader = URLClassLoader(jooqClasspath.toTypedArray(), this.javaClass.classLoader)
```

...which I guessed may be adding the full classpath of the internal project, which would include that newer version of the dependency, and unfortunately the behavior of `URLClassLoader` is such that it first delegates to the parent classloader, and only then tries its own classpath. That might explain why anything I tried at the gradle dependency resolution level had no effect, but unfortunately, if my guess is right, I can't think of a way to fix it without changing the plugin code. A counterpoint to that theory: if the gradle resolution rules were working, then the scan should show the version sticking to `1.1.1`, which it doesn't.


build.gradle.kts
```diff
@@ -35,7 +35,12 @@ subprojects {
     apply(plugin = "nebula.kotlin")
 
     repositories {
-      jcenter()
+      jcenter() {
+        metadataSources {
+          artifact()
```
luispollo: This is required to pull `net.jodah:failsfe` which doesn't have a POM, just the artifact in jcenter.

luispollo:
@marchello2000 I've tried my best to fix this but got stuck. 😭 
If you have any ideas, please let me know.

marchello2000:
sorry, i am way out of my knowledge on this one... only idea i have is what about a format like this:
https://github.com/spinnaker/clouddriver/blob/fe1ab9b29f1331d3a19973cd53e48d92c87ef65e/clouddriver-elasticsearch/clouddriver-elasticsearch.gradle#L11

luispollo:
The Netflix build team helped address the dependency conflict issue and I'm now able to use the composite build.

PR #875, opened 2020-03-12
fix(sql): fix issue with queued constraint approval

asher:


luispollo:
Thanks for fixing this!

PR #876, opened 2020-03-12
fix(export): Fix export API to handle new kind format

luispollo:
Fixes the export API in a backwards-compatible manner with the old resource kind format. Introduces new behavior in terms of looking up the correct handler by defaulting to the one that supports the latest version for the specified group and kind.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
@@ -150,7 +150,7 @@ data class SupportedKind<SPEC : ResourceSpec>(
 )
 
 /**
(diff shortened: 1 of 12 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
```diff
@@ -150,7 +150,7 @@ data class SupportedKind<SPEC : ResourceSpec>(
 )
 
 /**
- * Searches a list of `ResourceHandler`s and returns the first that supports [kind].
+ * Searches a list of []ResourceHandler]s and returns the first that supports [kind].
```
asher:
```suggestion
 * Searches a list of [ResourceHandler]s and returns the first that supports [kind].
```

PR #877, opened 2020-03-13
fix(db): Better DB types and conversions

luispollo:
This PR addresses the following issues:
- Modifies `TIMESTAMP` columns to `TIMESTAMP(3)` to add millisecond precision.
   _Note: we have timestamps that are stored as `BIGINT` in other tables. I'm not sure what the behavior is in those cases, but haven't touched them._
- Modifies `LONGTEXT` columns used to store JSON to the native MySQL `JSON` type for efficiency.
- ~~Adds a converter from `LocalDateTime` (the type used by JOOQ to store `Instant` objects) to `Instant`, so that conversion happens in a single place throughout the code-base and is consistent (especially with regards to the clock and timezone).~~ (no can do because of JOOQ bugs 😞)


keel-sql/src/main/resources/db/changelog/20200312-fix-timestamps.yml
```diff
@@ -0,0 +1,63 @@
...
+              alter table keel.environment_artifact_constraint
+              modify judged_at timestamp(3) NULL default NULL
+        - modifyDataType:
+            tableName: environment_artifact_versions
+            columnName: approved_at
+            newDataType: timestamp(3)
+        - sql:
+            sql: >
+              alter table keel.environment_artifact_versions
+              modify deployed_at timestamp(3) NULL default NULL
```
luispollo: Some of these had to be written in SQL due to this error:
```
> liquibase.exception.MigrationFailedException: Migration failed for change set /Users/lpollo/netflix/spinnaker/workspace/keel/keel-sql/src/main/resources/db/changelog/20200312-fix-timestamps.yml::fix-timestamps::lpollo:
       Reason: liquibase.exception.DatabaseException: Invalid default value for 'deployed_at' [Failed SQL: (1067) ALTER TABLE keel.environment_artifact_versions MODIFY deployed_at timestamp(3)]
```

luispollo:
@erikmunson, any chance the extra bits at the end of the timestamp might break parsing in the UI?
```
timestamp: "2020-03-13T03:21:25.787Z"
```

keel-sql/src/main/resources/db/changelog/20200312-fix-timestamps.yml
robfletcher: Another way to handle that would be to use Liquibase's `dropNotNullConstraint` but this is fine too.

robfletcher:
It would be great if we could stop using `BIGINT` as well

keel-sql/src/main/resources/jooqConfig.xml
```diff
@@ -14,6 +14,24 @@
...
         | DATABASECHANGELOGLOCK
       </excludes>
+      <forcedTypes>
+        <!-- Automatically convert JSON columns to String -->
+        <forcedType>
+          <name>LONGVARCHAR</name>
+          <includeTypes>JSON</includeTypes>
+        </forcedType>
+        <!-- Automatically convert TIMESTAMP to Instant -->
+<!--        <forcedType>-->
```
luispollo: It turns out JOOQ has built-in support for converting between time types. This causes the generator to declare the type of the field as `org.jooq.impl.SQLDataType.INSTANT` instead of org.jooq.impl.SQLDataType.LOCALDATETIME`, and sounded like it would be exactly what we needed, except inserts fail with errors like these:
```
org.jooq.exception.DataAccessException: SQL [insert into `keel`.`delivery_config_last_checked` (`delivery_config_uid`, `at`) values (?, ?) on duplicate key update `keel`.`delivery_config_last_checked`.`at` = ?]; Data truncation: Incorrect datetime value: '2020-03-13 21:24:06.244+00:00' for column 'at' at row 1

	at org.jooq_3.12.3.MYSQL.debug(Unknown Source)
	at org.jooq.impl.Tools.translate(Tools.java:2717)
	at org.jooq.impl.DefaultExecuteContext.sqlException(DefaultExecuteContext.java:755)
	at org.jooq.impl.AbstractQuery.execute(AbstractQuery.java:382)
        ...
Caused by: com.mysql.cj.jdbc.exceptions.MysqlDataTruncation: Data truncation: Incorrect datetime value: '2020-03-13 21:24:06.244+00:00' for column 'at' at row 1
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:104)
        ...
```
😢

keel-sql/src/main/resources/jooqConfig.xml
```diff
@@ -14,6 +14,24 @@
...
+<!--          <name>INSTANT</name>-->
+<!--          <includeTypes>TIMESTAMP.*</includeTypes>-->
+<!--        </forcedType>-->
+        <!-- Convert LocalDateTime to Instant -->
+        <forcedType>
+          <includeTypes>TIMESTAMP.*</includeTypes>
+          <userType>java.time.Instant</userType>
+          <converter>com.netflix.spinnaker.keel.sql.JooqTimestampsToInstantConverter</converter>
```
luispollo: This works a bit better, except that JOOQ gets lost with the types in certain cases and tries to pass a `java.sql.Timestamp` to the converter instead of a `java.time.LocalDateTime` as declared, which causes errors like these:
```
org.jooq.exception.DataAccessException: SQL [select 'keel' as `deliveryConfigName`, 'staging' as `environmentName`, `keel`.`environment_artifact_constraint`.`artifact_version`, `keel`.`environment_artifact_constraint`.`type`, `keel`.`environment_artifact_constraint`.`status`, `keel`.`environment_artifact_constraint`.`created_at`, `keel`.`environment_artifact_constraint`.`judged_by`, `keel`.`environment_artifact_constraint`.`judged_at`, `keel`.`environment_artifact_constraint`.`comment`, `keel`.`environment_artifact_constraint`.`attributes` from `keel`.`environment_artifact_constraint` where `keel`.`environment_artifact_constraint`.`environment_uid` = ? order by `keel`.`environment_artifact_constraint`.`created_at` desc limit ?]; Error while reading field: "keel"."environment_artifact_constraint"."created_at", at JDBC index: 6
	at org.jooq_3.12.3.MYSQL.debug(Unknown Source)
	... 
Caused by: java.lang.ClassCastException: java.sql.Timestamp cannot be cast to java.time.LocalDateTime
	at com.netflix.spinnaker.keel.sql.JooqTimestampsToInstantConverter.from(JooqConverters.kt:11)
	... 
```
😭

PR #880, opened 2020-03-14
fix(api): Fix/simplify retrieval of artifact summaries

luispollo:
Fixes issues reported by @erikmunson with the `/application?entities=artifacts` API:
- I see two (seemingly identical?) entries for the keeldemo debian — is there something going on there?
Inside the versions list, I think what I’m seeing is each version getting an item in the list for each environment it’s referenced by — so if version h92 applies to test and prestaging, there’s a versions list entry for h2 that has all the test information, and another separate entry for h2 that has all the prestaging information. If I’m interpreting that correctly, can we talk about how to get all the metadata across all environments rolled into a single list item for each distinct version?
- Within the environments list for a version, I’m seeing multiple entries for the same environment, each with a different deployedAt and replacedAt. Any idea what’s being shown there? I think for the UI at least all we’d want is the ‘latest’ state of that version + environment — so a single entry per environment

Turns out I had way overcomplicated the logic to transform the data from environment summaries to artifact summaries, so I pretty much rewrote that from scratch. The new implementation is much simpler and hopefully easier to follow as it follows the path that things show up in the results (artifacts -> versions -> environments).

Tests have been updated to cover multiple environments and different artifact versions with different states across then.


luispollo:
Sample JSON output is in the tests:
https://github.com/spinnaker/keel/blob/779197343895ead00419dc2634b6c278c9732301/keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt#L317-L377

luispollo:
@erikmunson Sorry this took so long.

luispollo:
Here's another sample. It turns out there's still a little 🐞 in there because the `environments` for `pending` versions are coming up empty:
```yaml
applicationPaused: false
hasManagedResources: true
currentEnvironmentConstraints:
- deliveryConfigName: "myspinapp"
  environmentName: "test"
  artifactVersion: "myspinapp-0.0.1~dev.10-h12.7ca62a0"
  type: "manual-judgement"
  status: "PENDING"
  createdAt: "2020-03-13T22:40:58.435Z"
environments:
- artifacts:
  - name: "myspinapp"
    type: "deb"
    statuses:
    - "SNAPSHOT"
    versions:
      pending:
      - "myspinapp-0.0.1~dev.10-h12.7ca62a0"
      - "myspinapp-0.0.1~dev.10-h10.7ca62a0"
      approved: []
      previous: []
      vetoed: []
  name: "test"
  resources:
  - "ec2:cluster:test:myspinapp"
  - "bakery:image:myspinapp"
artifacts:
- name: "myspinapp"
  type: "deb"
  versions:
  - version: "myspinapp-0.0.1~dev.10-h12.7ca62a0"
    environments: []
  - version: "myspinapp-0.0.1~dev.10-h10.7ca62a0"
    environments: []
resources:
- id: "ec2:cluster:test:myspinapp"
  kind: "ec2/cluster@v1"
  status: "DIFF"
  moniker:
    app: "myspinapp"
  locations:
    account: "test"
    vpc: "vpc0"
    regions:
    - name: "us-east-1"
- id: "bakery:image:myspinapp"
  kind: "bakery/image@v1"
  status: "UNHAPPY"
```

luispollo:
Fixed! ✅ 
```yaml
applicationPaused: false
hasManagedResources: true
currentEnvironmentConstraints:
- deliveryConfigName: "myspinapp"
  environmentName: "test"
  artifactVersion: "myspinapp-0.0.1~dev.10-h12.7ca62a0"
  type: "manual-judgement"
  status: "PENDING"
  createdAt: "2020-03-13T22:40:58.435Z"
artifacts:
- name: "myspinapp"
  type: "deb"
  versions:
  - version: "myspinapp-0.0.1~dev.10-h12.7ca62a0"
    environments:
    - name: "test"
      state: "pending"
  - version: "myspinapp-0.0.1~dev.10-h10.7ca62a0"
    environments:
    - name: "test"
      state: "pending"
resources:
- id: "ec2:cluster:test:myspinapp"
  kind: "ec2/cluster@v1"
  status: "DIFF"
  moniker:
    app: "myspinapp"
  locations:
    account: "test"
    vpc: "vpc0"
    regions:
    - name: "us-east-1"
- id: "bakery:image:myspinapp"
  kind: "bakery/image@v1"
  status: "ACTUATING"
```

erikmunson:
woot! this looks way more concise/de-duped than the current output. just to confirm (cause figuring out behavior from example data is hard) — does this mean that _every environment_ in a delivery config which includes a reference to that artifact will be reflected in each version's `environments` list, and if nothing has happened in that environment yet the version will be listed as `pending`?

luispollo:
Yeah, that’s actually a change I made so it would show up like that, based on what @robfletcher had done for the environment summaries, but if you’d prefer I could just not include those environments in a version where they haven’t even evaluated yet.

erikmunson:
I actually asked because this behavior (including in-evaluated environments) is _far_ preferable, as it means we can rely on this same dataset to populate a “predictive” part of a version’s journey without needing to join onto the environments dataset to find out where it might go - so this is perfect!

The only thing that would be undesirable (but not merge blocking) would be if environments show up in the list as pending a when they contain no resources that reference that artifact (I.e. a version could never flow into that environment)

luispollo:
> The only thing that would be undesirable (but not merge blocking) would be if environments show up in the list as pending a when they contain no resources that reference that artifact (I.e. a version could never flow into that environment)

Yeah, tying deployments with resources is next on my list. We don't have that linkage in the DB today. Then we can come back to this.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -68,93 +69,51 @@ class ApplicationService(
...
+    return deliveryConfig.artifacts.map { artifact ->
+      artifactRepository.versions(artifact).map { version ->
+        val artifactSummariesInEnvironments = mutableSetOf<ArtifactSummaryInEnvironment>()
 
-          versionSummariesByArtifact[artifact.key]!!.addAll(
-            versions.map { version ->
-              val summaryInEnvironment = artifactRepository.getArtifactSummaryInEnvironment(
+        environmentSummaries.forEach { environmentSummary ->
+          environmentSummary.getArtifactPromotionStatus(artifact, version)?.let { status ->
+            if (status == PromotionStatus.PENDING) {
```
luispollo: Avoid a query to the database for status of `PENDING` because we already have all the info we need.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -68,93 +69,51 @@ class ApplicationService(
...
-          "previous" to artifact.versions.previous,
-          "vetoed" to artifact.versions.vetoed
-        )
-
-        versionsByState.entries.map {
-          val state = it.key
-          val versions = it.value
+    return deliveryConfig.artifacts.map { artifact ->
+      artifactRepository.versions(artifact).map { version ->
+        val artifactSummariesInEnvironments = mutableSetOf<ArtifactSummaryInEnvironment>()
```
luispollo: Hopefully this is now much simpler to follow as it's just looping through the entities in the order they appear in the result.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -68,93 +69,51 @@ class ApplicationService(
...
-   *
-   * For each environment in the delivery config:
-   * |    For each artifact in the environment:
-   * |    |    Build a map of artifact versions by state (e.g. pending, current, previous, etc.).
-   * |    |    For each pair of (state, versions):
-   * |    |    |    For each version in this state:
-   * |    |    |    |    Get and cache an [ArtifactSummaryInEnvironment] for this version in this environment.
-   * |    |    |    |    Create and cache the [ArtifactVersionSummary] for this version.
-   * |    |    Finally, create the [ArtifactSummary] by looking up the "inner" summaries that were built above.
-   * Return a flat list of all the artifact summaries for all environments in the delivery config.
```
robfletcher: Is this no longer right?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
robfletcher: Certainly seems a ton easier

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: Nope, algorithm is much simpler now, so I thought it was unnecessary to describe in pseudo-code.

PR #886, opened 2020-03-17
Constraint to ensure AMI is available for a Debian artifact

robfletcher:


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -20,6 +20,7 @@ data class DebianArtifact(
   override val name: String,
   override val deliveryConfigName: String? = null,
   override val reference: String = name,
+  val vmOptions: VirtualMachineOptions,
```
robfletcher: New property on `DebianArtifact`

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/VirtualMachineOptions.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import com.netflix.spinnaker.keel.api.artifacts.BaseLabel.RELEASE
+import com.netflix.spinnaker.keel.api.artifacts.StoreType.EBS
+
+data class VirtualMachineOptions(
+  val baseLabel: BaseLabel = RELEASE,
+  val baseOs: String,
+  val regions: Set<String>,
+  val storeType: StoreType = EBS
+)
```
robfletcher: Details needed for bake config on a `DebianArtifact`. As @erikmunson pointed out ideally we want to get rid of `regions` here and just work it out based on the resources we are deploying.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
```diff
@@ -33,7 +34,7 @@ internal class ArtifactListenerTests : JUnit5Minutests {
     .provenance("https://my.jenkins.master/jobs/fnord-release/58")
     .build()
 
-  val debianArtifact = DebianArtifact(name = "fnord", deliveryConfigName = "fnord-config")
+  val debianArtifact = DebianArtifact(name = "fnord", deliveryConfigName = "fnord-config", vmOptions = VirtualMachineOptions(baseOs = "bionic", regions = setOf("us-west-2")))
```
robfletcher: There are a lot of places where I'm having to specify this. Not sure if there's a sensible way to default it and I don't feel like it should be nullable.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
```diff
@@ -134,6 +141,10 @@ internal class DeliveryConfigControllerTests : JUnit5Minutests {
         |artifacts:
         |- name: keel
         |  type: deb
+        |  vmOptions:
+        |    baseOs: bionic
+        |    regions:
+        |    - ap-south-1
         |environments:
         |- name: test
         |  resources:
```
robfletcher: Example of the new config in use. Note that `baseLabel` and `storeType` seem to me to have sensible defaults so I've omitted them.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
emjburns: I think it's sensible to default it, can we default from properties in the config file? it's going to be pretty installation specific I think.

luispollo: +1. Not unlike what we already do with base images.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
```diff
@@ -0,0 +1,62 @@
...
+      true
+    } else {
+      log.debug("Searching for baked image for %s in %s", version, regions.joinToString())
+
+      runBlocking {
+        imageService.getLatestNamedImageWithAllRegionsForAppVersion(
+          AppVersion.parseName(version),
+          defaultImageAccount,
+          regions
+        ) != null
```
luispollo: Don't we need to check that we found one that actually matches the version being evaluated? This is just searching by name, account and regions, right?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
+  }
+
+  private val defaultImageAccount: String
+    get() = dynamicConfigService.getConfig("images.default-account", "test")
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  private inline fun <reified T> DynamicConfigService.getConfig(configName: String, defaultValue: T) =
```
luispollo: This seems like it would be useful elsewhere. Move to a utility class somewhere?

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
robfletcher: Yeah, that might work. I don't think we can default the regions, but ultimately I'd like that  to go away anyway (hopefully later in this PR, maybe in a future one).

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
robfletcher: Yeah, I copied it from another place in our codebase. Would be good to remove that duplication

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
luispollo: Never mind. `parseName` is a really confusing name for that function, IMHO. Maybe `fromAmiName` instead?

robfletcher: the version is the first argument to this call so I'm not sure we're proving anything by checking the result

robfletcher: The function is in an external library so we can't change it.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/VirtualMachineOptions.kt
asher: Thoughts for a future PR - It'd be neat if keel could determine across all resources/envs every combination of (artifact, baseLabel, baseOs, storeType, regions) and enforce/trigger breaks for the superset of regions. That, or we should eliminate the requirement that image names match across all regions.

If resource in app1 uses artifact `fnord` in regions A, B and a resource in app2 uses artifact `fnord` in regions B, C, whichever gets evaluated second is going to have a hard time since the image created in region A or C will differ from B.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
```diff
@@ -0,0 +1,73 @@
...
+      it.locations.regions.map(RegionSpec::name)
+    }
+
+    return if (regions.isEmpty()) {
+      true
+    } else {
+      log.debug("Searching for baked image for %s in %s", version, regions.joinToString())
+
+      runBlocking {
+        imageService.getLatestNamedImageWithAllRegionsForAppVersion(
```
asher: I don't think this method cares which (baseOs, baseLabel) the artifact version was baked onto, which I think is required behavior.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
```diff
@@ -0,0 +1,73 @@
...
+import kotlinx.coroutines.runBlocking
+import org.slf4j.LoggerFactory
+import org.springframework.context.ApplicationEventPublisher
+
+class BakeConstraintEvaluator(
+  override val eventPublisher: ApplicationEventPublisher,
+  private val imageService: ImageService,
+  private val dynamicConfigService: DynamicConfigService,
+  repository: KeelRepository
+) : StatefulConstraintEvaluator<BakeConstraint>(repository) {
```
asher: `BakeConstraint` is a subtype of `Constraint` (but not `StatefulConstraint`) and the evaluator doesn't do anything with `ConstraintState`, so you may want to change this to subclass `ConstraintEvaluator` instead. 

If you leave this as a `StatefulConstraintEvaluator`, a `ConstraintState` db row is created in `PENDING` status, and you'll need to explicitly persist it to `PASS`. But since the constraint evaluator isn't triggering bakes and has no need to do things like track and monitor orca task id's, it's only checking if a condition is met, I think it is a better fit for the stateless interface.

asher: We should also make the combination of a stateless constraint type with the stateful evaluator type impossible.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/VirtualMachineOptions.kt
robfletcher: I was hoping we'd evaluate all regions in the entire delivery config and just bake 'em all in one go

asher: There may still be times where different delivery configs for different apps reference the same artifact with a different set of regions. Consider how many spinnaker@netflix services used to have a "spinredis" package deployed amongst them, so I don't think sharing artifacts across delivery-configs will necessarily be a rare corner case.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
robfletcher: Good point. Will fix

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
robfletcher: Do we actually need `StatefulConstraint` to exist?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
```diff
@@ -0,0 +1,178 @@
...
+      ))
+      return true
+    }
+
+    val locatableResources = targetEnvironment
+      .resources
+      .map(Resource<*>::spec)
+      .filterIsInstance<Locatable<*>>()
+
+    val regions: List<String> = locatableResources.flatMap {
```
luispollo: If we're inferring the regions from the resources here (which I think is the right approach), why do you need to specify the regions in the artifact's `vmOptions`?

robfletcher: Ultimately we want to get rid of it… baby steps

robfletcher: We're going to move in that direction eventually but there are many edge cases. For example, it's not enough just to  consider the immediate environment we're deploying to -- future environments may have different regions but would expect the AMI to already exist. It's also possible the AMI might be used outside the application or outside of the managed parts.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
```diff
@@ -0,0 +1,178 @@
...
+            )
+          )
+        ),
+        artifacts = listOf(
+          mapOf(
+            "type" to "DEB",
+            "customKind" to false,
+            "name" to artifact.name,
+            "version" to appVersion.version,
+            "location" to "rocket",
```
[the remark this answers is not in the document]
robfletcher: I think this is to do we the provenance of the artifact. Is that right @emjburns ?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
luispollo: Got it. Good point.

keel-sql/src/main/resources/db/changelog/20200325-create-artifact-last-checked-table.yml
```diff
@@ -0,0 +1,38 @@
...
+            constraints:
+              unique: true
+              nullable: false
+        - column:
+            name: at
+            type: timestamp(3)
+            # MySQL is stupid and won't let you insert a zero valued TIMESTAMP
+            defaultValueDate: "1970-01-01T00:00:01"
+            constraints:
+              nullable: false
```
robfletcher: A "last checked" table for artifacts just like we have for resources & delivery configs / environments

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ArtifactHandler.kt
```diff
@@ -0,0 +1,10 @@
+package com.netflix.spinnaker.keel.actuation
+
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+
+interface ArtifactHandler {
+  val name: String
+    get() = javaClass.simpleName
+
+  fun handle(artifact: DeliveryArtifact)
+}
```
robfletcher: New interface I'll use for checking artifacts on a schedule. `ImageHandler` will become one of these instead of a `ResourceHandler`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -119,6 +121,36 @@ class CheckScheduler(
...
+            }
+          }
+      }
+
+      runBlocking { job.join() }
+      log.debug("Scheduled artifact validation complete")
+    } else {
+      log.debug("Scheduled artifact validation disabled")
+    }
+  }
```
robfletcher: Artifact checking loop, very similar to the ones for resources and environments

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -1,53 +1,62 @@
...
+        )
+      } else {
+        val latestVersion = artifact.findLatestVersion()
+        val image = imageService.getLatestImageWithAllRegions(artifact.name, "test", artifact.vmOptions.regions.toList())
+        if (image?.appVersion != latestVersion) {
+          launchBake(artifact, latestVersion)
+        }
+      }
+    }
+  }
```
robfletcher: Core logic of `ImageHandler` flips from looking at resource to artifact.

keel-bakery-plugin/src/test/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandlerTests.kt
```diff
@@ -0,0 +1,336 @@
...
+import strikt.api.expectThat
+import strikt.assertions.failed
+import strikt.assertions.first
+import strikt.assertions.get
+import strikt.assertions.hasSize
+import strikt.assertions.isA
+import strikt.assertions.isEqualTo
+import strikt.assertions.isTrue
+
+internal class ImageHandlerTests : JUnit5Minutests {
```
robfletcher: Ended up completely rewriting the tests

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -0,0 +1,160 @@
...
+
+  private val DebianArtifact.taskAuthenticationDetails: BakeCredentials
+    get() = deliveryConfigName?.let {
+      deliveryConfigRepository.get(it).run {
+        BakeCredentials(serviceAccount, application)
+      }
+    } ?: defaultCredentials
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+}
```
robfletcher: This is the meat of it -- totally new version of `ImageHandler` that looks at the artifact rather than the `Resource<ImageSpec>`

keel-sql/src/main/resources/db/changelog/20200325-create-artifact-last-checked-table.yml
```diff
@@ -0,0 +1,70 @@
...
+            and environment.delivery_config_uid = delivery_config.uid
+            and delivery_artifact.delivery_config_name = delivery_config.name
+            and delivery_artifact.`name` = resource.spec->'$.artifactName'
+          )
+          where delivery_artifact.type = 'deb'
+    rollback:
+    - sql:
+        sql: >
+          update delivery_artifact
+          set details = json_remove(details, '$.vmOptions')
```
robfletcher: This migrates the bake settings from image resources to the artifact table

keel-web/src/test/resources/examples/cluster-example.yml
```diff
@@ -11,6 +11,11 @@ spec:
     deliveryArtifact:
       name: keeldemo
       type: deb
+      vmOptions:
+        baseOs: bionic-classic
+        regions:
+        - us-east-1
+        - us-west-2
```
robfletcher: This is what a `deb` artifact definition looks like now

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.netflix.spinnaker.keel.clouddriver.ImageService
+import com.netflix.spinnaker.keel.clouddriver.model.NamedImage
+import com.netflix.spinnaker.keel.constraints.ConstraintEvaluator
+import com.netflix.spinnaker.keel.constraints.SupportedConstraintType
+import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
+import kotlinx.coroutines.runBlocking
+import org.slf4j.LoggerFactory
+import org.springframework.context.ApplicationEventPublisher
+
+class BakeConstraintEvaluator(
```
luispollo: This should go away now, no? The cluster will remain in a `HAPPY` state until a new image is build by the `ImageHandler`, causing a diff for the cluster, right?

keel-api-jackson/src/test/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifactTests.kt
```diff
@@ -18,7 +18,15 @@ internal class DeliveryArtifactTests : JUnit5Minutests {
       {
         "name": "fnord",
         "type": "deb",
-        "deliveryConfigName": "my-delivery-config"
+        "deliveryConfigName": "my-delivery-config",
```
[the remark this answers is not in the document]
emjburns: We need delivery config name on artifacts to tie them to the delivery config. It's set internally and not exposed to users. However, if you're creating manifests for tests/not going through the normal persisting flow etc you'll need to set it yourself like done here

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/config/BakeryConfiguration.kt
```diff
@@ -25,24 +26,24 @@ class BakeryConfiguration {
   @Bean
   fun imageHandler(
     artifactRepository: ArtifactRepository,
+    deliveryConfigRepository: DeliveryConfigRepository,
```
emjburns: If you're going to use both of these repositories here I'd suggest you just wire in the `KeelRepository`

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/BakeConstraint.kt
```diff
@@ -0,0 +1,5 @@
+package com.netflix.spinnaker.keel.bakery.api
+
+import com.netflix.spinnaker.keel.api.Constraint
+
+class BakeConstraint : Constraint("bake")
```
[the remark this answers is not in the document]
robfletcher: I've renamed it to `ImageExistsConstraint` and it's now implicit. It's still required though. It's separate from the thing that launches the bake but just ensures we don't start trying to deploy an artifact before the bake has happened. We do now and the cluster resource errors because there's no AMI.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -83,6 +86,12 @@ class SqlArtifactRepository(
         .set(DELIVERY_ARTIFACT.REFERENCE, artifact.reference)
         .set(DELIVERY_ARTIFACT.DETAILS, artifact.detailsAsJson())
         .execute()
+      jooq.insertInto(ARTIFACT_LAST_CHECKED)
```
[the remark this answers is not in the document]
robfletcher: Fixing now

keel-sql/src/main/resources/db/changelog/20200325-create-artifact-last-checked-table.yml
emjburns: This is awesome!!!

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/BakeConstraintAttributes.kt
```diff
@@ -0,0 +1,7 @@
+package com.netflix.spinnaker.keel.bakery.api
+
+import com.netflix.spinnaker.keel.constraints.ConstraintStateAttributes
+
+data class BakeConstraintAttributes(
```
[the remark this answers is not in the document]
robfletcher: Yes, fixed

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/BakeConstraintEvaluator.kt
robfletcher: No, because the cluster works on the artifact version not the AMI. We still need to prevent promotion until the AMI exists.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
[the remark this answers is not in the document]
robfletcher: Actually I'm going to change things so that row does not need to exist so I don't need to do multiple inserts here.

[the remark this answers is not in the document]
robfletcher: I've been poking at this for a while now & can't get a query like that to work. This is the same as in the resource repository so I suggest we handle the change separately

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -0,0 +1,158 @@
...
+  private val repository: KeelRepository,
+  private val baseImageCache: BaseImageCache,
+  private val igorService: ArtifactService,
+  private val imageService: ImageService,
+  private val publisher: ApplicationEventPublisher,
+  private val taskLauncher: TaskLauncher,
+  private val defaultCredentials: BakeCredentials
+) : ArtifactHandler {
+
+  override suspend fun handle(artifact: DeliveryArtifact) {
```
emjburns: Do you have any ideas on how to make sure we don't keep rebaking forever? I just added this pr (https://github.com/spinnaker/keel/pull/929) to make sure we only try to rebake once. Moving baking out of the resource handler means that we can't take advantage of the `UnhappyVeto`. I'd really like that issue to not come back.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -0,0 +1,160 @@
...
+    if (artifact is DebianArtifact) {
+      if (taskLauncher.correlatedTasksRunning(artifact.correlationId)) {
+        publisher.publishEvent(
+          ArtifactCheckSkipped(artifact.type, artifact.name, "ActuationInProgress")
+        )
+      } else {
+        val latestVersion = artifact.findLatestVersion()
+        val latestBaseImageVersion = artifact.getLatestBaseImageVersion()
+        val image = imageService.getLatestImageWithAllRegions(artifact.name, "test", artifact.vmOptions.regions.toList())
```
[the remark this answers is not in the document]
robfletcher: This is just what was already in the old `ImageHandler`. We should probably do what you suggest, but maybe separate PR?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/BakeConstraint.kt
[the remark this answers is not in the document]
luispollo: Got it. I was under the impression the the cluster would only have a diff once the AMI existed, which is not the case.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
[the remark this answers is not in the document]
luispollo: So long as we follow-up on it in the not-too-distant future, sure.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -0,0 +1,167 @@
...
+        .firstOrNull()
+      if (knownVersion != null) {
+        log.debug("Latest known version of $name = $knownVersion")
+        return knownVersion
+      }
+    } catch (e: NoSuchArtifactException) {
+      log.debug("Latest known version of $name = null")
+      if (!repository.isRegistered(name, type)) {
+        // we clearly care about this artifact, let's register it.
+        repository.register(this)
```
luispollo: This could `throw` and prevent execution of the block below. Wrap it in a `try/catch`?

robfletcher: why would we want to carry on trying to do anything if this throws an exception?

PR #890, opened 2020-03-18
fix(core): scheduler thread pool and okhttp3 4.x upgrade

asher:


keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
```diff
@@ -113,4 +115,12 @@ class DefaultConfiguration {
     resolvers
       .filter { ComputeResourceSpec::class.java.isAssignableFrom(it.supportedKind.specClass) }
       .map { it as Resolver<ComputeResourceSpec> }
+
+  @Bean
+  fun taskScheduler(): TaskScheduler {
+    val scheduler = ThreadPoolTaskScheduler()
+    scheduler.threadNamePrefix = "scheduler-"
+    scheduler.poolSize = 5
```
luispollo: Should we make this configurable?

build.gradle.kts
```diff
@@ -76,6 +76,16 @@ subprojects {
...
+
+    resolutionStrategy {
+      var okHttpVersion = "4.4.1"
+      force(
+        "com.squareup.okhttp3:okhttp:$okHttpVersion",
+        "com.squareup.okhttp3:okhttp-urlconnection:$okHttpVersion",
+        "com.squareup.okhttp3:okhttp-sse:$okHttpVersion",
+        "com.squareup.okhttp3:mockwebserver:$okHttpVersion",
+        "com.squareup.okhttp3:logging-interceptor:$okHttpVersion")
+    }
```
robfletcher: Do you intend to update this in `kork/spinnaker-dependencies`?

asher: I don't think this should be propagated to other services yet since it's a potentially breaking change (despite 4.x being an okhttp3 version) for any service calling from kotlin.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
asher: Good call, will update

PR #892, opened 2020-03-18
chore(core): timeouts for scheduled check coroutines

asher:


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -54,7 +62,18 @@ class CheckScheduler(
         repository
           .resourcesDueForCheck(resourceCheckMinAgeDuration, resourceCheckBatchSize)
           .forEach {
-            launch { resourceActuator.checkResource(it) }
+            try {
+              /**
+               * Allow individual resource checks to timeout but catch the `CancellationException`
+               * to prevent the cancellation of all coroutines under [job]
+               */
```
asher: This comment may make more sense with the context that coroutine jobs launched within the `withTimeout` below are themselves children of the job created on line 61, which isn't visible in the diff by default. It's my understanding that a cancel exception thrown by a coroutine launched within the `withTimeout` would cascade up and cancel the check of all jobs that are children of parent job to all resource checks.

robfletcher: Could always quickly test that assumption by making the timeout 1ms!

emjburns:
LGTM

Issue #894, opened 2020-03-18
Unhappy veto throwing DateTimeParseException


lorin:
Seeing the following exception:

```
java.time.format.DateTimeParseException: Text cannot be parsed to a Duration
	at java.time.Duration.parse(Duration.java:412)
	at com.netflix.spinnaker.keel.veto.unhappy.UnhappyVeto.getWaitingTime(UnhappyVeto.kt:117)
	at com.netflix.spinnaker.keel.veto.unhappy.UnhappyVeto.waitingTime(UnhappyVeto.kt:112)
	at com.netflix.spinnaker.keel.veto.unhappy.UnhappyVeto.check(UnhappyVeto.kt:69)
	at com.netflix.spinnaker.keel.veto.VetoEnforcer.canCheck(VetoEnforcer.kt:37)
	at com.netflix.spinnaker.keel.actuation.ResourceActuator$checkResource$2.invokeSuspend(ResourceActuator.kt:79)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.internal.ScopeCoroutine.afterResume(Scopes.kt:32)
	at kotlinx.coroutines.AbstractCoroutine.resumeWith(AbstractCoroutine.kt:113)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:46)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:561)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:727)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:667)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:655)
```

See #866 for additional logging to help debug this issue.

luispollo:
@lorin Was there anything remaining to close #894?

lorin:
@luispollo I believe this (really #866) will just improve the logging to help us debug what's going wrong, rather than an actual fix, so this should remain open until we've figured out the nature of hte failure mode.

lorin:
[...]

PR #895, opened 2020-03-18
fix(veto): fix Value annotation syntax

lorin:
Add missing syntax for variable substitution

Fixes an issue in #866. This is work to address #894

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
@@ -45,8 +45,8 @@ class UnhappyVeto(
   private val unhappyVetoRepository: UnhappyVetoRepository,
   private val dynamicConfigService: DynamicConfigService,
 
-  @Value("\${veto.unhappy.waiting-time:PT10M}") // tests fail if default not specified here
-  private val configuredWaitingTime: String = "PT10M"
+  @Value("\${veto.unhappy.waiting-time:PT10M}")
+  private val configuredWaitingTime: String
 ) : Veto {
 
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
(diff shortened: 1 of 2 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
```diff
@@ -44,7 +44,8 @@ class UnhappyVeto(
   private val diffFingerprintRepository: DiffFingerprintRepository,
   private val unhappyVetoRepository: UnhappyVetoRepository,
   private val dynamicConfigService: DynamicConfigService,
-  @Value("veto.unhappy.waiting-time")
+
+  @Value("\${veto.unhappy.waiting-time:PT10M}") // tests fail if default not specified here
```
lorin: The duplication of `PT10M` is gross, but if it's not there, multiple tests will fail with the following exception:

```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'vetoController' defined in file [/Users/lhochstein/netflix/spinnaker/workspace/keel/keel-web/out/production/classes/com/netflix/spinnaker/keel/rest/VetoController.class]: Unsatisfied dependency expressed through constructor parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'unhappyVeto' defined in file [/Users/lhochstein/netflix/spinnaker/workspace/keel/keel-core/out/production/classes/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.class]: Unexpected exception during bean creation; nested exception is java.lang.IllegalArgumentException: Could not resolve placeholder 'veto.unhappy.waiting-time' in value "${veto.unhappy.waiting-time}"
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
```diff
@@ -44,7 +44,8 @@ class UnhappyVeto(
...
   private val configuredWaitingTime: String = "PT10M"
```
asher: You should be able to remove this PT10M now that the Value annotation has a default value.

lorin: I was going to write, "I can't, because the constructor is invoked directly in the code without  configuredWaitingTime as an arugment", but that only happens in test code, where there's no harm in modifying the test to pass the argument explicitly. I'll update the PR.

lorin: Addressed in f021ed6

robfletcher: I mean, not necessarily even as an automated test, just to confirm the behavior. If it's an automated test all the better, sure.

PR #883, opened 2020-03-17
feat(api): Add artifact info to resource summaries

luispollo:
Adds artifact details on compute resources where they were (or will be) deployed.

This PR achieves that by leveraging the existing awareness that certain resource specs have of artifacts (`VersionedArtifact` -> `VersionedArtifactContainer` interface), and resolvers that figure out what to populate in the corresponding fields of the specs.

Closes #882

PR #896, opened 2020-03-18
fix(api): More efficient linking of resources and artifacts in summaries API

luispollo:
Following a suggestion by @asher, this PR revisits the implementation in #883 by leveraging the event history to find a link between compute resources and the desired artifact versions to deploy on them.

Note that this will not have any effect until new `ResourceCreated` or `ResourceValid` events are persisted in the database, as old events will not have that information stored.


erikmunson:
so if i'm reading this correctly (entirely possible I'm not), when someone submits a brand new delivery config or resource and it enters a `CREATED` status, it won't have an artifact reference? And it will remain without a reference until a diff happens and that diff is resolved? Does that also mean if keel starts managing an existing resource that's already in its desired state it could go a very long time without getting linked up?

luispollo:
> so if i'm reading this correctly (entirely possible I'm not), when someone submits a brand new delivery config or resource and it enters a `CREATED` status, it won't have an artifact reference? And it will remain without a reference until a diff happens and that diff is resolved?

Hmm, great question. Yes. I think we could solve that by storing the artifact info on `ResourceValid` events instead. @asher? IIUC, a `ResourceValid` always follows a `ResourceDeltaResolved`, but it also follows `ResourceCreated` when the resource is already in the desired state.

> Does that also mean if keel starts managing an existing resource that's already in its desired state it could go a very long time without getting linked up?

If we do what I mentioned above, then that will solve this case as well.

erikmunson:
ok thanks for clarifying — is it possible we could also add to `ResourceCreated` to avoid ever having a period of time (even if it's just 1 minute) where this is sending back inaccurate data? Having this be wrong is very disruptive to the UI and my impression is that folks often go straight to the UI to check on their stuff right after they submit their first delivery config, increasing the likelihood they'll see things in a CREATED status.

Separately, is there another less roundabout path to accomplishing this where we just read straight out of the stored delivery config/resource definition and don't add a `desiredVersion` (so just name+type)? We don't actually need `desiredVersion` right now.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -287,6 +278,23 @@ open class SqlResourceRepository(
...
   }
 
+  override fun <T : ResourceEvent> lastEventByType(id: String, eventType: Class<T>): T? {
+    return sqlRetry.withRetry(READ) {
+      jooq
+        .select(EVENT.JSON)
+        .from(EVENT, RESOURCE)
+        .where(EVENT.SCOPE.eq(Scope.RESOURCE.name))
+        .and(RESOURCE.ID.eq(id))
+        .and(RESOURCE.UID.eq(EVENT.UID))
```
asher: This line looks like an accident to me (i.e. remove entirely instead fixing the `eq` condition)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -287,6 +278,23 @@ open class SqlResourceRepository(
...
+        .and("event.json->'\$.type' = '${ResourceDeltaResolved::class.simpleName}'")
```
asher: Nice to see the json column type put to use! Since this will be in the path of a regularly polled UI route, it may make sense to index `.type` as a virtual column. The json type is stored as binary and avoids serialization overhead but is always going to be costlier to scan than a btree lookup. An example on how to do this: https://dev.mysql.com/doc/refman/5.7/en/create-table-secondary-indexes.html#json-column-indirect-index

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: This is the join condition between the two tables?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: Awesome, I didn't know about that! Thanks for letting me know!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
asher: Sorry, I deleted that comment after realizing this is right, just not quickly enough. I like to be explicit about join type and condition. The implicit inner join style here looks cleaner, I'm just less use to thinking in that way.

luispollo:
> ok thanks for clarifying — is it possible we could also add to `ResourceCreated` to avoid ever having a period of time (even if it's just 1 minute) where this is sending back inaccurate data? Having this be wrong is very disruptive to the UI and my impression is that folks often go straight to the UI to check on their stuff right after they submit their first delivery config, increasing the likelihood they'll see things in a CREATED status.

I need to look into that. I _think_ the resource spec will have resolved already and container the artifact info when `ResourceCreated` is emitted, but need to double check. If it is, then yes.

> Separately, is there another less roundabout path to accomplishing this where we just read straight out of the stored delivery config/resource definition and don't add a `desiredVersion` (so just name+type)? We don't actually need `desiredVersion` right now.

I can't think of anything without adding additional data to the database because the only place we currently link the two things together (resource and artifact) are during resolution (first approach I took) and actuation (where we get the resolved resources and emit events -- approach I took here). It wouldn't be a big deal to add a new table with that association, I don't think, but let me see if we can fix this approach first before I move on to a 3rd one. 😄

luispollo:
Actually... I need to take a moment to rethink this over again. Resolution is ephemeral. It is always applied on-the-fly, its results never persisted. Every time we load a resource from the database, the `spec` is still the original one submitted by the user.

Let me think about this some more... @asher @robfletcher if you have ideas, please let me know.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
```diff
@@ -47,12 +47,16 @@ abstract class ResourceHandler<S : ResourceSpec, R : Any>(
...
-   * The value returned by this method is used as the basis of the diff (with the result of
-   * [current] in order to decide whether to call [create]/[update]/[upsert].
+   * Returns a pair of values, the first of which is the concrete representation of the resource spec which is used as
+   * the basis of the diff (with the result of [current]) in order to decide whether to call [create]/[update]/[upsert],
+   * the second the resolved resource type, which can be useful to understand what defaults and opinions were applied.
    *
    * @param resource the resource as persisted in the Keel database.
    */
-  suspend fun desired(resource: Resource<S>): R = toResolvedType(applyResolvers(resource))
+  suspend fun desired(resource: Resource<S>): Pair<C, Resource<S>> {
```
luispollo: This function now returns a pair of the concrete type (which is mostly backwards compatible with the previous signature) and the resolved resource, which was previously invisible to callers and can be very useful, as it was in this PR.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
```diff
@@ -64,7 +68,7 @@ abstract class ResourceHandler<S : ResourceSpec, R : Any>(
    * @param resource a fully-resolved version of the persisted resource spec. You can assume that
    * [applyResolvers] has already been called on this object.
    */
-  protected abstract suspend fun toResolvedType(resource: Resource<S>): R
+  protected abstract suspend fun toConcreteType(resource: Resource<S>): C
```
luispollo: I renamed this to remove any ambiguity with the `resolve` function implemented in handlers and the ad hoc differ. We now make a distinction between a "resolved resource" being just a resource that has passed by resolvers, and the "concrete type" that is the output of a handler translating a resource/spec into a diff-able type.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -70,35 +72,33 @@ class ResourceActuator(
       }
 
       try {
-        val (desired, current) = plugin.resolve(resource)
+        val (resolvedResource, desired, current) = plugin.resolve(resource)
```
luispollo: Here's where that resolved resource comes in handy. Now we can pass the resolved resource, which will include information about artifacts, into event constructors.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -70,35 +72,33 @@ class ResourceActuator(
...
-                    val versioned = (desired as Map<String, VersionedArtifactContainer>).values.first()
-                    Pair(versioned.artifactVersion, versioned.deliveryArtifact)
-                  } else {
-                    Pair(null, null)
-                  }
-                }
-                is VersionedArtifactContainer -> Pair(desired.artifactVersion, desired.deliveryArtifact)
-                else -> Pair(null, null)
-              }
+              val (version, artifact) = versionedArtifact.artifactVersion to versionedArtifact.deliveryArtifact
```
luispollo: With the resolved resource available above, we no longer need to look into the concrete type to find the artifact info, so I was able to simplify this block.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -111,16 +111,6 @@ open class SqlResourceRepository(
...
-  fun getUidByApplication(application: String): List<String> {
-    return sqlRetry.withRetry(READ) {
-      jooq
-        .select(RESOURCE.UID)
-        .from(RESOURCE)
-        .where(RESOURCE.APPLICATION.eq(application))
-        .fetch(RESOURCE.UID)
-    }
-  }
-
```
luispollo: Unused.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -48,21 +47,17 @@ class ApplicationService(
...
-                  type = computeResource.spec.deliveryArtifact!!.type,
-                  desiredVersion = computeResource.spec.artifactVersion!!
-                ))
-              } else {
-                summary
-              }
-            }
+          (resource.spec as ComputeResourceSpec).let {
+            val lastRelevantEvent =
+              repository.resourceLastEventByType(resource.id, ResourceValid::class.java)
```
luispollo: This gives preference to a `ResourceValid` event to get the artifact information. If the resource is not yet valid, we use the information from the `ResourceCreated` event.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
robfletcher: This has implications on our plugin contract. Not sure it's a problem but we should definitely be pretty conservative in changing this API or requiring too much of the implementor.

asher: Important points from Rob, but I do think this is a useful change.

luispollo: Yeah, I realized that, and struggled with the idea of making this change. I had initially made `applyResolvers` public and renamed it to `resolve` to allow clients access to the resolved resource, but then ran into name conflicts with the extension functions defined elsewhere. The two things were also very coupled together for good reason, so I ended up with this approach. But I understand it's a braking change. My statement above that it's "mostly backwards compatible" was bullocks. 😛

luispollo:
I've been unable to finish a new version of this PR today, so I' merging #901 to unblock new releases.

PR #914, opened 2020-03-24
chore(docs): Add release status badge to README

luispollo:


luispollo:
I think that _is_ the (release) build status. I’ll see about including the release number. Also looking into a code coverage badge.

Issue #887, opened 2020-03-18
Give each artifact version a "display name" for use in the UI


erikmunson:
Right now the canonical representation of an artifact version in keel is the fully qualified name of the artifact — e.g. `keeldemo-0.0.1~dev.95-h95.9493fe4`. This is useful as a unique key for internal operations, but in the UI it's very long, hard to parse visually, and hard to work with in layouts.

It's brittle (and currently in the case of Docker images impossible) to derive the right display name from a full version string in the UI, so let's do it in the API!

I propose we add a `displayName` field to each item in [the list of `versions` under each `ArtifactSummary`](https://github.com/spinnaker/keel/blob/master/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt#L16). That display name could be either:
- The actual version part of the string (e.g. `0.0.1~dev.95-h95.9493fe4` for debian packages
- The capture group of [the regex we use for deriving versions from tags](https://www.spinnaker.io/guides/user/managed-delivery/artifacts/#docker-images) for Docker images

erikmunson:
Note that in many places the UI will prefer using build numbers and commit SHAs which are planned to be attached as metadata — those take more time to implement and can't happen right now, so for now this unblocks us. I think there will still be useful cases for this field regardless, but we can re-evaluate later whether that's true.

PR #917, opened 2020-03-24
feat(summary): giving artifact versions a display name and metadata

emjburns:
closes https://github.com/spinnaker/keel/issues/887.

Adding some info to artifact versions so the UI has great data to display. Fallback is to not display any metadata if it doesn't exist.

Right now I'm parsing things from the version string which is brittle but okay for Netflix and this stage. In the future we will want a sustainable strategy for pulling info from stash/jenkins/git/etc and maybe storing it somewhere. We will need a refactor for that. 

Refactored the tests to have an actual app version so that we test the debian parsing with frigga.

In a future PR I'll look at also adding some docker tests.

Looks like: (leaving the list of environments empty here to make the example smaller)

```yaml
---
artifacts:
- name: emburns/spin-titus-demo
  type: docker
  versions:
  - version: master-h7.d6760b2
    displayName: master-h7.d6760b2
    environments: []
    build:
      id: 7
    git:
      commit: d6760b2
- name: deb-sample-app-server
  type: deb
  versions:
  - version: deb-sample-app-server-0.115.0-h345.2add2c9
    displayName: 0.115.0
    environments: []
    build:
      id: 345
    git:
      commit: 2add2c9
```

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -1,6 +1,8 @@
 package com.netflix.spinnaker.keel.core.api
 
 import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonInclude.Include
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
 import java.time.Instant
@@ -14,9 +16,29 @@ data class ArtifactSummary(
   val versions: Set<ArtifactVersionSummary> = emptySet()
 )
 
+@JsonInclude(Include.NON_NULL)
 data class ArtifactVersionSummary(
   val version: String,
-  val environments: Set<ArtifactSummaryInEnvironment>
+  val displayName: String,
+  val environments: Set<ArtifactSummaryInEnvironment>,
+  val build: BuildMetadata? = null,
+  val git: GitMetadata? = null
+)
+
+/**
+ * todo eb: other information should go here, like a link to the jenkins build. But that needs to be done
+ * in a scalable way. For now, this is just a minimal container for information we can parse from the version.
+ */
+data class BuildMetadata(
+  val id: Int
+)
+
+/**
+ * todo eb: other information should go here, like a link to the commit. But that needs to be done
+ * in a scalable way. For now, this is just a minimal container for information we can parse from the version.
+ */
+data class GitMetadata(
+  val commit: String
 )
 
 data class ArtifactSummaryInEnvironment(
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -1,11 +1,18 @@
 package com.netflix.spinnaker.keel.services
 
+import com.netflix.frigga.ami.AppVersion
 import com.netflix.spinnaker.keel.api.DeliveryConfig
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactType.deb
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactType.docker
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.artifacts.DockerArtifact
 import com.netflix.spinnaker.keel.core.api.ArtifactSummary
 import com.netflix.spinnaker.keel.core.api.ArtifactSummaryInEnvironment
 import com.netflix.spinnaker.keel.core.api.ArtifactVersionSummary
 import com.netflix.spinnaker.keel.core.api.ArtifactVersions
+import com.netflix.spinnaker.keel.core.api.BuildMetadata
 import com.netflix.spinnaker.keel.core.api.EnvironmentSummary
+import com.netflix.spinnaker.keel.core.api.GitMetadata
 import com.netflix.spinnaker.keel.core.api.PromotionStatus
 import com.netflix.spinnaker.keel.core.api.ResourceSummary
 import com.netflix.spinnaker.keel.persistence.ArtifactRepository
@@ -90,10 +97,7 @@ class ApplicationService(
           }
         }
 
-        ArtifactVersionSummary(
-          version = version,
-          environments = artifactSummariesInEnvironments.toSet()
-        )
+        return@map versionToSummary(artifact, version, artifactSummariesInEnvironments.toSet())
       }.let { artifactVersionSummaries ->
         ArtifactSummary(
           name = artifact.name,
@@ -104,7 +108,54 @@ class ApplicationService(
     }
   }
 
-  private fun getFirstDeliveryConfigFor(application: String): DeliveryConfig? =
+  /**
+   * Takes an artifact version, plus information about the type of artifact, and constructs a summary view.
+   * This should be supplemented/re-written to use actual data from stash/git/etc instead of parsing everything
+   * from the version string.
+   */
+  private fun versionToSummary(
+    artifact: DeliveryArtifact,
+    version: String,
+    environments: Set<ArtifactSummaryInEnvironment>
+  ): ArtifactVersionSummary =
+    when (artifact.type) {
+      deb -> {
+        val appversion = AppVersion.parseName(version)
+        ArtifactVersionSummary(
+          version = version,
+          environments = environments,
+          displayName = appversion?.version ?: version.removePrefix("${artifact.name}-"),
+          build = if (appversion != null) BuildMetadata(id = appversion.buildNumber.toInt()) else null,
+          git = if (appversion != null) GitMetadata(commit = appversion.commit) else null
+        )
+      }
+      docker -> {
+        var build: BuildMetadata? = null
+        var git: GitMetadata? = null
+        val dockerArtifact = artifact as DockerArtifact
+        if (dockerArtifact.hasBuild()) {
+          // todo eb: this could be less brittle
+          val regex = Regex("""^.*-h(\d+).*$""")
+          val result = regex.find(version)
+          if (result != null && result.groupValues.size == 2) {
+            build = BuildMetadata(id = result.groupValues[1].toInt())
+          }
+        }
+        if (dockerArtifact.hasCommit()) {
+          // todo eb: this could be less brittle
+          git = GitMetadata(commit = version.substringAfterLast("."))
+        }
+        ArtifactVersionSummary(
+          version = version,
+          environments = environments,
+          displayName = version,
+          build = build,
+          git = git
+        )
+      }
+    }
+
+  fun getFirstDeliveryConfigFor(application: String): DeliveryConfig? =
     repository.getDeliveryConfigsByApplication(application).also {
       if (it.size > 1) {
         log.warn("Application $application has ${it.size} delivery configs. " +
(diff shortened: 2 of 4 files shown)


keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
```diff
@@ -171,17 +171,17 @@ internal class ApplicationControllerTests : JUnit5Minutests {
         deliveryConfig.environments.flatMap { it.resources }.forEach { resource ->
           resourceRepository.appendHistory(ResourceValid(resource))
         }
```
emjburns: Using actual app versions here makes the tests harder to read, but ensures that we test important string parsing for debians.

erikmunson:
omgomgomg

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -90,10 +97,7 @@ class ApplicationService(
           }
         }
 
-        ArtifactVersionSummary(
-          version = version,
-          environments = artifactSummariesInEnvironments.toSet()
-        )
+        return@map versionToSummary(artifact, version, artifactSummariesInEnvironments.toSet())
```
erikmunson: > return@map

holy guacamole this is some freaky stuff

emjburns: The other options is very `groovy` and you don't have to put `return@map` there. But, I like stating it explicitly because I'm bad at reading code and this helps me.

erikmunson: had to go look it up on SO — the way all these JVM langs handle closures totally breaks my brain coming from functional languages like JS. I love reading Kotlin PRs it's always an adventure 👨‍🔬

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
```diff
@@ -14,9 +16,29 @@ data class ArtifactSummary(
...
   val version: String,
-  val environments: Set<ArtifactSummaryInEnvironment>
+  val displayName: String,
+  val environments: Set<ArtifactSummaryInEnvironment>,
+  val build: BuildMetadata? = null,
+  val git: GitMetadata? = null
+)
+
+/**
+ * todo eb: other information should go here, like a link to the jenkins build. But that needs to be done
```
luispollo: I'd envisioned this would actually be a link to the new CI view within Spinnaker. That should take care of the "scalable" part, no?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -104,7 +108,54 @@ class ApplicationService(
...
+    version: String,
+    environments: Set<ArtifactSummaryInEnvironment>
+  ): ArtifactVersionSummary =
+    when (artifact.type) {
+      deb -> {
+        val appversion = AppVersion.parseName(version)
+        ArtifactVersionSummary(
+          version = version,
+          environments = environments,
+          displayName = appversion?.version ?: version.removePrefix("${artifact.name}-"),
```
luispollo: I was hoping that we might be able to extract the "clean" version string just using the regex in the `TagVersionStrategy`. Is there a reason that wouldn't work?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -104,7 +108,54 @@ class ApplicationService(
...
+          git = if (appversion != null) GitMetadata(commit = appversion.commit) else null
+        )
+      }
+      docker -> {
+        var build: BuildMetadata? = null
+        var git: GitMetadata? = null
+        val dockerArtifact = artifact as DockerArtifact
+        if (dockerArtifact.hasBuild()) {
+          // todo eb: this could be less brittle
+          val regex = Regex("""^.*-h(\d+).*$""")
```
luispollo: Same comment as above on using the `TagVersionStrategy`.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -104,7 +108,54 @@ class ApplicationService(
...
+          val result = regex.find(version)
+          if (result != null && result.groupValues.size == 2) {
+            build = BuildMetadata(id = result.groupValues[1].toInt())
+          }
+        }
+        if (dockerArtifact.hasCommit()) {
+          // todo eb: this could be less brittle
+          git = GitMetadata(commit = version.substringAfterLast("."))
```
luispollo: Same comment as above on using the `TagVersionStrategy`.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
luispollo: Funny enough, I had come across this exact problem earlier, and added proper version strings, but forgot to port those over when I rewrote the PR. https://github.com/spinnaker/keel/pull/883/files#diff-fc6833fb02ee833fc028c7cdb6c7037a

Sorry to give you unnecessary work. 🙁

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
emjburns: I think all that is up for debate! All I mean is this will contain a link to _something_. Since the CI in spinnaker view might not be available for everyone, but jenkins will be, it seems reasonable to leave the comment text as is now.

I mean that we need to figure out how to get this info by querying some system, and I doubt that will be hammering jenkins/stash for it. Maybe we build a cache and populate it as new versions come it? That kind of thing. We will address the design of that later though.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: This here is for a debian, and using frigga to parse the app version is definitely the right call here.

Re docker images: my first pass of this means that the 'friendly name' is the full tag. I thought about using the sorting regex to pull out info, but I think that is more hairy. I expect most of this code here to be replaced with a better way of populating the build id and commit, so I want to keep it as simple as possible

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: oh, answered above :D

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
luispollo: > Since the CI in spinnaker view might not be available for everyone, but jenkins will be, it seems reasonable to leave the comment text as is now.

Topic for a separate discussion for sure, but I was surprised by this. I thought the whole point of the CI view was to abstract Jenkins (and other CI systems). Based on conversation with @erikmunson and @gal-yardeni recently, it seems everyone was in agreement that that's where those links would point, hence my comment. But let's chat offline. 😃

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: OK, my only other suggestion would be to move the regex to a constant/val in the companion object. 

Oh, also, if I recall correctly, the `-h{number}` part of the expression is a Netflix-specific convention where the number is the Jenkins build, no? Could we maybe change that slightly to just look for the first set of one or more consecutive digits anywhere?

```suggestion
          val regex = Regex("""^.*(\d+).*$""")
```

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
```diff
@@ -171,17 +171,17 @@ internal class ApplicationControllerTests : JUnit5Minutests {
...
-        artifactRepository.markAsSuccessfullyDeployedTo(deliveryConfig, artifact, "1.0.0", "staging")
-        artifactRepository.markAsSuccessfullyDeployedTo(deliveryConfig, artifact, "1.0.0", "production")
-        artifactRepository.markAsSuccessfullyDeployedTo(deliveryConfig, artifact, "1.0.1", "test")
-        artifactRepository.markAsSuccessfullyDeployedTo(deliveryConfig, artifact, "1.0.1", "staging")
-        artifactRepository.markAsSuccessfullyDeployedTo(deliveryConfig, artifact, "1.0.2", "test")
-        artifactRepository.approveVersionFor(deliveryConfig, artifact, "1.0.3", "test")
+        artifactRepository.store(artifact, "fnord-1.0.0-h0.2add2c9", ArtifactStatus.RELEASE)
+        artifactRepository.store(artifact, "fnord-1.0.1-h1.2add2c9", ArtifactStatus.RELEASE)
+        artifactRepository.store(artifact, "fnord-1.0.2-h2.2add2c9", ArtifactStatus.RELEASE)
+        artifactRepository.store(artifact, "fnord-1.0.3-h3.2add2c9", ArtifactStatus.RELEASE)
```
luispollo: Nit, but supposedly each version would have originated from a different commit?

luispollo: Also, consider adding Docker artifacts to these tests for added coverage.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: I actually _only_ want this to be used here - I don't want it widely used because it's not a great way of doing things and I don't want it to spread.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
emjburns: Yes but copy pasting things is hard lol and that didn't add much value.

Also, yes! Adding docker is on my to-do list, which I noted in my PR description :) I'm tracking down a bug that shows up when i have both a deb and a docker docker image, so it's great motivation to add a test for that. I wanted to keep this PR small and focused though

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
gal-yardeni: @luispollo CI in Spinnaker will be available for rocket users first. We do have a plan to make it generic (and to open source it), but it will take a while to do so. We could find how to link between the two, but it still very early in the process.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: Thanks for explaining!

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -104,7 +108,54 @@ class ApplicationService(
...
+        }
+        ArtifactVersionSummary(
+          version = version,
+          environments = environments,
+          displayName = version,
+          build = build,
```
gal-yardeni: If we have the build id here, we can link it to the logs within the ci tab (I guess we will need to do it in -nflx only?)

erikmunson: once we have enough details to put together a full build link (so controller + build name + number) I intend to do a -nflx only link to our CI view as you suggest. Currently all we have is the number that's already in the version string, so unfortunately I can't choose the right build name to look for yet.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: I couldn't help but read that and think of 🦠😷...
Going nuts... 😄

PR #919, opened 2020-03-25
feat(api): test endpoint that throws an exception

lorin:
Add a `/test/error`endpoint that throws an exception (and, consequently, returns a 500 error) when hit.

This is intended to assist in testing error alerts, by giving us a reliable way to induce errors.


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/TestController.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.netflix.spinnaker.keel.yaml.APPLICATION_YAML_VALUE
+import com.netflix.spinnaker.kork.exceptions.SystemException
+import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping(path = ["/test"])
+class TestController {
```
gal-yardeni: I'm not quite sure if we need it in main, but if you can add some comments here about the purpose of this controller, it will be great!

lorin: The purpose of the controller is to make it easier to test out that alerts are configured correctly. For example, if we have an alert on exceptions, or 500 errors, we can test it out by inducing the error and then checking if the alert fired.

The reason I'd like to have it in main is that I want to be able to test that the alert is actually working correctly in the main environment. I worry about making a mistake configuring an alert for the main environment which would result in it never firing.

gal-yardeni: Oh I see, make sense to me! If you can just add this to the actual code (for future reference, and for other developers to know), it will be great.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/TestController.kt
```diff
@@ -0,0 +1,21 @@
...
+
+  @GetMapping(
+    path = ["/error"],
+    produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  fun throwAnException(): String {
```
gal-yardeni: It maybe will make sense it should return `Map<String, Any>` and not `string`, like other controllers in keel that throws an exception?

emjburns: Agreed!

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/TestController.kt
luispollo: Consider renaming this to `ErrorSimulationController` for clarity.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/TestController.kt
luispollo: Would it be useful to support different types of exceptions/status codes? If so, you could add a parameter with the desired status code and throw an exception you know would cause that code to be returned.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/TestController.kt
luispollo: Consider adding a @ConditionalOnProperty("tests.error-simulation:false") that you can override in config to enable the endpoint in specific environments. I'm a little uncomfortable just leaving this always on everywhere.

lorin: @gal-yardeni @luispollo Those are good ideas. I'll add comments in the code and protect it with a config value.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/TestController.kt
lorin: Will do.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/TestController.kt
lorin: Will do.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/TestController.kt
lorin: Yeah, I think it's potentially useful to support simulations of different kinds of errors, and I'd like to hold off on adding this functionality until we have tests that are consuming them (this meets my current testing needs for the exception alerts).

lorin:
My test is failing because I have to enable the new `tests.error-simulation` property to get it to work. Working on that now.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/TestController.kt
lorin: Addressed in 595a47f.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/TestController.kt
lorin: Addressed in 2914d241

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/TestController.kt
lorin: Addressed in 595a47f

Issue #833, opened 2020-02-25
DeliveryConfig's with multiple artifacts can improperly apply constraints


asher:
Current situation:
- Artifacts are declared at the delivery-config root
- Environments subscribe to artifacts via cluster resource references
- Constraints are evaluated for all artifact/environment combinations, regardless of use within an environment.

Result:
- Canary constraints can deploy the incorrect artifact into a production environment.
- Manual judgements are requested for incorrect artifact/environment combinations.

asher:
Idea to fix: adding an `artifacts` list property to `Environment`, with validation that it contains a subset of the `DeliveryConfig` artifacts list. If null or empty on an env, the parent DeliveryConfig list is used.

PR #920, opened 2020-03-25
Artifact approval limiting

emjburns:
**The setup:**
- You have a delivery config 
- It has two environments: `ec2-test` and `titus-test`
- Each has a cluster (`ec2-test` has an ec2 cluster, `titus-test` has a titus cluster)
- Your delivery config has two artifacts: a debian artifact and a docker artifact

**The "problem":**
Docker versions will be approved for `ec2-test` environment, and debians will be approved for the `titus-test` environment. (_note_: they will not be deployed, it's just a visualization problem when you query the `/application/{applicationName}?entities=artifacts,resources,environments` endpoint)

**The solution, in two parts:**
(1) A stateless constraint that only approves an artifact version if a resource in the environment uses that artifact.
(2) Support for _implicit_ constraints. These constraints can be either stateless or stateful, and are applied to _every single environment_. They are run before user-defined constraints. They are opaque to the user (i.e. a user never knows that we don't allow debians to be approved for titus-only environments).

Update: fixes https://github.com/spinnaker/keel/issues/833.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
@@ -22,9 +22,18 @@ class EnvironmentPromotionChecker(
   private val constraints: List<ConstraintEvaluator<*>>,
   private val publisher: ApplicationEventPublisher
 ) {
-  private val statefulEvaluators: List<ConstraintEvaluator<*>> = constraints
+  private val implicitConstraints: List<ConstraintEvaluator<*>> = constraints.filter { it.isImplicit() }
+  private val explicitConstraints: List<ConstraintEvaluator<*>> = constraints - implicitConstraints
+
+  // constraints that are only run if they are defined in a delivery config
+  private val statefulEvaluators: List<ConstraintEvaluator<*>> = explicitConstraints
     .filterIsInstance<StatefulConstraintEvaluator<*>>()
-  private val statelessEvaluators = constraints - statefulEvaluators
+  private val statelessEvaluators = explicitConstraints - statefulEvaluators
+
+  // constraints that run for every environment in a delivery config but aren't shown to the user.
+  private val implicitStatefulEvaluators: List<ConstraintEvaluator<*>> = implicitConstraints
+    .filterIsInstance<StatefulConstraintEvaluator<*>>()
+  private val implicitStatelessEvaluators: List<ConstraintEvaluator<*>> = implicitConstraints - implicitStatefulEvaluators
 
   suspend fun checkEnvironments(deliveryConfig: DeliveryConfig) {
     val pinnedEnvs: Map<String, PinnedEnvironment> = repository
@@ -70,7 +79,7 @@ class EnvironmentPromotionChecker(
                 hasPin = true
                 pinnedEnvs.versionFor(environment.name, artifact)
               }
-              environment.constraints.isEmpty() -> {
+              environment.constraints.isEmpty() && implicitConstraints.isEmpty() -> {
                 versions.firstOrNull { v ->
                   !vetoedVersions.contains(v)
                 }
@@ -89,8 +98,8 @@ class EnvironmentPromotionChecker(
                        * deployed to a required environment or outside of an allowed time.
                        */
                       val passesConstraints =
-                        checkConstraints(statelessEvaluators, artifact, deliveryConfig, v, environment) &&
-                          checkConstraints(statefulEvaluators, artifact, deliveryConfig, v, environment)
+                        checkStatelessConstraints(artifact, deliveryConfig, v, environment) &&
+                          checkStatefulConstraints(artifact, deliveryConfig, v, environment)
 
                       versionIsPending = when (environment.constraints.anyStateful) {
                         true -> repository
@@ -116,8 +125,8 @@ class EnvironmentPromotionChecker(
                 .sortedWith(artifact.versioningStrategy.comparator.reversed()) // oldest first
                 .forEach {
                   val passesConstraints: Boolean =
-                    checkConstraints(statelessEvaluators, artifact, deliveryConfig, it, environment) &&
-                    checkConstraints(statefulEvaluators, artifact, deliveryConfig, it, environment)
+                    checkStatelessConstraints(artifact, deliveryConfig, it, environment) &&
+                      checkStatefulConstraints(artifact, deliveryConfig, it, environment)
 
                   if (passesConstraints) {
                     approveVersion(deliveryConfig, artifact, it, environment.name)
@@ -132,7 +141,7 @@ class EnvironmentPromotionChecker(
                    * We don't need to re-invoke stateful constraint evaluators for these, but we still
                    * check stateless constraints to avoid approval outside of allowed-times.
                    */
-                  if (checkConstraints(statelessEvaluators, artifact, deliveryConfig, v, environment)) {
+                  if (checkStatelessConstraints(artifact, deliveryConfig, v, environment)) {
                     approveVersion(deliveryConfig, artifact, v, environment.name)
                     repository.deleteQueuedConstraintApproval(deliveryConfig.name, environment.name, v)
                   }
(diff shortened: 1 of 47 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -179,6 +188,50 @@ class EnvironmentPromotionChecker(
     }
   }
```
emjburns: Utility functions to make sure it's easy to check constraints in the right order (implicit comes before user defined)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ArtifactTypeConstraintEvaluator.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.stereotype.Component
+
+@Component
+class ArtifactTypeConstraintEvaluator(
+  override val eventPublisher: ApplicationEventPublisher
+) : ConstraintEvaluator<ArtifactTypeConstraint> {
+  override val supportedType = SupportedConstraintType<ArtifactTypeConstraint>("artifact-type")
+  override fun isImplicit() = true
+
```
emjburns: The new constraint! It runs always to make sure that artifacts are only approved for the environment if they are the type that a resource uses.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -22,9 +22,18 @@ class EnvironmentPromotionChecker(
   private val constraints: List<ConstraintEvaluator<*>>,
   private val publisher: ApplicationEventPublisher
 ) {
```
emjburns: Separating out the four different type of constraint evaluators here so that they can be called in a specific order.

gal-yardeni:
Love your description and the solution, look good to me! My only question is how the delivery config files will have these `implicit` constraints? are those being injected automatically?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ArtifactTypeConstraintEvaluator.kt
```diff
@@ -0,0 +1,33 @@
...
+  override fun canPromote(artifact: DeliveryArtifact, version: String, deliveryConfig: DeliveryConfig, targetEnvironment: Environment): Boolean {
+    val allowedTypes = mutableSetOf<ArtifactType>()
+    deliveryConfig
+      .environments
+      .firstOrNull { it.name == targetEnvironment.name }
+      ?.resources
+      ?.forEach { resource ->
+        val usedArtifact = resource.findAssociatedArtifact(deliveryConfig)
+        if (usedArtifact != null) {
+          allowedTypes.add(usedArtifact.type)
```
asher: If you do the same sort of thing to build an `allowedArtifacts` list, and asserted that `artifact` is in the list, this would also solve #833.

emjburns: @asher GREAT THOUGHT!

emjburns:
@gal-yardeni they are being injected into the environment evaluation loop. Basically, a user who submits a delivery config won't realize we also evaluate this constraint for every environment. It's something that's used just by us in the background. That is done in `EnvironmentPromotionChecker.kt` ln225. The normal situation (user defines a constraint, we evaluate that constraint) is handled on `EnvironmentPromotionChecker.kt` ln246: you'll notice we only evaluate if the constraint exists on the environment.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ArtifactUsedConstraintEvaluator.kt
```diff
@@ -0,0 +1,32 @@
...
+      .firstOrNull { it.name == targetEnvironment.name }
+      ?.resources
+      ?.forEach { resource ->
+        resource.findAssociatedArtifact(deliveryConfig)
+          ?.let { usedArtifact ->
+            allowedArtifacts.add(usedArtifact)
+          }
+      }
+
+    return artifact in allowedArtifacts
```
lorin: If you wanted to increase your functional programming score, I think you can rewrite this function like this:

```kotlin
  override fun canPromote(artifact: DeliveryArtifact, version: String, deliveryConfig: DeliveryConfig, targetEnvironment: Environment): Boolean {
    return deliveryConfig
      .environments
      .firstOrNull { it.name == targetEnvironment.name }
      ?.resources
      ?.map { it.findAssociatedArtifact(deliveryConfig) }
      ?.any { it == artifact }
      ?: false
  }
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ArtifactUsedConstraintEvaluator.kt
```diff
@@ -0,0 +1,32 @@
...
+
+@Component
+class ArtifactUsedConstraintEvaluator(
+  override val eventPublisher: ApplicationEventPublisher
+) : ConstraintEvaluator<ArtifactUsedConstraint> {
+  override val supportedType = SupportedConstraintType<ArtifactUsedConstraint>("artifact-used")
+  override fun isImplicit() = true
+
+  override fun canPromote(artifact: DeliveryArtifact, version: String, deliveryConfig: DeliveryConfig, targetEnvironment: Environment): Boolean {
+    val allowedReferences = mutableSetOf<String>()
```
lorin: If you want to improve your functional programming score, I think you can simplify the body of this function to:

```kotlin
return deliveryConfig
      .environments
      .firstOrNull { it.name == targetEnvironment.name }
      ?.resources
      ?.mapNotNull { it.findAssociatedArtifact(deliveryConfig) }
      ?.any { it.reference == artifact.reference }
      ?: false
```

lorin:
The injection mechanism was something I was unsure of as well. My guess was that if you annotated it with `@Component`, then it would somehow get injected into the `constraints` field  of `EnvironmentPromotionChecker` when that object is constructed, by some sort of Spring DI magic, but I don't know how that mechanism works.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ArtifactUsedConstraintEvaluator.kt
emjburns: My personal preference is to not do anything too functional. That is maybe controversial, but I personally find functional programming hard to read and I'm bad at it. Also, the functional style is really different from the rest of our code, so I find it hard to debug when switching back and forth

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
```diff
@@ -33,6 +33,12 @@ interface ConstraintEvaluator<T : Constraint> {
 
   val eventPublisher: ApplicationEventPublisher
 
+  /**
+   * returns true if a constraint should be run for every environment in every delivery config, without being
+   * exposed to delivery config author.
+   */
+  fun isImplicit(): Boolean = false
```
robfletcher: How would this work for a constraint that was implicit but only applied to certain artifact types?

emjburns: It would run for every environment, but if that artifact wasn't there then it would just return true. If the artifact or thing you cared about was there, then you'd add logic to decide what to do in that case

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -70,7 +79,7 @@ class EnvironmentPromotionChecker(
                 hasPin = true
                 pinnedEnvs.versionFor(environment.name, artifact)
               }
-              environment.constraints.isEmpty() -> {
+              environment.constraints.isEmpty() && implicitConstraints.isEmpty() -> {
```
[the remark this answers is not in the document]
emjburns: 🤷‍♀ I sort of like that it's an acknowledgment that there are two separate things, and they must be both empty. Doesn't seem worth it to create an extension function for this one `&&` statement.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ConstraintEvaluator.kt
robfletcher: So, for example the "AMI exists" constraint should just return `true` for a Docker artifact?

emjburns: Yep! but in this case "true" is more like "ehh idk sure go ahead pal"

robfletcher: Right, that's what I figured. Makes sense.

PR #926, opened 2020-03-26
feat(api): Add stateful constraint information to artifact summaries

### LINKED ISSUE #888: Surface stateful constraints for the UI via ArtifactSummaryInEnvironment
In order to show users the current state of constraints for the artifacts they care about, we need to surface _all_ the relevant constraints for a given artifact + environment combination in the API. Mockups look roughly like this:

Screen Shot 2020-03-17 at 6 45 05 PM

[The `ArtifactSummaryInEnvironment` class](https://github.com/spinnaker/keel/blob/368312c2962076c06c44465fe09937888f879707/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt#L25-L34) already summarizes similar information — when a version was deployed to a specific environment, when it was replaced by another version, etc. It makes sense to also summarize constraints here.

Here's some example JSON that @luispollo and I sketched out a while ago, all the fields are already present in the database:
```ts
{
  name: "test",
  state: "current",
  deployedAt: "2019-12-04T18:49:58.748Z",
  replacedAt: "2019-12-04T18:49:58.748Z",
  replacedBy: "keel-0.344.0-h280.966f877",
  constraints: [
    {
      // we need some kind of 'StatefulConstraintSummary' type
      // similar to ResourceSummary.
      type: "manual-judgement",
      // can be one of: NOT_STARTED | PENDING | PASS | FAIL | OVERRIDE_PASS | OVERRIDE_FAIL
      status: "PASS",
      startedAt: "2019-12-04T18:49:58.748Z",
      judgedBy: "emunson@netflix.com",
      judgedAt: "2019-12-04T18:49:58.748Z",
      comment: "It's so cool we can declaratively deliver",
      attributes: {
        // arbitrary key/value metadata, but strongly typed by constraint type
        // so we can get type safety on the client
      }
    }
  ]
},
```

An important nuance: we don't actually write these out to the DB until they get kicked off and enter a `PENDING` status. In the API/UI, we need to surface them _before_ that so users can see a grey indicator that the constraint will eventually run. Our current proposal is to add a 'synthetic' `NOT_STARTED` status for constraints that are defined in the delivery config, but haven't run and ended up in the DB yet.

luispollo:
Adds information about stateful constraints applicable to an artifact version within an environment, in support of the UI.

Closes #888.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/time/MutableClock.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/time/MutableClock.kt
@@ -8,7 +8,8 @@ import java.time.temporal.TemporalAmount
 
 class MutableClock(
   private var instant: Instant = Instant.now(),
-  private val zone: ZoneId = ZoneId.of("UTC")
+  private val zone: ZoneId = ZoneId.of("UTC"),
+  private val start: Instant = instant
 ) : Clock() {
 
   override fun withZone(zone: ZoneId): MutableClock {
@@ -35,5 +36,9 @@ class MutableClock(
 
   fun instant(newInstant: Instant) {
     instant = newInstant
+  }
+
+  fun reset() {
+    instant = start
   }
 }
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/time/MutableClock.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/time/MutableClock.kt
@@ -9,7 +9,7 @@ import java.time.temporal.TemporalAmount
 class MutableClock(
   private var instant: Instant = Instant.now(),
   private val zone: ZoneId = ZoneId.of("UTC"),
-  private val start: Instant = instant
+  val start: Instant = instant
 ) : Clock() {
 
   override fun withZone(zone: ZoneId): MutableClock {
@@ -28,11 +28,11 @@ class MutableClock(
     instant = instant.plus(amount)
   }
 
-  fun tickSeconds(seconds: Long) = incrementBy(Duration.ofSeconds(seconds))
+  fun tickSeconds(seconds: Long) = incrementBy(Duration.ofSeconds(seconds)).let { instant }
 
-  fun tickMinutes(minutes: Long) = incrementBy(Duration.ofMinutes(minutes))
+  fun tickMinutes(minutes: Long) = incrementBy(Duration.ofMinutes(minutes)).let { instant }
 
-  fun tickHours(hours: Long) = incrementBy(Duration.ofHours(hours))
+  fun tickHours(hours: Long) = incrementBy(Duration.ofHours(hours)).let { instant }
 
   fun instant(newInstant: Instant) {
     instant = newInstant
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -6,6 +6,8 @@ import com.fasterxml.jackson.annotation.JsonInclude.Include
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.annotation.JsonPropertyOrder
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
+import com.netflix.spinnaker.keel.constraints.ConstraintStateAttributes
+import com.netflix.spinnaker.keel.constraints.ConstraintStatus
 import java.time.Instant
 
 /**
@@ -44,6 +46,7 @@ data class GitMetadata(
   val commit: String
 )
 
+@JsonInclude(Include.NON_EMPTY)
 data class ArtifactSummaryInEnvironment(
   @JsonProperty("name")
   val environment: String,
@@ -52,5 +55,17 @@ data class ArtifactSummaryInEnvironment(
   val state: String,
   val deployedAt: Instant? = null,
   val replacedAt: Instant? = null,
-  val replacedBy: String? = null
+  val replacedBy: String? = null,
+  val constraints: List<EnvironmentConstraintSummary> = emptyList()
+)
+
+@JsonInclude(Include.NON_NULL)
+data class EnvironmentConstraintSummary(
+  val type: String,
+  val status: ConstraintStatus,
+  val createdAt: Instant,
+  val judgedBy: String? = null,
+  val judgedAt: Instant? = null,
+  val comment: String? = null,
+  val attributes: ConstraintStateAttributes? = null
 )
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -4,12 +4,14 @@ import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.annotation.JsonInclude.Include
 import com.fasterxml.jackson.annotation.JsonProperty
+import com.fasterxml.jackson.annotation.JsonPropertyOrder
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
 import java.time.Instant
 
 /**
  * Summarized data about a specific artifact, mostly for use by the UI.
  */
+@JsonPropertyOrder(value = ["name", "type", "versions"])
 data class ArtifactSummary(
   val name: String,
   val type: ArtifactType,
@@ -17,6 +19,7 @@ data class ArtifactSummary(
 )
 
 @JsonInclude(Include.NON_NULL)
+@JsonPropertyOrder(value = ["name", "version", "state"])
 data class ArtifactVersionSummary(
   val version: String,
   val displayName: String,
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
@@ -1,6 +1,7 @@
 package com.netflix.spinnaker.keel.core.api
 
 import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonPropertyOrder
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactStatus
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
@@ -10,6 +11,7 @@ import com.netflix.spinnaker.keel.api.id
 /**
  * Summarized data about a specific environment, mostly for use by the UI.
  */
+@JsonPropertyOrder(value = ["name"])
 data class EnvironmentSummary(
   @JsonIgnore val environment: Environment,
   val artifacts: Set<ArtifactVersions>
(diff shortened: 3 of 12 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
```diff
@@ -38,7 +39,7 @@ class InMemoryArtifactRepository(
   private val artifacts = mutableMapOf<UUID, DeliveryArtifact>()
 
   private val approvedVersions = mutableMapOf<EnvironmentVersionsKey, MutableList<String>>()
-  private val deployedVersions = mutableMapOf<EnvironmentVersionsKey, MutableList<String>>()
+  private val deployedVersions = mutableMapOf<EnvironmentVersionsKey, MutableList<Pair<String, Instant>>>()
```
luispollo: We now track the deployment times alongside the versions so that we can fill in the `deployedAt` and `replacedAt` fields in the artifact summaries.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
```diff
@@ -49,9 +51,11 @@ class DefaultConfiguration {
   fun idGenerator(): ULID = ULID()
 
   @Bean
+  @Primary
```
luispollo: I added this while tracking down what I thought might have been an issue with wiring our customized object mapper into Spring REST controllers, which wasn't an issue in the end. I don't think this hurts, but I can remove it if you feel I should. Let me know.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
```diff
@@ -49,9 +51,11 @@ class DefaultConfiguration {
...
   fun objectMapper(): ObjectMapper = configuredObjectMapper()
 
   @Bean
+  @Qualifier("yamlMapper")
```
luispollo: To differentiate from the (JSON) `ObjectMapper` above since both beans qualify as `YAMLMapper` is a sub-class. This eliminates any ambiguity.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -26,14 +29,15 @@ import org.springframework.stereotype.Component
  */
 @Component
 class ApplicationService(
-  private val repository: KeelRepository,
-  private val artifactRepository: ArtifactRepository
+  private val resourceRepository: ResourceRepository,
+  private val artifactRepository: ArtifactRepository,
+  private val deliveryConfigRepository: DeliveryConfigRepository
```
luispollo: Little side-note, but wanted to call your attention to this class (and potentially a pattern) here. Personally, I really feel like lumping all the various repository methods into a giant interface like `KeelRepository` is not the best approach. Not only does it make that interface and the implementation extremely hard to read and understand, but it violates the single responsibility principle by adding business logic into what, by definition, should be a class concerned exclusively with persistence. 

The approach taken here, instead, is to have an object whose specific and declared purpose is to be a service object responsible for that kind of business logic. Upstream callers, like the `ApplicationController` in this case, call into the high-level methods presented by this class to abstract details related to calling multiple persistence repositories and adding additional processing of the data.

I would propose that we reconsider the combined repository approach in favor of purpose-built service objects for specific sub-areas of the domain model.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
```diff
@@ -4,19 +4,24 @@ import com.fasterxml.jackson.annotation.JsonIgnore
...
+import com.fasterxml.jackson.annotation.JsonPropertyOrder
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
+import com.netflix.spinnaker.keel.constraints.ConstraintStateAttributes
+import com.netflix.spinnaker.keel.constraints.ConstraintStatus
 import java.time.Instant
 
 /**
  * Summarized data about a specific artifact, mostly for use by the UI.
  */
+@JsonPropertyOrder(value = ["name", "type", "versions"])
```
emjburns: Thoughts on just using `@JsonPropertyOrder(alphabetic = true)` so that this doesn't have to be updated all the time?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
```diff
@@ -10,6 +11,7 @@ import com.netflix.spinnaker.keel.api.id
 /**
  * Summarized data about a specific environment, mostly for use by the UI.
  */
+@JsonPropertyOrder(value = ["name"])
```
emjburns: what is the purpose of only having one value here?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: But, the `KeelRepository` class is meant to deal with business logic - i.e. how do the objects relate, what order must things be done in, what happens when we delete one thing. Without a way to deal with all of the objects at once (delivery config, artifacts, resources) we will have consistency and concurrency bugs. We should talk about this offline.

For now, please use the `KeelRepository` here.

emjburns: You could in fact think of `KeelRepository` as a purpose built service object that itself is concerned with persistence so that things like the `ApplicationController` don't have to be.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
luispollo: I wanted names to always be on top because it's freaking annoying to find those in YAML down a bunch of other keys, especially when there's a lot of data. I don't mind keeping these up-to-date, and I don't expect the classes to change that frequently either.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
luispollo: It makes sure name is at the top. All other properties are arranged in whatever way Jackson thinks is the right way.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: Let's take it offline. I'll open an issue and transfer these comments there. I didn't want to loose the thought but should have done that in the first place. My bad.

luispollo: Opened #927

luispollo:
> Please use the `KeelRepository`. I understand that you find it clunky for now, but it is not a permanent solution. If you deviate from the pattern in `ApplicationService` it will make it a lot harder for me to clean things up later.

Done.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -26,8 +27,7 @@ import org.springframework.stereotype.Component
  */
 @Component
 class ApplicationService(
-  private val repository: KeelRepository,
-  private val artifactRepository: ArtifactRepository
+  private val repository: KeelRepository
```
luispollo: @emjburns As requested.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
emjburns: okay! Makes sense. That is very annoying.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
emjburns: oh cool!

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
```diff
@@ -171,17 +199,38 @@ internal class ApplicationControllerTests : JUnit5Minutests {
...
-        artifactRepository.markAsSuccessfullyDeployedTo(deliveryConfig, artifact, "fnord-1.0.0-h0.2add2c9", "staging")
-        artifactRepository.markAsSuccessfullyDeployedTo(deliveryConfig, artifact, "fnord-1.0.0-h0.2add2c9", "production")
-        artifactRepository.markAsSuccessfullyDeployedTo(deliveryConfig, artifact, "fnord-1.0.1-h1.2add2c9", "test")
-        artifactRepository.markAsSuccessfullyDeployedTo(deliveryConfig, artifact, "fnord-1.0.1-h1.2add2c9", "staging")
-        artifactRepository.markAsSuccessfullyDeployedTo(deliveryConfig, artifact, "fnord-1.0.2-h2.2add2c9", "test")
-        artifactRepository.approveVersionFor(deliveryConfig, artifact, "fnord-1.0.3-h3.2add2c9", "test")
+        artifactRepository.store(artifact, "fnord-1.0.0-h0.a0a0a0a", ArtifactStatus.RELEASE)
+        artifactRepository.store(artifact, "fnord-1.0.1-h1.b1b1b1b", ArtifactStatus.RELEASE)
+        artifactRepository.store(artifact, "fnord-1.0.2-h2.c2c2c2c", ArtifactStatus.RELEASE)
+        artifactRepository.store(artifact, "fnord-1.0.3-h3.d3d3d3d", ArtifactStatus.RELEASE)
```
luispollo: @emjburns I took the trouble of using different hashes for each version. 🎁 😄

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: Thank you! :)

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
emjburns: wowie! So realistic 😍

erikmunson:
🎉 💃 

Just for my own planning/coordination purposes on the UI side: there was also a requirement discussed in #888 around creating a 'synthetic' (i.e. not explicitly recorded directly in the constraint table) `NOT_STARTED` status for stateful constraints when exposed in the API so that we can indicate that a constraint _will run later_ even though it's not actually written to the database yet. Do you intend to add that functionality as a follow up change directly or come back to it later? I can flex/adjust my implementation plan depending on the lead time for getting that additional status.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/time/MutableClock.kt
```diff
@@ -27,13 +28,17 @@ class MutableClock(
     instant = instant.plus(amount)
   }
 
-  fun tickSeconds(seconds: Long) = incrementBy(Duration.ofSeconds(seconds))
+  fun tickSeconds(seconds: Long) = incrementBy(Duration.ofSeconds(seconds)).let { instant }
```
gal-yardeni: I don't understand this part - what does `.let { instant }` is doing?

luispollo:
> Just for my own planning/coordination purposes on the UI side: there was also a requirement discussed in #888 around creating a 'synthetic' (i.e. not explicitly recorded directly in the constraint table) `NOT_STARTED` status for stateful constraints when exposed in the API so that we can indicate that a constraint _will run later_ even though it's not actually written to the database yet. Do you intend to add that functionality as a follow up change directly or come back to it later? I can flex/adjust my implementation plan depending on the lead time for getting that additional status.

Oh! Oh! I totally forgot! 🤦‍♂ Will open another small PR for that before releasing.

erikmunson:
ok! feel free to cut a release now and get it out for testing — i'm not in a huge hurry to get that extra status as long as I have a rough sense of when it'll be available, there's _tons_ of work I can do with this data as-is while you're working on a follow-up.

PR #929, opened 2020-03-26
fix(bake): let's not keep rebaking forever.

emjburns:
Rebaking is extremely unlikely to fix any bake problems. Let's not rebake unless the diff changes.

This pr uses some unhappy controls that asher put into place.

Alternatively to not rebaking, I could change this to rebake every day or every 6 hours instead. Thoughts?

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
@@ -2,7 +2,9 @@ package com.netflix.spinnaker.keel.bakery.api
 
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.netflix.spinnaker.keel.api.ResourceSpec
+import com.netflix.spinnaker.keel.api.UnhappyControl
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactStatus
+import java.time.Duration
 
 data class ImageSpec(
   val artifactName: String,
@@ -12,7 +14,13 @@ data class ImageSpec(
   val storeType: StoreType,
   val artifactStatuses: Set<ArtifactStatus> = enumValues<ArtifactStatus>().toSet(),
   override val application: String // the application an image is baked in
-) : ResourceSpec {
+) : ResourceSpec, UnhappyControl {
   @JsonIgnore
   override val id: String = artifactName
+
+  @JsonIgnore
+  override val maxDiffCount = 2
+
+  @JsonIgnore
+  override val unhappyWaitTime = Duration.ZERO
 }


lorin:
What happens if the user wants to manually attempt a rebake? (e.g., they look at the bake error, it looks to them like a transient issue). How would they do that?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
```diff
@@ -12,7 +14,14 @@ data class ImageSpec(
   val storeType: StoreType,
   val artifactStatuses: Set<ArtifactStatus> = enumValues<ArtifactStatus>().toSet(),
   override val application: String // the application an image is baked in
-) : ResourceSpec {
+) : ResourceSpec, UnhappyControl {
   @JsonIgnore
   override val id: String = artifactName
+
+  @JsonIgnore
+  override val maxDiffCount = 2
```
lorin: Does this mean it will retry baking twice if it fails?

emjburns:
@lorin currently there's no way to rebake except changing the diff - so you'd need to push a new artifact version to rebake. Or, you could create a pipeline to rebake for you. That's a shitty experience. I'm not convinced it's shittier than randomly seeing 1000 bake tasks cause we're doing something wrong on our end. Thoughts?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
emjburns: Yes it does I think. I just stole whatever number we were using for clusters w/o digging too much into it

lorin:
@emjburns My first instinct is to provide some sort of UI to allow the user to manually override an "UnhappyVeto" state. That would provide a way for them to try the bake again in this case. (I wouldn't gate this PR on that, although I'd like to know how to manually do that behind the scenes by tweaking the database if somebody asked for this in a support context).

emjburns:
We have the capability to do that! you can "pass a message" to a veto (bad naming...it's my fault). So, we could theoretically have an api you could hit to trigger a rebake. We'd have to make some changes in `UnhappyVeto` to support that.

As far as what to do to unblock a user, i'm pretty sure you could just delete the right row from the `unhappy_veto` table

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
```diff
@@ -12,7 +14,14 @@ data class ImageSpec(
...
+
+  // rebaking does not help remedy any of the bake problems we see.
```
luispollo: So zero means "wait forever or until the diff changes". I feel the comment doesn't quite fit with that. Maybe just remove the comment?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
asher: It's the number of attempts in total, so 1 retry.

robfletcher: That seems to conflict with what the PR says which is that there's no point retrying. Shouldn't this be `1`?

emjburns: Ehhh, maybe i'll leave it at 2 so it retries once. I can always bump it down.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
emjburns: Sure, I"ll remove it.

emjburns:
added an issue for a rebake api

Issue #915, opened 2020-03-24
Versions being deployed are stuck in the 'approved' state on EnvironmentSummary and ArtifactSummary


erikmunson:
In theory, the way the EnvironmentSummary and ArtifactSummary APIs work is that when an artifact is `approved` for an environment, it subsequently enters the [`deploying`](https://github.com/spinnaker/keel/blob/master/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt#L48) state as tasks are launched and resources flip to the `ACTUATING` status.

In practice, it appears that right now versions enter `approved` and get stuck there until deployment is done. Resources flip to `DIFF` then `ACTUATING` as tasks are launched, and then eventually to `HAPPY` once the diff is resolved. The whole time the version stays `approved`, and then when things are all over it switches directly to `current` without ever hitting `deployed`.

Because this API is new and not thoroughly used, it's not clear to me whether this ever worked or if we're looking at some sort of regression. Either way it's critical that we be able to show things in a `deploying` state to properly communicate what's happening, so I consider this a blocker for getting the UI out to customers.

PR #932, opened 2020-03-27
fix(artifacts): mark as deploying, version in task description

emjburns:
Fixes https://github.com/spinnaker/keel/issues/915:
(1) Adds "DEPLOYING" when we launch tasks to deploy new versions to clusters. Note: for Titus that means we have to translate the digest to the desired app version. 

(2) Since we're already calculating the version to send in an event right after the task launches, also include the version in the task description so it shows up in the UI (like `Upsert server group emburnstest-md-test2 to deb-sample-app-server-0.117.0-h347.3cc480c in account/region`

### DIFF
--- a/keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployingListenerTests.kt
+++ b/keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployingListenerTests.kt
@@ -1,0 +1,74 @@
+package com.netflix.spinnaker.keel.artifact
+
+import com.netflix.spinnaker.keel.api.Resource
+import com.netflix.spinnaker.keel.api.id
+import com.netflix.spinnaker.keel.events.ArtifactVersionDeploying
+import com.netflix.spinnaker.keel.test.DummyResourceSpec
+import com.netflix.spinnaker.keel.test.combinedMockRepository
+import com.netflix.spinnaker.keel.test.deliveryConfig
+import com.netflix.spinnaker.keel.test.resource
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import io.mockk.every
+import io.mockk.spyk
+import io.mockk.verify
+
+internal class ArtifactDeployingListenerTests : JUnit5Minutests {
+  class Fixture {
+    val r: Resource<DummyResourceSpec> = spyk(resource())
+    val config = deliveryConfig(resource = r)
+    val artifact = config.artifacts.first()
+    val repository = combinedMockRepository()
+
+    val event = ArtifactVersionDeploying(
+      resourceId = r.id,
+      artifactVersion = "1.1.1"
+    )
+
+    val subject = ArtifactDeployingListener(repository)
+  }
+
+  fun tests() = rootContext<Fixture> {
+    fixture { Fixture() }
+
+    before {
+      every { repository.getResource(r.id) } returns r
+      every { repository.deliveryConfigFor(r.id) } returns config
+      every { repository.environmentFor(r.id) } returns config.environments.first()
+    }
+
+    context("no artifact associated with the resource") {
+      test("nothing is marked deploying") {
+        subject.onArtifactVersionDeploying(event)
+        verify(exactly = 0) { repository.isApprovedFor(config, any(), event.artifactVersion, any()) }
+        verify(exactly = 0) { repository.markAsDeployingTo(config, any(), event.artifactVersion, any()) }
+      }
+    }
+
+    context("an artifact is associated with a resource") {
+      before {
+        every { r.findAssociatedArtifact(config) } returns artifact
+      }
+      context("artifact is approved for env") {
+        before {
+          every { repository.isApprovedFor(any(), any(), event.artifactVersion, any()) } returns true
+        }
+
+        test("version is marked deploying") {
+          subject.onArtifactVersionDeploying(event)
+          verify(exactly = 1) { repository.markAsDeployingTo(any(), any(), event.artifactVersion, any()) }
+        }
+      }
+
+      context("artifact is not approved for env") {
+        before {
+          every { repository.isApprovedFor(any(), any(), event.artifactVersion, any()) } returns false
+        }
+        test("nothing is marked as deploying") {
+          subject.onArtifactVersionDeploying(event)
+          verify(exactly = 0) { repository.markAsDeployingTo(config, any(), event.artifactVersion, any()) }
+        }
+      }
+    }
+  }
+}
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ArtifactVersionDeploying.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ArtifactVersionDeploying.kt
@@ -1,0 +1,9 @@
+package com.netflix.spinnaker.keel.events
+
+/**
+ * An event fired when we launch a task to deploy a new version to a cluster
+ */
+data class ArtifactVersionDeploying(
+  val resourceId: String,
+  val artifactVersion: String
+)
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/MissingAppVersionException.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/MissingAppVersionException.kt
@@ -1,0 +1,7 @@
+package com.netflix.spinnaker.keel.ec2
+
+import com.netflix.spinnaker.kork.exceptions.IntegrationException
+
+class MissingAppVersionException(
+  val resourceId: String
+) : IntegrationException("Resource $resourceId is missing version information. Is the appVersion tag missing from the AMI?")
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -53,8 +53,10 @@ import com.netflix.spinnaker.keel.core.orcaClusterMoniker
 import com.netflix.spinnaker.keel.core.serverGroup
 import com.netflix.spinnaker.keel.diff.toIndividualDiffs
 import com.netflix.spinnaker.keel.ec2.CLOUD_PROVIDER
+import com.netflix.spinnaker.keel.ec2.MissingAppVersionException
 import com.netflix.spinnaker.keel.ec2.SPINNAKER_EC2_API_V1
 import com.netflix.spinnaker.keel.events.ArtifactVersionDeployed
+import com.netflix.spinnaker.keel.events.ArtifactVersionDeploying
 import com.netflix.spinnaker.keel.exceptions.ExportError
 import com.netflix.spinnaker.keel.orca.OrcaService
 import com.netflix.spinnaker.keel.orca.dependsOn
@@ -117,14 +119,24 @@ class ClusterHandler(
           modifyInPlace(resource, modifyDiffs))
       }
 
+      val version = diffs.first().desired.launchConfiguration.appVersion
+        ?: throw MissingAppVersionException(resource.id)
+
       if (resource.spec.deployWith.isStaggered && createDiffs.isNotEmpty()) {
-        val tasks = upsertStaggered(resource, createDiffs)
+        val tasks = upsertStaggered(resource, createDiffs, version)
         return@coroutineScope tasks + deferred.map { it.await() }
       }
 
       deferred.addAll(
-        upsertUnstaggered(resource, createDiffs)
+        upsertUnstaggered(resource, createDiffs, version)
       )
+
+      if (createDiffs.isNotEmpty()) {
+        publisher.publishEvent(ArtifactVersionDeploying(
+          resourceId = resource.id,
+          artifactVersion = version
+        ))
+      }
 
       return@coroutineScope deferred.map { it.await() }
     }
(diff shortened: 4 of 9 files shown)


keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployingListenerTests.kt
```diff
@@ -0,0 +1,66 @@
...
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.verify
+
+internal class ArtifactDeployingListenerTests : JUnit5Minutests {
+  class Fixture {
+    val realResource: Resource<DummyResourceSpec> = resource()
+    val r: Resource<DummyResourceSpec> = mockk()
```
luispollo: I think you can use a Spy to simplify this: https://github.com/mockk/mockk#spy

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployingListenerTests.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    context("no artifact") {
+      before {
+        every { r.findAssociatedArtifact(config) } returns null
+      }
+
+      test("nothing is marked deploying") {
+        subject.onArtifactVersionDeploying(event)
+        verify(exactly = 0) { repository.isApprovedFor(config, any(), event.artifactVersion, any()) }
```
luispollo: I recently learned from @robfletcher that you can write this in a much more human-friendly way:
```kotlin
verify { repository.isApprovedFor(config, any(), event.artifactVersion, any()) wasNot Called }
```

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployingListenerTests.kt
```diff
@@ -0,0 +1,66 @@
...
+        verify(exactly = 0) { repository.markAsDeployingTo(config, any(), event.artifactVersion, any()) }
+      }
+    }
+
+    context("an artifact") {
```
luispollo:
```suggestion
    context("an artifact is approved") {
```

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployingListenerTests.kt
```diff
@@ -0,0 +1,66 @@
...
+      before {
+        every { r.findAssociatedArtifact(config) } returns artifact
+        every { repository.isApprovedFor(any(), any(), event.artifactVersion, any()) } returns true
```
luispollo: It'd be nice to add another test for when an artifact is associated with the resource but the version has not been approved.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ArtifactVersionDeploying.kt
```diff
@@ -0,0 +1,10 @@
+package com.netflix.spinnaker.keel.events
+
+/**
+ * An event fired when we launch a task to deploy a new version to a cluster
+ */
+data class ArtifactVersionDeploying(
+  val resourceId: String,
+  val artifactVersion: String,
```
luispollo: Would it be useful to have a reference to the `DeliveryArtifact` here, even if only for informational purposes (e.g. logging)?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ArtifactVersionDeploying.kt
```diff
@@ -0,0 +1,10 @@
...
+  val provider: String = "aws"
```
luispollo: Why default this? I don't think we should have a preference for providers. 😄

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/MissingAppVersionException.kt
```diff
@@ -0,0 +1,7 @@
+package com.netflix.spinnaker.keel.ec2
+
+import com.netflix.spinnaker.kork.exceptions.IntegrationException
+
+class MissingAppVersionException(
+  val resourceId: String
+) : IntegrationException("Resource $resourceId is missing the appVersion, is it present on the AMI?")
```
luispollo: TIL about this `IntegrationException`.
Question: I think the exception message may be a little cryptic if it bubbles up all the way. Can we maybe make that a bit more friendly? Where does the `appVersion` come from in the AMI, is it a tag? If it is, we could use something like "Resource X is missing version information. Is the appVersion tag missing from the AMI?".

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -189,7 +202,7 @@ class ClusterHandler(
           async {
             taskLauncher.submitJob(
               resource = resource,
-              description = "Upsert server group ${diff.desired.moniker} in " +
+              description = "Upsert server group ${diff.desired.moniker} to $version in " +
```
luispollo: ❤️

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -261,7 +275,7 @@ class ClusterHandler(
         val deferred = async {
           taskLauncher.submitJob(
             resource = resource,
-            description = "Upsert server group ${diff.desired.moniker} in " +
+            description = "Upsert server group ${diff.desired.moniker} to $version in " +
```
luispollo: ❤️

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -126,16 +127,45 @@ class TitusClusterHandler(
...
+          var tags: Set<String> = emptySet()
+          val version = when {
+            diff.isCapacityOnly() -> null
+            else -> {
+              // calculate the version for the digest
+              tags = getTagsForDigest(desired.container, desired.location.account).filterNot { it == "latest" }.toSet()
+              if (tags.size == 1) {
+                tags.first()
+              } else {
+                log.debug("Container digest ${desired.container} has multiple tags: $tags")
```
luispollo: Would it be safe to sort the tags in descending order and pick the first? As in, the latest tag applied?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -126,16 +127,45 @@ class TitusClusterHandler(
...
               resource = resource,
-              description = "Upsert server group ${desired.moniker} in ${desired.location.account}/${desired.location.region}",
+              description = description,
               correlationId = "${resource.id}:${desired.location.region}",
               job = job
             )
           }
+
+          tags.forEach { tag ->
+            publisher.publishEvent(ArtifactVersionDeploying(
```
luispollo: Why one event for every tag? We just went through all this logic above to pick the one version string to use. Shouldn't we use that to generate a single event here?

luispollo: BTW, whatever the case, there should be tests covering the expected behavior when there are multiple tags for a digest.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -319,25 +349,33 @@ class TitusClusterHandler(
...
                 provider = "titus"
               ))
             }
-          }
         }
       }
 
+  /**
+   * Get all tags that match a digest.
+   * Note: there may be more than one tag with the same digest (like `latest` and a version tag)
```
luispollo: I see that we end up filtering the `"latest"` tag at least for the `ArtifactVersionDeploying` event above. Should we filter it out here, at the "source"?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
luispollo: On second thought, never mind. I'm assuming the UI will have a link that takes the user to the image, where they'll see it has multiple tags. Perhaps just use the short hash for the digest instead?

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployingListenerTests.kt
emjburns: oh thank goodness, i was trying to figure out how to do that but i didn't know the right words to search for!!! thank you!

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployingListenerTests.kt
emjburns: updating the wording to make the test clearer

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ArtifactVersionDeploying.kt
emjburns: No, we need to look it up in the handler for this event because that's where we have access to the delivery config repository.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/MissingAppVersionException.kt
emjburns: This should never happen TM, but i have seen this situation maybe once... It'll need us to debug regardless since it requires a bunch of knowledge of the system and our netflix assumptions. It does come from the ami tag, i'll update the exception to be the language you suggested.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
emjburns: 👍 I'll shorten the hash here so it's not so long. Until we see a bunch of situations where this happens i'm not sure what the right behavior is - but I think you're right, you'll just need to look at the image yourself.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
emjburns: Well, we know what digest we're deploying. But, for docker images, there can be multiple tags for each digest. Say you publish your new image, and you tag it with the actual version `1.1.1`, and the `latest` tag happens automatically, but you're also doing some testing so you tag it `my-test-v1.1.1`. We have no way of knowing here which version you care about for managed delivery's sake. So the `getTags` list for this digest will return [`1.1.1`, `my-test-v1.1.1`, `latest`]. We filter out `latest`, cause we never care about that. But, in this context, how can we know which version string is the one that matches the format we're tracking, and which is the format that we can discard? `.first()` is meaningless. So, we send out an event for each version. In the handler for this event, we only mark as `DEPLOYING` if it's already been approved for the environment (which is how we ignore versions we're not explicitly caring about).

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
emjburns: hmm, ya it's probably better to do here since we'll never want `latest` in this class.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ArtifactVersionDeploying.kt
emjburns: ehh, was just copying from elsewhere. I can remove

emjburns: Actually I don't even need this field now. I'll just remove it!

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
gal-yardeni: Nice!

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
emjburns: Added tests for each publishing

Issue #940, opened 2020-03-28
Allow for updating constraint status via application name (instead of manifest name)


erikmunson:
similar to #836, currently [the endpoint for updating constraint status](https://github.com/spinnaker/keel/blob/master/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt#L98-L124) requires a specific manifest name to be supplied in order to work. On the UI side we ideally need a version of this endpoint that accepts only an application name + environment name.

Until we're able to make this call without a manifest name, we can't update the state of constraints like `manual-judgement` interactively in the UI.

Issue #816, opened 2020-02-20
Check service account permissions in relevant controller methods


luispollo:
Need to enable user's authorization for "write" permissions against the service account for the following:
- callback from interactive notification to change manual judgement constraint state (`InteractiveNotificationCallbackController`)
- change constraint states in general (`DeliveryConfigController`)
- create, update and delete delivery configs (`DeliveryConfigController`)

...any other relevant REST calls (possibly, nearly every call).

PR #942, opened 2020-03-30
chore(api): endpoint to judge constraints by application

emjburns:
Closes https://github.com/spinnaker/keel/issues/940.

The UI needs to be able to judge constraints by using the application name instead of the delivery config name. This PR adds a new endpoint that is the same as the delivery config endpoint (see linked issue). This new endpoint has auth protections as well.

I removed the old endpoint, and also updated the tests for this functionality. I moved all tests for constraints and generating the views to the `ApplicationServiceTests` class. I kept actual controller tests (like auth rejection) in the controller tests.

### DIFF
--- a/keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/CombinedRepository.kt
+++ b/keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/CombinedRepository.kt
@@ -14,15 +14,17 @@ import java.time.Clock
  * A util for generating a combined repository with in memory implementations for tests
  */
 fun combinedInMemoryRepository(
-  deliveryConfigRepository: DeliveryConfigRepository = InMemoryDeliveryConfigRepository(),
-  artifactRepository: ArtifactRepository = InMemoryArtifactRepository(),
-  resourceRepository: ResourceRepository = InMemoryResourceRepository()
+  clock: Clock = Clock.systemUTC(),
+  deliveryConfigRepository: DeliveryConfigRepository = InMemoryDeliveryConfigRepository(clock),
+  artifactRepository: ArtifactRepository = InMemoryArtifactRepository(clock),
+  resourceRepository: ResourceRepository = InMemoryResourceRepository(clock)
 ): CombinedRepository =
-  CombinedRepository(deliveryConfigRepository, artifactRepository, resourceRepository, Clock.systemUTC(), mockk(relaxed = true))
+  CombinedRepository(deliveryConfigRepository, artifactRepository, resourceRepository, clock, mockk(relaxed = true))
 
 fun combinedMockRepository(
   deliveryConfigRepository: DeliveryConfigRepository = mockk(relaxed = true),
   artifactRepository: ArtifactRepository = mockk(relaxed = true),
-  resourceRepository: ResourceRepository = mockk(relaxed = true)
+  resourceRepository: ResourceRepository = mockk(relaxed = true),
+  clock: Clock = Clock.systemUTC()
 ): CombinedRepository =
-  CombinedRepository(deliveryConfigRepository, artifactRepository, resourceRepository, Clock.systemUTC(), mockk(relaxed = true))
+  CombinedRepository(deliveryConfigRepository, artifactRepository, resourceRepository, clock, mockk(relaxed = true))
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
@@ -17,15 +17,13 @@
  */
 package com.netflix.spinnaker.keel.rest
 
+import com.netflix.spinnaker.keel.constraints.ConstraintState
 import com.netflix.spinnaker.keel.constraints.UpdatedConstraintStatus
-import com.netflix.spinnaker.keel.exceptions.InvalidConstraintException
 import com.netflix.spinnaker.keel.pause.ActuationPauser
-import com.netflix.spinnaker.keel.persistence.KeelRepository
-import com.netflix.spinnaker.keel.persistence.NoDeliveryConfigForApplication
+import com.netflix.spinnaker.keel.persistence.ResourceRepository.Companion.DEFAULT_MAX_EVENTS
 import com.netflix.spinnaker.keel.services.ApplicationService
 import com.netflix.spinnaker.keel.yaml.APPLICATION_YAML_VALUE
 import com.netflix.spinnaker.kork.web.exceptions.InvalidRequestException
-import java.time.Instant
 import org.slf4j.LoggerFactory
 import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
 import org.springframework.security.access.prepost.PreAuthorize
@@ -43,8 +41,7 @@ import org.springframework.web.bind.annotation.RestController
 @RequestMapping(path = ["/application"])
 class ApplicationController(
   private val actuationPauser: ActuationPauser,
-  private val applicationService: ApplicationService,
-  private val repository: KeelRepository
+  private val applicationService: ApplicationService
 ) {
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
@@ -90,23 +87,19 @@ class ApplicationController(
     @PathVariable("environment") environment: String,
     @RequestBody status: UpdatedConstraintStatus
   ) {
-    // we assume there's one manifest per application
-    val manifest = repository.getDeliveryConfigsByApplication(application).firstOrNull()
-      ?: throw NoDeliveryConfigForApplication(application)
-    val currentState = repository.getConstraintState(
-      manifest.name,
-      environment,
-      status.artifactVersion,
-      status.type) ?: throw InvalidConstraintException(
-      "${manifest.name}/$environment/${status.type}/${status.artifactVersion}", "constraint not found")
-
-    repository.storeConstraintState(
-      currentState.copy(
-        status = status.status,
-        comment = status.comment ?: currentState.comment,
-        judgedAt = Instant.now(),
-        judgedBy = user))
+    applicationService.updateConstraintStatus(user, application, environment, status)
   }
+
+  @GetMapping(
+    path = ["/{application}/environment/{environment}/constraints"],
+    produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  fun getConstraintState(
+    @PathVariable("application") application: String,
+    @PathVariable("environment") environment: String,
+    @RequestParam("limit") limit: Int?
+  ): List<ConstraintState> =
+    applicationService.getConstraintStatesFor(application, environment, limit ?: DEFAULT_MAX_EVENTS)
 
   @PostMapping(
     path = ["/{application}/pause"]
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
@@ -17,16 +17,24 @@
  */
 package com.netflix.spinnaker.keel.rest
 
+import com.netflix.spinnaker.keel.constraints.UpdatedConstraintStatus
+import com.netflix.spinnaker.keel.exceptions.InvalidConstraintException
 import com.netflix.spinnaker.keel.pause.ActuationPauser
+import com.netflix.spinnaker.keel.persistence.KeelRepository
+import com.netflix.spinnaker.keel.persistence.NoDeliveryConfigForApplication
 import com.netflix.spinnaker.keel.services.ApplicationService
 import com.netflix.spinnaker.keel.yaml.APPLICATION_YAML_VALUE
 import com.netflix.spinnaker.kork.web.exceptions.InvalidRequestException
+import java.time.Instant
 import org.slf4j.LoggerFactory
 import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
+import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.web.bind.annotation.DeleteMapping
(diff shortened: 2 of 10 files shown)


asher:
This requires making environment names unique per application. If that's adopted, it may make sense to revisit whether allowing multiple delivery configs per application makes sense. If not, `DeliveryConfigController` could be deprecated entirely, in favor of an application centric api.

emjburns:
@asher ya my plan is to make it a requirement at the api layer today, but keep everything in the back end still referring to the delivery config so that it's easy to back that assumption out later. 

Good point about removing the `DeliveryConfigController` entirely. I'll think about that.

emjburns:
Basic limiting: https://github.com/spinnaker/keel/pull/944

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
```diff
@@ -146,6 +146,7 @@ interface DeliveryConfigRepository : PeriodicallyCheckedRepository<DeliveryConfi
 
 sealed class NoSuchDeliveryConfigException(message: String) : UserException(message)
 class NoSuchDeliveryConfigName(name: String) : NoSuchDeliveryConfigException("No delivery config named $name exists in the repository")
+class NoDeliveryConfigForApplication(application: String) : NoSuchDeliveryConfigException("No delivery config for application $application exists in the repository")
```
luispollo:
```suggestion
class NoDeliveryConfigForApplication(application: String) : NoSuchDeliveryConfigException("No delivery config for application $application exists in the database")
```

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -35,7 +43,8 @@ import org.springframework.web.bind.annotation.RestController
 @RequestMapping(path = ["/application"])
 class ApplicationController(
   private val actuationPauser: ActuationPauser,
-  private val applicationService: ApplicationService
+  private val applicationService: ApplicationService,
+  private val repository: KeelRepository
```
luispollo: The idea of introducing `ApplicationService` was to avoid the controller doing any business logic. Consider adding a new method to `ApplicationService` that you call from the controller for the new API.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -69,6 +78,36 @@ class ApplicationController(
     }
   }
 
+  @PostMapping(
+    path = ["/{application}/environment/{environment}/constraint"],
+    consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
+    produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  @PreAuthorize("@authorizationSupport.userCanModifyApplication(#application)")
```
luispollo: Cool! I'm getting started on #816 right now -- this gives me a good example to follow! 👍

luispollo: P.S. It'd be nice to add tests for the pass/fail auth check.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -69,6 +78,36 @@ class ApplicationController(
...
+    @PathVariable("environment") environment: String,
+    @RequestBody status: UpdatedConstraintStatus
+  ) {
+    // we assume there's one manifest per application
+    val manifest = repository.getDeliveryConfigsByApplication(application).firstOrNull()
+      ?: throw NoDeliveryConfigForApplication(application)
+    val currentState = repository.getConstraintState(
+      manifest.name,
+      environment,
+      status.artifactVersion,
```
erikmunson: a couple very tactical notes/questions about the current state of this endpoint:
1. How is this going to handle cases where there's multiple constraints of the same type? At the moment we have no way of uniquely identifying a constraint when there's multiple of the same type, and I expect we'll need that in the foreseeable future (not in this PR)
2. Is a version _on its own_ enough to globally identify a specific artifact version in a config? I realize at Netflix most debs/docker images have the artifact name encoded into the version, but in general my expectation/understanding was that you needed a tuple of name + type + version to uniquely identify a version within a config.

If one or both of those is worth a GitHub issue, I'll happily start one up. Again I doubt we'll need to worry about it in the next few weeks, but might in the next few months.

erikmunson:
Can't speak to the impl, but controller contract/signature looks good. Thank you!

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
```diff
@@ -46,13 +46,26 @@ class AuthorizationSupport(
...
+    userCanModifyDeliveryConfig(deliveryConfig.serviceAccount, application, deliveryConfig.name)
+  } catch (e: NoSuchElementException) {
+    // If there are no delivery configs for that application return true so this is handled correctly in the controller.
+    true
+  }
+
+  fun userCanModifyDeliveryConfig(serviceAccount: String, application: String, manifestName: String): Boolean {
+    val auth = SecurityContextHolder.getContext().authentication
+    return userCanAccessServiceAccount(auth, serviceAccount, "Delivery config for application $application: $manifestName")
   }
```
luispollo: You can simplify this and make it useful for other use cases like so:
```suggestion
  fun userCanModifyDeliveryConfig(deliveryConfig: DeliveryConfig): Boolean {
    val auth = SecurityContextHolder.getContext().authentication
    return userCanAccessServiceAccount(
      auth,
      deliveryConfig.serviceAccount,
      "Delivery config for application ${deliveryConfig.application}: ${deliveryConfig.name}"
    )
  }
```

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
emjburns: Please start github issues for problems. This is strictly a port of the existing api. I'm happy to redesign functionality or improve it in the future

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
emjburns: Done, also added tests and moved tests from the `ApplicationControllerTests` to the new `ApplicationServiceTests`

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
emjburns: Done but only for failing, because that's the only relevant thing

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -84,44 +77,5 @@ class DeliveryConfigController(
     //  in [ExceptionHandler])
     mapOf("status" to "valid")
```
emjburns: These are moved to the `ApplicationController` for ease of use in the UI.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
```diff
@@ -159,26 +157,6 @@ internal class ApplicationControllerTests : JUnit5Minutests {
               ),
               _defaults = clusterDefaults
             )
```
emjburns: Moved a lot of these tests out of this class, but the actual tests stayed the same in `ApplicationServiceTests`

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
```diff
@@ -252,84 +244,6 @@ internal class DeliveryConfigControllerTests : JUnit5Minutests {
                 ?.constraints
             ).isNotNull()
           }
```
emjburns: Moved to `ApplicationServiceTests`, and refactored to not be rest tests

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
luispollo: Just a heads-up that you may have a conflict with #949 which is about to merge.

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/CombinedRepository.kt
```diff
@@ -14,15 +14,17 @@ import java.time.Clock
  * A util for generating a combined repository with in memory implementations for tests
  */
 fun combinedInMemoryRepository(
-  deliveryConfigRepository: DeliveryConfigRepository = InMemoryDeliveryConfigRepository(),
-  artifactRepository: ArtifactRepository = InMemoryArtifactRepository(),
-  resourceRepository: ResourceRepository = InMemoryResourceRepository()
+  clock: Clock = Clock.systemUTC(),
```
luispollo: Thanks!

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
luispollo: That will break the API contract and require changes to gate, no?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/ApplicationServiceTests.kt
```diff
@@ -0,0 +1,358 @@
...
+          setOf(
+            DependsOnConstraint("staging"),
+            ManualJudgementConstraint(),
+            PipelineConstraint(pipelineId = "fakePipeline")
+          )
+        } else {
+          emptySet()
+        },
+        resources = setOf(
+          // cluster with new-style artifact reference
```
luispollo: The "old-style" cluster spec with image details that existed in the original tests was important to check that retrieving cluster -> artifact linkage worked for that case too.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/ApplicationServiceTests.kt
```diff
@@ -0,0 +1,358 @@
...
+            artifact.name,
+            artifact.type,
+            artifact.statuses,
+            ArtifactVersionStatus(
+              current = "fnord-1.0.2-h2.c2c2c2c",
+              pending = listOf(),
+              approved = listOf("fnord-1.0.3-h3.d3d3d3d"),
+              previous = listOf("fnord-1.0.0-h0.a0a0a0a", "fnord-1.0.1-h1.b1b1b1b"),
+              vetoed = listOf(),
+              deploying = null
```
luispollo: These turned out pretty easy to read in this format! 🏆

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/ApplicationServiceTests.kt
```diff
@@ -0,0 +1,358 @@
...
+        }
+      }
+
+      test("no constraints have been evaluated") {
+        val states = applicationService.getConstraintStatesFor(application, "prod", 10)
+        expectThat(states).isEmpty()
+      }
+
+      test("pending manual judgement") {
+        val judgement = ConstraintState(deliveryConfig.name, "production", "fnord-1.0.2-h2.c2c2c2c", "manual-judgement", PENDING)
```
luispollo: Any reason you can't just use the one stored in the big `before` block for all tests? Then you don't have to check that size is 2 below, which was confusing to me.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
emjburns: yep, it will, but that's only used in the slack mj response api, right?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/ApplicationServiceTests.kt
gal-yardeni: SO MUCH better!

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/ApplicationServiceTests.kt
emjburns: ah ok, i'll add it back in

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/ApplicationServiceTests.kt
luispollo: Oh, thank you very much for my part, hahaha! 🤣

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
emjburns: Will resolve the conflict now

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
luispollo: Well, it was used up until now as the only way to set constraint state _besides_ manual judgement. But since the whole point of the PR is to unblock the new UI to do this, I think we're good.

PR #944, opened 2020-03-30
feat(api): limit 1 delivery config per app

emjburns:
Closes https://github.com/spinnaker/keel/issues/676.

A first pass at limiting delivery configs to one per app.

Right now, we're making the choice to limit delivery configs to one per app so that it's easy to work with from the UI. This assumption might change as we get more adoption. I've put this one easy limit together now, which will prevent you from saving a new delivery config if one exists for your app. 

Do you think we need more precautions than this?

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
@@ -379,6 +379,39 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
           expectThat(subject.allResourceNames().size).isEqualTo(0)
         }
       }
+
+      context("a second delivery config for an app fails to persist") {
+        val submittedConfig1 = SubmittedDeliveryConfig(
+          name = configName,
+          application = "keel",
+          serviceAccount = "keel@spinnaker",
+          artifacts = setOf(DockerArtifact(name = "org/thing-1", deliveryConfigName = configName, reference = "thing")),
+          environments = setOf(
+            SubmittedEnvironment(
+              name = "test",
+              resources = setOf(
+                SubmittedResource(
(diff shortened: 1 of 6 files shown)


asher:
Consider adding a unique index on the `delivery_config.application` column as well. Could be instead of the existing `delivery_config.name` unique index, if part of a broader change to address delivery configs by application.`name` no longer seems to be meaningful attribute for delivery configs.

emjburns:
Good point @asher. I've added a unique index for `delivery_config.application`. I will leave the change like this for now. I don't want to just address configs by application, because I think that's too big of a change. I also don't know where beta adoption will take us, and how building delivery configs that can be for multiple applications will work.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -379,6 +379,39 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
...
+                  kind = TEST_API_V1.qualify("whatever"),
+                  spec = DummyResourceSpec(data = "o hai")
+                )
+              ),
+              constraints = emptySet()
+            )
+          )
+        )
+
+        val submittedConfig2 = submittedConfig1.copy(name = "double-trouble")
```
luispollo: Love the name. 😄

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -379,6 +379,39 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
...
+        test("an error is thrown and config is deleted") {
```
luispollo: Technically I think this might be more accurate:
```suggestion
        test("an error is thrown and config is not persisted") {
```

PR #946, opened 2020-03-30
feat(tests): Add code coverage report

luispollo:
This PR adds the JaCoCo plugin and sets up an aggregate report of code coverage for all sub-projects, that we can use, among other things, to publish a code coverage badge on the `README`.


emjburns:
Why are you adding this?

luispollo:
> Why are you adding this?

So we can see what kind of code coverage we have in the project?

emjburns:
I guess my question is more - do you think that adding code coverage is valuable? I personally don't really think it's valuable, but I haven't used it that much. What scenarios do you imagine it helping with?

luispollo:
> I guess my question is more - do you think that adding code coverage is valuable? I personally don't really think it's valuable, but I haven't used it that much. What scenarios do you imagine it helping with?

I find it valuable not so much from the detailed report, but because it gives us the ability to generate a nice badge like this (actual current coverage):
Codecov branch

I've always found that giving visibility to this kind of metric naturally makes developers on the team more aware of where we are and nudges them to improve. No one likes to see a red label. 😄 

As a secondary benefit, even though we're fortunate to have nice built-in reports in IntelliJ Ultimate, not everyone in the open-source community will, and so using an open-source code coverage tool allows them access to the detailed information as well. I have yet to figure out how to publish the reports as part of the GitHub builds.

emjburns:
I find that a badge like this incentives perhaps the wrong thing. Does test coverage give you an indication of how well things are working? I don't think so ... I think the best indication is whether someone is using it in production and whether active development is happing. Sure, test coverage helps things run smoothly, but a culture of good tests is I think better built by all buying in and enforcing it with code review that putting a metric up that shames everyone. 

Just my 2 cents.

luispollo:
> I find that a badge like this incentives perhaps the wrong thing. Does test coverage give you an indication of how well things are working? I don't think so ... I think the best indication is whether someone is using it in production and whether active development is happing. 
Personally, I don't think the two are mutually exclusive. 

> Sure, test coverage helps things run smoothly, but a culture of good tests is I think better built by all buying in and enforcing it with code review that putting a metric up that shames everyone.

That's a pretty strong word. The intention is never to point blame or publicly shame anyone. In fact, it's an entirely anonymous/opaque number. The goal, as I said before, is to give visibility. It's only one amongst a variety of metrics we can use to make keel better. Again, it doesn't have to be one thing to the exclusion of another. BTW, 80% is a pretty good number for a project where we haven't been tracking it. If anything, it should be a cause for pride, not shame.

I'm not hung up on pushing this. If others think it's a bad idea, I'll just close the PR.

emjburns:
If it's a useful tool I'm okay with you adding it, I just probably will never look at it. I guess I'm more worried about the ways I feel it could be misused (like, "we really gotta get our code coverage up, let's focus on that"). But, we can always talk about that if it happens!

PR #947, opened 2020-03-31
fix(configs): get by app returns one delivery config

emjburns:
Delivery config changes to follow enforcing one config per app (https://github.com/spinnaker/keel/pull/944). This will make it easier to work with delivery configs.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
@@ -21,10 +21,8 @@ import strikt.api.expectCatching
 import strikt.api.expectThat
 import strikt.assertions.contains
 import strikt.assertions.failed
-import strikt.assertions.first
 import strikt.assertions.hasSize
 import strikt.assertions.isA
-import strikt.assertions.isEmpty
 import strikt.assertions.isEqualTo
 import strikt.assertions.succeeded
 
@@ -141,8 +139,8 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
 
       test("retrieving config by application returns an empty list") {
         getByApplication()
-          .succeeded()
-          .isEmpty()
+          .failed()
+          .isA<NoSuchDeliveryConfigException>()
       }
     }
 
@@ -163,8 +161,6 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
       test("the config can be retrieved by application") {
         getByApplication()
           .succeeded()
-          .hasSize(1)
-          .first()
           .get(DeliveryConfig::name)
           .isEqualTo(deliveryConfig.name)
       }
@@ -229,7 +225,6 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
         test("artifacts are attached when retrieved by application") {
           getByApplication()
             .succeeded()
-            .first()
             .get { artifacts }.isEqualTo(deliveryConfig.artifacts)
         }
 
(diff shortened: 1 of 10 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -243,7 +247,7 @@ class CombinedRepository(
   override fun deliveryConfigFor(resourceId: String): DeliveryConfig =
     deliveryConfigRepository.deliveryConfigFor(resourceId)
 
-  override fun getDeliveryConfigsByApplication(application: String): Collection<DeliveryConfig> =
+  override fun getDeliveryConfigForApplication(application: String): DeliveryConfig =
```
emjburns: This is the big change.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -141,8 +139,8 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
 
       test("retrieving config by application returns an empty list") {
```
luispollo:
```suggestion
      test("retrieving config by application fails") {
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
```diff
@@ -146,11 +145,12 @@ interface DeliveryConfigRepository : PeriodicallyCheckedRepository<DeliveryConfi
 
 sealed class NoSuchDeliveryConfigException(message: String) : UserException(message)
 class NoSuchDeliveryConfigName(name: String) : NoSuchDeliveryConfigException("No delivery config named $name exists in the repository")
+class NoDeliveryConfigForApplication(application: String) : NoSuchDeliveryConfigException("No delivery config for application $application exists in the repository")
```
luispollo:
```suggestion
class NoDeliveryConfigForApplication(application: String) : NoSuchDeliveryConfigException("No delivery config for application $application exists in the database")
```
(I don't think users will know what a "repository" means in this context if the exception ever bubbles up.)

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
gal-yardeni: why should it fail? maybe an empty list is better user experience?

gal-yardeni: ok I just saw that it supposed to fail and return an exception, so nevermind!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -68,10 +69,10 @@ class SqlDeliveryConfigRepository(
         )
         .from(DELIVERY_CONFIG)
         .where(DELIVERY_CONFIG.APPLICATION.eq(application))
-        .fetch { (uid, name, application, serviceAccount) ->
+        .fetchOne { (uid, name, application, serviceAccount) ->
```
gal-yardeni: why it was changed from `fetch` to `fetchOne`? Is it possible to have more than one?

emjburns: It's not - I removed the ability to have more than one. If there is it'll be a database failure. So This is just updating this method call to reflect that.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
emjburns: I'll just remove "in the repsotiory"

PR #948, opened 2020-03-31
fix(artifact): more resilient app version parsing

emjburns:
This PR name might lead you to think that this is less hacky than before. It's not. It's just more wacky-input-safe.

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -154,14 +154,25 @@ class ApplicationService(
   ): ArtifactVersionSummary =
     when (artifact.type) {
       deb -> {
-        val appversion = AppVersion.parseName(version)
-        ArtifactVersionSummary(
+        var summary = ArtifactVersionSummary(
           version = version,
           environments = environments,
-          displayName = appversion?.version ?: version.removePrefix("${artifact.name}-"),
-          build = if (appversion != null) BuildMetadata(id = appversion.buildNumber.toInt()) else null,
-          git = if (appversion != null) GitMetadata(commit = appversion.commit) else null
+          displayName = version.removePrefix("${artifact.name}-")
         )
+
+        // attempt to parse helpful info from the appversion.
+        // todo: replace, this is brittle
+        val appversion = AppVersion.parseName(version)
+        if (appversion?.version != null) {
+          summary = summary.copy(displayName = appversion.version)
+        }
+        if (appversion?.buildNumber != null) {
+          summary = summary.copy(build = BuildMetadata(id = appversion.buildNumber.toInt()))
+        }
+        if (appversion?.commit != null) {
+          summary = summary.copy(git = GitMetadata(commit = appversion.commit))
+        }
+        summary
       }
       docker -> {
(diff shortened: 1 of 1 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -154,14 +154,25 @@ class ApplicationService(
...
+        if (appversion?.version != null) {
+          summary = summary.copy(displayName = appversion.version)
+        }
+        if (appversion?.buildNumber != null) {
+          summary = summary.copy(build = BuildMetadata(id = appversion.buildNumber.toInt()))
+        }
+        if (appversion?.commit != null) {
+          summary = summary.copy(git = GitMetadata(commit = appversion.commit))
+        }
+        summary
```
luispollo: I'm failing to see how this is different from what you had before, other than setting `displayName` outside this block above. I find this format harder to read than the one with the property names and inline if statements to the left. It's also potentially copying the whole object 3 times over which I know is not a huge deal but I feel if we just let these spread they may become a problem.

lorin: I believe the current problem is that `appversion.buildNumber` is null for one of the artifacts, so the `appversion.buildNumber.toInt()` call fails.on line 162. In the new implementation, this call is guarded by a null chcek in line 169.

lorin:
I'd love to see a unit test against the problematic artifact version to verify that the code no longer throws an exception on it.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -154,14 +154,25 @@ class ApplicationService(
...
-          build = if (appversion != null) BuildMetadata(id = appversion.buildNumber.toInt()) else null,
-          git = if (appversion != null) GitMetadata(commit = appversion.commit) else null
+          displayName = version.removePrefix("${artifact.name}-")
         )
+
+        // attempt to parse helpful info from the appversion.
+        // todo: replace, this is brittle
+        val appversion = AppVersion.parseName(version)
+        if (appversion?.version != null) {
+          summary = summary.copy(displayName = appversion.version)
```
lorin: Do we ever expect  fields of `appversion` to be null during normal operation? IF not, I'd put a `WARN` level log statement with the  value of`version` that's parsed to make it easier to debug in the future.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: Lorin is correct. Again, this code should be replaced in Q2.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: we might fail, we might not fail if that's the case. it definitely puts us into a weird state. once we pull the appversion off the ami though it shouldn't be null. we don't rely on any of the other fields that i'm using here (if i remember correctly). Logging would probably be helpful, you're right.

PR #949, opened 2020-03-31
feat(api): Add stateless constraint information to artifact summary

### LINKED ISSUE #889: Surface stateless constraints for the UI via ArtifactSummaryInEnvironment
Similar to https://github.com/spinnaker/keel/issues/888, in order to show users information about stateless constraints in the UI we'll need to surface them via artifact summaries in the API. It seems to make the most sense for them to live within [the `ArtifactSummaryInEnvironment` class](https://github.com/spinnaker/keel/blob/368312c2962076c06c44465fe09937888f879707/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt#L25-L34).

Here's a rough sketch of what the JSON might look like:
```ts
{
  name: "test",
  state: "current",
  deployedAt: "2019-12-04T18:49:58.748Z",
  replacedAt: "2019-12-04T18:49:58.748Z",
  replacedBy: "keel-0.344.0-h280.966f877",
  statelessConstraints: [
    {
      type: "depends-on",
      currentlyPassing: true,
      attributes: {
        // arbitrary key/value metadata, like time windows for allowed-times
        // or the dependent environment for depends-on. Strongly typed by
        // constraint type so we can get type safety on the client
      }
    }
  ]
},
```

Given the way things are currently modeled, it's seeming like we'll want to differentiate between stateless and stateful constraints at the API level for a couple reasons:
- Stateless constraints don't really have a 'pending' or 'failed' condition the way stateful constraints do. They're either passing right now, or they're not. That might be bad, it might be good, but it could change the next time we check. It's worth being able to differentiate that (e.g. show grey/green visuals for stateless, and grey/blue/green/red for stateful depending on status)
- There's an important and useful contract we give constraints: stateful constraints only kick off once all the stateless constraints have been satisfied. This prevents things like kicking off a canary too soon, or prompting for manual judgement before a version has come up healthy in the previous environment. Keeping them separate for now allows us to visually order constraints in a way that matches their behavior

gal-yardeni:
Closes #889 
Adding stateless constraints info the artifact summary API.
an example of the output would be:
```json
              "statelessConstraints": [
                {
                  "type": "depends-on",
                  "currentlyPassing": true,
                  "attributes": {
                    "environment": "testing"
                  }
                },
                {
                  "type": "allowed-times",
                  "currentlyPassing": false,
                  "attributes": {
                    "windows": [
                      {
                        "days": "Wed",
                        "hours": "12-14"
                      }
                    ],
                    "tz": "America/Los_Angeles"
                  }
                }
              ]
```

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -75,8 +75,6 @@ data class StatefulConstraintSummary(
 data class StatelessConstraintSummary(
   val type: String,
   val currentlyPassing: Boolean,
-  val status: ConstraintStatus? = null,
-  val comment: String? = null,
   val attributes: ConstraintMetadata? = null
 )
 
@@ -88,5 +86,5 @@ data class DependOnConstraintMetadata(
 
 data class AllowedTimesConstraintMetadata(
   val windows: List<TimeWindow>,
-  val tz: String? = null
+  val timezone: String? = null
 ) : ConstraintMetadata()
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -56,7 +56,8 @@ data class ArtifactSummaryInEnvironment(
   val deployedAt: Instant? = null,
   val replacedAt: Instant? = null,
   val replacedBy: String? = null,
-  val statefulConstraints: List<StatefulConstraintSummary> = emptyList()
+  val statefulConstraints: List<StatefulConstraintSummary> = emptyList(),
+  val statelessConstraints: List<StatelessConstraintSummary?> = emptyList()
 )
 
 @JsonInclude(Include.NON_NULL)
@@ -69,3 +70,23 @@ data class StatefulConstraintSummary(
   val comment: String? = null,
   val attributes: ConstraintStateAttributes? = null
 )
+
+@JsonInclude(Include.NON_NULL)
+data class StatelessConstraintSummary(
+  val type: String,
+  val currentlyPassing: Boolean,
+  val status: ConstraintStatus? = null,
+  val comment: String? = null,
+  val attributes: ConstraintMetadata? = null
+)
+
+abstract class ConstraintMetadata(val type: String)
+
+data class DependOnConstraintMetadata(
+  val environment: String
+) : ConstraintMetadata("depends-on")
+
+data class AllowedTimesConstraintMetadata(
+  val windows: List<TimeWindow>,
+  val tz: String? = null
+) : ConstraintMetadata("allowed-times")
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -80,13 +80,13 @@ data class StatelessConstraintSummary(
   val attributes: ConstraintMetadata? = null
 )
 
-abstract class ConstraintMetadata(val type: String)
+abstract class ConstraintMetadata()
 
 data class DependOnConstraintMetadata(
   val environment: String
-) : ConstraintMetadata("depends-on")
+) : ConstraintMetadata()
 
 data class AllowedTimesConstraintMetadata(
   val windows: List<TimeWindow>,
   val tz: String? = null
-) : ConstraintMetadata("allowed-times")
+) : ConstraintMetadata()
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -38,14 +38,14 @@ import org.springframework.stereotype.Component
 @Component
 class ApplicationService(
   private val repository: KeelRepository,
-  private val constraints: List<ConstraintEvaluator<*>>
+  private val constraintEvaluators: List<ConstraintEvaluator<*>>
 ) {
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
-  private val explicitConstraints: List<ConstraintEvaluator<*>> = constraints.filter { !it.isImplicit() }
-  private val statefulEvaluators: List<ConstraintEvaluator<*>> = explicitConstraints
+  private val explicitConstraintEvaluators: List<ConstraintEvaluator<*>> = constraintEvaluators.filter { !it.isImplicit() }
+  private val statefulEvaluators: List<ConstraintEvaluator<*>> = explicitConstraintEvaluators
     .filterIsInstance<StatefulConstraintEvaluator<*>>()
 
-  private val statelessEvaluators = explicitConstraints - statefulEvaluators
+  private val statelessEvaluators = explicitConstraintEvaluators - statefulEvaluators
 
   fun hasManagedResources(application: String) = repository.hasManagedResources(application)
 
@@ -177,8 +177,8 @@ class ApplicationService(
           type = constraint.type,
           currentlyPassing = it.canPromote(artifact, version = version, deliveryConfig = deliveryConfig, targetEnvironment = environment),
           attributes = when (constraint) {
-              is DependsOnConstraint -> DependOnConstraintMetadata(constraint.environment)
-              is TimeWindowConstraint -> AllowedTimesConstraintMetadata(constraint.windows, constraint.tz)
+            is DependsOnConstraint -> DependOnConstraintMetadata(constraint.environment)
+            is TimeWindowConstraint -> AllowedTimesConstraintMetadata(constraint.windows, constraint.tz)
             else -> null
           }
         )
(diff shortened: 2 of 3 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -30,9 +33,15 @@ import org.springframework.stereotype.Component
  */
 @Component
 class ApplicationService(
-  private val repository: KeelRepository
+  private val repository: KeelRepository,
+  private val constraints: List<ConstraintEvaluator<*>>
 ) {
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
+  private val explicitConstraints: List<ConstraintEvaluator<*>> = constraints.filter { !it.isImplicit() }
```
gal-yardeni: the part of filtering out the right evaluators is pretty much taken from `EnvironmentPromotionChecker` class. Any suggestions on how to make it prettier?

luispollo: Looks readable enough to me. I didn't even know `isImplicit()` was a thing (probably new?). 😃

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -30,9 +37,15 @@ import org.springframework.stereotype.Component
  */
 @Component
 class ApplicationService(
-  private val repository: KeelRepository
+  private val repository: KeelRepository,
+  private val constraints: List<ConstraintEvaluator<*>>
```
luispollo: For naming consistency:
```suggestion
  private val constraintEvaluators: List<ConstraintEvaluator<*>>
```

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -30,9 +37,15 @@ import org.springframework.stereotype.Component
...
 ) {
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
+  private val explicitConstraints: List<ConstraintEvaluator<*>> = constraints.filter { !it.isImplicit() }
```
luispollo: For naming consistency:
```suggestion
  private val explicitConstraintEvaluators: List<ConstraintEvaluator<*>> = constraints.filter { !it.isImplicit() }
```

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -97,10 +110,12 @@ class ApplicationService(
               )
             }?.let { artifactSummaryInEnvironment ->
               addStatefulConstraintSummaries(artifactSummaryInEnvironment, deliveryConfig, environment, version)
-              // TODO: add stateless constraint summaries
-            }?.also { artifactSummaryInEnvironment ->
-              artifactSummariesInEnvironments.add(artifactSummaryInEnvironment)
             }
+              ?.let { artifactSummaryInEnvironment ->
```
luispollo: You don't need the extra `.let` here -- you can just call `addStatelessConstraintSummaries` right below `addStatefulConstraintSummaries` above, where the `//TODO` was.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -142,6 +157,39 @@ class ApplicationService(
...
+   * Adds details about any stateless constraints in the given environment to the [ArtifactSummaryInEnvironment].
+   */
+  private fun addStatelessConstraintSummaries(
+    artifactSummaryInEnvironment: ArtifactSummaryInEnvironment,
+    deliveryConfig: DeliveryConfig,
+    environment: Environment,
+    version: String,
+    artifact: DeliveryArtifact
+  ): ArtifactSummaryInEnvironment {
+    val statelessConstraints: List<StatelessConstraintSummary?> = environment.constraints.filter { constraint ->
```
luispollo: Why the `?` for the type of the list items? There shouldn't be any nulls in the list, right?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -142,6 +157,39 @@ class ApplicationService(
...
+      constraint !is StatefulConstraint
+    }.map { constraint ->
```
luispollo: You can use `.mapNotNull` here, then you won't have any nulls in your list above.

luispollo:
@gal-yardeni Just a heads-up that @emjburns was planning to replace `ApplicationControllerTests` with a new `ApplicationServiceTests` class in #942 since it's simpler to work with tests if you don't need to have a full Spring context loaded, MVC mocked, etc. Depending on which of you merges first, you'll have some conflicts to resolve.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: Yep it's pretty new, Emily added it in one of her previous prs.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: actually I've tried it! but when doing so, the StatefulConstraint got disappeared!

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: the `find` can still return null, so the compiler is asking me to play it safe

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: Tried changing it and it still didn't help with reducing the `null` checks..

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
```diff
@@ -69,3 +70,23 @@ data class StatefulConstraintSummary(
...
+)
+
+abstract class ConstraintMetadata()
+
+data class DependOnConstraintMetadata(
+  val environment: String
+) : ConstraintMetadata()
+
+data class AllowedTimesConstraintMetadata(
+  val windows: List<TimeWindow>,
```
erikmunson: does this preserve whatever string format existed on the persisted delivery config itself? i.e. if they put in `2-20` as the `hours` string will it be represented exactly like that?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
```diff
@@ -69,3 +70,23 @@ data class StatefulConstraintSummary(
   val comment: String? = null,
   val attributes: ConstraintStateAttributes? = null
 )
+
+@JsonInclude(Include.NON_NULL)
+data class StatelessConstraintSummary(
+  val type: String,
+  val currentlyPassing: Boolean,
+  val status: ConstraintStatus? = null,
+  val comment: String? = null,
```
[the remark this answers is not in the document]
gal-yardeni: I left it here if we decide to add it in the future, but I can take it off :)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
erikmunson: nevermind — just noticed your updated PR description has an example with a dashedd.

I'm not sure whether it's appropriate to do it in this PR or not, but for UI purposes it'd be better to split this up into a more structured object, e.g.:
```js
{
  "days": "Wed",
  "hours": {
    from: "12"
    to: "14"
  }
}
```

gal-yardeni: yes, it does.. do you prefer it to be shown differently?

emjburns:
@luispollo I'll be resolving the conflicts :) it's my lucky day!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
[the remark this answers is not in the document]
emjburns: Correct - `comment` and `status` are only for stateful

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
erikmunson: I'm gonna open a GH issue for this as I don't think it's critical to getting things working — but I would like to make the attributes data for these much more structured before we fully move on from working on constraint APIs so it doesn't stay this way forever :D

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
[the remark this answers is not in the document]
gal-yardeni: removed!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
gal-yardeni: Yeah, I agree with you - I prefer it more structured as well. I can make the change in a separate pr.

erikmunson: 👍 thanks gal!

PR #953, opened 2020-03-31
fix(api): Handle non-managed applications gracefully in summaries API

luispollo:
Tiny PR to handle the case of applications not known/managed by keel gracefully when the application summaries API is called.

(Also simplified the stateless constraint code a bit.)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -58,12 +58,14 @@ class ApplicationController(
       if (includeDetails && !entities.contains("resources")) {
         entities.add("resources")
       }
-      entities.forEach { entity ->
-        results[entity] = when (entity) {
-          "resources" -> applicationService.getResourceSummariesFor(application)
-          "environments" -> applicationService.getEnvironmentSummariesFor(application)
-          "artifacts" -> applicationService.getArtifactSummariesFor(application)
-          else -> throw InvalidRequestException("Unknown entity type: $entity")
+      if (true == results["hasManagedResources"]) {
```
gal-yardeni: You can remove the `true` here :)

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -38,14 +38,11 @@ import org.springframework.stereotype.Component
 @Component
 class ApplicationService(
   private val repository: KeelRepository,
-  private val constraintEvaluators: List<ConstraintEvaluator<*>>
+  constraintEvaluators: List<ConstraintEvaluator<*>>
```
gal-yardeni: just out of curiosity, why was `private val` removed?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -38,14 +38,11 @@ import org.springframework.stereotype.Component
...
 ) {
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
-  private val explicitConstraintEvaluators: List<ConstraintEvaluator<*>> = constraintEvaluators.filter { !it.isImplicit() }
-  private val statefulEvaluators: List<ConstraintEvaluator<*>> = explicitConstraintEvaluators
-    .filterIsInstance<StatefulConstraintEvaluator<*>>()
-
-  private val statelessEvaluators = explicitConstraintEvaluators - statefulEvaluators
+  private val statelessEvaluators: List<ConstraintEvaluator<*>> =
+    constraintEvaluators.filter { !it.isImplicit() && it !is StatefulConstraintEvaluator }
```
gal-yardeni: Nice!

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
luispollo: I can't because the map can return `Any?`.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: IntelliJ told me this wasn't accessed as a property anywhere (and, in fact, it's only used to build `statelessContraintEvaluators` below).

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
gal-yardeni: oh I see now, thanks

PR #955, opened 2020-04-01
feat(security): API permission checks

### LINKED ISSUE #816: Check service account permissions in relevant controller methods

luispollo:
Upon discussion with the team, we agreed on the following general guidelines for API authorization in keel:
   - Operations that read data from keel’s database exclusively should require READ access to the application.
   - Operations that read data from keel’s database and from cloud infrastructure should require READ access to the application AND, for each resource returned, READ access to the corresponding cloud account.
 - Operations that trigger storing data in keel’s database, but do NOT trigger infrastructure changes, should require WRITE access to the application.
 - Operations that trigger actuation, whereby the service account will take action on behalf of the original caller, should check the caller has access to the service account, in addition to WRITE access to the application.

This PR covers the following:
- Adds `@PreAuthorize` annotations to almost all API endpoints (excluding those that are called strictly internally from Spinnaker services such as `/artifacts/events`)
- Revisits the implementation of `AuthorizationSupport` by defining authorization functions that align with the guidelines above
- Adds tests for all authorized controller functions

Closes #816


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/SubmittedResource.kt
```diff
@@ -36,10 +36,10 @@ data class SubmittedResource<T : ResourceSpec>(
   @Description("The specification of the resource")
   @JsonTypeInfo(use = Id.NAME, include = As.EXTERNAL_PROPERTY, property = "kind")
   val spec: T
-)
-
-val <T : ResourceSpec> SubmittedResource<T>.id: String
-  get() = generateId(kind, spec)
+) {
+  val id: String
```
luispollo, line 40: This avoids the need to import the `id` extension all over the place, especially since it was already defined within the class itself and didn't have any external dependencies.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/PersistenceExceptions.kt
```diff
@@ -0,0 +1,54 @@
+package com.netflix.spinnaker.keel.exceptions
+
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import com.netflix.spinnaker.kork.exceptions.SystemException
+import com.netflix.spinnaker.kork.exceptions.UserException
+
+sealed class NoSuchEntityException(override val message: String?) :
```
luispollo, line 8: I've consolidated all exceptions from the various repositories in this file so that it's easy to follow the hierarchy, and have refactored the hierarchy so that all "not found" type exceptions inherit from `NoSuchEntityException`. This allows me to handle the "not found" case in a generic way in `AuthorizationSupport` so that we can allow 404s to be returned from the APIs instead of 403s. To be honest, I'm not sure that's the right thing to do -- if you don't have permission, in theory we shouldn't even query the database at all. Thoughts?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -49,6 +49,7 @@ class ApplicationController(
     path = ["/{application}"],
     produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
   )
+  @PreAuthorize("@authorizationSupport.userCan('READ', 'APPLICATION', #application)")
```
luispollo, line 52: Example of what calling the new `userCan` function looks like.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -151,6 +157,9 @@ class ArtifactController(
     @PathVariable name: String,
     @PathVariable type: ArtifactType
   ): List<String> =
+    // FIXME: we can't authorize this call at the moment because neither the application name nor the delivery config
+    //  name are passed into this method. This means this method would also currently return artifact versions across
+    //  applications if the artifact names and types match, which I think is a bug.
```
emjburns, line 162: I don't think we need permissions on this call. Artifact versions are stored independently of applications, which is intentional and needed to prevent a potentially huge amount of data duplication. The available versions of an artifact is not restricted data.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
```diff
@@ -18,59 +18,73 @@
...
+          log.debug("[ACCESS {}] User {}: access to service account {}.",
+            if (it) "ALLOWED" else "DENIED", auth.principal, serviceAccount)
+        }
+      val hasApplicationAccess = permissionEvaluator.hasPermission(serviceAccount, application, "APPLICATION", action.name)
+        .also {
+          log.debug("[ACCESS {}] Service account {}: permission to {} {} in application {}.",
+            if (it) "ALLOWED" else "DENIED", serviceAccount, action.name, "${entity.name} $identifier", application)
+        }
+      hasServiceAccountAccess && hasApplicationAccess
+    } catch (e: NoSuchEntityException) {
```
luispollo, line 81: This is where `NoSuchEntityException` as the ancestor of all "not found"-type repository exceptions comes into play.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
```diff
@@ -18,59 +18,73 @@
...
-      serviceAccount,
-      if (hasPermission) "" else " DO NOT",
-      specOrName
-    )
-    return hasPermission
+  fun userCan(action: Action, entity: Entity, identifier: String): Boolean {
+    return try {
+      val auth = SecurityContextHolder.getContext().authentication
+      val serviceAccount = when (entity) {
+        Entity.RESOURCE -> repository.getResource(identifier).serviceAccount
```
emjburns, line 61: nit: please add imports for all these so that this is `RESOURCE` instead of `Entity.Resource`

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/InteractiveNotificationCallbackController.kt
```diff
@@ -18,29 +21,31 @@ import org.springframework.web.bind.annotation.RestController
...
-  //  the e-mail address without a Spinnaker user having authenticated here.
+  //  the e-mail address without a user having actually authenticated with Spinnaker.
   fun handleInteractionCallback(
     @RequestHeader("X-SPINNAKER-USER") user: String,
     @RequestBody callback: EchoNotification.InteractiveActionCallback
   ) {
     val currentState = repository.getConstraintStateById(parseUID(callback.messageId))
       ?: throw InvalidConstraintException("constraint@callbackId=${callback.messageId}", "constraint not found")
 
+    if (!authorizationSupport.userCan(Action.WRITE, Entity.DELIVERY_CONFIG, currentState.deliveryConfigName)) {
```
luispollo, line 45: Unfortunately, for this API, since it must be kept generic to comply with the contract for interactive notifications, I have to look up the constraint state from the database to then check permissions. But, it's an example showing you can use `AuthorizationSupport` directly instead of the annotations, if needed.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -64,6 +67,7 @@ class DeliveryConfigController(
     consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
     produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
   )
+  @PreAuthorize("@authorizationSupport.userCan('WRITE', 'DELIVERY_CONFIG', #deliveryConfig.name)")
```
emjburns, line 70: this is just a get call, although it does do calculations on the back end. Why did you chose 'WRITE' permissions here?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
```diff
@@ -0,0 +1,50 @@
...
+  apis: Map<ApiRequest, AuthorizationSupport.Permission>
+) {
+  apis.forEach { (api, permission) ->
+    val (method, uri) = api.request.split(Regex("""\s+"""))
+    context("API permission check for $api") {
+      before {
+        every {
+          authorizationSupport.userCan(permission.action.name, permission.entity.name, any())
+        } returns false
+      }
```
luispollo, line 27: @robfletcher This is what I ended up with that worked, but was not what I originally intended. What I would like to do is to have a single context like "API permission checks" for the entire controller, with multiple tests underneath, one per API check. However, what I found is that trying to override the mock behavior in the `before` block inside each test does not work.

If you can think of something I'm missing in setting this up, please let me know.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
```diff
@@ -55,9 +58,14 @@ internal class ResourceControllerTests {
...
 
+  @BeforeEach
+  fun mockAuthz() {
+    every { authorizationSupport.userCan(Action.READ.name, Entity.RESOURCE.name, any()) } returns true
+  }
+
   @Test
   fun `can't diff a resource when unauthorized`() {
-    every { authorizationSupport.userCanModifySpec("fzlem@netflix.com", any()) } returns false
+    every { authorizationSupport.userCan(Action.READ.name, Entity.RESOURCE.name, any()) } returns false
```
luispollo, line 68: This is more or less what I was trying to accomplish in the minutest-version of controller tests. It works fine here just running with JUnit... 🤷‍♂

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put
+import org.springframework.test.web.servlet.result.MockMvcResultMatchers.status
+
+fun ContextBuilder<*>.testApiPermissions(
+  mvc: MockMvc,
+  jsonMapper: ObjectMapper,
+  authorizationSupport: AuthorizationSupport,
+  apis: Map<ApiRequest, AuthorizationSupport.Permission>
+) {
+  apis.forEach { (api, permission) ->
```
emjburns, line 20: I find this test very hard to understand, and I would be pretty lost with how to update it. Could you try to make it easier to understand, and add comments?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/SubmittedResource.kt
robfletcher: Are you sure this isn't going to require users to specify the id on submission?

Also you could use `val id: String by lazy { generateId(kind,  spec) }` since you don't really need to re-generate the id every time someone calls `mySubmittedResource.id`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/PersistenceExceptions.kt
emjburns: If the resource in question doesn't exist - a delivery config, and application, etc - we should return a `..NotFound` instead of failing because you "aren't authorized". If the thing does exist enough to check permissions and you're not authorized, we should return unauthorized

robfletcher: I'm generally not a fan of packaging things according to what type of thing they are (exceptions in this case) rather than what bit of functionality they're involved in. Is there any reason to not co-locate these with the repository interfaces themselves?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
luispollo: OK, I'll remove the `FIXME`. What about the question on returning versions across applications? Isn't the UI always going to have to index things based on application anyway? Seems like a waste of bandwidth to return the artifact versions for all apps if the artifact names happen to match (I have no idea how common that is).

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
robfletcher: I don't understand. There's only one test in here, and I don't see anything overriding what this mock does.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/PersistenceExceptions.kt
luispollo: Okey dokey! So this is as should be. 😃

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
cfieber: why is there a new way to do this here when everywhere else in Spinnaker this would be

` @PreAuthorize("hasPermission(#application, 'APPLICATION', 'READ')")`

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
luispollo: Copy/paste mistake. Will fix.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
robfletcher: You know you can nest `context` blocks, right? You could have a structure like:

```kotlin
context("API auth checks") {
  context("user is not authorized") {
    before {
      // set up mock to return false
    }
    test("user can't do the thing") {
      // ...
    }
  }
  context("user is totally authorized") {
    before {
      // set up mock to return true
    }
    test("user is allowed to do the thing") {
      // ...
    }
  }
}
```

Am I missing something?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
luispollo: Sure. The basic idea is that I want to avoid a ton of repetition defining these API permission check tests in the various controller tests. This function just loops through the APIs defined by the controller tests and dynamically adds a context/test for each, so you don't have to do it yourself.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
emjburns: What do you mean? This is an api that pulls the artifact versions for a specific name and type of artifact. Artifacts are not scoped to an application.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
cfieber: I just looked in authorization support, so I get what is going on in there now and it makes a bit more sense, I'm still not sure that for some of these checks they shouldn't just follow the standard model, looking a bit more

emjburns: This was originally done this way because we didn't have all the info in the API call that we needed to look up the permissions. I had to do some additional querying. I bet we could follow the standard model for some of them/maybe we could have done it all along

cfieber: I think it is worth a discussion on ApplicationController what is required to view/pause/resume management (I'm not totally sure it is the same as having access to manage the resource configuration of the application / needs to be gated on access to the service account)

cfieber: for example if I can see your application but can't modify it, I should maybe be able to tell it is a managed application?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/SubmittedResource.kt
luispollo: Ah. Good question. I'll just revert that. Not worth the risk.

cfieber:
This seems like a reasonable technical implementation for checks that need to take into consideration the ability of someone able to customize declarative management (the `userCan` method)

I feel like this method has been too broadly applied in the context of UX for an application under declarative management.. I'd hesitate to recommend this goes in without careful consideration of each of the personas that might interact with the application and what they should be able to see or do.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/PersistenceExceptions.kt
luispollo: @robfletcher Not technically. I just thought it would be very easy to forget to extend `NoSuchEntityException` if it's not near the other exception types. But I agree with you on the principle for code organization in general. Let me see how to better organize them.

P.S. In my defense, we already had a `com.netflix.spinnaker.keel.exceptions` package elsewhere.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
luispollo: @cfieber I had the same reaction when initially working on this, and that was one of the reasons I refactored it to be as close as possible to what the other services already use. But as you pointed out, there's some additional logic to find out the service account, and after my changes also the application name, so we can actually call the permission evaluator. If you can think of an easier/better way to do this, I'm all 👂s.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
luispollo: Oh, that's right, I totally forgot about that! I'm even using one of your container images in one of my delivery configs. 😄 I guess my confusion came from the fact that our top-level concept is the delivery config, and _it_ is scoped to an application. But the artifacts defined within it are not.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
cfieber: I think if I understand correctly that everything funnelling through userCan is a test on whether the user has permission to the service account for the entity and that the service account has the permission being tested.

I think that only accounts for the persona of an application owner that can manage the delivery configuration for the application (they have permission to the service account).

I suspect (in application controller at least) that access to details on whether the application is under declarative management would be relevant to anyone who has permission to view the application.

I'd just propose going through each of the places where auth has been applied and considering the security annotation under the lens of what specifically that endpoint is serving and make sure we haven't locked other users out

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
luispollo: The original code was not generalized in this function, it was just in `ApplicationControllerTests`, and looked something like this:
```kotlin
    fixture { Fixture() }

    before {
      every { authorizationSupport.userCan(Action.READ.name, Entity.APPLICATION.name, any()) } returns true      
    }

    after {
      clearAllMocks()
    }

    context("other controller tests") {
      // ...
    }

    context("when user is not authorized") {
      test("Some GET API is forbidden") {
        before {
          every { authorizationSupport.userCan(Action.READ.name, Entity.APPLICATION.name, any()) } returns true      
        }
        // mvc.perform...
      }

      test("Some POST API is forbidden") {
        before {
          every { authorizationSupport.userCan(Action.WRITE.name, Entity.APPLICATION.name, any()) } returns true      
        }
        // mvc.perform...
      }
    }
  }
```

This did not work.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
luispollo: No, and I've tried all sorts of combinations like that. I suspect the issue is when you define a `before` block directly under `tests` (i.e. scoped to all `context`s and `test`s underneath).

robfletcher: But why would you initialize the mock to return either true or false there?

luispollo:
> I feel like this method has been too broadly applied in the context of UX for an application under declarative management.. I'd hesitate to recommend this goes in without careful consideration of each of the personas that might interact with the application and what they should be able to see or do.

Thanks for the review, @cfieber! Can you elaborate a bit on what you expect those other personas/use cases might be? As an example, do you mean as in our team operating Spinnaker and having to make API calls? My thought process there was that all our API calls are authenticated one way or another, so we have a user identity, and my understanding was that generally our team has access to all the accounts. What I failed to grasp there, I think, is that we _don't_ have access to all service accounts, so that would likely cause calls with our personal identities to fail (even though, as you know, we have the ability to work around that internally). Is that the kind of scenario you were thinking about, or something else entirely?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
robfletcher: Why not wrap the cases where the user _is_ authorized in their own context?

Also you *cannot* put a `before` inside a `test` like that. I don't understand what the point is anyway. The point of grouping things into contexts is so they can share common setup. Why does each test in that context need its own special `before` block? Just generalize the mock initialization. Either:

```kotlin
every { authorizationSupport.userCan(Action.READ.name, Entity.APPLICATION.name, any()) } returns true 
every { authorizationSupport.userCan(Action.WRITE.name, Entity.APPLICATION.name, any()) } returns true
```

or…

```kotlin
every { authorizationSupport.userCan(any(), Entity.APPLICATION.name, any()) } returns true 
```

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
luispollo: Added comments. Hopefully they clarify.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
luispollo: Well, I had 4 contexts in the class, and in only 1 of each I wanted the behavior to be different. Is the common setup to repeat the same code in all 3 contexts where the behavior is the same? That's what was a bit surprising to me.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
luispollo: > Also you **cannot** put a `before` inside a `test` like that.

Ohhh! There you go! So that's what I was missing. This feels like a bug that the DSL allows you to even do that.

Thanks for the example. In my mind, besides sharing common setup, another reason to use contexts was simply to logically organize your tests in a hierarchical way. But point taken! Will revisit.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt
robfletcher: so have

context 1
  context 1a
  context 1b
  context 1c
context 2

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
robfletcher: I'm not sure how it could prevent you doing that. It's kind of a limitation of the fact that even when a closure overrides `this` the `this` from any outer context is still visible.

Totally, organizing according to a logical hierarchy is the point, but I don't see that as being different from common setup. Usually the one requires the other, no?

cfieber:
> Can you elaborate a bit on what you expect those other personas/use cases might be?

Literally every different type of user of Spinnaker...

concretely:
* I own a service, and own its declarative configuration (I think you have implemented for this case)
* I am a collaborator with a service, I don't know about or care about it's declarative configuration, but if it is in a bad or broken state that would be useful information for me. If it has changed recently that would be useful information for me.
* I am on a central team, I don't want to edit your declarative configuration, but I might need to pause actuation of your service
* ... etc

Who might interact with the Spinnaker UI? If they can see my application, what should they be about it from a declarative perspective?

luispollo:
Thanks, this helps @cfieber. So let's see:
> * I own a service, and own its declarative configuration (I think you have implemented for this case)

Right.

> * I am a collaborator with a service, I don't know about or care about it's declarative configuration, but if it is in a bad or broken state that would be useful information for me. If it has changed recently that would be useful information for me.

If you're a collaborator, my understanding was that the expectation was you'd have access to the service account, and there have the same permissions as the owner, no?

> * I am on a central team, I don't want to edit your declarative configuration, but I might need to pause actuation of your service

This is the case I referred to above where I think we could relax the permission checks to be an "or" and not an "and": if the caller has access to the application with the right level of permission (READ/WRITE), then it shouldn't matter if they have access to the service account? That would take care of this case, I think.

> Who might interact with the Spinnaker UI? If they can see my application, what should they be about it from a declarative perspective?

Same answer as above. Let me know if that makes sense. Very curious if @emjburns and @robfletcher have thoughts on this. Also @erikmunson (whom I forgot to invite earlier, sorry).

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
luispollo: I've revisited the implementation here, please take a look: https://github.com/spinnaker/keel/pull/955/commits/09bd23dc3325c896c670947a3cd22b88b186dc69
Looks like this for `ApplicationControllerTests`:
image

emjburns:
I agree with @cfieber regarding general authorization. I think we should allow many more people to do things (pause/resume/view/pin/unpin etc). I think we should only use write permissions to restrict submitting new declarative configs. 

It's easy to tighten permissions if we have a need, but I don't see the need right now. I share Cam's  worry that by adding these permissions we are locking people out of doing what they need to do.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
```diff
@@ -18,59 +18,76 @@
...
 import org.slf4j.LoggerFactory
-import org.springframework.security.core.Authentication
 import org.springframework.security.core.context.SecurityContextHolder
 import org.springframework.stereotype.Component
 
+/**
+ * Support for authorization of REST API calls.
+ *
+ * Permission is granted based on the user (as identified by the X-SPINNAKER-USER request header, or X509 client
+ * certificate, or metatron identity) having access to the "service account" associated with keel's entities
```
lorin, line 37: Non-Netflixers likely won't know what "metatron" refers to, so may want to note that it's a Netflix-internal credential management service.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
```diff
@@ -350,5 +342,13 @@ internal class ApplicationControllerTests : JUnit5Minutests {
           ))
       }
     }
+
+    testApiPermissions(mvc, jsonMapper, authorizationSupport, mapOf(
```
emjburns, line 346: what is this doing?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get
+import org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post
+import org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put
+import org.springframework.test.web.servlet.result.MockMvcResultMatchers.status
+
+/**
+ *  This function loops through the specified API requests and permissions and adds a test for each,
+ *  so that you don't have to manually write those tests in your REST controller test classes.
+ */
+fun ContextBuilder<*>.testApiPermissions(
```
emjburns, line 19: I understand that your goal here is to make it so we don't have to repeat code for tests. However, I am having a really hard time (even with your comments) understanding what this is testing, how it's testing it, and how to make changes or diagnose failures if I'm refactoring. I think that it should be very obvious what tests are testing - for me, looping and auto-generating things is much harder to understand than a verbose test or a series of similar tests. I prefer the tests that you've deleted - where the api and the description or behavior is explicit.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -74,6 +75,7 @@ class ArtifactController(
     path = ["/pin"]
   )
   @ResponseStatus(ACCEPTED)
+  @PreAuthorize("@authorizationSupport.hasApplicationPermission('WRITE', 'DELIVERY_CONFIG', #pin.deliveryConfigName)")
```
emjburns, line 78: It looks like pin and delete pin are different permissions. Is that intentional?

luispollo: No, good spot. Will fix.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
luispollo: https://github.com/spinnaker/keel/blob/2ac3cb709e91f0ca50c5773e8293d8e25fcefb0f/keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt#L121-L142

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationTestSupport.kt
luispollo: I appreciate your concern with readability and maintainability of the tests. However, a question I would ask back to you is: how often do you go back and refactor tests, especially those for  behavior that you take for granted once put in place, such as these authorization checks?

What I attempted to give you here is almost like a small DSL where you can actually clearly see what APIs are being tested and with which permissions, as you can see in the example below (or the one I gave you above):
https://github.com/spinnaker/keel/blob/2ac3cb709e91f0ca50c5773e8293d8e25fcefb0f/keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ResourceControllerTests.kt#L116-L135

I frankly cannot see how this is harder to understand than a bunch of `test` blocks that specify those API calls and checks, but also have a lot of that boilerplate code I abstracted away.

luispollo: P.S. Instead of this:
```kotlin
      ApiRequest("GET /resources/${resource.id}") to setOf(
        Permission(APPLICATION_AUTHZ, READ, RESOURCE),
        Permission(CLOUD_ACCOUNT_AUTHZ, READ, RESOURCE)
      )
```
...is this easier to read?
```kotlin
      ApiRequest("GET /resources/${resource.id}") to setOf(
        ApplicationPermission(READ, RESOURCE),
        CloudAccountPermission(READ, RESOURCE)
      )
```

luispollo:
@emjburns Upon further reflection, I've decided to rewrite all the tests using just the usual syntax. Although the test code was much more concise with my little framework, I agree that it was an "alien" concept in the middle of tests that are otherwise just using the minutest, mockk and strikt DSLs.

I hope this will be easier on your eyes now. Here's an example of what the tests look like in IntelliJ now:
image

P.S. I used my function to generate the code for me, though. 😉

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
```diff
@@ -52,6 +52,7 @@ class DefaultConfiguration {
 
   @Bean
   @Primary
+  @Qualifier("jsonMapper")
```
gal-yardeni, line 55: why was `@Qualifier("jsonMapper")` added?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -93,6 +98,7 @@ class ApplicationController(
     path = ["/{application}/environment/{environment}/constraints"],
     produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
   )
+  @PreAuthorize("@authorizationSupport.hasApplicationPermission('READ', 'APPLICATION', #application)")
```
gal-yardeni, line 101: why does here and below there are only 2 quotes `"@authorizationSupport..` as oppose to the rest that have 4 `"""@authorizationSupport..` ?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
```diff
@@ -18,59 +18,171 @@
...
+    override fun toString() = name.toLowerCase()
   }
 
-  fun userCanModifyApplication(application: String): Boolean = try {
-    val deliveryConfig = repository.getDeliveryConfigForApplication(application)
-    userCanModifyDeliveryConfig(deliveryConfig.serviceAccount, application, deliveryConfig.name)
-  } catch (e: NoSuchElementException) {
-    // If there are no delivery configs for that application return true so this is handled correctly in the controller.
-    true
+  private fun enabled() = dynamicConfigService.isEnabled("keel.authorization", true)
```
gal-yardeni, line 71: just out of curiosity, why didn't you use `@ConditionalOnProperty("xxx.enabled")` as in other classes?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
```diff
@@ -18,59 +18,171 @@
...
+
+  private fun passes(authorizationCheck: () -> Unit) =
+    try {
+      authorizationCheck()
+      true
+    } catch (e: AccessDeniedException) {
+      false
+    }
+
+  val Boolean.str: String
```
gal-yardeni, line 186: what is this magic?? can you add some comments / name it better for future reference 🙏?

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
luispollo: Because I rather name the object `jsonMapper`, and since the object created by default configuration comes from a function called `objectMapper`, Spring can't find it automatically unless I use a qualifier. I prefer `jsonMapper` to contrast it with the `yamlMapper` used in some of the classes (both of which are instances of `ObjectMapper`, so I prefer to avoid confusion).

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
luispollo: Because this has a single permission check, as opposed to the other ones which have two and I wanted to break in two lines for readability. It's just a string (a SpEL expression, to be precise), in both cases.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
luispollo: If I used `@ConditionalOnProperty`, then the entire bean would be missing, and the `@PreAuthorize` annotations would break, so I opted for checking inside the methods.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
luispollo: No magic, really. I just wanted something that could a `Boolean` that is `true` as `"ALLOWED"`, and `"DENIED"` otherwise to make it easier to log above. Unfortunately, you cannot override the `toString()` method with extensions like this, so I used this `str` property instead.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
luispollo: Ah, also, more importantly, this is using a Fast Property, not Spring config! 😃

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/SubmittedResource.kt
```diff
@@ -38,7 +38,7 @@ data class SubmittedResource<T : ResourceSpec>(
   val spec: T
 )
 
-val <T : ResourceSpec> SubmittedResource<T>.id: String
+val SubmittedResource<*>.id: String
```
emjburns, line 41: just curious, why the loosening of types here?

luispollo: Because the spec type doesn't matter to get the `id`. But this was a side-change, will undo.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
```diff
@@ -18,59 +18,178 @@
...
+   */
+  fun hasCloudAccountPermission(action: String, target: String, identifier: String) =
+    passes { hasCloudAccountPermission(Action.valueOf(action), TargetEntity.valueOf(target), identifier) }
+
+  /**
+   * Verifies that the caller has the specified permission (action) to access the application associated with the
+   * specified target object.
+   *
+   * @throws AccessDeniedException if caller does not have the required permission.
+   */
```
emjburns, line 98: For these functions, I'd expect a function starting with `has` to return a `Boolean`, not throw an exception if `false`. What do you think about changing these to just return false if the thing doesn't have access, since you're just swallowing the exception in `passes {}` anyways? Then you could remove the `passes()` function, and I think this would get simpler.

luispollo: The reason I wrote it this was was so that these functions could be used from the body of controllers and would take care of creating the exception with the right information, as opposed to letting each controller do that (including because the controller won't have all the info).

You can see a usage example here:
https://github.com/spinnaker/keel/blob/e313b5bbb5c0ee9d1d3836cd930000092dab7c76/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/InteractiveNotificationCallbackController.kt#L44

But I agree the name is bad. Will replace `has` with `verify` (as the comment says).

emjburns: Oh I see. I think since that specific controller seems like a special case I would have just thrown the exception there, instead of throwing it everywhere but swallowing it most places. Do you anticipate this method being called from controllers a lot?

luispollo: I had no idea if it will be common or not. I hope it won't, but it didn't seem to hurt to encapsulate that in `AuthorizationSupport`. I think we would be tempted to just copy/paste a "if this then throw exception" from one controller to another if I didn't, and end up with inconsistent error messages. From the point of view of consumers of this class, most uses should be using `@PreAuthorize` annotations with the `hasXyz` form and they shouldn't care about the implementation underneath.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
robfletcher: I definitely feel like the name could be more meaningful… `toAuthorization` maybe?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupportTests.kt
```diff
@@ -0,0 +1,236 @@
...
+          combinedRepository.getDeliveryConfigForApplication(application)
+        } returns deliveryConfig
+      }
+
+      listOf(Action.READ, Action.WRITE).forEach { action ->
+        context("user has no ${action.name} access to application") {
+          before {
+            every {
+              dynamicConfigService.isEnabled("keel.authorization", true)
+            } returns true
```
robfletcher, line 81: Why is it necessary to re-do this when it's done in the parent `before`?

lorin:
I'd love it if we could document the overall authorization strategy somewhere more permanent than this PR description. Do we have a canonical location for documenting these kinds of details?

luispollo:
> I'd love it if we could document the overall authorization strategy somewhere more permanent than this PR description. Do we have a canonical location for documenting these kinds of details?
> 
> _Edit: I just noticed that this info is in a comment in AuthorizationSupport.kt_

@lorin Do you feel that's enough, or do you have any suggestions for another place (documentation maybe?). I'd be happy to put that somewhere more relevant.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupportTests.kt
luispollo: Good catch. Will remove.

lorin:
@luispollo My vote would be to have something like a docs folder that lives in the repo that has developer-relevant documentation like this.

luispollo:
> @luispollo My vote would be to have something like a docs folder that lives in the repo that has developer-relevant documentation like this.

It was pointed out to me yesterday that we have an official home for developer documentation in the OSS docs: https://www.spinnaker.io/community/contributing/back-end-code/

luispollo:
@robfletcher Hopefully the doc I added helps with context on where we arrived with the guidelines for authorization: https://github.com/spinnaker/keel/blob/ecf9086692afebb2cbd77742126817115a6f3b1c/docs/authorization.md

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
```diff
@@ -18,59 +18,168 @@
...
+
+  private fun passes(authorizationCheck: () -> Unit) =
+    try {
+      authorizationCheck()
+      true
+    } catch (e: AccessDeniedException) {
+      false
+    }
+
+  private fun Boolean.toAuthorization() = if (this) "ALLOWED" else "DENIED"
```
lorin, line 184: Kotlin is wild!

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupportTests.kt
```diff
@@ -0,0 +1,233 @@
...
+            checkServiceAccountAccess(TargetEntity.DELIVERY_CONFIG, deliveryConfig.name)
+          }
+          expectThat(
+            hasServiceAccountAccess(TargetEntity.DELIVERY_CONFIG.name, deliveryConfig.name)
+          ).isFalse()
+        }
+      }
+    }
+
+    context("authorization is disabled") {
```
lorin, line 186: Are there any tests were authorization is enabled and the user has the required permissions? I didn't see any.

lorin: Never mind, I see these in the Controller tests.

PR #957, opened 2020-04-01
If we can't parse a resource in the database, stop checking

robfletcher:
After I promoted the change where we no longer check image resources, but instead check artifacts, we discovered that resource checking halted because Keel was no longer able to parse resources with `ImageSpec` payloads.

This change makes it so that if we hit that condition, we'll flag the resource as unreadable and stop checking it. That way resources we _can_ handle will continue being processed.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
@@ -9,10 +9,7 @@ import com.netflix.spinnaker.keel.persistence.KeelRepository
 import com.netflix.spinnaker.keel.telemetry.ArtifactCheckTimedOut
 import com.netflix.spinnaker.keel.telemetry.EnvironmentsCheckTimedOut
 import com.netflix.spinnaker.keel.telemetry.ResourceCheckTimedOut
-import java.time.Duration
-import java.util.concurrent.atomic.AtomicBoolean
-import kotlin.coroutines.CoroutineContext
-import kotlin.math.max
+import com.netflix.spinnaker.keel.telemetry.ResourceLoadFailed
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.TimeoutCancellationException
@@ -26,6 +23,10 @@ import org.springframework.context.ApplicationEventPublisher
 import org.springframework.context.event.EventListener
 import org.springframework.scheduling.annotation.Scheduled
 import org.springframework.stereotype.Component
+import java.time.Duration
+import java.util.concurrent.atomic.AtomicBoolean
+import kotlin.coroutines.CoroutineContext
+import kotlin.math.max
 
 @Component
 class CheckScheduler(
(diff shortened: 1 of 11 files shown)


keel-sql/src/main/resources/db/changelog/20200401-resource-check-veto.yml
```diff
@@ -0,0 +1,18 @@
...
+      changes:
+      - addColumn:
+          tableName: resource_last_checked
+          columns:
+          - column:
+              name: unreadable
+              type: boolean
+              defaultValue: false
+              constraints:
+                nullable: false
```
robfletcher, line 14: Add a new DB column so we can just say "don't keep trying to check this"

robfletcher:
Hold off reviewing for now. There's a reason that transaction was there… and I need to put some telemetry in place.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -309,29 +310,38 @@ open class SqlResourceRepository(
...
-          .also {
-            it.forEach { (uid, _, _, _) ->
-              insertInto(RESOURCE_LAST_CHECKED)
-                .set(RESOURCE_LAST_CHECKED.RESOURCE_UID, uid)
-                .set(RESOURCE_LAST_CHECKED.AT, now.toLocal())
-                .onDuplicateKeyUpdate()
-                .set(RESOURCE_LAST_CHECKED.AT, now.toLocal())
-                .execute()
-            }
+      jooq.select(RESOURCE.UID, RESOURCE.KIND, RESOURCE.METADATA, RESOURCE.SPEC)
```
emjburns, line 313: Does this need to still be in a transaction?

emjburns: LOL I see it did.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -61,20 +62,27 @@ class CheckScheduler(
...
+                /**
+                 * Allow individual resource checks to timeout but catch the `CancellationException`
+                 * to prevent the cancellation of all coroutines under [job]
+                 */
+                withTimeout(checkTimeout.toMillis()) {
+                  launch { resourceActuator.checkResource(it) }
+                }
+              } catch (e: TimeoutCancellationException) {
+                log.error("Timed out checking resource ${it.id}", e)
+                publisher.publishEvent(ResourceCheckTimedOut(it.kind, it.id, it.application))
```
lorin, line 84: Does the timeout metric currently appear on our keel dashboard? I don't think I've seen it there.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/CheckSchedulerTests.kt
```diff
@@ -111,20 +113,37 @@ internal object CheckSchedulerTests : JUnit5Minutests {
...
+        context("resources cannot be loaded from the database") {
+          before {
+            every {
+              repository.resourcesDueForCheck(any(), any())
+            } throws UnsupportedKind("some-invalid-kind")
+
+            checkResources()
+          }
+
+          test("all resources are checked") {
```
luispollo, line 143:
```suggestion
          test("a ResourceLoadFailed event is published") {
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -328,10 +329,20 @@ open class SqlResourceRepository(
                 .execute()
             }
           }
-          .map { (_, kind, metadata, spec) ->
+      }
+        .map { (uid, kind, metadata, spec) ->
+          try {
             constructResource(kind, metadata, spec)
+          } catch (e: Exception) {
```
lorin, line 336: This logic will ignore updates regardless of the exception that occurs. I worry about situations where constructResource throws an exception for a transient issue (e.g., networking issue connecting with the database). Couple of thoughts around that:

* We need to feed back info to a human being when a resource gets ignored. It should hopefully be rare enough that we can handle these as they come in. Since we're logging this as an error when it happens, I think we're good here, because we'll get alerted by RADAR.
* Human operator needs to know how to fix things when a resource gets into a bad state due to a transient issue. Looks like the operation is simple enough (just set ignore to `false` in the database record), but we need to make sure everyone knows how to do this.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
robfletcher: I think it's pretty new and I don't believe it got added to the dashboard, no

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
lorin: (Note that I saw we're logging as an error when the exception gets caught upstream, so no need to add any additional log statements).

PR #958, opened 2020-04-01
fix(core): Add supervisorScope around coroutines in CheckScheduler

luispollo:
...so that failing parallel coroutines under the same scope do not interrupt their siblings, which can cause unexpected side-effects like downstream requests killed mid-flight that we get no exception information for in the logs. Using `supervisorScope` ensures that the exception causing failure can be caught and logged.


emjburns:
Oh wow, interesting. I don't feel I have enough coroutine knowledge for my LGTM to be meaningful :)

luispollo:
> Oh wow, interesting. I don't feel I have enough coroutine knowledge for my LGTM to be meaningful :)

I'm no expert either, but I remember this from https://github.com/spinnaker/keel/pull/649

And it was actually good that I went back to look at that just now, because unlike what I said in the description above, `supervisorScope` does _not_ propagate any errors. I'm not sure that's the behavior we want. I'll go back to the docs to understand what options exist.

PR #959, opened 2020-04-02
feat(artifact): introduce SKIPPED status, more data for previous

emjburns:
Recently we were talking about artifact statuses. We have an `APPROVED` status which means "this artifact is ready to be deployed". But, we sometimes have the situation where an artifact is approved for an environment but never deployed to that environment. This is probably because the artifact was superseded by another version of an artifact (both were ready to deploy, but we deployed the latest one). 

We also have a `PENDING` status for artifacts to represent that something might be going on in the future, but we have no idea what. That is calculated like this: we look at all versions we know about, and versions that we have data for in an environment (like `APPROVED`/`CURRENT` etc). The versions that we know about but are not in the environment are marked `PENDING`. This is weird, because some versions are never going to happen (for example: two artifacts are waiting on manual judgment. You approve the latest one. We ignore that other version for ever and ever).

This PR introduces a `SKIPPED` status which will get set when an artifact is marked as `CURRENT` in an environment. When an artifact is marked as deployed, we
- set the existing `CURRENT` artifact to `PREVIOUS` (this pr also updates the `promotion_reference` column to contain what artifact replaced it)
- set any versions that are lower than the existing version and have a status of `APPROVED` to `SKIPPED` (for example: 1.1.1, 1.2.1, and 1.3.1 are all approved. 1.2.1 is marked as `CURRENT`. 1.1.1 would be marked as `SKIPPED`. 1.3.1 would be untouched)

This will make our existing data a bit wonkey (i.e. every place we now have `APPROVED` but that's out of date and confusing). It should be straightforward going forward.

This `SKIPPED` status is also used for pending judgments. That works like this:
- we calculate `PENDING` in the same way as described above
- we mark any versions that are < the current version in the environment as `SKIPPED`

After this PR is merged, we should have a better view of artifacts that will never be deployed.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
@@ -1,0 +1,112 @@
+package com.netflix.spinnaker.keel.persistence
+
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactStatus.RELEASE
+import com.netflix.spinnaker.keel.api.artifacts.DebianArtifact
+import com.netflix.spinnaker.keel.api.artifacts.VirtualMachineOptions
+import com.netflix.spinnaker.keel.core.api.DependsOnConstraint
+import com.netflix.spinnaker.keel.core.api.ManualJudgementConstraint
+import com.netflix.spinnaker.time.MutableClock
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import java.time.Clock
+import strikt.api.expect
+import strikt.assertions.containsExactlyInAnyOrder
+import strikt.assertions.isEmpty
+import strikt.assertions.isEqualTo
+
+/**
+ * In the artifact repository we have several methods that generate summary views of data
+ * that are returned in the /application/{application} endpoint.
+ * This class tests some of that data generation.
+ */
+abstract class ApplicationSummaryGenerationTests<T : ArtifactRepository> : JUnit5Minutests {
+
+  abstract fun factory(clock: Clock): T
+
+  val clock = MutableClock()
+
+  open fun T.flush() {}
+
+  data class Fixture<T : ArtifactRepository>(
+    val subject: T
+  ) {
+    // the artifact built off a feature branch
+    val artifact = DebianArtifact(
+      name = "keeldemo",
+      deliveryConfigName = "my-manifest",
+      reference = "my-artifact",
+      vmOptions = VirtualMachineOptions(baseOs = "bionic", regions = setOf("us-west-2")),
+      statuses = setOf(RELEASE)
+    )
+
+    val environment1 = Environment("aa")
+    val environment2 = Environment(
+      name = "bb",
+      constraints = setOf(DependsOnConstraint("test"), ManualJudgementConstraint())
+    )
+    val manifest = DeliveryConfig(
+      name = "my-manifest",
+      application = "fnord",
+      serviceAccount = "keel@spinnaker",
+      artifacts = setOf(artifact),
+      environments = setOf(environment1, environment2)
+    )
+    val version1 = "keeldemo-1.0.1-h11.1a1a1a1" // release
+    val version2 = "keeldemo-1.0.2-h12.2b2b2b2" // release
+//    val version3 = "keeldemo-1.0.3-h13.3c3c3c3" // release
+  }
+
+  open fun Fixture<T>.persist() {
+    with(subject) {
+      register(artifact)
+      setOf(version1, version2).forEach {
+        store(artifact, it, RELEASE)
+      }
+    }
+    persist(manifest)
+  }
+
+  abstract fun persist(manifest: DeliveryConfig)
+
+  fun tests() = rootContext<Fixture<T>> {
+    fixture { Fixture(factory(clock)) }
+
+    before {
+      persist()
+    }
+
+    after {
+      subject.flush()
+    }
+
+    context("artifact 1 skipped in env 1, mj before env 2") {
+      before {
+        // version 1 and 2 are approved in env 1 approved in env 1
+        subject.approveVersionFor(manifest, artifact, version1, environment1.name)
+        subject.approveVersionFor(manifest, artifact, version2, environment1.name)
+        // only version 2 is approved in env 2
+        subject.approveVersionFor(manifest, artifact, version2, environment2.name)
+        // version 1 has been skipped in env 1 by version 2
+        subject.markAsSkipped(manifest, artifact, version1, environment1.name, version2)
+        // version 2 was successfully deployed to both envs
+        subject.markAsSuccessfullyDeployedTo(manifest, artifact, version2, environment1.name)
+        subject.markAsSuccessfullyDeployedTo(manifest, artifact, version2, environment2.name)
+      }
+
+      test("skipped versions don't get a pending status in the next env") {
+        val summaries = subject.getEnvironmentSummaries(manifest).sortedBy { it.name }
+        expect {
+          that(summaries.size).isEqualTo(2)
+          that(summaries[0].artifacts.first().versions.current).isEqualTo(version2)
+          that(summaries[0].artifacts.first().versions.pending).isEmpty()
+          that(summaries[0].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
+          that(summaries[1].artifacts.first().versions.current).isEqualTo(version2)
+          that(summaries[1].artifacts.first().versions.pending).isEmpty()
+          that(summaries[1].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
+        }
+      }
+    }
+  }
+}
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
@@ -41,8 +41,8 @@ abstract class ApplicationSummaryGenerationTests<T : ArtifactRepository> : JUnit
       statuses = setOf(RELEASE)
     )
 
-    val environment1 = Environment("aa")
-    val environment2 = Environment(
+    val environmentA = Environment("aa")
+    val environmentB = Environment(
       name = "bb",
       constraints = setOf(DependsOnConstraint("test"), ManualJudgementConstraint())
     )
@@ -51,11 +51,10 @@ abstract class ApplicationSummaryGenerationTests<T : ArtifactRepository> : JUnit
       application = "fnord",
       serviceAccount = "keel@spinnaker",
       artifacts = setOf(artifact),
-      environments = setOf(environment1, environment2)
+      environments = setOf(environmentA, environmentB)
     )
     val version1 = "keeldemo-1.0.1-h11.1a1a1a1" // release
     val version2 = "keeldemo-1.0.2-h12.2b2b2b2" // release
-//    val version3 = "keeldemo-1.0.3-h13.3c3c3c3" // release
   }
 
   open fun Fixture<T>.persist() {
@@ -81,30 +80,30 @@ abstract class ApplicationSummaryGenerationTests<T : ArtifactRepository> : JUnit
       subject.flush()
     }
 
-    context("artifact 1 skipped in env 1, mj before env 2") {
+    context("artifact 1 superseded in envA, manual judgement before envB") {
       before {
-        // version 1 and 2 are approved in env 1 approved in env 1
-        subject.approveVersionFor(manifest, artifact, version1, environment1.name)
-        subject.approveVersionFor(manifest, artifact, version2, environment1.name)
-        // only version 2 is approved in env 2
-        subject.approveVersionFor(manifest, artifact, version2, environment2.name)
-        // version 1 has been skipped in env 1 by version 2
-        subject.markAsSkipped(manifest, artifact, version1, environment1.name, version2)
+        // version 1 and 2 are approved in env A
+        subject.approveVersionFor(manifest, artifact, version1, environmentA.name)
+        subject.approveVersionFor(manifest, artifact, version2, environmentA.name)
+        // only version 2 is approved in env B
+        subject.approveVersionFor(manifest, artifact, version2, environmentB.name)
+        // version 1 has been superseded in env A by version 2
+        subject.markAsSuperseded(manifest, artifact, version1, environmentA.name, version2)
         // version 2 was successfully deployed to both envs
-        subject.markAsSuccessfullyDeployedTo(manifest, artifact, version2, environment1.name)
-        subject.markAsSuccessfullyDeployedTo(manifest, artifact, version2, environment2.name)
+        subject.markAsSuccessfullyDeployedTo(manifest, artifact, version2, environmentA.name)
+        subject.markAsSuccessfullyDeployedTo(manifest, artifact, version2, environmentB.name)
       }
 
-      test("skipped versions don't get a pending status in the next env") {
+      test("superseded versions don't get a pending status in the next env") {
         val summaries = subject.getEnvironmentSummaries(manifest).sortedBy { it.name }
         expect {
           that(summaries.size).isEqualTo(2)
           that(summaries[0].artifacts.first().versions.current).isEqualTo(version2)
           that(summaries[0].artifacts.first().versions.pending).isEmpty()
-          that(summaries[0].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
+          that(summaries[0].artifacts.first().versions.superseded).containsExactlyInAnyOrder(version1)
           that(summaries[1].artifacts.first().versions.current).isEqualTo(version2)
           that(summaries[1].artifacts.first().versions.pending).isEmpty()
-          that(summaries[1].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
+          that(summaries[1].artifacts.first().versions.superseded).containsExactlyInAnyOrder(version1)
         }
       }
     }
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
@@ -80,30 +80,30 @@ abstract class ApplicationSummaryGenerationTests<T : ArtifactRepository> : JUnit
       subject.flush()
     }
 
-    context("artifact 1 superseded in envA, manual judgement before envB") {
+    context("artifact 1 skipped in envA, manual judgement before envB") {
       before {
         // version 1 and 2 are approved in env A
         subject.approveVersionFor(manifest, artifact, version1, environmentA.name)
         subject.approveVersionFor(manifest, artifact, version2, environmentA.name)
         // only version 2 is approved in env B
         subject.approveVersionFor(manifest, artifact, version2, environmentB.name)
-        // version 1 has been superseded in env A by version 2
-        subject.markAsSuperseded(manifest, artifact, version1, environmentA.name, version2)
+        // version 1 has been skipped in env A by version 2
+        subject.markAsSkipped(manifest, artifact, version1, environmentA.name, version2)
         // version 2 was successfully deployed to both envs
         subject.markAsSuccessfullyDeployedTo(manifest, artifact, version2, environmentA.name)
         subject.markAsSuccessfullyDeployedTo(manifest, artifact, version2, environmentB.name)
       }
 
-      test("superseded versions don't get a pending status in the next env") {
+      test("skipped versions don't get a pending status in the next env") {
         val summaries = subject.getEnvironmentSummaries(manifest).sortedBy { it.name }
         expect {
           that(summaries.size).isEqualTo(2)
           that(summaries[0].artifacts.first().versions.current).isEqualTo(version2)
           that(summaries[0].artifacts.first().versions.pending).isEmpty()
-          that(summaries[0].artifacts.first().versions.superseded).containsExactlyInAnyOrder(version1)
+          that(summaries[0].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
           that(summaries[1].artifacts.first().versions.current).isEqualTo(version2)
           that(summaries[1].artifacts.first().versions.pending).isEmpty()
-          that(summaries[1].artifacts.first().versions.superseded).containsExactlyInAnyOrder(version1)
+          that(summaries[1].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
         }
       }
     }
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
@@ -54,6 +54,7 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
       vmOptions = VirtualMachineOptions(baseOs = "bionic", regions = setOf("us-west-2")),
       statuses = setOf(SNAPSHOT)
     )
+
     // the artifact built off of master
     val artifact2 = DebianArtifact(
       name = "keeldemo",
@@ -417,14 +418,14 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
               subject.markAsSuccessfullyDeployedTo(manifest, artifact1, version3, environment1.name)
             }
 
-            test("the lower version was marked as skipped") {
+            test("the lower version was marked as superseded") {
               val result = versionsIn(environment1)
               expectThat(result) {
                 get(ArtifactVersionStatus::pending).isEmpty()
                 get(ArtifactVersionStatus::current).isEqualTo(version3)
                 get(ArtifactVersionStatus::deploying).isNull()
                 get(ArtifactVersionStatus::previous).containsExactly(version1)
-                get(ArtifactVersionStatus::skipped).containsExactly(version2)
+                get(ArtifactVersionStatus::superseded).containsExactly(version2)
               }
             }
           }
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
@@ -408,6 +408,26 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
               }
             }
           }
+
+          context("there are two approved versions for the environment and the latter was deployed") {
+            before {
+              clock.incrementBy(Duration.ofHours(1))
+              subject.approveVersionFor(manifest, artifact1, version2, environment1.name)
+              subject.approveVersionFor(manifest, artifact1, version3, environment1.name)
+              subject.markAsSuccessfullyDeployedTo(manifest, artifact1, version3, environment1.name)
+            }
+
+            test("the lower version was marked as skipped") {
+              val result = versionsIn(environment1)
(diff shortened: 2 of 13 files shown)


asher:
promotion_reference has a special meaning currently, see i.e. https://github.com/spinnaker/keel/blob/master/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt#L488-L494. Repurposing it this way, without rewriting the existing use will probably break automated artifact version vetoing and rollbacks.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -408,6 +408,26 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
               }
             }
           }
+
+          context("there are two approved versions for the environment and the latter was deployed") {
+            before {
+              clock.incrementBy(Duration.ofHours(1))
```
luispollo, line 414: If you think it's easier to read, you can now do `clock.tickHours(1)`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/PromotionStatus.kt
```diff
@@ -1,5 +1,32 @@
 package com.netflix.spinnaker.keel.core.api
 
 enum class PromotionStatus {
-  PENDING, APPROVED, DEPLOYING, CURRENT, PREVIOUS, VETOED
+  /**
+   * Waiting on constraint evaluation before being approved
```
luispollo, line 5: Thanks for documenting the enum values! ❤️

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -395,15 +396,36 @@ class SqlArtifactRepository(
...
         .update(ENVIRONMENT_ARTIFACT_VERSIONS)
         .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS, PREVIOUS.name)
+        .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_REFERENCE, version)
         .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(environmentUid))
         .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifact.uid))
         .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.eq(CURRENT.name))
         .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.ne(version))
         .execute()
     }
+    sqlRetry.withRetry(WRITE) {
```
luispollo, line 410: Should wrap these in a transaction now that there's more than one write operation.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -395,15 +396,36 @@ class SqlArtifactRepository(
...
+      // update any past artifacts that were "APPROVED" to be "SKIPPED
+      // because the new version takes precedence
+      val approvedButOld = jooq.select(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
+        .from(ENVIRONMENT_ARTIFACT_VERSIONS)
+        .where(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.eq(APPROVED.name))
+        .fetch(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
+        .filter { artifact.versioningStrategy.comparator.compare(it, version) > 0 } // get the lower versions
```
luispollo, line 417: This is very confusing to read because the contract of any `Comparator` is to return `> 0` when the item on the left is higher than the one on the right, but because we've hard-coded the version strategy comparators to be descending by default, you have to use that in the backwards sense here. How about adding an `ascendingComparator` to `versionStrategy` which returns `comparator.inverse()` so that the original semantics is maintained? Even better, if you could add helper methods like `versioningStrategy.isHigher(ver1, ver2)` and `versioningStrategy.isLower(ver1, ver2)`, that would make it much more readable IMHO.

erikmunson:
😍 fantastic!

Just for my own clarification: does this also modify the behavior of `pending` in a similar way to `approved`? Don't know enough about the code to tell fo sure from reading the diff. I know we'd talked about both and couldn't tell if you were thinking about each state separately or also changing that here.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/PromotionStatus.kt
gal-yardeni: +1!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -395,15 +396,36 @@ class SqlArtifactRepository(
...
+
+      jooq
+        .update(ENVIRONMENT_ARTIFACT_VERSIONS)
+        .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS, SKIPPED.name)
+        .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_REFERENCE, version)
+        .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(environmentUid))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifact.uid))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.eq(APPROVED.name))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.`in`(*approvedButOld.toTypedArray()))
```
gal-yardeni, line 426: what does the star in `*approvedButOld`? does? just wondering since I haven't seen this pattern here before

emjburns: It takes a list of strings and turns it into whatever form sql needs to actually be able to tell whether a string is in a list of strings :D

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: I'm going to add a helper method to just the sql artifact repository to make it easier to read, but not have to do a big change. It's not super straightforward to add helper methods onto versioning strategy.

emjburns:
Updated this PR to address my accidental misuse of the `PROMOTION_REFERENCE` field and deal with `PENDING` status.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
```diff
@@ -0,0 +1,112 @@
...
+    val manifest = DeliveryConfig(
+      name = "my-manifest",
+      application = "fnord",
+      serviceAccount = "keel@spinnaker",
+      artifacts = setOf(artifact),
+      environments = setOf(environment1, environment2)
+    )
+    val version1 = "keeldemo-1.0.1-h11.1a1a1a1" // release
+    val version2 = "keeldemo-1.0.2-h12.2b2b2b2" // release
+//    val version3 = "keeldemo-1.0.3-h13.3c3c3c3" // release
```
gal-yardeni, line 58: is `version3` supposed to be here?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -208,6 +220,46 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
...
+    artifact.versioningStrategy.comparator.compare(new, existing) > 0
+
+  /**
+   * Returns true if the new version is higher than the existing version.
+   * Note: the artifact comparitors are decending by default
+   */
+  fun isHigher(artifact: DeliveryArtifact, new: String, existing: String): Boolean =
+    artifact.versioningStrategy.comparator.compare(new, existing) < 0
+
+  /**
```
gal-yardeni, line 238: General question - we are relaying here on the fact that a newer version will get the higher number. is it always the case?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -208,6 +220,46 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
    * different values.
    */
   override fun itemsDueForCheck(minTimeSinceLastCheck: Duration, limit: Int): Collection<DeliveryArtifact>
+
+  /**
+   * Returns true if the new version is lower than the existing version.
+   * Note: the artifact comparitors are decending by default
+   */
+  fun isLower(artifact: DeliveryArtifact, new: String, existing: String): Boolean =
```
gal-yardeni, line 228: suggestion - maybe use `old` and `new` instead of `low` and `high`? I saw that you used these terms in other places in the code and it made sense to me better..

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
```diff
@@ -0,0 +1,112 @@
...
+      persist()
+    }
+
+    after {
+      subject.flush()
+    }
+
+    context("artifact 1 skipped in env 1, mj before env 2") {
+      before {
+        // version 1 and 2 are approved in env 1 approved in env 1
```
lorin, line 86: nit: "approved in env 1" appears twice in this comment

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    before {
+      persist()
+    }
+
+    after {
+      subject.flush()
+    }
+
+    context("artifact 1 skipped in env 1, mj before env 2") {
```
luispollo, line 84:
```suggestion
    context("artifact 1 skipped in env 1, manual judgement before env 2") {
```
I always read "mj" as "Michael Jordan", can't help it. 🏀

luispollo, line 84: nit: I find it a bit hard to read the tests when we use the same naming convention for different types of objects ("environment 1", "artifact 1"). I've been using letters for environments and numbers for artifacts ("environment A", "artifact 1") because numbers kind of imply a sense of ordering/versions, and for environments we just need different names.

luispollo: Hmm... actually artifacts are != versions in this case. Maybe "environment E1", "artifact A1"?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
```diff
@@ -26,12 +33,13 @@ data class EnvironmentSummary(
     artifacts.find { it.name == artifact.name && it.type == artifact.type }
       ?.let {
         when (version) {
-          it.versions.current -> PromotionStatus.CURRENT
-          it.versions.deploying -> PromotionStatus.DEPLOYING
-          in it.versions.previous -> PromotionStatus.PREVIOUS
-          in it.versions.approved -> PromotionStatus.APPROVED
-          in it.versions.pending -> PromotionStatus.PENDING
-          in it.versions.vetoed -> PromotionStatus.VETOED
+          it.versions.current -> CURRENT
```
luispollo, line 36: I've now configured IntelliJ to do this for me automatically. Didn't know it was a thing, sorry. 😄

lorin:
The PR description says:

> This PR introduces a `SKIPPED` status which will get set when an artifact is marked as `DEPLOYED` to an environment.

What does *when an artifact is marked as `DEPLOYED`* mean? At first, I thought it was about PromotionStatus, but there's no `DEPLOYED` enum value.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -382,27 +383,49 @@ class SqlArtifactRepository(
...
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID, environmentUid)
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID, artifact.uid)
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION, version)
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT, currentTimestamp())
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS, CURRENT.name)
+          .onDuplicateKeyUpdate()
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT, currentTimestamp())
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS, CURRENT.name)
+          .execute()
+        // update old "CURRENT" to "PREVIOUS, set promotionReference for use in summary data
```
luispollo, line 399:
```suggestion
        // update old "CURRENT" to "PREVIOUS
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -382,27 +383,49 @@ class SqlArtifactRepository(
...
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS, PREVIOUS.name)
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.REPLACED_BY, version)
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.REPLACED_AT, currentTimestamp())
+          .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(environmentUid))
+          .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifact.uid))
+          .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.eq(CURRENT.name))
+          .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.ne(version))
+          .execute()
+        // update any past artifacts that were "APPROVED" to be "SKIPPED
+        //        // because the new version takes precedence
```
luispollo, line 411:
```suggestion
        // because the new version takes precedence
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -162,6 +163,17 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
...
 
+  /**
+   * Marks a version of an artifact as skipped for an environment, with information on what version skipped it.
+   */
+  fun markAsSkipped(
+    deliveryConfig: DeliveryConfig,
+    artifact: DeliveryArtifact,
+    version: String,
+    targetEnvironment: String,
+    skippedByVersion: String
```
luispollo, line 174:
```suggestion
    supersededByVersion: String
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -162,6 +163,17 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
     targetEnvironment: String
   )
 
+  /**
+   * Marks a version of an artifact as skipped for an environment, with information on what version skipped it.
```
luispollo, line 167:
```suggestion
   * Marks a version of an artifact as skipped for an environment, with information on what version superseded it.
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -162,6 +163,17 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
...
+   */
+  fun markAsSkipped(
```
luispollo, line 169: This function is currently only used in tests. Do you anticipate it being used elsewhere? If not, should probably move it to the test classes.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
```diff
@@ -495,6 +519,12 @@ class InMemoryArtifactRepository(
       .toList()
   }
 
+  private data class Skipped(
+    val version: String,
+    val skippedByVersion: String,
+    val skippedAt: Instant
```
luispollo, line 525:
```suggestion
    val supersededByVersion: String,
    val supersededAt: Instant
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -675,17 +724,22 @@ class SqlArtifactRepository(
...
           name = artifact.name,
           type = artifact.type,
           statuses = releaseStatuses,
           versions = ArtifactVersionStatus(
-            current = versions[CURRENT]?.firstOrNull(),
+            current = currentVersion,
             deploying = versions[DEPLOYING]?.firstOrNull(),
-            pending = versions[PENDING] ?: emptyList(),
+            // take out stateful constraint values that will never happen
+            pending = removeLowerIfCurrentExists(artifact, currentVersion, versions[PENDING]),
```
luispollo, line 737: This is pretty confusing. The comment and function name seem disconnected. I'm not sure what to suggest, but wanted to point it out because I think it'll be hard for anyone maintaining this code to understand.

luispollo: Maybe `higherThanCurrent` to follow the pattern you used naming the other function? I think that would make much more sense.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -813,42 +866,23 @@ class SqlArtifactRepository(
...
-                .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.ne(VETOED.name))
-                .orderBy(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT.asc())
-                .limit(1)
-                .fetchOne { (replacedBy, replacedAt) ->
-                  Pair(replacedBy, replacedAt)
-                } ?: Pair(null, null)
-            }
-            else -> Pair(null, null)
-          }
+        .fetchOne { (version, deployedAt, promotionStatus, replacedBy, replacedAt) ->
```
luispollo, line 885: Oh! did I reinvent the wheel there?! Nooo... 😞

luispollo: Ah, no, I see it now. You're updating it in `markAsSuccessfullyDeployedTo`. Smart! 👍

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
emjburns: Yes! It is always the case that we are deciding which is the newer version by this comparator, and "higher" version will always be newer. It's one of our core artifact assumptions. It's definitely an assumption, but it's what allows us to do cool artifact things.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
emjburns: Ah, I'll remove. Good call

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
emjburns: You're right, that's sooooo much better.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
emjburns: 🏀 lol will change

emjburns: Moved to environments A and B (was already doing that for the names), versions will stay with numbers. That's a good idea in general I'll try and remember to keep doing it!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
emjburns: I got it from fletch, I didn't know it was a thing before him either. You'll have to show me how to make intellij do that automatically

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
emjburns: I anticipate that it might, which is why I kept it here.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: I updated all these with gal's suggesting of wording above - I agree it's confusing.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: Yes! So we didn't have to do the nested query here! Just making an improvement :) plus it means that it's nicer to troubleshoot cause you can see things in the database

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
emjburns: Oh, this is better wording. Let me see how to work this in.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/PromotionStatus.kt
```diff
@@ -1,5 +1,38 @@
...
+   */
+  PENDING,
+
+  /**
+   * Has passed constraints successfully, and wil be deployed imminently
```
erikmunson, line 10:
```suggestion
   * Has passed constraints successfully, and will be deployed imminently
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/PromotionStatus.kt
```diff
@@ -1,5 +1,38 @@
...
+
+  /**
+   * Never allowed to be in the environment again
+   */
+  VETOED,
+
+  /**
+   * Was approved for the environment, but a newer version was deployed instead
+   */
+  SUPERSEDED
```
erikmunson, line 37: fwiw I will most likely continue to use "skipped" in the UI — while in isolation superseded is the most precise term for this concept, when taken in the context of it living alongside the "previous" status it gets pretty ambiguous. Part of the meaning held in the term "skipped" is that _it was never deployed_ (but was rather _skipped_ over in favor of another version) — superseded does not carry that connotation and will likely be perceived as meaning the same thing as previous (which we call "decommissioned" in deck) when folks are looking at it on a UI.

erikmunson: not to say that that has anything to do with whether it's useful as a naming scheme in keel's codebase

luispollo: I think "skipped" is better in the UI, and was fine even as a status enum value. I just suggested that the term in the `skippedBy` and `skippedAt` fields could be replaced with "superseded" because it sounded weird in that context.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -382,27 +383,49 @@ class SqlArtifactRepository(
...
-    sqlRetry.withRetry(WRITE) {
-      jooq
-        .update(ENVIRONMENT_ARTIFACT_VERSIONS)
-        .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS, PREVIOUS.name)
-        .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(environmentUid))
-        .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifact.uid))
-        .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.eq(CURRENT.name))
-        .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.ne(version))
-        .execute()
+      jooq.transaction { config ->
```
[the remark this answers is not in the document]
emjburns: `@Transactional` is something that spring does (we use it on the api layer sometimes), and the `jooq` one is for when you're actually in the sql repository and can control every part of the transaction

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -594,6 +617,32 @@ class SqlArtifactRepository(
     }
   }
 
+  override fun markAsSkipped(
```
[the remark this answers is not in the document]
emjburns: it's just in tests (there's a comment on the interface method for that) I wanted to easily call it in the tests with a separate impl in each repository

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/PromotionStatus.kt
emjburns: Ooo I see. I have no opinion on this - is it worth me changing the enum back to `SKIPPED` then, and just using `skippedBy` and `skippedAt` instead of superseded everywhere?

erikmunson: I really have no stake / opinion on this code specifically, so I'll let ya'll decide how you feel 😄

lorin: I'm a big fan of consistency in terminology between implementation in the code and what's exposed in the UI (I was very influenced by Eric Evans's book _Domain-Driven Design_). If the UI is going to use the term "skipped", I think that the code should as well.

emjburns: Okay, I'll do that then! Thanks for weighing in.

PR #972, opened 2020-04-06
chore(artifacts): make deployed image approver cloud agnostic

emjburns:
This makes the way we react to existing artifacts deployed in clusters cloud agnostic. This refactor was made possible by some work that @luispollo did on getting an artifact associated with a cluster. No change in logic, just simplifying and cleaning.


keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployedListener.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.context.event.EventListener
+import org.springframework.stereotype.Component
+
+@Component
+class ArtifactDeployedListener(
+  private val repository: KeelRepository
+) {
+  private val log = LoggerFactory.getLogger(javaClass)
+
+  @EventListener(ArtifactVersionDeployed::class)
```
gal-yardeni, line 16: Trying to understand the logic here - `onArtifactVersionDeployed` is called when an `ArtifactVersionDeployed` event is emitted, correct? If this event has published, it means that the artifact is deployed? if so, why the call to `isApprovedFor` is done after it being deployed?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/image/CurrentlyDeployedDockerImageApprover.kt
```diff
@@ -1,61 +0,0 @@
...
-
-  @EventListener(ArtifactVersionDeployed::class)
-  fun onArtifactVersionDeployed(event: ArtifactVersionDeployed) =
-    runBlocking {
-      val resourceId = event.resourceId
-      val resource = repository.getResource(resourceId)
-      val deliveryConfig = repository.deliveryConfigFor(resourceId)
-      val env = repository.environmentFor(resourceId)
-
-      (resource.spec as? TitusClusterSpec)?.let { spec ->
```
gal-yardeni, line 28: why can we skip the cloud providers' checks? (or maybe just a link to the corresponding PR that you've mentioned in the description)

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployedListener.kt
emjburns: Great question. I had the same question myself as I was porting this logic over!

The event is emitted when we check resources and get the current server group from clouddriver. We emit this event for the version we see in that current server group. I thought it might be good to keep this logic so that we would essentially mark the version as approved then deployed as we went through pushing the artifact through the environments. I think this logic avoids us marking an artifact as `DEPLOYED` if it has not met the constraints/etc. 

However, now that I explain it to you I'm wondering if we should take away that check. It seems we'd want to know if an artifact was already in the environment asap. Let me think about this some more.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/image/CurrentlyDeployedDockerImageApprover.kt
emjburns: Oh! Before, like a month ago, there was no way to get the artifact from the resource without knowing what cloud provider it was and reaching into the spec. That's why you see the spec stuff here. But Luis put in a method to get the artifact associated with a resource. So, you no longer need to know the type of the resource. So, we don't need the cloud provider stuff in here anymore and can make this generic.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployedListener.kt
luispollo: +1 for removing the check. Just as an example, if it had been deployed there outside of keel's control, and the environment in keel has a manual judgement, we can't make any assumptions about whether that constraint actually passed. But we do know it was deployed, so I'd just remove the check and accept that an artifact version may transition directly to deployed.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployedListenerTests.kt
```diff
@@ -0,0 +1,74 @@
...
+    fixture { Fixture() }
+
+    before {
+      every { repository.getResource(r.id) } returns r
+      every { repository.deliveryConfigFor(r.id) } returns config
+      every { repository.environmentFor(r.id) } returns config.environments.first()
+    }
+
+    context("no artifact associated with the resource") {
+      test("nothing is marked deploying") {
```
luispollo, line 41: "deploying" or "deployed"?

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployedListenerTests.kt
```diff
@@ -0,0 +1,74 @@
...
+    context("an artifact is associated with a resource") {
+      before {
+        every { r.findAssociatedArtifact(config) } returns artifact
+      }
+      context("artifact is approved for env") {
+        before {
+          every { repository.isApprovedFor(any(), any(), event.artifactVersion, any()) } returns true
+        }
+
+        test("version is marked deploying") {
```
luispollo, line 57: "deploying" or "deployed"?

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployedListenerTests.kt
```diff
@@ -0,0 +1,74 @@
...
+          subject.onArtifactVersionDeployed(event)
+          verify(exactly = 1) { repository.markAsSuccessfullyDeployedTo(any(), any(), event.artifactVersion, any()) }
+        }
+      }
+
+      context("artifact is not approved for env") {
+        before {
+          every { repository.isApprovedFor(any(), any(), event.artifactVersion, any()) } returns false
+        }
+        test("nothing is marked as deploying") {
```
luispollo, line 67: "deploying" or "deployed"?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/image/CurrentlyDeployedDockerImageApprover.kt
emjburns: What do you mean?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactDeployedListener.kt
emjburns: My worry is that our check for whether or not an artifact was approved for an environment right now just checks if a row exists in the `environment_artifact_versions` table, and that the status is not set to veto. So I think changing the logic here would have unintended consequences. I don't want to make a logic change in a PR that is just a clean up.

luispollo: Fair enough. You said you were gonna think about it, so I wasn't sure what you had decided. Fine by me. 😃

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/image/CurrentlyDeployedDockerImageApprover.kt
gal-yardeni: forget about it :) nevermind

PR #974, opened 2020-04-06
Schema upgrade

robfletcher:
There's a new interface `SchemaMigrator` that defines an `input` and `output` `ResourceKind` and has a function for converting `input` to `output`. We'd have one of these for each version bump, e.g. `ClusterV1ToV2Migrator`, etc.

The `SchemaMigrator` instances are injected into `SqlResourceRepository` so that when a resource is read from the database we can seamlessly "upgrade" it to the latest schema before returning it. That way handlers only need to support the current version of a resource kind. We'll keep the models around for the older versions but they will not be referenced by any handlers, only by migrators.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/SpecMigrator.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * A component used to migrate an older version of a [com.netflix.spinnaker.keel.api.ResourceSpec]
+ * to a current one.
+ */
+interface SpecMigrator {
+  val supportedKind: ResourceKind
+
+  fun migrate(spec: Map<String, Any?>): Pair<ResourceKind, Map<String, Any?>>
+}
```
robfletcher, line 13: This is the interface for migrating a spec from vn to vn+1

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/SpecMigrator.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Recursively applies [SpecMigrator]s to bring a [spec] of [kind] up to the latest version.
+ */
+fun Collection<SpecMigrator>.migrate(kind: ResourceKind, spec: Map<String, Any?>): Pair<ResourceKind, Map<String, Any?>> =
+  find { it.supportedKind == kind }
+    ?.migrate(spec)
+    ?.let { (kind, spec) -> migrate(kind, spec) }
+    ?: kind to spec
```
robfletcher, line 22: Extension function for migrating vn -> vn+1 -> vn+2 and so on

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -85,11 +88,15 @@ open class SqlResourceRepository(
...
-    )
+    specMigrators
+      .migrate(parseKind(kind), objectMapper.readValue(spec))
+      .let { (endKind, endSpec) ->
+        Resource(
+          endKind,
+          objectMapper.readValue<Map<String, Any?>>(metadata).asResourceMetadata(),
+          objectMapper.convertValue(endSpec, resourceTypeIdentifier.identify(endKind))
+        )
+      }
```
robfletcher, line 99: Inside of `SqlResourceRepository` this function turns the row results into resources with the correct spec type. Now I'm just using the migrators up front so that we only ever try to parse the latest type of spec. We don't need to keep the spec classes for the older versions around.

luispollo:
This is pretty neat and straightforward -- I like the general approach a lot. However, I wondered about just using `Map`s versus keeping the old versions of spec types around and using generic types for the `SpecMigrator`s, so we don't put ourselves in a position of having to handle specs that are just 💩and/or reimplement a lot of custom validation code based on maps in these migrators. 

I believe that is the approach Kubernetes takes, and they have a clear deprecation policy of API resources:
- https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api (see rule 4.a)
- https://kubernetes.io/docs/setup/release/version-skew-policy/

TL;DR: 
> GA: 12 months or 3 releases (whichever is longer)
> Beta: 9 months or 3 releases (whichever is longer)
> Alpha: 0 releases

robfletcher:
Yeah, it's an interesting question. I wanted to get this working and see how sane it looked first. It does feel like writing migrators that deal with maps would be nasty.

One thing I'd like us to do in that case is formalize a naming convention for newer versions of spec implementations. I can see it getting messy and confusing if it's just a free-for-all and we're not sure which version of a particular kind's spec is the current one.

I wonder if we should also enforce that we cannot have handlers for multiple versions of the same kind simultaneously. Obviously we could always revisit that later but I'd like to stick to the practice of always evaluating upgraded versions of resources even if the user has not updated their definitions.

robfletcher:
I think we also need a `kind` property on each `ResourceSpec` implementation's companion object or something

luispollo:
> I wonder if we should also enforce that we cannot have handlers for multiple versions of the same kind simultaneously. Obviously we could always revisit that later but I'd like to stick to the practice of always evaluating upgraded versions of resources even if the user has not updated their definitions.

I think so! Multiple spec versions, yes, but multiple handlers, no. Handlers should only ever be concerned with the latest/current version of the specs.

robfletcher:
> I think it would be prudent to include information that can be passed to handlers somehow about the original version submitted, just in case you need to have special handling for corner cases for incompatible changes in the spec format inside that handler's code.

I'd actually really like to avoid that kind of complexity if at all possible. I get the temptation to say it might be useful but it makes my skin crawl thinking of the horrible special cases that would end up littering the codebase.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
```diff
@@ -144,11 +144,6 @@ abstract class ResourceHandler<S : ResourceSpec, R : Any>(
   open suspend fun actuationInProgress(resource: Resource<S>): Boolean = false
 }
 
-data class SupportedKind<SPEC : ResourceSpec>(
-  val kind: ResourceKind,
-  val specClass: Class<SPEC>
-)
```
robfletcher, line 150: Moved to its own file as it's more important now

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ec2.kt
```diff
@@ -16,6 +16,19 @@
...
 val SPINNAKER_EC2_API_V1 = ApiVersion("ec2")
 const val CLOUD_PROVIDER = "aws"
+
+val CLUSTER_V1 = kind<ClusterSpec>("ec2/cluster@v1")
+
+val SECURITY_GROUP_V1 = kind<SecurityGroupSpec>("ec2/security-group@v1")
+
+val CLASSIC_LOAD_BALANCER_V1 = kind<ClassicLoadBalancerSpec>("ec2/classic-load-balancer@v1")
+
+val APPLICATION_LOAD_BALANCER_V1 = kind<ApplicationLoadBalancerSpec>("ec2/application-load-balancer@v1")
```
robfletcher, line 34: Made constants for all the resource kinds. These are exported as beans in `EC2Config` so we no longer rely on `ResourceHandler` implementations to work out what kinds we support. That means we can have kinds with no associated handler (e.g. the handler supports the latest version, but we still have the class for the older versions).

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ApplicationLoadBalancerDefaultsResolver.kt
```diff
@@ -4,12 +4,12 @@ import com.netflix.spinnaker.keel.api.Resource
...
 import com.netflix.spinnaker.keel.api.plugins.Resolver
 import com.netflix.spinnaker.keel.clouddriver.model.ApplicationLoadBalancerModel
-import com.netflix.spinnaker.keel.ec2.SPINNAKER_EC2_API_V1
+import com.netflix.spinnaker.keel.ec2.APPLICATION_LOAD_BALANCER_V1
 import org.springframework.stereotype.Component
 
 @Component
 class ApplicationLoadBalancerDefaultsResolver : Resolver<ApplicationLoadBalancerSpec> {
-  override val supportedKind = SPINNAKER_EC2_API_V1.qualify("application-load-balancer")
+  override val supportedKind = APPLICATION_LOAD_BALANCER_V1
```
robfletcher, line 12: `ResourceHandler` implementations can now reference one of the constants to indicate what kind they support

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/LegacySpecUpgradeTests.kt
```diff
@@ -0,0 +1,145 @@
...
+import dev.minutest.rootContext
+import java.time.Clock.systemUTC
+import java.time.Instant
+import java.time.Instant.EPOCH
+import strikt.api.expectCatching
+import strikt.assertions.isA
+import strikt.assertions.isEqualTo
+import strikt.assertions.succeeded
+
+internal class LegacySpecUpgradeTests : JUnit5Minutests {
```
robfletcher, line 24: Tests for running legacy specs in the database through migrators

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/KindController.kt
```diff
@@ -10,9 +10,9 @@ import org.springframework.web.bind.annotation.RestController
...
 @RequestMapping(path = ["/kinds"])
 class KindController(
-  val plugins: List<ResourceHandler<*, *>>
+  val kinds: List<SupportedKind<*>>
 ) {
   @GetMapping(produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE])
   fun get(): Set<String> =
-    plugins.map { it.supportedKind.kind.toString() }.toSet()
+    kinds.map { it.kind.toString() }.toSet()
 }
```
robfletcher, line 18: This is an example of using the kinds directly from Spring rather than having to look at the handlers.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/config/EC2Config.kt
```diff
@@ -42,6 +46,18 @@ import org.springframework.context.annotation.Configuration
 @ConditionalOnProperty("keel.plugins.ec2.enabled")
 class EC2Config {
 
+  @Bean
+  fun clusterV1() = CLUSTER_V1
```
luispollo, line 50: This feels a bit cumbersome to maintain. What would happen if we had a single `@Bean` that is the set of all supported kinds?
```kotlin
@Bean
fun supportedKinds() = setOf(CLUSTER_V1, SECURITY_GROUP_V1, CLASSIC_LOAD_BALANCER_V1, APPLICATION_LOAD_BALANCER_V1) 
```

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ec2.kt
```diff
@@ -16,6 +16,19 @@
...
+import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec
+import com.netflix.spinnaker.keel.api.ec2.ClassicLoadBalancerSpec
+import com.netflix.spinnaker.keel.api.ec2.ClusterSpec
+import com.netflix.spinnaker.keel.api.ec2.SecurityGroupSpec
+import com.netflix.spinnaker.keel.api.plugins.kind
 
 val SPINNAKER_EC2_API_V1 = ApiVersion("ec2")
 const val CLOUD_PROVIDER = "aws"
+
+val CLUSTER_V1 = kind<ClusterSpec>("ec2/cluster@v1")
```
luispollo, line 28: I'd name this `EC2_CLUSTER_V1` for consistency with the Titus implementation and added clarity.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -85,11 +88,15 @@ open class SqlResourceRepository(
    * Constructs a resource object from its database representation
    */
   private fun constructResource(kind: String, metadata: String, spec: String) =
-    Resource(
-      parseKind(kind),
-      objectMapper.readValue<Map<String, Any?>>(metadata).asResourceMetadata(),
-      objectMapper.readValue(spec, resourceTypeIdentifier.identify(parseKind(kind)))
-    )
+    specMigrators
+      .migrate(parseKind(kind), objectMapper.readValue(spec, resourceTypeIdentifier.identify(parseKind(kind))))
```
luispollo, line 92: I hadn't paid too much attention to `ResourceTypeIdentifier` before, and find this a bit hard to read. Does it define the canonical kind this repository knows how to handle? If that's the case, could we perhaps rename the property for clarity? Either way, it might be useful to add a line comment describing it above.

robfletcher:
> This feels cumbersome to maintain. Would be nice to automate the creation of these beans. What would happen if you just had one `@Bean` function that returned the set of kinds?

I tried that and Spring ignores it. I am not quite done experimenting though. The best way to do it might be with a callback or bean factory of some kind.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ec2.kt
robfletcher: Yep, I think I'll stick an `EC2_` prefix on all the kinds

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
robfletcher: It's really just a function `ResourceKind -> Class<out ResourceSpec>` in other words, figure out what spec is the implementation for this kind. I feel like I can probably short circuit this entirely using the `SupportedKind` type, tho.

luispollo: Never mind. I see how it works from the tests now. The name is still a bit confusing to me, though. Maybe `ResourceSpecTypeMapper` or at least `ResourceSpecIdentifier`, since what it returns is a `ResourceSpec` sub-class?

robfletcher: I'm going to change it to `ResourceSpecIdentifier`. It also really does not need to be an interface any more. Changing to a `@Component` class where the `SupportedKind` instances just get injected.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -300,6 +275,9 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
             }
           }
         }
+
+        context("creation using an older schema") {
```
emjburns, line 279: Do you want to remove this section?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
```diff
@@ -0,0 +1,25 @@
...
+ * things like reading resources from the database, or parsing JSON.
+ */
+@Component
+class ResourceSpecIdentifier @Autowired constructor(
+  private val kinds: List<SupportedKind<*>>
+) {
+  fun identify(kind: ResourceKind): Class<out ResourceSpec> =
+    kinds.find { it.kind == kind }?.specClass ?: throw UnsupportedKind(kind)
+
+  /**
```
emjburns, line 21: I do not understand this comment

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/LegacySpecUpgradeTests.kt
```diff
@@ -0,0 +1,137 @@
...
+    val resourceTypeIdentifier = ResourceSpecIdentifier(v1, v2, v3)
+
+    val v1to2Migrator = object : SpecMigrator<SpecV1, SpecV2> {
+      override val input = v1
+      override val output = v2
+
+      override fun migrate(spec: SpecV1): SpecV2 =
+        SpecV2(spec.name, 1)
+    }
+    val v2to3Migrator = object : SpecMigrator<SpecV2, SpecV3> {
```
emjburns, line 71: Nice, this looks really easy to use.

robfletcher: Thanks. Once I got going on it, this was easy to implement too. I was overthinking it last week. @luispollo suggested to use the actual types as well which made it much nicer.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
robfletcher: Fixed

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
robfletcher: Oh yeah, this is a future PR

robfletcher: Fixed

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/config/EC2Config.kt
robfletcher: Couldn't get this to work in any way that was _less_ complicated than what I'd already done

PR #976, opened 2020-04-06
fix(metrics): record drift from last resource check

### LINKED ISSUE #950: Calculate drift metric based on last time resource was checked
The way that keel currently computes the drift metric, when the timeout fires the drift metric resets to zero. This is misleading, because it looks like drift has gone down to zero, even though the time since last resource check is still increasing.

It would be better to change `TelemetryListener` to calculate the drift metric based on the last time a resource was checked, instead of basing it on the start of an actuation loop.

(Suggested by @asher).

lorin:
Calculate the resource check drift metric (keel.resource.check.drift)
by using the last time resource check succeeded as the start time.

Previously, it was calculating based on the beginning of the actuation
loop, which would drop to zero when the loop restarted due to timeout.

Fixes #950

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
@@ -8,7 +8,6 @@ import com.netflix.spinnaker.keel.persistence.AgentLockRepository
 import com.netflix.spinnaker.keel.persistence.KeelRepository
 import com.netflix.spinnaker.keel.telemetry.ArtifactCheckTimedOut
 import com.netflix.spinnaker.keel.telemetry.EnvironmentsCheckTimedOut
-import com.netflix.spinnaker.keel.telemetry.ResourceCheckCompleted
 import com.netflix.spinnaker.keel.telemetry.ResourceCheckTimedOut
 import com.netflix.spinnaker.keel.telemetry.ResourceLoadFailed
 import java.time.Duration
@@ -61,7 +60,6 @@ class CheckScheduler(
   fun checkResources() {
     if (enabled.get()) {
       log.debug("Starting scheduled resource validation…")
-      publisher.publishEvent(ScheduledResourceCheckStarting)
 
       val job = launch {
         supervisorScope {
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
@@ -8,6 +8,7 @@ import com.netflix.spinnaker.keel.persistence.AgentLockRepository
 import com.netflix.spinnaker.keel.persistence.KeelRepository
 import com.netflix.spinnaker.keel.telemetry.ArtifactCheckTimedOut
 import com.netflix.spinnaker.keel.telemetry.EnvironmentsCheckTimedOut
+import com.netflix.spinnaker.keel.telemetry.ResourceCheckCompleted
 import com.netflix.spinnaker.keel.telemetry.ResourceCheckTimedOut
 import com.netflix.spinnaker.keel.telemetry.ResourceLoadFailed
 import java.time.Duration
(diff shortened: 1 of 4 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -79,7 +80,10 @@ class CheckScheduler(
                    * to prevent the cancellation of all coroutines under [job]
                    */
                   withTimeout(checkTimeout.toMillis()) {
-                    launch { resourceActuator.checkResource(it) }
+                    launch {
+                      resourceActuator.checkResource(it)
+                      publisher.publishEvent(ResourceCheckCompleted)
```
lorin, line 85: I created a new event that gets published after checkResource completes successfully.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryListener.kt
```diff
@@ -115,8 +114,8 @@ class TelemetryListener(
     ).safeIncrement()
   }
 
-  @EventListener(ScheduledResourceCheckStarting::class)
```
gal-yardeni, line 118: Is the new event in addition or instead `ScheduledResourceCheckStarting`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
emjburns: I think that because this is in a coroutine (`launch {` starts the coroutine in the background) this will publish right when the check starts. It would be worth verifying.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
```diff
@@ -11,6 +11,8 @@ data class ResourceCheckSkipped(
   val skipper: String = "unknown"
 ) : TelemetryEvent()
 
+object ResourceCheckCompleted : TelemetryEvent()
```
gal-yardeni, line 14: I wonder if `ResourceCheckCompleted` belongs here? is this a `TelemetryEvent` or just an object? if its part of the `TelemetryEvent`, it might be `data class` and not object, and it might need to hold more data, like resource info?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
lorin: @emjburns The `publishEvent` is called after the `checkResource` method returns.

emjburns: ahh 🤦‍♂ great.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryListener.kt
lorin: Instead of. Nothing is currently consuming `ScheduledResourceCheckStarting` events. Maybe I should just remove that class altogether.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
lorin: Good call: I'm going to move it to the file that contains other objects that are used as events but don't have any members.

lorin: Addressed in 5ad23592

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryListener.kt
lorin: Addressed in 5ad23592

lorin:
rebased against master

PR #977, opened 2020-04-06
fix(artifacts): add skipped summary to env summary

emjburns:
Turns out skipped versions weren't correctly getting an entry for the environment in `ArtifactVersionSummary`. This PR should address that. 

I added a skipped artifact into our `ArtifactServiceTests` to make sure that skipped info was tested. I also made the version numbers variables.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
@@ -32,7 +32,6 @@ abstract class ApplicationSummaryGenerationTests<T : ArtifactRepository> : JUnit
   data class Fixture<T : ArtifactRepository>(
     val subject: T
   ) {
-    // the artifact built off a feature branch
     val artifact = DebianArtifact(
       name = "keeldemo",
       deliveryConfigName = "my-manifest",
@@ -95,15 +94,15 @@ abstract class ApplicationSummaryGenerationTests<T : ArtifactRepository> : JUnit
       }
 
       test("skipped versions don't get a pending status in the next env") {
-        val summaries = subject.getEnvironmentSummaries(manifest).sortedBy { it.name }
+        val envSummaries = subject.getEnvironmentSummaries(manifest).sortedBy { it.name }
         expect {
-          that(summaries.size).isEqualTo(2)
-          that(summaries[0].artifacts.first().versions.current).isEqualTo(version2)
-          that(summaries[0].artifacts.first().versions.pending).isEmpty()
-          that(summaries[0].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
-          that(summaries[1].artifacts.first().versions.current).isEqualTo(version2)
-          that(summaries[1].artifacts.first().versions.pending).isEmpty()
-          that(summaries[1].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
+          that(envSummaries.size).isEqualTo(2)
+          that(envSummaries[0].artifacts.first().versions.current).isEqualTo(version2)
+          that(envSummaries[0].artifacts.first().versions.pending).isEmpty()
+          that(envSummaries[0].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
+          that(envSummaries[1].artifacts.first().versions.current).isEqualTo(version2)
+          that(envSummaries[1].artifacts.first().versions.pending).isEmpty()
+          that(envSummaries[1].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
         }
       }
     }
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
@@ -12,7 +12,7 @@ import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
 import java.time.Clock
 import strikt.api.expect
-import strikt.assertions.containsExactlyInAnyOrder
+import strikt.assertions.containsExactly
 import strikt.assertions.isEmpty
 import strikt.assertions.isEqualTo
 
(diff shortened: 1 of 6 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
```diff
@@ -485,14 +485,25 @@ class InMemoryArtifactRepository(
...
     val statuses = statusByEnvironment.getOrDefault(key, emptyMap<String, String>())
+
     val artifactDeployedVersions = deployedVersions[key]
       ?.sortedBy { (_, deployedAt) -> deployedAt } // ascending because we want to get the replacement deployment using `firstOrNull` below
     val deployedAt = artifactDeployedVersions?.find { (ver, _) -> ver == version }?.second
-    val (replacedBy, replacedAt) = artifactDeployedVersions?.firstOrNull { (ver, at) ->
+    var (replacedBy, replacedAt) = artifactDeployedVersions?.firstOrNull { (ver, at) ->
       ver != version && deployedAt != null && at.isAfter(deployedAt)
     }
       ?: Pair(null, null)
```
emjburns, line 496: This logic was present in the sql repository but not in the memory repository

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -115,45 +116,60 @@ class ApplicationService(
...
-            if (status == PromotionStatus.PENDING) {
-              ArtifactSummaryInEnvironment(
+            var artEnvSummary = when (status) {
+              PENDING -> ArtifactSummaryInEnvironment(
                 environment = environmentSummary.name,
                 version = version,
                 state = status.name.toLowerCase()
               )
-            } else {
-              repository.getArtifactSummaryInEnvironment(
```
emjburns, line 131: This was the problem - we weren't correctly handling the skipped status, because when we went to get the summary from the database it said it had a `PENDING` status.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -115,45 +116,60 @@ class ApplicationService(
     val environmentSummaries = getEnvironmentSummariesFor(application)
 
     return deliveryConfig.artifacts.map { artifact ->
```
emjburns, line 118: I simplified some of this logic a bit to make it less functional.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
```diff
@@ -95,15 +94,15 @@ abstract class ApplicationSummaryGenerationTests<T : ArtifactRepository> : JUnit
...
-          that(summaries[0].artifacts.first().versions.current).isEqualTo(version2)
-          that(summaries[0].artifacts.first().versions.pending).isEmpty()
-          that(summaries[0].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
-          that(summaries[1].artifacts.first().versions.current).isEqualTo(version2)
-          that(summaries[1].artifacts.first().versions.pending).isEmpty()
-          that(summaries[1].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
+          that(envSummaries.size).isEqualTo(2)
+          that(envSummaries[0].artifacts.first().versions.current).isEqualTo(version2)
+          that(envSummaries[0].artifacts.first().versions.pending).isEmpty()
+          that(envSummaries[0].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
```
luispollo, line 102: If the list has a single element you could just use `containsExactly()`, right?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
```diff
@@ -95,15 +94,15 @@ abstract class ApplicationSummaryGenerationTests<T : ArtifactRepository> : JUnit
...
+          that(envSummaries[1].artifacts.first().versions.current).isEqualTo(version2)
+          that(envSummaries[1].artifacts.first().versions.pending).isEmpty()
+          that(envSummaries[1].artifacts.first().versions.skipped).containsExactlyInAnyOrder(version1)
```
luispollo, line 105: Same comment.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -115,45 +116,60 @@ class ApplicationService(
...
+                if (potentialSummary == null || potentialSummary.state == "pending") {
+                  ArtifactSummaryInEnvironment(
+                    environment = environmentSummary.name,
+                    version = version,
+                    state = status.name.toLowerCase()
+                  )
+                } else {
+                  potentialSummary
+                }
+              }
```
luispollo, line 150: Struggling a bit to understand how this is different from the default case handled by the `else` block just below. Can you please explain?

emjburns: Sure - is the comment on it not clear? Suggestions on how I could rephrase?

"Skipped" artifacts are a little weird - sometimes when you get a summary for them from the database it returns "pending". That's because when we have no information about a version we return "pending" status (for example, the version was skipped in envA, which envB depends-on, we would have no info about if it was skipped in envB so we'd return a "pending" summary). 

But, if we already know it's skipped, that's not correct . That means that it the info doesn't correctly get added to the summary. So, first we check if the summary exists correctly in the database (this would be the case where there's a "skipped" entry for the env, like if there were two artifacts up for manual judgement but we only approved the latest version in that environment). If that doesn't exist, we give a summary with minimal information (like we do for pending).

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApplicationSummaryGenerationTests.kt
emjburns: Yes

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: An example of this situation is in the tests for this - `ApplicationServiceTests` `ln360` shows that in the first env we have data about which version replaced the version that was skipped, but the following two environments don't have that information.

PR #978, opened 2020-04-07
chore(pin): move pinning to the application controller

emjburns:
Move pin operations to the application controller so that they're consistent with the other things the UI is going to call. 

Will add auth once the big PR by @luispollo (#955) has been merged in.

I did not add tests because all these methods are just passthrough, and testing should be done on the artifact repository.

Part of https://github.com/spinnaker/keel/issues/979

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
@@ -3,7 +3,6 @@ package com.netflix.spinnaker.keel.core.api
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 
 data class EnvironmentArtifactPin(
-  val deliveryConfigName: String,
   val targetEnvironment: String,
   val reference: String,
   val type: String,
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -4,6 +4,7 @@ import com.netflix.frigga.ami.AppVersion
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.StatefulConstraint
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactType.deb
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactType.docker
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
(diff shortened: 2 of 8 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -73,6 +75,21 @@ class ApplicationService(
         judgedBy = user))
   }
 
+  fun pin(application: String, pin: EnvironmentArtifactPin, user: String) {
+    val config = repository.getDeliveryConfigForApplication(application)
```
gal-yardeni, line 79: can the `config` here and below be null/empty? are we throwing an exception in those cases?

robfletcher: Good question. Feels like a 404 would make sense in that case.

emjburns: I can, and `repository.getDeliveryConfigForApplication` will throw an exception (should be 404)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
```diff
@@ -3,7 +3,6 @@ package com.netflix.spinnaker.keel.core.api
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 
 data class EnvironmentArtifactPin(
-  val deliveryConfigName: String,
```
luispollo, line 6: This was the only way to look up the delivery config for authorization in the `@PreAuthorize` annotation:
https://github.com/spinnaker/keel/blob/ebad5af538bca45ec1d6a6c3a9c80946ec5ec53a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt#L78-L81

luispollo:
LGTM. Will approve after authz checks are added.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
```diff
@@ -3,7 +3,6 @@ package com.netflix.spinnaker.keel.core.api
...
   val targetEnvironment: String,
   val reference: String,
```
erikmunson, line 7: is this `reference` string the name given to the artifact in the delivery config? if so LMK and I will open up a GH issue for switching this API up a little bit as we don't have easy access to reference names and have been identifying versions by type + name + version

emjburns: Yes it is - let me switch it up now if it's easy

emjburns: it's not easy

emjburns:
@luispollo ported over the authz checks from your earlier PR!

PR #983, opened 2020-04-08
fix(api): Fix authorization for POST /delivery-configs

luispollo:
I had a bad reference in the SPeL expression, and forgot that the submitted delivery config is not in the database yet. 😝

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
@@ -32,8 +32,8 @@ class DeliveryConfigController(
     consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
     produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
   )
-  @PreAuthorize("""@authorizationSupport.hasApplicationPermission('WRITE', 'DELIVERY_CONFIG', #name)
-    and @authorizationSupport.hasServiceAccountAccess('DELIVERY_CONFIG', #name)"""
+  @PreAuthorize("""@authorizationSupport.hasApplicationPermission('WRITE', 'APPLICATION', #deliveryConfig.application)
+    and @authorizationSupport.hasServiceAccountAccess(#deliveryConfig.serviceAccount)"""
   )
   fun upsert(
     @RequestBody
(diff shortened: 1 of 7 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -32,8 +32,8 @@ class DeliveryConfigController(
     consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
     produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
   )
```
emjburns, line 34: I'm confused how this one api change led to so many test changes?

emjburns: Oh you just updated the function signature for the test helpers, right?

emjburns:
Ahh good spot. Tricky tricky.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
luispollo: I just realized when fixing this that, for tests, it doesn't matter where we get the service account from when we're mocking the calls to allow/deny authorization, so I just removed the parameter from the test helper. Sorry if it was confusing, but the tests are a bit simpler for it.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -32,8 +32,8 @@ class DeliveryConfigController(
...
-  @PreAuthorize("""@authorizationSupport.hasApplicationPermission('WRITE', 'DELIVERY_CONFIG', #name)
-    and @authorizationSupport.hasServiceAccountAccess('DELIVERY_CONFIG', #name)"""
+  @PreAuthorize("""@authorizationSupport.hasApplicationPermission('WRITE', 'APPLICATION', #deliveryConfig.application)
+    and @authorizationSupport.hasServiceAccountAccess(#deliveryConfig.serviceAccount)"""
```
luispollo, line 36: This was the fix: we can't read information about the delivery config from the database because it doesn't exist yet, so we get it from the submitted object.

Issue #951, opened 2020-03-31
Keel instances stop checking resources


lorin:
## Overview

Keel instances are regularly getting into a state where they stop checking resources. This is apparent when the `keel.resource.check.drift` metric starts to increase unbounded, until it reaches the timeout. Even once it reaches the timeout, it still doesn't make progress.

<details><summary>Example graph of an instance getting into a bad state</summary>

![image](https://user-images.githubusercontent.com/446305/78854921-63f2af80-79d7-11ea-9d39-780daae3b994.png)

You can see that all successful checks stop, with a timeout error happening every two minutes.

</details>

<details><summary>Atlas query</summary>

```
name,keel.resource.check.drift,:eq,
1000,:div,
2,:lw,
$name,:legend,

name,keel.resource.checked,:eq,
(,resourceState,),:by,
:stack,
:per-step,
44,:alpha,
$resourceState,:legend,
1,:axis,

:list,
(,nf.node,<INSTANCE ID GOES HERE>,:eq,:cq,),:each
```
</details>

## Current theory: bug in OkHttp library 4.4.1

My current hypothesis is that the issue is a bug in OkHttp version 4.4.1. Here's my reasoning:

### Thread dump: spinning in findHealthyConnection

Thread dump on a stuck instance shows that it's in the OkHttp library's [`findHealthyConnection`][1] method, which is an infinite loop.

[1]: https://github.com/square/okhttp/blob/parent-4.4.1/okhttp/src/main/java/okhttp3/internal/connection/ExchangeFinder.kt#L108

<details><summary>stack trace</summary>

```
"OkHttp https://clouddriver/..." #2130 prio=5 os_prio=0 tid=0x00007f596474c800 nid=0x23fc runnable [0x00007f5914dbe000]
   java.lang.Thread.State: RUNNABLE
	at okhttp3.internal.connection.ExchangeFinder.findHealthyConnection(ExchangeFinder.kt:108)
	at okhttp3.internal.connection.ExchangeFinder.find(ExchangeFinder.kt:76)
	at okhttp3.internal.connection.RealCall.initExchange$okhttp(RealCall.kt:245)
	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:32)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:82)
[...]

PR #985, opened 2020-04-08
fix(actuator): add coroutine id to help debug

lorin:
Create an ad-hoc coroutine id to help gather related log messages. This is a temporary change to make it easier to go through the log messages for debugging #951.

Context: I'm seeing different thread names that seem to be associated with the same coroutine. Since there's no guarantee of 1:1 mapping between threads and coroutines, I don't know if it's the same coroutine or a different one, this change will make it easier to figure that out.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -32,6 +32,7 @@ import com.netflix.spinnaker.keel.telemetry.ResourceCheckSkipped
 import com.netflix.spinnaker.keel.veto.VetoEnforcer
 import com.netflix.spinnaker.keel.veto.VetoResponse
 import java.time.Clock
+import java.util.UUID
 import kotlinx.coroutines.async
 import kotlinx.coroutines.supervisorScope
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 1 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -54,17 +55,18 @@ class ResourceActuator(
 
   suspend fun <T : ResourceSpec> checkResource(resource: Resource<T>) {
     withTracingContext(resource) {
+      val coid = UUID.randomUUID().toString() // coroutine id for log messages to help debug #951
       val id = resource.id
       val plugin = handlers.supporting(resource.kind)
 
       if (actuationPauser.isPaused(resource)) {
-        log.debug("Actuation for resource {} is paused, skipping checks", id)
+        log.debug("Actuation for resource {} is paused, skipping checks [$coid]", id)
```
emjburns, line 63: One thought - if you want to group all the messages that have a coid in them together, you might want each log message to be `[coid: $coid]` so it's easy to search for all of them. If you don't want to do that, then ignore me.

lorin: My plan is to find the `"Resource check for $id failed [$coid]" ` log message (which always appears when instances go bad), and then just search for the coid in that message to get the remaining relevant messages.

luispollo:
I think you can just get the coroutine IDs automatically added to the thread names by turning on `-Dkotlinx.coroutines.debug` in the JVM:
https://kotlinlang.org/docs/reference/coroutines/coroutine-context-and-dispatchers.html#debugging-coroutines-and-threads

Also, just to make sure you're aware, the behavior of coroutines sharing threads can be changed, if we feel like that's an approach worth pursuing as a fix: https://kotlinlang.org/docs/reference/coroutines/coroutine-context-and-dispatchers.html#dispatchers-and-threads

lorin:
@luispollo Thanks, I'm going to enable coroutine ids. However, even with the ability to turn on coroutine ids, I'd like to have unique ids for each invocation in the log to help me debug this.

PR #988, opened 2020-04-09
Compare base AMI versions properly

robfletcher:
Previously I was trying to compare the base AMI version from the current application AMI to the base AMI _name_ from `BaseImageCache` resulting in infinite rebaking.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
@@ -16,6 +16,7 @@ import com.netflix.spinnaker.keel.model.Job
 import com.netflix.spinnaker.keel.persistence.KeelRepository
 import com.netflix.spinnaker.keel.persistence.NoSuchArtifactException
 import com.netflix.spinnaker.keel.telemetry.ArtifactCheckSkipped
+import com.netflix.spinnaker.kork.exceptions.IntegrationException
 import org.slf4j.LoggerFactory
 import org.springframework.context.ApplicationEventPublisher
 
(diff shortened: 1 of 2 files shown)


keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -53,8 +54,25 @@ class ImageHandler(
...
+    return findBaseAmi(version)
+  }
+
+  private suspend fun findBaseAmi(baseImage: String): String {
+    return imageService.getLatestNamedImage(baseImage, "test")
+      ?.let { namedImage ->
+        val tags = namedImage
+          .tagsByImageId
+          .values
+          .first { it?.containsKey("base_ami_version") ?: false }
```
luispollo, line 68:
```suggestion
          .firstOrNull { it?.containsKey("base_ami_version") ?: false }
```
...otherwise you'd get a `NoSuchElementException` if you can't find the tag, right?

luispollo, line 68:
```suggestion
          .find { it?.containsKey("base_ami_version") ?: false }
```
...otherwise you'd get a `NoSuchElementException` if you can't find the tag, right?

luispollo: Actually, maybe rewrite the whole block like:
```kotlin
namedImage
  .tagsByImageId
  .values
  .find { it.containsKey("base_ami_version") }
  ?.map { it.get("base_ami_version") }
  ?: throw BaseAmiNotFound(baseImage)
```

PR #991, opened 2020-04-09
Make BaseImageCache map to base_ami_version rather than just imageName

robfletcher:
Obviously, this will require equivalent change in `keel-nflx` where we'll use `ImageService.findBaseAmiVersion` to go from the image name we get from the Bakery's fast property to the actual `base_ami_version` tag on that image.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/BaseImageCache.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/BaseImageCache.kt
@@ -6,8 +6,9 @@ interface BaseImageCache {
   /**
    * @param os the desired base image operating system.
    * @param label the desired base image label.
-   * @return the id of the requested base image, if it exists.
+   * @return the base AMI version (i.e. the "appversion" of the base AMI) of the requested base
+   * image, if it exists.
    * @throws UnknownBaseImage if there is no known base image for [os] and [label].
    */
-  fun getBaseImage(os: String, label: BaseLabel): String
+  fun getBaseAmiVersion(os: String, label: BaseLabel): String
 }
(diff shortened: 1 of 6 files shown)


keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/BaseImageCache.kt
```diff
@@ -6,8 +6,9 @@ interface BaseImageCache {
   /**
    * @param os the desired base image operating system.
    * @param label the desired base image label.
-   * @return the id of the requested base image, if it exists.
+   * @return the base AMI version (i.e. the "appversion" of the base AMI) of the requested base
+   * image, if it exists.
    * @throws UnknownBaseImage if there is no known base image for [os] and [label].
    */
-  fun getBaseImage(os: String, label: BaseLabel): String
+  fun getBaseAmiVersion(os: String, label: BaseLabel): String
```
robfletcher, line 13: Clarified the name of the method

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -179,20 +179,17 @@ class ImageService(
         amiMatches(allTags, buildHost, buildName, buildNumber)
       }
 
-  suspend fun findBaseAmi(baseImage: String): String {
-    return cloudDriverService.namedImages(DEFAULT_SERVICE_ACCOUNT, baseImage, "test")
+  suspend fun findBaseAmiVersion(baseImageName: String): String {
+    return cloudDriverService.namedImages(DEFAULT_SERVICE_ACCOUNT, baseImageName, "test")
```
robfletcher, line 183: I felt like this name was a little clearer

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -179,20 +179,17 @@ class ImageService(
...
       .lastOrNull()
       ?.let { namedImage ->
```
emjburns, line 185: You just simplified here, no change in logic, correct?

robfletcher: Correct. Although as @luispollo pointed out on my PR yesterday I think there was a risk of the `first` method throwing an exception in the unlikely event that there was no matching element, whereas `find` will return `null`

PR #992, opened 2020-04-09
chore(bake): add a metric to alert on for when we bake too much

emjburns:
Tagging a bake counter with the appversion so that we can see if we're re-baking the same app version over and over again. I'm going to set up an alert for baking the same appversion like... more than 10 times or something.


lorin:
Wooo, metrics!

PR #997, opened 2020-04-10
Don't keep re-trying the same bake

### LINKED ISSUE #856: Fix unnecessary (no-op) bakes
From Michael:
> After an initial bake, there are lots of redundant bakes which fail after a few seconds. I believe this is not a new issue, but wanted to flag it anyway.

Discussion notes:
> The problem is we're waiting to see the image reflected in the cache, and it takes a while because we have lots of images. This is made worse when swabbie isn't running. Hurts the quality of the experience, and could overwhelm the bakery with lots of users.

robfletcher:
Closes #856


keel-bakery-history-sql/keel-bakery-history-sql.gradle.kts
```diff
@@ -0,0 +1,20 @@
...
+  testImplementation("io.strikt:strikt-core")
+  testImplementation(project(":keel-spring-test-support"))
+  testImplementation(project(":keel-test"))
+  testImplementation(project(":keel-core-test"))
+
+  testImplementation(project(":keel-web")) {
+    // avoid circular dependency which breaks Liquibase
+    exclude(module = "keel-bakery-history-sql")
+  }
+}
```
robfletcher, line 20: New module because I didn't want `keel-sql` to depend on `keel-bakery-plugin`

keel-bakery-history-sql/src/test/kotlin/com/netflix/spinnaker/keel/bakehistory/sql/spring/SpringStartupTests.kt
```diff
@@ -0,0 +1,42 @@
...
+  webEnvironment = MOCK,
+  properties = [
+    "keel.plugins.bakery.enabled=true",
+    "sql.enabled=true",
+    "orca.enabled=true",
+    "orca.base-url=http://localhost:8087",
+    "echo.base-url=http://localhost:8083",
+    "clouddriver.enabled=true",
+    "clouddriver.base-url=http://localhost:8085",
+    "igor.base-url=http://localhost:8084",
```
robfletcher, line 27: I'd like to figure out if I can avoid having to set all this up

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeHistory.kt
```diff
@@ -0,0 +1,6 @@
+package com.netflix.spinnaker.keel.bakery.artifact
+
+interface BakeHistory {
+  fun contains(appVersion: String, baseAmiVersion: String): Boolean
+  fun add(appVersion: String, baseAmiVersion: String, regions: Collection<String>, taskId: String)
+}
```
robfletcher, line 6: Simple interface to record and query whether we've baked a particular appversion / base_ami_version combination

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -36,24 +37,32 @@ class ImageHandler(
...
+          if (bakeHistory.contains(latestArtifactVersion, latestBaseAmiVersion)) {
+            log.warn("Artifact version {} and base AMI version {} were baked previously", latestArtifactVersion, latestBaseAmiVersion)
+            publisher.publishEvent(RecurrentBakeDetected(latestArtifactVersion, latestBaseAmiVersion))
+          } else {
+            log.info("Versions or regions for {} differ, launching bake", artifact.name)
+            launchBake(artifact, latestArtifactVersion)
+              .also { tasks ->
+                bakeHistory.add(latestArtifactVersion, latestBaseAmiVersion, artifact.vmOptions.regions, tasks.first().id)
+              }
+          }
```
robfletcher, line 65: Don't bake if we've done it before

keel-bakery-history-sql/keel-bakery-history-sql.gradle.kts
robfletcher: Wondering if I should instead have `keel-bakery-plugin` depend on `keel-sql`. Not sure. I'm not entirely happy with adding this new module.

keel-bakery-history-sql/src/test/kotlin/com/netflix/spinnaker/keel/bakehistory/sql/spring/SpringStartupTests.kt
luispollo: We could just use an `application.properties` or `keel.yml` for tests, right?
You can use `@TestPropertySource` to specify the location: https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/test/context/TestPropertySource.html

robfletcher:
> I understand what you're trying to do here, but don't we have that information in orca already? We even have the orca task IDs stored in the `task_tracking` table (maybe missing some columns for cross-referencing). My feeling is that we could just ask orca about the bake tasks for a particular artifact version. It could also very well be that I'm missing the point entirely. 😄

Can we rely on Orca not having cleaned up those tasks? I'm kind of looking at this as an interim fix anyway. It seems like the reason we keep re-baking is because we don't really have a way to handle missing regions on an AMI. I'm going to do some digging in to that but for now this should stop us spamming bake tasks.

luispollo:
> Can we rely on Orca not having cleaned up those tasks?

I think we only care about the short term, since we'll only be baking the latest approved version, no? In any case, our cleanup config looks like this right now (i.e. orchestration tasks live forever, apparently):
```yaml
pollers:
  oldPipelineCleanup:
    enabled: false
    thresholdDays: 90
  pipelineCleanup:
    enabled: false
  orchestrationCleanup:
    enabled: false
```

luispollo:
P.S. Not trying to discourage you, just wanted to understand if we had that information available elsewhere already.

robfletcher:
> P.S. Not trying to discourage you, just wanted to understand if we had that information available elsewhere already.

I don't think it's accessible in any very convenient way. I think the change to Orca to enable us to access it would maybe be bigger than this change is.

keel-bakery-history-sql/src/test/kotlin/com/netflix/spinnaker/keel/bakehistory/sql/spring/SpringStartupTests.kt
robfletcher: Damn, I had totally missed that this was why I wasn't having the same problems in `keel-sql` or `keel-web`. Thanks. Good spot.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -36,24 +37,32 @@ class ImageHandler(
...
+        val versionsDiffer = image?.appVersion != latestArtifactVersion || image?.baseAmiVersion != latestBaseAmiVersion
         val regionsDiffer = !(image?.regions ?: emptySet()).containsAll(artifact.vmOptions.regions)
         if (image == null || versionsDiffer || regionsDiffer) {
-          log.info("Versions or regions for {} differ, launching bake", artifact.name)
-          launchBake(artifact, latestVersion)
+          if (bakeHistory.contains(latestArtifactVersion, latestBaseAmiVersion)) {
+            log.warn("Artifact version {} and base AMI version {} were baked previously", latestArtifactVersion, latestBaseAmiVersion)
+            publisher.publishEvent(RecurrentBakeDetected(latestArtifactVersion, latestBaseAmiVersion))
+          } else {
+            log.info("Versions or regions for {} differ, launching bake", artifact.name)
```
emjburns, line 60: can you make this log message tell which of the two differ?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -36,24 +37,32 @@ class ImageHandler(
...
 
-        val versionsDiffer = image?.appVersion != latestVersion || image?.baseAmiVersion != latestBaseImageVersion
+        val versionsDiffer = image?.appVersion != latestArtifactVersion || image?.baseAmiVersion != latestBaseAmiVersion
         val regionsDiffer = !(image?.regions ?: emptySet()).containsAll(artifact.vmOptions.regions)
         if (image == null || versionsDiffer || regionsDiffer) {
-          log.info("Versions or regions for {} differ, launching bake", artifact.name)
-          launchBake(artifact, latestVersion)
+          if (bakeHistory.contains(latestArtifactVersion, latestBaseAmiVersion)) {
+            log.warn("Artifact version {} and base AMI version {} were baked previously", latestArtifactVersion, latestBaseAmiVersion)
+            publisher.publishEvent(RecurrentBakeDetected(latestArtifactVersion, latestBaseAmiVersion))
```
emjburns, line 58: Could you add an event listener into https://github.com/spinnaker/keel/blob/master/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/telemetry/BakeryTelemetryListener.kt that increments a counter and tags by something like `artifactVersion+baseAmiVersion` (just one tag so it's easy to graph)?

keel-bakery-history-sql/keel-bakery-history-sql.gradle.kts
emjburns: Is it reasonable to keep all sql in one module? or does that present a lot of problems?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
emjburns: Actually, could you just use the differ's `toJson` view to print the actual difference? I know we're not using it here, but it might help us get to the bottom of why we re-bake. As I've been debugging this week I've really wanted to know what we expect vs what exists.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
robfletcher: We probably want regions on there too

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/telemetry/BakeryTelemetryListener.kt
```diff
@@ -42,10 +43,21 @@ class BakeryTelemetryListener(private val spectator: Registry) {
...
 
+  @EventListener(RecurrentBakeDetected::class)
+  fun onRecurrentBakeDetected(event: RecurrentBakeDetected) {
+    spectator.counter(
+      RECURRENT_BAKE_DETECTED_ID,
+      listOf(
+        BasicTag("versions", "${event.appVersion}+${event.baseAmiVersion}")
+      )
+    )
+  }
```
robfletcher, line 54: @emjburns new telemetry

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
robfletcher: I was just thinking the same thing

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -36,30 +39,33 @@ class ImageHandler(
...
+          } else {
+            log.info("Image for {} delta: {}", artifact.name, diff.toDebug())
+          }
+
+          if (diffFingerprintRepository.seen(artifact.name, diff)) {
+            log.warn("Artifact version {} and base AMI version {} were baked previously", latestArtifactVersion, latestBaseAmiVersion)
+            publisher.publishEvent(RecurrentBakeDetected(latestArtifactVersion, latestBaseAmiVersion))
+          } else {
+            launchBake(artifact, latestArtifactVersion)
+            diffFingerprintRepository.store(artifact.name, diff)
```
luispollo, line 61: Great idea!

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -36,30 +39,33 @@ class ImageHandler(
...
+        val diff = DefaultResourceDiff(desired, current)
+
+        if (diff.hasChanges()) {
+          if (current == null) {
+            log.info("No AMI found for {}", artifact.name)
+          } else {
+            log.info("Image for {} delta: {}", artifact.name, diff.toDebug())
+          }
+
+          if (diffFingerprintRepository.seen(artifact.name, diff)) {
```
emjburns, line 56: You should hash this diff, since that's what the diff fingerprint repository stores

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
luispollo: Do these ever get cleaned up from the database?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -36,30 +39,33 @@ class ImageHandler(
...
+            diffFingerprintRepository.store(artifact.name, diff)
+          }
         } else {
           log.debug("Existing image for {} is up-to-date", artifact.name)
         }
 
-        if (image != null && regionsDiffer) {
-          publisher.publishEvent(ImageRegionMismatchDetected(image, artifact.vmOptions.regions))
+        if (current != null && diff.affectedRootPropertyNames == setOf("regions")) {
+          publisher.publishEvent(ImageRegionMismatchDetected(current, artifact.vmOptions.regions))
```
luispollo, line 68: Great instrumentation. We'd been missing data on the frequency of this forever. 👏

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDiffFingerprintRepository.kt
```diff
@@ -82,4 +82,15 @@ class SqlDiffFingerprintRepository(
     }
     return count ?: 0
   }
+
+  override fun seen(resourceId: String, diff: ResourceDiff<*>): Boolean =
+    sqlRetry.withRetry(READ) {
+      jooq
```
lorin, line 88: Note that jooq has a [fetchExists](https://www.jooq.org/javadoc/latest/org.jooq/org/jooq/DSLContext.html#fetchExists(org.jooq.Select)) method, which you could use to rewrite the query like this:

```
jooq.fetchExists(
  selectFrom(DIFF_FINGERPRINT)
  .where(DIFF_FINGERPRINT.RESOURCE_ID.eq(resourceId))
  .and(DIFF_FINGERPRINT.HASH.eq(diff.generateHash())))
```

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
robfletcher: It does that internally in the method, same as `store` does

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDiffFingerprintRepository.kt
robfletcher: oh nice, that's  much better

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintEvaluator.kt
```diff
@@ -54,6 +55,7 @@ import org.springframework.stereotype.Component
  *    failureCancelsRunningRegions: false
  */
 @Component
+@ConditionalOnProperty("orca.enabled")
```
lorin, line 58: Is this related to the bake retry issue?

robfletcher: Ah, no. I should back this out

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
robfletcher: There's only ever one per resource but no

keel-sql/src/main/resources/db/changelog/20200410-rename-diff-fingerprint-column.yml
```diff
@@ -0,0 +1,16 @@
+databaseChangeLog:
+- changeSet:
+    id: rename-diff-fingerprint-column
+    author: fletch
+    changes:
+    - renameColumn:
+        tableName: diff_fingerprint
+        oldColumnName: resource_id
+        newColumnName: entity_id
+        columnDataType: varchar(255)
```
robfletcher, line 10: @emjburns did the column name change

emjburns: Thanks!

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
emjburns: LOL right. duh.

PR #1000, opened 2020-04-10
fix(api): Better 403 error messages

### LINKED ISSUE #998: Better import error message for 403s from keel
When a 403 is returned by keel when `POST /delivery-config` is called, i.e. the user is forbidden from submitting the delivery config given their permissions, the stage execution is showing an unhelpful (and actually inaccurate) message:
image

Looking at the execution JSON we see it was a 403 from keel:
```json
    {
      "id": "01E5G75ZNRKTKCM8Z7FS7VCMD8",
      "refId": "2",
      "type": "importDeliveryConfig",
      "name": "Import Delivery Config",
      "startTime": 1586461081410,
      "endTime": 1586461082132,
      "status": "TERMINAL",
      "context": {
        "manifest": "spinnaker.yml",
        "error": "Non-retryable HTTP response 403 received from downstream service: HTTP 403 https://keel-main.us-west-2.spinnaker.mgmt.netflix.net/delivery-configs/: 403 "
      }
    }
```
That error message in the stage execution could be much better in this case. I added a special case for 401 coming from igor, but not 403 coming from keel. Need to add special handling for that case as well.

luispollo:
Unfortunately, since the `@PreAuthorize` behavior in Spring is just to check boolean-returning functions, there's no way to propagate the actual permission error up the stack. Authorizationt errors are all turned into "Access is denied", which is not very informative.

I could change the permission check methods to throw my own `AccessDeniedException`, but that would short-circuit Spring's `AccessDecisionManager`, which didn't seem like a good idea, so my approach here is just to use a generic message that's better than the default.

Closes #998

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
@@ -85,8 +85,8 @@ class ExceptionHandler(
     log.error(e.message)
     return ApiError(
       if (e.message == null || e.message == "Access is denied") {
-        "Access denied. Please make sure you have access to the service account specified in your delivery config, " +
-          "and that the service account has access to the target application and cloud account."
+        "Access denied. Please make sure you have access to the service account specified in your delivery config. " +
+          "If you do have access, check that the service account has access to this application along with all the cloud accounts included in the delivery config."
       } else {
         e.message!!
       }
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
@@ -23,10 +23,12 @@ import java.time.format.DateTimeParseException
 import org.slf4j.LoggerFactory
 import org.springframework.http.HttpStatus.BAD_REQUEST
 import org.springframework.http.HttpStatus.CONFLICT
+import org.springframework.http.HttpStatus.FORBIDDEN
 import org.springframework.http.HttpStatus.NOT_FOUND
 import org.springframework.http.HttpStatus.UNPROCESSABLE_ENTITY
 import org.springframework.http.converter.HttpMessageConversionException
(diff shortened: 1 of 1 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -77,6 +79,20 @@ class ExceptionHandler(
...
+  @ExceptionHandler(AccessDeniedException::class)
+  @ResponseStatus(FORBIDDEN)
+  fun onAccessDenied(e: AccessDeniedException): ApiError {
+    log.error(e.message)
+    return ApiError(
+      if (e.message == null || e.message == "Access is denied") {
+        "Access denied. Please make sure you have access to the service account specified in your delivery config, " +
+          "and that the service account has access to the target application and cloud account."
+      } else {
+        e.message!!
```
luispollo, line 91: This allows messages that are _not_ the default (which may come from using the `AuthorizationSupport.checkXyZ()`-type functions) to be propagated.

luispollo:
@erikmunson I changed the wording on the error box label in the import delivery config stage execution details so that it accommodates for errors that are not parsing JSON: https://github.com/spinnaker/deck/pull/8154

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -77,6 +79,20 @@ class ExceptionHandler(
...
   }
 
+  @ExceptionHandler(AccessDeniedException::class)
+  @ResponseStatus(FORBIDDEN)
+  fun onAccessDenied(e: AccessDeniedException): ApiError {
+    log.error(e.message)
+    return ApiError(
+      if (e.message == null || e.message == "Access is denied") {
+        "Access denied. Please make sure you have access to the service account specified in your delivery config, " +
+          "and that the service account has access to the target application and cloud account."
```
erikmunson, line 89:
```suggestion
        "Access denied. Please make sure you have access to the service account specified in your delivery config. " +
          "If you do have access, check that the service account has access to this application along with all the cloud accounts included in the delivery config."
```

PR #1001, opened 2020-04-10
fix(resource): tweak thread model, add more debug logging

lorin:
With the recent changes to coroutine logging, the logs reveal that, for issue #951, the code is timing out in the [resolve](https://github.com/spinnaker/keel/blob/87028e7638fefcdd9dfbee0c449042f932b6970f/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt#L177) function:

<details><summary>stack trace</summary>

```
kotlinx.coroutines.TimeoutCancellationException: Timed out waiting for 120000 ms
	at (Coroutine boundary.()
	at kotlinx.coroutines.DeferredCoroutine.await$suspendImpl(Builders.common.kt:99)
	at com.netflix.spinnaker.keel.actuation.ResourceActuator$resolve$2.invokeSuspend(ResourceActuator.kt:195)
	at com.netflix.spinnaker.keel.actuation.ResourceActuator$checkResource$2.invokeSuspend(ResourceActuator.kt:75)
	at com.netflix.spinnaker.keel.logging.TracingSupport$Companion.withTracingContext(TracingSupport.kt:39)
```

</details>

I suspect the timeout is happening in the `current` function call, but there isn't enough evidence in the logs to be certain.

This PR makes a few changes to that method:

* adds debug log statements to narrow down where the code is timing out
* splits two `await` invocations across two separate lines
* change the threading model of the async call that wraps the `current` function invocation.

The threading model change is just a stab in the dark, to see if there's some strange coroutine/thread race condition interaction that's biting us that may not happen if the coroutine to thread mapping changes.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -34,6 +34,7 @@ import com.netflix.spinnaker.keel.veto.VetoResponse
 import java.time.Clock
 import java.util.UUID
 import kotlinx.coroutines.async
+import kotlinx.coroutines.newSingleThreadContext
 import kotlinx.coroutines.supervisorScope
 import org.slf4j.LoggerFactory
 import org.springframework.context.ApplicationEventPublisher
@@ -57,6 +58,7 @@ class ResourceActuator(
     withTracingContext(resource) {
       val coid = UUID.randomUUID().toString() // coroutine id for log messages to help debug #951
       val id = resource.id
+      log.debug("checkResource $id [$coid]")
       val plugin = handlers.supporting(resource.kind)
 
       if (actuationPauser.isPaused(resource)) {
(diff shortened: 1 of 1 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -174,25 +176,37 @@ class ResourceActuator(
...
-  private suspend fun <T : Any> ResourceHandler<*, T>.resolve(resource: Resource<out ResourceSpec>): Pair<T, T?> =
+  private suspend fun <T : Any> ResourceHandler<*, T>.resolve(resource: Resource<out ResourceSpec>, coid: String): Pair<T, T?> =
     supervisorScope {
       val desired = async {
         try {
-          desired(resource)
+          log.debug("before desired: ${resource.id}: [$coid]")
+          val result = desired(resource)
+          log.debug("after desired: ${resource.id}: [$coid]")
+          result
```
gal-yardeni, line 186: here and below, what is the meaning of the stand alone `result`?

lorin: AIUI, blocks passed to an `async` have to evaluate to a value:

```
val desired = async {
```


Before my change, the block passed to async would evaluate to the return value of `desired(resource)`.

However, if I just added a log statement afterwards, the block would evaluate to the return value of the log statement. So, I needed to add a line afterwards that would evaluate to the return value of `desired` to preserve the behavior.

The `result` variable is just a temp for doing this.

PR #1007, opened 2020-04-13
feat(export): Export delivery config from pipelines

luispollo:
**:bulb: TL;DR:** for a considerable chunk of the applications at Netflix, which happen to coincide with those we're going after for Beta, I believe we can simplify the migration process by exporting a delivery config from their (relatively simple) pipelines.

----

As I was reviewing our quarterly plans and goals for Managed Delivery, in particular around our desire to simplify the onboarding process for existing applications, it occurred to me that, instead of focusing on resources for the export process, since the goal is really to migrate teams from _pipelines_ to delivery configs, we could look at what an export process based on pipelines might look like. While we had assumed it would be very hard to figure out what the user’s intent was with pipelines, I had a hunch it might not be too far-fetched for applications with “simple” pipelines. 

I went back to the script and pipeline data that @michaelgalloway gathered a few months back to check some assumptions, starting with what our definition of “simple” was. In a nutshell, a simple pipeline as defined there is one with a linear delivery flow (i.e. no branches in the stage graph) that is composed of a stage that creates or identifies an artifact (a bake or a “find image”-type stage), and one or more subsequent deploy stages, sometimes with a manual judgement before deployments. There are several patterns the script matches against, but they all basically boil down to variations of this pattern.

Once I looked at the data, I was convinced it was possible to export an entire delivery config for these types of application based solely on their pipeline configurations, without the need to query the actual state of cloud resources. Starting from pipelines actually makes the process much easier to codify and reason about.

This PR introduces an initial implementation of this idea via the `/export/{application}` API, as a way to get feedback from the team and hopefully from select users.

The following are supported:
- Extracting Debian artifact information from bake stages.
- Extracting Debian artifact information indirectly from findImage stages. In this case, since the image must have been baked by some pipeline, we find the pipeline which deploys the cluster referenced in the find image config, then find the bake stage in that pipeline to extract the artifact information.
- Extracting environments with EC2 cluster resources from deploy stages.
- Extracting `manual-judgment` constraints from manual judgment stages.
- Extracting `manual-judgment` and `depends-on` constraints from pipeline triggers. 

These are _not_ currently supported:
- Deploy stages with more than one server group definition. I’m unsure how common this case is, but we could look at the data to find out. 
- Extracting artifact information with two or more levels of indirection via findImage stages. I assumed this to be a corner case.
- Load balancer resources. 
- Security group resources (although we do populate the SGs as cluster dependencies).
- Titus clusters and Docker artifacts. 
- Time window constraints (the data is available, it’s just work).
- Notifications (the data is available, it's just work).

I would love feedback on whether this is a good idea overall, and on specifics at your leisure. I would be thrilled to continue working on this if it makes sense. 

To test the API locally, just point your local keel to production front50 (keel uses strictly read-only APIs and checks authorization with fiat).

TODO:
- [ ] Tests
- [ ] Addressing any of the gaps above that are "must haves" for a v1


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ExportService.kt
```diff
@@ -0,0 +1,324 @@
...
+    val constraints = mutableSetOf<Constraint>()
+
+    val trigger = pipeline.triggers.also {
+      if (it.size > 1) {
+        log.warn("Pipeline has more than 1 trigger. Using only the first.")
+      }
+    }.firstOrNull()
+
+    if (trigger == null || !trigger.enabled) {
+      // if there's no trigger, the pipeline is triggered manually, i.e. the equivalent of a manual judgment
```
michaelgalloway, line 237: In most simple cases I've seen, production environments should have two constraints: depends-on (testing), and manual-judgement.  It looks like from this if/else block, it'd only be configured to have the one constraint.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
```diff
@@ -33,7 +33,7 @@ class MemoryCloudDriverCache(
 
   private val securityGroupSummariesById = Caffeine.newBuilder()
     .maximumSize(1000)
-    .expireAfterWrite(1, MINUTES)
+    .expireAfterWrite(5, MINUTES)
```
luispollo, line 36: I tried to increase this because, as far as I can tell testing locally, the cache is having no effect on calls to be security groups by ID. Still trying to debug that...

keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Pipeline.kt
```diff
@@ -0,0 +1,69 @@
...
+  val id: String,
+  val disabled: Boolean = false,
+  val fromTemplate: Boolean = false,
+  val triggers: List<Trigger> = emptyList(),
+  stages: List<Stage> = emptyList(),
+  val lastModifiedBy: String? = null,
+  val updateTs: Long? = null
+) {
+  val stages: List<Stage> = stages
+    get() = field.sortedBy { if (it.requisiteStageRefIds.isEmpty()) "" else it.requisiteStageRefIds.first() }
```
luispollo, line 14: This sorts the stages by cross-stage dependencies.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ExportService.kt
```diff
@@ -0,0 +1,324 @@
...
+
+@Component
+class ExportService(
+  private val handlers: List<ResourceHandler<*, *>>,
+  private val front50Service: Front50Service
+) {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  companion object {
+    val EXPORTABLE_PIPELINE_PATTERNS = listOf(
```
luispollo, line 32: Thank you @michaelgalloway for your research and coming up with this list of patterns! 🙏

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ExportService.kt
```diff
@@ -0,0 +1,324 @@
...
+  ): DebianArtifact {
+    return when (stage) {
+      is BakeStage -> stage.artifact
+      is FindImageStage -> run {
+        // Look for another pipeline in the app that has a deploy stage for the cluster referenced in findImage.
+        // The other pipeline must have either baked or found the image itself.
+        // TODO: implement the case for the other pipeline also having a findImage, not a bake.
+        val (otherPipeline, deploy) = pipelines.findPipelineWithDeployForCluster(stage)
+          ?: error("Deploy stage with cluster definition matching find image stage '${stage.name}' not found.")
+        log.debug("Found other pipeline with deploy stage for cluster ${stage.moniker}")
```
luispollo, line 150:
```suggestion
        log.debug("Found other pipeline with deploy stage for cluster ${cluster.moniker}")
```

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ExportService.kt
```diff
@@ -0,0 +1,324 @@
...
+
+        val bake = otherPipeline.findUpstreamBake(deploy)
+          ?: error("Upstream bake stage from deploy stage '${deploy.name}' not found.")
+        log.debug("Found other pipeline with deploy stage for cluster ${stage.moniker}")
```
luispollo, line 154:
```suggestion
        log.debug("Found other pipeline with deploy stage for cluster ${cluster.moniker}")
```

luispollo:
### Sample pipelines and output
Given the following pipelines:
image
image

...keel generates the following export result:
```yaml
deliveryConfig:
  name: "myapp"
  application: "myapp"
  serviceAccount: "delivery-engineering@netflix.com"
  artifacts:
    - name: "myapp-server"
      reference: "myapp-server"
      vmOptions:
        baseLabel: "RELEASE"
        baseOs: "xenial"
        regions:
          - "us-east-1"
        storeType: "EBS"
      statuses:
        - "RELEASE"
      type: "deb"
  environments:
    - name: "test"
      resources:
        - metadata: {}
          kind: "ec2/cluster@v1"
          spec:
            moniker:
              app: "myapp"
              stack: "test"
            imageProvider:
              reference: "myapp-server"
            deployWith:
              strategy: "red-black"
              rollbackOnFailure: true
              resizePreviousToZero: false
              maxServerGroups: 2
              delayBeforeDisable: "PT0S"
              delayBeforeScaleDown: "PT0S"
            locations:
              account: "test"
              subnet: "internal (vpc0)"
              vpc: "vpc0"
              regions:
                - name: "us-east-1"
                  availabilityZones:
                    - "us-east-1c"
                    - "us-east-1d"
                    - "us-east-1e"
            launchConfiguration:
              instanceType: "m5.2xlarge"
              ebsOptimized: true
              iamRole: "BaseIAMRole"
              keyPair: "<redacted>"
              instanceMonitoring: false
            capacity:
              min: 2
              max: 2
              desired: 2
            dependencies:
              securityGroupNames:
                - "<redacted>"
                - "<redacted>"
                - "myapp"
            health:
              cooldown: "PT10S"
              healthCheckType: "EC2"
              terminationPolicies:
                - "OldestInstance"
      constraints: []
      notifications: []
    - name: "prod-eu-west-1"
      resources:
        - metadata: {}
          kind: "ec2/cluster@v1"
          spec:
            moniker:
              app: "myapp"
              stack: "prod"
            imageProvider:
              reference: "myapp-server"
            deployWith:
              strategy: "red-black"
              rollbackOnFailure: true
              resizePreviousToZero: false
              maxServerGroups: 2
              delayBeforeDisable: "PT0S"
              delayBeforeScaleDown: "PT0S"
            locations:
              account: "prod"
              subnet: "internal (vpc0)"
              vpc: "vpc0"
              regions:
                - name: "eu-west-1"
                  availabilityZones:
                    - "eu-west-1b"
                    - "eu-west-1c"
                    - "eu-west-1a"
            launchConfiguration:
              instanceType: "m5.2xlarge"
              ebsOptimized: true
              iamRole: "BaseIAMRole"
              keyPair: "nf-prod-keypair-a"
              instanceMonitoring: false
            capacity:
              min: 2
              max: 2
              desired: 2
            dependencies:
              securityGroupNames:
                - "<redacted>"
                - "<redacted>"
                - "myapp"
            health:
              cooldown: "PT10S"
              healthCheckType: "EC2"
              terminationPolicies:
                - "OldestInstance"
      constraints:
        - timeout: "PT168H"
          type: "manual-judgement"
      notifications: []
    - name: "prod-us-east-1"
      resources:
        - metadata: {}
          kind: "ec2/cluster@v1"
          spec:
            moniker:
              app: "myapp"
              stack: "prod"
            imageProvider:
              reference: "myapp-server"
            deployWith:
              strategy: "red-black"
              rollbackOnFailure: true
              resizePreviousToZero: false
              maxServerGroups: 2
              delayBeforeDisable: "PT0S"
              delayBeforeScaleDown: "PT0S"
            locations:
              account: "prod"
              subnet: "internal (vpc0)"
              vpc: "vpc0"
              regions:
                - name: "us-east-1"
                  availabilityZones:
                    - "us-east-1c"
                    - "us-east-1d"
                    - "us-east-1e"
            launchConfiguration:
              instanceType: "m5.2xlarge"
              ebsOptimized: true
              iamRole: "BaseIAMRole"
              keyPair: "nf-prod-keypair-a"
              instanceMonitoring: false
            capacity:
              min: 2
              max: 2
              desired: 2
            dependencies:
              securityGroupNames:
                - "<redacted>"
                - "<redacted>"
                - "myapp"
            health:
              cooldown: "PT10S"
              healthCheckType: "EC2"
              terminationPolicies:
                - "OldestInstance"
      constraints:
        - timeout: "PT168H"
          type: "manual-judgement"
      notifications: []
    - name: "prod-us-west-2"
      resources:
        - metadata: {}
          kind: "ec2/cluster@v1"
          spec:
            moniker:
              app: "myapp"
              stack: "prod"
            imageProvider:
              reference: "myapp-server"
            deployWith:
              strategy: "red-black"
              rollbackOnFailure: true
              resizePreviousToZero: false
              maxServerGroups: 2
              delayBeforeDisable: "PT0S"
              delayBeforeScaleDown: "PT0S"
            locations:
              account: "prod"
              subnet: "internal (vpc0)"
              vpc: "vpc0"
              regions:
                - name: "us-west-2"
                  availabilityZones:
                    - "us-west-2a"
                    - "us-west-2b"
                    - "us-west-2c"
            launchConfiguration:
              instanceType: "m5.2xlarge"
              ebsOptimized: true
              iamRole: "BaseIAMRole"
              keyPair: "nf-prod-keypair-a"
              instanceMonitoring: false
            capacity:
              min: 2
              max: 2
              desired: 2
            dependencies:
              securityGroupNames:
                - "<redacted>"
                - "<redacted>"
                - "myapp"
            health:
              cooldown: "PT10S"
              healthCheckType: "EC2"
              terminationPolicies:
                - "OldestInstance"
      constraints:
        - timeout: "PT168H"
          type: "manual-judgement"
      notifications: []
pipelines:
  exported:
    - id: "e3f3443f-a985-4b14-a9c5-3317220763f2"
      name: "aws-deploy-to-prod"
      environments:
        - "prod-eu-west-1"
        - "prod-us-west-2"
        - "prod-us-east-1"
    - id: "239bb180-9344-4cf6-b11a-de8b518c6d52"
      name: "aws-deploy-to-test"
      environments:
        - "test"
  skipped: []
```

luispollo:
> The other thing to note about this pipeline is that it has multiple deploy stages, one for each region where the app owner wanted to deploy the artifact. We map these as multiple environments in the delivery config, each named after the corresponding region.

I just realized these could just be different cluster resources underneath a single `prod` environment. Will fix.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ExportService.kt
luispollo: Good point. I'm already having to find the upstream environment below for pipeline triggers, but could do something similar for pipelines with no triggers, starting from their first deploy stage.

erikmunson:
This is a very cool idea, and if it's possible to deliver it to customers in a way that meets our expectations for reliability and UX I would advocate for investing time in it. With that, I have some questions and concerns around what it might look like to productize the overall approach which I'll leave here for you to consider.

The thing that's top of mind for me here is that the approach is — mostly by necessity — very brittle and explicitly does not work in a sizable percentage of cases. While the analysis Michael did was useful as a way to get a rough sense of the target population at Netflix, I'm skeptical of using that data to make decisions (except in the narrow case of evaluating potential early adopters through deeper, thorough human spot-checking), and _even more_ skeptical of using it programmatically in any way. Two concerns:
- I worry we're falling into a selection bias trap where we look for a very rigid pattern, then pick some examples that look good and try to extrapolate them to every other seemingly similar example
- I worry that due to the selection bias problem we're underestimating how much variation and unpredictability is hiding in those "simple" pipelines, and overestimating the percentage of applications we'll migrate in Q2/Q3 that could be solved by automation like this due to additional unforeseen nuances in the way those apps are setup

Anecdotally, I rarely see pipelines in the wild that fit these patterns exactly, even though they're likely doing nearly the same thing a supported pipeline is trying to do. Does that mean that the data we have is not accurate? I trust its accuracy in the sense that it's accurately measuring what it's measuring. Do I trust that our actual adopters will end up being in line with the data from Michael's analysis, or that making tactical decisions based on that data will lead to predictable outcomes? I don't think any of us has the ability to answer with confidence _precisely because it's impossible to truly understand pipeline intent reliably through automation_. Any analysis or anecdotal impression (like mine) is going to suffer from the exact same selection problem.

What comes to mind given those concerns is a set of questions about what it would look like to _actually_ use this thing as a first-class part of the onboarding experience (e.g. part of the newt export flow), not just an API endpoint our team can hit as an internal convenience:
1. What is the experience like for the times where someone has no pipelines that fit this limited set of structures?
2. _Much_ more importantly in my eyes: what is the experience like when it _appears_ the export worked fine (i.e. endpoint returns a delivery config), and yet something actually isn't right because there's nuance, intent, or human-level context that didn't make it into the automation? While the resource export flow is simple and exceedingly predictable, I imagine that there are many cases where this endpoint might behave in a way that's mismatched with the expectation of someone using it for the first time e.g.:
  - There are a bunch of pipelines that do very similar things, maybe several of which match the patterns this endpoint handles, but only one of which would be the 'right' one to export from the human's perspective. Or worse! There are actually _multiple_ pipelines that the human would want to pick and choose context-specific behavior from
- There's some out of band way that the pipeline gets triggered, or the triggering relationship is otherwise misleading
- We pick a _slightly_ wrong artifact/bake options, but not one that's wrong enough to call immediate attention to itself
- The automation "sees" a small subsection of the pipelines that someone uses to deliver (I suspect this might happen with the setups that use a bunch of find image tricks, which in my experience are shockingly commonplace), exports the things it thinks are needed, again doesn't catch the eye of the human, and then inexplicably the manifest that was exported doesn't work as expected or misses part of the delivery workflow entirely

3. What does it look like to have _two_ distinct onboarding flows, with different levels of predictability and stability?
4. What impact does all of the above have on our ability to support, document, and ruthlessly simplify the end-to-end onboarding experience?

For me #2 is the most critical and worrying set of questions to understand before investing time in productizing. I've worked on revenue-critical onboarding flows for more time than I care to think about and my experience has been that — no matter the context — a predictable workflow with very little room for error that takes a bit more upfront effort will usually get more people successfully down the adoption funnel than a less predictable and more automated approach, even when the unpredictable approach can have 2x or 3x the effectiveness in the cases that work as intended. When I've supported unpredictable onboarding components in the past the biggest pain came from the sour taste users had in their mouth after a horrendously bad experience (i.e. trying, thinking it was working, things going wrong, contacting us, us not understanding why things didn't work, us discovering it was a confusing bug with the onboarding logic and needing to do manual damage control, etc.) that could have been avoided with just a little extra upfront configuration work on their part.

That was an eye-full, so I'll repeat again: I will be the first to champion this work if we can mitigate the potentially profound UX risks, and I have no intent of shutting the idea down. My only desire is to inject a little bit of the hard-won experience I have with onboarding in past roles and the skepticism I bring about how easily we can programmatically understand pipelines — if it was easy to programmatically derive pipeline intent the idea of Managed Delivery wouldn't be nearly as exciting! 

Consider what it would really look like to _fully_ productize this, and whether the end result would
1. Have a positive ROI after the required hardening work when compared with spending time on other work and leveraging the current resource export flow
2. Meet our very high UX standards — if you'd like input or additional brainpower around how to do that, I'm here to help!

luispollo:
Hi @erikmunson, thank you so much for taking the time to write such incredibly thoughtful comments and questions. I don't feel discouraged but challenged, although I have to say it's clear to me after reading your comments that there are many angles I did not consider when coming up with this idea, and context I lacked mostly from a lack of experience seeing a large number of pipelines at Netflix.

The other thing that's absolutely clear to me is that, _if_ we were to continue to invest in this, this API would be but a small part of a larger onboarding flow that would necessarily involve human intervention from users, and likely consulting with our team. I think part of the opportunity of onboarding to Managed Delivery is _precisely_ that perhaps the pipelines you have cobbled together to achieve your delivery goals can be rethought in terms of much simpler declarative delivery flows, which we would miss out on altogether if it was at all possible to fully automate the process.

I think it will be more effective to discuss in more detail in our upcoming meeting, but I did want to make sure I clarify a nuance you may have potentially missed about the approach:

> * There are a bunch of pipelines that do very similar things, maybe several of which match the patterns this endpoint handles, but only one of which would be the 'right' one to export from the human's perspective. Or worse! There are actually _multiple_ pipelines that the human would want to pick and choose context-specific behavior from

To be clear, the algorithm does not look at pipelines individually and in isolation, but rather the entire set of pipelines for an application (which is why the API takes the application name as a parameter, not a pipeline). That doesn't mean some of the pipelines are not "wrong" to include, but I just wanted to clarify that the core of the algorithm is an attempt to find the relationships between the environments represented by _all_ the pipelines in the app.

lorin:
I really like the idea of having support for this functionality, but more as a "power tool" that's only accessible by experts. I wouldn't want to expose it to end-users, at least not initially.

luispollo:
After discussing within the team, we've decided not to pursue this idea further for the time being, for the following main reasons:
1. Yes, there is manual intervention involved in migrating an app from pipelines to a delivery config today, but the fact that it creates an opportunity for teams to stop and rethink their delivery flows in terms of Managed Delivery (potentially even simplifying things greatly) is a hugely valuable thing we don't want to give up.
2. Even as a "power tool" for us to use internally to expedite the process of migrating an app while hand-holding a team through onboarding, we would run the risk of following into the trap of (1) above.
3. If this were to be exposed to end-users as a product feature, the risk of exporting a delivery config that looks "right" but totally misses out on nuances in the pipeline configs is very high, and could severely damage our image and trust from the teams if/when it went wrong.

@lorin Pointed out that there might be a use for a similar but also different kind of tool, to be used internally only, which would help us cross check a delivery config generated manually/via newt against the existing pipelines to look for those missed nuances.

@michaelgalloway FYI.

PR #1009, opened 2020-04-13
fix(retrofit): Call propagate to get request ID, set user origin

robzienert:
Adding `AuthenticatedRequest.propagate` to interceptor. This command will propagate a bunch of headers, but also automates creation of a request ID. Also added a user origin, which is somewhat important as well - if left unset, the rest of Spinnaker will assume this is a gate API-originating request, which it isn't.

### DIFF
--- a/keel-retrofit/src/main/kotlin/com/netflix/spinnaker/keel/retrofit/AuthorizationHeadersInterceptor.kt
+++ b/keel-retrofit/src/main/kotlin/com/netflix/spinnaker/keel/retrofit/AuthorizationHeadersInterceptor.kt
@@ -18,13 +18,16 @@ class AuthorizationHeadersInterceptor(private val fiatPermissionEvaluator: FiatP
   override fun intercept(chain: Chain): Response {
     var request = chain.request()
     request.header(Header.USER.header)?.also { user ->
-      AuthenticatedRequest.allowAnonymous {
-        val accounts = fiatPermissionEvaluator.getPermission(user).accounts.joinToString(",") { it.name }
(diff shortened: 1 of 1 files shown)


keel-retrofit/src/main/kotlin/com/netflix/spinnaker/keel/retrofit/AuthorizationHeadersInterceptor.kt
```diff
@@ -18,13 +18,16 @@ class AuthorizationHeadersInterceptor(private val fiatPermissionEvaluator: FiatP
...
     request.header(Header.USER.header)?.also { user ->
-      AuthenticatedRequest.allowAnonymous {
-        val accounts = fiatPermissionEvaluator.getPermission(user).accounts.joinToString(",") { it.name }
-        log.trace("Adding X-SPINNAKER-ACCOUNTS: $accounts to ${request.method} ${request.url}")
-        request = request
-          .newBuilder()
-          .addHeader(Header.ACCOUNTS.header, accounts)
-          .build()
+      AuthenticatedRequest.setUserOrigin("keel")
+      AuthenticatedRequest.propagate {
```
luispollo, line 22: Glad you were able to hook this into this interceptor. It didn't exist until last week. 😄

luispollo:
Hey @robzienert, it looks like the addition of `AuthenticatedRequest.propagate` caused a regression in adding the `X-SPINNAKER-ACCOUNTS` header. As far as I can see, the closure within is never called. Any thoughts?

luispollo:
Never mind, I got it. MDC with coroutines doesn't work without extra shenanigans.

Issue #295, opened 2019-05-19
ResourceDeltaResolved is not consistently added


robfletcher:
It's only added right after a `ResourceDeltaDetected` but other events (e.g. a `ResourceUpdate`) could happen first.

spinnakerbot:
This issue hasn't been updated in 111 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

PR #1012, opened 2020-04-13
fix(resourceEvent): emit ResourceDeltaResolved when actually needed

gal-yardeni:
fixes https://github.com/spinnaker/keel/issues/295.

We observe some event mismatch behavior in the resource history API - for some resources, the event chain is `Drift detected` --> `valid` without `ResourceDeltaResolved` prior to `valid` (see screenshot).

Solution: prior to emitting `ResourceDeltaResolved` event, we check what was the resource last event? 
`ResourceActuationLaunched`? -- do nothing and wait for task completion.
`ResourceTaskSucceeded` or `ResourceTaskFailed`? --> emit the `resolved` event (as we have a clear status and the delta is resolved).
`ResourceDeltaDetected`? but no actuation was launched, emit the `resolved` event (as probably we resolved it without taking any explicit action).

Screen Shot 2020-04-13 at 10 03 21 AM
Screen Shot 2020-04-13 at 10 03 58 AM

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -19,6 +19,7 @@ import com.netflix.spinnaker.keel.events.ResourceCheckUnresolvable
 import com.netflix.spinnaker.keel.events.ResourceDeltaDetected
 import com.netflix.spinnaker.keel.events.ResourceDeltaResolved
 import com.netflix.spinnaker.keel.events.ResourceMissing
+import com.netflix.spinnaker.keel.events.ResourceTaskFailed
 import com.netflix.spinnaker.keel.events.ResourceTaskSucceeded
 import com.netflix.spinnaker.keel.events.ResourceValid
 import com.netflix.spinnaker.keel.logging.TracingSupport.Companion.withTracingContext
@@ -93,16 +94,16 @@ class ResourceActuator(
           if (response.vetoArtifact && resource.spec is VersionedArtifactProvider) {
             try {
               val versionedArtifact = when (desired) {
-                      is Map<*, *> -> {
-                        if (desired.size > 0) {
-                          (desired as Map<String, VersionedArtifactProvider>).values.first()
-                        } else {
-                          null
-                        }
-                      }
-                      is VersionedArtifactProvider -> desired
-                      else -> null
-                    }?.completeVersionedArtifactOrNull()
+                is Map<*, *> -> {
+                  if (desired.size > 0) {
+                    (desired as Map<String, VersionedArtifactProvider>).values.first()
+                  } else {
+                    null
+                  }
+                }
+                is VersionedArtifactProvider -> desired
+                else -> null
+              }?.completeVersionedArtifactOrNull()
 
               if (versionedArtifact != null) {
                 with(versionedArtifact) {
@@ -155,14 +156,16 @@ class ResourceActuator(
               }
           }
           else -> {
+            log.info("Resource {} is valid [$coid]", id)
             val lastEvent = resourceRepository.lastEvent(id)
-            if (lastEvent is ResourceDeltaDetected || lastEvent is ResourceActuationLaunched ||
-              lastEvent is ResourceTaskSucceeded) {
-              log.info("Resource {} delta resolved [$coid]", id)
-              publisher.publishEvent(ResourceDeltaResolved(resource, clock))
-            } else {
-              log.info("Resource {} is valid [$coid]", id)
-              publisher.publishEvent(ResourceValid(resource, clock))
+            when (lastEvent) {
+              is ResourceActuationLaunched -> log.debug("waiting for actuating task to be completed") // do nothing and wait
+              is ResourceDeltaDetected, is ResourceTaskSucceeded, is ResourceTaskFailed -> {
+                // if a delta was detected and a task wasn't launched, the delta is resolved
+                // if a task was launched and it completed, either successfully or not, the delta is resolved
+                publisher.publishEvent(ResourceDeltaResolved(resource, clock))
+              }
+              else -> publisher.publishEvent(ResourceValid(resource, clock))
             }
           }
         }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -19,6 +19,7 @@ import com.netflix.spinnaker.keel.events.ResourceCheckUnresolvable
 import com.netflix.spinnaker.keel.events.ResourceDeltaDetected
 import com.netflix.spinnaker.keel.events.ResourceDeltaResolved
 import com.netflix.spinnaker.keel.events.ResourceMissing
+import com.netflix.spinnaker.keel.events.ResourceTaskSucceeded
 import com.netflix.spinnaker.keel.events.ResourceValid
 import com.netflix.spinnaker.keel.logging.TracingSupport.Companion.withTracingContext
 import com.netflix.spinnaker.keel.pause.ActuationPauser
@@ -92,18 +93,16 @@ class ResourceActuator(
           if (response.vetoArtifact && resource.spec is VersionedArtifactProvider) {
             try {
               val versionedArtifact = when (desired) {
-                is Map<*, *> -> {
-                  if (desired.size > 0) {
-                    (desired as Map<String, VersionedArtifactProvider>).values.first()
-                  } else {
-                    null
-                  }
-                }
-                is VersionedArtifactProvider -> desired
-                else -> null
-              }?.let {
-                it.completeVersionedArtifactOrNull()
-              }
+                      is Map<*, *> -> {
+                        if (desired.size > 0) {
+                          (desired as Map<String, VersionedArtifactProvider>).values.first()
+                        } else {
+                          null
+                        }
+                      }
+                      is VersionedArtifactProvider -> desired
+                      else -> null
+                    }?.completeVersionedArtifactOrNull()
 
               if (versionedArtifact != null) {
                 with(versionedArtifact) {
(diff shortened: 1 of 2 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -92,18 +93,16 @@ class ResourceActuator(
...
-                is Map<*, *> -> {
-                  if (desired.size > 0) {
-                    (desired as Map<String, VersionedArtifactProvider>).values.first()
-                  } else {
-                    null
-                  }
-                }
-                is VersionedArtifactProvider -> desired
-                else -> null
-              }?.let {
```
gal-yardeni, line 104: the compiler suggested removing the redundant `let` here

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -156,12 +155,13 @@ class ResourceActuator(
               }
           }
           else -> {
-            log.info("Resource {} is valid [$coid]", id)
-            // TODO: not sure this logic belongs here
             val lastEvent = resourceRepository.lastEvent(id)
-            if (lastEvent is ResourceDeltaDetected || lastEvent is ResourceActuationLaunched) {
+            if (lastEvent is ResourceDeltaDetected || lastEvent is ResourceActuationLaunched ||
+              lastEvent is ResourceTaskSucceeded) {
```
emjburns, line 160: I don't think that a task succeeding means a resource delta is resolved - for example, a bake task can succeed but definitely not resolve a delta.

emjburns: oh wait, this is in the section where we've already determined the resource is valid, right?

gal-yardeni: Its a bit tricky - as for my understanding, this is the only place that we are publishing the `ResourceDeltaResolved` event. This is being determined only after checking if the resource should be created/ updated, so I guess the answer to your question is yes. My only fear with this solution is that we might send duplicated `ResourceDeltaResolved`, so I'm wondering if there's some logic in the UI to handle it (or should we handle it here in some way?)

luispollo:
> `ResourceTaskSucceeded` or `ResourceTaskFailed`? --> emit the `resolved` event (as we have a clear status and the delta is resolved).

If the task failed, how would the delta be resolved?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -95,6 +97,9 @@ internal class ResourceActuatorTests : JUnit5Minutests {
       resourceRepository.dropAll()
     }
 
+    context("resource events are ") {
```
emjburns, line 100: extra?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -157,12 +157,15 @@ class ResourceActuator(
...
-            if (lastEvent is ResourceDeltaDetected || lastEvent is ResourceActuationLaunched) {
-              publisher.publishEvent(ResourceDeltaResolved(resource, clock))
-            } else {
-              publisher.publishEvent(ResourceValid(resource, clock))
+            when (lastEvent) {
+              is ResourceActuationLaunched -> log.debug("waiting for actuating task to be completed") // do nothing and wait
+              is ResourceDeltaDetected, is ResourceTaskSucceeded, is ResourceTaskFailed -> {
+                // if a delta was detected and a task wasn't launched, the delta is resolved
+                // if a task was launched and it completed, either successfully or not, the delta is resolved
+                publisher.publishEvent(ResourceDeltaResolved(resource, clock))
```
luispollo, line 166: I asked this question in another comment, but I'm confused as to how a delta would have been resolved if the task launched to resolve the delta has failed...

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -95,6 +97,9 @@ internal class ResourceActuatorTests : JUnit5Minutests {
...
+    }
```
luispollo, line 101: Are you planning to add any tests here?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -196,6 +201,42 @@ internal class ResourceActuatorTests : JUnit5Minutests {
               }
             }
 
+            context("when the a actuation launched event exists in resource history, check other events before publishing ResourceDeltaResolved") {
```
luispollo, line 204:
```suggestion
            context("when the last event in resource history is an actuation launched, check other events before publishing ResourceDeltaResolved") {
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -196,6 +201,42 @@ internal class ResourceActuatorTests : JUnit5Minutests {
...
+                resourceRepository.appendHistory(ResourceActuationLaunched(resource, plugin1.name, emptyList()))
+                runBlocking {
+                  subject.checkResource(resource)
+                }
+              }
+              test("ResourceDeltaResolved was not published since the task is still running") {
+                verify(exactly = 0) { publisher.publishEvent(ofType<ResourceDeltaResolved>()) }
+              }
+
+              context("the task was finished successfully") {
```
luispollo, line 215: This should be outdented to be at the same level as the context for "last event is actuation launched", right?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -196,6 +201,42 @@ internal class ResourceActuatorTests : JUnit5Minutests {
...
+                before {
+                  resourceRepository.appendHistory(ResourceTaskSucceeded(resource, emptyList()))
+                  runBlocking {
+                    subject.checkResource(resource)
+                  }
+                }
+                test("a telemetry event is published") {
```
luispollo, line 222:
```suggestion
                test("a resource delta resolved event is published") {
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: So this case is only hit when there is no delta (aka the resource is in a good state). It would be weird if this happened (maybe the task timed out? maybe it failed on some other step like scaling down an old server group?) but regardless there was no diff detected, so we should output the right event.

All the other events will still stay in the task history.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -196,6 +201,42 @@ internal class ResourceActuatorTests : JUnit5Minutests {
...
+                  verify { publisher.publishEvent(ofType<ResourceDeltaResolved>()) }
+                }
+              }
+
+              context("the task was finished with an error") {
```
luispollo, line 227: Same comment on nesting of the tests.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: Sorry, I don't follow. If a task was launched and failed it's because there was a delta to resolve, no? Why do you say this is only hit when there is no delta?

erikmunson: Funny enough, Emily's example case has happened to me multiple times! Both times the scenario has been:
1. Delta detected
2. Deploy task is launched
3. Server group comes up healthy and fine, but something about the deploy made the task itself fail (one time it was emily's timeout example, the other time it was something wonky with the disable step)
4. There's now a failed task in the history events
5. Keel comes back around and sees things are back to the right state, and considers things resolved, it adds a `ResourceValid` event

The relationship between orca's view of the world (did this unit of work go the way i expected) and keel's view of the world (is the current state of this thing the same as how i'd ideally like it to be) are very different in areas like this.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
gal-yardeni: oops yes!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
gal-yardeni: @luispollo the `ResourceTaskFailed` event is connected to orca tasks. When there's a diff, and there's an actuation, the task can either succeed or failed. In both cases. the diff is resolved.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
gal-yardeni: discussed privately.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
gal-yardeni: same

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: Thanks, @erikmunson. That clarified it for me! @gal-yardeni, that's not entirely true. If an actuation task is launched (say, for a deployment), and the deployment itself fails, the task will be marked as failed, and the diff will _not_ have been resolved because current state will differ from desired state. However, in this case, the block of code you're in is only reached when we've already compared and there is no diff, so in this specific spot it's safe to make that assumption.

I think what threw me off was reading the comment just above these lines, without realizing that we were in this `else` block that's only reached when `current == desired`.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -196,6 +198,42 @@ internal class ResourceActuatorTests : JUnit5Minutests {
...
+                before {
+                  resourceRepository.appendHistory(ResourceTaskFailed(resource, "error", emptyList()))
+                  runBlocking {
+                    subject.checkResource(resource)
+                  }
+                }
+                test("a telemetry event is published") {
```
luispollo, line 231:
```suggestion
                test("a resource delta resolved event is published") {
```

PR #1014, opened 2020-04-13
feat(artifacts): add pinned info, only use reference

emjburns:
I'm doing several things in this PR:
- Change pinning to only need artifact reference, not reference and type. Artifact reference is unique. Artifact name + type is not unique.
- Clean up some methods around ^
- Add pinning data into the summary view: both in the environment summary and the artifact summary

A request to pin an env will now look like:

`POST /application/{appName}/pin`
```json
{
	"targetEnvironment": "testing",
	"reference" : "my-artifact",
	"version" : "master-h13.73d1d93"
}
```
(you don't need the `type` in the body anymore, will update gate in a future PR)

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
@@ -14,6 +14,7 @@ import com.netflix.spinnaker.keel.api.artifacts.VirtualMachineOptions
 import com.netflix.spinnaker.keel.core.api.ArtifactVersionStatus
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactPin
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVetoes
+import com.netflix.spinnaker.keel.core.api.Pinned
 import com.netflix.spinnaker.keel.core.api.PinnedEnvironment
 import com.netflix.spinnaker.time.MutableClock
 import dev.minutest.junit.JUnit5Minutests
@@ -502,7 +503,7 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
           )
           expectThat(envArtifactSummary)
             .isNotNull()
-            .get { pinned }
+            .get { isPinned }
             .isFalse()
         }
 
@@ -518,7 +519,7 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
           }
 
           test("pinned version cannot be vetoed") {
-            expectThat(subject.markAsVetoedIn(manifest, artifact2, pin1.version!!, pin1.targetEnvironment))
+            expectThat(subject.markAsVetoedIn(manifest, artifact2, pin1.version, pin1.targetEnvironment))
               .isFalse()
           }
 
@@ -532,7 +533,7 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
                 artifact = artifact2,
                 version = version4,
                 pinnedBy = pin1.pinnedBy,
-                pinnedAt = clock.millis(),
+                pinnedAt = clock.instant(),
                 comment = pin1.comment
               )))
           }
@@ -544,10 +545,11 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
               artifactReference = artifact2.reference,
               version = version4
             )
-            expectThat(envArtifactSummary)
-              .isNotNull()
-              .get { pinned }
-              .isTrue()
+            expect {
+              that(envArtifactSummary).isNotNull()
+              that(envArtifactSummary?.isPinned).isTrue()
+              that(envArtifactSummary?.pinned).isEqualTo(Pinned(by = pin1.pinnedBy, at = clock.instant(), comment = pin1.comment))
+            }
           }
         }
       }
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
@@ -14,6 +14,7 @@ import com.netflix.spinnaker.keel.api.artifacts.VirtualMachineOptions
 import com.netflix.spinnaker.keel.core.api.ArtifactVersionStatus
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactPin
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVetoes
+import com.netflix.spinnaker.keel.core.api.PinnedEnvironment
 import com.netflix.spinnaker.time.MutableClock
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
@@ -32,6 +33,7 @@ import strikt.assertions.isEmpty
 import strikt.assertions.isEqualTo
 import strikt.assertions.isFalse
 import strikt.assertions.isNotEqualTo
+import strikt.assertions.isNotNull
 import strikt.assertions.isNull
 import strikt.assertions.isTrue
 import strikt.assertions.succeeded
@@ -87,7 +89,6 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
     val pin1 = EnvironmentArtifactPin(
       targetEnvironment = environment2.name, // staging
       reference = artifact2.reference,
-      type = artifact2.type.name,
       version = version4, // the older release build
       pinnedBy = "keel@spinnaker",
       comment = "fnord")
@@ -492,17 +493,62 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
             .isNotEqualTo(pin1.version)
         }
 
-        test("latestVersionApprovedIn prefers a pinned version over the latest approved version") {
-          subject.pinEnvironment(manifest, pin1)
-          expectThat(subject.latestVersionApprovedIn(manifest, artifact2, environment2.name))
-            .isEqualTo(version4)
-            .isEqualTo(pin1.version)
+        test("get env artifact version shows that artifact is not pinned") {
+          val envArtifactSummary = subject.getArtifactSummaryInEnvironment(
+            deliveryConfig = manifest,
+            environmentName = pin1.targetEnvironment,
+            artifactReference = artifact2.reference,
+            version = version4
+          )
+          expectThat(envArtifactSummary)
+            .isNotNull()
+            .get { pinned }
+            .isFalse()
         }
 
-        test("pinned version cannot be vetoed") {
-          subject.pinEnvironment(manifest, pin1)
-          expectThat(subject.markAsVetoedIn(manifest, artifact2, pin1.version!!, pin1.targetEnvironment))
-            .isFalse()
+        context("once pinned") {
+          before {
+            subject.pinEnvironment(manifest, pin1)
+          }
+
+          test("latestVersionApprovedIn prefers a pinned version over the latest approved version") {
+            expectThat(subject.latestVersionApprovedIn(manifest, artifact2, environment2.name))
+              .isEqualTo(version4)
+              .isEqualTo(pin1.version)
+          }
+
+          test("pinned version cannot be vetoed") {
+            expectThat(subject.markAsVetoedIn(manifest, artifact2, pin1.version!!, pin1.targetEnvironment))
+              .isFalse()
+          }
+
+          test("getting pinned environments shows the pin") {
+            val pins = subject.getPinnedEnvironments(manifest)
+            expectThat(pins)
+              .hasSize(1)
+              .isEqualTo(listOf(PinnedEnvironment(
+                deliveryConfigName = manifest.name,
+                targetEnvironment = pin1.targetEnvironment,
+                artifact = artifact2,
+                version = version4,
+                pinnedBy = pin1.pinnedBy,
+                pinnedAt = clock.millis(),
+                comment = pin1.comment
+              )))
+          }
+
+          test("get env artifact version shows that artifact is pinned") {
+            val envArtifactSummary = subject.getArtifactSummaryInEnvironment(
+              deliveryConfig = manifest,
+              environmentName = pin1.targetEnvironment,
+              artifactReference = artifact2.reference,
+              version = version4
+            )
+            expectThat(envArtifactSummary)
+              .isNotNull()
+              .get { pinned }
+              .isTrue()
+          }
         }
       }
     }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -57,9 +57,16 @@ data class ArtifactSummaryInEnvironment(
   val deployedAt: Instant? = null,
   val replacedAt: Instant? = null,
   val replacedBy: String? = null,
-  val pinned: Boolean = false,
+  val isPinned: Boolean = false,
+  val pinned: Pinned? = null,
   val statefulConstraints: List<StatefulConstraintSummary> = emptyList(),
   val statelessConstraints: List<StatelessConstraintSummary> = emptyList()
+)
+
+data class Pinned(
+  val at: Instant,
+  val by: String?,
+  val comment: String?
 )
 
 @JsonInclude(Include.NON_NULL)
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -57,6 +57,7 @@ data class ArtifactSummaryInEnvironment(
   val deployedAt: Instant? = null,
   val replacedAt: Instant? = null,
   val replacedBy: String? = null,
+  val pinned: Boolean = false,
   val statefulConstraints: List<StatefulConstraintSummary> = emptyList(),
   val statelessConstraints: List<StatelessConstraintSummary> = emptyList()
 )
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
@@ -5,7 +5,6 @@ import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 data class EnvironmentArtifactPin(
   val targetEnvironment: String,
   val reference: String,
-  val type: String,
   val version: String?,
   val pinnedBy: String?,
   val comment: String?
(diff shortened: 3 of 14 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
```diff
@@ -57,6 +57,7 @@ data class ArtifactSummaryInEnvironment(
   val deployedAt: Instant? = null,
   val replacedAt: Instant? = null,
   val replacedBy: String? = null,
```
emjburns, line 59: Adding a field to the artifact summary to indicate whether this version is pinned in this environment

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
```diff
@@ -15,7 +14,10 @@ data class PinnedEnvironment(
   val deliveryConfigName: String,
   val targetEnvironment: String,
   val artifact: DeliveryArtifact,
-  val version: String
+  val version: String,
+  val pinnedBy: String?,
+  val pinnedAt: Long?,
+  val comment: String?
 )
```
emjburns, line 21: Adding more data here so that we could maybe display this info (not now, I definitely didn't expose this in the UI at all) (but @erikmunson maybe you could tell me if you even want this data about a pinned artifact and if so where you want it)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
```diff
@@ -51,7 +51,8 @@ data class ArtifactVersions(
   val type: ArtifactType,
   val reference: String,
   val statuses: Set<ArtifactStatus>,
-  val versions: ArtifactVersionStatus
+  val versions: ArtifactVersionStatus,
+  val pinnedVersion: String?
```
emjburns, line 55: Added an optional field on the environment summary that will show you the pinned version, if it exists

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -867,28 +873,13 @@ class SqlArtifactRepository(
   override fun getArtifactSummaryInEnvironment(
     deliveryConfig: DeliveryConfig,
     environmentName: String,
-    artifactName: String,
-    artifactType: ArtifactType,
+    artifactReference: String,
     version: String
   ): ArtifactSummaryInEnvironment? {
     return sqlRetry.withRetry(READ) {
```
emjburns, line 880: There were 3 sql queries in this function. I removed these two, and added that sql into the third query. Then I added a nested query to get the pinned information. I'd love ideas on how I could remove the nested query.

Basically, I want to have a `true` or `false` if there's a row that matches. This will be used to populate the `pinned` field.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -492,17 +493,62 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
...
+          }
+
+          test("latestVersionApprovedIn prefers a pinned version over the latest approved version") {
+            expectThat(subject.latestVersionApprovedIn(manifest, artifact2, environment2.name))
+              .isEqualTo(version4)
+              .isEqualTo(pin1.version)
+          }
+
+          test("pinned version cannot be vetoed") {
+            expectThat(subject.markAsVetoedIn(manifest, artifact2, pin1.version!!, pin1.targetEnvironment))
```
luispollo, line 521: I was surprised `EnvironmentArtifactPin.version` could be `null`, since the whole point is to pin a specific version. Do you know if that was intentional? Might be worth fixing in this PR if not.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
```diff
@@ -15,7 +14,10 @@ data class PinnedEnvironment(
   val deliveryConfigName: String,
   val targetEnvironment: String,
   val artifact: DeliveryArtifact,
-  val version: String
+  val version: String,
+  val pinnedBy: String?,
+  val pinnedAt: Long?,
```
luispollo, line 19: This should be an `Instant?`, no?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -24,7 +24,7 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
 
   fun get(name: String, type: ArtifactType, reference: String, deliveryConfigName: String): DeliveryArtifact
 
-  fun get(deliveryConfigName: String, reference: String, type: ArtifactType): DeliveryArtifact
+  fun get(deliveryConfigName: String, reference: String): DeliveryArtifact
```
luispollo, line 27: I think this is a good move. The only concern I have is that currently `DeliveryArtifact.reference` defaults to `DeliveryArtifact.name`, which, to your point in the PR description, is only unique when combined with the artifact type. I feel like we should add an additional check in the delivery config validation to ensure artifact references are unique. Maybe in a separate PR.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -280,12 +279,12 @@ class NoSuchArtifactException(name: String, type: ArtifactType) :
...
 
-class ArtifactReferenceNotFoundException(deliveryConfig: String, reference: String, type: ArtifactType) :
-  NoSuchEntityException("No $type artifact with reference $reference in delivery config $deliveryConfig is registered")
+class ArtifactReferenceNotFoundException(deliveryConfig: String, reference: String) :
+  NoSuchEntityException("No artifact with reference $reference in delivery config $deliveryConfig is registered")
 
-class ArtifactNotFoundException(name: String, type: ArtifactType, reference: String, deliveryConfig: String?) :
-  NoSuchEntityException("No $type artifact named $name with reference $reference in delivery config $deliveryConfig is registered") {
-  constructor(artifact: DeliveryArtifact) : this(artifact.name, artifact.type, artifact.reference, artifact.deliveryConfigName)
+class ArtifactNotFoundException(reference: String, deliveryConfig: String?) :
```
luispollo, line 285: It's confusing that these take the same arguments now, but in different orders, especially since they're both strings. Consider reordering for consistency.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: > There were 3 sql queries in this function. I removed these two, and added that sql into the third query. Then I added a nested query to get the pinned information. I'd love ideas on how I could remove the nested query.

Yeah, I know they weren't the most efficient approach, but I struggle reading nested or really large queries, and nesting also makes it impossible to throw specific errors like I did here, so that was on purpose. 😄

Also, the things you replaced them with both cause queries to be executed, so in terms of performance it's the same. Having said that, it does make things shorter, which might be easier to read, so I don't have any major objections, just wanted to explain what my rationale was.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -910,7 +901,14 @@ class SqlArtifactRepository(
             state = promotionStatus.toLowerCase(),
             deployedAt = deployedAt?.toInstant(ZoneOffset.UTC),
             replacedAt = replacedAt?.toInstant(ZoneOffset.UTC),
-            replacedBy = replacedBy
+            replacedBy = replacedBy,
+            pinned = jooq
+              .fetchExists(
+                ENVIRONMENT_ARTIFACT_PIN,
+                ENVIRONMENT_ARTIFACT_PIN.ENVIRONMENT_UID.eq(deliveryConfig.getUidFor(environmentName))
+                  .and(ENVIRONMENT_ARTIFACT_PIN.ARTIFACT_UID.eq(artifact.uid))
```
luispollo, line 909: Here, you're actually repeating those two queries to get the UIDs again. Consider just declaring the variables up top like:
```kotlin
val environmentUid = deliveryConfig.getUidFor(environmentName)
val artifactUid = artifact.uid
```
Then you can reuse them in both places.

luispollo: As far as the nested "exists` query, I think this is fine.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -987,6 +985,12 @@ class SqlArtifactRepository(
       .where(ENVIRONMENT.NAME.eq(environment.name))
       .and(ENVIRONMENT.DELIVERY_CONFIG_UID.eq(uid))
 
+  private fun DeliveryConfig.getUidFor(environmentName: String): Select<Record1<String>> =
```
luispollo, line 988: Is there a specific reason to return this type here instead of just `String`? You can just add a `fetchOne(ENVIRONMENT.UID)` at the end of the query.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/ApplicationServiceTests.kt
```diff
@@ -264,7 +264,8 @@ class ApplicationServiceTests : JUnit5Minutests {
               vetoed = listOf(),
               deploying = null,
               skipped = listOf()
-            )
+            ),
+            null
```
luispollo, line 268: Consider adding the property name here to make it clear what we're assigning `null` to.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
erikmunson: This is all excellent and I would love to see every field make it to the client! Lines it up very nicely to show a card like we do for judgements:

Screen Shot 2020-04-13 at 7 06 55 PM

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
erikmunson: I have no idea whether this code would actually impact what gets serialized to clients but I would strongly prefer the API continue to only return ISO timestamps

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
erikmunson: lovely

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
gal-yardeni: I love this field addition! I think it's valuable and I even for doing it mandatory... since if the user is pinning some artifact, shouldn't he explain why for others to see?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -358,8 +358,8 @@ class CombinedRepository(
   override fun getArtifact(name: String, type: ArtifactType, reference: String, deliveryConfigName: String): DeliveryArtifact =
     artifactRepository.get(name, type, reference, deliveryConfigName)
 
-  override fun getArtifact(deliveryConfigName: String, reference: String, type: ArtifactType): DeliveryArtifact =
```
gal-yardeni, line 361: just for my understanding, what used to be the values for `type`?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: Oooh it still executes them like that? Darn!! Ok I'll have to rethink this. Thanks.

I just want to speed the generation of this view up, because it's definitely getting slow. I struggle to read the big queries too, but I think it's worth it because making multiple sql connections is slower.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
emjburns: We throw an exception in the controller if it's null. So, I don't know why it can be null in the first place. I might look into it now, good suggestion.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
emjburns: Ah yes, I'll fix. Good reminder. I removed my `// todo eb: type? fix?` comment too soon 😆

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
emjburns: There is already a check validation to make sure references are unique, so we're all set here. Defaulting to name is okay because we will fail the validation if you have two artifacts with the same name and you don't define a different reference.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
emjburns: Oh wow that is confusing, didn't realize I did this! I'll fix!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: Actually, the functions I used just generate other sql and embed it in the query. So I think it is reducing the number of connections to sql.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
emjburns: `deb` or `docker`. It used to be that artifacts didn't have a reference, so you could refer to them by name + type. But now, we have a reference, and we validate that it's unique in your delivery config. So we need to use that everywhere

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
emjburns: Ok I'll send it to you.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: These two things just generate the jooq sql, so they're not actually making a new connection to the database. AFAIK making a connection to the database is the problem. But, I could be totally wrong, because I'm new at this

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: Yes, this makes it so that you can use it to build another query, instead of making the call right then

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
emjburns: Also, now, information about who pinned it and when they pinned it

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
emjburns: @gal-yardeni re: making a comment mandatory - while I think that is nice in theory, I think it would be full of friction in practice. I think that's more of a culture thing w/in teams to enforce. For example, if it's just you maintaining an app, it would be so annoying to have to fill in the comment every time.

gal-yardeni: @emjburns Yeah I think you're right :) We should though highly encourage people to explain why there are pinning it..

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -899,18 +890,30 @@ class SqlArtifactRepository(
...
         .fetchOne { (version, deployedAt, promotionStatus, replacedBy, replacedAt) ->
+          val pinned: Pinned? = jooq
+            .select(ENVIRONMENT_ARTIFACT_PIN.PINNED_BY, ENVIRONMENT_ARTIFACT_PIN.PINNED_AT, ENVIRONMENT_ARTIFACT_PIN.COMMENT)
+            .from(ENVIRONMENT_ARTIFACT_PIN)
+            .where(ENVIRONMENT_ARTIFACT_PIN.ENVIRONMENT_UID.eq(deliveryConfig.getUidFor(environmentName)))
+            .and(ENVIRONMENT_ARTIFACT_PIN.ARTIFACT_UID.eq(artifact.uid))
+            .and(ENVIRONMENT_ARTIFACT_PIN.ARTIFACT_VERSION.eq(version))
+            .fetchOne { (pinnedBy, pinnedAt, comment) ->
+              Pinned(at = Instant.ofEpochMilli(pinnedAt), by = pinnedBy, comment = comment)
+            }
```
robfletcher, line 906: I agree it would be good to eliminate this sub-query but it probably makes sense to do that as an optimization later. I'll try to find some time to have a play with doing that.

emjburns: Thanks! I'd also love to pair with you on this

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: FWIW, I don't think connecting to the database is the problem, but rather executing queries. Jooq uses a pool of reusable connections.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: Ah, TIL! Thanks!

luispollo:
Thanks for addressing my comments on this, @emjburns!

Issue #1015, opened 2020-04-14
InvalidTypeIdException thrown after kork version bump


lorin:
Starting with kork version [7.35.0], keel regularly throws the following exception if run against the prestaging database:

[7.35.0]: https://github.com/spinnaker/kork/compare/v7.34.0...v7.35.0

```
com.fasterxml.jackson.databind.exc.InvalidTypeIdException:
Could not resolve type id 'depends-on' as a subtype of 
`com.netflix.spinnaker.keel.api.Constraint`: known type ids = [] at 
[Source: (String)"[
  {"type": "depends-on", "environment": "pretest"},
  {"type": "manual-judgement", "timeout": "PT168H"}
 ]"; line: 1, column: 11]
(through reference chain: java.util.HashSet[0])
```

<details>
<summary>Complete stack trace</summary>

```
org.jooq.exception.DataAccessException: Rollback caused
	at org.jooq.impl.DefaultDSLContext.lambda$transactionResult0$0(DefaultDSLContext.java:578)
	at org.jooq.impl.Tools$12$1.block(Tools.java:4843)
	at java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3323)
	at org.jooq.impl.Tools$12.get(Tools.java:4840)
	at org.jooq.impl.DefaultDSLContext.transactionResult0(DefaultDSLContext.java:588)
	at org.jooq.impl.DefaultDSLContext.transactionResult(DefaultDSLContext.java:505)
	at com.netflix.spinnaker.keel.sql.JooqKt.inTransaction(Jooq.kt:10)
	at com.netflix.spinnaker.keel.sql.SqlDeliveryConfigRepository$itemsDueForCheck$1.invoke(SqlDeliveryConfigRepository.kt:913)
	at com.netflix.spinnaker.keel.sql.SqlDeliveryConfigRepository$itemsDueForCheck$1.invoke(SqlDeliveryConfigRepository.kt:52)
	at com.netflix.spinnaker.keel.sql.SqlRetry$sam$java_util_function_Supplier$0.get(SqlRetry.kt)
	at io.github.resilience4j.retry.Retry.lambda$decorateSupplier$2(Retry.java:177)
	at io.vavr.control.Try.of(Try.java:75)
	at io.vavr.control.Try.ofSupplier(Try.java:92)
	at com.netflix.spinnaker.keel.sql.SqlRetry.withRetry(SqlRetry.kt:24)
[...]

PR #1017, opened 2020-04-14
fix(core): Ensure there's only ever 1 of each type of configured ObjectMapper

luispollo:
We're calling `configuredObjectMapper()` all over the codebase to obtain an instance of an `ObjectMapper` with our custom configuration. This should generally yield new copies of a mapper with the same configuration, except that in the application startup code we do additional configuration of the ones registered with the Spring context, which exclude all the ones instantiated directly.

This PR ensures that there's only ever a single copy of each type of `ObjectMapper` by using a singleton for each.

Attempt to close https://github.com/spinnaker/keel/issues/1015

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
@@ -19,17 +19,20 @@ import de.huxhorn.sulky.ulid.ULID
 import java.text.SimpleDateFormat
 import java.util.TimeZone
 
+object ObjectMapperSingleton : ObjectMapper()
+object YamlMapperSingleton : YAMLMapper()
+
 /**
  * Factory method for [ObjectMapper]s configured how we like 'em.
  */
 fun configuredObjectMapper(): ObjectMapper =
-  ObjectMapper().configureMe()
+  ObjectMapperSingleton.configureMe()
 
 /**
  * Factory method for [YAMLMapper]s configured how we like 'em.
  */
 fun configuredYamlMapper(): YAMLMapper =
-  YAMLMapper()
+  YamlMapperSingleton
     .configureMe()
     .disable(USE_NATIVE_TYPE_ID)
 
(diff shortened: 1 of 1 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
```diff
@@ -19,17 +19,30 @@ import de.huxhorn.sulky.ulid.ULID
 import java.text.SimpleDateFormat
 import java.util.TimeZone
 
+class ObjectMapperSingleton private constructor() {
+  companion object {
+    val instance: ObjectMapper by lazy { ObjectMapper() }
+  }
+}
```
robfletcher, line 26: You could literally just do ```private val objectMapperInstance by lazy { ObjectMapper().configureMe() }```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
```diff
@@ -19,17 +19,30 @@ import de.huxhorn.sulky.ulid.ULID
...
+    val instance: YAMLMapper by lazy { YAMLMapper() }
+  }
+}
+
 /**
  * Factory method for [ObjectMapper]s configured how we like 'em.
  */
 fun configuredObjectMapper(): ObjectMapper =
-  ObjectMapper().configureMe()
+  ObjectMapperSingleton.instance.configureMe()
```
robfletcher, line 38: This will still call `configureMe` each time. Feels like you could roll that into the creation of the instance

luispollo: Good point. Will fix.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
luispollo: You mean in place of the singleton class? But this is a module, not a class...

robfletcher: Kotlin allows package level variables

luispollo: Ah! TIL.

PR #1022, opened 2020-04-14
chore(veto): move veto apis to application controller

emjburns:
Moving the apis for vetoing an artifact in an environment to the application controller in preparation for exposing them through gate. Removed the need to specify the artifact type and delivery config name, just like the other apis.


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -91,6 +93,31 @@ class ApplicationService(
     repository.deletePin(config, targetEnvironment)
   }
 
+  fun markAsVetoedIn(application: String, veto: EnvironmentArtifactVeto, force: Boolean) {
```
luispollo, line 96: Thank you for adding these here. I _really_ feel service objects are a good pattern to follow because not only do they make the controllers lean (as they should be), but makes it much easier to write tests without the whole Spring paraphernalia! ❤️

emjburns: Yes! I love having the service objects!

emjburns:
Question answered in person :)

PR #1025, opened 2020-04-15
chore(usability): Use consistent formats for task user

luispollo:
This makes the orca task user for a bake look the same as for other tasks launched by keel.


emjburns:
I _think_ that just setting `user = "Spinnaker` is actually the behavior you want. Could you test it out locally and verify or tell me why I'm wrong? :D

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -153,7 +153,6 @@ class ImageHandler(
               "package" to artifactRef.substringAfterLast("/"),
               "regions" to artifact.vmOptions.regions,
               "storeType" to artifact.vmOptions.storeType.name.toLowerCase(),
-              "user" to "keel",
```
luispollo, line 156: @emjburns Turns out the only difference from this orca task and all the others is that we overriding `user` here (which is set to `"Spinnaker"` by default in the `Job` type alias).

emjburns: Ahhh great :)

PR #1030, opened 2020-04-15
don't trigger region mismatch if there are more regions than we need

robfletcher:


keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -48,7 +48,12 @@ class ImageHandler(
         val diff = DefaultResourceDiff(desired, current)
 
         if (current != null && diff.isRegionsOnly()) {
-          publisher.publishEvent(ImageRegionMismatchDetected(current, artifact.vmOptions.regions))
+          if (current.regions.containsAll(desired.regions)) {
+            log.debug("Image for ${current.appVersion} contains more regions than we need, which is fine")
```
emjburns, line 52: 😆 love it.

PR #1031, opened 2020-04-15
send image diff as a parameter rather than task description

robfletcher:


keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -164,12 +163,13 @@ class ImageHandler(
...
         ),
-        artifacts = listOf(artifactPayload)
+        artifacts = listOf(artifactPayload),
+        parameters = mapOf("delta" to diff.toDebug())
       )
       publisher.publishEvent(BakeLaunched(desiredVersion))
       return listOf(Task(id = taskRef.id, name = description))
     } catch (e: Exception) {
-      log.error("Error launching bake for: ${description.toLowerCase()}")
+      log.error("Error launching bake for: $description")
```
robfletcher, line 172: I can't even imagine why we did this

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -146,7 +145,7 @@ class ImageHandler(
         user = serviceAccount,
         application = application,
         notifications = emptySet(),
-        subject = subject,
+        subject = "bakery:image:$artifact.name",
```
robfletcher, line 148: Everywhere else this is a resource id so I thought I'd make it clearly something consistent with that

PR #1034, opened 2020-04-16
default delivery config name

### LINKED ISSUE #1019: Make manifest names an optional implementation detail
We've observed that in the majority case, the `name` of a manifest is a sort of unfamiliar implementation detail of configuring your app (which is what you likely _actually_ care about) and trying to understand its intended use, what inputs are valid, and how its uniqueness is determined is very tough to digest as someone getting on board and already learning lots of new concepts. Even our own team was going back and forth on the uniqueness of names, which is a clear indication the feature is hard to intuitively grasp.

Can we make the name optional and default it to something like `<appName>-manifest` (or if we do multi-app manifests later perhaps `<concatenatedAppNames>-manifest`), and then allow querying for a manifest via application name? This would effectively make the name an invisible implementation detail for the majority case where an app is the thing you're thinking about (I'm not even sure we've found a use case where the name is super important to understand or needs human attention yet)

robfletcher:
Fixes #1019

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,5 +3,5 @@ fiatVersion=1.18.1
 kapt.use.worker.api=true
 korkVersion=7.36.0
 liquibaseTaskPrefix=liquibase
-org.gradle.parallel=true
+#org.gradle.parallel=true
 spinnakerGradleVersion=7.11.5
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryConfig.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryConfig.kt
@@ -4,8 +4,8 @@ import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 
 data class DeliveryConfig(
-  val name: String,
   val application: String,
+  val name: String,
   val serviceAccount: String,
   val artifacts: Set<DeliveryArtifact> = emptySet(),
   val environments: Set<Environment> = emptySet(),
--- a/keel-bakery-plugin/src/test/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandlerTests.kt
+++ b/keel-bakery-plugin/src/test/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandlerTests.kt
@@ -239,19 +239,7 @@ internal class ImageHandlerTests : JUnit5Minutests {
               }
 
               test("no bake is launched") {
-                verify(exactly = 0) {
-                  taskLauncher.submitJob(
-                    user = any(),
-                    application = any(),
-                    notifications = any(),
-                    subject = any(),
-                    description = any(),
-                    correlationId = any(),
-                    stages = any(),
-                    artifacts = any(),
-                    parameters = any()
-                  )
-                }
+                expectThat(bakeTask).isNotCaptured()
               }
             }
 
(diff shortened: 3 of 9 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/DeliveryConfig.kt
```diff
@@ -4,8 +4,8 @@ import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 
 data class DeliveryConfig(
-  val name: String,
   val application: String,
+  val name: String,
```
robfletcher, line 8: Just changed the order to be consistent with `SubmittedDeliveryConfig`

keel-bakery-plugin/src/test/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandlerTests.kt
```diff
@@ -239,19 +239,7 @@ internal class ImageHandlerTests : JUnit5Minutests {
...
-                    notifications = any(),
-                    subject = any(),
-                    description = any(),
-                    correlationId = any(),
-                    stages = any(),
-                    artifacts = any(),
-                    parameters = any()
-                  )
-                }
+                expectThat(bakeTask).isNotCaptured()
```
robfletcher, line 242: D'oh, this was an uncommitted change from yesterday. Pls to excuse.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/SubmittedDeliveryConfig.kt
```diff
@@ -12,8 +12,8 @@ const val DEFAULT_SERVICE_ACCOUNT = "keel@spinnaker.io"
 
 @Description("A manifest specifying the environments and resources that comprise an application.")
 data class SubmittedDeliveryConfig(
-  val name: String,
   val application: String,
+  val name: String = "$application-manifest",
```
robfletcher, line 16: Default the name here (had to change property order as otherwise I can't use `application` in the definition of `name`)

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
```diff
@@ -144,7 +144,6 @@ internal class DeliveryConfigControllerTests : JUnit5Minutests {
     context("submitting a delivery config manifest") {
       val yamlPayload =
         """---
-        |name: keel-manifest
```
robfletcher, line 147: Nice easy way to test is to just omit it from the submitted data when calling the controller. I didn't relax the requirement for the `name` to be specified on `DeliveryConfig` itself, only at the edge.

robfletcher:
>     * Which docs need to change?

Will take a look in a minute

>     * Should we PR our internal users' configs to remove their names? I seriously doubt anyone is using names in a way that would preclude us from removing them, but curious to hear whether folks think doing that is necessary.

IMO it would be sufficient to announce it in the #managed-delivery channel. I don't feel like it's an onerous thing for people to self-serve on.

emjburns:
@robfletcher it would be a great thing to announce along with a follow up to our users, like, "let us know when you complete this by reacting with X emoji" - I suspect people might have some permissions issues because we just put more auth in place, and you now need a service account on your submitted pipelines. This would flush out all of those problems.

gradle.properties
```diff
@@ -3,5 +3,5 @@ fiatVersion=1.18.1
 kapt.use.worker.api=true
 korkVersion=7.36.0
 liquibaseTaskPrefix=liquibase
-org.gradle.parallel=true
+#org.gradle.parallel=true
```
plumpy, line 6: Did you mean to change this? I only ask because it wasn't mentioned at all in the PR and seems like the kind of thing you might have done for testing and accidentally committed...

(I only noticed it because it caused a merge conflict for an autobump)

robfletcher: We had a test behaving strangely only on CI and this was an attempt to isolate it. Going to revert in another PR today.

PR #1036, opened 2020-04-16
refactor(events): Improve resource events for error cases

### LINKED ISSUE #686: Map ResourceCheckUnresolvable history event to ERROR resource status
We recently ran into an issue where a cluster resource couldn't find its docker image, and we published a `ResourceCheckUnresolvable` event to the history feed as a result. While looking at that event is fairly explanatory, the resource itself remained in a `DIFF` status the whole time.

In theory that particular error (no docker image) shouldn't really send a resource into an error state the way it did in this case, but generally any time a resource has a `ResourceCheckUnresolvable` event as its most recent event, it makes sense to clearly communicate that in the status so users are aware something is wrong/unexpected. This will be especially helpful in cases where there's a system level bug with keel that might not get caught until a human notices a resource is in a bad state (our notifications don't currently cover this type of problem).

Absent other ideas/proposals, I recommend we always map resources where a `ResourceCheckUnresolvable` event is the most recent event to an `ERROR` status.

luispollo:
Fixes #686.

Starting from these goals (as articulated by @erikmunson):
- How we can un-overload that event so the history gives very clear, single-purpose messages to folks
- How that can bubble up to clear statuses — that might mean we end up still mapping everything to ERROR, or it might mean we need something else too (what i'm seeing in that list is some stuff falls into a sort of 'invalid config' error status)

...this PR:
- Revisits all keel base exception types to inherit from base exception types from kork with well-defined meanings, specifically:
  - `UserException` (_An exception thrown when the cause is a user (e.g. bad input)._)
  - `SystemException` (_An exception thrown by core Spinnaker; this will represent a system-level defect, interruption, etc._)
- Revisits base classes for exceptions currently inheriting from `ResourceCurrentlyUnresolvable` (_Exception that may be thrown during resource desired state resolution implementations to indicate that some object the resource relies on is not (yet) available. (...) Exceptions of this type are not treated as fatal as they are expected to be transient._) such that only exceptions that actually fit that description inherit from it. This, by itself, will take care of avoiding to publish misleading `ResourceCheckUnresolvable` events for things that are actually final errors (like bad config input from the user).
- Revisits `ResourceCheckError` event by adding an `origin` field that is set to an enum value of either `USER` or `SYSTEM` matching the base type of the exception causing the error (which, per above, will either be `UserException` or `SystemException`, respectively). This is in addition to the `exceptionType: Class<out SpinnakerException>` already present in the event, with the specific goal of making parsing by the UI simpler.
- Introduces a new `CURRENTLY_UNRESOLVABLE` resource status to match the `ResourceCheckUnresolvable` event and represent transient resolution errors, thereby distinguishing from unrecoverable errors represented by the `ERROR` status.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -174,7 +177,11 @@ class ResourceActuator(
         publisher.publishEvent(ResourceCheckUnresolvable(resource, e, clock))
       } catch (e: Exception) {
         log.error("Resource check for $id failed [$coid]", e)
-        publisher.publishEvent(ResourceCheckError(resource, e, clock))
+        val normalizedException = when (e) {
+          is UserException, is SystemException -> e
+          else -> SystemException(e)
+        } as SpinnakerException
+        publisher.publishEvent(ResourceCheckError(resource, normalizedException, clock))
```
luispollo, line 184: This ensures that the exception type of a `ResourceCheckError` will always be `UserException` or `SystemException`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -55,7 +58,9 @@ import java.time.Instant
   Type(value = ResourceTaskFailed::class, name = "ResourceTaskFailed"),
   Type(value = ResourceTaskSucceeded::class, name = "ResourceTaskSucceeded")
 )
-abstract class ResourceEvent : PersistentEvent() {
+abstract class ResourceEvent(
+  open val message: String? = null
```
luispollo, line 62: @erikmunson, this should take care of:
```typescript
  // We really should not have 3 different versions of basically
  // the same field, but right now we do.
  message?: string;
  reason?: string;
  exceptionMessage?: string;
```
https://github.com/spinnaker/deck/blob/9db0ef63071b110bd98a26f6f31e5108f6c9f15c/app/scripts/modules/core/src/domain/IManagedEntity.ts#L160-L164

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -367,23 +374,37 @@ data class ResourceCheckUnresolvable(
...
   override val kind: ResourceKind,
   override val id: String,
   override val application: String,
   override val timestamp: Instant,
-  val exceptionType: Class<out Throwable>,
-  val exceptionMessage: String?
-) : ResourceCheckResult() {
+  val exceptionType: Class<out SpinnakerException>,
+  val exceptionMessage: String?,
+  val origin: ResourceCheckErrorOrigin
```
luispollo, line 396: The purpose of `origin` is to give a really simple hint to the UI as to what the original cause of the problem was. Was it something the user did wrong? Or something keel did wrong/couldn't do? Hopefully that helps render better error messages on the UI.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -367,23 +374,37 @@ data class ResourceCheckUnresolvable(
...
+) : ResourceCheckResult(message = exceptionMessage) {
   @JsonIgnore
   override val state = Error
 
-  constructor(resource: Resource<*>, exception: Throwable, clock: Clock = Companion.clock) : this(
+  constructor(resource: Resource<*>, exception: SpinnakerException, clock: Clock = Companion.clock) : this(
```
luispollo, line 401: Changing the exception type here to give us some basic compiler protection against not using the right base-types. Not perfect, but better than `Throwable`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -367,23 +374,37 @@ data class ResourceCheckUnresolvable(
...
     resource.kind,
     resource.id,
     resource.application,
     clock.instant(),
     exception.javaClass,
-    exception.message
+    exception.message,
+    ResourceCheckErrorOrigin.fromException(exception)
```
luispollo, line 408: This line will actually `throw` if the exception is not a `SystemException` or `UserException` with a proper error message. We'll only catch it a runtime, but at least we'll know if we messed up the types.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/InvalidRegexException.kt
```diff
@@ -18,14 +18,15 @@
...
+import com.netflix.spinnaker.kork.exceptions.UserException
 
 /**
  * The regex provided produced too many capture groups.
  */
 class InvalidRegexException(
   val regex: String,
   val tag: String
-) : ResourceCurrentlyUnresolvable("The provided regex ($regex) produced did not produce one capture group on tag $tag")
+) : UserException("The provided regex ($regex) produced did not produce one capture group on tag $tag")
```
luispollo, line 29: This will no longer result in a `ResourceCheckUnresolvable` event.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/NoDockerImageFound.kt
```diff
@@ -18,9 +18,10 @@
...
 
 import com.netflix.spinnaker.keel.core.ResourceCurrentlyUnresolvable
+import com.netflix.spinnaker.kork.exceptions.IntegrationException
 
 class NoDigestFound(repository: String, tag: String) :
   ResourceCurrentlyUnresolvable("No digest found for docker image $repository:$tag in any registry")
 
 class RegistryNotFound(titusAccount: String) :
-  ResourceCurrentlyUnresolvable("Unable to find docker registry for titus account $titusAccount")
+  IntegrationException("Unable to find docker registry for titus account $titusAccount")
```
luispollo, line 27: This will no longer result in a `ResourceCheckUnresolvable` event.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/VetoController.kt
```diff
@@ -97,11 +95,4 @@ class VetoController(
     log.error(e.message)
     return mapOf("message" to e.message)
   }
-
-  @ExceptionHandler(MalformedMessageException::class)
```
luispollo, line 101: This exception was not being thrown anywhere.

luispollo:
Note: the CI build error is unrelated and is happening in all PR builds. 🤷‍♂️

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
luispollo: Actually, since the original properties are still there, and are the values used to fill in the `message` property in the base class, deserialization should work transparently and you should start seeing the `message` field in the JSON _in addition to_ the other ones. I'll test this out to make sure.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/NoDockerImageFound.kt
luispollo: Yes. `IntegrationException` is a sub-class of `SystemException`. For a moment, I was tempted to use just `SystemException` and `UserException` everywhere, but it seemed like a waste to not use descendent types that were more accurate.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
```diff
@@ -175,6 +176,6 @@ fun <T : ResourceSpec> Collection<ResourceHandler<*, *>>.supporting(
   find { it.supportedKind.specClass == specClass }
 
 class UnsupportedKind(kind: String) :
-  IllegalStateException("No resource handler supporting \"$kind\" is available") {
+  SystemException("No resource handler supporting \"$kind\" is available") {
```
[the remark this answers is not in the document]
luispollo: Yeah, this is a tricky one, because it happens in both situations. For example, when reading a resource from the DB, if the `kind` is not recognized, it is thrown, but also when you call the `/diff` endpoint with a bogus kind, it will get thrown. The vast majority of cases where it appears are "system" cases, though. I'd propose we wrap it with a `UserException` in REST controllers.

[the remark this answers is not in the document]
erikmunson: that seems like a great idea — no need to conflate the two contexts/use cases if we don't have to

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/NoDockerImageFound.kt
erikmunson: 👍 I'm also in favor of using the more granular exceptions kork has to offer, was mostly just curious to learn what it would look like as an API client. I think sticking with `SYSTEM` for now is fine and later we can decide whether there's usefulness in supporting more granular `origin` values in the events themselves.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
erikmunson: oooooooooooooooooo that is great. thx for clarifying.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/ImageProviderDeserializer.kt
```diff
@@ -35,4 +35,4 @@ internal class ImageProviderDeserializer :
     }
 }
 
-class InvalidPayload(strategy: String) : UserException("Payload is not valid for $strategy")
+class InvalidPayload(strategy: String) : SystemException("Payload is not valid for $strategy")
```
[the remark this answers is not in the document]
luispollo: Sorry, I should have added a comment since I went from one extreme to the other. This is only thrown by the custom polymorphic deserialization code @robfletcher wrote, which is why I thought it'd be better represented as a system exception. Rob, what was your rationale to use `UserException` originally?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -267,5 +273,5 @@ class NoSuchApplication(application: String) :
   NoSuchEntityException("No resource with application name $application exists in the database")
 
 enum class ResourceStatus {
-  HAPPY, ACTUATING, UNHAPPY, CREATED, DIFF, ERROR, PAUSED, VETOED, RESUMED, UNKNOWN
+  HAPPY, ACTUATING, UNHAPPY, CREATED, DIFF, ERROR, CURRENTLY_UNRESOLVABLE, PAUSED, VETOED, RESUMED, UNKNOWN
```
erikmunson, line 276: yeeeeeah i like this a lot. nice solve.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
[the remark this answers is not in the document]
luispollo: Actually the API case is already covered by our REST exception handler, in which case there's no impact on resource events, so I think we're good:
https://github.com/spinnaker/keel/blob/60b97474b4d6d637ef9eb444557a6dddd0a78d2e/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt#L57-L62

luispollo:
Side-note: I don't know why, but every time I add the `spinnaker/managed-delivery` team as a reviewer GitHub removes it. 😞

luispollo:
4

> Side-note: I don't know why, but every time I add the `spinnaker/managed-delivery` team as a reviewer GitHub removes it. 😞

Ah, I see it now!
image

That's dumb... It should only do that if someone actually publishes a review, not just a comment!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
luispollo: I'm glad you brought this up! When testing, I found I broke deserialization, not because of the `message` field (which worked as expected), but because of the new `origin` field! It's fixed now. The output looks like this:
```yaml
- type: "ResourceCheckError"
  kind: "titus/cluster@v1"
  id: "titus:cluster:titustestvpc:lpollomdtest-test"
  application: "lpollomdtest"
  timestamp: "2020-04-10T21:42:34.028Z"
  exceptionType: "com.netflix.spinnaker.keel.plugin.CannotResolveCurrentState"
  exceptionMessage: "Unable to resolve current state of titus:cluster:titustestvpc:lpollomdtest-test\
    \ due to: timeout (through reference chain: java.util.ArrayList[10326])"
  scope: "resource"
  message: "Unable to resolve current state of titus:cluster:titustestvpc:lpollomdtest-test\
    \ due to: timeout (through reference chain: java.util.ArrayList[10326])"
  origin: "SYSTEM"
  uid: "titus:cluster:titustestvpc:lpollomdtest-test"
```

I can turn the `origin` enum values to lowercase if you prefer.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -174,7 +178,15 @@ class ResourceActuator(
         publisher.publishEvent(ResourceCheckUnresolvable(resource, e, clock))
       } catch (e: Exception) {
         log.error("Resource check for $id failed [$coid]", e)
-        publisher.publishEvent(ResourceCheckError(resource, e, clock))
+        val normalizedException = when (e) {
```
gal-yardeni, line 181: Is it possible to extract this logic to a method?

gal-yardeni:
I think this is great as well! Thanks for doing it. It would be super valuable to have some kind of doc explaining the meaning of each exception, and how it is rendering by the UI/API, for future referance.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -172,7 +173,7 @@ class ArtifactListener(
    */
   private fun debStatus(artifact: Artifact): ArtifactStatus {
     val status = artifact.metadata["releaseStatus"]?.toString()
-      ?: throw IllegalStateException("Artifact event received without 'releaseStatus' field")
+      ?: throw IntegrationException("Artifact event received without 'releaseStatus' field")
```
emjburns, line 176: I think here `IntegrationException` is less clear. I think it would be more clear if we had an `IllegalStateException` that extended the `IntegrationException`. Thoughts?

keel-api/keel-api.gradle.kts
```diff
@@ -10,6 +10,7 @@ plugins {
  * dependencies along for the ride -- ideally nothing at all.
  */
 dependencies {
+  api("com.netflix.spinnaker.kork:kork-core")
```
robfletcher, line 13: Is this compatible with the pf4j plugin stuff? My understanding is we should not have any dependencies in here.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
robfletcher: Also seems like a system exception to me

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/UnknownBaseImage.kt
```diff
@@ -1,6 +1,7 @@
 package com.netflix.spinnaker.keel.bakery
 
 import com.netflix.spinnaker.keel.api.artifacts.BaseLabel
-import com.netflix.spinnaker.kork.exceptions.UserException
+import com.netflix.spinnaker.kork.exceptions.ConfigurationException
 
-class UnknownBaseImage(os: String, label: BaseLabel) : UserException("Could not identify base image for os $os and label $label")
+class UnknownBaseImage(os: String, label: BaseLabel) :
+  ConfigurationException("Could not identify base image for os $os and label $label")
```
robfletcher, line 7: I'm not sure that's a `ConfigurationException` is it?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -367,18 +376,35 @@ data class ResourceCheckUnresolvable(
...
+enum class ResourceCheckErrorOrigin {
+  @JsonProperty("user") USER,
+  @JsonProperty("system") SYSTEM;
+
+  companion object {
+    fun fromException(exceptionType: Class<out SpinnakerException>) =
+      when {
+        UserException::class.java.isAssignableFrom(exceptionType) -> USER
+        SystemException::class.java.isAssignableFrom(exceptionType) -> SYSTEM
+        else -> throw SystemException("All keel exceptions should inherit from UserException or SystemException. This is a bug.")
```
emjburns, line 388: I think it would be better to log a message here, set this to "UNKNOWN", and then set up an alert to watch for the message. If I'm not mistaken, this code here means that If we miss something in a PR review or accidentally do it wrong, then the whole history view will error out.  

I strongly prefer logging messages when something is in a weird but not harmful state, because that means it's more likely a view will remain complete for users.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -195,7 +197,7 @@ interface ResourceRepository : PeriodicallyCheckedRepository<Resource<out Resour
   }
 
   private fun List<ResourceEvent>.isDiff(): Boolean {
-    return first() is ResourceDeltaDetected || first() is ResourceMissing || first() is ResourceCheckUnresolvable
+    return first() is ResourceDeltaDetected || first() is ResourceMissing
```
emjburns, line 200: Much better lol

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
luispollo: You know, I used `IntegrationException` here because of this as artifact events coming from echo or igor, but this is an internal event handler. I looked at the other available kork exceptions to see if there was something closer in meaning but didn't find anything. Would you be ok with just changing to `SystemException`?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/ImageProviderDeserializer.kt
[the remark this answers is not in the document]
emjburns: I think this happens in the case where you've submitted a resource with an invalid `imageProvide`, in which case I think this a `UserException`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
luispollo: Good point. Will change.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -367,18 +376,35 @@ data class ResourceCheckUnresolvable(
...
+      }
+  }
+}
```
robfletcher: Can you explain how this works?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: Actually, now that I read through your PR more, I think it's okay as it is. It is an `IntegrationException` because this is coming from echo.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/ImageProviderDeserializer.kt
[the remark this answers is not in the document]
robfletcher: Isn't this used on submission of a delivery config? e.g. if your resource kind doesn't match the spec?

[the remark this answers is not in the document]
luispollo: Yes. I was gonna say "but also when we load it from the db", but that's not a thing. Will revert.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
luispollo: So this looks at the `exceptionType` passed to the `ResourceCheckError` constructor to determine a high-level "type" describing where the error originated, which I called `origin`. If it's a sub-class of `UserException`, it's a `USER` error, if it's a sub-class of `SystemException`, it's a `SYSTEM` error.

robfletcher: Can the `else` ever happen? It looks like it will always be one or the other, right?

luispollo: It can because the type of `exceptionType` is `Class<SpinnakerException>`, and there are other sub-classes of `SpinnakerException` beyond those two.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/UnknownBaseImage.kt
luispollo: Sub-class of `UserException` specifically for when the error is with something the user mis-configured. I went back and forth on using this type. I can change it to `UserException` if you think it's better.

keel-api/keel-api.gradle.kts
luispollo: Good question. I'll have to defer to @robzienert on that... I added this so I could use `SystemException` instead of `Throwable` in the handler interface.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/UnknownBaseImage.kt
luispollo: Sorry, I misread your comment. You're right. Will change to `SystemException`.

keel-api/keel-api.gradle.kts
robzienert: `kork-core` is definitely a no-no module. You can use `kork-exceptions`, however.

luispollo: Thanks! Good catch, @robfletcher!

keel-api/keel-api.gradle.kts
```diff
@@ -10,6 +10,7 @@ plugins {
  * dependencies along for the ride -- ideally nothing at all.
  */
 dependencies {
+  api("com.netflix.spinnaker.kork:kork-exceptions")
```
luispollo, line 13: So we can use `SystemException` in `ResourceHandler`.

PR #1037, opened 2020-04-17
fix test that leaks data

robfletcher:


gradle.properties
```diff
@@ -3,5 +3,5 @@ fiatVersion=1.18.1
 kapt.use.worker.api=true
 korkVersion=7.37.0
 liquibaseTaskPrefix=liquibase
-#org.gradle.parallel=true
+org.gradle.parallel=true
```
robfletcher, line 6: Restored this… wasn't the problem

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -79,14 +80,10 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
       Fixture(subject = factory(clock))
     }
 
-//    after { confirmVerified(callback) }
+    after { confirmVerified(callback) }
```
robfletcher, line 83: Put this back

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -79,14 +80,10 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
...
     after { flush() }
 
     context("no resources exist") {
-      before {
-        flush()
-      }
```
robfletcher, line 88: This was my workaround, flush the database before the tests because something else wasn't flushing after

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/LegacySpecUpgradeTests.kt
```diff
@@ -87,11 +88,19 @@ internal class LegacySpecUpgradeTests : JUnit5Minutests {
...
+      cleanupDb(jooq)
+    }
   }
 
   fun tests() = rootContext<Fixture> {
     fixture { Fixture }
 
+    after {
+      flush()
+    }
```
robfletcher, line 102: This was the culprit

PR #1038, opened 2020-04-17
fix(errors): Improve authorization error messages further

luispollo:
Make authorization error messages as detailed as possible.

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
@@ -7,6 +7,8 @@ import com.fasterxml.jackson.databind.exc.InvalidFormatException
 import com.fasterxml.jackson.databind.exc.InvalidTypeIdException
 import com.fasterxml.jackson.databind.exc.MismatchedInputException
 import com.fasterxml.jackson.module.kotlin.MissingKotlinParameterException
+import com.netflix.spinnaker.fiat.model.resources.ResourceType.SERVICE_ACCOUNT
+import com.netflix.spinnaker.fiat.shared.FiatPermissionEvaluator
 import com.netflix.spinnaker.keel.api.ResourceSpec
 import com.netflix.spinnaker.keel.api.plugins.ResourceHandler
 import com.netflix.spinnaker.keel.api.plugins.UnsupportedKind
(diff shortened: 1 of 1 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -86,14 +88,28 @@ class ExceptionHandler(
   @ResponseStatus(FORBIDDEN)
   fun onAccessDenied(e: AccessDeniedException): ApiError {
     log.error(e.message)
-    return ApiError(FORBIDDEN,
+    val authFailure = FiatPermissionEvaluator.getAuthorizationFailure().get()
+    val message = if (authFailure != null) {
+      val user = "User ${AuthenticatedRequest.getSpinnakerUser().orElse("")}"
+      val permission = "${authFailure.authorization.let { if (it == null) "access" else "$it permission" }}"
+      val resourceType = authFailure.resourceType.name.toLowerCase().replace('_', ' ')
+      "Access denied. $user does not have $permission to the ${authFailure.resourceName} $resourceType specified in your request. " +
```
luispollo, line 96: If the authorization failure info from Fiat is available, this will now include the username, permission and resource in the error message. For example, "User joe@acme.com does not have READ access to the myteam@acme.com service account specified in your request."

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -86,14 +88,28 @@ class ExceptionHandler(
...
+        if (authFailure.resourceType == SERVICE_ACCOUNT) {
```
luispollo, line 97: This tailors the remainder of the message based on whether the missing permission is to the service account or other resources.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -86,14 +88,28 @@ class ExceptionHandler(
...
+          "Please make sure you have access to the service account specified in your delivery config."
+        } else {
+          "Please check that the service account specified in your delivery config has access to this application along " +
+            "with all the cloud accounts included in the delivery config."
+        }
+    } else {
       if (e.message == null || e.message == "Access is denied") {
         "Access denied. Please make sure you have access to the service account specified in your delivery config. " +
-          "If you do have access, check that the service account has access to this application along with all the cloud accounts included in the delivery config."
+          "If you do have access, check that the service account has access to this application along with all the " +
```
luispollo, line 106: Generic message remains here as a fallback.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
gal-yardeni: I like it, make it much more clear.

erikmunson:
sailormoon

Issue #212, opened 2019-02-27
Don't attempt to anneal if dependent resources don't exist


robfletcher:
e.g. security groups referenced by a cluster

robfletcher:
Exception: default security group which gets created automatically if a server group is not being attached to any other security groups.

spinnakerbot:
This issue hasn't been updated in 111 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

robfletcher:
@spinnakerbot remove-label stale

PR #1040, opened 2020-04-17
Vetos for ensuring dependent security groups / load balancers exist

robfletcher:
See #212

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
@@ -16,6 +16,7 @@
 package com.netflix.spinnaker.keel.clouddriver
 
 import com.netflix.spinnaker.keel.clouddriver.model.ActiveServerGroup
+import com.netflix.spinnaker.keel.clouddriver.model.AmazonLoadBalancer
 import com.netflix.spinnaker.keel.clouddriver.model.ApplicationLoadBalancerModel
 import com.netflix.spinnaker.keel.clouddriver.model.ClassicLoadBalancerModel
 import com.netflix.spinnaker.keel.clouddriver.model.Credential
@@ -78,6 +79,12 @@ interface CloudDriverService {
     @Path("account") account: String,
     @Header("X-SPINNAKER-USER") user: String = DEFAULT_SERVICE_ACCOUNT
   ): Credential
+
+  @GET("/applications/{application}/loadBalancers")
+  suspend fun loadBalancersForApplication(
+    @Header("X-SPINNAKER-USER") user: String,
+    @Path("application") application: String
+  ): List<AmazonLoadBalancer>
 
   @GET("/{provider}/loadBalancers/{account}/{region}/{name}")
   suspend fun getClassicLoadBalancer(
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/AmazonLoadBalancer.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/AmazonLoadBalancer.kt
@@ -1,11 +1,26 @@
 package com.netflix.spinnaker.keel.clouddriver.model
 
+import com.fasterxml.jackson.annotation.JsonSubTypes
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import com.fasterxml.jackson.annotation.JsonTypeInfo.As
+import com.fasterxml.jackson.annotation.JsonTypeInfo.Id
 import com.netflix.spinnaker.keel.api.Moniker
 
+@JsonTypeInfo(
+  use = Id.NAME,
+  include = As.PROPERTY,
+  property = "loadBalancerType",
+  defaultImpl = ClassicLoadBalancerModel::class // CloudDriver's response does not specify the loadBalancerType on classic load balancers
+)
+@JsonSubTypes(
+  JsonSubTypes.Type(ClassicLoadBalancerModel::class, name = "classic"),
+  JsonSubTypes.Type(ApplicationLoadBalancerModel::class, name = "application")
+)
 interface AmazonLoadBalancer {
   val moniker: Moniker?
   val loadBalancerName: String
   val loadBalancerType: String
+  val region: String
   val availabilityZones: Set<String>
   val vpcId: String
   val subnets: Set<String>
--- a/keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverServiceTests.kt
+++ b/keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverServiceTests.kt
@@ -1,0 +1,51 @@
+package com.netflix.spinnaker.keel.clouddriver
+
+import com.netflix.spinnaker.keel.core.api.DEFAULT_SERVICE_ACCOUNT
+import com.netflix.spinnaker.keel.serialization.configuredObjectMapper
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import kotlinx.coroutines.runBlocking
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import retrofit2.Retrofit
+import retrofit2.converter.jackson.JacksonConverterFactory
+import retrofit2.create
+import strikt.api.expectThat
+import strikt.assertions.hasSize
+
+class CloudDriverServiceTests : JUnit5Minutests {
+  class Fixture {
+    val server = MockWebServer()
+    val service: CloudDriverService by lazy {
+      Retrofit.Builder()
+        .addConverterFactory(JacksonConverterFactory.create(configuredObjectMapper()))
+        .baseUrl(server.url("/"))
+        .build()
+        .create<CloudDriverService>()
+    }
+  }
+
+  fun tests() = rootContext<Fixture> {
+    fixture { Fixture() }
+
+    after {
+      server.shutdown()
+    }
+
+    context("getting load balancers for an application") {
+      before {
+        with(server) {
+          enqueue(MockResponse().setBody(javaClass.getResource("/loadBalancers.json").readText()))
+          start()
+        }
+      }
+
+      test("can handle a mixture of application and classic load balancers") {
+        val results = runBlocking {
+          service.loadBalancersForApplication(DEFAULT_SERVICE_ACCOUNT, "keeldemo")
+        }
+        expectThat(results).hasSize(10)
+      }
+    }
+  }
+}
--- a/keel-clouddriver/src/test/resources/loadBalancers.json
+++ b/keel-clouddriver/src/test/resources/loadBalancers.json
@@ -1,0 +1,1357 @@
+[
+  {
+    "account": "test",
+    "name": "keeldemo-main-clb",
+    "region": "us-east-1",
+    "vpcId": "vpc-11111111",
+    "serverGroups": [],
+    "targetGroups": [],
+    "type": "aws",
+    "cloudProvider": "aws",
+    "moniker": {
+      "app": "keeldemo",
+      "cluster": "keeldemo-main-clb",
+      "detail": "clb",
+      "stack": "main"
+    },
+    "labels": {},
+    "canonicalHostedZoneNameID": "XXXXXXXXXXXXXX",
+    "scheme": "internal",
+    "instances": [],
+    "policies": {
+      "appCookieStickinessPolicies": [],
+      "otherPolicies": [],
+      "lbcookieStickinessPolicies": []
+    },
+    "availabilityZones": [
+      "us-east-1c",
+      "us-east-1d",
+      "us-east-1e"
+    ],
+    "backendServerDescriptions": [],
+    "sourceSecurityGroup": {
+      "ownerAlias": "999999999999",
+      "groupName": "keeldemo-elb"
+    },
+    "loadBalancerName": "keeldemo-main-clb",
+    "listenerDescriptions": [
+      {
+        "listener": {
+          "protocol": "HTTP",
+          "loadBalancerPort": 80,
+          "instanceProtocol": "HTTP",
+          "instancePort": 8080
+        },
+        "policyNames": []
+      }
+    ],
+    "healthCheck": {
+      "target": "HTTP:8080/admin/health",
+      "interval": 10,
+      "timeout": 5,
+      "unhealthyThreshold": 2,
+      "healthyThreshold": 5
+    },
+    "vpcid": "vpc-11111111",
+    "idleTimeout": 60,
+    "dnsname": "internal-keeldemo-main-clb-1444444440.us-east-1.elb.amazonaws.com",
+    "createdTime": 1572286917070,
+    "subnets": [
+      "subnet-12345678",
+      "subnet-55555555",
+      "subnet-44444444"
+    ],
+    "securityGroups": [
+      "sg-11111111111111111"
+    ],
+    "attributes": {
+      "crossZoneLoadBalancing": {
+        "enabled": true
+      },
+      "accessLog": {
+        "enabled": false
+      },
+      "connectionDraining": {
+        "enabled": true,
+        "timeout": 60
+      },
+      "connectionSettings": {
+        "idleTimeout": 60
+      },
+      "additionalAttributes": []
+    }
+  },
+  {
+    "account": "test",
+    "name": "keeldemo-main-clb",
+    "region": "us-west-2",
+    "vpcId": "vpc-12345678",
+    "serverGroups": [
+      {
+        "name": "keeldemo-main-v065",
+        "isDisabled": true,
+        "detachedInstances": [],
+        "instances": [
+          {
+            "id": "i-0144444444abcdef0",
+            "name": "i-0144444444abcdef0",
+            "zone": "us-west-2a",
+            "health": {}
+          }
+        ],
+        "cloudProvider": "aws"
+      },
+      {
+        "name": "keeldemo-main-v066",
+        "isDisabled": false,
+        "detachedInstances": [],
+        "instances": [
+          {
+            "id": "i-144444444abcdef00",
+            "name": "i-144444444abcdef00",
+            "zone": "us-west-2c",
+            "health": {
+              "state": "InService",
+              "reasonCode": "N/A",
+              "description": "N/A"
+            }
+          }
+        ],
+        "cloudProvider": "aws"
+      }
+    ],
+    "targetGroups": [],
+    "type": "aws",
+    "cloudProvider": "aws",
+    "moniker": {
+      "app": "keeldemo",
+      "cluster": "keeldemo-main-clb",
+      "detail": "clb",
+      "stack": "main"
+    },
+    "labels": {},
+    "canonicalHostedZoneNameID": "ZZZZZZZZZZZZZ",
+    "scheme": "internal",
+    "instances": [
+      "i-144444444abcdef00"
+    ],
+    "policies": {
+      "appCookieStickinessPolicies": [],
+      "otherPolicies": [],
+      "lbcookieStickinessPolicies": []
+    },
+    "availabilityZones": [
+      "us-west-2c",
+      "us-west-2b",
+      "us-west-2a"
+    ],
+    "backendServerDescriptions": [],
+    "sourceSecurityGroup": {
+      "ownerAlias": "999999999999",
+      "groupName": "keeldemo-elb"
+    },
+    "loadBalancerName": "keeldemo-main-clb",
+    "listenerDescriptions": [
+      {
+        "listener": {
+          "protocol": "HTTP",
+          "loadBalancerPort": 80,
+          "instanceProtocol": "HTTP",
+          "instancePort": 8080
+        },
+        "policyNames": []
+      }
+    ],
+    "healthCheck": {
+      "target": "HTTP:8080/admin/health",
+      "interval": 10,
+      "timeout": 5,
+      "unhealthyThreshold": 2,
+      "healthyThreshold": 5
+    },
+    "vpcid": "vpc-12345678",
+    "idleTimeout": 60,
+    "dnsname": "internal-keeldemo-main-clb-222222222.us-west-2.elb.amazonaws.com",
+    "createdTime": 1572286851780,
+    "subnets": [
+      "subnet-11111111",
+      "subnet-22222222",
+      "subnet-33333333"
+    ],
+    "securityGroups": [
+      "sg-11111111111111111"
+    ],
+    "attributes": {
+      "crossZoneLoadBalancing": {
+        "enabled": true
+      },
+      "accessLog": {
+        "enabled": false
+      },
+      "connectionDraining": {
+        "enabled": true,
+        "timeout": 60
+      },
+      "connectionSettings": {
+        "idleTimeout": 60
+      },
+      "additionalAttributes": []
+    }
+  },
+  {
+    "account": "test",
+    "name": "keeldemo-prestaging-alb",
+    "region": "us-east-1",
+    "vpcId": "vpc-11111111",
+    "serverGroups": [],
+    "targetGroups": [
+      {
+        "account": "test",
+        "name": "keeldemo-prestaging",
+        "region": "us-east-1",
+        "vpcId": "vpc-11111111",
+        "serverGroups": [
+          {
+            "name": "keeldemo-prestaging-v041",
+            "isDisabled": true,
+            "detachedInstances": [],
+            "instances": [],
+            "cloudProvider": "aws"
+          },
+          {
+            "name": "keeldemo-prestaging-v042",
+            "isDisabled": false,
+            "detachedInstances": [],
+            "instances": [
+              {
+                "id": "i-22222222222222222",
+                "name": "i-22222222222222222",
+                "zone": "us-east-1e",
+                "health": {
+                  "state": "healthy"
+                }
+              }
+            ],
+            "cloudProvider": "aws"
+          }
+        ],
+        "type": "aws",
+        "cloudProvider": "aws",
+        "moniker": {
+          "app": "keeldemo",
+          "cluster": "keeldemo-prestaging",
+          "stack": "prestaging"
+        },
+        "labels": {},
+        "healthCheckEnabled": true,
+        "loadBalancerNames": [
+          "keeldemo-prestaging-alb"
+        ],
+        "instances": [
+          "i-22222222222222222"
+        ],
+        "targetGroupArn": "arn:aws:elasticloadbalancing:us-east-1:999999999999:targetgroup/keeldemo-prestaging/4444444444444444",
+        "healthCheckTimeoutSeconds": 5,
+        "healthCheckPort": "8080",
+        "targetType": "instance",
+        "matcher": {
+          "httpCode": "200-299"
+        },
+        "targetGroupName": "keeldemo-prestaging",
+        "healthCheckProtocol": "HTTP",
+        "protocol": "HTTP",
+        "healthCheckPath": "/admin/health",
+        "port": 8080,
+        "healthCheckIntervalSeconds": 10,
+        "healthyThresholdCount": 10,
+        "attributes": {
+          "stickiness.enabled": "false",
+          "deregistration_delay.timeout_seconds": "300",
+          "stickiness.type": "lb_cookie",
+          "stickiness.lb_cookie.duration_seconds": "86400",
+          "slow_start.duration_seconds": "0",
+          "load_balancing.algorithm.type": "round_robin"
+        },
+        "unhealthyThresholdCount": 2
+      }
+    ],
+    "type": "aws",
+    "cloudProvider": "aws",
+    "moniker": {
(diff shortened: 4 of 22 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredSecurityGroupVeto.kt
```diff
@@ -0,0 +1,85 @@
...
+    TODO("not implemented")
+  }
+
+  override fun currentRejections(): List<String> {
+    TODO("not implemented")
+  }
+
+  override fun currentRejectionsByApp(application: String): List<String> {
+    TODO("not implemented")
+  }
```
robfletcher, line 84: Can anyone help me understand what these methods are supposed to do?

emjburns: So, the messages are a generic way to tell a veto to do something. The veto tells you what data it accepts (`messageFormat()`), and you send it that data via `passMessage(...)`. That avoids having to make a new controller for each veto, and could let us do remote calls to vetos. For example, that's how I first implemented pausing an application, and resuming an application.

`currentRejections` is a list of things the veto is rejecting (so you don't have to look at the database). Like, when there was a veto for pausing applications, this would return a list of the applications. `currentRejectionsByApp` returns a list of things the veto is rejecting by app like if the veto was rejecting resources. 

Maybe a lot has changed since I wrote this generic stuff. The idea was to make it easy to add new vetos without having to do a lot of wiring through keel and gate. However, that might have not been worth it.

As it is, you don't have to implement any of those things for the veto to veto stuff. You just won't be able to interact with it.

robfletcher: Okay, it sounds like it would make sense for me to implement `currentRejectionsByApp` but I probably don't need any of the others. Does that make sense?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -42,12 +42,10 @@ import com.netflix.spinnaker.kork.exceptions.SystemException
 import com.netflix.spinnaker.kork.exceptions.UserException
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
-import io.mockk.coEvery
-import io.mockk.coVerify
-import io.mockk.every
+import io.mockk.coEvery as every
+import io.mockk.coVerify as verify
```
robfletcher, line 46: Useful trick if you want to not care whether the mock method is a `suspend` method

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredSecurityGroupVeto.kt
emjburns: I think so. If you don't want the user to interact with the veto (for example, be able to say "i know best please actually check this resource") then that seems right.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/Veto.kt
```diff
@@ -28,32 +28,32 @@ import com.netflix.spinnaker.keel.api.Resource
...
+   * The name of the veto.
    */
   fun name(): String = javaClass.simpleName
 
   /**
-   * Check whether the resource can be checked according to this veto
+   * Check whether the resource can be checked according to this veto.
    */
-  fun check(resource: Resource<*>): VetoResponse
+  suspend fun check(resource: Resource<*>): VetoResponse
```
robfletcher, line 38: This method is always called from a suspending context, so we might as well allow implementations to freely use coroutines.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredSecurityGroupVeto.kt
```diff
@@ -0,0 +1,147 @@
...
+        region,
+        if (vpcName == null) null else cloudDriverCache.networkBy(vpcName, account, region).id
+      )
+    }
+      .onFailure { ex ->
+        if (ex is HttpException && ex.code() == 404) {
+          log.warn("security group $name not found in $account/$vpcName/$region")
+        } else {
+          log.error("error finding security group $name in $account/$vpcName/$region", ex)
+        }
```
robfletcher, line 80: I want to distinguish in the logs between a 404, which is fine, and any other type of error response, which indicates some kind of failure to talk to Clouddriver.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredSecurityGroupVeto.kt
robfletcher: Yeah, I think in this case it doesn't make sense for a user to be able to say "no, I'm 100% sure that security group exists"!

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredSecurityGroupVeto.kt
```diff
@@ -0,0 +1,147 @@
...
+              result[it] = (acc[it] ?: emptyList()) + (map[it] ?: emptyList())
+            }
+            result
+          }
+        SecurityGroupDependencies(
+          securityGroupsInRegions = allSecurityGroups,
+          account = locations.account,
+          subnet = locations.subnet
+        )
+      }
```
robfletcher, line 144: It sure feels like this should be shorter, but my brain is not firing on all cylinders today.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredSecurityGroupVeto.kt
robfletcher: I still think this API seems kinda weird in that each veto has to track its own rejections rather than having that centralized. I don't want to tackle any changes to it here since I'd like to keep this PR focused in scope, and I'm not yet sure exactly how I'd flip that around.

emjburns: Let's talk about this in person, I think we can definitely make this better

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/Veto.kt
```diff
@@ -28,34 +28,34 @@ import com.netflix.spinnaker.keel.api.Resource
...
 
   /**
-   * The message format a veto accepts
+   * The message format a veto accepts.
    */
-  fun messageFormat(): Map<String, Any>
+  fun messageFormat(): Map<String, Any> = TODO("not implemented")
```
lorin, line 43: nice

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
```diff
@@ -79,6 +80,12 @@ interface CloudDriverService {
     @Header("X-SPINNAKER-USER") user: String = DEFAULT_SERVICE_ACCOUNT
   ): Credential
 
+  @GET("/applications/{application}/loadBalancers")
+  suspend fun loadBalancersForApplication(
+    @Header("X-SPINNAKER-USER") user: String,
+    @Path("application") application: String
+  ): List<AmazonLoadBalancer>
```
robfletcher, line 87: This CloudDriver endpoint lists all the load balancers for an app. Because there's no individual endpoint for target groups, we need to grab those from any ALBs returned by this list.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/AmazonLoadBalancer.kt
```diff
@@ -1,11 +1,26 @@
...
+  use = Id.NAME,
+  include = As.PROPERTY,
+  property = "loadBalancerType",
+  defaultImpl = ClassicLoadBalancerModel::class // CloudDriver's response does not specify the loadBalancerType on classic load balancers
+)
+@JsonSubTypes(
+  JsonSubTypes.Type(ClassicLoadBalancerModel::class, name = "classic"),
+  JsonSubTypes.Type(ApplicationLoadBalancerModel::class, name = "application")
+)
 interface AmazonLoadBalancer {
```
robfletcher, line 19: I added this so we can have mixed collections of load balancers

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverServiceTests.kt
```diff
@@ -0,0 +1,51 @@
...
+        }
+      }
+
+      test("can handle a mixture of application and classic load balancers") {
+        val results = runBlocking {
+          service.loadBalancersForApplication(DEFAULT_SERVICE_ACCOUNT, "keeldemo")
+        }
+        expectThat(results).hasSize(10)
+      }
+    }
```
robfletcher, line 49: This is just testing my Jackson config is doing the right thing when we encounter a mix of classic and application load balancers in a response.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/Veto.kt
robfletcher: I don't love the existence of these methods. They seem to be "optional" for implementations which tells me we're missing a layer of abstraction. I didn't want to change too much at once though.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.netflix.spinnaker.keel.api.Locatable
+import com.netflix.spinnaker.keel.api.ResourceSpec
+import com.netflix.spinnaker.keel.api.SubnetAwareLocations
+import com.netflix.spinnaker.keel.core.api.ClusterDependencies
+
+/**
+ * A [ResourceSpec] that has default and override cluster dependencies (i.e. security groups, load
+ * balancers, and/or target groups).
+ */
+interface OverrideableClusterDependencyContainer<T : ClusterDependencyContainer> :
```
robfletcher, line 12: Please help me think of a better name for this

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
```diff
@@ -0,0 +1,95 @@
...
+    val spec = resource.spec as? OverrideableClusterDependencyContainer<*>
+      ?: return allowedResponse()
+
+    // avoid making call to CloudDriver to get all the load balancers if we don't need to
+    if (spec.loadBalancersByRegion.isEmpty() && spec.targetGroupsByRegion.isEmpty()) {
+      return allowedResponse()
+    }
+
+    val loadBalancers = loadBalancers(spec)
+    val missingLoadBalancers = loadBalancers.checkLoadBalancers(spec)
```
robfletcher, line 35: This works a bit differently to the security group veto as I'm getting all the load balancers at once. There are a couple of reasons:

1. For ALBs we actually specify the target group rather than the load balancer name, and there's no endpoint to just get target groups and no way for us to know the name of the ALB they belong to without just going through them all.
2. Security groups may not belong to the same app, load balancers have to, so we need to check security groups individually.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
```diff
@@ -0,0 +1,74 @@
...
+  Locatable<SubnetAwareLocations> {
+  val defaults: T
+  val overrides: Map<String, T>
```
lorin, line 15: I recommend adding a comment to indicate that the key is a region name (if that is, indeed, the case).

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
```diff
@@ -0,0 +1,74 @@
...
+  val regions = locations.regions.map { it.name }.toSet()
+  val defaults = (defaults.dependencies?.let(fn) ?: emptySet())
+    .associateWith { regions }
+  val overrides = overrides
+    .map { (region, spec) ->
+      spec.dependencies?.let(fn)?.associateWith { setOf(region) } ?: emptyMap()
+    }
+    .zip()
+  return listOf(defaults, overrides)
+    .zip()
```
lorin, line 50: I found the name "zip" confusing here because I expect a function named `zip` to return a list of pairs. (It took me a while to find the custom zip function was defined just below).

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * Reduces a collection of `Map<*, Set>` down to a single map where the values contain the combined
+ * set of values from all elements in the original collection with a common key.
+ */
+private fun <K, VE> Collection<Map<K, Set<VE>>>.zip(): Map<K, Set<VE>> =
+  reduceOrDefault(emptyMap()) { acc, map ->
```
gal-yardeni, line 58: what does `acc` stands for?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
```diff
@@ -0,0 +1,74 @@
...
+      result[it] = (acc[it] ?: emptySet()) + (map[it] ?: emptySet())
+    }
+    result
+  }
+
+/**
+ * @return [defaultValue] if this collection is empty, otherwise as [reduce].
+ */
+private inline fun <S, T : S> Collection<T>.reduceOrDefault(defaultValue: S, operation: (acc: S, T) -> S): S =
+  if (isEmpty()) {
```
lorin, line 70: I think this method is equivalent to:

```kotlin
fold(defaultValue, operation)
```

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
lorin: My guess is `accumulator`. It's a common name for the first argument of a `reduce` or `fold` function.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
```diff
@@ -0,0 +1,95 @@
...
+    } else {
+      deniedResponse(
+        message = missingLoadBalancers.joinToString(separator = "\n", transform = MissingDependency::message),
+        vetoArtifact = false
+      )
+    }
+  }
+
+  private suspend fun loadBalancers(spec: OverrideableClusterDependencyContainer<*>) =
+    withContext(IO) {
```
lorin, line 48: This is the first time I've seen the `IO` context used in keel. Should we be using it in more places?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
robfletcher: Accumulator. I think I just inherited that from the Kotlin method I'm implementing

keel-clouddriver/src/test/resources/loadBalancers.json
```diff
@@ -0,0 +1,1357 @@
+[
+  {
+    "account": "test",
+    "name": "keeldemo-main-clb",
+    "region": "us-east-1",
+    "vpcId": "vpc-46f5a223",
```
luispollo, line 6: You might want to obfuscate real AWS IDs for potentially sensitive things like VPCs, Route53 hosted zones, as well as ELB hostnames.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
luispollo: Will the compiler resolve the original `every` and `verify` OK in the case where you have a mix of suspending and non-suspending functions?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
```diff
@@ -0,0 +1,95 @@
...
+      runCatching {
+        cloudDriver.loadBalancersForApplication(DEFAULT_SERVICE_ACCOUNT, spec.application)
+      }
+        .onFailure { ex ->
+          log.error("error finding load balancers for ${spec.application}", ex)
+        }
+        .getOrDefault(emptyList())
+    }
+
+  private suspend fun List<AmazonLoadBalancer>.checkLoadBalancers(spec: OverrideableClusterDependencyContainer<*>): Collection<MissingDependency> {
```
lorin, line 58: I think something like `missingLoadBalancers` would be a better name here.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/veto/UnhappyVetoTests.kt
```diff
@@ -80,17 +83,17 @@ class UnhappyVetoTests : JUnit5Minutests {
...
-          expectThat(subject.check(r).allowed).isEqualTo(false)
+          expectThat(check().allowed).isEqualTo(false)
         }
 
         test("resources are checked once every wait time") {
           unhappyRepository.markUnhappyForWaitingTime(r.id, r.spec.application)
 
-          val response1 = subject.check(r)
+          val response1 = check()
           clock.incrementBy(Duration.ofMinutes(11))
```
luispollo, line 93: Just FYI, I added `MutableClock.tickMinutes(minutes)` to make this a bit more convenient.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
luispollo: I think this is fine.

lorin: :shrug:

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
robfletcher: d'oh, yes it is

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
robfletcher: My understanding is that any time we're doing blocking IO (as in here, where I'm calling CloudDriver), that's the appropriate one to use.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
robfletcher: Will rename to `merge`?

luispollo: I think this would be much more readable (albeit perhaps less "kotlinesque") if you removed `zip()` and did something like:
```kotlin
val result = defaults.toMutableMap()
overrides.forEach { (dep, regions} ->
  if (result.contains(dep)) {
    result[dep].add(regions)
  } else {
    result[dep] = regions
  }
}
```

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
luispollo: Reference(s), please?

keel-clouddriver/src/test/resources/loadBalancers.json
robfletcher: I've obscured it all now

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
robfletcher: Yes. Effectively you're _always_ using the suspending ones. It makes no difference when the mock function is non-suspending. I raised this with the library author a while back but he said he wasn't very knowledgeable about coroutines and had assumed there might be some performance penalty. I don't believe that is the case, and even if there were it's unlikely to amount to anything given the number of calls in a unit test scenario.

luispollo: Got it. Yeah, seems like a nice simplification for the author to apply. Thanks for the tip!

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
robfletcher: Javadoc on `IO`

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
```diff
@@ -0,0 +1,95 @@
...
+      runCatching {
+        cloudDriver.loadBalancersForApplication(DEFAULT_SERVICE_ACCOUNT, spec.application)
+      }
+        .onFailure { ex ->
+          log.error("error finding load balancers for ${spec.application}", ex)
+        }
+        .getOrDefault(emptyList())
+    }
+
+  private suspend fun List<AmazonLoadBalancer>.findMissingLoadBalancers(spec: OverrideableClusterDependencyContainer<*>): Collection<MissingDependency> {
```
luispollo, line 58: Technically, this is looking for dependencies that are not just load balancers. Consider renaming to `findMissingDependencies`.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
```diff
@@ -0,0 +1,95 @@
...
+    val spec = resource.spec as? OverrideableClusterDependencyContainer<*>
+      ?: return allowedResponse()
+
+    // avoid making call to CloudDriver to get all the load balancers if we don't need to
+    if (spec.loadBalancersByRegion.isEmpty() && spec.targetGroupsByRegion.isEmpty()) {
+      return allowedResponse()
+    }
+
+    val loadBalancers = loadBalancers(spec)
+    val missingLoadBalancers = loadBalancers.findMissingLoadBalancers(spec)
```
luispollo, line 35: See comment below -- consider renaming this variable to `missingDependencies` for consistency with the function if you rename it as suggested.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredSecurityGroupVeto.kt
```diff
@@ -0,0 +1,115 @@
...
+
+  override suspend fun check(resource: Resource<*>) =
+    resource.spec.toSecurityGroupDependencies()
+      ?.let {
+        val missingSecurityGroupRegions = checkSecurityGroups(it)
+        VetoResponse(
+          allowed = missingSecurityGroupRegions.isEmpty(),
+          vetoName = name(),
+          message = missingSecurityGroupRegions.entries.joinToString(separator = "\n") { (securityGroup, missingRegions) ->
+            "Security group $securityGroup is not found in ${missingRegions.joinToString()}"
```
luispollo, line 36: Did you want to separate the regions with commas here?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredSecurityGroupVeto.kt
```diff
@@ -0,0 +1,115 @@
...
+        user = DEFAULT_SERVICE_ACCOUNT,
+        account = account,
+        type = CLOUD_PROVIDER,
+        securityGroupName = name,
+        region = region,
+        vpcId = if (vpcName == null) null else cloudDriverCache.networkBy(vpcName, account, region).id
+      )
+    }
+      .onFailure { ex ->
+        if (ex is HttpException && ex.code() == 404) {
```
luispollo, line 82: Is this a subclass of `RetrofitError` when the error is at the HTTP layer? I hadn't seen this before, but I ask because I could simplify code I wrote elsewhere to do similar things...

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+    .merge()
+  return listOf(defaults, overrides)
+    .merge()
+}
+
+/**
+ * Reduces a collection of `Map<*, Set>` down to a single map where the values contain the combined
+ * set of values from all elements in the original collection with a common key.
+ */
```
emjburns, line 60: I feel like the things in this file could benefit from some typing. It's pretty hard for me to reason about `Map<String, Set<String>>` etc. Could you create a data class to make working with this a little easier?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
robfletcher: Well, it's CLBs and ALB target groups. Not really any other type of dependency.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredSecurityGroupVeto.kt
robfletcher: Default separator on `joinToString` is `", "` so it will

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredSecurityGroupVeto.kt
robfletcher: Yeah, it is. I think it's any `RetrofitError` where you got an HTTP status code (i.e. not a socket closing or that kind of thing).

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterDependencyContainer.kt
robfletcher: I think that would really only be able to replace the `Map` part but I'll give it a shot & see if it makes things clearer.

PR #1041, opened 2020-04-18
fix(export): Fix launch config export

### LINKED ISSUE #918: Export API omitting Instance Type for EC2 Clusters
No `Instance Type` is included in an EC2 cluster export. If no instance type is specified, Managed Delivery is unable to obtain the desired state.

Steps: 

1. I created a test app created through a standard newt app-type (java-project)
1. I used pipelines to spin up server groups, then disabled the pipelines
1. I used `newt export` to create a new config for this app
    1. Exported clusters didn’t include a `launchconfig` block with a defined `instance type`
1. I published the config using `newt publish`
1. Cluster showed an error, indicating that `instance type` was missing and no default exists

Screen Shot 2020-03-25 at 9 49 43 AM

Work-around:

Manually edit the delivery configuration to include a launchConfiguration block with an instance type detail.

luispollo:
The code to export a launch config was including the server group's own instance type to compare with itself, which... er... would always be the same, and therefore get omitted. 🤦 

Fixes #918

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -922,7 +922,7 @@ class ClusterHandler(
     )
 
     val thisSpec = ServerGroupSpec(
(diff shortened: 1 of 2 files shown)


emjburns:
Nice find :)

PR #1050, opened 2020-04-20
feat(configs): fetch config by app

emjburns:
I keep wanting to look up a delivery config by app instead of by manifest name. So I made that api.

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
@@ -17,6 +17,7 @@
  */
 package com.netflix.spinnaker.keel.rest
 
+import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.constraints.ConstraintState
 import com.netflix.spinnaker.keel.constraints.UpdatedConstraintStatus
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactPin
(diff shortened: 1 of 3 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -81,6 +82,15 @@ class ApplicationController(
     }
   }
 
+  @GetMapping(
+    path = ["/{application}/config"],
+    produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  @PreAuthorize("""@authorizationSupport.hasApplicationPermission('READ', 'APPLICATION', #application)""")
```
luispollo, line 89: I think in `DeliveryConfigController` we also added `READ` permission to the account (because the delivery config container cloud resources).

luispollo: The other thought that occurred to me is that maybe we could just add a fallback to `DeliveryConfigController.get()` to lookup the delivery config by application name if we can't find it by delivery config name? Then we wouldn't need a separate API, which I suspect may be confusing to users.

emjburns: I was thinking that you should be able to see the config, even if you can't see the resources. That way it'll promote easy finding and sharing of examples. Also, I was thinking we might want to get rid of the "look up by manifest name" endpoint, because we are experimenting with phasing out the name

luispollo: That sounds reasonable, but remember that we did not version the API endpoints (yet?), and we’ve broken the API contract before. My number 1 concern is newt. I’m not trying to play API 👮🏻‍♂️, but I think we should start following the golden rule of never breaking contracts without a deprecation warning as we approach beta. Hence my suggestion of sticking with the original endpoint. I’m not opposed to deprecating it, but we need to coordinate with newt.

emjburns: No one is using the old api (get by manifest name), so it's safe to remove. However I'm going to leave it for now since it's not causing anyone harm.

PR #1055, opened 2020-04-21
feat(export): Export deployment strategy

### LINKED ISSUE #625: Default options for red-black strategy don't get omitted during export
When you export a cluster with all the default red/black options, the fields still make it into the spec. e.g.:

```yaml
deployWith:
  delayBeforeDisable: PT0S
  delayBeforeScaleDown: PT0S
  maxServerGroups: 2
  resizePreviousToZero: false
  rollbackOnFailure: true
  strategy: red-black
```

### LINKED ISSUE #753: /managed/resources/export exports incorrect deployWith strategy
My cluster is currently managed with:
```
     deployWith:
       strategy: highlander
```
but if I export the same cluster it returns with the default redblack:
```
      deployWith:
        delayBeforeDisable: PT0S
        delayBeforeScaleDown: PT0S
        maxServerGroups: 2
        resizePreviousToZero: false
        rollbackOnFailure: true
        strategy: red-black
```

The export should honor the currently managed deployWith strategy.

Ideally we would be able to inspect current deployment pipelines for clusters to infer the correct strategy when exporting a non-managed cluster to be managed, but for now we need it to be consistent with the currently managed strategy.

luispollo:
Adds support for exporting the deployment strategy associated with a cluster based on information retrieved from pipelines (for not-yet-migrated apps) and tasks (for already-managed apps or clusters created manually in the UI).

Limitations: currently, the export supports `createServerGroup` tasks and `deploy` pipeline stages only. Support for clone operations, if necessary, will be addressed in a separate PR.

Closes #753. As I was working in this area of the code anyway, I've taken the opportunity to also fix #625.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
@@ -28,15 +28,17 @@ sealed class ClusterDeployStrategy {
   open val isStaggered: Boolean = false
   @get:JsonInclude(NON_EMPTY)
   open val stagger: List<StaggeredRegion> = emptyList()
+  @get:JsonIgnore
+  abstract val defaults: ClusterDeployStrategy
 }
 
 @JsonTypeName("red-black")
 data class RedBlack(
-  val rollbackOnFailure: Boolean = true,
-  val resizePreviousToZero: Boolean = false,
-  val maxServerGroups: Int = 2,
-  val delayBeforeDisable: Duration = ZERO,
-  val delayBeforeScaleDown: Duration = ZERO,
+  val rollbackOnFailure: Boolean? = true,
+  val resizePreviousToZero: Boolean? = false,
+  val maxServerGroups: Int? = 2,
+  val delayBeforeDisable: Duration? = ZERO,
+  val delayBeforeScaleDown: Duration? = ZERO,
   // The order of this list is important for pauseTime based staggers
   @JsonInclude(NON_EMPTY)
   override val stagger: List<StaggeredRegion> = emptyList()
@@ -53,19 +55,24 @@ data class RedBlack(
         delayBeforeDisable = Duration.ofSeconds((context["delayBeforeDisableSec"] as Int).toLong()),
         delayBeforeScaleDown = Duration.ofSeconds((context["delayBeforeScaleDownSec"] as Int).toLong())
       )
+
+    val DEFAULTS = RedBlack()
   }
+
+  override val defaults: RedBlack
+    get() = DEFAULTS
 
   override fun toOrcaJobProperties() = mapOf(
     "strategy" to "redblack",
     "maxRemainingAsgs" to maxServerGroups,
-    "delayBeforeDisableSec" to delayBeforeDisable.seconds,
-    "delayBeforeScaleDownSec" to delayBeforeScaleDown.seconds,
+    "delayBeforeDisableSec" to delayBeforeDisable?.seconds,
+    "delayBeforeScaleDownSec" to delayBeforeScaleDown?.seconds,
     "scaleDown" to resizePreviousToZero,
     "rollback" to mapOf("onFailure" to rollbackOnFailure)
   )
 
   override val isStaggered: Boolean
-    get() = stagger.isNotEmpty()
+    get() = stagger?.isNotEmpty() ?: false
 }
 
 @JsonTypeName("highlander")
@@ -73,6 +80,9 @@ object Highlander : ClusterDeployStrategy() {
   override fun toOrcaJobProperties() = mapOf(
     "strategy" to "highlander"
   )
+
+  override val defaults: Highlander
+    get() = this
 }
 
 /**
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
@@ -23,13 +23,12 @@ import java.time.Duration.ZERO
   Type(Highlander::class)
 )
 sealed class ClusterDeployStrategy {
-  abstract fun toOrcaJobProperties(): Map<String, Any?>
   @get:JsonIgnore
   open val isStaggered: Boolean = false
   @get:JsonInclude(NON_EMPTY)
   open val stagger: List<StaggeredRegion> = emptyList()
-  @get:JsonIgnore
-  abstract val defaults: ClusterDeployStrategy
+  abstract fun toOrcaJobProperties(): Map<String, Any?>
+  abstract fun withDefaultsOmitted(): ClusterDeployStrategy
 }
 
 @JsonTypeName("red-black")
@@ -59,9 +58,6 @@ data class RedBlack(
     val DEFAULTS = RedBlack()
   }
 
-  override val defaults: RedBlack
-    get() = DEFAULTS
-
   override fun toOrcaJobProperties() = mapOf(
     "strategy" to "redblack",
     "maxRemainingAsgs" to maxServerGroups,
@@ -73,6 +69,35 @@ data class RedBlack(
 
   override val isStaggered: Boolean
     get() = stagger?.isNotEmpty() ?: false
+
+  override fun withDefaultsOmitted() =
+    RedBlack(
+      maxServerGroups = if (maxServerGroups == DEFAULTS.maxServerGroups) {
+        null
+      } else {
+        maxServerGroups
+      },
+      delayBeforeDisable = if (delayBeforeDisable == DEFAULTS.delayBeforeDisable) {
+        null
+      } else {
+        delayBeforeDisable
+      },
+      delayBeforeScaleDown = if (delayBeforeScaleDown == DEFAULTS.delayBeforeScaleDown) {
+        null
+      } else {
+        delayBeforeScaleDown
+      },
+      resizePreviousToZero = if (resizePreviousToZero == DEFAULTS.resizePreviousToZero) {
+        null
+      } else {
+        resizePreviousToZero
+      },
+      rollbackOnFailure = if (rollbackOnFailure == DEFAULTS.rollbackOnFailure) {
+        null
+      } else {
+        rollbackOnFailure
+      }
+    )
 }
 
 @JsonTypeName("highlander")
@@ -81,8 +106,7 @@ object Highlander : ClusterDeployStrategy() {
     "strategy" to "highlander"
   )
 
-  override val defaults: Highlander
-    get() = this
+  override fun withDefaultsOmitted() = this
 }
 
 /**
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
@@ -41,6 +41,20 @@ data class RedBlack(
   @JsonInclude(NON_EMPTY)
   override val stagger: List<StaggeredRegion> = emptyList()
 ) : ClusterDeployStrategy() {
+
+  companion object {
+    fun fromOrcaStageContext(context: Map<String, Any?>) =
+      RedBlack(
+        rollbackOnFailure = context["rollback"]
+          ?.let { it as Map<String, Any> }
+          ?.get("onFailure") as Boolean,
+        resizePreviousToZero = context["scaleDown"] as Boolean,
+        maxServerGroups = context["maxRemainingAsgs"] as Int,
+        delayBeforeDisable = Duration.ofSeconds((context["delayBeforeDisableSec"] as Int).toLong()),
+        delayBeforeScaleDown = Duration.ofSeconds((context["delayBeforeScaleDownSec"] as Int).toLong())
+      )
+  }
+
   override fun toOrcaJobProperties() = mapOf(
     "strategy" to "redblack",
     "maxRemainingAsgs" to maxServerGroups,
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
@@ -68,35 +68,15 @@ data class RedBlack(
   )
 
   override val isStaggered: Boolean
-    get() = stagger?.isNotEmpty() ?: false
+    get() = stagger.isNotEmpty() ?: false
 
   override fun withDefaultsOmitted() =
     RedBlack(
-      maxServerGroups = if (maxServerGroups == DEFAULTS.maxServerGroups) {
-        null
-      } else {
-        maxServerGroups
-      },
-      delayBeforeDisable = if (delayBeforeDisable == DEFAULTS.delayBeforeDisable) {
-        null
-      } else {
-        delayBeforeDisable
-      },
-      delayBeforeScaleDown = if (delayBeforeScaleDown == DEFAULTS.delayBeforeScaleDown) {
-        null
-      } else {
-        delayBeforeScaleDown
-      },
-      resizePreviousToZero = if (resizePreviousToZero == DEFAULTS.resizePreviousToZero) {
-        null
-      } else {
-        resizePreviousToZero
-      },
-      rollbackOnFailure = if (rollbackOnFailure == DEFAULTS.rollbackOnFailure) {
-        null
-      } else {
-        rollbackOnFailure
-      }
+      maxServerGroups = nullIfDefault(maxServerGroups, DEFAULTS.maxServerGroups),
+      delayBeforeDisable = nullIfDefault(delayBeforeDisable, DEFAULTS.delayBeforeDisable),
+      delayBeforeScaleDown = nullIfDefault(delayBeforeScaleDown, DEFAULTS.delayBeforeScaleDown),
+      resizePreviousToZero = nullIfDefault(resizePreviousToZero, DEFAULTS.resizePreviousToZero),
+      rollbackOnFailure = nullIfDefault(rollbackOnFailure, DEFAULTS.rollbackOnFailure)
     )
 }
 
@@ -144,3 +124,6 @@ data class StaggeredRegion(
   val allowedHours: Set<Int>
     get() = AllowedTimesConstraintEvaluator.parseHours(hours)
 }
+
+private fun <T> nullIfDefault(value: T, default: T): T? =
+  if (value == default) null else value
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceSpecExportHelper.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceSpecExportHelper.kt
@@ -1,6 +1,7 @@
 package com.netflix.spinnaker.keel.plugin
 
 import com.netflix.spinnaker.keel.diff.DefaultResourceDiff
+import kotlin.reflect.KClass
 import kotlin.reflect.full.memberProperties
 import kotlin.reflect.full.primaryConstructor
 
@@ -59,4 +60,14 @@ inline fun <reified T : Any, reified S : Any> buildSpecFromDiff(
   } else {
     ctor.callBy(params)
   }
+}
+
+/**
+ * Helper function to return an instance of any type with all nullable constructor parameters set to null.
+ * Note that this call will fail if any remaining parameters to the primary constructor are not optional.
+ */
+fun <T : Any> emptyShell(klass: KClass<T>): T {
+  val ctor = klass.primaryConstructor!!
+  val params = ctor.parameters.filter { it.type.isMarkedNullable }.associateWith { null }
+  return ctor.callBy(params)
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceSpecExportHelper.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceSpecExportHelper.kt
@@ -1,7 +1,6 @@
 package com.netflix.spinnaker.keel.plugin
 
 import com.netflix.spinnaker.keel.diff.DefaultResourceDiff
-import kotlin.reflect.KClass
 import kotlin.reflect.full.memberProperties
 import kotlin.reflect.full.primaryConstructor
 
@@ -60,14 +59,4 @@ inline fun <reified T : Any, reified S : Any> buildSpecFromDiff(
   } else {
     ctor.callBy(params)
   }
-}
-
-/**
- * Helper function to return an instance of any type with all nullable constructor parameters set to null.
- * Note that this call will fail if any remaining parameters to the primary constructor are not optional.
- */
-fun <T : Any> emptyShell(klass: KClass<T>): T {
-  val ctor = klass.primaryConstructor!!
-  val params = ctor.parameters.filter { it.type.isMarkedNullable }.associateWith { null }
-  return ctor.callBy(params)
 }
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -313,7 +313,7 @@ class ClusterHandler(
     }
 
   override suspend fun export(exportable: Exportable): ClusterSpec {
-    // Get existing infrastructure -- this is a very costly call
+    // Get existing infrastructure
     val serverGroups = cloudDriverService.getServerGroups(
       account = exportable.account,
       moniker = exportable.moniker,
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -65,6 +65,7 @@ import com.netflix.spinnaker.keel.orca.dependsOn
 import com.netflix.spinnaker.keel.orca.restrictedExecutionWindow
 import com.netflix.spinnaker.keel.orca.waitStage
 import com.netflix.spinnaker.keel.plugin.buildSpecFromDiff
+import com.netflix.spinnaker.keel.plugin.emptyShell
 import com.netflix.spinnaker.keel.retrofit.isNotFound
 import com.netflix.spinnaker.keel.serialization.configuredObjectMapper
 import com.netflix.spinnaker.kork.exceptions.SystemException
@@ -316,7 +317,7 @@ class ClusterHandler(
     }
 
   override suspend fun export(exportable: Exportable): ClusterSpec {
-    // Get existing infrastructure
+    // Get existing infrastructure -- this is a very costly call
     val serverGroups = cloudDriverService.getServerGroups(
       account = exportable.account,
       moniker = exportable.moniker,
@@ -374,7 +375,7 @@ class ClusterHandler(
         null
       },
       locations = locations,
-      deployWith = base.discoverDeploymentStrategy() ?: RedBlack(),
+      deployWith = (base.discoverDeploymentStrategy() ?: RedBlack()).let { it.withDefaultsOmitted() },
       _defaults = base.exportSpec(exportable.account, exportable.moniker.app),
       overrides = mutableMapOf()
     )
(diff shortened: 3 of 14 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
```diff
@@ -23,42 +23,90 @@ import java.time.Duration.ZERO
...
-  val rollbackOnFailure: Boolean = true,
-  val resizePreviousToZero: Boolean = false,
-  val maxServerGroups: Int = 2,
-  val delayBeforeDisable: Duration = ZERO,
-  val delayBeforeScaleDown: Duration = ZERO,
+  val rollbackOnFailure: Boolean? = true,
+  val resizePreviousToZero: Boolean? = false,
+  val maxServerGroups: Int? = 2,
+  val delayBeforeDisable: Duration? = ZERO,
+  val delayBeforeScaleDown: Duration? = ZERO,
```
luispollo, line 40: I had to make these nullable in order to fix #625

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
```diff
@@ -23,42 +23,90 @@ import java.time.Duration.ZERO
...
 data class RedBlack(
-  val rollbackOnFailure: Boolean = true,
-  val resizePreviousToZero: Boolean = false,
-  val maxServerGroups: Int = 2,
-  val delayBeforeDisable: Duration = ZERO,
-  val delayBeforeScaleDown: Duration = ZERO,
+  val rollbackOnFailure: Boolean? = true,
+  val resizePreviousToZero: Boolean? = false,
+  val maxServerGroups: Int? = 2,
+  val delayBeforeDisable: Duration? = ZERO,
```
robfletcher, line 39: Why do these need to be nullable now?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
```diff
@@ -23,42 +23,90 @@ import java.time.Duration.ZERO
...
     "scaleDown" to resizePreviousToZero,
     "rollback" to mapOf("onFailure" to rollbackOnFailure)
   )
 
   override val isStaggered: Boolean
-    get() = stagger.isNotEmpty()
+    get() = stagger?.isNotEmpty() ?: false
+
+  override fun withDefaultsOmitted() =
+    RedBlack(
```
luispollo, line 74: I tried to use my `ResourceSpecExportHelper` to nullify defaults on these objects, like we do for other types, but the challenge is that, in this case, we want the type-identifying field `strategy` to always be present, even if the other fields are all `null`. After spending an hour trying to fiddle with that code, I gave up and went for the simple thing that works, even though it hurts my 👀 to look at it. Oh well. 🤷‍♂️

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/plugin/ResourceSpecExportHelper.kt
```diff
@@ -60,3 +61,13 @@ inline fun <reified T : Any, reified S : Any> buildSpecFromDiff(
     ctor.callBy(params)
   }
 }
+
+/**
+ * Helper function to return an instance of any type with all nullable constructor parameters set to null.
+ * Note that this call will fail if any remaining parameters to the primary constructor are not optional.
+ */
+fun <T : Any> emptyShell(klass: KClass<T>): T {
```
luispollo, line 69: Oops, this is a left-over from my attempt to fix the specifics of nullifying only some fields of `ClusterDeployStrategy`. Will remove.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -310,7 +316,7 @@ class ClusterHandler(
     }
 
   override suspend fun export(exportable: Exportable): ClusterSpec {
-    // Get existing infrastructure
+    // Get existing infrastructure -- this is a very costly call
```
luispollo, line 319: I added this note because it routinely takes more than 30 seconds to retrieve the details for a single server group from clouddriver, which is insane... I'm not sure we could do much from our end, since it's not really information we would want to cache... Or maybe we do?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -382,11 +389,106 @@ class ClusterHandler(
...
+          .isNotEmpty()
+      }
+
+  /**
+   * Retrieve server group entity tags from clouddriver to identity the "source" of the server group (pipeline or
+   * orchestration), then retrieve execution details from orca to determine the deployment strategy used for that
+   * server group.
+   */
+  private fun ServerGroup.discoverDeploymentStrategy(): ClusterDeployStrategy? {
+    val entityTags = runBlocking {
```
luispollo, line 410: First step is to look for the `spinnaker:metadata` entity tag.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -382,11 +389,106 @@ class ClusterHandler(
...
+      spinnakerMetadata["executionId"] == null
+    ) {
+      log.warn("Unable to find Spinnaker metadata for server group $name in application ${moniker.app}, " +
+        "account ${location.account} in entity tags.")
+      return null
+    }
+
+    val executionType = spinnakerMetadata["executionType"]!!.toString()
+    val executionId = spinnakerMetadata["executionId"]!!.toString()
+    val execution = runBlocking {
```
luispollo, line 449: Once found, we look up the corresponding execution in orca.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
robfletcher, line 410: Can we do this without blocking?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -382,11 +389,106 @@ class ClusterHandler(
...
+      when (executionType) {
+        "orchestration" -> orcaService.getOrchestrationExecution(executionId)
+        "pipeline" -> orcaService.getPipelineExecution(executionId)
+        else -> throw SystemException("Unsupported execution type $executionType in Spinnaker metadata.")
+      }
+    }
+
+    // get context from the appropriate execution stage for the execution type, drilling down into the data as needed
+    val context = if (executionType == "pipeline") {
```
luispollo, line 458: This set of statements finds the right spot in the data structure (which is different between a task execution and a pipeline execution) to extract the context data we need.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -382,11 +389,106 @@ class ClusterHandler(
...
+      ?.let { context ->
+        if (executionType == "pipeline") {
+          (context["clusters"] as List<Map<String, Any>>?)?.first()
+        } else {
+          context
+        }
+      }
+
+    return when (val strategy = context?.get("strategy")) {
+      "redblack" -> RedBlack.fromOrcaStageContext(context)
```
luispollo, line 482: Finally, once we've found the strategy + other context bits, convert those into an instance of the corresponding `ClusterDeployStrategy` sub-type.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -400,18 +502,6 @@ class ClusterHandler(
       }
     }
 
-  override suspend fun actuationInProgress(resource: Resource<ClusterSpec>) =
```
luispollo, line 403: Just moved up above to come before the private extension functions.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
luispollo: Sorry @robfletcher, I was still adding some additional inline comments: https://github.com/spinnaker/keel/pull/1055#discussion_r412556993

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: Sure.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -382,10 +388,107 @@ class ClusterHandler(
...
+
+      if (entityTags.isEmpty()) {
+        log.warn("Unable to find entity tags for server group $name in application ${moniker.app}, " +
+          "account ${location.account}.")
+        return@coroutineScope null
+      }
+
+      val spinnakerMetadata = entityTags.first().tags
+        .find { it.name == "spinnaker:metadata" }
+        ?.let {
```
lorin, line 430: I think this let block can be simplified using the [as?](https://kotlinlang.org/docs/reference/typecasts.html#safe-nullable-cast-operator) operator

```kotlin
let { it.value as? Map<*, *> }
```

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -382,10 +388,107 @@ class ClusterHandler(
...
+          application = moniker.app,
+          entityType = "servergroup",
+          entityId = name
+        )
+      }.await()
+
+      if (entityTags.isEmpty()) {
+        log.warn("Unable to find entity tags for server group $name in application ${moniker.app}, " +
+          "account ${location.account}.")
+        return@coroutineScope null
```
lorin, line 425: What's the difference in behavior between `return` and `return@coroutineScope`?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -382,10 +388,107 @@ class ClusterHandler(
...
+      }
+
+  /**
+   * Retrieve server group entity tags from clouddriver to identity the "source" of the server group (pipeline or
+   * orchestration), then retrieve execution details from orca to determine the deployment strategy used for that
+   * server group.
+   */
+  private suspend fun ServerGroup.discoverDeploymentStrategy(): ClusterDeployStrategy? {
+    return coroutineScope {
+      val entityTags = async {
```
lorin, line 410: Note that IntelliJ Is flagging the two `async` blocks in this method with the warning: _Redundant 'async' call may be reduced to 'kotlinx.coroutines.withContext'_

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -382,10 +388,107 @@ class ClusterHandler(
...
+      val execution = async {
+        when (executionType) {
+          "orchestration" -> orcaService.getOrchestrationExecution(executionId)
+          "pipeline" -> orcaService.getPipelineExecution(executionId)
+          else -> throw SystemException("Unsupported execution type $executionType in Spinnaker metadata.")
+        }
+      }.await()
+
+      // get context from the appropriate execution stage for the execution type, drilling down into the data as needed
+      val context = if (executionType == "pipeline") {
```
lorin, line 458: Given the differences in logic for pipelines vs tasks, I would find it easier to read if there were separate `pipeline` and `task` helper functions, rather than having multiple branches in the method.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -382,10 +388,107 @@ class ClusterHandler(
...
+        // TODO: or clone?
+        execution.stages?.find { stage -> stage["type"] == "deploy" }
+      } else { // orchestration (i.e. a task)
+        // TODO: or cloneServerGroup?
+        execution.execution.stages?.find { stage -> stage["type"] == "createServerGroup" }
+      }
+        ?.let { stage -> stage["context"] }
+        ?.let { context ->
+          if (context is Map<*, *>) {
```
lorin, line 467: Another place where `as?` operator could be used.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
```diff
@@ -23,42 +23,90 @@ import java.time.Duration.ZERO
...
-    "delayBeforeScaleDownSec" to delayBeforeScaleDown.seconds,
+    "delayBeforeDisableSec" to delayBeforeDisable?.seconds,
+    "delayBeforeScaleDownSec" to delayBeforeScaleDown?.seconds,
     "scaleDown" to resizePreviousToZero,
     "rollback" to mapOf("onFailure" to rollbackOnFailure)
   )
 
   override val isStaggered: Boolean
-    get() = stagger.isNotEmpty()
+    get() = stagger?.isNotEmpty() ?: false
```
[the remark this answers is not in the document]
luispollo: Ah, good catch! I reverted the nullability change for `stagger` and forgot to undo this. Will fix.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
```diff
@@ -23,42 +23,90 @@ import java.time.Duration.ZERO
...
+    )
 }
 
 @JsonTypeName("highlander")
 object Highlander : ClusterDeployStrategy() {
   override fun toOrcaJobProperties() = mapOf(
     "strategy" to "highlander"
   )
+
+  override fun withDefaultsOmitted() = this
```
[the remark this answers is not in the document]
luispollo: So the `=` part is just a shorthand-form for a return type + function body when the return type can be inferred from the expression following (`this`, in this case). The `this` part is because, in this case, `Highlander` is not actually a class, but an `object` (singleton instance), and it has no properties. If it was a class, `this` would be replaced with `Highlander()` to instantiate an object. Make sense?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: Good question... One of my goals for the week is to take some time to teach myself metrics and Atlas (for real this time, unlike during onboarding 😝), so I could definitely add one here.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: Yeah, it did for me too, but I thought that form was much less readable/obvious than `async`.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: I had `return` originally but after adding `coroutineScope` up top I could no longer use that and had to specify the return point. I could rewrite this entire block as `return coroutineScope { ... }` without any returns inside, but the function is relatively long and I felt it would be less readable.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: TIL! Thanks!

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: Makes sense. Will refactor.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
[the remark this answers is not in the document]
lorin: Yeah, makes sense. That was a little mind-bendy.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
erikmunson: > it routinely takes more than 30 seconds to retrieve the details for a single server group from clouddriver

wut? unless we're talking like p99 latency, that seems unbelievably wrong and bad. do you have a sense of what percentile 'routinely' maps to, and if it's not something crazy high like p99 have you considered reaching out to our platform team folks to figure out why we're seeing such unreasonable latency on a regular basis?

luispollo: Routinely as in > 90% of the calls. I believe I did reach out in chat about this at one point, and meant to do it again when I was testing this, but ended up forgetting. Will do.

luispollo: Update: collected some data from the logs and it turns out it's not a single call, but a combination of dozens of calls with near-second response times. But it looks like we can improve on our end, so digging into that now.

emjburns: @luispollo would you mind creating a separate issue to look into why this is taking 30+ seconds? You can assign it to me - I'm happy to look into it. I also feel it shouldn't take that long, maybe I can make it better.

emjburns:
It looks like you're only doing this for ec2 - do you have any plans to implement this in a more generic way that would work for both ec2 and titus?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -382,10 +390,100 @@ class ClusterHandler(
...
+        orcaService
+          .getCorrelatedExecutions("${resource.id}:$region")
+          .isNotEmpty()
+      }
+
+  /**
+   * Retrieve server group entity tags from clouddriver to identity the "source" of the server group (pipeline or
+   * orchestration), then retrieve execution details from orca to determine the deployment strategy used for that
+   * server group.
+   */
```
emjburns, line 409: Do you think this function could be moved to a generic class of its own, and then used for both ec2 and titus? That would help keep this long bit of logic out of the ec2 specific cluster handler, and would make it easier to split up into smaller functions if that was helpful.

luispollo:
> It looks like you're only doing this for ec2 - do you have any plans to implement this in a more generic way that would work for both ec2 and titus?

You know, I completely forgot that Titus clusters had a deployment strategy too. 🤦‍♂️

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: Here it is: https://github.com/spinnaker/keel/issues/1060
I'm already on it: https://github.com/spinnaker/clouddriver/pull/4540

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: Yes, totally! I completely forgot about Titus. 🤦‍♂️

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/ClusterExportHelper.kt
```diff
@@ -0,0 +1,106 @@
...
+import com.netflix.spinnaker.keel.core.api.ClusterDeployStrategy
+import kotlinx.coroutines.async
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+
+/**
+ * Provides common logic for multiple cloud plugins to export aspects of compute clusters.
+ */
+@Component
+class ClusterExportHelper(
```
luispollo, line 13: I wasn't sure in which module to put this class since it has a dependency on `keel-orca` and `keel-clouddriver`. It turned out that `keel-orca` already had a dependency on `keel-clouddriver`, so I decided to put it here. 🤷‍♂️

emjburns: Sounds reasonable to me!

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/ClusterExportHelper.kt
```diff
@@ -0,0 +1,106 @@
...
+   * orchestration), then retrieve execution details from orca to determine the deployment strategy used for that
+   * server group.
+   */
+  suspend fun discoverDeploymentStrategy(
+    cloudProvider: String,
+    account: String,
+    application: String,
+    serverGroupName: String
+  ): ClusterDeployStrategy? {
+    return kotlinx.coroutines.coroutineScope {
```
emjburns, line 30: nit: does this import need to to be so fully qualified?

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/ClusterExportHelper.kt
```diff
@@ -0,0 +1,106 @@
...
+      val context = if (executionType == "pipeline") {
+        // TODO: or clone?
+        execution.getDeployStageContext()
+      } else { // orchestration (i.e. a task)
+        // TODO: or cloneServerGroup?
+        execution.getTaskContext("createServerGroup")
+      }
+
+      when (val strategy = context?.get("strategy")) {
+        "redblack" -> com.netflix.spinnaker.keel.core.api.RedBlack.fromOrcaStageContext(context)
```
emjburns, line 81: nit: these all don't need to be fully qualified.

my intellij started doing this too, but only for `gate`. I've yet to figure out why.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -382,10 +395,17 @@ class ClusterHandler(
     return spec
   }
 
-  private fun List<ResourceDiff<ServerGroup>>.createsNewServerGroups() =
-    any {
-      !it.isCapacityOrAutoScalingOnly()
-    }
+  override suspend fun actuationInProgress(resource: Resource<ClusterSpec>) =
```
luispollo, line 398: Just moved up above to come before the private extension functions.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
emjburns: Worth removing this comment if you've fixed the problem? or do we still need to see if this fixes the problem?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -192,12 +195,22 @@ class TitusClusterHandler(
...
 
+    val deployStrategy = clusterExportHelper.discoverDeploymentStrategy(
+      cloudProvider = "titus",
+      account = exportable.account,
+      application = exportable.moniker.app,
+      serverGroupName = base.name
+    )
+      ?.withDefaultsOmitted()
+      ?: RedBlack().withDefaultsOmitted()
+
```
emjburns, line 206: Nit: this reads kind of weird to me - i'd probably remove both the `.withDefaultsOmitted()` calls here (and in the ec2 cluster handler) and then just put it on the spec call below (so here line 213 would have `deployWith = deployStrategy.withDefaultsOmitted()`). I think that makes it a bit easier to read, because I don't have to reason about when `.withDefaultsOmitted()` would be null, then realize that's only the case when actually the `deployStrategy` was null.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/ClusterExportHelper.kt
```diff
@@ -0,0 +1,110 @@
...
+      if (spinnakerMetadata == null ||
+        spinnakerMetadata["executionType"] == null ||
+        spinnakerMetadata["executionId"] == null
+      ) {
+        log.warn("Unable to find Spinnaker metadata for server group $serverGroupName in application $application, " +
+          "account $account in entity tags.")
+        return@coroutineScope null
+      }
+
+      val executionType = spinnakerMetadata["executionType"]!!.toString()
```
emjburns, line 66: I think kotlin will help you out if you extract these to variables (like on line 56), then the nullcheck & return if null should let you not have to do the `!!` here.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/ClusterExportHelper.kt
```diff
@@ -0,0 +1,110 @@
...
+      val executionId = spinnakerMetadata["executionId"]!!.toString()
+      val execution = async {
+        when (executionType) {
+          "orchestration" -> orcaService.getOrchestrationExecution(executionId)
+          "pipeline" -> orcaService.getPipelineExecution(executionId)
+          else -> throw SystemException("Unsupported execution type $executionType in Spinnaker metadata.")
```
emjburns, line 72: Is this worth throwing an exception here? I'd rather just log an error and return null. It doesn't seem worth it to bork the whole export just because we didn't account for a case where we're trying to be extra helpful.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/ClusterExportHelper.kt
```diff
@@ -0,0 +1,110 @@
...
+        execution.getDeployStageContext()
+      } else { // orchestration (i.e. a task)
+        // TODO: or cloneServerGroup?
+        execution.getTaskContext("createServerGroup")
+      }
+
+      when (val strategy = context?.get("strategy")) {
+        "redblack" -> RedBlack.fromOrcaStageContext(context)
+        "highlander" -> Highlander
+        null -> throw InvalidRequestException("Deployment strategy information not found for server group $serverGroupName " +
```
emjburns, line 87: same comment re: exceptions here. Is it worth throwing them and borking the export when we're just trying to be extra helpful? I think a log message might be better

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/ClusterExportHelper.kt
luispollo: Yeah, this happened because I used the refactor tool to move it. Fixing.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: Still testing. Will remove the comment anyway as it's not actionable.

PR #1065, opened 2020-04-23
fix(titus): filtering out non matching tags

emjburns:
Right now, we sort tags based on the docker strategy you've chosen. If you choose the wrong strategy, we don't necessarily care, and even though we can't sort in a meaningful way we don't stop trying! We just put that un-sortable data at the bottom of the list. That seems not great.

This PR filters out non-matching tags based on the versioning strategy you've chosen. One caveat - this doesn't protect from user error. For example, you could tell us that your tag is and will always be a semantic version. Then, you could start pushing tags with colors (or other string garbage). Because you've told us that your whole tag is what we should use to sort, we say "sure! let's do it!". This PR doesn't change that (except that we explicitly ignore the "latest" tag).

If you've given us a custom regex or chosen a strategy with a custom regex (branch-job-commit-by-job, for example), and a tag that you push fails to have something extracted from it with regex (like "latest"), then we will cut that version out for rendering the summary and also approving the artifacts. This is similar functionality to the debian status filtering.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
@@ -176,7 +176,7 @@ class InMemoryArtifactRepository(
           // we only want to give valid versions, so this will filter out tags like "latest"
           // and others that don't fit the chosen versioning strategy.
           try {
-            TagComparator.parseWithRegex(it.version, artifact.tagVersionStrategy, artifact.captureGroupRegex) != null
+            it.version != "latest" && TagComparator.parseWithRegex(it.version, artifact.tagVersionStrategy, artifact.captureGroupRegex) != null
           } catch (e: InvalidRegexException) {
             false
           }
@@ -424,7 +424,7 @@ class InMemoryArtifactRepository(
                     artifact.statuses.isEmpty() || it.status in artifact.statuses
                   }
                   is DockerArtifact -> {
-                    TagComparator.parseWithRegex(it.version, artifact.tagVersionStrategy, artifact.captureGroupRegex) != null
+                    it.version != "latest" && TagComparator.parseWithRegex(it.version, artifact.tagVersionStrategy, artifact.captureGroupRegex) != null
                   }
                 }
               }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
@@ -5,6 +5,8 @@ import com.netflix.spinnaker.keel.api.artifacts.ArtifactStatus
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
 import com.netflix.spinnaker.keel.api.artifacts.DebianArtifact
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.artifacts.DockerArtifact
+import com.netflix.spinnaker.keel.core.TagComparator
 import com.netflix.spinnaker.keel.core.api.ArtifactSummaryInEnvironment
 import com.netflix.spinnaker.keel.core.api.ArtifactVersionStatus
 import com.netflix.spinnaker.keel.core.api.ArtifactVersions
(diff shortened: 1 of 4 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
```diff
@@ -169,6 +172,14 @@ class InMemoryArtifactRepository(
       .filter {
         if (artifact is DebianArtifact && artifact.statuses.isNotEmpty()) {
           it.status in artifact.statuses
+        } else if (artifact is DockerArtifact) {
+          // we only want to give valid versions, so this will filter out tags like "latest"
+          // and others that don't fit the chosen versioning strategy.
+          try {
+            it.version != "latest" && TagComparator.parseWithRegex(it.version, artifact.tagVersionStrategy, artifact.captureGroupRegex) != null
```
luispollo, line 179: Wouldn’t “latest” be filtered out naturally as a result of the regexp not matching? Why check for it explicitly?

emjburns: Ha! You would think! And I did think that! But, it turns out, no. With some tag strategies you say "hey I promise I'll give you a real tag that's a number" and that regex looks like `(^.*$)`. And that regex just means "grab everything". We could make the regex more complicated, to filter out other things. And a user could totally do that also. But for now, I thought I'd just filter out latest because I know we definitely don't want that ever.

gal-yardeni:
General question:
For better user experience - should we give users a set of valid values for sorting tags? doing so, we can validate it in the backend. It feels a bit scary to not let users know if they have entered some invalid string

emjburns:
@gal-yardeni what do you mean by 
> should we give users a set of valid values for sorting tags?

gal-yardeni:
> @gal-yardeni what do you mean by
> 
> > should we give users a set of valid values for sorting tags?

I mean - we have specific values that we support, correct? like `ranch-job-commit-by-job` and so on. Is there a list of all of those values? Are we supporting all kinds of custom values?

emjburns:
Ah, yes, we have some documentation about it ([here](https://www.spinnaker.io/guides/user/managed-delivery/artifacts/#docker-images)). And we could totally make the parsing of tags and validating more complicated - for example, in the `INCREASING_TAG` strategy we could validate that something is a number instead of just accepting all tag input. That would probably be a better experience. But I'd like to worry about improving that later I think, in a separate PR. 

I wonder if this is eventually something we could work into the validation endpoint - a little detection to see if you're set up correctly. Not sure.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -247,11 +249,34 @@ class SqlArtifactRepository(
...
+  /**
+   * Given a docker artifact and a list of docker tags, filters out all tags that don't produce exactly one capture
+   * group with the provided regex.
+   *
+   * This means that this will filter out tags like "latest" from the list.
+   */
+  private fun filterDockerVersions(artifact: DockerArtifact, versions: List<String>): List<String> =
+    versions.filter { version ->
+      try {
+        version != "latest" && TagComparator.parseWithRegex(version, artifact.tagVersionStrategy, artifact.captureGroupRegex) != null
```
lorin, line 273: I suggest pulling this line out into a function, since it's used twice (here and on line 762) and it's a complex enough concept that I think it's worth encapsulating into a function.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -247,11 +249,34 @@ class SqlArtifactRepository(
           .getValues(ARTIFACT_VERSIONS.VERSION)
       }
         .sortedWith(artifact.versioningStrategy.comparator)
+        .also { versions ->
+          return if (artifact is DockerArtifact) {
+            filterDockerVersions(artifact, versions)
+          } else {
+            versions
```
lorin, line 256: On line 247, there's some debian artifact filtering that happens in the query side so there's no debian filtering necessary here. However, on line 758, there is debian artifact filtering logic. Might want to modify this code so it filters debian artifacts using the same logic as on 758, and then you could pull that out into a function.

emjburns: Good spot. I think I'd rather move the filtering into the sql query that populates line 758 if possible, so we only fetch the data we need.. looking into it.

emjburns: Done. Didn't pull it out into a function, just copied the same sql filtering. That's... not excellent, but I think it's fine. I'm not quite sure how I would extract out that apply statement.

PR #1066, opened 2020-04-24
fix(security): Fix regression in adding X-SPINNAKER-ACCOUNTS

luispollo:
Fixes a regression introduced by #1009.


luispollo:
> LGTM. Could use some tests, though.

I know... Got lazy... Will add some tomorrow.

Issue #927, opened 2020-03-26
Proposal: replace KeelRepository with domain-specific service objects


luispollo:
### Problem statement
We recently introduced a new all-encompassing repository interface (`KeelRepository`) which is a proxy for all the various repository methods, and also adds some business logic around them. IMHO, that is not the best approach. Not only does it make that interface and the implementation extremely hard to read and understand, causing additional toil by forcing all repository functions to be defined in up to 5 different places (`MyRepository`, `MyInMemoryRepository`, `MySqlRepository`, `KeelRepository`, `CombinedRepository`), but it violates the single responsibility principle by adding business logic into what, by definition, should be a class concerned exclusively with persistence (otherwise, it shouldn't be named "Repository").

### Proposed solution
As attempted in the `ApplicationService` class, IMHO a better approach is to have classes whose specific and declared purpose is to be a service object responsible for that kind of business logic. Upstream callers, like the `ApplicationController` in this example, call into the high-level methods presented by this class to abstract details related to calling multiple persistence repositories and adding additional processing of the data.

I would propose that we reconsider the combined repository approach in favor of purpose-built service objects for specific sub-areas of the domain model. I'd be happy to open an issue to track that refactoring if all agree.

P.S. This is obviously not my original idea. In fact, we use this pattern in many other components of Spinnaker already. Another reason why I don't feel like we need to come up with a different design pattern of our own. 😄

### Early feedback
From @emjburns:
> But, the KeelRepository class is meant to deal with business logic - i.e. how do the objects relate, what order must things be done in, what happens when we delete one thing. Without a way to deal with all of the objects at once (delivery config, artifacts, resources) we will have consistency and concurrency bugs. We should talk about this offline.
>
> For now, please use the KeelRepository here.
> 
> You could in fact think of KeelRepository as a purpose built service objec> t that itself is concerned with persistence so that things like the ApplicationController don't have to be.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

luispollo:
I think we're in a much better place at this point in terms of abstracting things with service objects, although `KeelRepository` lives on. I suspect as we create more service objects, we may end up naturally moving functions out of the repository and into those objects, and it may just go away eventually.

PR #1069, opened 2020-04-25
fix(events): Fix paused events, again

### LINKED ISSUE #1043: Pausing, deleting a manifest, and then recreating gets keel into a wonky state
Very related to the UX discussion in #780

While one of our internal early adopters was trying things out, they paused their cluster resource to do some things, deleted their test manifest, and then resubmitted a modified manifest. The result of this was that their _new_ cluster resource definition was successfully created, and as a result the resource itself had a single `ResourceCreated` event in the history which led to it getting a `CREATED` status.

Thing is, there was still a row in the pause table for that cluster resource ID. Because of that, keel came around to check the cluster and skipped it because it was paused. This is the behavior I think we'd all want and expect (it seems scary to automatically start actuating something you had previously paused just because you resubmitted it), but because there was no `ResourcePaused` event in event history the resource **appeared** to be stuck in `CREATED`/not working at all, and all the UI / API was effectively making it look like things were totally broken. In order for our team to even understand what was happening we had to look at keel's own logs for that resource ID and discover a bunch of `Actuation for resource <resourceId> is paused, skipping checks` log lines.

Based on what I know about the history event system as it relates to pausing, I have a suspicion that something about the state where a resource is newly created just slips through our existing logic to add a 'synthetic' pause event to the history, which bubbles up to the status being wrong too. Worth also testing out whether application-level pausing has the same experience right now, but given how the UI handles those cases I'm skeptical it suffers from the same problem.

luispollo:
This PR fixes #1043, as follows:
- Revisits the logic in `ActuationPauser` that adds application pause/resume events to resource history by looking _before_ the oldest resource event in the list to see if the application or the resource had been paused and not yet resumed, in which case an `ApplicationActuationPaused` or `ResourceActuationPaused` event with a timestamp matching the time of the pause is added as the last of the events. 

  This is to account for cases where an application or resource is paused, the delivery config subsequently deleted, and then resubmitted before the application or resource is unpaused, in which case the resource status should remain paused.

- Moves business logic around resource status out of `ResourceRepository` and into a new `ResourceHistoryService` bean, which can interact both with repositories as well as other service beans like the `ActuationPauser`.

- Refactors (by moving them into test classes matching the responsible beans) and adds additional tests around resource events and statuses

I've also moved `ApplicationService` out of `keel-web` and into `keel-core`, alongside `ResourceService` for consistency, and simplified its tests so it doesn't depend on `keel-ec2-plugin` by adding easy-to-use new dummy resources with artifacts.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
@@ -48,13 +48,12 @@ import strikt.assertions.isTrue
  *
  * Tests that only apply to one repository should live in the repository-specific test classes.
  */
-abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository, P : PausedRepository> :
+abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository> :
   JUnit5Minutests {
 
   abstract fun createDeliveryConfigRepository(resourceSpecIdentifier: ResourceSpecIdentifier): D
   abstract fun createResourceRepository(resourceSpecIdentifier: ResourceSpecIdentifier): R
   abstract fun createArtifactRepository(): A
-  abstract fun createPausedRepository(): P
 
   open fun flush() {}
 
@@ -73,11 +72,10 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
     environments = setOf(firstEnv)
   )
 
-  data class Fixture<D : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository, P : PausedRepository>(
+  data class Fixture<D : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository>(
     val deliveryConfigRepositoryProvider: (ResourceSpecIdentifier) -> D,
     val resourceRepositoryProvider: (ResourceSpecIdentifier) -> R,
-    val artifactRepositoryProvider: () -> A,
-    val pausedRepositoryProvider: () -> P
+    val artifactRepositoryProvider: () -> A
   ) {
     internal val clock = MutableClock()
     val publisher: ApplicationEventPublisher = mockk(relaxUnitFun = true)
@@ -85,15 +83,13 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
     internal val deliveryConfigRepository: D = deliveryConfigRepositoryProvider(DummyResourceSpecIdentifier)
     internal val resourceRepository: R = resourceRepositoryProvider(DummyResourceSpecIdentifier)
     internal val artifactRepository: A = artifactRepositoryProvider()
-    internal val pausedRepository: P = pausedRepositoryProvider()
 
     val subject = CombinedRepository(
       deliveryConfigRepository,
       artifactRepository,
       resourceRepository,
       clock,
-      publisher,
-      ActuationPauser(resourceRepository, pausedRepository, publisher, clock)
+      publisher
     )
 
     fun resourcesDueForCheck() =
@@ -106,13 +102,12 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
         }
   }
 
-  fun tests() = rootContext<Fixture<D, R, A, P>> {
+  fun tests() = rootContext<Fixture<D, R, A>> {
     fixture {
       Fixture(
         deliveryConfigRepositoryProvider = this@CombinedRepositoryTests::createDeliveryConfigRepository,
         resourceRepositoryProvider = this@CombinedRepositoryTests::createResourceRepository,
-        artifactRepositoryProvider = this@CombinedRepositoryTests::createArtifactRepository,
-        pausedRepositoryProvider = this@CombinedRepositoryTests::createPausedRepository
+        artifactRepositoryProvider = this@CombinedRepositoryTests::createArtifactRepository
       )
     }
 
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
@@ -17,7 +17,6 @@ import com.netflix.spinnaker.keel.events.ResourceCreated
 import com.netflix.spinnaker.keel.events.ResourceUpdated
 import com.netflix.spinnaker.keel.exceptions.DuplicateArtifactReferenceException
 import com.netflix.spinnaker.keel.exceptions.DuplicateResourceIdException
-import com.netflix.spinnaker.keel.pause.ActuationPauser
 import com.netflix.spinnaker.keel.exceptions.MissingEnvironmentReferenceException
 import com.netflix.spinnaker.keel.resources.ResourceSpecIdentifier
 import com.netflix.spinnaker.keel.test.DummyResourceSpec
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
@@ -17,16 +17,17 @@ import com.netflix.spinnaker.keel.events.ResourceCreated
 import com.netflix.spinnaker.keel.events.ResourceUpdated
 import com.netflix.spinnaker.keel.exceptions.DuplicateArtifactReferenceException
 import com.netflix.spinnaker.keel.exceptions.DuplicateResourceIdException
+import com.netflix.spinnaker.keel.pause.ActuationPauser
 import com.netflix.spinnaker.keel.exceptions.MissingEnvironmentReferenceException
 import com.netflix.spinnaker.keel.resources.ResourceSpecIdentifier
 import com.netflix.spinnaker.keel.test.DummyResourceSpec
 import com.netflix.spinnaker.keel.test.TEST_API_V1
 import com.netflix.spinnaker.keel.test.resource
+import com.netflix.spinnaker.time.MutableClock
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
 import io.mockk.mockk
 import io.mockk.verify
-import java.time.Clock
 import java.time.Duration
 import org.springframework.context.ApplicationEventPublisher
 import strikt.api.expect
@@ -47,17 +48,17 @@ import strikt.assertions.isTrue
  *
  * Tests that only apply to one repository should live in the repository-specific test classes.
  */
-abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository> :
+abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository, P : PausedRepository> :
   JUnit5Minutests {
 
   abstract fun createDeliveryConfigRepository(resourceSpecIdentifier: ResourceSpecIdentifier): D
   abstract fun createResourceRepository(resourceSpecIdentifier: ResourceSpecIdentifier): R
   abstract fun createArtifactRepository(): A
+  abstract fun createPausedRepository(): P
 
   open fun flush() {}
 
   val configName = "my-config"
-
   val artifact = DockerArtifact(name = "org/image", deliveryConfigName = configName)
   val newArtifact = artifact.copy(reference = "myart")
   val firstResource = resource()
@@ -72,24 +73,27 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
     environments = setOf(firstEnv)
   )
 
-  data class Fixture<D : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository>(
+  data class Fixture<D : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository, P : PausedRepository>(
     val deliveryConfigRepositoryProvider: (ResourceSpecIdentifier) -> D,
     val resourceRepositoryProvider: (ResourceSpecIdentifier) -> R,
-    val artifactRepositoryProvider: () -> A
+    val artifactRepositoryProvider: () -> A,
+    val pausedRepositoryProvider: () -> P
   ) {
-    private val clock: Clock = Clock.systemUTC()
+    internal val clock = MutableClock()
     val publisher: ApplicationEventPublisher = mockk(relaxUnitFun = true)
 
     internal val deliveryConfigRepository: D = deliveryConfigRepositoryProvider(DummyResourceSpecIdentifier)
     internal val resourceRepository: R = resourceRepositoryProvider(DummyResourceSpecIdentifier)
     internal val artifactRepository: A = artifactRepositoryProvider()
+    internal val pausedRepository: P = pausedRepositoryProvider()
 
     val subject = CombinedRepository(
       deliveryConfigRepository,
       artifactRepository,
       resourceRepository,
       clock,
-      publisher
+      publisher,
+      ActuationPauser(resourceRepository, pausedRepository, publisher, clock)
     )
 
     fun resourcesDueForCheck() =
@@ -102,12 +106,13 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
         }
   }
 
-  fun tests() = rootContext<Fixture<D, R, A>> {
+  fun tests() = rootContext<Fixture<D, R, A, P>> {
     fixture {
       Fixture(
         deliveryConfigRepositoryProvider = this@CombinedRepositoryTests::createDeliveryConfigRepository,
         resourceRepositoryProvider = this@CombinedRepositoryTests::createResourceRepository,
-        artifactRepositoryProvider = this@CombinedRepositoryTests::createArtifactRepository
+        artifactRepositoryProvider = this@CombinedRepositoryTests::createArtifactRepository,
+        pausedRepositoryProvider = this@CombinedRepositoryTests::createPausedRepository
       )
     }
 
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
@@ -18,7 +18,6 @@ package com.netflix.spinnaker.keel.persistence
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Resource
 import com.netflix.spinnaker.keel.api.id
-import com.netflix.spinnaker.keel.core.api.ResourceSummary
 import com.netflix.spinnaker.keel.events.ResourceActuationLaunched
 import com.netflix.spinnaker.keel.events.ResourceCreated
 import com.netflix.spinnaker.keel.events.ResourceDeltaDetected
@@ -54,7 +53,6 @@ import strikt.assertions.isEqualTo
 import strikt.assertions.isFailure
 import strikt.assertions.isGreaterThanOrEqualTo
 import strikt.assertions.isNotEmpty
-import strikt.assertions.isNotNull
 import strikt.assertions.map
 
 abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests {
@@ -159,18 +157,6 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
 
         test("full resource returned for the application") {
           expectThat(subject.getResourcesByApplication("toast")).hasSize(1)
-        }
-
-        test("resource summary is formatted correctly") {
-          val summaries = subject.getResourceSummaries(deliveryConfig)
-
-          expectThat(summaries) {
-            hasSize(1)
-            with(first()) {
-              get(ResourceSummary::id).isEqualTo(lr.id)
-              get(ResourceSummary::locations).isNotNull()
-            }
-          }
         }
       }
 
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceHistoryListener.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceHistoryListener.kt
@@ -1,14 +1,30 @@
 package com.netflix.spinnaker.keel.events
 
+import com.netflix.spinnaker.keel.api.serviceAccount
+import com.netflix.spinnaker.keel.pause.ActuationPauser
 import com.netflix.spinnaker.keel.persistence.ResourceRepository
+import java.time.Clock
 import org.springframework.context.event.EventListener
 import org.springframework.stereotype.Component
 
 @Component
-class ResourceHistoryListener(private val resourceRepository: ResourceRepository) {
+class ResourceHistoryListener(
+  private val resourceRepository: ResourceRepository,
+  private val actuationPauser: ActuationPauser,
+  private val clock: Clock
+) {
 
   @EventListener(ResourceEvent::class)
   fun onResourceEvent(event: ResourceEvent) {
     resourceRepository.appendHistory(event)
+
+    // Account for the case where a resource was paused, then deleted (i.e. removed from management), then
+    // resubmitted, where we don't want to inadvertently resume actuation without the user knowing and giving
+    // explicit consent, by injecting a ResourceActuationPaused event.
+    // ApplicationActuationPaused events are injected in the resource event history dynamically.
+    if (event is ResourceCreated && actuationPauser.resourceIsPaused(event.id)) {
+      val resource = resourceRepository.get(event.id)
+      resourceRepository.appendHistory(ResourceActuationPaused(resource, resource.serviceAccount, clock))
+    }
   }
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceHistoryListener.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceHistoryListener.kt
@@ -1,30 +1,16 @@
 package com.netflix.spinnaker.keel.events
 
-import com.netflix.spinnaker.keel.api.serviceAccount
-import com.netflix.spinnaker.keel.pause.ActuationPauser
 import com.netflix.spinnaker.keel.persistence.ResourceRepository
-import java.time.Clock
 import org.springframework.context.event.EventListener
 import org.springframework.stereotype.Component
 
 @Component
 class ResourceHistoryListener(
-  private val resourceRepository: ResourceRepository,
-  private val actuationPauser: ActuationPauser,
-  private val clock: Clock
+  private val resourceRepository: ResourceRepository
 ) {
 
   @EventListener(ResourceEvent::class)
   fun onResourceEvent(event: ResourceEvent) {
     resourceRepository.appendHistory(event)
-
-    // Account for the case where a resource was paused, then deleted (i.e. removed from management), then
-    // resubmitted, where we don't want to inadvertently resume actuation without the user knowing and giving
-    // explicit consent, by injecting a ResourceActuationPaused event.
-    // ApplicationActuationPaused events are injected in the resource event history dynamically.
-    if (event is ResourceCreated && actuationPauser.resourceIsPaused(event.id)) {
-      val resource = resourceRepository.get(event.id)
-      resourceRepository.appendHistory(ResourceActuationPaused(resource, resource.serviceAccount, clock))
-    }
   }
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
@@ -25,11 +25,11 @@ import com.netflix.spinnaker.keel.events.ApplicationActuationResumed
 import com.netflix.spinnaker.keel.events.PersistentEvent
 import com.netflix.spinnaker.keel.events.ResourceActuationPaused
 import com.netflix.spinnaker.keel.events.ResourceActuationResumed
+import com.netflix.spinnaker.keel.events.ResourceCreated
 import com.netflix.spinnaker.keel.events.ResourceEvent
+import com.netflix.spinnaker.keel.pause.PauseScope.APPLICATION
+import com.netflix.spinnaker.keel.pause.PauseScope.RESOURCE
 import com.netflix.spinnaker.keel.persistence.PausedRepository
-import com.netflix.spinnaker.keel.persistence.PausedRepository.Scope
-import com.netflix.spinnaker.keel.persistence.PausedRepository.Scope.APPLICATION
-import com.netflix.spinnaker.keel.persistence.PausedRepository.Scope.RESOURCE
 import com.netflix.spinnaker.keel.persistence.ResourceRepository
 import java.time.Clock
 import org.slf4j.LoggerFactory
@@ -38,6 +38,7 @@ import org.springframework.stereotype.Component
 
 @Component
 class ActuationPauser(
+  // TODO: replace use of individual repos with KeelRepository
   val resourceRepository: ResourceRepository,
   val pausedRepository: PausedRepository,
   val publisher: ApplicationEventPublisher,
@@ -53,7 +54,7 @@ class ActuationPauser(
     return isPaused(resource)
   }
 
-  fun getPauseScope(resource: Resource<*>): Scope? =
+  fun getPauseScope(resource: Resource<*>): PauseScope? =
     when {
       applicationIsPaused(resource.application) -> APPLICATION
       resourceIsPaused(resource.id) -> RESOURCE
@@ -66,9 +67,9 @@ class ActuationPauser(
   fun resourceIsPaused(id: String): Boolean =
     pausedRepository.resourcePaused(id)
 
-  fun pauseApplication(application: String) {
+  fun pauseApplication(application: String, user: String) {
     log.info("Pausing application $application")
-    pausedRepository.pauseApplication(application)
+    pausedRepository.pauseApplication(application, user)
     publisher.publishEvent(ApplicationActuationPaused(application, clock))
   }
 
@@ -78,16 +79,15 @@ class ActuationPauser(
     publisher.publishEvent(ApplicationActuationResumed(application, clock))
   }
 
-  fun pauseResource(id: String) {
+  fun pauseResource(id: String, user: String) {
     log.info("Pausing resource $id")
-    pausedRepository.pauseResource(id)
+    pausedRepository.pauseResource(id, user)
     publisher.publishEvent(ResourceActuationPaused(resourceRepository.get(id), clock))
   }
 
   fun resumeResource(id: String) {
     log.info("Resuming resource $id")
     pausedRepository.resumeResource(id)
-    // helps a user not be confused by an out of date status from before a pause
     publisher.publishEvent(ResourceActuationResumed(resourceRepository.get(id), clock))
   }
 
@@ -102,10 +102,35 @@ class ActuationPauser(
     mutableListOf<PersistentEvent>()
       .let { events ->
         events.addAll(originalEvents)
-        val relevantAppEvents = resourceRepository
-          .applicationEventHistory(resource.application, events.last().timestamp)
+
+        val oldestResourceEvent = events.last()
+
+        if (oldestResourceEvent is ResourceCreated) {
+          val appPause = pausedRepository.getPause(APPLICATION, resource.application)
+          val resourcePause = pausedRepository.getPause(RESOURCE, resource.id)
+
+          // If the app was paused before the resource was created, and hasn't yet been resumed,
+          // we add that event to the history so the resource event history makes sense against its status
+          if (appPause != null && appPause.pausedAt.isBefore(oldestResourceEvent.timestamp)) {
+            events.add(ApplicationActuationPaused(resource.application, appPause.pausedAt))
+          }
+
+          // Similarly, if the resource itself had been paused, then was deleted and recreated,
+          // we add a corresponding event to the history
+          if (resourcePause != null && resourcePause.pausedAt.isBefore(oldestResourceEvent.timestamp)) {
+            events.add(ResourceActuationPaused(resource, resourcePause.pausedAt))
+          }
+        }
+
+        // Add all other app paused/resumed events from after the oldest resource event
+        resourceRepository
+          .applicationEventHistory(resource.application, oldestResourceEvent.timestamp)
           .filter { it is ApplicationActuationPaused || it is ApplicationActuationResumed }
-        events.addAll(relevantAppEvents)
+          .also { relevantAppEvents ->
+            events.addAll(relevantAppEvents)
+          }
+
+        // Always sort by timestamp to put events in the right positions
         events.sortedByDescending { it.timestamp }
       }
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
@@ -72,10 +72,10 @@ class ActuationPauser(
     publisher.publishEvent(ApplicationActuationPaused(application, user, clock))
   }
 
-  fun resumeApplication(application: String) {
+  fun resumeApplication(application: String, user: String) {
     log.info("Resuming application $application")
     pausedRepository.resumeApplication(application)
-    publisher.publishEvent(ApplicationActuationResumed(application, clock))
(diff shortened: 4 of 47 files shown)


luispollo:
⚠️ After sleeping on it, I realized my algorithm changes to inject application events won’t work. But I also had an idea that I think can help simplify this whole thing a lot. Will revisit on Monday, stay tuned!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/Pause.kt
```diff
@@ -0,0 +1,18 @@
+package com.netflix.spinnaker.keel.pause
+
+import java.time.Instant
+
+/**
+ * A pause in an application or resource's actuation, requested by a user.
+ */
+data class Pause(
```
luispollo, line 8: `PausedRepository` now has a corresponding model to work with, like other repositories.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -50,7 +52,8 @@ class CombinedRepository(
   val artifactRepository: ArtifactRepository,
   val resourceRepository: ResourceRepository,
   override val clock: Clock,
-  override val publisher: ApplicationEventPublisher
+  override val publisher: ApplicationEventPublisher,
+  override val actuationPauser: ActuationPauser
```
luispollo, line 56: This is an example of where I see `KeelRepository` and `CombinedRepository` leading us down a path that I don't think is the right one, by creating a "catch-all" bucket where to place service-level functions, which IMHO would be better placed in smaller beans that are focused on specific sub-areas of the domain model (like `ActuationPauser` in this case). Increasingly, I fear we'll start adding dependencies to other service beans here to achieve what we need, and I don't think a "repository" is the right place to be adding that kind of logic. 🤷‍♂️

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -313,7 +316,10 @@ class CombinedRepository(
     resourceRepository.getResourcesByApplication(application)
 
   override fun getResourceSummaries(deliveryConfig: DeliveryConfig): List<ResourceSummary> =
-    resourceRepository.getResourceSummaries(deliveryConfig)
+    getResourcesByApplication(deliveryConfig.application)
```
luispollo, line 319: There was no value in adding an additional layer of indirection to get resource summaries from a `ResourceRepository`, plus I needed to use `ActuationPauser` below to determine status (which goes into the summaries), so that logic resides here now.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
```diff
@@ -61,6 +93,13 @@ interface KeelRepository {
       log.debug("Creating $resource")
       storeResource(resource)
       publisher.publishEvent(ResourceCreated(resource, clock))
+
+      // Account for the case where a resource was paused, then deleted (i.e. removed from management), then
```
luispollo, line 97: Calling attention to this comment.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
```diff
@@ -231,4 +270,108 @@ interface KeelRepository {
...
+  fun getResourceEventHistory(id: String, limit: Int = ResourceRepository.DEFAULT_MAX_EVENTS): List<PersistentEvent> {
+    val resource = getResource(id)
+    return resourceEventHistory(id, limit)
+      .let { events ->
+        actuationPauser.addApplicationActuationEvents(events, resource)
+      }
+  }
+
+  fun getResourceStatus(id: String): ResourceStatus {
+    // For the PAUSED status, we look at the `paused` table as opposed to events, since these records
```
luispollo, line 283: Calling attention to this comment.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
```diff
@@ -231,4 +270,108 @@ interface KeelRepository {
   ): ArtifactSummaryInEnvironment?
 
   // END ArtifactRepository methods
+
+  fun getResourceEventHistory(id: String, limit: Int = ResourceRepository.DEFAULT_MAX_EVENTS): List<PersistentEvent> {
```
luispollo, line 274: Functions dealing with resource event history and status now live here, which is the closest place we have at the moment to a service object for these kinds of operations.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/PausedRepository.kt
```diff
@@ -17,24 +17,27 @@
...
 
+import com.netflix.spinnaker.keel.pause.Pause
+import com.netflix.spinnaker.keel.pause.PauseScope
+import java.time.Clock
+
 /**
  * A repository to track what scopes are paused, starting with application
  */
 interface PausedRepository {
+  val clock: Clock
```
luispollo, line 28: The clock is used to timestamp the paused records in the database.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/PausedRepository.kt
```diff
@@ -17,24 +17,27 @@
...
+    get() = Clock.systemUTC()
+
+  fun getPause(scope: PauseScope, name: String): Pause?
 
-  fun pauseApplication(application: String)
+  fun pauseApplication(application: String, user: String)
```
luispollo, line 33: Pause functions now take a user as an argument so we can record that in the database.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -66,9 +67,9 @@ class ActuationPauser(
   fun resourceIsPaused(id: String): Boolean =
     pausedRepository.resourcePaused(id)
 
-  fun pauseApplication(application: String) {
+  fun pauseApplication(application: String, user: String) {
```
luispollo, line 70: Functions that pause things now receive a user as an argument, so we can record who paused them in the database.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -102,10 +102,27 @@ class ActuationPauser(
...
+        val oldestResourceEvent = events.last()
+
+        if (oldestResourceEvent is ResourceCreated) {
+          val pause = pausedRepository.getPause(APPLICATION, resource.application)
+          if (pause != null && pause.pausedAt.isBefore(oldestResourceEvent.timestamp)) {
+            // If the app was paused before the resource was created, and hasn't yet been resumed,
+            // we add that event to the history so the resource event history makes sense against its status
+              events.add(ApplicationActuationPaused(resource.application, pause.pausedAt))
+          }
+        }
```
luispollo, line 115: This is much simplified from my previous attempt. Now that we have timestamps associated with records in the `paused` table, which are _not_ deleted upon deleting delivery configs or resources, we can synthesize paused events in a more straightforward fashion.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
luispollo: This is the only change in logic in this function and the ones just below. Otherwise, they were just moved over from `ResourceRepository`.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
```diff
@@ -1,16 +1,10 @@
-package com.netflix.spinnaker.keel
+package com.netflix.spinnaker.keel.services
```
luispollo, line 1: I removed all dependencies from `ApplicationServiceTests` with the `keel-ec2-plugin` module, which allowed me to move it, and the service class itself, to `keel-core`.

keel-sql/src/main/resources/db/changelog/20200427-add-info-to-paused-table.yml
```diff
@@ -0,0 +1,19 @@
+databaseChangeLog:
+  - changeSet:
+      id: add-timestamp-to-paused-table
+      author: lpollo
+      changes:
+        - addColumn:
+            tableName: paused
+            columns:
+              - column:
+                  name: paused_by
```
luispollo, line 10: Adds the two new columns that allow us to track who paused things and when.

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/CombinedRepository.kt
```diff
@@ -1,30 +1,62 @@
...
-  deliveryConfigRepository: DeliveryConfigRepository = InMemoryDeliveryConfigRepository(clock),
-  artifactRepository: ArtifactRepository = InMemoryArtifactRepository(clock),
-  resourceRepository: ResourceRepository = InMemoryResourceRepository(clock)
-): CombinedRepository =
-  CombinedRepository(deliveryConfigRepository, artifactRepository, resourceRepository, clock, mockk(relaxed = true))
+  deliveryConfigRepository: InMemoryDeliveryConfigRepository = InMemoryDeliveryConfigRepository(clock),
+  artifactRepository: InMemoryArtifactRepository = InMemoryArtifactRepository(clock),
+  resourceRepository: InMemoryResourceRepository = InMemoryResourceRepository(clock),
+  pausedRepository: InMemoryPausedRepository = InMemoryPausedRepository(clock)
+): InMemoryCombinedRepository =
```
luispollo, line 50: This avoids the need to be typecasting things in other places, since everything is explicitly declared as the in-memory type.

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/CombinedRepository.kt
```diff
@@ -1,30 +1,62 @@
...
 
 fun combinedMockRepository(
   deliveryConfigRepository: DeliveryConfigRepository = mockk(relaxed = true),
   artifactRepository: ArtifactRepository = mockk(relaxed = true),
   resourceRepository: ResourceRepository = mockk(relaxed = true),
-  clock: Clock = Clock.systemUTC()
+  pausedRepository: PausedRepository = mockk(relaxed = true),
+  clock: Clock = Clock.systemUTC(),
+  publisher: ApplicationEventPublisher = mockk(relaxed = true),
+  actuationPauser: ActuationPauser = ActuationPauser(resourceRepository, pausedRepository, publisher, clock)
```
luispollo, line 60: Another example of where putting service methods in `CombinedRepository` leads to weird patterns. Here, I'm adding a dependency on both `PausedRepository` and to `ActuationPauser` since the latter depends on the former, all so that I can add service methods that can query pause information. I had previously introduced a `ResourceService` that would be a new home for such service methods, but went back on it because I think we need to agree on a path forward as a team before I make further changes of that nature.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
```diff
@@ -34,8 +32,6 @@ class EventController(
     @RequestParam("limit") limit: Int?
   ): List<PersistentEvent> {
     log.debug("Getting state history for: $id")
-    val resource = repository.getResource(id)
-    val events = repository.resourceEventHistory(id, limit ?: DEFAULT_MAX_EVENTS)
-    return actuationPauser.addApplicationActuationEvents(events, resource)
+    return repository.getResourceEventHistory(id, limit ?: DEFAULT_MAX_EVENTS)
```
luispollo, line 35: Logic to manipulate the resource event history for pauses/resumes is now centralized in `KeelRepository`.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
```diff
@@ -66,19 +66,15 @@ class ResourceController(
     and @authorizationSupport.hasCloudAccountPermission('READ', 'RESOURCE', #id)"""
   )
   fun getStatus(@PathVariable("id") id: String): ResourceStatus =
-    if (actuationPauser.isPaused(id)) { // todo eb: we could make determining status easier and more straight forward.
-      PAUSED
-    } else {
-      repository.getResourceStatus(id)
-    }
+    repository.getResourceStatus(id)
```
luispollo, line 69: Logic to determine resource status is now centralized in `KeelRepository`.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -160,18 +158,6 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
         test("full resource returned for the application") {
           expectThat(subject.getResourcesByApplication("toast")).hasSize(1)
         }
-
-        test("resource summary is formatted correctly") {
-          val summaries = subject.getResourceSummaries(deliveryConfig)
```
luispollo, line 165: Covered by `ResourceServiceTests` now.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -69,24 +70,27 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
     environments = setOf(firstEnv)
   )
 
-  data class Fixture<D : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository>(
+  data class Fixture<D : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository, P : PausedRepository>(
```
luispollo, line 73: Tests now depend on `PausedRepository` to create an `ActuationPauser`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -161,99 +128,6 @@ interface ResourceRepository : PeriodicallyCheckedRepository<Resource<out Resour
    */
   override fun itemsDueForCheck(minTimeSinceLastCheck: Duration, limit: Int): Collection<Resource<out ResourceSpec>>
 
-  fun getStatus(id: String): ResourceStatus {
```
luispollo, line 164: Functions moved out of repository and into `ResourceService` bean as per discussions around rethinking responsibility of repositories, in particular not having any "business logic" in them and letting them focus on data-access-related stuff.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -119,19 +118,6 @@ class ApplicationService(
       targetEnvironment = targetEnvironment)
   }
 
-  /**
-   * Returns a list of [ResourceSummary] for the specified application.
-   *
-   * This function assumes there's a single delivery config associated with the application.
-   */
-  fun getResourceSummariesFor(application: String): List<ResourceSummary> =
```
luispollo, line 127: Also moved to `ResourceService`.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/veto/UnhappyVetoTests.kt
```diff
@@ -101,20 +98,6 @@ class UnhappyVetoTests : JUnit5Minutests {
             that(response3.allowed).isFalse()
           }
         }
-
-        test("a happy resource should no longer be skipped ") {
```
luispollo, line 105: This flow was no longer present in `UnhappyVeto`, in fact `ResouceRepository` was unused there, so I removed the test.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -119,24 +117,6 @@ open class SqlResourceRepository(
     }
   }
 
-  override fun getResourceSummaries(deliveryConfig: DeliveryConfig): List<ResourceSummary> {
```
luispollo, line 122: Moved to `ResourceService`.

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/Resources.kt
```diff
@@ -122,6 +123,36 @@ fun <T : ResourceSpec> submittedResource(
     spec = spec
   )
 
+fun versionedArtifactResource(
```
luispollo, line 126: New dummy resource types that can be used in tests in `keel-core` where we need artifact info.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/EventControllerTests.kt
```diff
@@ -183,100 +177,6 @@ internal class EventControllerTests : JUnit5Minutests {
       }
     }
 
-    context("with application paused at various times") {
```
luispollo, line 186: Tests moved to `ResourceServiceTests`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceService.kt
```diff
@@ -0,0 +1,192 @@
...
+        actuationPauser.addApplicationActuationEvents(events, resource)
+      }
+  }
+
+  /**
+   * Returns the status of the specified resource by first checking whether or not it or the parent application are
+   * paused, then looking into the last few events in the resource's history.
+   */
+  fun getStatus(id: String): ResourceStatus {
+    // For the PAUSED status, we look at the `paused` table as opposed to events, since these records
```
luispollo, line 73: Calling attention to this comment. This is the only change in logic in this function and the ones just below. Otherwise, they were just moved over from `ResourceRepository`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -102,10 +102,35 @@ class ActuationPauser(
     mutableListOf<PersistentEvent>()
       .let { events ->
         events.addAll(originalEvents)
-        val relevantAppEvents = resourceRepository
-          .applicationEventHistory(resource.application, events.last().timestamp)
+
+        val oldestResourceEvent = events.last()
+
+        if (oldestResourceEvent is ResourceCreated) {
```
luispollo, line 108: This is much simplified from my previous attempt. Now that we have timestamps associated with records in the `paused` table (which are not deleted upon deleting delivery configs or resources), we can synthesize paused events in a more straightforward fashion.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
```diff
@@ -34,8 +33,6 @@ class EventController(
     @RequestParam("limit") limit: Int?
   ): List<PersistentEvent> {
     log.debug("Getting state history for: $id")
-    val resource = repository.getResource(id)
-    val events = repository.resourceEventHistory(id, limit ?: DEFAULT_MAX_EVENTS)
-    return actuationPauser.addApplicationActuationEvents(events, resource)
+    return resourceService.getEnrichedEventHistory(id, limit ?: DEFAULT_MAX_EVENTS)
```
luispollo, line 36: Logic to manipulate the resource event history for pauses/resumes is now centralized in `ResourceService`.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ResourceController.kt
```diff
@@ -66,19 +68,15 @@ class ResourceController(
     and @authorizationSupport.hasCloudAccountPermission('READ', 'RESOURCE', #id)"""
   )
   fun getStatus(@PathVariable("id") id: String): ResourceStatus =
-    if (actuationPauser.isPaused(id)) { // todo eb: we could make determining status easier and more straight forward.
-      PAUSED
-    } else {
-      repository.getResourceStatus(id)
-    }
+    resourceService.getStatus(id)
```
luispollo, line 71: Logic to determine resource status is now centralized in `ResourceService`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
luispollo: P.S. If you feel this is not the right place to discuss this, please feel free to add your thoughts in https://github.com/spinnaker/keel/issues/927.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceService.kt
```diff
@@ -0,0 +1,192 @@
...
+import com.netflix.spinnaker.keel.persistence.ResourceStatus.PAUSED
+import com.netflix.spinnaker.keel.persistence.ResourceStatus.RESUMED
+import com.netflix.spinnaker.keel.persistence.ResourceStatus.UNHAPPY
+import com.netflix.spinnaker.keel.persistence.ResourceStatus.UNKNOWN
+import com.netflix.spinnaker.keel.persistence.ResourceStatus.VETOED
+import org.springframework.stereotype.Component
+
+/**
+ * Service object that offers high-level APIs for resource-related operations.
+ */
```
emjburns, line 49: It seems like this class deals only with the resource history and status. I think that it makes sense to have a class that deals with those things - they have definitely outgrown living in the `ResourceRepository`. I'd suggest naming this class to reflect that it does those things - I don't think it's a catch-all for resource operations. Maybe `ResourceHistoryService` or `ResourceEventsService`?

I like separating these thing out into their own class because in the future we could decide to change how we do calculate status, and it would be easy to replace that implementation detail in this class.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/PausedRepository.kt
```diff
@@ -17,24 +17,27 @@
...
   fun resumeApplication(application: String)
   fun applicationPaused(application: String): Boolean
 
-  fun pauseResource(id: String)
+  fun pauseResource(id: String, user: String)
   fun resumeResource(id: String)
```
emjburns, line 38: Should resume have the same capturing of user identity?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: Why? This is specifically in support of the application summary endpoint, I think it makes sense to stay here with all of the related functions

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceService.kt
```diff
@@ -0,0 +1,192 @@
...
+      // there are other events, we're not thrashing.
+      return false
+    }
+    return true
+  }
+
+  /**
+   * Returns a list of [ResourceSummary] for the specified application.
+   *
+   * This function assumes there's a single delivery config associated with the application.
```
emjburns, line 155: This comment is not needed, because that's an application-wide assumption

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceService.kt
```diff
@@ -0,0 +1,192 @@
...
+   */
```
emjburns, line 156: I don't agree that this goes here, but I have a question on the changed implementation. Before this had a sql implementation, now it does not. Is that because you have all the information needed in the delivery config, and we actually didn't need to look in the database?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -38,6 +38,7 @@ import org.springframework.stereotype.Component
 
 @Component
 class ActuationPauser(
+  // TODO: replace use of individual repos with KeelRepository
```
emjburns, line 41: If this only cares about resources, I don't see the need to replace it with the `KeelRepository`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
emjburns: Wouldn't it be easier to inject the `ApplicationPauser` into the place where we process these events? That way, we don't have to worry about it here, instead we can worry about it when we handle these events.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
```diff
@@ -39,7 +38,6 @@ import org.springframework.stereotype.Component
  */
 @Component
 class UnhappyVeto(
-  private val resourceRepository: ResourceRepository,
```
emjburns, line 42: was this just leftover and not used anymore?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/Pause.kt
```diff
@@ -0,0 +1,18 @@
...
+  val scope: PauseScope,
+  val name: String,
```
gal-yardeni, line 10: what does `name` means?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
luispollo: That's a good idea. I hand't thought about that. Will give it a try.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceService.kt
luispollo: Yeah, it's hard to find the right boundaries for these things... I'm not sure having extremely fine-grained service objects (classes with one or two methods) is the best approach either, because you'll end up having to inject a lot of them in other components higher in the stack, like controllers, but I definitely agree that we don't want to create "catch-all" components -- that was my whole rationale for opposing `KeelRepository`.

Let me ask you a question back. You mentioned you liked `ApplicationService`, which started small but now has several more methods. My thinking here was to follow that pattern. What do you think is different in this case that warrants a more narrowly-scoped class versus that case?

Also curious if @robfletcher has any thoughts on this.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/PausedRepository.kt
luispollo: Yes. I initially didn't think of adding it because there's no "resume" table to record it, but this will end up in the events. Was thinking it could be a separate PR since this one is pretty big already...

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceService.kt
luispollo: Where is the assumption written, though? 🤔

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceService.kt
luispollo: Correct. It was redundant with the `ResourceRepository.get()` method.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
luispollo: Correct.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/Pause.kt
luispollo: The "name" of the thing that was paused (resource or application). It's the name of the column in the database as well. It's more like an identifier, really, but I haven't changed that table, just added the model.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceService.kt
luispollo: Correct.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: Yeah, I didn't like this either, but the reason I moved it was that the `Resource<*>.toResourceSummary()` extension function has to call `getStatus(id)` for the resource, which, in turn, depends on other beans. Note that this function here was just delegating to `ResourceRepository` previously, where all those other functions used to live.

The alternative was to substitute `resourceService` for `repository` here, but that will introduce a dependency between `ApplicationService` and `ResourceService` which also felt a bit awkward, so I opted for moving it and just injecting both beans into `ApplicationController`.

Happy to undo this if you think the dependency between the two service beans is OK. Or let me know if you have any other ideas!

luispollo:
<video src="https://drive.google.com/file/d/1B9VgLAn0vzIjVvzHjdbssVhIRdWakpXq/preview" width="640" height="480" />

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceService.kt
emjburns: What do you mean? We have specific checks for it (like [here](https://github.com/spinnaker/keel/blob/master/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt#L101)), and a unique index constraint on the application field of the delivery config table. If you mean is there text anywhere that says that, I'm not sure, but regardless this comment is not the place to note an application wide assumption

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceService.kt
emjburns: I think the thing I like about `ApplicationService` is that it is build around a purpose (providing views for the application controller), vs just a collection of application-related methods. It has definitely grown to also include passthroughs for other methods, but the actual logic is all around rendering the views. I'd like to keep that same feeling here with the purpose of doing the event/history things. I don't like the grouping of "resource-related things" because it doesn't seem to have a focused purpose (and is more like the disliked `KeelRepository`).

I also like small and focused classes because they're easier to test and reason about. I don't mind injecting several to do a bigger thing. It's easy to refactor small classes.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: Hmm I see. Your original point about the `ApplicationService` was that the controller should be super light, and all the logic should be in the service. I agree with that, and I think based on that reasoning it would make sense to wire in the new history/event service into this class (because you're using that information to generate the application views).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -102,10 +102,35 @@ class ActuationPauser(
     mutableListOf<PersistentEvent>()
       .let { events ->
         events.addAll(originalEvents)
-        val relevantAppEvents = resourceRepository
-          .applicationEventHistory(resource.application, events.last().timestamp)
+
+        val oldestResourceEvent = events.last()
```
gal-yardeni, line 105: watching you explaining it in the video -- I think the name here is a bit confusing since it's not the `oldest` event, it's actually `latest` event, meaning the last event for that application, correct?

gal-yardeni:
@luispollo correct me if I'm wrong, but besides the mega refactoring that you did, the essence of this fix is to inject a synthetic event + timestamp if the application/resource was paused, and keep it paused in case the delivery config gets deleted and recreate?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
luispollo: Nope, it really is the oldest. The list is ordered in descending order of timestamp, so the first is the newest, and the last is the oldest.

gal-yardeni: isn't `events.last` returns the last publish event?

luispollo:
> @luispollo correct me if I'm wrong, but besides the mega refactoring that you did, the essence of this fix is to inject a synthetic event + timestamp if the application/resource was paused, and keep it paused in case the delivery config gets deleted and recreate?

Thanks for the "mega refactoring", haha! 😆 

Yes, I think you got the gist of it. We actually already had code to inject the pause/resume events in history, but we did not account for this case where the delivery config was deleted and then resubmitted, so the focus here is to cater to that case.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceService.kt
luispollo: Fair enough. Let's go with `ResourceHistoryService`, then!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
luispollo: No, that would be `events.first()` because the list is in descending order of timestamp, like I mentioned above.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/PausedRepository.kt
luispollo: I ended up adding this.

luispollo:
@gal-yardeni if it helps, here are the most important bits:

Adding paused events to history on-the-fly based on records in the `event` table:
https://github.com/spinnaker/keel/blob/d56abc35636790a6569e8c43fc0992081b36ccef/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt#L107-L122

Adding paused event to history when resource is re-created:
https://github.com/spinnaker/keel/blob/d56abc35636790a6569e8c43fc0992081b36ccef/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceHistoryListener.kt#L21-L28

Changing resource status logic to give higher precedence to evaluating paused status based on the `paused` table (not events):
https://github.com/spinnaker/keel/blob/d56abc35636790a6569e8c43fc0992081b36ccef/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceHistoryService.kt#L73-L79

Adding who paused an app/resource and when to the `paused` table:
https://github.com/spinnaker/keel/pull/1069/files#diff-18f7632f0ca5032d4877623591f2322c

Adding who triggered events:
https://github.com/spinnaker/keel/pull/1069/files#diff-2205a5946ca65d2b29a11a32433ed4a6

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -73,7 +75,7 @@ class ApplicationController(
       }
       entities.forEach { entity ->
         results[entity] = when (entity) {
-          "resources" -> applicationService.getResourceSummariesFor(application)
+          "resources" -> resourceHistoryService.getResourceSummariesFor(application)
```
emjburns, line 78: I still don't love this but it's fine

luispollo: I think I may have misread your previous comments on this. Will add the wrapper function back in `ApplicationService`.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -174,7 +174,7 @@ open class SqlResourceRepository(
         .from(EVENT)
         .where(EVENT.SCOPE.eq(Scope.APPLICATION.name))
         .and(EVENT.UID.eq(application))
-        .and(EVENT.TIMESTAMP.lessOrEqual(LocalDateTime.ofInstant(after, ZoneOffset.UTC)))
+        .and(EVENT.TIMESTAMP.greaterOrEqual(LocalDateTime.ofInstant(after, ZoneOffset.UTC)))
```
luispollo, line 177: This was another bug: the query was backwards! 😱

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -228,4 +229,14 @@ class ApplicationController(
   ) {
     applicationService.deleteVeto(application, targetEnvironment, reference, version)
   }
+
+  @GetMapping(
+    path = ["/{application}/events"],
```
luispollo, line 234: I realized we didn't have an endpoint for this and it helps at the very least with testing.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceHistoryListener.kt
```diff
@@ -1,30 +1,16 @@
...
     resourceRepository.appendHistory(event)
-
-    // Account for the case where a resource was paused, then deleted (i.e. removed from management), then
-    // resubmitted, where we don't want to inadvertently resume actuation without the user knowing and giving
-    // explicit consent, by injecting a ResourceActuationPaused event.
-    // ApplicationActuationPaused events are injected in the resource event history dynamically.
-    if (event is ResourceCreated && actuationPauser.resourceIsPaused(event.id)) {
-      val resource = resourceRepository.get(event.id)
-      resourceRepository.appendHistory(ResourceActuationPaused(resource, resource.serviceAccount, clock))
-    }
```
luispollo: This turned out to be redundant with the code I added to `ApplicationPauser`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
```diff
@@ -132,27 +132,25 @@ interface KeelRepository {
...
   fun getResourcesByApplication(application: String): List<Resource<*>>
 
-  fun getResourceSummaries(deliveryConfig: DeliveryConfig): List<ResourceSummary>
-
   fun storeResource(resource: Resource<*>)
 
   fun deleteResource(id: String)
 
-  fun applicationEventHistory(application: String, limit: Int): List<ApplicationEvent>
+  fun applicationEventHistory(application: String, limit: Int = DEFAULT_MAX_EVENTS): List<ApplicationEvent>
```
gal-yardeni, line 139: why the `DEFAULT_MAX_EVENTS` was added?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
gal-yardeni: hmm... didn't we send here the `until` time? how come it used to work?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
luispollo: Because you can’t query events from the database without a limit, and the controller receives it as a parameter, which needs to be passed down all the way to the repository.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: Yes, I just renamed `until` to `after` for clarity. But it turns out that the query never worked (there wasn’t a test for it, I should probably add one).

gal-yardeni: ha I just saw that the only place we use this function is in `ActuationPauser:addApplicationActuationEvents` and you changed the logic there, so I guess it used to work with that query? :)

luispollo: No... that's what I meant, I think we just got lucky that there weren't many apps paused, or no one was paying attention to those events being present in the event history. 🤷‍♂️

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
luispollo: Sorry, I misread your comment. We may not need the default here, actually. Let me check.

Issue #1063, opened 2020-04-23
Add admin endpoint for getting a list of managed apps


erikmunson:
As we move toward having many more apps managed by keel, it's becoming difficult to keep track of which apps are managed vs. not unless you have a _specific_ application you're interested in (at which point you can just hit one of the `/application/{application}` endpoints).

This is going to become increasingly painful as we try to do things like build dashboards or automation/scripts to do something with managed apps in bulk. What if we added an `/admin` endpoint (could even be internal-only, so no gate access) that returned a list of all the applications we know have delivery configs in the DB?

There's likely some performance implications we'd need to discuss and prepare for as the number of apps scales, but without this we'll likely either need to hammer the API with app names checking for configs or connect to the DB to make queries (which seems excessive for this use case).

PR #1070, opened 2020-04-27
feat(getAllApps): Endpoint to get all managed applications

gal-yardeni:
issue https://github.com/spinnaker/keel/issues/1063.

This is a proposal for an endpoint to retrieve all of the managed applications. 
I've created a new response type called `ApplicationSummary` , 
and the new endpoint will return a list of these. Here's an example:
`    "name": "gyardeni-manifest",
    "application": "gyardeniapp",
    "serviceAccount": "delivery-engineering@netflix.com",
    "apiVersion": "delivery.config.spinnaker.netflix.com/v1",
    "isPaused": true`

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ApplicationSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ApplicationSummary.kt
@@ -1,22 +1,14 @@
 package com.netflix.spinnaker.keel.core.api
 
-import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.annotation.JsonPropertyOrder
-import com.netflix.spinnaker.keel.api.DeliveryConfig
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder(value = ["name", "application", "serviceAccount", "apiVersion", "paused"])
+@JsonPropertyOrder(value = ["name", "application", "serviceAccount", "apiVersion", "isPaused"])
 data class ApplicationSummary(
-  @JsonIgnore val config: DeliveryConfig,
+  val name: String,
+  val application: String,
+  val serviceAccount: String,
+  val apiVersion: String,
   val isPaused: Boolean = false
-) {
-  val name: String
-    get() = config.name
-  val application: String
-    get() = config.application
-  val serviceAccount: String
-    get() = config.serviceAccount
-  val apiVersion: String
-    get() = config.apiVersion
-}
+)
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ApplicationSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ApplicationSummary.kt
@@ -4,9 +4,9 @@ import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.annotation.JsonPropertyOrder
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder(value = ["name", "application", "serviceAccount", "apiVersion", "isPaused"])
+@JsonPropertyOrder(value = ["deliveryConfigName", "application", "serviceAccount", "apiVersion", "isPaused"])
 data class ApplicationSummary(
-  val name: String,
+  val deliveryConfigName: String,
   val application: String,
   val serviceAccount: String,
   val apiVersion: String,
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ApplicationSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ApplicationSummary.kt
@@ -1,0 +1,22 @@
+package com.netflix.spinnaker.keel.core.api
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonPropertyOrder
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder(value = ["name", "application", "serviceAccount", "apiVersion", "paused"])
+data class ApplicationSummary(
+  @JsonIgnore val config: DeliveryConfig,
+  val isPaused: Boolean = false
+) {
+  val name: String
+    get() = config.name
+  val application: String
+    get() = config.application
+  val serviceAccount: String
+    get() = config.serviceAccount
+  val apiVersion: String
+    get() = config.apiVersion
+}
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
@@ -11,6 +11,7 @@ import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.artifacts.DockerArtifact
 import com.netflix.spinnaker.keel.api.id
 import com.netflix.spinnaker.keel.constraints.ConstraintState
+import com.netflix.spinnaker.keel.core.api.ApplicationSummary
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactPin
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVetoes
 import com.netflix.spinnaker.keel.core.api.EnvironmentSummary
@@ -295,8 +296,8 @@ class CombinedRepository(
   override fun deliveryConfigsDueForCheck(minTimeSinceLastCheck: Duration, limit: Int): Collection<DeliveryConfig> =
     deliveryConfigRepository.itemsDueForCheck(minTimeSinceLastCheck, limit)
 
-  override fun getAll(): Collection<DeliveryConfig> =
-    deliveryConfigRepository.getAll()
+  override fun getApplicationSummaries(): Collection<ApplicationSummary> =
+    deliveryConfigRepository.getApplicationSummaries()
   // END DeliveryConfigRepository methods
 
   // START ResourceRepository methods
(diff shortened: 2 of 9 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
```diff
@@ -36,4 +39,14 @@ class AdminController(
...
   fun getPausedApplications() =
     actuationPauser.pausedApplications()
+
+  @GetMapping(
+    path = ["/applications"],
+    produces = [MediaType.APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  fun getManagedApplications() =
+    repository.getAll().map {
+      ApplicationSummary(config = it,
```
emjburns, line 49: I think it's probably better to create a method on the pauser that returns all paused applications. Then, we can make only one query to the database to get all paused apps. Merging the data here is much cheaper than making a lot of sql calls.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ApplicationSummary.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonPropertyOrder
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder(value = ["name", "application", "serviceAccount", "apiVersion", "paused"])
+data class ApplicationSummary(
+  @JsonIgnore val config: DeliveryConfig,
```
robfletcher, line 11: I think it would make sense to specify all the properties (`name`, `application`, `serviceAccount`, `apiVersion`) in the data class constructor, then have a secondary constructor that extracts them from a delivery config.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -294,6 +294,9 @@ class CombinedRepository(
 
   override fun deliveryConfigsDueForCheck(minTimeSinceLastCheck: Duration, limit: Int): Collection<DeliveryConfig> =
     deliveryConfigRepository.itemsDueForCheck(minTimeSinceLastCheck, limit)
+
+  override fun getAll(): Collection<DeliveryConfig> =
+    deliveryConfigRepository.getAll()
```
robfletcher, line 299: Is it worth disambiguating the type in the method name in case we add a `getAll` for resources, or whatever?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
robfletcher: It might be worth optimizing the call path by making the repository methods return `Collection<ApplicationSummary>` rather than loading up a bunch of `DeliveryConfigs` and then just plucking a few fields from them. That way you could also include the paused status in a single query.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        context("with paused application") {
+          before {
+            actuationPauser.pauseApplication(application1)
+          }
+
+//          after {
+//            actuationPauser.resumeApplication(application)
+//          }
```
robfletcher, line 98: Is this supposed to be commented out?

gal-yardeni:
@emjburns @robfletcher - I've addressed your comments, appreciate another look.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ApplicationSummary.kt
```diff
@@ -0,0 +1,14 @@
+package com.netflix.spinnaker.keel.core.api
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonPropertyOrder
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder(value = ["name", "application", "serviceAccount", "apiVersion", "isPaused"])
+data class ApplicationSummary(
+  val name: String,
+  val application: String,
```
luispollo, line 10: What's the difference between `name` and `application`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryDeliveryConfigRepository.kt
```diff
@@ -221,6 +222,15 @@ class InMemoryDeliveryConfigRepository(
       .map { name -> configs[name] ?: error("No delivery config named $name") }
   }
 
+  // TODO[GY]: this does not take into account puased applications, as we can't wire it here.
```
luispollo, line 225: As discussed in chat, if #1069 gets merged, then you can take advantage of the changes in the following file and implement this directly in `InMemoryCombinedRepository`, where you'll have access to both delivery configs and pauses.
https://github.com/spinnaker/keel/pull/1069/files#diff-0501cf0c59a58dc72aee2333bb484a56

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ApplicationSummary.kt
robfletcher: That's the delivery config name, right? Maybe be just need to clarify that?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
```diff
@@ -0,0 +1,159 @@
...
+
+@ExtendWith(SpringExtension::class)
+@SpringBootTest(
+  classes = [KeelApplication::class, MockEurekaConfiguration::class],
+  webEnvironment = SpringBootTest.WebEnvironment.MOCK
+)
+@AutoConfigureMockMvc
+internal class AdminControllerTests : JUnit5Minutests {
+  @Configuration
+  class TestConfiguration {
```
luispollo, line 37: It doesn't look like you're referencing this configuration class anywhere?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
```diff
@@ -0,0 +1,159 @@
...
+    @Bean
+    fun clock(): Clock = MutableClock(
+      Instant.parse("2020-03-25T00:00:00.00Z"),
+      ZoneId.of("UTC")
+    )
+  }
+
+  @MockkBean
+  lateinit var authorizationSupport: AuthorizationSupport
```
luispollo, line 46: None of the `AdminController` APIs are authorized (which is a bit scary) so you don't need this...

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
```diff
@@ -0,0 +1,159 @@
...
+          serviceAccount = "keel@spinnaker",
+          artifacts = emptySet(),
+          environments = emptySet()
+        )
+        combinedRepository.upsertDeliveryConfig(deliveryConfig)
+        authorizationSupport.allowAll()
+      }
+
+      test("can get basic summary of unpaused application") {
+        val request = MockMvcRequestBuilders.get("/admin/applications/")
```
luispollo, line 85: nit: add an import statement for `MockMvcRequestBuilders.get` so you can just use `get()` here like in other tests.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
```diff
@@ -0,0 +1,159 @@
...
+          .accept(MediaType.APPLICATION_JSON_VALUE)
+        mvc
+          .perform(request)
+          .andExpect(MockMvcResultMatchers.status().isOk)
```
luispollo, line 89: Same comment on imports.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
```diff
@@ -0,0 +1,159 @@
...
+            application = application2,
+            serviceAccount = "keel@spinnaker",
+            artifacts = emptySet(),
+            environments = emptySet()
+          )
+          combinedRepository.upsertDeliveryConfig(deliveryConfig2)
+          authorizationSupport.allowAll()
+        }
+
+        test("can get basic summary of 2 applications, one paused") {
```
luispollo, line 116: You say "one paused" here but both have `"isPaused":false` below...

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ApplicationSummary.kt
gal-yardeni: changed `name` to `deliveryConfigName`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryDeliveryConfigRepository.kt
gal-yardeni: That would be great! I will make a note to myself to change it.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
gal-yardeni: should it be authorized?

gal-yardeni:
> Hi @gal-yardeni, this generally looks OK but I think there are a few things that need fixing as per my inline comments.
> 
> Also, on the issue of making the in-memory repo aware of pauses, an alternative approach you could take, since all the data you need is already exposed one way or another by existing repository functions, would be to move the `getApplicationSummaries()` function out of `KeelRepository` and into `ApplicationService`, where you could then call the repository to retrieve the delivery configs and the `ActuationPauser` to check which ones are paused, which would work regardless of the repository implementation you're using (in-memory or SQL).

@luispollo Thanks for reviewing and the suggestion here. I think I'd rather leave it as is and fix later on the in-memory implementation (as it only used for testing). I will make a note to fix the in-memory part + the tests after we'll have your pr in place.

PR #1071, opened 2020-04-27
feat(export): export artifacts from clusters

emjburns:
This PR adds an endpoint to export artifacts given a cluster. The existing cluster export will now return a reference provider with a reference to the name of the artifact instead of the old style full artifact (which is no longer supported)

These endpoints should be used together to produce the artifact and cluster that are both needed to put a cluster correctly in a delivery config.

A `GET /export/artifact/{provider}/{account}/{cluster}` will return something like:
```json
{
  "name": "deb-sample-app-server",
  "reference": "deb-sample-app-server",
  "vmOptions": {
    "baseLabel": "RELEASE",
    "baseOs": "xenial",
    "regions": [
      "us-east-1",
      "us-west-2"
    ],
    "storeType": "EBS"
  },
  "statuses": [],
  "type": "deb"
}
```
Notes:
- Name and reference will be the same, and reference can be excluded or changed.
- (`deb`) I'm guessing the baseOs from netflix naming conventions. Awesome Hackery. Turns out the image details from clouddriver do not tell us what OS an image is.
- (`deb`) No idea how to calculate the store type. This is defaulted.
- (`deb`) No idea how to calculate the statuses. This might need to be a user prompt.
- (`docker`) I'm trying to determine the type of tag versioning strategy to use based on the running tag. This won't work if the cluster is a one-off. I throw an exception in the case I can't determine the strategy, because I don't know how else to signal that the output wouldn't be correct.

### DIFF
--- a/keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroupTest.kt
+++ b/keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroupTest.kt
@@ -413,7 +413,10 @@ object ActiveServerGroupTest : ModelParsingTestSupport<CloudDriverService, Activ
     image = ActiveServerGroupImage(
       imageId = ami,
       appVersion = "$app-3.16.0-h205.121d4ac",
-      baseImageVersion = "nflx-base-5.308.0-h1044.b4b3f78"
+      baseImageVersion = "nflx-base-5.308.0-h1044.b4b3f78",
+      description = "name=$app, arch=x86_64, ancestor_name=xenialbase-x86_64-201811142132-ebs, ancestor_id=ami-0fe383ecea7f75eac, ancestor_version=nflx-base-5.308.0-h1044.b4b3f78",
+      imageLocation = "$owner/$app-3.16.0-h205.121d4ac-x86_64-20181115184054-xenial-hvm-sriov-ebs-ebs-ebs",
+      name = "$app-3.16.0-h205.121d4ac-x86_64-20181115184054-xenial-hvm-sriov-ebs-ebs-ebs"
     ),
     launchConfig = LaunchConfig(
       ramdiskId = "",
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -1,6 +1,7 @@
 package com.netflix.spinnaker.keel.ec2.resource
 
 import com.fasterxml.jackson.module.kotlin.convertValue
+import com.netflix.frigga.ami.AppVersion
 import com.netflix.spinnaker.keel.api.Exportable
 import com.netflix.spinnaker.keel.api.Moniker
 import com.netflix.spinnaker.keel.api.Resource
@@ -9,6 +10,9 @@ import com.netflix.spinnaker.keel.api.SubnetAwareLocations
 import com.netflix.spinnaker.keel.api.SubnetAwareRegionSpec
 import com.netflix.spinnaker.keel.api.actuation.Task
 import com.netflix.spinnaker.keel.api.actuation.TaskLauncher
+import com.netflix.spinnaker.keel.api.artifacts.DebianArtifact
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.artifacts.VirtualMachineOptions
 import com.netflix.spinnaker.keel.api.ec2.ClusterSpec
 import com.netflix.spinnaker.keel.api.ec2.ClusterSpec.LaunchConfigurationSpec
 import com.netflix.spinnaker.keel.api.ec2.ClusterSpec.ServerGroupSpec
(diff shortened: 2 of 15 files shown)


keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroupTest.kt
```diff
@@ -413,7 +413,10 @@ object ActiveServerGroupTest : ModelParsingTestSupport<CloudDriverService, Activ
     image = ActiveServerGroupImage(
       imageId = ami,
       appVersion = "$app-3.16.0-h205.121d4ac",
```
emjburns, line 415: Here's an example of what the data looks like for debs. To find the baseOS, I look for either "xenial" or "bionic" as strings in the `name`, `description`, and `imageLocation` fields. It's excellent garbage hackery. For netflix, these files are autogenerated by the bakery so it's reliable.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -491,7 +509,7 @@ class TitusClusterHandler(
       capacity = capacity,
       capacityGroup = capacityGroup,
       constraints = constraints,
-      container = generateContainer(container, location.account),
+      container = ReferenceProvider(container.repository()),
```
emjburns, line 512: Now when we export clusters we return them with a reference provider.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExportController.kt
```diff
@@ -45,7 +46,7 @@ class ExportController(
     "securitygroup" to "security-group",
     "securitygroups" to "security-group",
     "cluster" to "cluster",
```
emjburns, line 48: Pretty sure this was a typo.

robfletcher:
> These endpoints should be used together to produce the artifact and cluster that are both needed to put a cluster correctly in a delivery config.

This is something we expect Newt to do internally?

robfletcher:
> I'm guessing the baseOs from netflix naming conventions. Awesome Hackery. Turns out the image details from clouddriver do not tell us what OS an image is.

You should be able to reverse engineer it from the `base_ami_version` tag

robfletcher:
> No idea how to calculate the store type. This is defaulted.

I guess the presence of `ebs` in the image name indicates a store type of `EBS`. I'm not sure what other options would look like. Doesn't seem super-reliable though.

emjburns:
@robfletcher 
(1) yes, I'm building this endpoint so that corey can do this from newt
(2) the `base_ami_version` tag only tells what version of the netflix base package it is, not the os type.
(3) hmm interesting - how do you feel if I default to `ebs` now and we fix it when we find a good example of someone doing something else?

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroupTest.kt
emjburns: Yes we definitely will. I agree, not great. 

Do you have any ideas if the operating system is present in any field in any clouddriver response? I checked the server group and image response. I don't see the data surfaced anywhere in either of those or the spinnaker UI, or the internal edda link for the image. So, I'm thinking we just don't have the data. 

If we don't have the data, do you have any ideas on how to do this better?

luispollo:
Have you considered looking into pipeline bake stages instead of clouddriver? The info you’re missing is available there, but I wasn’t sure the use case is only for new migrations or if Cory wants export to work with already-managed clusters.

If you think it’s worth considering, there’s code you can reuse from that PoC I did:

https://github.com/luispollo/keel/blob/579bb14f46debb86333174808424271acb32e33b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ExportService.kt#L145-L169

Also, #1055 introduced some code to find pipelines based on entity tags from clouddriver. 

Just a thought.

luispollo:
Have you considered looking into pipeline bake stages instead of clouddriver? The info you’re missing is available there, but I wasn’t sure the use case is only for new migrations or if Cory wants export to work with already-managed clusters.

If you think it’s worth considering, there’s code you can reuse from that PoC I did:

https://github.com/luispollo/keel/blob/579bb14f46debb86333174808424271acb32e33b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ExportService.kt#L145-L169

Also, #1055 introduced some code to find pipelines based on entity tags from clouddriver. 

Just a thought.

emjburns:
@luispollo thanks for the suggestion. Right now I don't think that it is worth it to try and find the pipeline bake stage to discover OS - users typically understand what OS their package needs to run on, and I think we can ask them if we don't have the information. 

If it turns out that this is a major migration pain point I think we can add more logic to how we find these things.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -393,6 +400,52 @@ class ClusterHandler(
...
+      throw ExportError("Server group ${base.name} doesn't have image information - unable to correctly export artifact.")
+    }
+
+    val artifactName = AppVersion.parseName(base.launchConfiguration.appVersion).packageName
+
+    // todo: can we also discover status information?
+    return DebianArtifact(
+      name = artifactName,
+      vmOptions = VirtualMachineOptions(regions = serverGroups.keys, baseOs = guessBaseOsFrom(base.image)))
+  }
```
gal-yardeni, line 430: lol that method name :)

gal-yardeni:
@emjburns sorry for coming late to the party... was wondering few things:

1. what is the flow we expect newt to be doing? do we expect it to call this endpoint for each resource the user wants to export and then merge everything to one delivery config? 
2. what was the behavior prior to this change? did we infer the artifact info in a different way? you wrote it's not supported anymore, but I'm curious to learn what we used to do. We can talk about it offline if you prefer!

gal-yardeni:
answered in private, thank you!

PR #1076, opened 2020-04-28
CloudDriver doesn't consistently return region, removed it from model

robfletcher:


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
```diff
@@ -75,6 +75,13 @@ class RequiredLoadBalancerVeto(
     }
     return missing
   }
+
+  /**
+   * I'd like to add a `region` property to [AmazonLoadBalancer] but CloudDriver doesn't
+   * consistently return it from all the endpoints we use.
+   */
+  private val AmazonLoadBalancer.region: String
+    get() = availabilityZones.first().dropLast(1)
```
gal-yardeni, line 84: will this guaranty that we'll get the region correctly each time?

robfletcher: Given AWS's naming, at least in the regions we use, yeah

PR #1079, opened 2020-04-29
Disambiguate 404s

robfletcher:
We have an issue in that `CloudDriverCache` is frequently called as part of determining the current state of a resource. Usually this is for converting ids to names for things like subnets and security groups. If the cache is not primed, we make a call to CloudDriver. If that call throws a `RetrofitException` it will propagate out of the `CloudDriverCache` method and be handled in the resource handler. If the exception is a 404 we can misinterpret that as meaning the resource itself was not found. 

For example, look at https://github.com/spinnaker/keel/blob/master/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt#L261-L267 where if we catch an `HttpException` with a code of 404, we return `null` for the current state of the ALB resource. That's fine when the exception was thrown from the CloudDriver endpoint we are using to get the load balancer itself, but there are `CloudDriverCache` calls inside of that `try` block that may themselves throw `HttpException` because, for example a security group does not exist. Those 2 different 404s are _not_ the same thing. One means the resource doesn't exist, one is probably an integration exception talking to CloudDriver or some lag with CloudDriver's cache.

This PR disambiguates the 2 conditions by ensuring `HttpException` does not get thrown from `CloudDriverCache`.

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverCache.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverCache.kt
@@ -35,3 +35,4 @@ interface CloudDriverCache {
 }
 
 class ResourceNotFound(message: String) : IntegrationException(message)
+class CacheLoadingException(message: String, cause: Throwable) : IntegrationException(message, cause)
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
@@ -26,6 +26,7 @@ import java.util.concurrent.TimeUnit.HOURS
 import java.util.concurrent.TimeUnit.MINUTES
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.runBlocking
+import retrofit2.HttpException
 
 class MemoryCloudDriverCache(
   private val cloudDriver: CloudDriverService
@@ -138,6 +139,15 @@ class MemoryCloudDriverCache(
     notFoundMessage: String,
     loader: suspend CoroutineScope.() -> T?
   ): T = get(key) {
-    runBlocking(block = loader)
+    runCatching {
+      runBlocking(block = loader)
+    }
+      .getOrElse { ex ->
+        if (ex is HttpException && ex.code() == 404) {
+          null
+        } else {
+          throw CacheLoadingException("Error loading cache for $key", ex)
+        }
+      }
   } ?: throw ResourceNotFound(notFoundMessage)
 }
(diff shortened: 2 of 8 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverCache.kt
```diff
@@ -35,3 +35,4 @@ interface CloudDriverCache {
 }
 
 class ResourceNotFound(message: String) : IntegrationException(message)
+class CacheLoadingException(message: String, cause: Throwable) : IntegrationException(message, cause)
```
robfletcher, line 38: Wraps any kind of general exception we catch when priming the cache

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCacheTest.kt
```diff
@@ -92,9 +94,31 @@ object MemoryCloudDriverCacheTest {
...
+  @Test
+  fun `any other exception is propagated`() {
+    every {
+      cloudDriver.getSecurityGroupSummaries("prod", "aws", "us-east-1")
+    } throws RETROFIT_SERVICE_UNAVAILABLE
+
+    expectThrows<CacheLoadingException> {
+      subject.securityGroupById("prod", "us-east-1", "sg-4")
+    }
+  }
```
robfletcher, line 117: Tests how the cache handles exceptions

keel-retrofit-test-support/src/main/kotlin/com/netflix/spinnaker/keel/retrofit/errors.kt
```diff
@@ -1,4 +1,4 @@
-package com.netflix.spinnaker.keel.ec2
+package com.netflix.spinnaker.keel.retrofit
```
robfletcher, line 1: Moved this somewhere it can be more widely referenced

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
```diff
@@ -138,6 +139,15 @@ class MemoryCloudDriverCache(
...
+      runBlocking(block = loader)
+    }
+      .getOrElse { ex ->
+        if (ex is HttpException && ex.code() == 404) {
+          null
+        } else {
+          throw CacheLoadingException("Error loading cache for $key", ex)
+        }
+      }
   } ?: throw ResourceNotFound(notFoundMessage)
```
luispollo, line 152: Should we _ever_ throw `ResourceNotFound` exceptions from the cache, since none of the objects it handles are actually exposed as first-class objects in keel's APIs? Maybe we need a new type for this case too (`ReferencedResourceNotFound`?).

robfletcher: Yeah, that would make sense

gal-yardeni:
@robfletcher thanks for the detailed description!

PR #1081, opened 2020-04-29
fix(performance): Avoid retrieving all SGs when looking for one (reintroduced)

luispollo:
Reverts #1074.

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
@@ -45,13 +45,23 @@ interface CloudDriverService {
     @Query("vpcId") vpcId: String? = null
   ): SecurityGroupModel
 
-  @GET("/securityGroups/{account}/{provider}")
-  suspend fun getSecurityGroupSummaries(
+  @GET("/securityGroups/{account}/{provider}/{region}/{id}?getById=true")
+  suspend fun getSecurityGroupSummaryById(
     @Path("account") account: String,
     @Path("provider") provider: String,
-    @Query("region") region: String,
+    @Path("region") region: String,
+    @Path("id") id: String,
     @Header("X-SPINNAKER-USER") user: String = DEFAULT_SERVICE_ACCOUNT
-  ): Collection<SecurityGroupSummary>
+  ): SecurityGroupSummary?
+
+  @GET("/securityGroups/{account}/{provider}/{region}/{name}")
+  suspend fun getSecurityGroupSummaryByName(
+    @Path("account") account: String,
+    @Path("provider") provider: String,
+    @Path("region") region: String,
+    @Path("name") name: String,
+    @Header("X-SPINNAKER-USER") user: String = DEFAULT_SERVICE_ACCOUNT
+  ): SecurityGroupSummary?
 
   @GET("/networks")
   suspend fun listNetworks(
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
@@ -52,7 +52,7 @@ interface CloudDriverService {
     @Path("region") region: String,
     @Path("id") id: String,
     @Header("X-SPINNAKER-USER") user: String = DEFAULT_SERVICE_ACCOUNT
-  ): SecurityGroupSummary?
+  ): SecurityGroupSummary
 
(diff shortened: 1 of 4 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
```diff
@@ -45,13 +45,23 @@ interface CloudDriverService {
...
+
+  @GET("/securityGroups/{account}/{provider}/{region}/{name}")
+  suspend fun getSecurityGroupSummaryByName(
+    @Path("account") account: String,
+    @Path("provider") provider: String,
+    @Path("region") region: String,
+    @Path("name") name: String,
     @Header("X-SPINNAKER-USER") user: String = DEFAULT_SERVICE_ACCOUNT
-  ): Collection<SecurityGroupSummary>
+  ): SecurityGroupSummary?
```
robfletcher, line 64: I don't think this should be nullable as Retrofit with throw `HttpException` if CloudDriver responds with a 404.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
```diff
@@ -32,7 +32,7 @@ class MemoryCloudDriverCache(
   private val cloudDriver: CloudDriverService
 ) : CloudDriverCache {
 
-  private val securityGroupSummariesById = Caffeine.newBuilder()
+  private val securityGroupSummariesByIdOrName = Caffeine.newBuilder()
```
robfletcher, line 35: I'm confused why there's only one cache here but it looks like it's only primed with a single key. If I ask for a security group by id, and then the same security group by name, that's 2 hits on CloudDriver, right?

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
```diff
@@ -68,29 +68,25 @@ class MemoryCloudDriverCache(
...
 
   override fun securityGroupByName(account: String, region: String, name: String): SecurityGroupSummary =
-    securityGroupSummariesById.getOrNotFound(
+    securityGroupSummariesByIdOrName.getOrNotFound(
       "$account:$region:$name",
       "Security group with name $name not found in the $account account and $region region"
     ) {
       val credential = credentialBy(account)
 
       // TODO-AJ should be able to swap this out for a call to `/search`
```
robfletcher, line 88: Can probably lose this TODO comment now

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
luispollo: Good catch.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
luispollo: Another good catch. Will fix.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
luispollo: Right. I actually tried to look up this API but couldn't not connect the dots as `/search` doesn't seem to support returning security groups... Do you know anything more about it?

robfletcher: Nope

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCacheTest.kt
```diff
@@ -69,35 +69,24 @@ object MemoryCloudDriverCacheTest {
...
-        get { name }.isEqualTo("bar")
-        get { id }.isEqualTo("sg-2")
+        get { name }.isEqualTo("foo")
+        get { id }.isEqualTo("sg-1")
       }
     }
   }
 
-  @Test
-  fun `an invalid security group id throws an exception`() {
```
luispollo, line 87: Removed because it ended up being the same test as the 404 test below after refactoring for the new method names/behavior.

PR #1082, opened 2020-04-29
Use stubs rather than in-mem repositories in tests

robfletcher:


luispollo:
Hey @robfletcher, thanks for giving this a try! It looks like a sensible approach to me, and the work to set up the mocks/stubs is not more than setting up the data in the in-memory repos, so in terms of effort it seems fine. I do worry a bit about the cost of refactoring all those tests, though, but I haven't looked to see how many we have that would fall under this category.

Curious to see what you come up next with ideas for refactoring our data access layer!

robfletcher:
> Hey @robfletcher, thanks for giving this a try! It looks like a sensible approach to me, and the work to set up the mocks/stubs is not more than setting up the data in the in-memory repos, so in terms of effort it seems fine. I do worry a bit about the cost of refactoring all those tests, though, but I haven't looked to see how many we have that would fall under this category.

The good news is that there are not all that many

> Curious to see what you come up next with ideas for refactoring our data access layer!

Still thinking

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
```diff
@@ -246,6 +253,12 @@ internal class DeliveryConfigControllerTests : JUnit5Minutests {
       ).forEach { (contentType, payload) ->
         derivedContext<ResultActions>("persisting a delivery config as $contentType") {
           fixture {
+            every {
+              repository.upsertDeliveryConfig(any<SubmittedDeliveryConfig>())
+            } answers {
+              firstArg<SubmittedDeliveryConfig>().toDeliveryConfig()
+            }
```
robfletcher, line 260: Doing stub setup here is actually kind of a revelation to me. It seems obvious that you can, but I've always tried to do that in `before` which inevitably happens _after_ `fixture`, making it impossible to have a fixture like this which is the result of some action.

emjburns:
I think this approach seems more sustainable to me than maintaining an in-memory repository. From a general glance I don't see any huge problems

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/EventControllerTests.kt
```diff
@@ -186,18 +191,17 @@ internal class EventControllerTests : JUnit5Minutests {
...
+        every { repository.resourceEventHistory(resource.id, any()) } returns listOf(
+          ResourceCreated(resource, nextTick()),
+          ResourceValid(resource, nextTick())
+        )
+        every { actuationPauser.addApplicationActuationEvents(any(), any()) } answers {
+          (firstArg<List<ResourceEvent>>() + listOf(
+            ApplicationActuationPaused(resource.application, nextTick()),
+            ApplicationActuationResumed(resource.application, nextTick()),
+            ApplicationActuationPaused(resource.application, nextTick())
+          )).sortedByDescending { it.timestamp }
```
robfletcher, line 204: ¡I sure hope this is tested properly in the `ActuationPauser` itself!

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
```diff
@@ -127,60 +121,111 @@ class ApplicationServiceTests : JUnit5Minutests {
...
-      clock.reset()
+      every { repository.getDeliveryConfigForApplication(application) } returns deliveryConfig
     }
 
-    after {
-      clearAllMocks()
+    context("artifact summaries by application") {
+      before {
+        // repository.artifactVersions(artifact)
+        every { repository.artifactVersions(artifact) } returns listOf(version0, version1, version2, version3, version4)
```
luispollo, line 130:
```suggestion
        every { repository.artifactVersions(artifact) } returns versions
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
```diff
@@ -127,60 +121,111 @@ class ApplicationServiceTests : JUnit5Minutests {
...
+                hasSize(versions.size)
+                all {
+                  with(ArtifactVersionSummary::environments) {
+                    hasSize(environments.size)
+                    all {
+                      get { state }.isEqualTo(PENDING.name.toLowerCase())
+                    }
+                  }
+                }
+              }
```
luispollo, line 167: This looks pretty good, @robfletcher! Thanks for taking the time to lend your expertise to improving these tests. My only feedback is that I find the `with` blocks a bit hard to read above, and wondered if there was a way to rewrite that in a simpler way. For example, would it work to replace `with(ArtifactSummary::versions) {` with `get { versions }.and {`, or even `with(this.versions) {`, or is `with` a strikt function here?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
```diff
@@ -123,250 +116,401 @@ class ApplicationServiceTests : JUnit5Minutests {
...
+        test("non-evaluated stateful constraint details are included in the summary") {
+          val summaries = applicationService.getArtifactSummariesFor(application)
+
+          expectThat(summaries.first { it.name == artifact.name })
+            .withVersionInEnvironment(version1, "production") {
+              get { statefulConstraints }
+                .first { it.type == "manual-judgement" }
+                .get { status }.isEqualTo(NOT_EVALUATED)
+            }
         }
```
robfletcher, line 366: @luispollo I'd love for you to help me understand this behavior. If I have the mock return a constraint status for `manual-judgement` another gets added with this state. I don't quite grok what's going on here.

keel-orca/src/test/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintEvaluatorTests.kt
```diff
@@ -39,18 +46,12 @@ import strikt.assertions.isEqualTo
 import strikt.assertions.isFalse
 import strikt.assertions.isNotNull
 import strikt.assertions.isTrue
+import strikt.assertions.last
+import strikt.mockk.captured
+import strikt.mockk.isCaptured
 
 internal class CanaryConstraintEvaluatorTests : JUnit5Minutests {
```
emjburns, line 54: oof this looks like it was v complicated to refactor

robfletcher: That and `ApplicationServiceTests` were the hardest ones

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
```diff
@@ -123,250 +116,401 @@ class ApplicationServiceTests : JUnit5Minutests {
...
-      clock.reset()
-    }
-
-    after {
-      deliveryConfigRepository.dropAll()
-      artifactRepository.dropAll()
-      resourceRepository.dropAll()
-      clearAllMocks()
+      every { repository.getDeliveryConfigForApplication(application) } returns deliveryConfig
     }
```
emjburns, line 121: The only thing I'm slightly concerned with in this PR is the changes to this class. It took a lot of work to replicate real situations and test both the db and the transforming for this presentation layer. It's really hard to reproduce these situations in practice (locally to test, or in a test to reproduce behavior), and I think switching to mocking everything here does us a disservice. 

Do you think there's a way that we could switch these tests to use the sql repositories, or perhaps extract them to a new test focused specifically on testing both the extraction from the database and the rendering for the api layer? That doesn't feel super "nice" to me, but I'd really rather have this way to test complicated situations

emjburns:
Capturing a comment here that we discussed offline -

I'm a bit worried about the repository logic changing out from under us and us not catching a change in behavior due to mocking all the data. That being said, I think it's more burdensome to maintain separate in memory implementations of repositories for this purpose (because it's likely/has happened that they'll diverge in some small way). I'm trying to think of ways to mitigate this risk, but the only think I can think of is ensuring we have good repository unit tests and maybe a mix of test that are a level higher and test some interactions/use of the db with the actual implementation.

robfletcher:
> Capturing a comment here that we discussed offline -
> 
> I'm a bit worried about the repository logic changing out from under us and us not catching a change in behavior due to mocking all the data. That being said, I think it's more burdensome to maintain separate in memory implementations of repositories for this purpose (because it's likely/has happened that they'll diverge in some small way). I'm trying to think of ways to mitigate this risk, but the only think I can think of is ensuring we have good repository unit tests and maybe a mix of test that are a level higher and test some interactions/use of the db with the actual implementation.

Yes I agree, it _is_ a concern. However, I'm not sure it outweighs the effort involved in  maintaining in-memory repositories.

 I discovered when doing this was that in many cases:

1. using the in-memory repositories requires us to do more complex data setup
2. hides the fact that a lot of what is being tested is functionality encapsulated in the repository (implying it would be better tested at that level)

I'm hopeful that if writing tests forces us into painfully complex mock setup that's a signal we can respond to that our data access could be modeled better.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
robfletcher: It's certainly a possibility. One thing I'd like to retain would be the hierarchical context structure I set up here, which I think makes the test a lot easier to follow.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
luispollo: TIL!

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuatorTests.kt
```diff
@@ -93,18 +93,13 @@ internal class ResourceActuatorTests : JUnit5Minutests {
       every { plugin2.supportedKind } returns SupportedKind(parseKind("plugin2/bar@v1"), DummyArtifactVersionedResourceSpec::class.java)
     }
 
-    after {
-      resourceRepository.dropAll()
```
luispollo, line 97: Just to make sure I understand, the reason we don't need a `clearAllMocks()` to replace this here is that the mocks are initialized in the fixture and the fixture is newly instantiated in every test?

robfletcher: Yes. I'm always suspicious of any use of `clearAllMocks` as it suggests that something is leaking.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauserTests.kt
```diff
@@ -23,52 +23,46 @@ import com.netflix.spinnaker.keel.events.ApplicationActuationPaused
...
 
     context("application wide") {
       test("pause is reflected") {
         subject.pauseApplication(resource1.application, user)
-        expect {
-          that(subject.isPaused(resource1)).isTrue()
-          that(subject.isPaused(resource2)).isTrue()
-        }
+
+        verify { pausedRepository.pauseApplication(resource1.application, user) }
```
luispollo, line 65: This is just verifying the call immediately above and so feels like a no-op test to me. I'd expect to see a mock of the dependency called by the subject (`pausedRepository.applicationPaused(application)`) and the previous `expect` block to remain as it was.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
luispollo: I think it's scary because the tests here now look so different, but I really like the way that @robfletcher structured the new tests and think they're pretty readable once you stop comparing them to the old implementation and look at them on their own.

robfletcher:
Yeah, the size getting out of control was the reason I wanted to stop and merge even though there are still a few uses of the in-memory things remaining. I'll try to bear that in mind in future.

Issue #1090, opened 2020-05-01
Do resources get stuck in an unhappy state and stay vetoed?


emjburns:
I think that if a resources is marked unhappy (maybe for a transient reason), there are some situations where it will never be checked again. 

For example, we encounter a problem getting clusters from clouddriver, which leads us to try and create a new cluster twice. This doesn't work (because we still can't get resources from clouddriver), so we mark the resource as unhappy. Then, the issue is resolved. The resource desired state hasn't changed, so I think we never remove the unhappy mark.

PR #1092, opened 2020-05-01
feat(admin): recheck endpoint

emjburns:
Sometimes we might need to trigger a recheck of a resource. This might happen after recovering from an incident or bug (like https://github.com/spinnaker/keel/issues/1090). I created an admin endpoint to do this instead of just going into the database.

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
@@ -8,6 +8,7 @@ import org.springframework.http.MediaType
 import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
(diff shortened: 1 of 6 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
```diff
@@ -25,4 +27,9 @@ class AdminService(
   fun getManagedApplications(): Collection<ApplicationSummary> {
     return repository.getApplicationSummaries()
   }
+
+  fun triggerRecheck(resourceId: String) {
+    log.info("Triggering a recheck for $resourceId by clearing our record of the diff")
+    diffFingerprintRepository.clear(resourceId)
```
luispollo, line 33: Just so I understand, the reason we do this by clearing the diff fingerprint is that `UnhappyVeto` uses the diff count to determine whether to veto or allow a resource to be checked?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
```diff
@@ -41,4 +42,11 @@ class AdminController(
   )
   fun getManagedApplications() =
     adminService.getManagedApplications()
+
+  @PostMapping(
+    path = ["/recheck/{resourceId}"]
+  )
+  fun triggerRecheck(resourceId: String) {
+    adminService.triggerRecheck(resourceId)
```
gal-yardeni, line 50: Not a review breaker - but maybe it worth adding a test for the recheck?

emjburns: i have a test for the underlying function that it calls, which is where the logic is

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
emjburns: Yep! https://github.com/spinnaker/keel/blob/master/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt#L56

PR #1096, opened 2020-05-01
fix(retrofit): OkHttpClient should not be PROTOTYPE scope

robzienert:
Making a `OkHttpClient` bean `PROTOTYPE` scope is a big no-no:

> OkHttp performs best when you create a single `OkHttpClient` instance and reuse it for all of your HTTP calls. This is because each client holds its own connection pool and thread pools. Reusing connections and threads reduces latency and saves memory. Conversely, creating a client for each request wastes resources on idle pools.


luispollo:
@robzienert can you please update the branch so it will merge?

luispollo:
@robzienert Ping!

PR #1101, opened 2020-05-01
chore(config): delete config by app

emjburns:
Adding an endpoint to delete delivery configs by app, because it's annoying to have to first make a call to find the config name

I'm not removing the other endpoint for backwards compatibility.

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -55,6 +55,14 @@ class ApplicationService(
 
   fun getDeliveryConfig(application: String) = repository.getDeliveryConfigForApplication(application)
 
+  fun deleteConfigByApp(application: String) =
+    try {
+      val config = repository.getDeliveryConfigForApplication(application)
+      repository.deleteDeliveryConfig(config.name)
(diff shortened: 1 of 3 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -55,6 +55,14 @@ class ApplicationService(
 
   fun getDeliveryConfig(application: String) = repository.getDeliveryConfigForApplication(application)
 
+  fun deleteConfigByApp(application: String) =
+    try {
+      val config = repository.getDeliveryConfigForApplication(application)
```
luispollo, line 60: I feel like the repository should be able to delete the delivery config based on the application name to avoid the extra query here.

luispollo: P.S. Not a blocker, though.

emjburns: fair point. I should just implement that lol

emjburns: HA! It exists. I shall use it. Oops.

PR #1105, opened 2020-05-01
fix(api): Avoid deleting equivalent artifacts when upserting delivery…

### LINKED ISSUE #1091: Upserting delivery config deletes artifact
After inadvertently publishing an artifact for keel from a feature branch, it got picked up and deployed as we had no filters on the status of the artifact. To fix this, I changed the artifact in the delivery config from:
```yaml
artifacts:
  - name: keel
    type: deb
    reference: keel
    # ...
```
...to:
```yaml
artifacts:
  - name: keel
    type: deb
    reference: keel
    statuses:
      - RELEASE
    # ...
```
For some reason, this had the side-effect of **deleting the `delivery_artifact` record** from the database. 😨 It's unclear why. Following the code path for upserts, I failed to identify any branch that would lead to deleting that record.

This manifested itself as various exceptions in the logs complaining about the artifact not being found, such as this one:
```
com.netflix.spinnaker.keel.persistence.NoSuchArtifactException: No deb artifact named keel is registered
at com.netflix.spinnaker.keel.sql.SqlArtifactRepository.versions(SqlArtifactRepository.kt:251)
at com.netflix.spinnaker.keel.persistence.CombinedRepository.artifactVersions(CombinedRepository.kt:380)
at com.netflix.spinnaker.keel.persistence.CombinedRepository$FastClassBySpringCGLIB$9cfcaa57.invoke(<generated>:-1)
at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:685)
at com.netflix.spinnaker.keel.persistence.CombinedRepository$EnhancerBySpringCGLIB$ab7d5b5d.artifactVersions(<generated>:-1)
at com.netflix.spinnaker.keel.artifact.ArtifactListener.onArtifactRegisteredEvent(ArtifactListener.kt:74)
at sun.reflect.NativeMethodAccessorImpl.invoke0(NativeMethodAccessorImpl.java:-2)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:498)
at org.springframework.context.event.ApplicationListenerMethodAdapter.doInvoke(ApplicationListenerMethodAdapter.java:305)
at org.springframework.context.event.ApplicationListenerMethodAdapter.processEvent(ApplicationListenerMethodAdapter.java:190)
at org.springframework.context.event.ApplicationListenerMethodAdapter.onApplicationEvent(ApplicationListenerMethodAdapter.java:153)
at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener(SimpleApplicationEventMulticaster.java:172)
at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:165)
at org.springframework.context.event.SimpleApplicationEventMulticaster.lambda$multicastEvent$0(SimpleApplicationEventMulticaster.java:136)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
at java.lang.Thread.run(Thread.java:748)
```

luispollo:
Fixes #1091.

The problem was that we were relying on an `equals()` comparison to determine whether an artifact had been removed between the old version of the delivery config and the new, and `DeliveryArtifact` sub-classes are data classes, where the compiler generates an override for `equals()` that compares all properties, so if you changed a property that was not really part of the object identity (like `statuses`, in this case), the function would consider the artifacts different when, for practical purposes, they were not.

My proposed fix is to override `equals()` to look only at the artifact `name`, `type`, `reference` and `deliveryConfigName`. Other properties can still be compared individually between two artifacts if that use case is ever needed. The alternative would be to use that same comparison when upserting the delivery config and leave the `DeliveryArtifact.equals()` alone.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -31,6 +42,10 @@ data class DebianArtifact(
   override val versioningStrategy: VersioningStrategy = DebianSemVerVersioningStrategy
 ) : DeliveryArtifact() {
   override val type = ArtifactType.deb
+
+  override fun equals(other: Any?): Boolean {
```
luispollo, line 46: This is required to override the compiler-generated `equals()` override.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -19,7 +19,18 @@ sealed class DeliveryArtifact {
...
   abstract val deliveryConfigName: String? // the delivery config this artifact is a part of
+
   override fun toString() = "$type:$name" + if (reference != null) " (ref: $reference)" else ""
+
+  override fun equals(other: Any?): Boolean {
+    if (other is DeliveryArtifact) {
+      return other.name == this.name &&
+        other.type == this.type &&
+        other.reference == this.reference &&
+        other.deliveryConfigName == this.deliveryConfigName
```
luispollo, line 30: This now compares only the fields that actually identify the object to determine if they're "equal". Note that I'm not overriding `hashCode()` here, so objects with differences in other properties would still be considered different keys if you add them to a map, for example.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -19,7 +19,18 @@ sealed class DeliveryArtifact {
...
+    }
+    return super.equals(other)
```
lorin, line 32: Why do this rather than just return false in the case where the other object isn't a DeliveryArtifact?

luispollo: Because I didn't want to handle other cases like `DeliveryArtifact?` (nullable type). This was my way of saying "other than the special case above, use the default behavior".

luispollo:
> Also, it would be nice to see tests.

OK. I'll add some.

luispollo:
On second thought, I'm a little uncomfortable changing the behavior of `equals()`. I'll push another commit changing the comparison in `CombinedRepository` instead.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
luispollo: @lorin I ended up changing tactics in this PR. Please take another look.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -146,10 +146,16 @@ class CombinedRepository(
...
    * delivery config and are not present in the [new] delivery config
    */
-  override fun removeResources(old: DeliveryConfig, new: DeliveryConfig) {
+  override fun removeDependents(old: DeliveryConfig, new: DeliveryConfig) {
     val newResources = new.resources.map { it.id }
+
     old.artifacts.forEach { artifact ->
-      if (artifact !in new.artifacts) {
+      val stillPresent = new.artifacts.any {
+        it.name == artifact.name &&
```
gal-yardeni, line 154: Should we maybe want to extract it into an `equals` method or something? I suspect we might want to use it in the future as well.

luispollo: I had initially overridden the `equals()` method but feared it may have unintended side-effects, and ended up adding the filter only in this spot. I propose we revisit this if and when other cases come up.

gal-yardeni: sounds good, thanks!

PR #1109, opened 2020-05-04
fix(veto): check again if previously was happy

emjburns:
Fixes https://github.com/spinnaker/keel/issues/1090

When I first created the diff fingerprinting, I wasn't sure if we'd need to clear it ever. Now I definitely feel we need to clear it. This pr is a very small fix plus a specific test reproducing the problem.

### DIFF
--- a/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
+++ b/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
@@ -17,7 +17,6 @@ import com.netflix.spinnaker.keel.persistence.memory.InMemoryDiffFingerprintRepo
 import com.netflix.spinnaker.keel.persistence.memory.InMemoryResourceRepository
 import com.netflix.spinnaker.keel.persistence.memory.InMemoryUnhappyVetoRepository
 import com.netflix.spinnaker.keel.test.DummyResourceSpec
-import com.netflix.spinnaker.keel.test.deliveryConfig
 import com.netflix.spinnaker.keel.test.resource
 import com.netflix.spinnaker.keel.veto.VetoEnforcer
 import com.netflix.spinnaker.keel.veto.unhappy.UnhappyVeto
@@ -53,8 +52,14 @@ class IntermittentFailureTests : JUnit5Minutests {
       coEvery { isPaused(any<String>()) } returns false
       coEvery { isPaused(any<Resource<*>>()) } returns false
     }
-    val plugin1 = mockk<ResourceHandler<DummyResourceSpec, DummyResourceSpec>>(relaxUnitFun = true)
-    val plugin2 = mockk<ResourceHandler<DummyResourceSpec, DummyResourceSpec>>(relaxUnitFun = true)
+    val plugin1 = mockk<ResourceHandler<DummyResourceSpec, DummyResourceSpec>>(relaxUnitFun = true) {
+      every { name } returns "plugin1"
+      every { supportedKind } returns SupportedKind(parseKind("plugin1/foo@v1"), DummyResourceSpec::class.java)
+    }
+    val plugin2 = mockk<ResourceHandler<DummyResourceSpec, DummyResourceSpec>>(relaxUnitFun = true) {
+      every { name } returns "plugin2"
+      every { supportedKind } returns SupportedKind(parseKind("plugin2/bar@v1"), DummyResourceSpec::class.java)
+    }
     val publisher = mockk<ApplicationEventPublisher>(relaxUnitFun = true)
     val clock = MutableClock()
     val vetoRepository = InMemoryUnhappyVetoRepository(clock)
@@ -97,13 +102,6 @@ class IntermittentFailureTests : JUnit5Minutests {
 
     fixture { Fixture() }
 
-    before {
-      every { plugin1.name } returns "plugin1"
-      every { plugin1.supportedKind } returns SupportedKind(parseKind("plugin1/foo@v1"), DummyResourceSpec::class.java)
-      every { plugin2.name } returns "plugin2"
-      every { plugin2.supportedKind } returns SupportedKind(parseKind("plugin2/bar@v1"), DummyResourceSpec::class.java)
-    }
-
     after {
       resourceRepository.dropAll()
     }
@@ -112,7 +110,6 @@ class IntermittentFailureTests : JUnit5Minutests {
       val resource: Resource<DummyResourceSpec> = resource(
         kind = parseKind("plugin1/foo@v1")
       )
-      val deliveryConfig = deliveryConfig(resource)
 
       before {
         resourceRepository.store(resource)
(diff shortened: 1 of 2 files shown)


keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
```diff
@@ -0,0 +1,170 @@
...
+import io.mockk.mockk
+import java.time.Clock
+import kotlinx.coroutines.runBlocking
+import org.springframework.context.ApplicationEventPublisher
+
+/**
+ * This class tests a specific scenario:
+ * (1) a resource has a diff which we cannot fix, so it is vetoed.
+ * (2) that diff goes away (it was an intermittent failure of another system, maybe) and we say it's happy
+ * (3) that same diff comes back, and we still try again because it went from happy to diff.
```
luispollo, line 40: Just so I understand, number (3) here is in opposition to not trying anymore because it stayed unhappy forever?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
```diff
@@ -0,0 +1,170 @@
...
+
+  fun tests() = rootContext<Fixture> {
+
+    fixture { Fixture() }
+
+    before {
+      every { plugin1.name } returns "plugin1"
+      every { plugin1.supportedKind } returns SupportedKind(parseKind("plugin1/foo@v1"), DummyResourceSpec::class.java)
+      every { plugin2.name } returns "plugin2"
+      every { plugin2.supportedKind } returns SupportedKind(parseKind("plugin2/bar@v1"), DummyResourceSpec::class.java)
```
luispollo, line 104: Since these apply to all tests, any reason not to have the `every` statements above where you declare the mocks, like you did for other mocks?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
```diff
@@ -0,0 +1,170 @@
...
+      )
+      val deliveryConfig = deliveryConfig(resource)
+
+      before {
+        resourceRepository.store(resource)
+        resourceRepository.appendHistory(ResourceCreated(resource))
+        resourceRepository.appendHistory(ResourceValid(resource))
+        coEvery { plugin1.actuationInProgress(resource) } returns false
+        // diff has happened once
+        diffFingerprintRepository.store(resource.id, diff)
```
luispollo, line 123: I feel like this is forcing part of the behavior you want to verify. Who calls this functions in a real flow? I think it would be best to call the higher-level component which triggers this call.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
```diff
@@ -0,0 +1,170 @@
...
+        runBlocking { subject.checkResource(resource) }
+        coVerify { plugin1.update(resource, any()) }
+      }
+
+      context("diff seen third time") {
+        before {
+          // diff has happened twice
+          diffFingerprintRepository.store(resource.id, diff)
+        }
+        test("marked unhappy, so no actuation") {
```
luispollo, line 140: According to the description here, the test should also check that the status of the resource is unhappy, right?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        context("then, resource is briefly in no diff state") {
+          before {
+            coEvery { plugin1.desired(resource) } returns desired
+            coEvery { plugin1.current(resource) } returns desired
+            runBlocking { subject.checkResource(resource) }
+          }
+
+          test("happy resource means nothing happens") {
```
luispollo, line 152: Again, it would be nice to check the resource status to match the test description.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
emjburns: yeah - that's correct.

luispollo: Also, judging from the code in `ResourceActuator`, it shouldn't matter if the diff here is the same as in (1) or a different one, right? If there's any change at all between desired and current, `diff.hasChanges()` will return true and the fingerprint will be stored.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
emjburns: yep changed

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
emjburns: changing the name of the tests so it doesn't have happy in the name

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
emjburns: It's just making it so I don't call `subject.checkResource(resource)` again - I don't think I need to test that again. Trying to only reproduce the behavior I'm testing, and not test a bunch of other things. I'd say this is part of the setup of the test

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
emjburns: yes, correct! however, the same diff is just a more mentally weird case that i'm testing out here

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
luispollo: I understand, but what happens if the behavior of `subject.checkResource(resource)` later changes and stops storing the diff for some reason (including a regression)? Then you won't catch those bugs because you're emulating the expected behavior here, not letting the actual implementation do the setup for you.

emjburns: but I'm not testing that `subject.checkResource(resource)` stores the diff

Issue #936, opened 2020-03-27
Delivery config validation allows bad depends-on environment references


luispollo:
Given a delivery config like this:
```yaml
name: myapp
application: myapp
serviceAccount: delivery-engineering@netflix.com
artifacts:
   # ...
environments:
  - name: envA
    constraints: []
    notifications: []
    resources:
      - kind: "titus/cluster@v1"
        metadata: {}
        spec:
          moniker:
            app: myapp
            stack: envA
  - name: envB
    constraints:
      - type: depends-on
        environment: badEnvReference
    notifications: []
    resources:
      - kind: "titus/cluster@v1"
        metadata: {}
        spec:
          moniker:
            app: myapp
            stack: envB
```

Keel is allowing the file to be submitted and then blowing up with this exception:
```
java.lang.IllegalArgumentException: No environment named test exists in the configuration mddemotitus
   at com.netflix.spinnaker.keel.constraints.DependsOnConstraintEvaluator.canPromote(DependsOnConstraintEvaluator.kt:31)
   at com.netflix.spinnaker.keel.actuation.EnvironmentPromotionChecker.checkConstraints(EnvironmentPromotionChecker.kt:247)
   at com.netflix.spinnaker.keel.actuation.EnvironmentPromotionChecker.checkStatelessConstraints(EnvironmentPromotionChecker.kt:198)
   at com.netflix.spinnaker.keel.actuation.EnvironmentPromotionChecker.checkEnvironments$suspendImpl(EnvironmentPromotionChecker.kt:101)
   at com.netflix.spinnaker.keel.actuation.EnvironmentPromotionChecker.checkEnvironments(EnvironmentPromotionChecker.kt)
   at com.netflix.spinnaker.keel.actuation.CheckScheduler$checkEnvironments$job$1$invokeSuspend$$inlined$forEach$lambda$1$1.invokeSuspend(CheckScheduler.kt:106)
   at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
   at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
   at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
   at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738)
[...]

PR #1110, opened 2020-05-04
fix(validation): Throw an exception when referencing a non existing env

gal-yardeni:
fixes https://github.com/spinnaker/keel/issues/936.

Adding a validation for the depends on constraint - if referencing to a non-existing environment, throw a bad request exception. 

I'm planning on extracting the delivery config validation into a new service but will do it in a separate PR.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
@@ -437,7 +437,10 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
           }.failed()
             .isA<MissingEnvironmentReferenceException>()
 
-          expectThat(subject.allResourceNames().size).isEqualTo(0)
+          expectCatching {
+            subject.getDeliveryConfig(configName)
+          }.failed()
+            .isA<NoSuchDeliveryConfigException>()
         }
       }
     }
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
@@ -426,7 +426,7 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
               SubmittedEnvironment(
               name = "test",
             resources = emptySet(),
-            constraints = setOf(DependsOnConstraint(environment = "notRealEnvironment"))
+            constraints = setOf(DependsOnConstraint(environment = "notARealEnvironment"))
           )
           )
         )
@@ -434,12 +434,12 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
         test("an error is thrown and config is not persisted") {
           expectCatching {
             subject.upsertDeliveryConfig(submittedConfig)
-          }.failed()
+          }.isFailure()
             .isA<MissingEnvironmentReferenceException>()
 
           expectCatching {
             subject.getDeliveryConfig(configName)
-          }.failed()
+          }.isFailure()
             .isA<NoSuchDeliveryConfigException>()
         }
       }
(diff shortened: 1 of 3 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -231,6 +233,21 @@ class CombinedRepository(
       log.error("Validation failed for ${config.name}, duplicate artifact references found: $duplicatesArtifactNameToRef")
       throw DuplicateArtifactReferenceException(duplicatesArtifactNameToRef)
     }
+
+    /**
+     * validating depends on environments
+     */
+    config.environments.forEach { environment ->
+      environment.constraints.forEach { constraint ->
+        when (constraint is DependsOnConstraint) {
```
emjburns, line 242: I'd just use an if statement here

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/MissingEnvironmentReferenceException.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.keel.exceptions
+
+class MissingEnvironmentReferenceException(
+  private val environmentName: String
+) : ValidationException(
+  "The depend-on environment named: $environmentName " +
+    "is missing in the config. Please ensure you reference to the right environment."
```
emjburns, line 7:
```suggestion
    "is missing in the config. Please ensure you reference the right environment."
```

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -408,6 +410,48 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
...
+        val submittedConfig = SubmittedDeliveryConfig(
+          name = configName,
+          application = "keel",
+          serviceAccount = "keel@spinnaker",
+          artifacts = setOf(DockerArtifact(name = "org/thing-1", deliveryConfigName = configName, reference = "thing")),
+          environments = setOf(
+            SubmittedEnvironment(
+              name = "test",
+              resources = setOf(
+                SubmittedResource(
```
emjburns, line 424: you can probably make these just empty sets (environments with no resources) - I think we allow that, right?

emjburns: (makes the test easier to understand and maintain

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -408,6 +410,48 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
...
+          )
+        )
+
+        test("an error is thrown and config is not persisted") {
+          expectCatching {
+            subject.upsertDeliveryConfig(submittedConfig)
+          }.failed()
+            .isA<MissingEnvironmentReferenceException>()
+
+          expectThat(subject.allResourceNames().size).isEqualTo(0)
```
emjburns, line 452: if you remove the resources like suggested above you can just check to make sure the delivery config is not persisted.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
gal-yardeni: good call, will change

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
gal-yardeni: done

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
gal-yardeni: I couldn't find the right statement for making sure the delivery config table is empty (I saw this check is used for the same thing, so I used it).. can you point me to the right statement?

emjburns: I would just do:
```
expectCatching {
            subject.getDeliveryConfig(name)
          }.failed()
            .isA<NoSuchDeliveryConfigException>()
```

gal-yardeni: oooo right! thank you!

PR #1095, opened 2020-05-01
fix(logs): Improve logging of error conditions

luispollo:
Closes #1089 and adds other small improvements to logs for error conditions. More details in inline comments.

PR #1112, opened 2020-05-04
fix(events): Fix deserialization of ResourceCheckError

### LINKED ISSUE #1107: Can't deserialize ResourceCheckError objects
In prestaging, when we attempt to deserialize a ResourceCheckError event from the events table, keel throws a long exception. The relevant parts of the error message are::

```
MissingKotlinParameterException: Instantiation of [simple type, class com.netflix.spinnaker.keel.events.ResourceCheckError] value failed for JSON property exception due to missing (therefore NULL) value for creator parameter exception which is a non-nullable type ...

(through reference chain: com.netflix.spinnaker.keel.events.ResourceCheckError["exception"])
```

I think what's happening is that the `ResourceCheckError` class has an `exception` field which is required but is `@JsonIgnore'd`, so there's no `exception` field in the JSON payload. When Jackson tries to instantiate the object, it cannot construct a proper `exception` argument to pass the constructor at runtime.

I think this was introduced by #1095.

luispollo:
Fixes #1107 by reverting related changes in `ResourceCheckError`.

Added serialization/deserialization tests for all sub-classes of `ResourceEvent` to prevent future regressions.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
@@ -386,12 +386,12 @@ enum class ResourceCheckErrorOrigin {
 
   companion object {
     val log: Logger by lazy { LoggerFactory.getLogger(ResourceCheckErrorOrigin.javaClass) }
-    fun fromException(exception: SpinnakerException) =
-      when (exception) {
-        is UserException -> USER
-        is SystemException -> SYSTEM
+    fun fromException(exceptionType: Class<out SpinnakerException>) =
+      when {
+        UserException::class.java.isAssignableFrom(exceptionType) -> USER
+        SystemException::class.java.isAssignableFrom(exceptionType) -> SYSTEM
         else -> UNKNOWN.also {
-          log.debug("All keel exceptions should inherit from UserException or SystemException (got ${exception.javaClass.name}). " +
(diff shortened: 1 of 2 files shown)


luispollo:
@lorin Added tests as discussed, if you want to take another look.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -386,12 +386,12 @@ enum class ResourceCheckErrorOrigin {
...
-      when (exception) {
-        is UserException -> USER
-        is SystemException -> SYSTEM
+    fun fromException(exceptionType: Class<out SpinnakerException>) =
+      when {
+        UserException::class.java.isAssignableFrom(exceptionType) -> USER
+        SystemException::class.java.isAssignableFrom(exceptionType) -> SYSTEM
         else -> UNKNOWN.also {
-          log.debug("All keel exceptions should inherit from UserException or SystemException (got ${exception.javaClass.name}). " +
+          log.trace("All keel exceptions should inherit from UserException or SystemException (got ${exceptionType.name}). " +
```
luispollo, line 394: This is just reverting to the implementation prior to #1095.

luispollo: P.S. So are other changes in this file.

PR #1114, opened 2020-05-05
refactor(tests): use functions from strikt-mockk

robfletcher:


luispollo:
Looks good to me, but there's a failing test that seems entirely unrelated. It worries me that it's a test around injecting paused/resumed events which I added, but I don't recall any changes to that lately other than in my pending PR... 🤷‍♂️

PR #1118, opened 2020-05-05
chore(events): add unresolvable state for metrics

lorin:
Previously, both `ResourceDeltaDetected` and `ResourceCheckUnresolvable` events were reported with identical tags in the `keel.resource.checked` metric, which meant we couldn't tell how many times we were getting transient resource check failures.

This change differentiates these two events by introducing a new resource state. I called it `unresolvable` to match the name `ResourceChecekUnresolvable`.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceState.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceState.kt
@@ -16,5 +16,5 @@
 package com.netflix.spinnaker.keel.events
 
 enum class ResourceState {
-  Ok, Diff, Missing, Error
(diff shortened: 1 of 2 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceState.kt
```diff
@@ -16,5 +16,5 @@
 package com.netflix.spinnaker.keel.events
 
 enum class ResourceState {
-  Ok, Diff, Missing, Error
+  Ok, Diff, Missing, Error, Unresolvable
```
luispollo, line 19: nit: this enum is pretty redundant with `ResourceStatus`.

PR #1119, opened 2020-05-06
chore(web): add controller invocation metrics

### LINKED ISSUE #868: Review instrumentation/metrics and add Atlas graph for API requests
Check that instrumentation is present/enabled for REST requests and add an Atlas graph for meaningful API requests.

lorin:
Enable Spring controller (server-side IPC) metrics, consistent with the other Spinnaker apps.

I implemented this using [fiat's config](https://github.com/spinnaker/fiat/blob/0d58386152ad78234b2554f5efdf12d26b77d57c/fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/FiatConfig.java#L47-L54) as a starting point.

Some of the calls that Fiat makes have been deprecated, so I'm using the non-deprecated equivalents  (I'm not using the deprecated `WebMvcConfigurerAdapter` or the deprecated constructor of `MetricsInterceptor`).

Fixes #868.

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
@@ -2,6 +2,7 @@ package com.netflix.spinnaker.config
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.dataformat.yaml.YAMLMapper
+import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.fiat.shared.EnableFiatAutoConfig
 import com.netflix.spinnaker.filters.AuthenticatedRequestFilter
 import com.netflix.spinnaker.keel.api.plugins.ResourceHandler
@@ -21,6 +22,7 @@ import com.netflix.spinnaker.keel.persistence.memory.InMemoryResourceRepository
 import com.netflix.spinnaker.keel.persistence.memory.InMemoryTaskTrackingRepository
 import com.netflix.spinnaker.keel.serialization.configuredObjectMapper
 import com.netflix.spinnaker.keel.serialization.configuredYamlMapper
+import com.netflix.spinnaker.kork.web.interceptors.MetricsInterceptor
 import de.huxhorn.sulky.ulid.ULID
 import java.time.Clock
 import org.springframework.beans.factory.annotation.Qualifier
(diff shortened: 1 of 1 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
```diff
@@ -33,10 +35,39 @@ import org.springframework.context.annotation.Primary
...
+
+  /**
+   * Enable controller metrics
+   */
+  override fun addInterceptors(interceptorRegistry: InterceptorRegistry) {
+
+    /**
+     * Path variables that will be added as tags to the metrics
+     *
+     * Lorin chose these particular variables arbitrary as potentially useful. Once we start consuming
```
lorin, line 57: I like to put my rationale in code comments and identify myself, especially when I don't have a particularly good reason for a decision.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
```diff
@@ -33,10 +35,39 @@ import org.springframework.context.annotation.Primary
...
 @Configuration
-class DefaultConfiguration {
+class DefaultConfiguration(
+  val spectatorRegistry: Registry
+) : WebMvcConfigurer {
+
+  /**
+   * Enable controller metrics
+   */
+  override fun addInterceptors(interceptorRegistry: InterceptorRegistry) {
```
luispollo, line 52: Does the parent's implementation add any interceptors (logging, auth, perhaps)? Wondering if we should call `super.addInterceptors(interceptorRegistry)` at the end of this method.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
luispollo, line 57:
```suggestion
     * Lorin chose these particular variables arbitrarily as potentially useful. Once we start consuming
```

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
```diff
@@ -33,10 +35,39 @@ import org.springframework.context.annotation.Primary
...
+     *
+     * Lorin chose these particular variables arbitrary as potentially useful. Once we start consuming
+     * these metrics, we should revisit whether to change these based on usefulness & cardinality
+     */
+    val pathVarsToTag = listOf("name", "application")
+
+    val queryParamsToTag = null
+
+    // exclude list copied from fiat: https://github.com/spinnaker/fiat/blob/0d58386152ad78234b2554f5efdf12d26b77d57c/fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/FiatConfig.java#L50
+    val exclude = listOf("BasicErrorController")
```
luispollo, line 65: Is that controller auto-magically added by `kork`, or an actual controller in `fiat`? If the latter, then we should probably remove this for `keel`.

lorin: I _think_ this is auto-magically added by the Spring Boot framework for serving error pages.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
cfieber: the parent is an interface with all the hook points having default noop implementations

PR #1124, opened 2020-05-06
feat(lb): don't keep trying to fix a load balancer

emjburns:
I feel like we shouldn't keep trying with load balancers if we can't fix the diff. Thoughts?


erikmunson:
:+1: is there a case to be made for making this the default behavior across all resource kinds so we don’t have to rely on resource authors adding it bespoke when they realize they want it? I have no clue what that would entail technically.

emjburns:
@erikmunson probably that would be a good thing. I'll look into it another day

erikmunson:
excellent, i don't think it's necessarily important to worry about until we start opening up resource authoring to engineers outside our core group.

PR #1128, opened 2020-05-06
chore(metrics): remove useless "name" path metrics variable

lorin:
Several keel REST controllers use "name" as a path variable, so I originally thought it would be useful to report those as tags on the controller invocations metric.

The problem is that "name" is already used as the metric key (`controller.invocations`), and so specifying it as a path variable tag has no effect. This commit removes it from the list of path variables to tag.

Longer term, if we want to capture this information as tags in the metrics, we should rename the variable in the controllers from "name" to something else so we can use it in a tag.


emjburns:
Oh, interesting. We could rename those easily to things like `applicationName` etc. Do you think it would be useful to capture those things with our metrics?

PR #1129, opened 2020-05-06
fix(delete): delete by app now deletes everything

emjburns:
I used the wrong endpoint for deleting things by app. The one I used didn't actually delete anything but the config. This one actually deletes everything.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
@@ -32,7 +32,7 @@ import java.time.Instant
 import org.slf4j.LoggerFactory
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Component
-import org.springframework.transaction.annotation.Propagation
+import org.springframework.transaction.annotation.Propagation.REQUIRED
 import org.springframework.transaction.annotation.Transactional
 
 /**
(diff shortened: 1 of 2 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -119,12 +119,24 @@ class CombinedRepository(
...
-    val deliveryConfig = deliveryConfigRepository.get(deliveryConfigName)
+  @Transactional(propagation = REQUIRED)
+  override fun deleteDeliveryConfigByApplication(application: String): DeliveryConfig =
+    deleteDeliveryConfigFull(deliveryConfigRepository.getByApplication(application))
 
+  /**
+   * Deletes a delivery config and everything in it.
+   */
+  @Transactional(propagation = REQUIRED)
+  internal fun deleteDeliveryConfigFull(deliveryConfig: DeliveryConfig): DeliveryConfig {
```
lorin, line 139: Not related to this PR, but I think the code below could be a lot shorter if we used proper foreign keys in the database. We could just do a cascading delete.

emjburns: ha, that's part of a longstanding argument about efficiency and database design between asher and fletch. we could revisit, but we should do that by properly talking to sql experts and understanding the tradeoffs about how that scales

PR #1137, opened 2020-05-07

spinnakerbot:


lorin:
One of the SQL-related test fails, I thought it might be a transient so I kicked it to see if it would pass if run a second time.

PR #1143, opened 2020-05-08
fix(unpinning): Fix unpinning API

luispollo:
This PR replaces the usage of a request body for the `DELETE /application/{app}/pin` API with request parameters (since a body is not supported in DELETEs by the version of the okhttp client in gate).

TO-DO: update `gate`.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
```diff
@@ -16,9 +16,10 @@ import java.time.Instant
   JsonSubTypes.Type(value = ApplicationActuationResumed::class, name = "ApplicationActuationResumed")
 )
 abstract class ApplicationEvent(
-  override val triggeredBy: String? = null
+  override val triggeredBy: String? = null,
+  override val message: String? = null
```
luispollo, line 20: Added `message` to the `PersistentEvent` contract, which, along with `triggeredBy`, create a nice easy contract for the UI around basic properties every event we'll return should always have. The defaults are `null` in abstract classes, but sub-classes are encouraged to implement a meaningful message.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
```diff
@@ -14,13 +14,15 @@ import java.time.Instant
 )
 @JsonSubTypes(
   JsonSubTypes.Type(value = ApplicationEvent::class, name = "application"),
-  JsonSubTypes.Type(value = ResourceEvent::class, name = "resource")
+  JsonSubTypes.Type(value = ResourceEvent::class, name = "resource"),
+  JsonSubTypes.Type(value = PinningEvent::class, name = "pinning")
```
luispollo, line 18: New scope for pinning events.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEventListener.kt
```diff
@@ -5,9 +5,9 @@ import org.springframework.context.event.EventListener
 import org.springframework.stereotype.Component
 
 @Component
-class ApplicationEventListener(private val resourceRepository: ResourceRepository) {
-    @EventListener(ApplicationEvent::class)
-    fun onResourceEvent(event: ApplicationEvent) {
+class PersistentEventListener(private val resourceRepository: ResourceRepository) {
```
luispollo, line 8: There was redundant code to persist different sub-classes of `PersistentEvent`. This simplifies that a bit.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -443,11 +443,12 @@ class CombinedRepository(
   override fun pinnedEnvironments(deliveryConfig: DeliveryConfig): List<PinnedEnvironment> =
     artifactRepository.getPinnedEnvironments(deliveryConfig)
 
-  override fun deletePin(deliveryConfig: DeliveryConfig, targetEnvironment: String) =
-    artifactRepository.deletePin(deliveryConfig, targetEnvironment)
-
-  override fun deletePin(deliveryConfig: DeliveryConfig, targetEnvironment: String, reference: String) =
-    artifactRepository.deletePin(deliveryConfig, targetEnvironment, reference)
+  override fun deletePin(deliveryConfig: DeliveryConfig, targetEnvironment: String, reference: String?) =
```
luispollo, line 446: There was no reason to have two different methods in the combined repo.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
```diff
@@ -112,12 +112,12 @@ class InMemoryResourceRepository(
...
-    appendHistory(resourceEvents, event)
-  }
-
-  override fun appendHistory(event: ApplicationEvent) {
-    appendHistory(applicationEvents, event)
+  override fun <T : PersistentEvent> appendHistory(event: T) {
+    when (event) {
+      is ResourceEvent -> appendHistory(resourceEvents, event)
+      is ApplicationEvent -> appendHistory(applicationEvents, event)
+      else -> throw IllegalArgumentException("Unsupported event type ${event.javaClass}. Test bug?")
```
luispollo, line 119: Purposefully not adding support for the new event scope here because @robfletcher will soon remove all in-memory repositories.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -85,16 +89,30 @@ class ApplicationService(
...
-    val config = repository.getDeliveryConfigForApplication(application)
-    repository.deletePin(config, pin.targetEnvironment, pin.reference)
-  }
-
-  fun deletePin(application: String, targetEnvironment: String) {
+  fun deletePin(application: String, targetEnvironment: String, reference: String?, user: String) {
     val config = repository.getDeliveryConfigForApplication(application)
-    repository.deletePin(config, targetEnvironment)
+    repository.deletePin(config, targetEnvironment, reference)
+    publisher.publishEvent(
```
luispollo, line 107: We could previously figure out when something was pinned by looking at the database, but we didn't have any historical record of unpinning. This event adds that.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -85,16 +89,30 @@ class ApplicationService(
   fun pin(application: String, pin: EnvironmentArtifactPin, user: String) {
     val config = repository.getDeliveryConfigForApplication(application)
     repository.pinEnvironment(config, pin.copy(pinnedBy = user))
+    publisher.publishEvent(
```
luispollo, line 92: You could argue this is redundant with the record in the `environment_artifact_pin` table, but this provides an easy way to follow the trail of breadcrumbs about pinning and other events from a single source (`event` table). It also has a counterpart for unpinning below.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -192,10 +178,13 @@ class ApplicationController(
   )
   @ResponseStatus(HttpStatus.ACCEPTED)
   fun deletePin(
+    @RequestHeader("X-SPINNAKER-USER") user: String,
     @PathVariable("application") application: String,
-    @PathVariable("targetEnvironment") targetEnvironment: String
+    @PathVariable("targetEnvironment") targetEnvironment: String,
+    @RequestParam reference: String? = null,
```
luispollo, line 184: There's now a single API to delete a pin, with `reference` being optional. If it's missing, all pins for that environment will be deleted, otherwise, only the one corresponding to the artifact reference.

luispollo:
P.S. To be honest, it was only after I'd added the new events that I realized that we obviously already had the information for the pinning case in the database. Doh. 🤦‍♂️ 

The unpinning event might still be useful, not sure. Let me know what you think. I'd be happy to remove them.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
erikmunson: what does scope do today in keel in terms of affecting behavior? is it purely informational or does it get used for something?

I ask because ever since the idea of a 'scope' was added I've sort of imagined it as a nice convenient way to serve the filtering/discovery experience we have in mind for the history UI. We think folks will want to jump into an activity feed that is 'scoped' (we've used this term in the design conversations as well) to a particular part of their delivery manifest's structure or domain. For example: show me everything that has happened to this resource (what we show today), show me everything from the `main` environment including the resources inside, or show me everything that's happened to my entire application including the environments/resources inside. While that would theoretically end up surfacing a ton of events at the higher level scopes, we expect to also add event type filtering to let you (or a UI deep link) filter down to only the event types that matter to you in that moment.

If we think the `scope` attribute is appropriate for powering that feature in the future, then it may be worth thinking about changing this event's scope to `environment` since pinning is an action done _to_ an environment _with_ a particular artifact version. If there's other behavioral/architectural needs tied to scope today, then it might be more useful to come up with a separate way of representing the UI/API's idea of 'scoping' later.

gal-yardeni:
General comment: I see two terms here referencing to the same thing - `unpinning` and `deletePin`. Maybe it's worth using a single term to reduce confusion?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
luispollo: Good question, @erikmunson. It is not purely informational in the backend. It is used to differentiate different "classes" of events while retaining the ability of using a single `event` table which has the entire history of events in keel that you could sort by various criteria.

Given that, I think it would support _some_ of the use cases you mentioned well, but it currently does not carry the semantics of events that affect an entity and all its "children" (or related/affected entities) as in some of the other examples you give. As I mentioned in our call earlier today, the "meat" of events is actually in the `json` column in the table, so I suspect that, between the `scope`, `timestamp` and `json` columns we should be able to construct queries to cater to all use cases for the UI, and given that MySQL has optimizations for querying into JSON columns, performance should be OK. But I'm sure we'll find out more when we get our hands dirty with it. 🙂

luispollo:
> General comment: I see two terms here referencing to the same thing - `unpinning` and `deletePin`. Maybe it's worth using a single term to reduce confusion?

Good point, Gal. I'll rename the methods for consistency.

luispollo:
> > General comment: I see two terms here referencing to the same thing - `unpinning` and `deletePin`. Maybe it's worth using a single term to reduce confusion?
> 
> Good point, Gal. I'll rename the methods for consistency.

Actually, the only place we use "unpin" is in the new event, `ArtifactUnpinnedEvent`, which I think is reasonable. `ArtifactPinDeletedEvent` seems unnecessarily long and complicated.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
luispollo: Upon discussion with @emjburns I'm removing this from this PR and saving it for a separate one. In any case, the notion of event scopes is already present/usable for resource and application events.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: Will open a separate PR for new events.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: Will open a separate PR for events.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -46,6 +47,7 @@ import org.springframework.stereotype.Component
 class ApplicationService(
   private val repository: KeelRepository,
   private val resourceHistoryService: ResourceHistoryService,
+  private val publisher: ApplicationEventPublisher,
```
emjburns, line 50: you can remove the publisher, since publishing the event is a todo

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
```diff
@@ -367,6 +376,49 @@ class ApplicationServiceTests : JUnit5Minutests {
           that(states.first().type).isEqualTo("manual-judgement")
         }
       }
+
```
emjburns, line 379: these all just passthrough to the repository events, with no logic in the `ApplicationService`. Could you remove these tests? They're already tested at the repository level.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
emjburns: Could you remove comment from here? It's not used (I know you'd like to add it back in later).
I generally don't love having comments as URL parameters. It feels weird to me, and they could potentially be long or weird

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
luispollo: I'll rewrite the tests to check that the repository is being called to persist the events.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
luispollo: But we do use it... It's the "reason" field in the UI, which I assumed we would add for unpinning too @erikmunson?
image

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
emjburns: Is that really necessary to check, though? It's the only thing those methods are doing. I have 100% confidence that the repository is being called.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
luispollo: P.S. This is where that new event would come in as unpinning currently just deletes a record from the DB and doesn't store any other information for us to query later.

emjburns: To be clear, I'm not suggesting that we don't have a comment. Just that you don't use it in this PR, so it probably doesn't belong here

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
luispollo: Trying to stick to a pattern. I see @robfletcher used this approach extensively in the big test refactor in #1082. You have 100% confidence at the time you wrote the code, but one of the points of tests is to catch regressions that may be introduced by others who come later without prior context, and change that behavior inadvertently.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
erikmunson: I don't really see any meaningful value in adding a comment to an unpin operation, especially if the only place it will show up is in a history event's details. Probably not worth worrying about supporting it, I doubt we will expose it in the UI.

Issue #1088, opened 2020-05-01
Old artifact version stays in "DEPLOYING" status


emjburns:
Sometimes an artifact version gets left with [status](https://github.com/spinnaker/keel/blob/master/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/PromotionStatus.kt#L17) `DEPLOYING`(`environment_artifact_versions` table). While a new artifact deploys there are two with status `DEPLOYING`, and the api isn't happy. We log [in this case](https://github.com/spinnaker/keel/blob/master/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt#L49), and cut that old artifact out of the view.

Why does an artifact get forgotten about? What causes it to get left in this state? That is the mystery we need to solve.

emjburns:
Best guess: if something goes wrong between when we try to kick off a deploy for a new version and when it gets deployed (like, it fails to be up, or we fail to deploy it, or the world fails to be cool) then nothing will ever go and clean up that record.

We don't have a lot of information on _why_ that would happen, so we can't tell the user definitively that this version was bad, or that it was fine and our fault. 

Two options for how to fix it, which both will happen in a new and improved `markAsDeployingTo` function ([here](https://github.com/spinnaker/keel/blob/master/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt#L404-L423)). We shall (in a transaction) check to see if another version is deploying. If so, we know this is a weird state. We can either:
(1) create a new PromotionStatus, like `FUNKY`, which indicates that something happened. This will be helpful to categorize all these issues and make it easy to look into them. We mark the old one as funky, then mark the new one as deploying.
(2) we mark the old one as skipped to indicate that it just was never actually successfully deployed but it might not be bad
(3) _rejected option_: we mark the other one as `VETOED`. Rejected because it might not actually be bad, doesn't feel like we know what's going on enough to make a definitive ruling on that artifact.

Questions: 
- What is nicer for the user? Skipped? A new status?
- How permanent is this / these kinds of issues? Do they need their own status so that we can easily look at them?

emjburns:
cc @lorin @gal-yardeni @erikmunson

emjburns:
[...]

PR #1144, opened 2020-05-08
fix(artifacts): clean up old deploying states

emjburns:
Fixes https://github.com/spinnaker/keel/issues/1088

The rough logic is that if a version is stuck in `deploying` in an environment, and it's no longer the latest, you probably don't care that it never deployed. So, we shall skip it but also log an error so that radar will alert us and we can look into it.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryArtifactRepository.kt
```diff
@@ -391,6 +391,12 @@ class InMemoryArtifactRepository(
     val artifactId = getId(artifact) ?: throw NoSuchArtifactException(artifact)
     val key = EnvironmentVersionsKey(artifactId, deliveryConfig, targetEnvironment)
     val statuses = statusByEnvironment.getOrPut(key, ::mutableMapOf)
+    // update all previous deploying things to skipped
+    val deployingVersions = statuses.filterValues { it == DEPLOYING }
+    deployingVersions.keys.forEach {
+      log.error("Stuck deploying version $it for artifact '${artifact.reference}' in delivery config ${deliveryConfig.name} found when deploying version $version")
```
luispollo, line 397: You mentioned Radar would alert us of this error, but I don’t think it will because it monitors exceptions, not any error logs.

emjburns: I'll figure out how to get something to alert us

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -409,16 +409,42 @@ class SqlArtifactRepository(
...
+          }
+
+        txn
+          .insertInto(ENVIRONMENT_ARTIFACT_VERSIONS)
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID, environmentUid)
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID, artifact.uid)
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION, version)
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS, DEPLOYING.name)
+          .onDuplicateKeyUpdate()
+          .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS, DEPLOYING.name)
```
luispollo, line 445: I don't quite understand how `onDuplicateKeyUpdate()` works. How do you know which columns to update here?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -409,16 +409,42 @@ class SqlArtifactRepository(
...
+              log.error("Stuck deploying version $stuckVersion for artifact '${artifact.reference}' in delivery config ${deliveryConfig.name} found when deploying version $version")
+              txn.update(ENVIRONMENT_ARTIFACT_VERSIONS)
+                .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS, SKIPPED.name)
+                .set(ENVIRONMENT_ARTIFACT_VERSIONS.REPLACED_BY, version)
+                .set(ENVIRONMENT_ARTIFACT_VERSIONS.REPLACED_AT, now.toLocal())
+                .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(environmentUid))
+                .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifact.uid))
+                .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.eq(DEPLOYING.name))
+                .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.eq(stuckVersion))
+                .execute()
```
luispollo, line 434: Can this run into the "duplicate key" case like the insert below?

emjburns: No, the `update` clause specifically selects rows and updates the data in them.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: The `.onDuplicateKeyUpdate()` is part of the `.insertInto` statement. So, if a row already exists with a key that matches what that insert would generate, instead we update that record and set the specific fields we list

emjburns:
@luispollo No, I don't think there is any other clean up. If this is happening there will not be any other pending tasks

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: I updated this query to be a bit more efficient and easier to reason about after a consult from @robfletcher

PR #1147, opened 2020-05-08
fix(events): Update events that disallow duplicates instead of ignoring

luispollo:
This PR is an attempt to fix the issue we found today where resources got stuck in `RESUMED` status. 

The cause of the problem was that we look at the event history to determine status, and, in this case, there were `ApplicationActuationResumed` events appearing last in the event history for those resources. (We also noticed the application paused/resumed events were not displayed on the UI, making the experience even more confusing, but that’s a separate issue). 

Upon investigation, we discovered that keel behaved as expected after an application is resumed and proceeded to check the resources and emitted `ResourceValid` events. The problem was that these events have the `ignoreRepeatedInHistory` flag set to `true`, which causes them not to be persisted in the database if the last resource event is of the same type (which it was). In other words, we were ignoring those application-level events when checking for duplicates in the resource event history, and so a new `ResourceValid` event was never written to the DB.

Based on a suggestion from @emjburns, this PR changes the behavior around the `ignoreRepeatedInHistory` flag by updating the timestamp of a matching equivalent event in the last position in history, as opposed to skipping the event altogether. The side-effect this will have, which we thought was probably a good bonus, is that the timestamp of that last event in history will "catch up" every time we check, until a new event comes along, which seemed good because it will be apparent to the user that we're doing something.

@erikmunson Curious if you have any objections to the above.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
@@ -205,23 +205,24 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
                 .isA<ResourceValid>()
             }
 
-            context("a subsequent identical event that should be ignored") {
+            context("a subsequent identical event with duplicates disallowed") {
               before {
                 tick()
                 subject.appendHistory(ResourceValid(resource, clock))
               }
 
-              test("the event is not included in the history") {
+              test("the existing event's timestamp is updated") {
                 expectThat(subject.eventHistory(resource.id))
                   .and {
                     first().isA<ResourceValid>()
+                      .and { get { timestamp }.isEqualTo(clock.instant()) }
                     second().not().isA<ResourceValid>()
                   }
               }
             }
           }
 
-          context("a subsequent identical event") {
+          context("a subsequent identical event with duplicates allowed") {
             before {
               tick()
               subject.appendHistory(ResourceUpdated(resource, emptyMap(), clock))
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
@@ -120,13 +120,19 @@ class InMemoryResourceRepository(
     appendHistory(applicationEvents, event)
   }
 
-  private fun <T : PersistentEvent> appendHistory(eventList: MutableMap<String, MutableList<T>>, event: T) {
+  private fun <T : PersistentEvent> appendHistory(
+    eventList: MutableMap<String, MutableList<T>>,
+    event: T
+  ) {
     eventList.computeIfAbsent(event.uid) {
       mutableListOf()
     }
       .let {
         val mostRecentEvent = it.firstOrNull() // we get the first because the list is in descending order
-        if (!event.ignoreRepeatedInHistory || event.javaClass != mostRecentEvent?.javaClass) {
+        if (event.javaClass == mostRecentEvent?.javaClass && event.ignoreRepeatedInHistory) {
+          it.removeAt(0)
+          it.add(0, event)
+        } else {
           it.add(0, event)
         }
       }
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
@@ -259,6 +259,8 @@ open class SqlResourceRepository(
           txn
             .update(EVENT)
             .set(EVENT.TIMESTAMP, event.timestamp.atZone(clock.zone).toLocalDateTime())
+            // we udpate the JSON as well because it also contains a timestamp
+            .set(EVENT.JSON, objectMapper.writeValueAsString(event))
             .where(EVENT.SCOPE.eq(Scope.RESOURCE.name))
             .and(EVENT.UID.eq(resourceUid))
             .orderBy(EVENT.TIMESTAMP.desc())
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
@@ -239,13 +239,14 @@ open class SqlResourceRepository(
     jooq.transaction { config ->
       val txn = DSL.using(config)
 
+      val resourceUid = txn.select(RESOURCE.UID).from(RESOURCE).where(RESOURCE.ID.eq(event.id)).fetchOne(RESOURCE.UID)
+
       if (event.ignoreRepeatedInHistory) {
         val previousEvent = txn
           .select(EVENT.JSON)
-          .from(EVENT, RESOURCE)
+          .from(EVENT)
           .where(EVENT.SCOPE.eq(Scope.RESOURCE.name))
-          .and(RESOURCE.ID.eq(event.id))
-          .and(RESOURCE.UID.eq(EVENT.UID))
+          .and(EVENT.UID.eq(resourceUid))
           .orderBy(EVENT.TIMESTAMP.desc())
           .limit(1)
           .fetchOne()
@@ -253,7 +254,18 @@ open class SqlResourceRepository(
             objectMapper.readValue<ResourceEvent>(json)
           }
 
-        if (event.javaClass == previousEvent?.javaClass) return@transaction
+        // If an equivalent event is already the latest in history, just update the timestamp
+        if (event.javaClass == previousEvent?.javaClass) {
+          txn
+            .update(EVENT)
+            .set(EVENT.TIMESTAMP, event.timestamp.atZone(clock.zone).toLocalDateTime())
+            .where(EVENT.SCOPE.eq(Scope.RESOURCE.name))
+            .and(EVENT.UID.eq(resourceUid))
+            .orderBy(EVENT.TIMESTAMP.desc())
+            .limit(1)
+            .execute()
+          return@transaction
+        }
       }
 
       txn


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -242,15 +249,19 @@ open class SqlResourceRepository(
...
-          .and(RESOURCE.ID.eq(event.id))
-          .and(RESOURCE.UID.eq(EVENT.UID))
+          .from(EVENT)
+          .where(EVENT_JSON_APPLICATION.eq(event.application))
+          // look for resource events that match the resource...
+          .and(EVENT.SCOPE.eq(Scope.RESOURCE.name)
+            .and(EVENT_JSON_RESOURCE_ID.eq(event.id))
+          )
+          // ...or application events that match the application as they apply to all resources
+          .or(EVENT.SCOPE.eq(Scope.APPLICATION.name))
```
luispollo, line 259: I need to add a condition here for the application event being later than the last resource event".

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -236,6 +237,25 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
...
+              subject.appendHistory(ApplicationActuationResumed("fnord", "keel@keel.io"))
+              tick()
+              subject.appendHistory(ResourceValid(resource, clock))
+            }
+
+            test("overrides the duplicate check and persists the event") {
+              expectThat(subject.eventHistory(resource.id))
+                .and {
+                  first().isA<ResourceValid>()
+                  second().isA<ResourceValid>()
```
luispollo, line 255: This is failing here because the second event is never inserted into the DB, judging from the jooq debug logs, and yet when I run it in the debugger I can follow it to the insert path... 🤔

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: Actually, this should work because of the descending sort order.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -205,23 +205,24 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
...
                 tick()
                 subject.appendHistory(ResourceValid(resource, clock))
               }
 
-              test("the event is not included in the history") {
+              test("the existing event's timestamp is updated") {
                 expectThat(subject.eventHistory(resource.id))
                   .and {
                     first().isA<ResourceValid>()
+                      .and { get { timestamp }.isEqualTo(clock.instant()) }
```
luispollo, line 218: This is failing, and I don't understand why... 😭 
The update is logged as expected:
```
-> with bind values      : update `keel`.`event` set `keel`.`event`.`timestamp` = {ts '2020-05-09 00:40:48.588'} where (`keel`.`event`.`scope` = 'RESOURCE' and `keel`.`event`.`uid` = '01E7VE1BQMRY0KET3XPYHDH1WH') order by `keel`.`event`.`timestamp` desc limit 1
Affected row(s)          : 1
```
But the test still fails:
```
org.opentest4j.AssertionFailedError: ▼ Expect that [ResourceValid(kind=test/whatever@v1, id=test:whatever:37396637, application=fnord, timestamp=2020-05-09T00:40:47.588Z), ResourceUpdated(kind=test/whatever@v1, id=test:whatever:37396637, application=fnord, delta={}, timestamp=2020-05-09T00:40:46.588Z), ResourceCreated(kind=test/whatever@v1, id=test:whatever:37396637, application=fnord, timestamp=2020-05-09T00:40:45.588Z)]:
  ▼ first element ResourceValid(kind=test/whatever@v1, id=test:whatever:37396637, application=fnord, timestamp=2020-05-09T00:40:47.588Z):
    ✓ is an instance of com.netflix.spinnaker.keel.events.ResourceValid
    ▼ 2020-05-09T00:40:47.588Z:
      ✗ is equal to 2020-05-09T00:40:48.588Z
              found 2020-05-09T00:40:47.588Z
```
What am I missing?!

luispollo: What I was missing was to update the redundant `timestamp` field within the JSON... 😛 We should really remove that redundancy, it's a recipe for bugs.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
```diff
@@ -120,13 +120,19 @@ class InMemoryResourceRepository(
...
+    event: T
+  ) {
     eventList.computeIfAbsent(event.uid) {
       mutableListOf()
     }
       .let {
         val mostRecentEvent = it.firstOrNull() // we get the first because the list is in descending order
-        if (!event.ignoreRepeatedInHistory || event.javaClass != mostRecentEvent?.javaClass) {
+        if (event.javaClass == mostRecentEvent?.javaClass && event.ignoreRepeatedInHistory) {
+          it.removeAt(0)
```
gal-yardeni, line 133: What is happening here with the `add`/`remove`? it feels a bit dangerous.. can you add some comments?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -239,21 +239,35 @@ open class SqlResourceRepository(
...
             objectMapper.readValue<ResourceEvent>(json)
           }
 
-        if (event.javaClass == previousEvent?.javaClass) return@transaction
+        // If an equivalent event is already the latest in history, just update the timestamp
+        if (event.javaClass == previousEvent?.javaClass) {
+          txn
+            .update(EVENT)
+            .set(EVENT.TIMESTAMP, event.timestamp.atZone(clock.zone).toLocalDateTime())
+            // we udpate the JSON as well because it also contains a timestamp
```
emjburns, line 262: Oof, you're right. That is a recipe for bugs. Will you create an issue and we can rethink it (not for this pr obviously)?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -239,21 +239,35 @@ open class SqlResourceRepository(
...
+            .set(EVENT.JSON, objectMapper.writeValueAsString(event))
+            .where(EVENT.SCOPE.eq(Scope.RESOURCE.name))
+            .and(EVENT.UID.eq(resourceUid))
+            .orderBy(EVENT.TIMESTAMP.desc())
+            .limit(1)
```
emjburns, line 267: I assume this ordering and limiting is so you grab the most recent event? I think it would be more efficient to just match on the timestamp of the `previousEvent`, something like `.and(EVENT.TIMESTAMP.eq(previousEvent.timestamp)`? Or, we might have to also fetch the timestamp from the previous event, but I think that's going to be more efficient than fetching all of them here and then ordering them again

luispollo: Good idea. We should really add a proper PK to this table (I believe it was one of the suggestions we got from DET Platform). Should I also add the event type to the query to avoid the risk of two resource events emitted at the “same time” (ResourceDeltaResolved followed by ResourceValid?), or would those happen in separate check iterations?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepository.kt
luispollo: Unfortunately there’s no `replace` method on `List`, and I can’t change the existing event because it’s a data class. But this is effective just replacing the event with another one of the same type but with a newer timestamp. And this class is going away next week after @robfletcher’s PR. 😊

erikmunson:
If I'm understanding the way this works correctly (which I'm certainly not certain of) I have thoughts on it from a couple different angles:

1. I'm sort of neutral or ambivalent about the UX impact, not necessarily because I don't think there's a meaningful change in behavior, but because I'm not sure which behavior is more intuitive to people. With the existing behavior, if something became valid 4 hours ago and you're trying to figure out whether things are happening for some reason (i.e. you're worried we got stuck), you'd probably wish the timestamp was updated so you could see action. In the proposed behavior, it could become annoying if you're trying to figure out when the resource _changed states_ — e.g. when it went from actuating/resumed/something else to a valid state. I don't really know which of those experiences will happen more frequently, how annoying either of them will be, or if there's other observations about either I'm missing

2. Separate from the experience, there are some technical considerations that make me a little uneasy about it. One of the original design goals (that really isn't reflected or documented in the code itself) that @robfletcher and I tried to stick to was keeping history events strictly immutable — at least at the API level. In other words events were originally conceived to be like log messages that represent a static, unchanging point in time, not mutable "identities" like tasks which show whatever the latest state is of the logical operation/thing they represent. That design assumption is baked in to the UI in some places, in part because being able to determine whether or not a bit of nested or complex data has changed based on one or two fields that are primitive types (like `timestamp` and/or `type`) is one of the most important performance optimizations we can make throughout UI code (it makes it possible to prevent rendering sections of the page that haven't changed). [Here's an example](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx#L118) of a spot where we prevent re-rendering the whole table on every polling cycle by looking for events by timestamp to give you an idea of what that looks like

Happy to talk through the nuances more and see if we can get more clarity, I don't know for sure if I'm actually understanding the nature of this PR fully.

luispollo:
> One of the original design goals (that really isn't reflected or documented in the code itself) that @robfletcher and I tried to stick to was keeping history events strictly immutable — at least at the API level.

I know... This was my one argument against this approach when discussing with @emjburns, but I actually forgot that one of the reasons was to allow the UI to optimize for performance. 

The initial approach I’d taken in the PR was to override the “ignore repeated” flag in resource events when there was a more recent application event, allowing the new resource event (`ResourceValid`, in the big we found) to be recorded in the database. But when Emily came up with the suggestion to update the existing event, it seemed like a simpler approach.

To be honest, my preference would be to refactor the way we’re modeling paused/resumed events at the resource and application level, and how they interact, because right now it feels kind of hacky. But that would take time and be a bigger change, which I didn’t think was the right thing to do at this particular time.

I can definitely go back to my original approach, which doesn’t suffer from the immutability problem, or if you have any other ideas about how we could approach this, I’d be happy to hear them. I’ll also update the description of the PR to make it clearer.

robfletcher:
> To be honest, my preference would be to refactor the way we’re modeling paused/resumed events at the resource and application level, and how they interact, because right now it feels kind of hacky.

Yeah, I'd definitely like to see us tackle this. I'm not sure what the correct approach is.

erikmunson:
👍 It's sounding like there's a strong desire to rework this whole area (which I totally agree with) so we can discuss that separately, but will chime in to say I think given the user impact it'd be good to try to get _some_ kind of hacky solution in (like the one Luis mentioned of re-emitting by ignoring app level events) that gets us the user experience we need now so there's sufficient time to consider a new approach without time pressure making things rushed. As-is we need to either rollback indefinitely or get a fix for the user-facing problem out.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
emjburns: Those would happen on separate check iterations so I think you should be good with just time!

luispollo:
Replaced with https://github.com/spinnaker/keel/pull/1159.

PR #1151, opened 2020-05-11
Misc DB changes following review

robfletcher:


keel-sql/src/main/resources/db/changelog/20200511-misc-platform-review-changes.yml
```diff
@@ -0,0 +1,45 @@
...
+        uniqueColumns: resource_uid
+    - addPrimaryKey:
+        tableName: artifact_last_checked
+        columnNames: artifact_uid
+    - addPrimaryKey:
+        tableName: delivery_config_last_checked
+        columnNames: delivery_config_uid
+    - addPrimaryKey:
+        tableName: resource_last_checked
+        columnNames: resource_uid
```
robfletcher, line 26: Replace unique keys with primary keys

keel-sql/src/main/resources/db/changelog/20200511-misc-platform-review-changes.yml
```diff
@@ -0,0 +1,45 @@
...
+    - renameColumn:
+        tableName: event
+        oldColumnName: uid
+        newColumnName: ref
+        columnDataType: varchar(255)
```
robfletcher, line 31: I think the intent of this column was causing confusion (it confused me). Renaming & will add a separate `uid` for the event itself in a following PR.

keel-sql/src/main/resources/db/changelog/20200511-misc-platform-review-changes.yml
```diff
@@ -0,0 +1,45 @@
...
+    - dropIndex:
+        tableName: artifact_versions
+        indexName: artifact_versions_artifact_idx
+    - dropIndex:
+        tableName: delivery_config_artifact
+        indexName: delivery_config_artifact_unique_idx
+    - dropIndex:
+        tableName: environment_resource
+        indexName: environment_resource_unique_idx
```
robfletcher, line 40: Dropped redundant indices

keel-sql/src/main/resources/db/changelog/20200511-misc-platform-review-changes.yml
```diff
@@ -0,0 +1,45 @@
...
+    - sql:
+        comment: make JSON columns last in the table for performance
+        sql: |
+          alter table event modify column json json after scope;
+          alter table delivery_artifact modify column details json after reference;
```
robfletcher, line 45: Moved columns to end of table

emjburns:
Do you have any worry about rolling these all out at once?

robfletcher:
> Do you have any worry about rolling these all out at once?

I deliberately didn't handle the couple of wider-reaching changes in this PR. I think these ones are pretty minor.

keel-sql/src/main/resources/db/changelog/20200511-misc-platform-review-changes.yml
```diff
@@ -0,0 +1,45 @@
...
+    - addPrimaryKey:
+        tableName: artifact_last_checked
+        columnNames: artifact_uid
+    - addPrimaryKey:
+        tableName: delivery_config_last_checked
+        columnNames: delivery_config_uid
+    - addPrimaryKey:
+        tableName: resource_last_checked
+        columnNames: resource_uid
+    - renameColumn:
```
luispollo, line 27: This should be a separate changeset in this changelog, no?

keel-sql/src/main/resources/db/changelog/20200511-misc-platform-review-changes.yml
luispollo: These should be a separate changeset in this changelog, no?

luispollo: P.S. I believe there was a redundant index in `event` as well (`event_scope_uid_idx`).

keel-sql/src/main/resources/db/changelog/20200511-misc-platform-review-changes.yml
```diff
@@ -0,0 +1,45 @@
...
+        uniqueColumns: artifact_uid
+    - dropUniqueConstraint:
+        tableName: delivery_config_last_checked
+        constraintName: delivery_config_uid
+        uniqueColumns: delivery_config_uid
+    - dropUniqueConstraint:
+        tableName: resource_last_checked
+        constraintName: resource_uid
+        uniqueColumns: resource_uid
+    - addPrimaryKey:
```
marchello2000, line 18: i assume this was generated by jooq? Will it just convert these multiple primary keys to a single composite primary key?

marchello2000: I am not sure, but I think the syntax here is wrong:
https://www.liquibase.org/documentation/changes/add_primary_key.html

i think it's supposed to be something like this:
```yaml
    - addPrimaryKey:
        tableName: delivery_config_last_checked
        columnNames: artifact_last_checked, delivery_config_uid, resource_last_checked
```
(instead of 3 adds)
also, assuming the order these columns is what you want - since they will comprise the PK index and will serve as the main lookup.

marchello2000: Sorry, ignore me, i am an idiot... these are different tables... 🤦

robfletcher:
I separated out the changesets

robfletcher:
> this makes sense to me, but going to caveat it with "this is the second time I am looking at keel source so not sure if anything else needs to change", but I assume since the model is encoded in code and if it compiles - it runs :)

Yeah, I did the simpler changes all together in this PR. There are a couple of more involved ones you guys suggested that I'll tackle separately.

PR #1152, opened 2020-05-11
Fix CI build on GitHub Actions

robfletcher:


robfletcher:
> @robfletcher Just a heads-up that this behavior had been explicitly changed because of another bug: #1017

It's important they're singleton Spring beans, but not the same object. If tests are going to reconfigure things that's 100% going to leak state between tests.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -92,37 +108,40 @@ class ArtifactListener(
...
-          }
-          if (latestVersion != null) {
-            val hasNew = when {
-              lastRecordedVersion == null -> true
-              latestVersion != lastRecordedVersion -> {
-                artifact.versioningStrategy.comparator.compare(lastRecordedVersion, latestVersion) > 0
-              }
-              else -> false
+  fun syncArtifactVersions() {
+    if (enabled.get()) {
```
robfletcher, line 112: I noticed some log traffic related to this scheduled method running when Spring tests were executing so I made it discovery-dependent like the other scheduled things

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
```diff
@@ -19,20 +19,15 @@ import de.huxhorn.sulky.ulid.ULID
...
-fun configuredObjectMapper(): ObjectMapper =
-  objectMapperInstance
+fun configuredObjectMapper(): ObjectMapper = ObjectMapper().configureMe()
 
 /**
  * Factory method for [YAMLMapper]s configured how we like 'em.
  */
-fun configuredYamlMapper(): YAMLMapper =
-  yamlMapperInstance.disable(USE_NATIVE_TYPE_ID)
+fun configuredYamlMapper(): YAMLMapper = YAMLMapper().configureMe().disable(USE_NATIVE_TYPE_ID)
```
robfletcher, line 30: This was the ultimate problem. Because `ExceptionHandlerTests` messes with the registration of subtypes on these mappers, and they're package-level variables rather than just singleton Spring beans, the changes that test makes leak into any subsequent tests

keel-sql/src/test/resources/application.properties
```diff
@@ -1,3 +1,4 @@
+eureka.defaultToUp=false
```
robfletcher, line 1: I've consistently specified this everywhere we use Spring tests so that our scheduled stuff should never be running in the background when tests are running

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -88,10 +89,10 @@ class ExceptionHandler(
   @ResponseStatus(FORBIDDEN)
   fun onAccessDenied(e: AccessDeniedException): ApiError {
     log.error(e.message)
-    val authFailure = FiatPermissionEvaluator.getAuthorizationFailure().get()
+    val authFailure: AuthorizationFailure? = FiatPermissionEvaluator.getAuthorizationFailure().orElse(null)
```
robfletcher, line 92: This was also throwing exceptions. It didn't seem to break any tests directly but it was certainly incorrect.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigControllerTests.kt
```diff
@@ -279,8 +282,8 @@ internal class DeliveryConfigControllerTests : JUnit5Minutests {
         derivedContext<ResultActions>("the submitted manifest is missing a required field") {
           fixture {
             val mapper = when (contentType) {
-              APPLICATION_YAML -> configuredYamlMapper()
-              else -> configuredObjectMapper()
+              APPLICATION_YAML -> yamlMapper
+              else -> jsonMapper
```
robfletcher, line 286: This didn't ultimately matter, but I think it makes sense to use the injected Spring bean mappers here rather than creating new ones

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
lorin: nice

luispollo:
> > @robfletcher Just a heads-up that this behavior had been explicitly changed because of another bug: #1017
> 
> It's important they're singleton Spring beans, but not the same object. If tests are going to reconfigure things that's 100% going to leak state between tests.

It's two different problems that play into each other. Understood and agreed with the reasoning around the tests, but as I pointed out in the description of that other PR, the problem is that we do further customization of the mappers beyond what the `configureMe()` function does, at Spring application startup time, and none of the instances obtained via `configuredXyzMapper()` in various places around the codebase (i.e. instances that are _not_ injected by Spring), will be missing that configuration.

https://github.com/spinnaker/keel/blob/0091096e4e1bfbe304c7c75b6b60f8bf03a818a5/keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt#L48-L57

robfletcher:
> It's two different problems that play into each other. Understood and agreed with the reasoning around the tests, but as I pointed out in the description of that other PR, the problem is that we do further customization of the mappers beyond what the `configureMe()` function does, at Spring application startup time, and all the instances obtained via `configuredXyzMapper()` in various places around the codebase (i.e. instances that are _not_ injected by Spring), will be missing that configuration.

Yes, but given you can't guarantee the mappers will have been thru a Spring test, I don't see what difference it makes. For any non-Spring tests that need subtypes registered we'll need to make that code callable outside the context of Spring.

luispollo:
> Yes, but given you can't guarantee the mappers will have been thru a Spring test, I don't see what difference it makes. For any non-Spring tests that need subtypes registered we'll need to make that code callable outside the context of Spring.

To be clear, my concern was not with tests, but with the [many other places](https://github.com/spinnaker/keel/search?q=configuredObjectMapper+-filename%3ATests&unscoped_q=configuredObjectMapper+-filename%3ATests) where we call `configuredObjectMapper()` to get an instance. Having said that, I see that, in the end, using a single instance was not what fixed the problem we had with that other bug, so I'll leave it to you to make that call. 🙂

keel-sql/src/test/resources/application.properties
emjburns: amazing

robfletcher:
> To be clear, my concern was not with tests, but with the [many other places](https://github.com/spinnaker/keel/search?q=configuredObjectMapper+-filename%3ATests&unscoped_q=configuredObjectMapper+-filename%3ATests) where we call `configuredObjectMapper()` to get an instance. Having said that, I see that, in the end, using a single instance was not what fixed the problem we had with that other bug, so I'll leave it to you to make that call. 🙂

Yeah, sorry I read your comment properly later and updated my response ^^^

Issue #1126, opened 2020-05-06
Flapping status/unhappy status not showing


emjburns:
A resource with history like this:
<img width="1191" alt="Screen Shot 2020-05-06 at 10 24 26 AM" src="https://user-images.githubusercontent.com/8454927/81208485-f8731200-8f83-11ea-813a-11cc65375eaf.png">

should have an unhappy/flapping status. But it doesn't, it just gets the "actuating" status 
<img width="682" alt="Screen Shot 2020-05-06 at 10 26 27 AM" src="https://user-images.githubusercontent.com/8454927/81208537-10e32c80-8f84-11ea-94f0-665faa950118.png">

erikmunson:
I will add to this that i just got into a situation where i had to pause, and noticed something similar (can't be sure if it's related). tl;dr:
1. paused
2. saw a pause event in the history, but status remained `ACTUATING` like it was before the pause
3. unpaused, status went to `RESUMED` and seemed fine again

previously the resource would've gone to a `PAUSED` status immediately after pausing

emjburns:
Update - this is what the UI shows when we're constantly rolling back on a resource. Can't say for sure if this is the actual scenario here, but just a fun fact.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

PR #1157, opened 2020-05-12
feat(status): vetoed resources get history event + unhappy status

emjburns:
I was investigating https://github.com/spinnaker/keel/issues/1126 (which I can't reproduce locally). I found it frustrating that we weren't persisting veto events for flapping resources, so i updated our code to add that event and then determine whether a resource is unhappy by either the fact that it's vetoed or that it has been flapping.

This means that a resource that is vetoed because it's flapping or because it doesn't have the required resources (sec group or lb) will have a history event that says vetoed and a status of `unhappy`. @erikmunson and I thought that `unhappy` was clearer to users than `vetoed`.

Thoughts?

UPDATE:
added a `MISSING_DEPENDENCY` status to be specific for the missing load balancer and missing sg stuff.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -133,7 +133,13 @@ class ResourceActuator(
           }
           log.debug("Skipping actuation for resource {} because it was vetoed: {}", id, response.message)
           publisher.publishEvent(ResourceCheckSkipped(resource.kind, id, response.vetoName))
-          publishVetoedEvent(response, resource)
+          publisher.publishEvent(
+            ResourceActuationVetoed(
+              resource.kind,
+              resource.id,
+              resource.spec.application,
+              response.message,
+              clock.instant()))
           return@withTracingContext
         }
 
@@ -217,24 +223,6 @@ class ResourceActuator(
       val d = desired.await()
       val c = current.await()
       d to c
-    }
-
-  /**
-   * We want a specific status for specific types of vetos. This function publishes the
-   * right event based on which veto said no.
-   */
-  private fun publishVetoedEvent(response: VetoResponse, resource: Resource<*>) =
-    when {
-      response.vetoName == "UnhappyVeto" -> {
-        // don't publish an event, we want the status to stay as "unhappy" for clarity
-      }
-      else -> publisher.publishEvent(
-        ResourceActuationVetoed(
-          resource.kind,
-          resource.id,
-          resource.spec.application,
-          response.message,
-          clock.instant()))
     }
 
   private fun DeliveryConfig.environmentFor(resource: Resource<*>): Environment? =
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -141,6 +141,7 @@ class ResourceActuator(
               resource.spec.application,
               response.message,
               response.vetoName,
+              response.suggestedStatus,
               clock.instant()))
           return@withTracingContext
         }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -94,6 +94,7 @@ class ResourceActuator(
            * containing [resource]. This ensures that the environment will be fully restored to
            * a prior good-state.
            */
+          // todo eb: limit this to not roll back forever
           if (response.vetoArtifact && resource.spec is VersionedArtifactProvider) {
             try {
               val versionedArtifact = when (desired) {
@@ -139,6 +140,7 @@ class ResourceActuator(
               resource.id,
               resource.spec.application,
               response.message,
+              response.vetoName,
               clock.instant()))
           return@withTracingContext
         }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
@@ -34,6 +34,7 @@ import com.netflix.spinnaker.keel.events.ResourceState.Error
 import com.netflix.spinnaker.keel.events.ResourceState.Missing
 import com.netflix.spinnaker.keel.events.ResourceState.Ok
 import com.netflix.spinnaker.keel.events.ResourceState.Unresolvable
+import com.netflix.spinnaker.keel.persistence.ResourceStatus
 import com.netflix.spinnaker.kork.exceptions.SpinnakerException
 import com.netflix.spinnaker.kork.exceptions.SystemException
 import com.netflix.spinnaker.kork.exceptions.UserException
@@ -238,25 +239,27 @@ data class ResourceActuationPaused(
  * Actuation on the managed resource has been vetoed.
  *
  * @property reason The reason why actuation was vetoed.
- * @property vetoer The veto that vetoed this resource
+ * @property veto The veto that vetoed this resource
  */
 data class ResourceActuationVetoed(
   override val kind: ResourceKind,
   override val id: String,
   override val application: String,
   val reason: String?,
-  val vetoer: String? = null,
+  val veto: String? = null,
+  val suggestedStatus: ResourceStatus? = null,
   override val timestamp: Instant
 ) : ResourceEvent(message = reason) {
   @JsonIgnore
   override val ignoreRepeatedInHistory = true
 
-  constructor(resource: Resource<*>, reason: String?, vetoer: String? = null, clock: Clock = Companion.clock) : this(
+  constructor(resource: Resource<*>, reason: String?, veto: String? = null, suggestedStatus: ResourceStatus? = null, clock: Clock = Companion.clock) : this(
     resource.kind,
     resource.id,
     resource.application,
     reason,
-    vetoer,
+    veto,
+    suggestedStatus,
     clock.instant()
   )
 }
@@ -395,8 +398,10 @@ data class ResourceCheckUnresolvable(
 enum class ResourceCheckErrorOrigin {
   @JsonProperty("user")
   USER,
+
   @JsonProperty("system")
   SYSTEM,
+
   @JsonProperty("unknown")
   UNKNOWN;
 
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
@@ -238,22 +238,25 @@ data class ResourceActuationPaused(
  * Actuation on the managed resource has been vetoed.
  *
  * @property reason The reason why actuation was vetoed.
+ * @property vetoer The veto that vetoed this resource
  */
 data class ResourceActuationVetoed(
   override val kind: ResourceKind,
   override val id: String,
   override val application: String,
   val reason: String?,
+  val vetoer: String? = null,
   override val timestamp: Instant
 ) : ResourceEvent(message = reason) {
   @JsonIgnore
   override val ignoreRepeatedInHistory = true
 
-  constructor(resource: Resource<*>, reason: String?, clock: Clock = Companion.clock) : this(
+  constructor(resource: Resource<*>, reason: String?, vetoer: String? = null, clock: Clock = Companion.clock) : this(
     resource.kind,
     resource.id,
     resource.application,
     reason,
+    vetoer,
     clock.instant()
   )
 }
@@ -390,9 +393,12 @@ data class ResourceCheckUnresolvable(
 }
 
 enum class ResourceCheckErrorOrigin {
-  @JsonProperty("user") USER,
-  @JsonProperty("system") SYSTEM,
-  @JsonProperty("unknown") UNKNOWN;
+  @JsonProperty("user")
+  USER,
+  @JsonProperty("system")
+  SYSTEM,
+  @JsonProperty("unknown")
+  UNKNOWN;
 
   companion object {
     val log: Logger by lazy { LoggerFactory.getLogger(ResourceCheckErrorOrigin::class.java) }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
@@ -145,5 +145,15 @@ class NoSuchResourceId(id: String) :
   NoSuchResourceException("No resource with id $id exists in the database")
 
 enum class ResourceStatus {
-  HAPPY, ACTUATING, UNHAPPY, CREATED, DIFF, ERROR, CURRENTLY_UNRESOLVABLE, PAUSED, VETOED, RESUMED, UNKNOWN
+  CREATED,
+  DIFF,
+  ACTUATING,
+  HAPPY,
+  UNHAPPY,
+  MISSING_DEPENDENCY,
+  CURRENTLY_UNRESOLVABLE,
+  ERROR,
+  PAUSED,
+  RESUMED,
+  UNKNOWN
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
@@ -137,8 +137,7 @@ class ResourceStatusService(
 
   /**
    * Looks at the veto event and determines if it was vetoed by any of the [Required*Veto]s, which indicate a
-   * missing dependency.
-   * Looks at both the [veto] and the [reason] for backwards compatibility.
+   * missing dependency. Parses this information from the [reason]. This is used for backwards compatibility.
    */
   private fun ResourceActuationVetoed.isMissingDependency(): Boolean =
     reason?.contains("is not found in", true) ?: false
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
@@ -23,6 +23,7 @@ import com.netflix.spinnaker.keel.persistence.ResourceStatus.CURRENTLY_UNRESOLVA
 import com.netflix.spinnaker.keel.persistence.ResourceStatus.DIFF
 import com.netflix.spinnaker.keel.persistence.ResourceStatus.ERROR
 import com.netflix.spinnaker.keel.persistence.ResourceStatus.HAPPY
+import com.netflix.spinnaker.keel.persistence.ResourceStatus.MISSING_DEPENDENCY
 import com.netflix.spinnaker.keel.persistence.ResourceStatus.PAUSED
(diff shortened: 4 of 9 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceHistoryService.kt
```diff
@@ -136,9 +130,17 @@ class ResourceHistoryService(
   }
 
   /**
-   * Checks last 10 events for flapping between only ResourceActuationLaunched and ResourceDeltaDetected
+   * Returns true if a resource has been vetoed, or if the last 10 events are only
+   * ResourceActuationLaunched or ResourceDeltaDetected events.
+   *
+   * Even though we could return the status of vetoed we think that unhappy is more clear to users
+   * and doesn't leak implementation terminology that might not be helpful to users.
```
luispollo, line 137: I totally agree. I'm not sure what wording/icon we were using to surface that status on the UI, but as a user I'd have no clue what "vetoed" means.

luispollo:
I agree 100% that "vetoed" is not helpful as a status for users. On the other hand, the wording around the "unhappy" status [currently](https://www.spinnaker.io/guides/user/managed-delivery/resource-status/#unhappy) is very specific and calls out that keel is unable to reconcile desired and current state for the resource, so I just wanted to make sure I understand where a resource is vetoed in the first place -- it is after we've attempted actuation a maximum configured number of times and failed to reconcile, correct? If that's the case, then I think "unhappy" fits perfectly for that case as well.

My 2 cents.

erikmunson:
Very much agreed with the approach here re: flapping and also Luis's commentary about making sure we're modeling the _cause_ of the veto accurately, and treating the veto itself as a somewhat meaningless (from a UX perspective) implementation detail.

Given that, [I opened an issue with a short proposal](https://github.com/spinnaker/keel/issues/1162) to add a new `MISSING_DEPENDENCY` status which would take over the case where a resource is waiting on things like LBs/security groups (or other kinds in the future). That would leave us with `UNHAPPY` for flapping, and `MISSING_DEPENDENCY` for missing things, and no exposure of the underlying veto.

Feedback welcome, if we think that's a good approach than I'd suggest trying to do it alongside this change (no idea if it makes sense to use the same PR or different) so we ship something that I can write fully baked UI code against.

emjburns:
Thanks @luispollo and @erikmunson for your thoughts! I really like the proposal of the specific missing dependency status, and I'll update this PR to add that status and do the right thing in the cases you both mentioned. (I'll wait on https://github.com/spinnaker/keel/pull/1159 though because I'm gonna have to update this pr anyways). Both cases (stopping because we're flapping and stopping because we have a missing dependency) are modeled as vetos now, and I might need to clean up some code around how we save that event. TBD! But fun things will come out of this.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -145,5 +145,15 @@ class NoSuchResourceId(id: String) :
   NoSuchResourceException("No resource with id $id exists in the database")
 
 enum class ResourceStatus {
-  HAPPY, ACTUATING, UNHAPPY, CREATED, DIFF, ERROR, CURRENTLY_UNRESOLVABLE, PAUSED, VETOED, RESUMED, UNKNOWN
+  CREATED,
+  DIFF,
+  ACTUATING,
+  HAPPY,
+  UNHAPPY,
+  MISSING_DEPENDENCY,
```
emjburns, line 153: new status

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
```diff
@@ -103,15 +103,46 @@ class ResourceStatusService(
...
+    // a catch all for vetoes that don't have an explicit status so they don't get unknown status.
+    return first() is ResourceActuationVetoed
+  }
+
+  /**
+   * Looks at the veto event and determines if it was vetoed by the UnhappyVeto.
+   * Looks at both the [vetoer] and the [reason] for backwards compatibility.
+   */
+  private fun ResourceActuationVetoed.isUnhappy(): Boolean =
+    vetoer?.contains("unhappy", true) ?: false || reason?.contains("unhappy", true) ?: false
```
emjburns, line 130: `vetoer` is a new field, which defaults to `null`. In order to be backwards compatible we look at both the `vetoer` or we parse the `reason`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
```diff
@@ -103,15 +103,46 @@ class ResourceStatusService(
...
+    first() is ResourceActuationVetoed && (first() as ResourceActuationVetoed).isMissingDependency()
+
+  /**
+   * Looks at the veto event and determines if it was vetoed by any of the [Required*Veto]s, which indicate a
+   * missing dependency.
+   * Looks at both the [vetoer] and the [reason] for backwards compatibility.
+   */
+  private fun ResourceActuationVetoed.isMissingDependency(): Boolean {
+    val isVetoNew = vetoer?.startsWith("Required", true) ?: false && vetoer?.endsWith("Veto", true) ?: false
+    val isVetoOld = reason?.contains("is not found in", true) ?: false
```
erikmunson, line 145: this is some deck level backwards compat shenanigans love it ❤️ 🙈

erikmunson:
👍 excited to see this happening! thanks for the fast followup, I'll try to get some UI in there ASAP after you ship.

only thing that jumped out to me at a high level was that doing string matching on values that (i think??) correspond to statically typed enums (i.e. the starts with `Resource` and ends with `Veto` checks) leaves open the possibility of things getting refactored/renamed in the rest of the code but someone accidentally not catching/realizing this code is also involved and has a reliance on that name. I have no idea:
1) if that's actually bad here
2) if the unit tests actually protect against that and I just can't tell because I don't know how to read them in depth
3) if there's a reasonable or low-cost way like some extra comments to to mitigate that risk

emjburns:
@erikmunson That's actually the class names and not enums: so even more brittle. But I can't think of a better way to do things right now :( I built in a fallback, so if we can't figure out what kind of veto it is with the string parsing we will just fallback to veto == `unhappy` status. I think that's a decent thing. That means things won't break if there's a refactor. However, I'll add a comment on each of the missing-dep vetos

erikmunson:
👍 👍 sounds good, I just noticed it as a code smell and wanted to point it out, a quick set of comments seems appropriately matched to the level of risk IMO

robfletcher:
I really like moving  toward surfacing the  dependency veto  information. I'm a little concerned about  some of  the complexity of  the logic around determining status.

emjburns:
@robfletcher same re: status. I think that'll evolve to something simpler after beta (and after we have a good idea for how to do it)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -219,24 +227,6 @@ class ResourceActuator(
       d to c
```
luispollo, line 227: Side-comment: _man_, this is hard to read!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -238,22 +238,25 @@ data class ResourceActuationPaused(
  * Actuation on the managed resource has been vetoed.
  *
  * @property reason The reason why actuation was vetoed.
+ * @property vetoer The veto that vetoed this resource
```
luispollo, line 241: nit: If it's the veto, why call it `vetoer` and not `veto`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
```diff
@@ -103,15 +103,46 @@ class ResourceStatusService(
   }
 
   /**
-   * Checks last 10 events for flapping between only ResourceActuationLaunched and ResourceDeltaDetected
+   * Returns true if a resource has been vetoed by the unhappy veto,
+   * or if the last 10 events are only ResourceActuationLaunched or ResourceDeltaDetected events,
+   * or if the resource has been vetoed by an unspecified veto that we don't have an explicit status mapping for.
    */
   private fun List<ResourceHistoryEvent>.isUnhappy(): Boolean {
+    if (first() is ResourceActuationVetoed && (first() as ResourceActuationVetoed).isUnhappy()) {
```
luispollo, line 111: I think you could simplify this as:
```suggestion
    if (first() as? ResourceActuationVetoed)?.isUnhappy()) {
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
```diff
@@ -103,15 +103,46 @@ class ResourceStatusService(
   }
 
   /**
-   * Checks last 10 events for flapping between only ResourceActuationLaunched and ResourceDeltaDetected
+   * Returns true if a resource has been vetoed by the unhappy veto,
+   * or if the last 10 events are only ResourceActuationLaunched or ResourceDeltaDetected events,
+   * or if the resource has been vetoed by an unspecified veto that we don't have an explicit status mapping for.
```
luispollo, line 108: Oof! This is a bit involved. But I don't have a great suggestion to simplify at the moment... 😕

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
```diff
@@ -103,15 +103,46 @@ class ResourceStatusService(
...
+   * Looks at both the [vetoer] and the [reason] for backwards compatibility.
+   */
+  private fun ResourceActuationVetoed.isUnhappy(): Boolean =
+    vetoer?.contains("unhappy", true) ?: false || reason?.contains("unhappy", true) ?: false
+
+  /**
+   * Determines if last event was a veto because of a missing dependency
+   */
+  private fun List<ResourceHistoryEvent>.isMissingDependency(): Boolean =
+    first() is ResourceActuationVetoed && (first() as ResourceActuationVetoed).isMissingDependency()
```
luispollo, line 136:
```suggestion
    (first() as? ResourceActuationVetoed)?.isMissingDependency()
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
luispollo: You could  add a new `isRequiredDependencyVeto()` function to the `Veto` interface that you check here, instead of relying on the class names.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
```diff
@@ -17,6 +17,10 @@ import kotlinx.coroutines.withContext
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
 
+/**
+ * Note: when resources are vetoed by this veto the [ResourceStatusService] infers the
+ * correct status by string matching the name of the veto.
```
luispollo, line 22: See comment above about adding a new interface function to `Veto` instead.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
luispollo: Actually, instead of storing the name of the veto, you could replace this with a boolean flag called `dueToMissingDependency` or similar, since that's actually what we want to check.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
emjburns: @luispollo I like the idea of adding to the interface, I'll think on that and hopefully improve this!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
emjburns: I think that's harder to read, personally

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -238,22 +239,27 @@ data class ResourceActuationPaused(
  * Actuation on the managed resource has been vetoed.
  *
  * @property reason The reason why actuation was vetoed.
+ * @property veto The veto that vetoed this resource
  */
 data class ResourceActuationVetoed(
   override val kind: ResourceKind,
   override val id: String,
   override val application: String,
   val reason: String?,
```
emjburns, line 248: Updated to `veto`, not using that field anymore, but I think it's really useful so I kept it.
Added `suggestedStatus`, which is set by the veto, so that it can tell you what it thinks the status should be (and the `ResourceStatusService` has the final say or can override it. I think this is a lot less brittle!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
```diff
@@ -103,15 +99,47 @@ class ResourceStatusService(
...
-    return true
+    return false
   }
+
+  /**
+   * Determines if last event was a veto because of a missing dependency
+   */
+  private fun List<ResourceHistoryEvent>.isMissingDependency(): Boolean =
+    first() is ResourceActuationVetoed && (first() as ResourceActuationVetoed).getStatus() == MISSING_DEPENDENCY
+
```
emjburns, line 124: This function now looks at the new `suggestedStatus` field and helps figure out what status a vetoed resource should have.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
```diff
@@ -103,15 +99,47 @@ class ResourceStatusService(
...
+      else -> UNHAPPY
+    }
+
+  /**
+   * Looks at the veto event and determines if it was vetoed by any of the [Required*Veto]s, which indicate a
+   * missing dependency.
+   * Looks at both the [veto] and the [reason] for backwards compatibility.
+   */
+  private fun ResourceActuationVetoed.isMissingDependency(): Boolean =
+    reason?.contains("is not found in", true) ?: false
```
emjburns, line 144: Some lovely backwards compatability.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
```diff
@@ -41,7 +42,8 @@ class RequiredLoadBalancerVeto(
     } else {
       deniedResponse(
         message = missingLoadBalancers.joinToString(separator = "\n", transform = MissingDependency::message),
-        vetoArtifact = false
+        vetoArtifact = false,
+        suggestedStatus = MISSING_DEPENDENCY
```
emjburns, line 46: Now the `RequiredLoadBalancerVeto` and the `RequiredSecurityGroupVeto` both set the correct status, which will propagate through

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
luispollo: I like!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
```diff
@@ -103,15 +99,47 @@ class ResourceStatusService(
...
+      // we can determine missing dependency by parsing the message
+      isMissingDependency() -> MISSING_DEPENDENCY
+      // all vetos get unhappy status if not specified
+      else -> UNHAPPY
+    }
+
+  /**
+   * Looks at the veto event and determines if it was vetoed by any of the [Required*Veto]s, which indicate a
+   * missing dependency.
+   * Looks at both the [veto] and the [reason] for backwards compatibility.
```
luispollo, line 141:
```suggestion
   * Looks at the [reason] of the veto event to determine if it was vetoed by any of the [Required*Veto]s, which indicate a missing dependency. This is done for backwards compatibility with old events prior to the introduction of the [suggestedStatus] field.
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusServiceTests.kt
```diff
@@ -192,13 +192,43 @@ class ResourceStatusServiceTests : JUnit5Minutests {
...
           }
 
-          test("returns VETOED") {
-            expectThat(subject.getStatus(resource.id)).isEqualTo(VETOED)
+          test("returns UNHAPPY") {
+            expectThat(subject.getStatus(resource.id)).isEqualTo(UNHAPPY)
+          }
+        }
+
+        context("when last event is ResourceActuationVetoed because new veto that doesn't provide a status") {
```
luispollo, line 205: This test seems redundant with the one above.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusServiceTests.kt
```diff
@@ -192,13 +192,43 @@ class ResourceStatusServiceTests : JUnit5Minutests {
...
+          before {
+            repository.appendResourceHistory(ResourceActuationVetoed(resource, "oh no", "meeeee", null, clock))
+          }
+
+          test("returns UNHAPPY") {
+            expectThat(subject.getStatus(resource.id)).isEqualTo(UNHAPPY)
+          }
+        }
+
+        context("when last event is ResourceActuationVetoed because missing dependency") {
```
luispollo, line 215:
```suggestion
        context("when last event is ResourceActuationVetoed due to a missing dependency") {
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusServiceTests.kt
```diff
@@ -192,13 +192,43 @@ class ResourceStatusServiceTests : JUnit5Minutests {
...
+          before {
+            repository.appendResourceHistory(ResourceActuationVetoed(resource, "vetoed", "RequiredBlahVeto", MISSING_DEPENDENCY, clock))
+          }
+
+          test("returns MISSING_DEPENDENCY") {
+            expectThat(subject.getStatus(resource.id)).isEqualTo(MISSING_DEPENDENCY)
+          }
+        }
+
+        context("when last event is ResourceActuationVetoed because missing dependency, but an older version of keel saved this code") {
```
luispollo, line 225:
```suggestion
        context("when last event is ResourceActuationVetoed due to a missing dependency, but saved by an older version of keel") {
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusServiceTests.kt
emjburns: Sure, combining the tests.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
emjburns: This got simplified a little, which is nice

PR #1159, opened 2020-05-13
fix(events): Simplify resource event history

luispollo:
Attempt to fix an issue identified recently where resources got stuck in RESUMED status, and hopefully prevent future pitfalls around resource events/status.

The cause of this particular problem was that we look at the event history to determine status, and, in this case, there were `ApplicationActuationResumed` events appearing last in the event history for those resources. (We also noticed the application paused/resumed events were not displayed on the UI, making the experience even more confusing, but that’s a separate issue).

Upon investigation, we discovered that keel behaved as expected after an application is resumed and proceeded to check the resources and emitted `ResourceValid` events. The problem was that these events have the `ignoreRepeatedInHistory` flag set to `true`, which causes them not to be persisted in the database if the last resource event is of the same type (which it was). In other words, we were ignoring those application-level events when checking for duplicates in the resource event history, and so a new `ResourceValid` event was never written to the DB.

This PR introduces the following changes to simplify the handling and interaction between resource events and application events that affect resource status, without going into a full-blown refactoring of this area of the system:
- Accept and embrace the fact that we have a single table in the database for all events and modify SQL queries around events so that we include _both_ resource events _and_ those application events that affect resource status. This allowed me to remove all the hacky event injection stuff (with certain assumptions, see below).
- To clearly indicate what events belong in the resource history, introduce a new `ResourceHistoryEvent` interface, which is implemented by all resource events, and only those application events that affect resource status (currently `ApplicationActuationPaused` and `ApplicationActuationResumed`).
- Change functions in related repository interfaces to use the new `ResourceHistoryEvent` type above.
- Refactored tests around resource event history in attempt to make them easier to read and understand.

Assumptions:
- When a resource is paused, and then deleted, the expectation is that the resource stops being managed, and so the pause record for that resource should also be deleted from the database. If the resource is later resubmitted, keel will start managing it.
- When an application is paused, and an associated delivery config is deleted, we cannot make the same assumption about an application pause record because, although we currently have a 1:1 mapping between applications and delivery configs, the UX does not expose the concept of delivery configs, only applications, and the user never explicitly unpaused the application, so we leave it alone.
- Similarly, resource events are deleted when a resource is deleted, but application events are not when a delivery config is deleted.

I hope these are reasonable assumptions and this PR provides an acceptable short-term compromise until we can redesign this further.

@erikmunson Tagging you to review the assumptions above.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
@@ -238,11 +238,10 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
           }
 
           test("the event is not included in the resource history") {
-            expectThat(subject.eventHistory(resource.id))
-              .and {
-                first().isA<ResourceValid>()
-                second().not().isA<ResourceValid>()
-              }
+            expectThat(subject.eventHistory(resource.id)) {
+              first().isA<ResourceValid>()
+              second().not().isA<ResourceValid>()
+            }
           }
         }
 
@@ -255,11 +254,10 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
           }
 
           test("the events are included in the resource history") {
-            expectThat(subject.eventHistory(resource.id))
-              .and {
-                first().isA<ApplicationActuationResumed>()
-                second().isA<ApplicationActuationPaused>()
-              }
+            expectThat(subject.eventHistory(resource.id)) {
+              first().isA<ApplicationActuationResumed>()
+              second().isA<ApplicationActuationPaused>()
+            }
           }
         }
 
@@ -276,13 +274,12 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
           }
 
           test("the event is included in the resource history") {
-            expectThat(subject.eventHistory(resource.id))
-              .and {
-                first().isA<ResourceValid>()
-                second().isA<ApplicationActuationResumed>()
-                third().isA<ApplicationActuationPaused>()
-                fourth().isA<ResourceValid>()
-              }
+            expectThat(subject.eventHistory(resource.id)) {
+              first().isA<ResourceValid>()
+              second().isA<ApplicationActuationResumed>()
+              third().isA<ApplicationActuationPaused>()
+              fourth().isA<ResourceValid>()
+            }
           }
         }
       }
@@ -363,7 +360,7 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
         }
 
         test("event history includes relevant application events") {
-          expectThat(subject.eventHistory(resource.id)).and {
+          expectThat(subject.eventHistory(resource.id)) {
             first().isA<ApplicationActuationResumed>()
             second().isA<ApplicationActuationPaused>()
           }
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
@@ -17,7 +17,10 @@ package com.netflix.spinnaker.keel.persistence
 
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Resource
+import com.netflix.spinnaker.keel.api.application
 import com.netflix.spinnaker.keel.api.id
+import com.netflix.spinnaker.keel.events.ApplicationActuationPaused
+import com.netflix.spinnaker.keel.events.ApplicationActuationResumed
 import com.netflix.spinnaker.keel.events.ResourceActuationLaunched
 import com.netflix.spinnaker.keel.events.ResourceCreated
 import com.netflix.spinnaker.keel.events.ResourceDeltaDetected
@@ -69,6 +72,7 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
       application = "toast",
       serviceAccount = "keel@spinnaker"
     ),
+    val user: String = "keel@keel.io",
     val subject: T,
     val callback: (ResourceHeader) -> Unit = mockk(relaxed = true) // has to be relaxed due to https://github.com/mockk/mockk/issues/272
   )
@@ -177,70 +181,116 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
         }
       }
 
-      context("updating the state of the resource") {
-        context("an event that should be persisted in history") {
+      context("updating the event history of the resource") {
+        context("appending a resource event") {
           before {
             tick()
-            // TODO: ensure persisting a map with actual data
-            subject.appendHistory(ResourceUpdated(resource, emptyMap(), clock))
+            subject.appendHistory(ResourceUpdated(resource, mapOf("delta" to "some-difference"), clock))
           }
 
-          test("the event is included in the history") {
+          test("the event is included in the resource history") {
             expectThat(subject.eventHistory(resource.id))
               .hasSize(2)
               .first()
               .isA<ResourceUpdated>()
+              .get { delta }.isEqualTo(mapOf("delta" to "some-difference"))
+          }
+        }
+
+        context("appending an identical resource event with duplicates allowed") {
+          before {
+            tick()
+            subject.appendHistory(ResourceUpdated(resource, emptyMap(), clock))
+            tick()
+            subject.appendHistory(ResourceUpdated(resource, emptyMap(), clock))
           }
 
-          context("updating the state again") {
-            before {
-              tick()
-              subject.appendHistory(ResourceValid(resource, clock))
-            }
-
-            test("the event is included in the history") {
-              expectThat(subject.eventHistory(resource.id))
-                .hasSize(3)
-                .first()
-                .isA<ResourceValid>()
-            }
-
-            context("a subsequent identical event that should be ignored") {
-              before {
-                tick()
-                subject.appendHistory(ResourceValid(resource, clock))
+          test("the event is included in the resource history") {
+            expectThat(subject.eventHistory(resource.id))
+              .hasSize(3)
+              .and {
+                first().isA<ResourceUpdated>()
+                second().isA<ResourceUpdated>()
               }
+          }
+        }
 
-              test("the event is not included in the history") {
-                expectThat(subject.eventHistory(resource.id))
-                  .and {
-                    first().isA<ResourceValid>()
-                    second().not().isA<ResourceValid>()
-                  }
-              }
-            }
+        context("appending a resource event with duplicates disallowed the first time") {
+          before {
+            tick()
+            subject.appendHistory(ResourceValid(resource, clock))
           }
 
-          context("a subsequent identical event") {
-            before {
-              tick()
-              subject.appendHistory(ResourceUpdated(resource, emptyMap(), clock))
-            }
+          test("the event is included in the resource history") {
+            expectThat(subject.eventHistory(resource.id))
+              .hasSize(2)
+              .first()
+              .isA<ResourceValid>()
+          }
+        }
 
-            test("the new state is included in the history") {
-              expectThat(subject.eventHistory(resource.id))
-                .hasSize(3)
-                .and {
-                  first().isA<ResourceUpdated>()
-                  second().isA<ResourceUpdated>()
-                }
-            }
+        context("appending an identical resource event with duplicates disallowed the second time") {
+          before {
+            tick()
+            subject.appendHistory(ResourceValid(resource, clock))
+            tick()
+            subject.appendHistory(ResourceValid(resource, clock))
+          }
+
+          test("the event is not included in the resource history") {
+            expectThat(subject.eventHistory(resource.id))
+              .and {
+                first().isA<ResourceValid>()
+                second().not().isA<ResourceValid>()
+              }
+          }
+        }
+
+        context("appending application events that affect resource history") {
+          before {
+            tick()
+            subject.appendHistory(ApplicationActuationPaused(resource.application, user, clock))
+            tick()
+            subject.appendHistory(ApplicationActuationResumed(resource.application, user, clock))
+          }
+
+          test("the events are included in the resource history") {
+            expectThat(subject.eventHistory(resource.id))
+              .and {
+                first().isA<ApplicationActuationResumed>()
+                second().isA<ApplicationActuationPaused>()
+              }
+          }
+        }
+
+        context("appending a resource event identical to the last resource event, but with a relevant application events in between") {
+          before {
+            tick()
+            subject.appendHistory(ResourceValid(resource, clock))
+            tick()
+            subject.appendHistory(ApplicationActuationPaused(resource.application, user, clock))
+            tick()
+            subject.appendHistory(ApplicationActuationResumed(resource.application, user, clock))
+            tick()
+            subject.appendHistory(ResourceValid(resource, clock))
+          }
+
+          test("the event is included in the resource history") {
+            expectThat(subject.eventHistory(resource.id))
+              .and {
+                first().isA<ResourceValid>()
+                second().isA<ApplicationActuationResumed>()
+                third().isA<ApplicationActuationPaused>()
+                fourth().isA<ResourceValid>()
+              }
           }
         }
       }
 
       context("deleting the resource") {
         before {
+          subject.appendHistory(ApplicationActuationPaused(resource.application, user, clock))
+          subject.appendHistory(ApplicationActuationResumed(resource.application, user, clock))
           subject.delete(resource.id)
         }
 
@@ -261,18 +311,27 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
             subject.eventHistory(resource.id)
           }
         }
+
+        test("events for the resource's parent application remain") {
+          expectThat(
+            subject.applicationEventHistory(resource.application)
+          ).hasSize(2)
+        }
       }
 
       context("fetching event history for the resource") {
         before {
           repeat(3) {
-            clock.incrementBy(TEN_MINUTES)
+            tick()
             subject.appendHistory(ResourceUpdated(resource, emptyMap(), clock))
             subject.appendHistory(ResourceDeltaDetected(resource, emptyMap(), clock))
             subject.appendHistory(ResourceActuationLaunched(resource, "whatever", emptyList(), clock))
             subject.appendHistory(ResourceDeltaResolved(resource, clock))
           }
-          clock.incrementBy(TEN_MINUTES)
+          tick()
+          subject.appendHistory(ApplicationActuationPaused(resource.application, user, clock))
+          tick()
+          subject.appendHistory(ApplicationActuationResumed(resource.application, user, clock))
         }
 
         test("default limit is 10 events") {
@@ -288,7 +347,7 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
 
         test("the limit can be higher than the default") {
           expectThat(subject.eventHistory(resource.id, limit = 20))
-            .hasSize(13)
+            .hasSize(15)
         }
 
         test("zero limit is not allowed") {
@@ -302,17 +361,23 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
             .isFailure()
             .isA<IllegalArgumentException>()
         }
+
+        test("event history includes relevant application events") {
+          expectThat(subject.eventHistory(resource.id)).and {
+            first().isA<ApplicationActuationResumed>()
+            second().isA<ApplicationActuationPaused>()
+          }
+        }
       }
     }
   }
 
   private fun tick() {
-    clock.incrementBy(ONE_SECOND)
+    clock.incrementBy(ONE_MINUTE)
   }
 
   companion object {
-    val ONE_SECOND: Duration = Duration.ofSeconds(1)
-    val TEN_MINUTES: Duration = Duration.ofMinutes(10)
+    val ONE_MINUTE: Duration = Duration.ofMinutes(1)
   }
 
   fun <T : Iterable<E>, E : ResourceEvent> Assertion.Builder<T>.areNoOlderThan(age: Period): Assertion.Builder<T> =
@@ -332,6 +397,12 @@ fun <T : Any> Assertion.Builder<T>.isNotIn(expected: Collection<T>) =
 
 fun <T : List<E>, E> Assertion.Builder<T>.second(): Assertion.Builder<E> =
   get("second element %s") { this[1] }
+
+fun <T : List<E>, E> Assertion.Builder<T>.third(): Assertion.Builder<E> =
+  get("third element %s") { this[2] }
+
+fun <T : List<E>, E> Assertion.Builder<T>.fourth(): Assertion.Builder<E> =
+  get("fourth element %s") { this[3] }
 
 fun randomString(length: Int = 8) =
   randomUUID()
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
@@ -33,7 +33,7 @@ data class ApplicationActuationPaused(
   override val application: String,
   override val timestamp: Instant,
   override val triggeredBy: String?
-) : ApplicationEvent() {
+) : ApplicationEvent(), ResourceHistoryEvent {
   @JsonIgnore
   override val ignoreRepeatedInHistory = true
 
@@ -51,7 +51,7 @@ data class ApplicationActuationResumed(
   override val application: String,
   override val triggeredBy: String?,
   override val timestamp: Instant
-) : ApplicationEvent() {
+) : ApplicationEvent(), ResourceHistoryEvent {
   @JsonIgnore override val ignoreRepeatedInHistory = true
 
   constructor(application: String, triggeredBy: String, clock: Clock = Companion.clock) : this(
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
@@ -13,8 +13,8 @@ import java.time.Instant
   include = JsonTypeInfo.As.PROPERTY
 )
 @JsonSubTypes(
-  JsonSubTypes.Type(value = ApplicationEvent::class, name = "application"),
-  JsonSubTypes.Type(value = ResourceEvent::class, name = "resource")
+  JsonSubTypes.Type(value = ApplicationEvent::class),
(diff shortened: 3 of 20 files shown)


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -177,70 +181,116 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
         }
       }
 
-      context("updating the state of the resource") {
-        context("an event that should be persisted in history") {
+      context("updating the event history of the resource") {
```
luispollo, line 184: There was never anything within this `context` about updating a resource, per se, but rather about updating the event history of the resource, so I reworded it and the nested contexts and tests accordingly.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
```diff
@@ -33,7 +33,7 @@ data class ApplicationActuationPaused(
   override val application: String,
   override val timestamp: Instant,
   override val triggeredBy: String?
-) : ApplicationEvent() {
+) : ApplicationEvent(), ResourceHistoryEvent {
```
luispollo, line 36: The new `ResourceHistoryEvent` interface provides a common type for events of any scope that belong in the resource event history.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
```diff
@@ -34,3 +34,24 @@ abstract class PersistentEvent {
...
+  use = JsonTypeInfo.Id.NAME,
+  property = "type",
+  include = JsonTypeInfo.As.PROPERTY
+)
+@JsonSubTypes(
+  JsonSubTypes.Type(value = ResourceEvent::class),
+  JsonSubTypes.Type(value = ApplicationActuationPaused::class),
+  JsonSubTypes.Type(value = ApplicationActuationResumed::class)
+)
+interface ResourceHistoryEvent {
```
luispollo, line 52: The new `ResourceHistoryEvent` interface provides a common type for events of any scope that belong in the resource event history.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -67,7 +67,7 @@ import org.slf4j.LoggerFactory
 abstract class ResourceEvent(
   open val message: String? = null,
   override val triggeredBy: String? = null
-) : PersistentEvent() {
+) : PersistentEvent(), ResourceHistoryEvent {
```
luispollo, line 70: The new `ResourceHistoryEvent` interface provides a common type for events of any scope that belong in the resource event history.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
```diff
@@ -92,44 +89,4 @@ class ActuationPauser(
 
   fun pausedApplications(): List<String> =
     pausedRepository.getPausedApplications()
-
-  /**
-   * Adds [ApplicationActuationPaused] and [ApplicationActuationResumed] events to the resource event history so that
-   * it reflects actuation pauses/resumes at the application level.
-   */
-  fun addApplicationActuationEvents(originalEvents: List<ResourceEvent>, resource: Resource<*>) =
```
luispollo, line 100: All this hacky stuff is now gone as we simply retrieve the right events from the data store.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -357,10 +358,10 @@ class CombinedRepository(
   override fun applicationEventHistory(application: String, downTo: Instant): List<ApplicationEvent> =
     resourceRepository.applicationEventHistory(application, downTo)
 
-  override fun resourceEventHistory(id: String, limit: Int): List<ResourceEvent> =
+  override fun resourceEventHistory(id: String, limit: Int): List<ResourceHistoryEvent> =
```
luispollo, line 361: Again, the resource event history is now composed of events implementing `ResourceHistoryEvent`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceHistoryService.kt
```diff
@@ -48,22 +45,10 @@ import org.springframework.stereotype.Component
...
-  private val repository: KeelRepository,
+  private val resourceRepository: ResourceRepository,
   private val actuationPauser: ActuationPauser
 ) {
 
-  /**
-   * Returns the history of events associated with the specified resource from the database, plus any applicable
-   * application-level pause/resume events.
-   */
-  fun getEnrichedEventHistory(id: String, limit: Int = DEFAULT_MAX_EVENTS): List<PersistentEvent> {
```
luispollo, line 59: Another function gone in favor of simply querying the right records from the data store.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ResourceHistoryServiceTests.kt
```diff
@@ -70,124 +84,176 @@ class ResourceHistoryServiceTests : JUnit5Minutests {
...
-          expectThat(events)
-            .map { it.javaClass }
-            .containsExactly(
-              ApplicationActuationPaused::class.java,
-              ApplicationActuationResumed::class.java,
-              ApplicationActuationPaused::class.java,
-              ResourceValid::class.java,
-              ResourceCreated::class.java
-            )
+      context("resource status") {
```
luispollo, line 95: I realized we didn't have tests for turning events or other information from the data store into resource statuses, so I added these.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -183,90 +191,83 @@ open class SqlResourceRepository(
...
-        .and(RESOURCE.UID.eq(EVENT.UID))
+        .from(EVENT)
+        // look for resource events that match the resource...
+        .where(EVENT.SCOPE.eq(Scope.RESOURCE.name)
+          .and(EVENT.UID.eq(resource.uid))
+        )
+        // ...or application events that match the application as they apply to all resources
+        .or(EVENT.SCOPE.eq(Scope.APPLICATION.name)
+          .and(EVENT_JSON_APPLICATION.eq(resource.application))
+        )
```
luispollo, line 210: This is the gist of the change to query the right event records from the database, which hopefully is pretty straightforward to read.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -183,90 +191,83 @@ open class SqlResourceRepository(
...
-          .and(RESOURCE.UID.eq(EVENT.UID))
+          .from(EVENT)
+          // look for resource events that match the resource...
+          .where(EVENT.SCOPE.eq(Scope.RESOURCE.name)
+            .and(EVENT.UID.eq(uid))
+          )
+          // ...or application events that match the application as they apply to all resources
+          .or(EVENT.SCOPE.eq(Scope.APPLICATION.name)
+            .and(EVENT_JSON_APPLICATION.eq(event.application))
+          )
```
luispollo, line 248: Similar query change here.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -291,6 +292,12 @@ open class SqlResourceRepository(
         .where(DIFF_FINGERPRINT.ENTITY_ID.eq(id))
         .execute()
     }
+    sqlRetry.withRetry(WRITE) {
+      jooq.deleteFrom(PAUSED)
```
luispollo, line 296: This was previously missing, and now enforces the assumption I mentioned in the PR description that, when a resource is deleted, a matching pause record should also be deleted as the resource effectively stops being managed.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -177,70 +181,116 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
...
+          test("the event is included in the resource history") {
             expectThat(subject.eventHistory(resource.id))
               .hasSize(2)
               .first()
               .isA<ResourceUpdated>()
+              .get { delta }.isEqualTo(mapOf("delta" to "some-difference"))
+          }
+        }
+
+        context("appending an identical resource event with duplicates allowed") {
```
luispollo, line 200: I pulled these out from under the context because I find nested tests hard to reason about. I know these causes some duplication of code in setting up the tests in the `before` block, but I much prefer that than having to read through multiple nested contexts to understand what the setup for a particular test is.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -183,90 +191,83 @@ open class SqlResourceRepository(
...
-          .orderBy(EVENT.TIMESTAMP.desc())
-          .limit(1)
-          .fetchOne()
-          ?.let { (json) ->
-            objectMapper.readValue<ApplicationEvent>(json)
-          }
-
-        if (event.javaClass == previousEvent?.javaClass) return@transaction
-      }
+  // todo: add sql retries once we've rethought repository structure: https://github.com/spinnaker/keel/issues/740
```
luispollo, line 222: @emjburns I kept this comment here because I didn't quite understand the reasoning behind it, but the issue linked is closed. I don't understand why we can't add retries here and elsewhere in this class. Could you please explain?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -183,90 +191,83 @@ open class SqlResourceRepository(
...
+  override fun appendHistory(event: ResourceEvent) {
+    // for historical reasons, we use the resource UID (not the ID) as an identifier in resource events
+    val uid = jooq.select(RESOURCE.UID).from(RESOURCE).where(RESOURCE.ID.eq(event.uid)).fetchOne(RESOURCE.UID)
```
gal-yardeni, line 225: nit: formating

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -183,90 +191,83 @@ open class SqlResourceRepository(
...
         .fetch()
         .map { (json) ->
-          objectMapper.readValue<ResourceEvent>(json)
-        }
-        .ifEmpty {
-          throw NoSuchResourceId(id)
+          objectMapper.readValue<PersistentEvent>(json)
         }
+        // filter out application events that don't affect resource history
+        .filterIsInstance<ResourceHistoryEvent>()
```
gal-yardeni, line 218: this is much easier to understand!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
```diff
@@ -34,3 +34,24 @@ abstract class PersistentEvent {
...
+  val scope: PersistentEvent.Scope
+  val uid: String // the resource ID or application name
```
emjburns, line 54: is `uid` the right name for this variable then? what about `identifier`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
```diff
@@ -127,7 +129,7 @@ interface KeelRepository {
   // START ResourceRepository methods
   fun allResources(callback: (ResourceHeader) -> Unit)
 
-  fun getResource(id: String): Resource<out ResourceSpec>
```
emjburns, line 130: Why remove the `out` here? Asking because I'm extremely bad at understanding what it does in the first place, so looking to learn :D

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -183,90 +191,83 @@ open class SqlResourceRepository(
     }
   }
 
-  override fun eventHistory(id: String, limit: Int): List<ResourceEvent> {
+  override fun eventHistory(id: String, limit: Int): List<ResourceHistoryEvent> {
     require(limit > 0) { "limit must be a positive integer" }
+
+    val resource = get(id)
```
gal-yardeni, line 197: do we need to catch here a `resourceNotFound` exception?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
```diff
@@ -16,7 +16,7 @@ import org.springframework.web.bind.annotation.RestController
 @RestController
 @RequestMapping(path = ["/resources/events"])
 class EventController(
-  private val resourceHistoryService: ResourceHistoryService
+  private val resourceRepository: ResourceRepository
```
gal-yardeni, line 19: wasn't the whole point of the service to encapsulate the repository logic? :)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
gal-yardeni: +1

gal-yardeni:
This solution makes sense to me, and it quite easy to understand and the tests are great. Thanks for pursuing it!

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/events/ResourceStatusTests.kt
```diff
@@ -50,7 +50,7 @@ internal class ResourceStatusTests : JUnit5Minutests {
     val repository = combinedInMemoryRepository(clock)
```
emjburns, line 50: should this just change to a `ResourceRepository`, or is the combined repository used in the rest of the tests?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ResourceHistoryServiceTests.kt
```diff
@@ -2,42 +2,56 @@ package com.netflix.spinnaker.keel.services
...
 import strikt.api.expectThat
-import strikt.assertions.containsExactly
-import strikt.assertions.isA
 import strikt.assertions.isEqualTo
-import strikt.assertions.map
 
 class ResourceHistoryServiceTests : JUnit5Minutests {
   companion object Fixture {
     val clock = MutableClock()
     val repository = combinedInMemoryRepository(clock)
```
emjburns, line 49: should this change to a `ResourceRepository`, or is the combined repo used in other parts of this file?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
emjburns: tbh I don't remember why we couldn't do sql retries here. maybe it's not true!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -183,90 +191,83 @@ open class SqlResourceRepository(
...
+        .set(EVENT.SCOPE, event.scope.name)
+        .set(EVENT.UID, uid)
         .set(EVENT.TIMESTAMP, event.timestamp.atZone(clock.zone).toLocalDateTime())
         .set(EVENT.JSON, objectMapper.writeValueAsString(event))
         .execute()
     }
   }
 
   override fun delete(id: String) {
+    // TODO: these should be run inside a transaction
```
emjburns, line 270: why don't you just put them in a transaction?

luispollo: Trying to avoid too many things in one PR, especially those not directly tied to the issue being addressed, as one Professor Burns taught me. 😉

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ResourceHistoryServiceTests.kt
luispollo: It's not. I could change it, but didn't want to spend too much time on in-memory repository-related stuff in tests because they're all going away...

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
luispollo: Yes. 😞 But... I think the service still has value in that it's encapsulating the logic around determining resource status based on events (which might warrant changing the name of the class, BTW), whereas here we really just needed to retrieve the list of events directly from the data source _without_ any additional processing, and IMHO it's best to avoid additional indirection levels just to proxy calls where we can, because it makes troubleshooting much easier.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: No, I specifically wanted that to bubble up.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
luispollo: > Asking because I'm extremely bad at understanding what it does in the first place, so looking to learn :D

Same. 😆 

I removed it because the compiler (who knows much more about Kotlin than I do), said it was redundant/unnecessary. 🙂

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
luispollo: It's really not. I didn't change it here because @robfletcher renamed it to `ref` in #1151.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: What's the formatting issue here?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EventController.kt
gal-yardeni: ok, make sense!

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -302,17 +361,23 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
             .isFailure()
             .isA<IllegalArgumentException>()
         }
+
+        test("event history includes relevant application events") {
+          expectThat(subject.eventHistory(resource.id)).and {
```
robfletcher, line 366: I think you should be able to omit `.and` there. There's an override of `expectThat` taking a lambda as a 2nd parameter.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
```diff
@@ -13,8 +13,8 @@ import java.time.Instant
   include = JsonTypeInfo.As.PROPERTY
 )
 @JsonSubTypes(
-  JsonSubTypes.Type(value = ApplicationEvent::class, name = "application"),
-  JsonSubTypes.Type(value = ResourceEvent::class, name = "resource")
+  JsonSubTypes.Type(value = ApplicationEvent::class),
+  JsonSubTypes.Type(value = ResourceEvent::class)
```
robfletcher, line 17: Does this change have an impact on the resulting JSON?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/pause/ActuationPauser.kt
robfletcher: That's a relief!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
robfletcher: …and it says that because I added `out` variance on the interface a few days ago making it unnecessary to specify it at the use-site.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceHistoryService.kt
```diff
@@ -92,53 +77,44 @@ class ResourceHistoryService(
     }
   }
 
-  private fun List<PersistentEvent>.isHappy(): Boolean {
+  private fun List<ResourceHistoryEvent>.isHappy(): Boolean {
     return first() is ResourceValid || first() is ResourceDeltaResolved
   }
```
robfletcher, line 82: Couldn't the first element be an application paused/resumed event and still we're in a "happy" state?

I think you could potentially do `filterIsInstance<ResourceEvent>()` first (assuming I'm right about this).

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -183,90 +191,83 @@ open class SqlResourceRepository(
...
-        .and(RESOURCE.UID.eq(EVENT.REF))
+        .from(EVENT)
+        // look for resource events that match the resource...
+        .where(EVENT.SCOPE.eq(Scope.RESOURCE.name)
+          .and(EVENT.REF.eq(resource.uid))
+        )
+        // ...or application events that match the application as they apply to all resources
+        .or(EVENT.SCOPE.eq(Scope.APPLICATION.name)
+          .and(EVENT_JSON_APPLICATION.eq(resource.application))
+        )
```
robfletcher, line 210: Would it make sense to add an `application` column to the table to make this easier?

luispollo: Probably, yes.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -126,12 +133,38 @@ class ApplicationService(
...
-      resourceStatusService.getResourceSummariesFor(deliveryConfig)
+      return deliveryConfig.resources.map { resource ->
+        resource.toResourceSummary(deliveryConfig)
+      }
     } catch (e: NoSuchDeliveryConfigException) {
       emptyList()
     }
   }
 
+  private fun Resource<*>.toResourceSummary(deliveryConfig: DeliveryConfig) =
```
luispollo, line 144: @emjburns I managed to move this back here and out of the (now renamed) `ResourceHistoryService`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
luispollo: No. I always have to go back and look at the docs for this, but the `name` parameter is used is the _"logical type name used as the type identifier for the class"_, which, for abstract classes and interfaces, never ends up serialized anywhere. In other words, this only really matters in concrete objects.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceHistoryService.kt
luispollo: For the pause case, it could, but I short-circuit that case right above:
https://github.com/spinnaker/keel/blob/1a498d9759729a19066ecd280ae9b5b3d023e91f/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt#L48-L54

In theory, if there's ever a pause record in the DB, there should be a matching event for that scope. In other words, if we ever get to evaluate other statuses, there should never be one of those at the top of the list.

As for the resume case, it has its own status, handled by the `isResumed()` function.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: BTW, do you know how much of a performance hit we get from querying within JSON blobs? I know MySQL has optimizations for it, but that's the extend of my knowledge...

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceHistoryService.kt
robfletcher: So we cannot be both paused and happy?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
robfletcher: No idea. I suspect that's something we need to measure and optimize if/when we feel like we need to

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
robfletcher: ah, yeah. Missed that this was abstract. Of course.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceHistoryService.kt
emjburns: Nope, we can't. We only display that we're paused. when we resume, we will check again and then see if we're happy, at which point the status will change

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
luispollo: Done.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: Added https://github.com/spinnaker/keel/issues/1171 to track.

erikmunson:
very nice approach, thanks for finding a way to get things working again without a total rethink.

Issue #1153, opened 2020-05-12
Returns 500 against /delivery-configs/diff when name field not in payload


lorin:
If a request is missing information needed for authenticating, we currently return a 500 error due to an uncaught IllegalArgumentException.

For example: POST to `/delivery-configs/diff` and the body is missing a `name` field:

```
java.lang.IllegalArgumentException: Parameter specified as non-null is null: method com.netflix.spinnaker.keel.rest.AuthorizationSupport.hasApplicationPermission, parameter identifier
```

The payload should not require the name field to be present, so we need to update the auth check logic here so it doesn't rely on this field.

gal-yardeni:
Marking it as a bug as it expects to get the delivery config name as an identifier, and it is not mandatory anymore. We will have to do the proper adjustments to the authz checks. cc: @luispollo

lorin:
@gal-yardeni I changed the title and body to reflect your comment.

gal-yardeni:
thank you @lorin ! I'm already fixing it.

PR #1163, opened 2020-05-13
fix(configAuthz): not relying on delivery config name for authorization

gal-yardeni:
fixes https://github.com/spinnaker/keel/issues/1153.

The delivery config name is not mandatory anymore, but the application is. So for authorization checks, we will use the application name instead.

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
@@ -100,8 +100,7 @@ class AuthorizationSupport(
     withAuthentication(target, identifier) { auth ->
       val application = when (target) {
         RESOURCE -> repository.getResource(identifier).application
-        APPLICATION -> identifier
-        DELIVERY_CONFIG -> repository.getDeliveryConfig(identifier).application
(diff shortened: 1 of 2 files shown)


gal-yardeni:
@lorin the `DeliveryConfigController.get` and `DeliveryConfigController.delete` are taking the `name` as a path param, so it guarantees it is there.

lorin:
@gal-yardeni I'm worried about this situation:

* app name: `foo`
* delivery config name: `foo-manifest`

The user calls:

`GET /delivery-configs/foo-manifest`

This calls:

```
authorizationSupport.hasApplicationPermission('READ', 'DELIVERY_CONFIG',  "foo-manifest")
```

This code sets the application name to the identifier, which in this case would be the delivery config name, not the application name:

```
      val application = when (target) {
...
        APPLICATION, DELIVERY_CONFIG -> identifier
permission check")
...
      }

```

Then the subsequent application check would check against the `foo-manifest` application (which doesn't exist) isntead of the `foo` application:

```
permissionEvaluator.hasPermission(auth, application, "APPLICATION", action.name)
```

If delivery config names aren't enforced to be the same as application names, this seems like a problem.

gal-yardeni:
Good point! you are right. I think it would make sense to change the check for the application as well for these cases.

PR #1166, opened 2020-05-13
fix(deliveryConfig): support both application and config name

gal-yardeni:
following PR https://github.com/spinnaker/keel/pull/1163/, fixing the endpoints that take the application names as the identifier to use `application` as the target, instead of `deliveryConfig`

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
@@ -22,7 +22,6 @@ import com.netflix.spinnaker.keel.api.Locatable
 import com.netflix.spinnaker.keel.api.application
 import com.netflix.spinnaker.keel.api.serviceAccount
 import com.netflix.spinnaker.keel.persistence.KeelRepository
-import com.netflix.spinnaker.keel.persistence.NoSuchDeliveryConfigName
 import com.netflix.spinnaker.keel.persistence.NoSuchEntityException
 import com.netflix.spinnaker.keel.rest.AuthorizationSupport.TargetEntity.APPLICATION
 import com.netflix.spinnaker.keel.rest.AuthorizationSupport.TargetEntity.DELIVERY_CONFIG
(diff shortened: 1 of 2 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AuthorizationSupport.kt
```diff
@@ -100,7 +101,8 @@ class AuthorizationSupport(
     withAuthentication(target, identifier) { auth ->
       val application = when (target) {
         RESOURCE -> repository.getResource(identifier).application
-        APPLICATION, DELIVERY_CONFIG -> identifier
+        APPLICATION -> identifier
+        DELIVERY_CONFIG -> getDeliveryConfigApplication(identifier)
```
luispollo, line 105: This looks OK to me, but I don't understand why you would fall back to the identifier if you fail to find the delivery config by name... There's no guarantee that the identifier will match the application.

I'd suggest just reverting back to the original behavior:
```kotlin
DELIVERY_CONFIG -> repository.getDeliveryConfig(identifier).application
```
Or am I missing something?

gal-yardeni: it can be either the name of the application OR the name of the delivery config.. so I'm not quite sure how `repository.getDeliveryConfig(identifier).application` would work if you pass the application name and not the delivery config name

luispollo: Hmm... But you specify which one to use in the annotation... There shouldn’t be a need to do this unless there’s a controller parameter that can be _either_ the delivery config name _or_ the application name, but I don’t recall seeing anything like that.

luispollo: Hmm... But you specify which one to use in the annotation... There shouldn’t be a need to do this unless there’s a controller parameter that can be _either_ the delivery config name _or_ the application name, but I don’t recall seeing anything like that.

PR #1170, opened 2020-05-14
Make the current OkHttpClient bean as primary till we migrate to the new provider impl.

srekapalli:
Making the existing OkHttpClient bean as primary till we migrate to the Client provider implementation.


keel-retrofit/src/main/kotlin/com/netflix/spinnaker/keel/retrofit/KeelRetrofitConfiguration.kt
```diff
@@ -40,6 +41,7 @@ class KeelRetrofitConfiguration {
   private val log = LoggerFactory.getLogger(javaClass)
 
   @Bean(name = ["retrofitClient", "okClient"])
+  @Primary // TODO: (SR) -- Refactor it to use the new client provider.
```
lorin, line 44: This is going to make @robzienert cry!

robfletcher: It's not hard to make @robzienert cry

PR #1172, opened 2020-05-14
Fix up a bunch of compiler warnings

robfletcher:


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -397,7 +395,7 @@ enum class ResourceCheckErrorOrigin {
   @JsonProperty("unknown") UNKNOWN;
 
   companion object {
-    val log: Logger by lazy { LoggerFactory.getLogger(ResourceCheckErrorOrigin.javaClass) }
+    val log: Logger by lazy { LoggerFactory.getLogger(ResourceCheckErrorOrigin::class.java) }
```
luispollo, line 398: The difference between these two always trips me up!

PR #1173, opened 2020-05-14
fix(orca): Fix stupid Retrofit error

robfletcher:
This really bugs me, but it's a blocker so I've just reverted the change I made. When just using a plain map Retrofit throws a runtime exception ```java.lang.IllegalArgumentException: Parameter type must not include a type variable or wildcard: java.util.Map<java.lang.String, ?> (parameter #3)```


lorin:
According to https://github.com/square/retrofit/issues/3275, you can use `@JvmSuppressWildcard` as a workaround.

lorin:
More details on workaround syntax here: https://github.com/square/retrofit/issues/1805

robfletcher:
> According to [square/retrofit#3275](https://github.com/square/retrofit/issues/3275), you can use `@JvmSuppressWildcard` as a workaround.

Seems like the best option. I'll write a test that actually stands up a dummy HTTP back-end because this problem is not exposed when you use the Retrofit interface as a mock in a unit test. That makes it particularly nasty since it only then happens at runtime.

PR #1178, opened 2020-05-14
fix(misc): Use OkHttpClientProvider instead of the configuration bean…

srekapalli:
- Moving to `OkHttpClientProvider` to build OkHttp3 clients instead of `OkHttp3ClientConfiguration` .
- New provider implementation will allow us to do metatron/gandalf based AuthnZ for IPC calls.
- To configure metatron auth between services that keel talks to we need to put in following config.
```yaml
metatron:
    client:
       applications: ['clouddriver', 'orca', 'echo', 'front50']
```
-  Without the config , we will use the existing cert/x509 authentication.
- Few Autobump PR's ( fiat & kork ) failed due to how we wire up OkHttp3 clients and interceptors and the change here addresses those issues as well.

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/config/ClouddriverConfiguration.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/config/ClouddriverConfiguration.kt
@@ -16,13 +16,13 @@
 package com.netflix.spinnaker.config
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.config.okhttp3.OkHttpClientProvider
 import com.netflix.spinnaker.keel.clouddriver.CloudDriverCache
 import com.netflix.spinnaker.keel.clouddriver.CloudDriverService
 import com.netflix.spinnaker.keel.clouddriver.ImageService
 import com.netflix.spinnaker.keel.clouddriver.MemoryCloudDriverCache
 import okhttp3.HttpUrl
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
-import okhttp3.OkHttpClient
 import org.springframework.beans.factory.BeanCreationException
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
@@ -45,13 +45,13 @@ class ClouddriverConfiguration {
   fun clouddriverService(
     clouddriverEndpoint: HttpUrl,
     objectMapper: ObjectMapper,
-    retrofitClient: OkHttpClient
+    clientProvider: OkHttpClientProvider
   ):
     CloudDriverService =
     Retrofit.Builder()
       .addConverterFactory(JacksonConverterFactory.create(objectMapper))
       .baseUrl(clouddriverEndpoint)
-      .client(retrofitClient)
+      .client(clientProvider.getClient(DefaultServiceEndpoint("clouddriver", clouddriverEndpoint.toString())))
       .build()
       .create(CloudDriverService::class.java)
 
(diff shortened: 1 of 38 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/config/ClouddriverConfiguration.kt
```diff
@@ -45,13 +45,13 @@ class ClouddriverConfiguration {
...
     objectMapper: ObjectMapper,
-    retrofitClient: OkHttpClient
+    clientProvider: OkHttpClientProvider
   ):
     CloudDriverService =
     Retrofit.Builder()
       .addConverterFactory(JacksonConverterFactory.create(objectMapper))
       .baseUrl(clouddriverEndpoint)
-      .client(retrofitClient)
+      .client(clientProvider.getClient(DefaultServiceEndpoint("clouddriver", clouddriverEndpoint.toString())))
```
luispollo, line 54: @srekapalli @cfieber This pattern is repeated when instantiating every client. Is this a common enough pattern that it would be worth extracting into an override of `getClient()` that takes just the service name and the endpoint `HttpUrl` object?

keel-retrofit/src/main/kotlin/com/netflix/spinnaker/keel/retrofit/KeelRetrofitConfiguration.kt
```diff
@@ -17,73 +17,21 @@ package com.netflix.spinnaker.keel.retrofit
...
-        connectionPool(ConnectionPool(
-          okHttpClientProperties.connectionPool.maxIdleConnections,
-          okHttpClientProperties.connectionPool.keepAliveDurationMs.toLong(),
-          MILLISECONDS
-        ))
-        retryOnConnectionFailure(okHttpClientProperties.isRetryOnConnectionFailure)
-      }
-
-    builder.interceptors().also {
-      // If the metrics interceptor (which complains about X-SPINNAKER-* auth headers missing in the request)
```
luispollo, line 75: Is this ordering problem not an issue with the new client provider?

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/config/ClouddriverConfiguration.kt
```diff
@@ -45,13 +45,13 @@ class ClouddriverConfiguration {
   fun clouddriverService(
     clouddriverEndpoint: HttpUrl,
     objectMapper: ObjectMapper,
-    retrofitClient: OkHttpClient
+    clientProvider: OkHttpClientProvider
```
luispollo, line 48: I don't see where this is being instantiated. Was that introduced in a previous PR?

srekapalli: Yes. Here is the PR in kork.
https://github.com/spinnaker/kork/pull/639

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/config/ClouddriverConfiguration.kt
srekapalli: Yh. Struggled to find a common way to do that because we have several places where we work with Retrofit1 `Endpoint` object as well and some places where we using `String`.   Probably we will revisit this again and settle on a common thing.

luispollo: I see. No worries, thanks for the feedback.

keel-retrofit/src/main/kotlin/com/netflix/spinnaker/keel/retrofit/KeelRetrofitConfiguration.kt
srekapalli: I made the SpinnakerHeadersInterceptor to go first in the chain so that OkHttp3Metrics interceptor can take advantage of the data we are adding.

srekapalli:
> LGTM now. Might be good to add a comment about the importance of that `@Lazy` annotation to avoid the circular dependency.

Done!

PR #1181, opened 2020-05-15
2nd phase of recommended DB changes

robfletcher:


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -221,8 +221,7 @@ open class SqlResourceRepository(
 
   // todo: add sql retries once we've rethought repository structure: https://github.com/spinnaker/keel/issues/740
   override fun appendHistory(event: ResourceEvent) {
-    // for historical reasons, we use the resource UID (not the ID) as an identifier in resource events
-    val uid = jooq.select(RESOURCE.UID).from(RESOURCE).where(RESOURCE.ID.eq(event.uid)).fetchOne(RESOURCE.UID)
+    val uid = jooq.select(RESOURCE.UID).from(RESOURCE).where(RESOURCE.ID.eq(event.ref)).fetchOne(RESOURCE.UID)
```
luispollo, line 224: Consider renaming `uid` to `ref` here as well (and in the parameter to `doAppendHistory()` if it's not renamed already.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
```diff
@@ -19,7 +19,7 @@ import java.time.Instant
 abstract class PersistentEvent {
   abstract val scope: Scope
   abstract val application: String
-  abstract val uid: String // The unique ID of the thing associated with the scope. Defined in sub-classes.
+  abstract val ref: String // The unique ID of the thing associated with the scope. Defined in sub-classes.
```
luispollo, line 22: This will break deserialization for existing events in the DB, right? Are you planning to write a migration, or add code to the functions that retrieve events to account for that?

One trick I've used in the past, which I know is not great in terms of performance, but is simple and works, is to do something like this in those methods:
```kotlin
objectMapper.readValue<Map<String, Any?>>(json)
  .toMutableMap()
  .let {
    it.putIfAbsent("ref", it["uid"])
    objectMapper.convertValue(it, PersistentEvent::class.java)
  }
```

keel-sql/keel-sql.gradle.kts
```diff
@@ -70,11 +70,8 @@ dependencies {
   jooqModelatorRuntime(platform("com.netflix.spinnaker.kork:kork-bom:${property("korkVersion")}"))
   jooqModelatorRuntime("mysql:mysql-connector-java")
 
-  liquibaseRuntime("org.liquibase:liquibase-core:3.8.5")
-  liquibaseRuntime("org.liquibase:liquibase-groovy-dsl:2.0.1")
-  liquibaseRuntime("org.yaml:snakeyaml:1.25")
-  liquibaseRuntime("mysql:mysql-connector-java:8.0.16")
-  liquibaseRuntime(sourceSets.main.get().output)
+  liquibaseRuntime(platform("com.netflix.spinnaker.kork:kork-bom:${property("korkVersion")}"))
+  liquibaseRuntime("mysql:mysql-connector-java")
```
robfletcher, line 74: I did a bunch of wrestling with Liquibase that ultimately came to nothing, but I did discover we don't need all of these dependencies.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -262,6 +262,7 @@ open class SqlResourceRepository(
 
       txn
         .insertInto(EVENT)
+        .set(EVENT.UID, ULID().nextULID(event.timestamp.toEpochMilli()))
```
robfletcher, line 265: Ensure we assign `uid` to new events

keel-sql/src/main/kotlin/com/netflix/spinnaker/migrations/EventUidAssigner.kt
```diff
@@ -0,0 +1,73 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import org.jooq.DSLContext
+import org.jooq.Record1
+import org.slf4j.LoggerFactory
+import org.springframework.boot.context.event.ApplicationReadyEvent
+import org.springframework.context.event.EventListener
+
+class EventUidAssigner(private val jooq: DSLContext) : CoroutineScope {
```
robfletcher, line 17: This will (slowly) plod through all the existing events and assign them a unique `uid`. We're not using that field anywhere (yet?) so I think it can harmlessly do this in the background. Once it's done them all (which by my calculations would take a couple of hours on a single instance) we can make `uid` a primary key column

keel-sql/src/main/kotlin/com/netflix/spinnaker/migrations/EventUidAssigner.kt
```diff
@@ -0,0 +1,73 @@
...
+  @EventListener(ApplicationReadyEvent::class)
+  fun onApplicationReady() {
+    launch {
+      while (countEventsWithNoUID() > 0) {
+        var count = 0
+        jooq.inTransaction {
+          select(EVENT.TIMESTAMP)
+            .from(EVENT)
+            .where(EVENT.UID.isNull)
+            .limit(25)
```
robfletcher, line 31: I did a bit of back of a napkin perf analysis and this batch size seemed optimal for not holding the transaction open too long. It can do 25 in about a second on my machine. Bumping to 100 makes it take ~10 seconds.

keel-sql/src/main/kotlin/com/netflix/spinnaker/migrations/EventUidAssigner.kt
```diff
@@ -0,0 +1,73 @@
...
+            .forShare()
```
robfletcher, line 32: This should as I understand it, make it safe to have this running on multiple instances. The updates are decoupled from the select anyway so it shouldn't clash in the first place.

keel-sql/src/main/kotlin/com/netflix/spinnaker/migrations/EventUidAssigner.kt
```diff
@@ -0,0 +1,73 @@
...
+                  // this might actually not update the same row, but 🤷‍
+                  .where(EVENT.TIMESTAMP.eq(ts.value1()))
+                  .and(EVENT.UID.isNull)
+                  .limit(1)
+                  .execute()
+              }
+                .onSuccess { count += it }
+                .onFailure { ex ->
+                  log.error("Error assigning uid to event with timestamp ${ts.value1()}", ex)
+                }
```
robfletcher, line 47: This updates a single row with a matching timestamp to the one we just selected with a new `uid`. It doesn't really matter if it's the same exact row. There's no good uniqueness as it stands, so it's practically impossible to ensure we are updating the same row. However, _most_ event timestamps are unique. The `limit(1)` ensures we always assign a unique uid.

keel-sql/src/main/resources/db/changelog/20200515-give-event-table-a-pk.yml
```diff
@@ -0,0 +1,16 @@
...
+        tableName: event
+        columns:
+        - name: uid
+          type: varchar(255)
+          constraints:
+            nullable: true
+    - modifySql:
+        dbms: mysql
+        append:
+          value: " first"
```
robfletcher, line 16: Add a new `uid` column. Currently it will be nullable. We'll do a 2nd migration once every row has a value assigned.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
robfletcher: Argh, yes. I need to migrate that

keel-sql/src/main/resources/db/changelog/20200515-add-unique-id-to-event.yml
```diff
@@ -0,0 +1,34 @@
...
+        tableName: event
+        columns:
+        - name: uid
+          type: varchar(255)
+          constraints:
+            nullable: true
+    - modifySql:
+        dbms: mysql
+        append:
+          value: " first"
```
robfletcher, line 16: adds a (for now) nullable `uid` column to `event`

keel-sql/src/main/resources/db/changelog/20200515-add-unique-id-to-event.yml
```diff
@@ -0,0 +1,34 @@
...
+- changeSet:
+    id: add-ref-to-events-with-no-uid
+    author: fletch
+    changes:
+    - sql:
+        sql: >
+          UPDATE EVENT
+          SET JSON = JSON_INSERT(JSON, '$.ref', JSON->'$.id')
+          WHERE NOT(JSON_CONTAINS_PATH(json, 'one', '$.uid'))
```
robfletcher, line 25: Older existing events had no `uid`.  For those I've synthesized a `ref` from the `id` (which would be the same value)

keel-sql/src/main/resources/db/changelog/20200515-add-unique-id-to-event.yml
```diff
@@ -0,0 +1,34 @@
...
+- changeSet:
+    id: update-uid-to-ref-in-event-json
+    author: fletch
+    changes:
+    - sql:
+        sql: >
+          UPDATE EVENT
+          SET JSON = JSON_REMOVE(JSON_INSERT(JSON, '$.ref', JSON->'$.uid'), '$.uid')
+          WHERE JSON_CONTAINS_PATH(json, 'one', '$.uid')
```
robfletcher, line 34: This migrates newer events' JSON to change the old `uid` to `ref`

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: I hadn't seen an example initializing the ULID based on a timestamp before. Neat!

robfletcher: Yeah, the nice thing is that because I am doing the same on the old events, we can sort by `uid` and they will be in timestamp order

keel-sql/src/main/kotlin/com/netflix/spinnaker/migrations/EventUidAssigner.kt
luispollo: TIL.

luispollo: Why do you say the updates are decoupled though, if they're both running in the same transaction?

robfletcher: See comment on the update below

keel-sql/src/main/kotlin/com/netflix/spinnaker/migrations/EventUidAssigner.kt
luispollo: Yeah, that was pretty much the query I used when I had to do this before while testing locally. The only other idea I might offer would be to include `scope`, `json->'$.id'` and `json->'$.type'` in the criteria, which _should_ give you a single row.

keel-sql/src/main/kotlin/com/netflix/spinnaker/migrations/EventUidAssigner.kt
robfletcher: Actually, it's running a ton faster locally now so who knows if this could be further optimized? It's not really important though.

keel-sql/src/main/kotlin/com/netflix/spinnaker/migrations/EventUidAssigner.kt
robfletcher: Yeah, ultimately it's not important though. Either way we'll end up with all the rows having a unique ULID.

keel-sql/src/main/resources/db/changelog/20200515-add-unique-id-to-event.yml
```diff
@@ -0,0 +1,34 @@
+databaseChangeLog:
+- changeSet:
+    id: add-unique-id-to-event
+    author: fletch
+    changes:
+    - addColumn:
+        tableName: event
+        columns:
+        - name: uid
+          type: varchar(255)
```
luispollo, line 10: I thought we were consistently using the size of the ULID for these kinds of columns (26?).

keel-sql/src/main/resources/db/changelog/20200515-add-unique-id-to-event.yml
luispollo, line 16: Nice!

keel-sql/src/main/resources/db/changelog/20200515-add-unique-id-to-event.yml
robfletcher: Yeah, I wondered about that. I can easily change it. We should check other tables too.

keel-sql/src/main/resources/db/changelog/20200515-add-unique-id-to-event.yml
```diff
@@ -0,0 +1,34 @@
...
+          value: " first"
+- changeSet:
+    id: add-ref-to-events-with-no-uid
+    author: fletch
+    changes:
+    - sql:
+        sql: >
+          update event
+          set json = json_insert(json, '$.ref', json->'$.id')
+          where not(json_contains_path(json, 'one', '$.uid'))
```
luispollo, line 25: I don't understand what this is checking, especially given the second argument...

keel-sql/src/main/resources/db/changelog/20200515-add-unique-id-to-event.yml
robfletcher: Yeah, it took me a while to work it out as the built in options in Liquibase didn't work for me and I had to figure out how to do a MySQL native thing.

keel-sql/src/main/resources/db/changelog/20200515-add-unique-id-to-event.yml
robfletcher: Yeah, that param is weird. Docs are here: https://dev.mysql.com/doc/refman/5.7/en/json-search-functions.html#function_json-contains-path

robfletcher: …It's basically saying "where there's no `$.uid` in the JSON"

luispollo: Never mind. Man, that's a crappy syntax for that function.

robfletcher: Sure is.

keel-sql/src/main/resources/db/changelog/20200515-add-unique-id-to-event.yml
robfletcher: Fixed

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: Awesome!

keel-sql/src/main/kotlin/com/netflix/spinnaker/migrations/EventUidAssigner.kt
robzienert: Seems like this PR comment should be a comment on the class itself.

keel-sql/src/main/kotlin/com/netflix/spinnaker/migrations/EventUidAssigner.kt
robzienert: Honestly, smaller batches for this migration process is probably better than more anyway.

Issue #1100, opened 2020-05-01
manual judgement approval != deploy 


emjburns:
Scenerio:
- artifact # 535 and # 536 both deploy to prestaging, and both have manual judgments sent out
- I approve # 535 
- it does not start deploying

this is a bug.

![Screen Shot 2020-05-01 at 2 33 14 PM](https://user-images.githubusercontent.com/8454927/80844894-aa918f00-8bbc-11ea-875a-54717af3fed8.png)

emjburns:
This seems to apply even earlier than I thought - it applies even when the pipeline constraints are evaluating for the latest artifact, not just when they both are up for manual judgement

PR #1182, opened 2020-05-15
chore(artifacts): write test for approving old version

emjburns:
A test that reproduces https://github.com/spinnaker/keel/issues/1100

note: this behavior only fails in the sql implementation, not the in memory implementation. cc @lorin in case you were curious.

Note: test is commented out because it's breaking. Putting this up for @gal-yardeni  to use, but it's probably a needed test because this flow is COMPLICATED!

...also, I can't spell judgement...


luispollo:
> ...also, I can't spell judgement...

image

😄

PR #1186, opened 2020-05-15
fix(db): Fix race condition for resource record

luispollo:
Event persistence is handled asynchronously, so when storing a resource for the first time, and attempting to persist the corresponding `ResourceCreated` event, there's a race condition to read the `resource` record from the DB.

This fixes that by introducing retries.


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -225,15 +225,17 @@ open class SqlResourceRepository(
...
+    }
     doAppendHistory(event, uid)
   }
 
   override fun appendHistory(event: ApplicationEvent) {
     doAppendHistory(event, event.application)
   }
 
-  private fun doAppendHistory(event: PersistentEvent, ref: String) {
+  private fun doAppendHistory(event: PersistentEvent, ref: String?) {
```
emjburns, line 238: why is ref now `?`

luispollo: ah, good catch, will revert

Issue #1189, opened 2020-05-16
Upgrade resilience4j


luispollo:
We're currently on resilience4j 1.0.0. As of this writing, the latest stable version is 1.0.4, with many bug fixes and improvements for kotlin in between. We should consider upgrading.

luispollo:
@robfletcher, @lorin FYI.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

PR #1190, opened 2020-05-16
test(sql): Test retries around race condition storing/reading resource

luispollo:
Complement to #1138.

See also #1189.

### DIFF
--- a/keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepositoryTests.kt
+++ b/keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepositoryTests.kt
@@ -1,18 +1,24 @@
 package com.netflix.spinnaker.keel.sql
 
+import com.netflix.spinnaker.keel.api.id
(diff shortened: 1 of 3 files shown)


keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepositoryTests.kt
```diff
@@ -30,6 +36,37 @@ internal object SqlResourceRepositoryTests : ResourceRepositoryTests<SqlResource
...
+        Thread {
+          try {
+            subject.getResourceUid(resource.id)
+          } catch (e: IllegalStateException) {
+            exception = e
+          }
+        }.also {
+          it.start()
+          it.join()
+          expectThat(exception).isNull()
```
luispollo, line 64: @robfletcher I tried to write this with Strikt's `runCatching` but the assertion error was supressed, I think because it happened in the thread. I also tried running the calls in coroutines instead of threads, but I couldn't make that work (delaying/sleeping in one coroutine never seemed to have an effect). 

If you have thoughts on how to write this in a different way, please let me know.

PR #1191, opened 2020-05-17
fix(ApprovedVersion): Approved versions are now actually get deployed

gal-yardeni:
fixes https://github.com/spinnaker/keel/issues/1100.

The problem: 
1. Artifact `deb` has 2 versions: `version1` and `version2`, where `version1` is older. 
2. We have `manual-judgement` constraint for the environment, and both versions are in `PENDING` state.
3. I approve `version1` but nothing is happening.

The fix:
When changing a constraint state, in this case, `PENDING` -> `PASS` (or `OVERIDE_PASS`), the following is happening:
1. we persist the new state to `ENVIRONMENT_ARTIFACT_CONSTRAINT`
2. we insert to `CURRENT_CONSTRAINT` (which holds the current constraints for an application per environment)
3. we have another table called `ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL` which actually holds the approved version. When we change the state to pass, assuming all other stateful constraints have passed, we expect the approved version to persist there. 
3.1. we check if all states are passed, by reading the states from `ENVIRONMENT_ARTIFACT_CONSTRAINT`. 

Since step 1 and step 3 are happening in the same transaction, we get `PENDING` and _not_ the new state as we expected.
By splitting the transaction into 2 parts, the condition's results in step 3.1 are now as expected.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
@@ -97,10 +97,17 @@ class EnvironmentPromotionChecker(
                        * want to request judgement or deploy a canary for artifacts that aren't
                        * deployed to a required environment or outside of an allowed time.
                        */
-                      val passesConstraints =
-                        checkStatelessConstraints(artifact, deliveryConfig, v, environment) &&
-                          checkStatefulConstraints(artifact, deliveryConfig, v, environment)
 
+                      val passesStatelessConstraints: Boolean = checkStatelessConstraints(artifact, deliveryConfig, v, environment)
+                      val passesStatefulConstraints: Boolean = checkStatefulConstraints(artifact, deliveryConfig, v, environment)
+
+                      log.info("for version: [$v] of artifact ${artifact.name}: " +
+                        "passesStatelessConstraints: [$passesStatelessConstraints]" +
+                        ", passesStatefulConstraints [$passesStatefulConstraints] in" +
+                        "for environment ${environment.name}")
+
+                      val passesConstraints =
+                        passesStatelessConstraints && passesStatefulConstraints
                       versionIsPending = when (environment.constraints.anyStateful) {
                         true -> repository
                           .constraintStateFor(deliveryConfig.name, environment.name, v)
@@ -121,12 +128,21 @@ class EnvironmentPromotionChecker(
              */
             var approvedPending = false
             if (!hasPin) {
+              log.info("pendingVersionsToCheck: [$pendingVersionsToCheck] of artifact ${artifact.name} for environment ${environment.name} ")
               pendingVersionsToCheck
                 .sortedWith(artifact.versioningStrategy.comparator.reversed()) // oldest first
                 .forEach {
-                  val passesConstraints: Boolean =
-                    checkStatelessConstraints(artifact, deliveryConfig, it, environment) &&
-                      checkStatefulConstraints(artifact, deliveryConfig, it, environment)
+
+                  val passesStatelessConstraints: Boolean = checkStatelessConstraints(artifact, deliveryConfig, it, environment)
+                  val passesStatefulConstraints: Boolean = checkStatefulConstraints(artifact, deliveryConfig, it, environment)
+
+                  log.info("for version: [$it] of artifact ${artifact.name}: " +
+                    "passesStatelessConstraints: [$passesStatelessConstraints]" +
+                    ", passesStatefulConstraints [$passesStatefulConstraints] in" +
+                    "for environment ${environment.name}")
+
+                  val passesConstraints =
+                    passesStatelessConstraints && passesStatefulConstraints
 
                   if (passesConstraints) {
                     approveVersion(deliveryConfig, artifact, it, environment.name)
@@ -141,6 +157,7 @@ class EnvironmentPromotionChecker(
                    * We don't need to re-invoke stateful constraint evaluators for these, but we still
                    * check stateless constraints to avoid approval outside of allowed-times.
                    */
+                  log.info("version: [$v] of artifact ${artifact.name} is in queuedForApproval for environment ${environment.name}")
                   if (checkStatelessConstraints(artifact, deliveryConfig, v, environment)) {
                     approveVersion(deliveryConfig, artifact, v, environment.name)
                     repository.deleteQueuedConstraintApproval(deliveryConfig.name, environment.name, v)
@@ -148,7 +165,13 @@ class EnvironmentPromotionChecker(
                 }
             }
             if (!approvedPending && versionIsPending || version == null) {
-              log.warn("No version of {} passes constraints for environment {}", artifact.name, environment.name)
+              if (version != null) {
+                val approvedVersion = repository.latestVersionApprovedIn(deliveryConfig, artifact, environment.name)
+                approvedVersion != null
+                  ?: log.info("version: [$approvedVersion] of artifact ${artifact.name} is currently approved for environment ${environment.name}")
+              } else {
+                log.warn("No version of {} passes constraints for environment {}", artifact.name, environment.name)
+              }
             } else {
               if (!versionIsPending) {
                 approveVersion(deliveryConfig, artifact, version, environment.name)
@@ -165,6 +188,7 @@ class EnvironmentPromotionChecker(
     version: String,
     targetEnvironment: String
   ) {
+    log.info("version [$version] of ${artifact.type} artifact ${artifact.name} for environment $targetEnvironment is approved")
     val isNewVersion = repository
       .approveVersionFor(deliveryConfig, artifact, version, targetEnvironment)
     if (isNewVersion) {
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
@@ -188,7 +188,7 @@ class EnvironmentPromotionChecker(
     version: String,
     targetEnvironment: String
   ) {
-    log.info("version [$version] of ${artifact.type} artifact ${artifact.name} for environment $targetEnvironment is approved")
+    log.debug("Approving version [$version] of ${artifact.type} artifact ${artifact.name} for environment $targetEnvironment")
     val isNewVersion = repository
       .approveVersionFor(deliveryConfig, artifact, version, targetEnvironment)
     if (isNewVersion) {
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
@@ -98,13 +98,13 @@ class EnvironmentPromotionChecker(
                        * deployed to a required environment or outside of an allowed time.
                        */
 
-                      val passesStatelessConstraints: Boolean = checkStatelessConstraints(artifact, deliveryConfig, v, environment)
-                      val passesStatefulConstraints: Boolean = checkStatefulConstraints(artifact, deliveryConfig, v, environment)
+                      val passesStatelessConstraints = checkStatelessConstraints(artifact, deliveryConfig, v, environment)
+                      val passesStatefulConstraints = checkStatefulConstraints(artifact, deliveryConfig, v, environment)
 
-                      log.info("for version: [$v] of artifact ${artifact.name}: " +
-                        "passesStatelessConstraints: [$passesStatelessConstraints]" +
-                        ", passesStatefulConstraints [$passesStatefulConstraints] in" +
-                        "for environment ${environment.name}")
+                      log.debug("Version $v of artifact ${artifact.name}: " +
+                        "passes stateless constraints: $passesStatelessConstraints, " +
+                        "passes stateful constraints: $passesStatefulConstraints " +
+                        "in environment ${environment.name}")
 
                       val passesConstraints =
                         passesStatelessConstraints && passesStatefulConstraints
@@ -128,13 +128,13 @@ class EnvironmentPromotionChecker(
              */
             var approvedPending = false
             if (!hasPin) {
-              log.info("pendingVersionsToCheck: [$pendingVersionsToCheck] of artifact ${artifact.name} for environment ${environment.name} ")
+              log.debug("pendingVersionsToCheck: [$pendingVersionsToCheck] of artifact ${artifact.name} for environment ${environment.name} ")
               pendingVersionsToCheck
                 .sortedWith(artifact.versioningStrategy.comparator.reversed()) // oldest first
                 .forEach {
 
-                  val passesStatelessConstraints: Boolean = checkStatelessConstraints(artifact, deliveryConfig, it, environment)
-                  val passesStatefulConstraints: Boolean = checkStatefulConstraints(artifact, deliveryConfig, it, environment)
+                  val passesStatelessConstraints = checkStatelessConstraints(artifact, deliveryConfig, it, environment)
+                  val passesStatefulConstraints = checkStatefulConstraints(artifact, deliveryConfig, it, environment)
 
                   log.debug("Version $it of artifact ${artifact.name}: " +
                     "passes stateless constraints: $passesStatelessConstraints, " +
@@ -157,7 +157,8 @@ class EnvironmentPromotionChecker(
                    * We don't need to re-invoke stateful constraint evaluators for these, but we still
                    * check stateless constraints to avoid approval outside of allowed-times.
                    */
-                  log.info("version: [$v] of artifact ${artifact.name} is in queuedForApproval for environment ${environment.name}")
+                  log.debug("Version $v of artifact ${artifact.name} is in queued for approval," +
+                    "and being evaluated for stateless constraints in environment ${environment.name}")
                   if (checkStatelessConstraints(artifact, deliveryConfig, v, environment)) {
                     approveVersion(deliveryConfig, artifact, v, environment.name)
                     repository.deleteQueuedConstraintApproval(deliveryConfig.name, environment.name, v)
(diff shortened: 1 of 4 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -97,10 +97,17 @@ class EnvironmentPromotionChecker(
...
                        * deployed to a required environment or outside of an allowed time.
                        */
-                      val passesConstraints =
-                        checkStatelessConstraints(artifact, deliveryConfig, v, environment) &&
-                          checkStatefulConstraints(artifact, deliveryConfig, v, environment)
 
+                      val passesStatelessConstraints: Boolean = checkStatelessConstraints(artifact, deliveryConfig, v, environment)
+                      val passesStatefulConstraints: Boolean = checkStatefulConstraints(artifact, deliveryConfig, v, environment)
+
+                      log.info("for version: [$v] of artifact ${artifact.name}: " +
```
gal-yardeni, line 104: added some logging statements, if it's not helpful I can remove it.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -141,14 +157,21 @@ class EnvironmentPromotionChecker(
...
                    */
+                  log.info("version: [$v] of artifact ${artifact.name} is in queuedForApproval for environment ${environment.name}")
                   if (checkStatelessConstraints(artifact, deliveryConfig, v, environment)) {
                     approveVersion(deliveryConfig, artifact, v, environment.name)
                     repository.deleteQueuedConstraintApproval(deliveryConfig.name, environment.name, v)
                   }
                 }
             }
             if (!approvedPending && versionIsPending || version == null) {
-              log.warn("No version of {} passes constraints for environment {}", artifact.name, environment.name)
```
gal-yardeni, line 151: This warning was not so true. The condition here does not cover the case where an older version gets approved (lets sat version 1 is approved and version 2 is pending, so: `approvedPending` is false (since version 1 is not pending approval anymore), `versionIsPending` is true (for version 2 which is pending), and vesion==2).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -141,14 +157,21 @@ class EnvironmentPromotionChecker(
...
+              if (version != null) {
+                val approvedVersion = repository.latestVersionApprovedIn(deliveryConfig, artifact, environment.name)
+                approvedVersion != null
+                  ?: log.info("version: [$approvedVersion] of artifact ${artifact.name} is currently approved for environment ${environment.name}")
```
lorin, line 171: I'm confused by the `approvedVersion != null ?:` part of this expression.  My understanding is that the expression to the left of the `?:` operator has to be a nullable, but here it's a Boolean (`approvedVersion != null`). 

Are you trying to guard the log statement against `approvedVersion` being null?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -477,11 +477,12 @@ class SqlDeliveryConfigRepository(
               .onDuplicateKeyUpdate()
               .set(CURRENT_CONSTRAINT.CONSTRAINT_UID, MySQLDSL.values(CURRENT_CONSTRAINT.CONSTRAINT_UID))
               .execute()
+          }
 
+          sqlRetry.withRetry(WRITE) {
             val allStates = constraintStateFor(state.deliveryConfigName, state.environmentName, state.artifactVersion)
```
lorin, line 483: If I understand this correctly (and this is a subtle bug!), it looks like `constraintStateFor` executes a SQL query that is outside of the transaction. Is that right?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
gal-yardeni: Yes :) I mean, if its null, the log statement will be `version null of xxx..` which is kinda weird in my opinion

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
gal-yardeni: yes, after this change, the transaction is being closed at line 480, and line 482 and after is outside that transaction.

gal-yardeni:
> Wow, this is a subtle bug! I imagine it was not easy to diagnose the issue here!
> 
> I worry a little about losing the transactional nature of this operation by breaking it up. I think it would be cleaner if the entire query was inside of a transaction, rather than splitting it across two transactions.
> 
> If I understand things correctly, it looks like the problem is that the `constraintStateFor` method executes a SQL query outside of the transaction. To do it all inside of a single transaction, we'd need another version of the `constraintStateFor` method that can take a jooq transaction context as an argument.

Can you elaborate more about your concerns?  why do we need it to be in the same transaction? I believe it should be safe calling it outside the first two `insert` calls, but maybe I'm missing something!

lorin:
> Can you elaborate more about your concerns? why do we need it to be in the same transaction? I believe it should be safe calling it outside the first two `insert` calls, but maybe I'm missing something!

I don't have a specific failure mode in mind that I'm concerned about. Rather, I have a more general concern when a logical operation is split up across multiple transactions. In particular, I don't know if it's dangerous or not if the inserts into the `environment_artifact_constraint` and `current_constraint` succeed but the insertion into the `environment_artifact_queued_approval` fails. (e.g., if there's a transient error that leads the section transaction to fail). The reason I lean towards using a single transaction is that it means that I don't have to reason about whether or not it's ok if some of the writes succeed but others fail.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -97,10 +97,17 @@ class EnvironmentPromotionChecker(
...
-                        checkStatelessConstraints(artifact, deliveryConfig, v, environment) &&
-                          checkStatefulConstraints(artifact, deliveryConfig, v, environment)
 
+                      val passesStatelessConstraints: Boolean = checkStatelessConstraints(artifact, deliveryConfig, v, environment)
+                      val passesStatefulConstraints: Boolean = checkStatefulConstraints(artifact, deliveryConfig, v, environment)
+
+                      log.info("for version: [$v] of artifact ${artifact.name}: " +
+                        "passesStatelessConstraints: [$passesStatelessConstraints]" +
+                        ", passesStatefulConstraints [$passesStatefulConstraints] in" +
+                        "for environment ${environment.name}")
```
luispollo, line 107:
```suggestion
                        " environment ${environment.name}")
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -97,10 +97,17 @@ class EnvironmentPromotionChecker(
                        * want to request judgement or deploy a canary for artifacts that aren't
                        * deployed to a required environment or outside of an allowed time.
                        */
-                      val passesConstraints =
-                        checkStatelessConstraints(artifact, deliveryConfig, v, environment) &&
-                          checkStatefulConstraints(artifact, deliveryConfig, v, environment)
 
+                      val passesStatelessConstraints: Boolean = checkStatelessConstraints(artifact, deliveryConfig, v, environment)
```
luispollo, line 101: Type declaration here is unnecessary.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -121,12 +128,21 @@ class EnvironmentPromotionChecker(
              */
             var approvedPending = false
             if (!hasPin) {
+              log.info("pendingVersionsToCheck: [$pendingVersionsToCheck] of artifact ${artifact.name} for environment ${environment.name} ")
```
luispollo, line 131:
```suggestion
              log.debug("pendingVersionsToCheck: [$pendingVersionsToCheck] of artifact ${artifact.name} for environment ${environment.name} ")
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
luispollo: I think they're definitely helpful! We typically don't have enough logs! But I'd lower the severity to `DEBUG`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -121,12 +128,21 @@ class EnvironmentPromotionChecker(
...
                 .sortedWith(artifact.versioningStrategy.comparator.reversed()) // oldest first
                 .forEach {
-                  val passesConstraints: Boolean =
-                    checkStatelessConstraints(artifact, deliveryConfig, it, environment) &&
-                      checkStatefulConstraints(artifact, deliveryConfig, it, environment)
+
+                  val passesStatelessConstraints: Boolean = checkStatelessConstraints(artifact, deliveryConfig, it, environment)
+                  val passesStatefulConstraints: Boolean = checkStatefulConstraints(artifact, deliveryConfig, it, environment)
+
+                  log.info("for version: [$it] of artifact ${artifact.name}: " +
```
luispollo, line 139:
```suggestion
                  log.debug("for version: [$it] of artifact ${artifact.name}: " +
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -121,12 +128,21 @@ class EnvironmentPromotionChecker(
...
+                    "passesStatelessConstraints: [$passesStatelessConstraints]" +
+                    ", passesStatefulConstraints [$passesStatefulConstraints] in" +
+                    "for environment ${environment.name}")
```
luispollo, line 142:
```suggestion
                  log.debug("Version $it of artifact ${artifact.name}: " +
                    "passes stateless constraints: $passesStatelessConstraints, " +
                    "passes stateful constraints: $passesStatefulConstraints " +
                    "in environment ${environment.name}")
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
lorin: I don't think the code achieves that. The `?:` is really only for handling cases where there's a null. If you want the logic to fire only when it's non-null, I think `?.let` is a better choice:

```kotlin
approvedVersion?.let {approved -> log.info("version: [$approved] of artifact ${artifact.name} is currently approved for environment ${environment.name}")}
```

luispollo: I'm pretty sure you're only having to do this due to precedence. `?:` is applying to just the `null` keyword on line 170, not the `approvedVersion != null` expression (which is a boolean expression that would never evaluate to null).

You could rewrite this as:
```kotlin
approvedVersion?.also {
  log.info("version: [$approvedVersion] of artifact ${artifact.name} is currently approved for environment ${environment.name}")
}
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -141,14 +157,21 @@ class EnvironmentPromotionChecker(
                    * We don't need to re-invoke stateful constraint evaluators for these, but we still
                    * check stateless constraints to avoid approval outside of allowed-times.
                    */
+                  log.info("version: [$v] of artifact ${artifact.name} is in queuedForApproval for environment ${environment.name}")
```
luispollo, line 160: This is a bit confusing to me since the statement immediately above is `approveVersion(...)`, which leads me to believe the version is approved, and not queued for approval. Consider rephrasing.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
luispollo: You description here seems to match exactly what the `if` statement expresses. Unless there's an operator precedence issue, where you might rewrite it as:
```kotlin
if ((!approvedPending && versionIsPending) || version == null) {
```
In your example:
```kotlin
if ((!false && true) || false)
```
...which would evaluate to `true` and enter this block, right?

luispollo: Never mind. You mean you _do_ enter the block as expected, it's just the warning was in the wrong place. Makes sense.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -165,6 +188,7 @@ class EnvironmentPromotionChecker(
     version: String,
     targetEnvironment: String
   ) {
+    log.info("version [$version] of ${artifact.type} artifact ${artifact.name} for environment $targetEnvironment is approved")
```
luispollo, line 191:
```suggestion
    log.debug("Approving version [$version] of ${artifact.type} artifact ${artifact.name} for environment $targetEnvironment")
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: I would highly recommend adding a comment here to explain why this is done outside the transaction, as it would be easy to fall into the trap of assuming this was done by mistake and adding this back into the transaction down the road.

luispollo:
> Wow, this is a subtle bug! I imagine it was not easy to diagnose the issue here!
> 
> I worry a little about losing the transactional nature of this operation by breaking it up. I think it would be cleaner if the entire query was inside of a transaction, rather than splitting it across two transactions.
> 
> If I understand things correctly, it looks like the problem is that the `constraintStateFor` method executes a SQL query outside of the transaction. To do it all inside of a single transaction, we'd need another version of the `constraintStateFor` method that can take a jooq transaction context as an argument.

Nice investigation, @gal-yardeni!

Overall, I understand why the change you made would fix the problem, but I have the same concern as @lorin. I would prefer to see the `txn` context object passed into `constraintStateFor` and executing that `select` within that context, than to break up the transaction. Consider giving that a try and seeing if does the trick as well!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
robfletcher: Couldn't it just be rewritten as an `if` statement? ```if (approvedVersion == null) {
  log.info("version: [$approvedVersion] of artifact ${artifact.name} is currently approved for environment ${environment.name}")
}```

gal-yardeni: @robfletcher  It was written as an `if` statement and then I wanted to make it fancier ;) I will just add that `if` back.

gal-yardeni:
@lorin @luispollo thanks for your suggestions! I've refactored the SQL code to use the transaction. Let me know if you have any other feedback or if it is safe to merge it.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -776,11 +779,21 @@ class SqlDeliveryConfigRepository(
...
     artifactVersion: String
   ): List<ConstraintState> {
+    return constraintStateForWithTransaction(deliveryConfigName, environmentName, artifactVersion)
+  }
+
+  private fun constraintStateForWithTransaction(
+    deliveryConfigName: String,
+    environmentName: String,
+    artifactVersion: String,
+    txn: DSLContext? = null
```
lorin, line 789: I think you can just change this to:

```
ctx: DSLContext? = jooq
```

Then you can remove this line:

```
val ctx = txn ?: jooq
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -141,14 +157,23 @@ class EnvironmentPromotionChecker(
                    * We don't need to re-invoke stateful constraint evaluators for these, but we still
                    * check stateless constraints to avoid approval outside of allowed-times.
                    */
+                  log.debug("Version $v of artifact ${artifact.name} is in queued for approval," +
```
luispollo, line 160:
```suggestion
                  log.debug("Version $v of artifact ${artifact.name} is in queued for approval, " +
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -478,20 +478,23 @@ class SqlDeliveryConfigRepository(
...
-            val allStates = constraintStateFor(state.deliveryConfigName, state.environmentName, state.artifactVersion)
+            /**
+             * Passing the transaction here since [constraintStateForWithTransaction] is querying [ENVIRONMENT_ARTIFACT_CONSTRAINT]
+             * table, and we need to make sure the new state was persisted prior to checking all states for a given artifact version.
+             */
+            val allStates = constraintStateForWithTransaction(state.deliveryConfigName, state.environmentName, state.artifactVersion, txn)
             if (allStates.allPass && allStates.size >= environment.constraints.statefulCount) {
-              txn
-                .insertInto(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL)
+              jooq.insertInto(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL)
```
luispollo, line 487:
```suggestion
              txn.insertInto(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL)
```

Issue #1167, opened 2020-05-13
Deb status export: do something better


emjburns:
Right now we can't determine deb status. We should be able to based on the name of the artifact. Reach out to the rocket team to see if they have something we can use?

It's pretty bad to not export all the status info.

emjburns:
Also, we need to actually figure out how to do the base os (`bionic-classic` etc)

jonsie:
I exported artifacts today, and I had two artifacts - one for each region:

```
- name: fiat
  type: deb
  reference: fiat
  vmOptions:
    baseLabel: RELEASE
    baseOs: bionic
    regions:
    - us-east-1
    storeType: EBS
- name: fiat
  type: deb
  reference: fiat
  vmOptions:
    baseLabel: RELEASE
    baseOs: bionic
    regions:
    - us-west-2
    storeType: EBS
```

This should be consolidated to:

```
- name: fiat
  type: deb
  reference: fiat
  vmOptions:
    baseLabel: RELEASE
    baseOs: bionic
    regions:
    - us-east-1
    - us-west-2
    storeType: EBS
```

jonsie:
Another small issue to note, the `statuses` field was removed when I exported the deb artifact.  I'm not sure why, but I added back `statuses: ["RELEASE"]`.

emjburns:
note: we should also be able to get the base os from hacky netflix parsing of the description

```"description": "name=fiat, arch=x86_64, ancestor_name=bionic-classicbase-x86_64-202005072313-ebs, ancestor_id=ami-1111, ancestor_version=nflx-base-5.540.0-h1708.eeeeeee",```

emjburns:
Fixed, will be rolled out in the next day or two

PR #1192, opened 2020-05-18
fix(deb): export base-os correctly

emjburns:
Step #1 to fixing https://github.com/spinnaker/keel/issues/1167 is guessing the base os more correctly and throwing an error if we can't determine it

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -430,21 +430,27 @@ class ClusterHandler(
   }
 
   /**
-   * This function attempts to use fields on an image to guess what OS the image is.
-   * If not found, it will default to "bionic".
+   * This function attempts to use image description to guess what OS the image is.
+   * If not found, it will throw an error.
    */
   fun guessBaseOsFrom(image: ActiveServerGroupImage): String =
-    when {
-      image.name.signalsBionic() || image.imageLocation.signalsBionic() || image.description?.signalsBionic() ?: false -> "bionic"
-      image.name.signalsXenial() || image.imageLocation.signalsXenial() || image.description?.signalsXenial() ?: false -> "xenial"
-      else -> {
-        log.error("Unable to determine OS from image $image, defaulting to bionic...")
-        "bionic"
-      }
-    }
+    parseBaseOsFrom(image.description)
+      ?: throw ExportError("Unable to determine the base image from image description: $image")
 
-  private fun String.signalsBionic(): Boolean = contains("bionic")
-  private fun String.signalsXenial(): Boolean = contains("xenial")
+  /**
+   * Parses the base os from the start of the ancestor name of a description like:
+   * "name=fiat, arch=x86_64, ancestor_name=bionic-classicbase-x86_64-202005072313-ebs,
+   *   ancestor_id=ami-1111, ancestor_version=nflx-base-5.540.0-h1708.eeeeeee"
+   */
+  private fun parseBaseOsFrom(description: String?): String? {
+    if (description == null) {
+      return null
+    }
+    val info: List<String> = description.split(",").map { it.trim() }
+    val ancestor = info.firstOrNull { it.startsWith("ancestor_name") } ?: return null
+    val ancestorValue = ancestor.split("=")[1]
+    return ancestorValue.substringBefore("base-x")
+  }
 
   override suspend fun actuationInProgress(resource: Resource<ClusterSpec>) =
     resource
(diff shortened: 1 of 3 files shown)


luispollo:
> Step #1

Haha -- loved this voyage through the tunnel of time. 😄

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -430,21 +430,27 @@ class ClusterHandler(
...
+   * "name=fiat, arch=x86_64, ancestor_name=bionic-classicbase-x86_64-202005072313-ebs,
+   *   ancestor_id=ami-1111, ancestor_version=nflx-base-5.540.0-h1708.eeeeeee"
+   */
+  private fun parseBaseOsFrom(description: String?): String? {
+    if (description == null) {
+      return null
+    }
+    val info: List<String> = description.split(",").map { it.trim() }
+    val ancestor = info.firstOrNull { it.startsWith("ancestor_name") } ?: return null
+    val ancestorValue = ancestor.split("=")[1]
```
luispollo, line 451: Maybe check the array size here just in case the format changes for whatever reason?

gal-yardeni, line 451: which `=` are you referring to in `ancestorValue`? where does this description string is coming from?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -430,21 +430,27 @@ class ClusterHandler(
...
+    return ancestorValue.substringBefore("base-x")
```
luispollo, line 452: This would return `bionic-classic` for your example in the comment, but the code above seemed to look only for `bionic` or `xenial`. Could this be rewritten as `ancestorValue.substringBefore("-")`, i.e. get the substring before the first dash, which in this case would be `bionic`?

luispollo: P.S. For my education, is there a different type of `bionic` other than `classic`?

lorin: @luispollo There are two bionic base images used inside of Netflix:
* `bionic-classic`
* `bionic`

The `bionic` one is a minimal bionic build.

lorin: @luispollo I assume that `bionic-classic` is what this method should return if `ancestor_name` looks something like `bionic-classicbase-x86_64-202005072313-ebs`, given the updated test assertion in ClusterExportTests. 

@emjburns can you confirm?

emjburns: we actually need to output `bionic-classic` here. That's the correct output

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
emjburns: Description is set by our bakery to always be something like the example. I split on commas first to separate each key=val pair, then find the `ancestor_name=bionic-classicbase-x86_64-202005072313-ebs` thing, then split _that_ by `=` to find the value of the ancestor. It's so exciting.

emjburns: tbh I don't anticipate a change here, because this is a pretty core thing to the bakery and we own the bakery and will not update it.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: Understood! Thanks for clarifying.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: 🤷‍♂️

gal-yardeni: ohhh I see, fun indeed.

emjburns:
@luispollo can you find an example of an app running on minimal bionic? I don't have an example payload, so I can't write the test

luispollo:
> @luispollo can you find an example of an app running on minimal bionic? I don't have an example payload, so I can't write the test

Actually, I just went to look on the UI and found out that it's called `bionic-minimal`, i.e. it has a dash just like `bionic-classic`, and so wouldn't add any value. Never mind.

PR #1193, opened 2020-05-19
fix(export): we now look at deb status in a hacky way

emjburns:
Part # 2 to https://github.com/spinnaker/keel/issues/1167

We now use a class adapted from the rocket team to give ourselves a hacky guess at what the debian status is. This is fine for now. 

Note: if you have a flow where you are deploying snapshot artifacts to feature branch clusters, you'll still need to edit your delivery config by hand. This change will just actually capture that you have two different statuses.

### DIFF
--- a/keel-core/src/main/java/com/netflix/rocket/api/artifact/internal/debian/DebianArtifactParser.java
+++ b/keel-core/src/main/java/com/netflix/rocket/api/artifact/internal/debian/DebianArtifactParser.java
@@ -1,0 +1,57 @@
+package com.netflix.rocket.api.artifact.internal.debian;
+
+import static com.netflix.spinnaker.keel.api.artifacts.ArtifactStatus.RELEASE;
+import static org.apache.commons.lang3.StringUtils.substringAfterLast;
+import static org.apache.commons.lang3.StringUtils.substringBefore;
+import static org.apache.commons.lang3.StringUtils.substringBetween;
+
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactStatus;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+
+public class DebianArtifactParser {
+  private static final Logger log = org.slf4j.LoggerFactory.getLogger(DebianArtifactParser.class);
+
+  public ArtifactStatus parseStatus(String raw) {
+    String version = parseVersion(raw);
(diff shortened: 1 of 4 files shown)


keel-core/src/main/java/com/netflix/rocket/api/artifact/internal/debian/DebianArtifactParser.java
```diff
@@ -0,0 +1,57 @@
+package com.netflix.rocket.api.artifact.internal.debian;
```
luispollo, line 1: Any particular reason to name this `rocket` and not `spinnaker.keel`?

keel-core/src/test/java/com/netflix/rocket/api/artifact/internal/debian/DebianArtifactParserTests.java
```diff
@@ -0,0 +1,130 @@
...
+    String rawVersion =
+        "debian-local:pool/o/mypackage-server/mypackage-server_1.100.5~rc.23-LOCAL_all.deb";
+
+    assertThat(parser.parseStatus(rawVersion)).isEqualTo(ArtifactStatus.CANDIDATE);
+  }
+
+  @Test
+  void shouldParseSnapshotVersionString() {
+    String rawVersion =
+        "debian-local:pool/o/mypackage-server/mypackage-server_0.0.1176~dev.34267+2345af-h1886.4246bc3_all.deb";
```
luispollo, line 53: Does the string always have the `~` before `deb`? I noticed it's not in the regex like it is for `~rc`.

keel-core/src/main/java/com/netflix/rocket/api/artifact/internal/debian/DebianArtifactParser.java
emjburns: It's a direct lift from their code, same as the DebianComparator. Just followed that pattern

keel-core/src/test/java/com/netflix/rocket/api/artifact/internal/debian/DebianArtifactParserTests.java
emjburns: I'm not sure, but I'm pretty sure this code is correct

PR #1195, opened 2020-05-19
chore(logs): pipeline logs + trigger

emjburns:
Adding a "start pipeline" log and putting the env name in the link text (as `Env {env}`). Any better link text ideas? Trying to find something short but informative

### DIFF
--- a/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluator.kt
+++ b/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluator.kt
@@ -155,12 +155,14 @@ class PipelineConstraintEvaluator(
     constraint: PipelineConstraint,
     serviceAccount: String
   ): String {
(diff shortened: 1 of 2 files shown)


keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluator.kt
```diff
@@ -155,12 +155,14 @@ class PipelineConstraintEvaluator(
...
     serviceAccount: String
   ): String {
+    log.info("Triggering pipeline ${constraint.pipelineId} for environment ${environment.name}")
 
     // using java.util.HashMap over kotlin.collections for retrofit2 compatibility
     val trigger = HashMap<String, Any>()
     trigger["type"] = "managed"
     trigger["user"] = "keel"
     trigger["parameters"] = constraint.parameters
+    trigger["linkText"] = "Env ${environment.name}"
```
luispollo, line 165: What about the link URL? Where does that come from?

PR #1197, opened 2020-05-19
fix(titus): export container resources

emjburns:
Fixing titus export where we don't correctly export the container resource section.

### DIFF
--- a/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
+++ b/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
@@ -540,16 +540,19 @@ class TitusClusterHandler(
       containerAttributes = containerAttributes,
       iamProfile = iamProfile,
       migrationPolicy = migrationPolicy,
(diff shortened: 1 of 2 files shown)


keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -540,16 +540,19 @@ class TitusClusterHandler(
...
       migrationPolicy = migrationPolicy,
-      resources = resources.exportSpec(),
+      resources = resources.toSpec(),
       tags = tags
     )
 
     return checkNotNull(buildSpecFromDiff(defaults, thisSpec))
   }
 
-  private fun Resources.exportSpec(): ResourcesSpec? {
```
gal-yardeni, line 550: can you explain what was wrong in the old implementation? did we not capture the right data?

emjburns: Well, it seemed like we were trying to only export the difference between the default for the resources field. but that didn't work. So just made it output the whole resources section there. TBH I don't understand how to debug the recursive diffing.

PR #1199, opened 2020-05-19
fix(clouddriver) LB Action: targetGroupName optional

### LINKED ISSUE #1198: Error exporting load balancer: missing targetGroupName
Keel assumes that a load balancer `Action` field will always have a `targetGroupName`, but that’s not true for redirect actions. Here's an example of a redirect action that doesn't have a `targetGroupName`.

```
$ jq  '.[0].listeners[0].defaultActions' load-balancer.json

[
  {
    "type": "redirect",
    "order": 1,
    "redirectConfig": {
      "protocol": "HTTPS",
      "port": "443",
      "host": "#{host}",
      "path": "/#{path}",
      "query": "#{query}",
      "statusCode": "HTTP_301"
    }
  }
]
```

This results in the following error:

```
MissingKotlinParameterException: Instantiation of [simple type, class com.netflix.spinnaker.keel.clouddriver.model.ApplicationLoadBalancerModel$Action] value failed for JSON property targetGroupName due to missing (therefore NULL) value for creator parameter targetGroupName which is a non-nullable type
 at [Source: (okhttp3.ResponseBody$BomAwareReader); line: 1, column: 526] (through reference chain: java.util.ArrayList[0]->com.netflix.spinnaker.keel.clouddriver.model.ApplicationLoadBalancerModel["listeners"]->java.util.ArrayList[0]->com.netflix.spinnaker.keel.clouddriver.model.ApplicationLoadBalancerModel$ApplicationLoadBalancerListener["defaultActions"]->java.util.ArrayList[0]->com.netflix.spinnaker.keel.clouddriver.model.ApplicationLoadBalancerModel$Action["targetGroupName"])
```

lorin:
Make targetGroupName optional in the Action model.

Keel assumed that a load balancer Action field will always have a targetGroupName, but that’s not true for redirect actions.

fixes #1198.

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ApplicationLoadBalancerModel.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ApplicationLoadBalancerModel.kt
@@ -34,7 +34,7 @@ data class ApplicationLoadBalancerModel(
   data class Action(
     val type: String,
     val order: Int,
-    val targetGroupName: String,
+    val targetGroupName: String?,
     val redirectConfig: RedirectConfig?
   )
 
--- a/keel-clouddriver/src/test/resources/loadBalancers.json
+++ b/keel-clouddriver/src/test/resources/loadBalancers.json
@@ -308,6 +308,18 @@
               }
             },
             "targetGroupName": "keeldemo-prestaging"
+          },
+          {
+            "type": "redirect",
+            "order": 1,
+            "redirectConfig": {
+              "protocol": "HTTPS",
+              "port": "443",
+              "host": "#{host}",
+              "path": "/#{path}",
+              "query": "#{query}",
+              "statusCode": "HTTP_301"
+            }
           }
         ],
         "loadBalancerName": "keeldemo-prestaging-alb",


keel-clouddriver/src/test/resources/loadBalancers.json
```diff
@@ -308,6 +308,18 @@
               }
             },
             "targetGroupName": "keeldemo-prestaging"
+          },
+          {
```
lorin, line 312: Updating this test fixture reproduces the problem.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ApplicationLoadBalancerModel.kt
```diff
@@ -34,7 +34,7 @@ data class ApplicationLoadBalancerModel(
   data class Action(
     val type: String,
     val order: Int,
-    val targetGroupName: String,
+    val targetGroupName: String?,
```
gal-yardeni, line 37: Is it possible to default it to something? or it is completely meanless?

lorin: My instinct is to have these models be as faithful as possible to the underlying JSON representation, because it makes it easier to figure out how the data was populated if something goes wrong.

gal-yardeni: makes sense.

PR #1201, opened 2020-05-19
fix(constraints): don't evaluate stateful unless stateless pass

emjburns:
Remove logging (sorry @gal-yardeni, I didn't want to take the time to ensure it would log but also not break the ordering), and revert back to only checking stateful if stateless pass.


keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionCheckerTests.kt
```diff
@@ -277,7 +277,7 @@ internal class EnvironmentPromotionCheckerTests : JUnit5Minutests {
           /**
            * Verify that stateful constraints are not checked if a stateless constraint blocks promotion
            */
-          verify(exactly = 1) {
+          verify(inverse = true) {
```
gal-yardeni, line 280: ohhhhh now I understand why it happened. thanks for fixing it!

emjburns: of course!

PR #1202, opened 2020-05-20

spinnakerbot:


lorin:
We're getting Kotlin compile errors:

```
> Task :keel-ec2-plugin:compileTestKotlin FAILED
e: org.jetbrains.kotlin.codegen.CompilationException: Back-end (JVM) Internal error: wrong bytecode generated
```

My suspicion is that it's related to https://github.com/spinnaker/kork/pull/655  cc @robfletcher

robzienert:
FWIW, this is the exact error that will occur with Keel plugins whenever the Kotlin version is incremented within Keel, but not updated for each plugin.

PR #1203, opened 2020-05-20
fix(clouddriver): healthCheckPort as string

### LINKED ISSUE #1200: Error deserializing ALB target group
Keel assumes the `healthCheckPort` field in an ApplicationLoadBalancer model object is an integer, but it turns out that `"traffic-port"` is also a valid value.

This results in the following error deserialization an application load balancer (ALB) object from clouddriver:

```
InvalidFormatException: Cannot deserialize value of
type `int` from String "traffic-port": not a valid Integer value
(through reference chain: java.util.ArrayList[0]→com.netflix.spinnaker.keel.clouddriver
.model.ApplicationLoadBalancerModel["targetGroups"]→java.util.ArrayList[0]→
com.netflix.spinnaker.keel.clouddriver.model.ApplicationLoadBalancerModel
$TargetGroup["healthCheckPort"])
```

(Error message lightly edited for clarity)

lorin:
## Summary

Change the `healthCheckPort` field on a clouddriver data model object from `Int` to `String`.

Fixes #1200 

## Problem

When we query clouddriver for an ApplicationLoadBalancer (ALB) model object, one of the model sub-objects, TargetGroup, has a field named `healthCheckPort`.

This field is encoded as a string in the JSON payload, but is almost always a string-quoted integer (e.g., `"8080"`). However, we recently discovered that it is not always an integer: a valid value is the string `"traffic-port"`, which means "just use the same port that the load balancer sends the regular traffic to."

Keel's `TargetGroup` model object currently represents this value as an `Int`, which leads to a deserialization error when an ALB is configured with `"traffic-port"`.

## Solution requirements

To fix this, we need to modify the `TargetGroup` model so that it can deserialize ALB models that have health check port set to the literal string "traffic-port", as well as string-encoded port numbers (e.g., `"8080"`).

## Implemented solution

I changed the field on the ApplicationLoadBalancer model object from `Int` to `String`, and I modified the `CloudDriverService.getApplicationLoadBalancer` method in ApplicationLoadBalancerHandler.kt to convert `"traffic-port"` to the numerical value of the traffic port.

I did **not** modify the corresponding field on the ApplicationLoadBalancerSpec.TargetGroup object, I left it as an `Int`

**Rationale**: I thought that preserving the current data model of the spec object would be less risky: Since the spec objects get persisted to the database, making changes to them carries a risk of the code and data going out of sync.

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ApplicationLoadBalancerModel.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ApplicationLoadBalancerModel.kt
@@ -68,7 +68,7 @@ data class ApplicationLoadBalancerModel(
     val port: Int,
     val healthCheckEnabled: Boolean,
     val healthCheckTimeoutSeconds: Int,
-    val healthCheckPort: Int,
+    val healthCheckPort: String, // quoted number (e.g., "8080") or "traffic-port"
     val healthCheckProtocol: String,
     val healthCheckPath: String,
     val healthCheckIntervalSeconds: Int,
--- a/keel-clouddriver/src/test/resources/loadBalancers.json
+++ b/keel-clouddriver/src/test/resources/loadBalancers.json
@@ -1039,7 +1039,7 @@
         ],
         "targetGroupArn": "arn:aws:elasticloadbalancing:us-east-1:999999999999:targetgroup/keeldemo-test/abababababababab",
         "healthCheckTimeoutSeconds": 5,
-        "healthCheckPort": "8080",
+        "healthCheckPort": "traffic-port",
         "targetType": "instance",
         "matcher": {
           "httpCode": "200-299"
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
@@ -246,7 +246,10 @@ class ApplicationLoadBalancerHandler(
                       port = tg.port,
                       healthCheckEnabled = tg.healthCheckEnabled,
                       healthCheckTimeoutSeconds = Duration.ofSeconds(tg.healthCheckTimeoutSeconds.toLong()),
-                      healthCheckPort = tg.healthCheckPort,
+                      healthCheckPort = when (tg.healthCheckPort) {
+                        "traffic-port" -> tg.port
+                        else -> tg.healthCheckPort.toInt()
+                      },
                       healthCheckProtocol = tg.healthCheckProtocol,
                       healthCheckHttpCode = tg.matcher.httpCode,
                       healthCheckPath = tg.healthCheckPath,
--- a/keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandlerTests.kt
+++ b/keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandlerTests.kt
@@ -1,9 +1,8 @@
 package com.netflix.spinnaker.keel.ec2.resource
 
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.module.kotlin.registerKotlinModule
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.Exportable
+import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancer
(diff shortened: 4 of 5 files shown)


keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandlerTests.kt
```diff
@@ -65,7 +66,6 @@ internal class ApplicationLoadBalancerHandlerTests : JUnit5Minutests {
     repository,
     publisher
   )
-  private val mapper = ObjectMapper().registerKotlinModule()
```
lorin, line 68: We weren't using `mapper` anywhere in this test so I assumed it was safe to delete.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandlerTests.kt
```diff
@@ -106,7 +106,7 @@ internal class ApplicationLoadBalancerHandlerTests : JUnit5Minutests {
   private val sub2 = Subnet("subnet-1", vpc.id, vpc.account, vpc.region, "${vpc.region}d", "internal (vpc0)")
   private val sg1 = SecurityGroupSummary("testapp-elb", "sg-55555", "vpc-1")
 
-  private val model = ApplicationLoadBalancerModel(
+  private fun model(port: Int = 7001, healthCheckPort: String = "7001") = ApplicationLoadBalancerModel(
```
lorin, line 109: Changed this from a variable to a function to make it easier to test with different port parameters.

keel-clouddriver/src/test/resources/loadBalancers.json
```diff
@@ -1039,7 +1039,7 @@
         ],
         "targetGroupArn": "arn:aws:elasticloadbalancing:us-east-1:999999999999:targetgroup/keeldemo-test/abababababababab",
         "healthCheckTimeoutSeconds": 5,
-        "healthCheckPort": "8080",
+        "healthCheckPort": "traffic-port",
```
lorin, line 1042: I changed this fixture to add an additional deserialization test  case.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ApplicationLoadBalancerModel.kt
```diff
@@ -68,7 +68,7 @@ data class ApplicationLoadBalancerModel(
     val port: Int,
     val healthCheckEnabled: Boolean,
     val healthCheckTimeoutSeconds: Int,
-    val healthCheckPort: Int,
+    val healthCheckPort: String, // quoted number (e.g., "8080") or "traffic-port"
```
gal-yardeni, line 71: is `traffic-port` the only string that can appear there?

lorin: I think so, based on [similar logic inside of clouddriver](https://github.com/spinnaker/clouddriver/blob/7c861e3b1c47db1e103df37dc410930b36aa61fb/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerV2AtomicOperation.groovy#L160).

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
```diff
@@ -246,7 +246,10 @@ class ApplicationLoadBalancerHandler(
                       port = tg.port,
                       healthCheckEnabled = tg.healthCheckEnabled,
                       healthCheckTimeoutSeconds = Duration.ofSeconds(tg.healthCheckTimeoutSeconds.toLong()),
-                      healthCheckPort = tg.healthCheckPort,
+                      healthCheckPort = when (tg.healthCheckPort) {
+                        "traffic-port" -> tg.port
+                        else -> tg.healthCheckPort.toInt()
```
gal-yardeni, line 251: can this might cause a class cast exception? (if it cannot be converted to int for some reason)

lorin: Yeah, If clouddriver ever returned a string other than an integer or `"traffic-port"`, it would throw an exception that looks like this:

```
java.lang.NumberFormatException: For input string: "unexpected-string-here"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.parseInt(Integer.java:615)
...
```

Since we never expect clouddriver to pass us a non-integer value, I think throwing an exception is the appropriate behavior here.

Issue #1021, opened 2020-04-14
Error handling: having more precise exceptions when validation fails


gal-yardeni:
We want to have a better user experience when submitting/validating a delivery config. Each endpoint is validating the config in a different way. We need to align those.
1. delivery-config/diff
2. delivery-config/validate
3. delivery-config

PR #1215, opened 2020-05-21
Chore(validations): Aligning config validations across multiple endpoints

gal-yardeni:
Closes https://github.com/spinnaker/keel/issues/1021.

This is a stab at aligning delivery config validations across endpoints:
POST /delivery-configs (upsert a config)
POST /delivery-configs/diff (diffing a new config to an existing one)
POST /delivery-configs/validate (static validation of config)

My logic was to extract the validation logic to a `DeliveryConfigValidator` class, which will be a centralized class for the different validations. To make it work with all the endpoints above, I've transformed the logic to apply on the `SubmittedDeliveryConfig` object rather than `DeliveryConfig` and wired the validator to the controller class rather than each individual implementation.

Next:
- [ ] Add more validations
- [ ] Clean up code

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
@@ -7,17 +7,11 @@ import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
 import com.netflix.spinnaker.keel.api.artifacts.DockerArtifact
 import com.netflix.spinnaker.keel.api.artifacts.TagVersionStrategy.BRANCH_JOB_COMMIT_BY_JOB
 import com.netflix.spinnaker.keel.api.id
-import com.netflix.spinnaker.keel.core.api.DependsOnConstraint
-import com.netflix.spinnaker.keel.core.api.SubmittedDeliveryConfig
-import com.netflix.spinnaker.keel.core.api.SubmittedEnvironment
 import com.netflix.spinnaker.keel.core.api.SubmittedResource
 import com.netflix.spinnaker.keel.core.api.normalize
 import com.netflix.spinnaker.keel.events.ArtifactRegisteredEvent
 import com.netflix.spinnaker.keel.events.ResourceCreated
 import com.netflix.spinnaker.keel.events.ResourceUpdated
-import com.netflix.spinnaker.keel.exceptions.DuplicateArtifactReferenceException
-import com.netflix.spinnaker.keel.exceptions.DuplicateResourceIdException
-import com.netflix.spinnaker.keel.exceptions.MissingEnvironmentReferenceException
 import com.netflix.spinnaker.keel.resources.ResourceSpecIdentifier
 import com.netflix.spinnaker.keel.test.DummyResourceSpec
 import com.netflix.spinnaker.keel.test.TEST_API_V1
@@ -38,7 +32,6 @@ import strikt.assertions.hasSize
 import strikt.assertions.isA
 import strikt.assertions.isEmpty
 import strikt.assertions.isEqualTo
-import strikt.assertions.isFailure
 import strikt.assertions.isSuccess
 import strikt.assertions.isTrue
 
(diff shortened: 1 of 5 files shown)


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -300,148 +293,6 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
     }
 
     context("persisting delivery config manifests") {
-
-      context("a delivery config with non-unique resource ids errors while persisting") {
```
gal-yardeni, line 304: Those tests have moved to `DeliveryConfigValidatorTest` class

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidator.kt
```diff
@@ -0,0 +1,84 @@
...
+ * - depends on environments are unique
+ *
+ * Throws an exception if config fails any checks
+ */
+@Component
+class DeliveryConfigValidator {
+
+  val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  fun validate(config: SubmittedDeliveryConfig) {
```
gal-yardeni, line 25: Changed the object to be `SubmittedDeliveryConfig`, all the logic remained the same.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidatorTests.kt
```diff
@@ -0,0 +1,164 @@
...
+import com.netflix.spinnaker.keel.exceptions.MissingEnvironmentReferenceException
+import com.netflix.spinnaker.keel.test.DummyResourceSpec
+import com.netflix.spinnaker.keel.test.TEST_API_V1
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import strikt.api.expectCatching
+import strikt.assertions.isA
+import strikt.assertions.isFailure
+
+internal class DeliveryConfigValidatorTests : JUnit5Minutests {
```
gal-yardeni, line 19: I wasn't sure if those tests should be at the controller level, or here, as here I can't verify that the config wasn't persisted (...this logic should prevent "bad" configs to even reach `upsert` function)

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidatorTests.kt
```diff
@@ -0,0 +1,164 @@
...
+                spec = DummyResourceSpec(data = "o hai")
+              )
+            ),
+            constraints = emptySet()
+          )
+        )
+      )
+
+      val submittedConfig2 = submittedConfig1.copy(name = "double-trouble")
+//      test("an error is thrown and config is not persisted") {
```
gal-yardeni, line 125: This is a scenario that cannot happen anymore -- since we are returning bad request even before reaching `upsert`, we cannot return `TooManyDeliveryConfigsException`.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -75,19 +79,23 @@ class DeliveryConfigController(
...
   )
   @PreAuthorize("@authorizationSupport.hasApplicationPermission('READ', 'APPLICATION', #deliveryConfig.application)")
-  fun validate(@RequestBody deliveryConfig: SubmittedDeliveryConfig) =
-  // TODO: replace with JSON schema/OpenAPI spec validation when ready (for now, leveraging parsing error handling
+  fun validate(@RequestBody deliveryConfig: SubmittedDeliveryConfig) {
+    // TODO: replace with JSON schema/OpenAPI spec validation when ready (for now, leveraging parsing error handling
     //  in [ExceptionHandler])
-    mapOf("status" to "valid")
+    validator.validate(deliveryConfig)
+    // mapOf("status" to "valid")
```
gal-yardeni, line 97: I think we can remove `mapOf("status" to "valid")` here and just return 200 if its valid?

lorin: I think [204](https://httpstatuses.com/204) would be a more appropriate status code if we aren't returning a body, but, otherwise, I agree we don't need to return anything in the body.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidatorTests.kt
lorin: I would leave these tests here, and have separate tests for the controller that mock out the validator, so you can verify that the controller doesn't call `upsert` if validation fails, but the controller tests don't have to deal with all of the different invalid cases.

(I would also remove the "while persisting" language from the contexts below, since you're just testing the behavior of the validator itself in these tests).

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
luispollo: Yeah, the trick is matching that to the return type of the function... Maybe if you make it nullable it just works?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidator.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.springframework.stereotype.Component
+
+/**
+ * Run validation checks against delivery config to ensure:
+ *
+ * - resources have unique ids
+ * - artifacts have unique references
+ * - depends on environments are unique
+ *
+ * Throws an exception if config fails any checks
```
luispollo, line 18: I think these comments belong in the function, not the class, especially since you mention the "throws" behavior. Here you could just say "Provides delivery config validation functions" or something like that.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidator.kt
```diff
@@ -0,0 +1,84 @@
...
+      ids.groupingBy { it }
+        .eachCount()
+        .filter { it.value > 1 }
+        .keys
+        .toList()
+
+    /**
+     * check: resources have unique ids
+     */
+
```
luispollo, line 38:
```suggestion
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidatorTests.kt
luispollo: Maybe just replace "errors while persisting" with "fails validation"?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidatorTests.kt
luispollo: Agreed. Just remove the test.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -75,19 +79,23 @@ class DeliveryConfigController(
...
     path = ["/validate"],
     consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
     produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
   )
   @PreAuthorize("@authorizationSupport.hasApplicationPermission('READ', 'APPLICATION', #deliveryConfig.application)")
-  fun validate(@RequestBody deliveryConfig: SubmittedDeliveryConfig) =
-  // TODO: replace with JSON schema/OpenAPI spec validation when ready (for now, leveraging parsing error handling
+  fun validate(@RequestBody deliveryConfig: SubmittedDeliveryConfig) {
+    // TODO: replace with JSON schema/OpenAPI spec validation when ready (for now, leveraging parsing error handling
     //  in [ExceptionHandler])
```
luispollo, line 95: Honestly, I think we can remove this `TODO` at this point. If we eventually use the OpenAPI spec here, that's fine, but this is our current validation framework and it's fine.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
luispollo: Never mind, I see you changed the return type of the function. Sure, go ahead and remove the map and change the status code to 204. 👍

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidator.kt
gal-yardeni: you are right!

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
gal-yardeni: I agree, removing those.

PR #1216, opened 2020-05-21
make event.uid a primary key

robfletcher:


keel-sql/src/main/resources/db/databaseChangeLog.yml
```diff
@@ -143,3 +143,6 @@ databaseChangeLog:
   - include:
       file: changelog/20200521-event-uid-timestamp-index.yml
       relativeToChangelogFile: true
+  - include:
+      file: changelog/20200521-add-event-pk.yml
```
lorin, line 147: Is it OK that `20200521-add-event-pk.yml` precedes `20200521-event-uid-timestamp-index.yml`in lexicographical order? I recall this biting us before.

robfletcher: I don't _think_ they should conflict (and the latter is rolled out in all our environments already)

PR #1220, opened 2020-05-21
Fixed Autobump korkVersion

robfletcher:


keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt
```diff
@@ -383,15 +382,15 @@ internal class SecurityGroupHandlerTests : JUnit5Minutests {
       setupVpc()
     }
 
-    sequenceOf(
+    mapOf(
       "create" to SecurityGroupHandler::create,
       "update" to SecurityGroupHandler::update
     )
-      .forEach { (methodName, handlerMethod) ->
+      .forEach { methodName, handlerMethod ->
```
robfletcher, line 389: Seems like there was a problem using the inline version of `forEach` when one of the parameters is a suspend function.

luispollo: Good find!

lorin: How did you ever figure this out???

PR #1223, opened 2020-05-21
fix(clouddriver): fetch multiple server groups

lorin:
Add support for being able to fetch multiple ASGs, so that we can ultimately handle the case where there are multiple active ASGs.

Initial work towards #1064

Notes: 

* this just adds a method for retrieving the ASGs from clouddriver. No code is calling this method yet.
* haven't added support for Titus yet, that's coming in a future PR.


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/PassThruController.kt
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * This controller passes through calls to downstream services (e.g., clouddriver).
+ *
+ * It's used only for testing.
+ */
+@RestController
+@ConditionalOnProperty("tests.passthru", matchIfMissing = false)
+@RequestMapping(path=["/test"])
+class PassThruController(
```
lorin, line 22: I wrote this controller to help me test this new logic, and figured it might be useful to have it in the codebase, but I can zap it if others feel differently.

robfletcher: I feel like we can zap it when we start using the new call.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -7,6 +7,57 @@ import com.netflix.spinnaker.keel.clouddriver.CloudDriverCache
 import com.netflix.spinnaker.keel.core.api.Capacity
 import com.netflix.spinnaker.kork.exceptions.SystemException
 
+data class ServerGroupCollection(
+  val accountName: String,
+  val serverGroups: Set<ServerGroup>
+)
+
+/**
+ * Objects that are returned when querying for all fo the server groups associated with a cluster.
```
luispollo, line 16:
```suggestion
 * Objects that are returned when querying for all of the server groups associated with a cluster.
```

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -7,6 +7,57 @@ import com.netflix.spinnaker.keel.clouddriver.CloudDriverCache
...
+ *
+ * Two differences from [ActiveServerGroup]:
+ *   - disabled flag
+ *   - no accountName field (since this is defined on the parent [ServerGroupCollection] object
```
luispollo, line 20: I feel like there's an opportunity here to take advantage of that `disabled` flag and merge those two types into one.

lorin: @luispollo What I'm struggling with for unifying them is that I want to have a ServerGroup class that has a non-nullable `accountName` field.

PR #1224, opened 2020-05-22
refactor(db): standardize timestamp types in all tables

robfletcher:
We've used a mixture of `bigint` and `timestamp` on our tables. This PR standardizes on the latter. It also ensures we have millisecond precision (which is probably only really important on `event` but makes sense as a standard for compatibility with `java.time` types).

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/timestamps.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/timestamps.kt
@@ -1,0 +1,9 @@
+package com.netflix.spinnaker.keel.sql
+
+import java.time.Clock
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneOffset.UTC
+
+fun Instant.toTimestamp(): LocalDateTime = atZone(UTC).toLocalDateTime()
+fun Clock.timestamp(): LocalDateTime = instant().toTimestamp()
--- a/keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
+++ b/keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
@@ -1,0 +1,48 @@
+databaseChangeLog:
+- changeSet:
+    id: event-uid-timestamp-index
+    author: fletch
+    changes:
+    - sql:
+        sql: |
+          alter table agent_lock change column expiry old bigint(13);
+          alter table agent_lock add column expiry timestamp(3) not null;
+          update agent_lock set expiry = from_unixtime(old / 1000);
+          alter table agent_lock drop column old;
+
+          alter table diff_fingerprint change column first_detection_time old bigint(13);
+          alter table diff_fingerprint add column first_detection_time timestamp(3) not null;
+          update diff_fingerprint set first_detection_time = from_unixtime(old / 1000);
+          alter table diff_fingerprint drop column old;
+
+          alter table environment_artifact_pin change column pinned_at old bigint(13);
+          alter table environment_artifact_pin add column pinned_at timestamp(3) not null;
+          update environment_artifact_pin set pinned_at = from_unixtime(old / 1000);
+          alter table environment_artifact_pin drop column old;
+
+          alter table environment_artifact_queued_approval change column queued_at old bigint(13);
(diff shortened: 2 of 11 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/timestamps.kt
```diff
@@ -0,0 +1,9 @@
+package com.netflix.spinnaker.keel.sql
+
+import java.time.Clock
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneOffset.UTC
+
+fun Instant.toTimestamp(): LocalDateTime = atZone(UTC).toLocalDateTime()
+fun Clock.timestamp(): LocalDateTime = instant().toTimestamp()
```
robfletcher, line 9: I created these standard functions that we can use across our SQL code

keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
```diff
@@ -0,0 +1,23 @@
...
+          alter table environment_artifact_pin modify pinned_at timestamp(3) not null;
+          alter table environment_artifact_queued_approval modify queued_at timestamp(3) not null;
+          alter table environment_artifact_versions modify approved_at timestamp(3) null;
+          alter table environment_artifact_versions modify deployed_at timestamp(3) null;
+          alter table environment_artifact_versions modify replaced_at timestamp(3) null;
+          alter table event modify timestamp timestamp(3) not null;
+          alter table paused modify paused_at timestamp(3) null;
+          alter table resource_last_checked modify at timestamp(3) not null;
+          alter table task_tracking modify timestamp timestamp(3) not null;
+          alter table unhappy_veto modify recheck_time timestamp(3) not null;
```
robfletcher, line 23: done with `sql` rather than `modifyDataType` as the latter does not preserve `null` / `not null` (and other column. settings) on MySQL.

keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
```diff
@@ -0,0 +1,48 @@
...
+          alter table artifact_last_checked modify at timestamp(3) not null;
+          alter table delivery_config_last_checked modify at timestamp(3) not null;
+          alter table environment_artifact_constraint modify created_at timestamp(3) not null;
+          alter table environment_artifact_constraint modify judged_at timestamp(3) null;
+          alter table environment_artifact_versions modify approved_at timestamp(3) null;
+          alter table environment_artifact_versions modify deployed_at timestamp(3) null;
+          alter table environment_artifact_versions modify replaced_at timestamp(3) null;
+          alter table event modify timestamp timestamp(3) not null;
+          alter table paused modify paused_at timestamp(3) null;
+          alter table resource_last_checked modify at timestamp(3) not null;
```
robfletcher, line 48: These are the columns that are already `timestamp`,  I'm converting them to `timestamp(3)` (millisecond precision).

Done with `sql` rather than `modifyDataType` as the latter does not preserve `null` / `not null` (and other column settings) on MySQL.

keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
```diff
@@ -0,0 +1,48 @@
...
+
+          alter table task_tracking change column timestamp old bigint(13);
+          alter table task_tracking add column timestamp timestamp(3) not null;
+          update task_tracking set timestamp = from_unixtime(old / 1000);
+          alter table task_tracking drop column old;
+
+          alter table unhappy_veto change column recheck_time old bigint(13);
+          alter table unhappy_veto add column recheck_time timestamp(3) not null;
+          update unhappy_veto set recheck_time = from_unixtime(old / 1000);
+          alter table unhappy_veto drop column old;
```
robfletcher, line 36: For columns that are currently `bigint` I needed to create the new column, update converting the unix millisecond to a timestamp value, then drop the original column.

robfletcher:
I think one of the beauties of Keel is that those kind of transient errors don't cause us any significant issues. We can live with the old server group experiencing errors for a few seconds because our regular check cycle means the newer server group will pick things up almost immediately.

That said, I think getting as many of these kinds of housekeeping changes done as we can before we broaden our user-base makes sense.

PR #1225, opened 2020-05-23
fix(db): Fix prior naive migration

robfletcher:


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepositoryTests.kt
```diff
@@ -132,3 +157,9 @@ abstract class UnhappyVetoRepositoryTests<T : UnhappyVetoRepository> : JUnit5Min
     }
   }
 }
+
+private val Assertion.Builder<UnhappyVetoStatus>.shouldSkip: Assertion.Builder<Boolean>
+  get() = get("should skip") { shouldSkip }
+
+private val Assertion.Builder<UnhappyVetoStatus>.shouldRecheck: Assertion.Builder<Boolean>
+  get() = get("should recheck") { shouldRecheck }
```
robfletcher, line 165: I was just finding it impossible to tell which field was which when the test was failing. This adds a description to the output.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepository.kt
```diff
@@ -67,9 +67,9 @@ abstract class UnhappyVetoRepository(
    */
   abstract fun getAllForApp(application: String): Set<String>
 
-  fun calculateExpirationTime(wait: Duration): Instant =
+  fun calculateExpirationTime(wait: Duration?): Instant? =
     when (wait) {
-      Duration.ZERO -> Instant.ofEpochMilli(Long.MAX_VALUE)
+      null -> null
```
lorin, line 72: Here it is, the first good use of `null` I've seen.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryUnhappyVetoRepository.kt
```diff
@@ -28,42 +28,42 @@ class InMemoryUnhappyVetoRepository(
...
     val record = resources[resourceId]
     if (record == null) {
       resources[resourceId] = Record(application, calculateExpirationTime(wait))
       return UnhappyVetoStatus(shouldSkip = true, shouldRecheck = false)
     }
     return UnhappyVetoStatus(
-      shouldSkip = record.recheckTime > clock.instant(),
-      shouldRecheck = record.recheckTime < clock.instant()
+      shouldSkip = record.recheckTime == null || record.recheckTime > clock.instant(),
+      shouldRecheck = record.recheckTime != null && record.recheckTime < clock.instant()
```
lorin, line 47: I think it would be easier to read if the logic (comparing the recheckTime to now) was encapsulated in a helper function. The same logic is used four times in this file, and I have trouble interpreting what `recheckTime < clock.instant()` means at a glance. Having a semantically meaningful function name instead would help.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
```diff
@@ -31,14 +32,30 @@ class SqlUnhappyVetoRepository(
...
 
-  override fun markUnhappyForWaitingTime(resourceId: String, application: String, wait: Duration) {
+  override fun markUnhappyForWaitingTime(resourceId: String, application: String, wait: Duration?) {
     sqlRetry.withRetry(WRITE) {
       jooq.insertInto(UNHAPPY_VETO)
         .set(UNHAPPY_VETO.RESOURCE_ID, resourceId)
         .set(UNHAPPY_VETO.APPLICATION, application)
-        .set(UNHAPPY_VETO.RECHECK_TIME, calculateExpirationTime(wait).toTimestamp())
+        .run {
+          calculateExpirationTime(wait)?.toTimestamp().let { expiryTime ->
```
lorin, line 41: I think it's a bit nicer if you use `?.let`:

```kotlin
calculateExpirationTime(wait)
  ?.toTimestamp()
  ?.let { expiryTime -> set(UNHAPPY_VETO.RECHECK_TIME, expiryTime) }
  ?: setNull(UNHAPPY_VETO.RECHECK_TIME)
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepository.kt
robfletcher: I don't think null per se is the problem. It's languages (like Java) that fail to distinguish the possibility of a variable being null so it's very easy to accidentally mishandle potentially null values. That's something Kotlin (mostly) fixes (not that it's the first language to do so).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryUnhappyVetoRepository.kt
robfletcher: I have many questions about this implementation, for example: why are there two separate booleans when it seems to me that one would suffice? I can certainly tidy this up but it also feels to me that a slightly more extensive refactor might not be the worst idea.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryUnhappyVetoRepository.kt
```diff
@@ -28,42 +28,45 @@ class InMemoryUnhappyVetoRepository(
...
   fun flush() = resources.clear()
 
   private data class Record(
     val application: String,
-    val recheckTime: Instant
+    val recheckTime: Instant?
   )
+
+  private fun Instant?.shouldSkip(now: Instant = clock.instant()) = this == null || this > now
+  private fun Instant?.shouldRecheck(now: Instant = clock.instant()) = this != null && this < now
```
robfletcher, line 71: @lorin encapsulated the logic in functions as suggested

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryUnhappyVetoRepository.kt
```diff
@@ -28,42 +28,45 @@ class InMemoryUnhappyVetoRepository(
...
-      shouldRecheck = record.recheckTime < clock.instant()
+      shouldSkip = record.recheckTime.shouldSkip(),
+      shouldRecheck = record.recheckTime.shouldRecheck()
     )
   }
 
   override fun getAll(): Set<String> {
     val now = clock.instant()
-    return resources.filter { it.value.recheckTime > now }.keys.toSet()
+    return resources.filter { it.value.recheckTime.shouldSkip(now) }.keys.toSet()
```
luispollo, line 53: This seemed backwards to me at first sight in my he context of this particular function. It’d be easier to read if the extension was named something like `isNullOrAfter()`, IMHO.

luispollo: Same thing with `shouldRecheck()`. I’d go with `isBefore()`, which even matches the existing function name/behavior in non-nullable `Instant`.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
```diff
@@ -51,18 +68,18 @@ class SqlUnhappyVetoRepository(
...
+          .from(UNHAPPY_VETO)
+          .where(UNHAPPY_VETO.RESOURCE_ID.eq(resourceId))
+          .fetchOne()
+      }
+      ?.let { (recheckTime: LocalDateTime?) ->
         return UnhappyVetoStatus(
-          shouldSkip = recheckTime > clock.timestamp(),
-          shouldRecheck = recheckTime < clock.timestamp()
+          shouldSkip = recheckTime == null || recheckTime > clock.timestamp(),
+          shouldRecheck = recheckTime != null && recheckTime < clock.timestamp()
```
luispollo, line 82: `recheckTime <= clock.timestamp()`? (One of them has to be inclusive, right?)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
```diff
@@ -75,7 +92,7 @@ class SqlUnhappyVetoRepository(
     return sqlRetry.withRetry(READ) {
       jooq.select(UNHAPPY_VETO.RESOURCE_ID)
         .from(UNHAPPY_VETO)
-        .where(UNHAPPY_VETO.RECHECK_TIME.greaterOrEqual(now))
+        .where(UNHAPPY_VETO.RECHECK_TIME.isNull.or(UNHAPPY_VETO.RECHECK_TIME.greaterOrEqual(now)))
```
luispollo, line 95: Can’t see the context around this line on the phone, but this needs to be consistent with my comment on one of the checks for skip or recheck being inclusive.

keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
```diff
@@ -1,39 +1,153 @@
 databaseChangeLog:
 - changeSet:
-    id: event-uid-timestamp-index
+    id: standardize-timestamps-agent_lock
```
luispollo, line 3: Did you rollback the migration in prestaging? If not, won’t reusing the migration filename cause a conflict when liquibase tries to run it again?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
robfletcher: For some reason this behaves differently and breaks the tests. I will figure it out tho.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
robfletcher: You'd think, but I didn't change the logic.

keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
robfletcher: I did roll it back.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
luispollo: Oh boy...

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
```diff
@@ -31,14 +32,30 @@ class SqlUnhappyVetoRepository(
...
-  override fun markUnhappyForWaitingTime(resourceId: String, application: String, wait: Duration) {
+  override fun markUnhappyForWaitingTime(resourceId: String, application: String, wait: Duration?) {
     sqlRetry.withRetry(WRITE) {
       jooq.insertInto(UNHAPPY_VETO)
         .set(UNHAPPY_VETO.RESOURCE_ID, resourceId)
         .set(UNHAPPY_VETO.APPLICATION, application)
-        .set(UNHAPPY_VETO.RECHECK_TIME, calculateExpirationTime(wait).toTimestamp())
+        .run {
+          calculateExpirationTime(wait)?.toTimestamp().let { expiryTime ->
+            if (expiryTime == null) {
```
luispollo, line 42: You're checking against nullability on the return of `calculateExpirationTime()`, so I assume any `Instant.toTimestamp()` would never be `null`?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
```diff
@@ -31,14 +32,30 @@ class SqlUnhappyVetoRepository(
...
+              setNull(UNHAPPY_VETO.RECHECK_TIME)
+            } else {
+              set(UNHAPPY_VETO.RECHECK_TIME, expiryTime)
+            }
+          }
```
luispollo, line 47:
```suggestion
          calculateExpirationTime(wait)?.let { expiryTime ->
            if (expiryTime == null) {
              setNull(UNHAPPY_VETO.RECHECK_TIME)
            } else {
              set(UNHAPPY_VETO.RECHECK_TIME, expiryTime.toTimestamp())
            }
          }
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
```diff
@@ -31,14 +32,30 @@ class SqlUnhappyVetoRepository(
...
         .onDuplicateKeyUpdate()
-        .set(UNHAPPY_VETO.RECHECK_TIME, calculateExpirationTime(wait).toTimestamp())
+        .run {
+          calculateExpirationTime(wait)?.toTimestamp().let { expiryTime ->
+            if (expiryTime == null) {
+              setNull(UNHAPPY_VETO.RECHECK_TIME)
+            } else {
+              set(UNHAPPY_VETO.RECHECK_TIME, expiryTime)
+            }
+          }
```
luispollo, line 57:
```suggestion
          calculateExpirationTime(wait).let { expiryTime ->
            if (expiryTime == null) {
              setNull(UNHAPPY_VETO.RECHECK_TIME)
            } else {
              set(UNHAPPY_VETO.RECHECK_TIME, expiryTime.toTimestamp())
            }
          }
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepository.kt
lorin: I'm a fan of Haskell's Maybe monad for dealing with this sort of thing.

keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
```diff
@@ -1,39 +1,153 @@
...
     author: fletch
     changes:
+    - addColumn:
+        tableName: agent_lock
+        columns:
+        - column:
+            name: expiry_new
+            type: timestamp(3)
+            constraints:
+            - nullable: false
```
lorin, line 13: Just to check my understanding: if you create a non-nullable timestamp column, MySQL will default the values to the current timestamp. Is that correct?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
robfletcher: d'oh. Of course

robfletcher: No, `toTimestamp` is only called if the result of `calculateExpirationTime` is not null. That means the overall  value of `calculateExpirationTime(wait)?.toTimestamp()` is `LocalDateTime?`.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
robfletcher: I like that better, but it doesn't fix the test!

keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
robfletcher: correct.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepository.kt
```diff
@@ -40,7 +40,7 @@ abstract class UnhappyVetoRepository(
   abstract fun markUnhappyForWaitingTime(
     resourceId: String,
     application: String,
-    wait: Duration = Duration.parse(waitingTime)
+    wait: Duration? = Duration.parse(waitingTime)
```
lorin, line 43: Probably worth a comment in the function header that passing `null` for the wait means that it is unhappy forever.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryUnhappyVetoRepository.kt
```diff
@@ -28,42 +28,45 @@ class InMemoryUnhappyVetoRepository(
 
   private val resources: MutableMap<String, Record> = mutableMapOf()
 
-  override fun markUnhappyForWaitingTime(resourceId: String, application: String, wait: Duration) {
+  override fun markUnhappyForWaitingTime(resourceId: String, application: String, wait: Duration?) {
```
lorin, line 31: As an aside, how close are we to eliminating the in memory repositories altogether?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
lorin: Look at the code, it's not obvious to me why `shouldRecheck` wasn't defined as `!shouldSkip` (or vice versa) in the first place. I understand the desire to not change existing behavior, but in a future PR I think maybe we should revisit this.

@emjburns Do you have any insight here?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepository.kt
robfletcher: I've made the observation before that Kotlin's nullable types are an option/maybe monad with some syntactic sugar.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryUnhappyVetoRepository.kt
robfletcher: I've got most of the usages removed. I'll continue with that work this week.

keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
```diff
@@ -1,39 +1,153 @@
...
+            name: first_detection_time
+            type: timestamp(3)
+            constraints:
+            - nullable: false
+    - sql:
+        sql: |
+          update diff_fingerprint set first_detection_time = from_unixtime(first_detection_time_old / 1000);
+    - dropColumn:
+        tableName: diff_fingerprint
+        columnName: first_detection_time_old
```
robfletcher, line 47: This is my template for migrating the bigint columns to timestamp:

1. rename it to `*_old`
2. add the new column
3. run an update to populate the new column
4. drop the `*_old` column

keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
```diff
@@ -1,39 +1,153 @@
...
+        oldColumnName: recheck_time
+        newColumnName: recheck_time_old
+        columnDataType: bigint(13)
+    - dropNotNullConstraint:
+        tableName: unhappy_veto
+        columnName: recheck_time_old
+        columnDataType: bigint(13)
+    - sql:
+        sql: |
+          update unhappy_veto set recheck_time_old = null where recheck_time_old = 9223372036854775807;
```
robfletcher, line 132: With `unhappy_veto` we no longer want to be using this `Long.MAX_VALUE` to represent "never re-check" since it can't be converted to a timestamp. Instead that will now be indicated by `null`.

keel-sql/src/main/resources/db/changelog/20200522-standardize-timestamps.yml
```diff
@@ -1,39 +1,153 @@
...
+    - dropColumn:
+        tableName: unhappy_veto
+        columnName: recheck_time_old
+- changeSet:
+    id: standardize-timestamps-add-subsecond-precision
+    author: fletch
+    changes:
     - sql:
         sql: |
           alter table artifact_last_checked modify at timestamp(3) not null;
```
robfletcher, line 153: For columns that are already `timestamp` we just need to add the precision

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/api/ImageSpec.kt
```diff
@@ -24,5 +24,5 @@ data class ImageSpec(
   override val maxDiffCount = 2
 
   @JsonIgnore
-  override val unhappyWaitTime = Duration.ZERO
+  override val unhappyWaitTime: Duration? = null
```
robfletcher, line 27: Made this change in the relevant handler types. It looks like nothing ever uses any other value.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
robfletcher: I've updated this now to match the query. As Luis pointed out, it was inconsistent.

PR #1237, opened 2020-05-27
Remove in-memory repositories

robfletcher:
This PR removes the in-memory repositories. Reasons for this:

- They take additional effort to maintain but we never use them outside of tests.
- They can result in convoluted tests that rely on more complex dependency wiring and data setup than would otherwise be necessary.
- They can result in tests that are indirectly testing behavior of the repositories themselves rather than the components that use them.


keel-web/src/test/resources/application.properties
```diff
@@ -7,3 +7,7 @@ front50.baseUrl=https://localhost:8082
 igor.baseUrl=https://localhost:8083
 echo.baseUrl=https://localhost:8086
 services.fiat.baseUrl=https://fiat.net
+sql.enabled=true
+sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename
+sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename
+spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver
```
robfletcher, line 13: This is the key interesting/contentious bit. We need to run the tests in `keel-web` (which pretty much all stand up the Spring container) using testcontainers. This may affect speed.

robfletcher: Rough speed comparison before & after on my laptop:
- `gradle :keel-web:clean :keel-web:test --no-build-cache  1.78s user 0.27s system 1% cpu 1:59.31 total`
- `gradle :keel-web:clean :keel-web:test --no-build-cache  1.82s user 0.27s system 1% cpu 2:15.35 total`

PR #1239, opened 2020-05-27
Include @Nullable @Nonnull annotations on JOOQ-generated code

robfletcher:


keel-sql/keel-sql.gradle.kts
```diff
@@ -48,7 +48,7 @@ dependencies {
   implementation("com.netflix.spinnaker.kork:kork-sql")
   implementation("org.springframework:spring-jdbc")
   implementation("org.springframework:spring-tx")
-  implementation("org.jooq:jooq:3.11.11")
+  implementation("org.jooq:jooq:3.13.2")
```
robfletcher, line 51: I don't know why we were pinning to a particular version before. I added it now until we can get 3.13.x in Kork.

keel-sql/src/main/resources/jooqConfig.xml
```diff
@@ -28,6 +28,15 @@
...
     <generate>
       <javaTimeTypes>true</javaTimeTypes>
+
+      <generatedAnnotation>true</generatedAnnotation>
+      <generatedAnnotationType>DETECT_FROM_JDK</generatedAnnotationType>
+
+      <nullableAnnotation>true</nullableAnnotation>
+      <nullableAnnotationType>javax.annotation.Nullable</nullableAnnotationType>
+      <nonnullAnnotation>true</nonnullAnnotation>
+      <nonnullAnnotationType>javax.annotation.Nonnull</nonnullAnnotationType>
```
robfletcher, line 38: These config options were added in JOOQ 3.13.x

lorin:
Interesting to me that these annotations didn't introduce any kotlin compile-time errors.

robfletcher:
I wouldn't expect errors (as that would imply we currently treat the types inappropriately which would be exploding at runtime). I'm surprised there are no warnings, though.

PR #1240, opened 2020-05-27
Stop pinning JOOQ to an old version

robfletcher:


keel-sql/keel-sql.gradle.kts
```diff
@@ -48,7 +48,7 @@ dependencies {
   implementation("com.netflix.spinnaker.kork:kork-sql")
   implementation("org.springframework:spring-jdbc")
   implementation("org.springframework:spring-tx")
-  implementation("org.jooq:jooq:3.11.11")
+  implementation("org.jooq:jooq")
```
robfletcher, line 51: I don't know why we are doing this

PR #1243, opened 2020-05-27
feat(deploy): Make red-black wait time configurable, orca stage timeout based on all delays

### LINKED ISSUE #1180: Update timeout on deploy stages if user adds wait before scale down
question from @jonsie:
> The deploy is configured to wait 1H before scaling down and terminating the previous instances.  During that time, I made another change to the MD config, and approved the deploy of a new artifact.  Is the expected behavior for the 1H wait of the previous deploy to terminate? 

It should wait, but the problem was the task failed with `Stage Wait Before Scale Down timed out after 30 minutes 29 seconds`. We should update that timeout if the user has selected a wait

luispollo:
Adds a `waitForInstancesUp` configuration to the `red-black` deployment strategy, and replaces the hard-coded 30 minute orca stage timeout for deployment jobs with the sum of this new duration plus existing delay fields in the config (`delayBeforeScaleDown` and `delayBeforeDisable`).

For example, the following config:
```yaml
          deployWith:
            delayBeforeScaleDown: PT1H
            resizePreviousToZero: true
            rollbackOnFailure: false
            strategy: red-black
```
...would translate to a 1 hour delay before scale down + (default) 30 min wait for instances up = 1h30min timeout for the orca task.

Fixes #1180.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
@@ -50,8 +50,6 @@ data class RedBlack(
 ) : ClusterDeployStrategy() {
 
   companion object {
-    val ORCA_STAGE_TIMEOUT_MARGIN: Duration = Duration.ofMinutes(5)
-
     fun fromOrcaStageContext(context: Map<String, Any?>) =
       RedBlack(
         rollbackOnFailure = context["rollback"]
@@ -79,8 +77,7 @@ data class RedBlack(
     "stageTimeoutMs" to (
       (waitForInstancesUp ?: DEFAULT_WAIT_FOR_INSTANCES_UP) +
       (delayBeforeDisable ?: ZERO) +
-      (delayBeforeScaleDown ?: ZERO) +
-      ORCA_STAGE_TIMEOUT_MARGIN
+      (delayBeforeScaleDown ?: ZERO)
     ).toMillis()
   )
 
@@ -101,7 +98,7 @@ data class RedBlack(
 object Highlander : ClusterDeployStrategy() {
   override fun toOrcaJobProperties() = mapOf(
     "strategy" to "highlander",
-    "stageTimeoutMs" to (DEFAULT_WAIT_FOR_INSTANCES_UP + RedBlack.ORCA_STAGE_TIMEOUT_MARGIN).toMillis()
+    "stageTimeoutMs" to DEFAULT_WAIT_FOR_INSTANCES_UP.toMillis()
   )
 
   override fun withDefaultsOmitted() = this
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
@@ -39,12 +39,16 @@ data class RedBlack(
   val maxServerGroups: Int? = 2,
   val delayBeforeDisable: Duration? = ZERO,
   val delayBeforeScaleDown: Duration? = ZERO,
+  val waitForInstancesUp: Duration? = DEFAULT_WAIT_FOR_INSTANCES_UP,
   // The order of this list is important for pauseTime based staggers
   @JsonInclude(NON_EMPTY)
   override val stagger: List<StaggeredRegion> = emptyList()
 ) : ClusterDeployStrategy() {
 
   companion object {
+    val DEFAULT_WAIT_FOR_INSTANCES_UP: Duration = Duration.ofMinutes(30)
+    val ORCA_STAGE_TIMEOUT_MARGIN: Duration = Duration.ofMinutes(5)
+
     fun fromOrcaStageContext(context: Map<String, Any?>) =
       RedBlack(
         rollbackOnFailure = context["rollback"]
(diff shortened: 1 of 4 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
```diff
@@ -87,7 +97,8 @@ data class RedBlack(
 @JsonTypeName("highlander")
 object Highlander : ClusterDeployStrategy() {
   override fun toOrcaJobProperties() = mapOf(
-    "strategy" to "highlander"
+    "strategy" to "highlander",
+    "stageTimeoutMs" to Duration.ofMinutes(30).toMillis()
```
emjburns, line 101: This should probably be `DEFAULT_WAIT_FOR_INSTANCES_UP` as well

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -537,15 +537,6 @@ class ClusterHandler(
           0 -> emptyList()
           else -> listOf(startingRefId.toString())
         },
```
emjburns, line 539: probably also needs to be changed in the `TitusClusterHandler`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
luispollo: OK, will fix.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: I was hoping to leave that for later, but will do.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ClusterDeployStrategy.kt
```diff
@@ -39,12 +39,16 @@ data class RedBlack(
...
   val delayBeforeScaleDown: Duration? = ZERO,
+  val waitForInstancesUp: Duration? = DEFAULT_WAIT_FOR_INSTANCES_UP,
   // The order of this list is important for pauseTime based staggers
   @JsonInclude(NON_EMPTY)
   override val stagger: List<StaggeredRegion> = emptyList()
 ) : ClusterDeployStrategy() {
 
   companion object {
+    val DEFAULT_WAIT_FOR_INSTANCES_UP: Duration = Duration.ofMinutes(30)
+    val ORCA_STAGE_TIMEOUT_MARGIN: Duration = Duration.ofMinutes(5)
```
jonsie, line 50: Thoughts on just removing `ORCA_STAGE_TIMEOUT_MARGIN`?.  

That way there is no additional timeout that needs to be reasoned about beyond what is configurable (delayBeforeDisable, delayBeforeScaleDown, and  waitForInstancesUp).  I'm not sure this additional timeout margin buys us much.

luispollo: Yeah, that's a good question. I think we could get rid of it, but @emjburns suggested it so I'll defer to her for a final call.

luispollo: You know what? Was the "informed captain" on this one, I'm making the call to remove it. 😄 
The user can adjust the timeout in a more predictive manner with `waitForInstancesUp`.

erikmunson:
This feels relatively reasonable to me, and certainly a lot less frustrating to reason about than adding up all these times in your head to give some sort of overall timeout.

From reading this is looks to me like the default setting is 30 mins, i.e. without any explicit config you get 30 + the default values for each other other two settings. Is that correct? Only question if so would be: how often do we think folks would need to explicitly configure this? It's always going to be a tough tradeoff between keeping the default short enough to make _most_ deploys faster, without going so low that lots of people have to deal with yet another config field.

luispollo:
> This feels relatively reasonable to me, and certainly a lot less frustrating to reason about than adding up all these times in your head to give some sort of overall timeout.

Hmm, actually this implementation implies that you _do_ have to add the times up in your head to get the overall timeout, as the PR description states...

> From reading this is looks to me like the default setting is 30 mins, i.e. without any explicit config you get 30 + the default values for each other other two settings. Is that correct? 

Correct. The defaults for both the other two delay fields is zero, so the default timeout overall is 30 minutes.

> Only question if so would be: how often do we think folks would need to explicitly configure this? It's always going to be a tough tradeoff between keeping the default short enough to make _most_ deploys faster, without going so low that lots of people have to deal with yet another config field.

Personally, I think it's kinda ridiculous for an instance to take longer anywhere near 30 minutes to come up, so given that we do provide a way to override the default, I think we should keep it short. On a related note, though, we really need to document these deployment strategy options better.

emjburns:
Retroactive LGTM, thanks for the fix @luispollo!

I think @erikmunson means adding the times in your head and then figuring out what your timeout should be, and putting that. It's good if the code adds for us.

I think 30 minutes is very reasonable and I wouldn't expect people to change this unless they're already doing something funky which they're aware of.

Issue #1084, opened 2020-04-30
We shouldn't allow two active server groups


emjburns:
We just discovered a situation where a cluster shows happy, but there are two active server groups in one region. We shouldn't consider this situation correct, we should instead scale down the other server group.

I'm wondering if this is the case because we just don't consider older server groups @robfletcher?

This is not great bug, we should fix it asap

robfletcher:
Yeah, this is because we literally just look at the latest server group. We can definitely handle this better.

lorin:
Yeah, looks like the `CloudDriverService.getServerGroups` method calls [activeServerGroup], that's a call agaisnt clouddriver that can only return a single ASG. We'll need to make a call that retrieves all of the active ASGs.


[activeServerGroup]: https://github.com/spinnaker/keel/blob/9cb5c4b0f7ad933cce3cbd32cf0ea3c237f26ad8/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt#L800

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

erikmunson:
Seemingly similar/related issue that came up internally, which I suspect might cause the same kinds of failure modes as this issue: https://github.com/spinnaker/keel/issues/1401

PR #1244, opened 2020-05-27
fix(clouddriver): Add support for fetching Titus server groups

lorin:
Add a method for retrieving Titus server groups.

Progress toward #1084.

This new code isn't being called anywhere yet, except for a controller only used for testing.


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/TitusActiveServerGroup.kt
```diff
@@ -20,6 +20,55 @@ package com.netflix.spinnaker.keel.clouddriver.model
...
+  val constraints: Constraints,
+  val tags: Map<String, String>,
+  val resources: Resources,
+  val capacityGroup: String,
+  val disabled: Boolean
+)
+
+fun TitusServerGroup.toActive() =
+  TitusActiveServerGroup(
+    name,
```
luispollo, line 50: I'd suggest adding the property names here. It's very hard to correlate the positional arguments with the properties, especially when there's a bunch of string values after one another.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/PassThruController.kt
```diff
@@ -32,14 +32,30 @@ class PassThruController(
     @PathVariable("cluster") cluster: String,
     @PathVariable("cloudProvider") cloudProvider: String,
     @PathVariable("region") region: String
-  ): List<ActiveServerGroup> =
+  ): List<Any> =
```
luispollo, line 35: Hmm... Is there an opportunity here for us to define a base class so that this could be typed `<T : BaseServerGroup>` instead of `Any`? You could also use the same base type in `ServerGroupCollection`.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/TitusActiveServerGroup.kt
lorin: Fixed in 306c3194

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/PassThruController.kt
lorin: I took a shot at this in 34a20d1

Issue #1242, opened 2020-05-27
Disallow multiple delivery configs from managing the same resource


lorin:
Currently, we don't guard against two delivery configs managing the same resource. If that happens, then the `SqlDeliveryConfigRepository.deliveryConfigFor` method will throw the following exception:

```
org.jooq.exception.TooManyRowsException: Cursor returned more than one result
```

We should add a check to prevent creating or modifying a delivery config that manages the same resource as an existing config.

PR #1245, opened 2020-05-28
fix(duplicateResources): don't allow resources to be managed in more than 1 config

gal-yardeni:
fixes https://github.com/spinnaker/keel/issues/1242.

Currently we allow a single resource to be managed by more than 1 delivery config file. This fix is blocking it by making sure the combination of a delivery config name + resource id is unique.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
@@ -3,7 +3,6 @@ package com.netflix.spinnaker.keel.persistence
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.Resource
-import com.netflix.spinnaker.keel.api.ResourceSpec
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactType
 import com.netflix.spinnaker.keel.api.artifacts.DockerArtifact
 import com.netflix.spinnaker.keel.api.artifacts.TagVersionStrategy.BRANCH_JOB_COMMIT_BY_JOB
@@ -259,7 +258,7 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
           context("after an update") {
             before {
               resourcesDueForCheck()
-              subject.upsertResource<ResourceSpec>(resource.copy(spec = DummyResourceSpec(id = resource.spec.id, data = "kthxbye")), deliveryConfig.name)
+              subject.upsertResource(resource.copy(spec = DummyResourceSpec(id = resource.spec.id, data = "kthxbye")), deliveryConfig.name)
             }
 
             test("stores the updated resource") {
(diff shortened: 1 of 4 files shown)


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -246,7 +259,7 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
           context("after an update") {
             before {
               resourcesDueForCheck()
-              subject.upsert(resource.copy(spec = DummyResourceSpec(id = resource.spec.id, data = "kthxbye")))
```
gal-yardeni, line 249: I think some of the tests handling a single resource creation/update, without a delivery config associated with it and not relevant anymore.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -294,5 +306,36 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
...
+      before {
+        subject.upsertDeliveryConfig(deliveryConfig)
+      }
+      test("trying to persist another config with the same resource") {
+        expectThrows<DuplicateManagedResourceException> { subject.upsertDeliveryConfig(secondDeliveryConfig) }
+        expectThrows<NoSuchDeliveryConfigException> { deliveryConfigRepository.get(secondConfigName) }
+        expectCatching { resourceRepository.get(firstResource.id) }.isSuccess()
+      }
+
+      context("delete the resource from the first config, and try to persist again") {
```
emjburns, line 320: This test probably isn't necessary, since it's just testing normal saving. Do you feel like it adds any other value?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/DuplicateManagedResourceException.kt
```diff
@@ -0,0 +1,10 @@
+package com.netflix.spinnaker.keel.exceptions
+
+class DuplicateManagedResourceException(
+  val id: String,
+  existingConfig: String,
+  newConfig: String
+) : ValidationException(
+  "Resource with id $id exist in a delivery config named ($existingConfig). " +
+    "Please ensure to remove reousrce $id from config named $existingConfig, if you wish this resource to be managed in a $newConfig."
```
emjburns, line 9: ``
```suggestion
    "Please ensure to remove reousrce $id from config named $existingConfig, if you wish this resource to be managed in $newConfig."
```

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
gal-yardeni: not really :) I can remove those tests.

emjburns: I lean towards no, because this is more of a unit test. I think there's a better place to test "if resource doesn't exist it is saved" (in the deliveryConfigRepositoryTests, already exists). I think it's really important to test that we error on a bad condition here, but not the remediation steps because that's just the happy path.

gal-yardeni: ok, that makes sense to me. will remove it

PR #1044, opened 2020-04-19
chore(build): update to newer spinnaker project plugin

cfieber:
updates to gradle 6.3

updates github actions workflows for new release flags

PR #1249, opened 2020-05-29
fix(coverage): Reenable generation of jacoco report

luispollo:
The `jacocoAggregateReport` task  was inadvertently removed in #1044.


luispollo:
Not sure what's going with Mergify. Merging by hand.

Issue #870, opened 2020-03-10
Refactor import delivery config flow


luispollo:
Currently, orca's `ImportDeliveryConfigStage` pulls the manifest file from source control via igor, then pushes it to keel to store/update. We'd like to change that behavior such that keel does the pulling by exposing a new API like `/delivery-configs/import`, which has 2 advantages:
1. Allows keel to store metadata about the source control repo in its database.
2. Brings us closer to having 1st-class support for GitOpsy flows _without_ pipelines.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

luispollo:
@spinnakerbot remove-label stale

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

PR #1261, opened 2020-06-02
feat(import): API to import delivery config from source

luispollo:
Implements a new `POST /delivery-configs/import` API that performs the same steps as the orca `ImportDeliveryConfigTask`, namely retrieving a specified delivery config from source control via igor, and upserting the delivery config into the keel database.

This PR also takes advantage of the fact that we now have metadata available in keel about the provenance of the delivery config, and adds a `metadata` column to the `delivery_config` table to store that information.

The goal of this PR is to make progress towards #870 and provide a more self-contained way to test the import flow locally by avoiding as many dependencies and steps as possible in the manual test flow. For example, with this change, it's straightforward to replicate the error from #1262 locally.

### DIFF
--- a/keel-igor/src/main/kotlin/com/netflix/spinnaker/igor/ScmService.kt
+++ b/keel-igor/src/main/kotlin/com/netflix/spinnaker/igor/ScmService.kt
@@ -1,0 +1,18 @@
+package com.netflix.spinnaker.igor
+
+import retrofit2.http.GET
+import retrofit2.http.Query
+
+/**
+ * Igor methods related to Source Control Management (SCM) operations.
+ */
+interface ScmService {
+  @GET("/delivery-config/manifest")
+  suspend fun getDeliveryConfigManifest(
+    @Query("scmType") repoType: String,
+    @Query("project") projectKey: String,
+    @Query("repository") repositorySlug: String,
+    @Query("manifest") manifestPath: String,
+    @Query("ref") ref: String? = null
+  ): Map<String, Any?>
+}
--- a/keel-igor/src/main/kotlin/com/netflix/spinnaker/igor/ScmService.kt
+++ b/keel-igor/src/main/kotlin/com/netflix/spinnaker/igor/ScmService.kt
@@ -1,5 +1,6 @@
 package com.netflix.spinnaker.igor
 
+import com.netflix.spinnaker.keel.core.api.SubmittedDeliveryConfig
 import retrofit2.http.GET
 import retrofit2.http.Query
 
@@ -7,6 +8,18 @@ import retrofit2.http.Query
  * Igor methods related to Source Control Management (SCM) operations.
  */
 interface ScmService {
+  /**
+   * Retrieves a delivery config manifest from a source control repository.
+   *
+   * @param repoType The type of SCM repository (e.g. "stash", "github")
+   * @param projectKey The "project" within the SCM system where the repository exists, which can be a user's personal
+   *        area (e.g. "SPKR", "~lpollo")
+   * @param repositorySlug The repository name (e.g. "myapp")
+   * @param manifestPath The path of the manifest file, relative to the base-path configured by the Spinnaker operator
+   *        in igor (which defaults to ".spinnaker"), for example "mydir/spinnaker.yml". The full path to the file
+   *        is determined by concatenating the base path with this relative path (e.g. ".spinnaker/mydir/spinnaker.yml").
+   * @param ref The git reference at which to retrieve to file (e.g. a commit hash, or a reference like "refs/heads/mybranch").
+   */
   @GET("/delivery-config/manifest")
   suspend fun getDeliveryConfigManifest(
     @Query("scmType") repoType: String,
@@ -14,5 +27,5 @@ interface ScmService {
     @Query("repository") repositorySlug: String,
     @Query("manifest") manifestPath: String,
     @Query("ref") ref: String? = null
-  ): Map<String, Any?>
+  ): SubmittedDeliveryConfig
 }
--- a/keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/DeliveryConfigImporter.kt
+++ b/keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/DeliveryConfigImporter.kt
@@ -1,0 +1,44 @@
+package com.netflix.spinnaker.keel.services
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.convertValue
+import com.netflix.spinnaker.igor.ScmService
+import com.netflix.spinnaker.keel.core.api.SubmittedDeliveryConfig
+import kotlinx.coroutines.runBlocking
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+
+/**
+ * Provides functionality to import delivery config manifests from source control repositories (via igor).
+ */
+@Component
+class DeliveryConfigImporter(
+  private val jsonMapper: ObjectMapper,
+  private val scmService: ScmService
+) {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  /**
+   * Imports a delivery config from source control via igor.
+   */
+  fun import(
+    repoType: String,
+    projectKey: String,
+    repoSlug: String,
+    manifestPath: String,
+    ref: String
+  ): SubmittedDeliveryConfig {
+    val manifestLocation = "$repoType://project:$projectKey/repo:$repoSlug/manifest:$manifestPath@$ref"
+
+    log.debug("Retrieving delivery config from $manifestLocation")
+    val submittedDeliveryConfig = runBlocking {
+      scmService.getDeliveryConfigManifest(repoType, projectKey, repoSlug, manifestPath, ref)
+        .let {
+          jsonMapper.convertValue<SubmittedDeliveryConfig>(it)
+        }
+    }
+
+    log.debug("Successfully retrieved delivery config from $manifestLocation.")
+    return submittedDeliveryConfig
+  }
+}
--- a/keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/DeliveryConfigImporter.kt
+++ b/keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/DeliveryConfigImporter.kt
@@ -39,6 +39,16 @@ class DeliveryConfigImporter(
     }
 
     log.debug("Successfully retrieved delivery config from $manifestLocation.")
-    return submittedDeliveryConfig
+    return submittedDeliveryConfig.copy(
+      metadata = mapOf("importedFrom" to
+        mapOf(
+          "repoType" to repoType,
+          "projectKey" to projectKey,
+          "repoSlug" to repoSlug,
+          "manifestPath" to manifestPath,
+          "ref" to ref
+        )
+      )
+    )
   }
 }
--- a/keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/DeliveryConfigImporter.kt
+++ b/keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/DeliveryConfigImporter.kt
@@ -1,7 +1,6 @@
 package com.netflix.spinnaker.keel.services
 
 import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.module.kotlin.convertValue
 import com.netflix.spinnaker.igor.ScmService
 import com.netflix.spinnaker.keel.core.api.SubmittedDeliveryConfig
 import kotlinx.coroutines.runBlocking
@@ -33,9 +32,6 @@ class DeliveryConfigImporter(
     log.debug("Retrieving delivery config from $manifestLocation")
     val submittedDeliveryConfig = runBlocking {
       scmService.getDeliveryConfigManifest(repoType, projectKey, repoSlug, manifestPath, ref)
-        .let {
-          jsonMapper.convertValue<SubmittedDeliveryConfig>(it)
-        }
     }
 
     log.debug("Successfully retrieved delivery config from $manifestLocation.")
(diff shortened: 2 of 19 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
```diff
@@ -61,13 +60,11 @@ class DefaultConfiguration(
   @Bean
   fun idGenerator(): ULID = ULID()
 
-  @Bean
+  @Bean(name = ["jsonMapper", "objectMapper"])
```
luispollo, line 63: Using a `@Qualifier` instead of bean names was causing issues with auto-detection of auto-wiring in IntelliJ.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -97,5 +102,22 @@ class DeliveryConfigController(
...
-  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+  @Operation(
+    description = "Imports a delivery config manifest from source control."
+  )
+  @PostMapping(
+    path = ["/import"],
+    consumes = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE],
+    produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  // Note: permissions are checked by the importer service
```
luispollo, line 113: This is because there's nothing to check the permissions against until we've retrieved the delivery config from SCM.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/DeliveryConfigImporter.kt
```diff
@@ -0,0 +1,58 @@
...
+          jsonMapper.convertValue<SubmittedDeliveryConfig>(it)
+        }
+    }
+
+    log.debug("Successfully retrieved delivery config from $manifestLocation. Checking permissions to upsert...")
+    authorizationSupport.checkApplicationPermission(WRITE, APPLICATION, submittedDeliveryConfig.application)
+    authorizationSupport.checkServiceAccountAccess(SERVICE_ACCOUNT, submittedDeliveryConfig.serviceAccount)
+
+    log.debug("Upserting delivery config '${submittedDeliveryConfig.name}' for app '${submittedDeliveryConfig.application}' from $manifestLocation")
+    val deliveryConfig = keelRepository.upsertDeliveryConfig(submittedDeliveryConfig)
```
gal-yardeni, line 53: can you please add `validation` here as well?

luispollo: Oh, good catch. I assumed `keelRepository.upsertDeliveryConfig` called the validator but I think that was before your change, right?

gal-yardeni: nope, it's decoupled from `keelRepository`.

luispollo: Done!

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
robfletcher: So, now we have 2 beans? The same bean registered twice? What happens if something injects `Collection<ObjectMapper>` (e.g. where we register sub-types), does the collection have 2 elements or 3? I think the point of naming it `jsonMapper` was to distinguish from `yamlMapper` which is also an instance of `ObjectMapper`.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -97,5 +107,26 @@ class DeliveryConfigController(
...
+    @RequestParam ref: String?
+  ): DeliveryConfig {
+    val deliveryConfig =
+      importer.import(repoType, projectKey, repoSlug, manifestPath, ref ?: "refs/heads/master")
+
+    authorizationSupport.checkApplicationPermission(WRITE, APPLICATION, deliveryConfig.application)
+    authorizationSupport.checkServiceAccountAccess(SERVICE_ACCOUNT, deliveryConfig.serviceAccount)
+
+    return upsert(deliveryConfig)
+  }
```
robfletcher, line 131: Something still needs to hit this endpoint then?

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
luispollo: That's still a problem, actually, because they're both the same type. This is not registering two beans, just having a primary name and an alias for the same bean. For consumers that only have 1 `ObjectMapper` injected, if it's named either of those names, auto-wiring will work automatically.

From [the docs](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html):
>**Bean Names**
> While a `name()` attribute is available, the default strategy for determining the name of a bean is to use the name of the `@Bean` method. This is convenient and intuitive, but if explicit naming is desired, the name attribute (or its alias value) may be used. Also note that name accepts an array of Strings, allowing for multiple names (i.e. a primary bean name plus one or more aliases) for a single bean.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
luispollo: Yes. I plan to refactor `orca` at some point to stop doing this work and just calling into this endpoint. For now, it's helped me troubleshoot and fix #1262.

keel-igor/src/main/kotlin/com/netflix/spinnaker/igor/ScmService.kt
```diff
@@ -0,0 +1,18 @@
...
+ */
+interface ScmService {
+  @GET("/delivery-config/manifest")
+  suspend fun getDeliveryConfigManifest(
+    @Query("scmType") repoType: String,
+    @Query("project") projectKey: String,
+    @Query("repository") repositorySlug: String,
+    @Query("manifest") manifestPath: String,
+    @Query("ref") ref: String? = null
+  ): Map<String, Any?>
```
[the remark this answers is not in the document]
luispollo: I started with `SubmittedDeliveryConfig` here, but I _think_ I saw the parsing error get wrapped in a `RetrofitError`, which meant that our custom `ExceptionHandler` wouldn't automatically kick in to format the response from the controller, so I added the explicit conversion there. I guess I could probably catch the `RetrofitError` there and look into the wrapped exception, but it seems harder to understand.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -97,5 +107,26 @@ class DeliveryConfigController(
...
+  )
+  @PostMapping(
+    path = ["/import"],
+    produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
+  )
+  fun import(
+    @RequestParam repoType: String,
+    @RequestParam projectKey: String,
+    @RequestParam repoSlug: String,
+    @RequestParam manifestPath: String,
```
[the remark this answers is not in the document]
luispollo: That's a good question. There's some history there. It's called a `path` because you can, indeed, have the file under a sub-directory. The catch is that we wanted to enforce the `.netflix` convention for the "root path" of the manifest, and so that is configured in `igor` and it won't look for manifests anywhere else in the source tree in SCM. This was also due to my concerns with just allowing other Spinnaker services to retrieve arbitrary files from source control via `igor`. So what `manifestPath` means is "the relative path to the manifest under the configured base path".

Examples:
- `spinnaker.yml` --> `.netflix/spinnaker.yml`
- `md/spinnaker.yml` --> `.netflix/md/spinnaker.yml`

If you have suggestions on how to improve the naming or anything here, let me know!

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
luispollo: This is not registering two beans, just having a primary name and an alias for the same bean. This now allows for auto-wiring with either of those names, without the need to use a `@Qualifier` at the point of injection. Auto-wiring for `yamlMapper` also continues to work.

From [the docs](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html):
>**Bean Names**
> While a `name()` attribute is available, the default strategy for determining the name of a bean is to use the name of the `@Bean` method. This is convenient and intuitive, but if explicit naming is desired, the name attribute (or its alias value) may be used. Also note that name accepts an array of Strings, allowing for multiple names (i.e. a primary bean name plus one or more aliases) for a single bean.

luispollo:
@erikmunson once this is integrated with orca, you'll get this kind of metadata with the delivery config:
```yaml
metadata:
  importedFrom:
    ref: "refs/heads/master"
    repoSlug: "server-lab-lpollo"
    repoType: "stash"
    projectKey: "~lpollo"
    manifestPath: "spinnaker.yml"
```

erikmunson:
NO. WAY.

I wanna get this into the `applications/{application}` if it’s alright with you. If it’s there I can add a link to the repo right in the UI.

Any chance we could have that application endpoint also provide a fully constructed URL that I could use in a link vs. needing to know how to build one?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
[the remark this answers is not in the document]
lorin: The only thing I can think of is adding a comment in a function header (e.g., in the `ScmService` interface).

keel-igor/src/main/kotlin/com/netflix/spinnaker/igor/ScmService.kt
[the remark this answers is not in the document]
lorin: Is there something specific about this use of Retrofit that leads to this problem or is this a potential issue with all of our retrofit clients? Should we be using `Map<String, Any?>` and explicitly doing the deserialization everywhere?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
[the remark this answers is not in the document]
luispollo: Will do. There's comments and user docs around this behavior in other places, FWIW.

keel-igor/src/main/kotlin/com/netflix/spinnaker/igor/ScmService.kt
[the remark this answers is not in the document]
luispollo: I will test this again to double check behavior.

erikmunson:
yes love it 💯

separate PR is great, just wanted to make sure doing that made sense first. As a side note I think it might also make sense to have the API give back a 'fallback' URL that just points to the API endpoint to see the manifest JSON, for times when it was submitted directly via the API and we can't point to a repo. I'd love to also have the repo information you've got listed here available in the application endpoint, but I think this config/manifest URL would probably be something of its own since the manifest isn't _always_ tied to version control.

gal-yardeni:
> @erikmunson once this is integrated with orca, you'll get this kind of metadata with the delivery config:
> 
> ```yaml
> metadata:
>   importedFrom:
>     ref: "refs/heads/master"
>     repoSlug: "server-lab-lpollo"
>     repoType: "stash"
>     projectKey: "~lpollo"
>     manifestPath: "spinnaker.yml"
> ```

What is the use case for using this information^? Just curious :)

erikmunson:
@gal-yardeni as we're currently thinking about it (subject to change) this is a good way to be able to link back to the actual source code of your delivery manifest in the UI — e.g. you could click a link on the Environments view like "View config" and get linked straight to the spinnaker.yml file in your repo. Useful if you aren't already looking at it and want to get there quickly, especially if you're not someone that is usually working with the manifest.

Later we might also come up with even more interesting ways of using it, but no idea what those might be 🤷

gal-yardeni:
@erikmunson ooo I see! Worth mentioning that we will have a link to stash (or other source code) in spinCI as well, and when will add the linking between pipeline <-> spinCI you can see it there.

luispollo:
> oh yes good call out — this will align pretty well with that i think. Maybe even good to have a repo link on this page that works the exact same way as the one on the integration tab.

Just a reminder that the delivery config manifest need not necessarily be in the same source code repo as the app. I've no idea what kinds of patterns will emerge as users start using MD more, but thought I should call that out. I think all the points above are still good/valid, just wanted to clarify that the link to the delivery config may not match the link(s) to the app's integration view.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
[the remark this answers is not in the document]
luispollo: Addressed in https://github.com/spinnaker/keel/pull/1261/commits/f84e2bf791c50f98b105b79b58318d55c569d7bf

keel-igor/src/main/kotlin/com/netflix/spinnaker/igor/ScmService.kt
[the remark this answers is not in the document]
luispollo: Turns out it was just a matter of extending `ExceptionHandler` a bit. Addressed in https://github.com/spinnaker/keel/pull/1261/commits/f84e2bf791c50f98b105b79b58318d55c569d7bf

Issue #1233, opened 2020-05-27
Add missing timestamps to stateful contraints and summaries


luispollo:
When things get "old" in the UI (e.g. constraint states, artifact vetoes), it's very hard to understand what's going on without timestamps in the cards in the artifact view.

In the example below, there's an artifact marked as bad, but there's no further information:
![image](https://user-images.githubusercontent.com/1323478/82964138-85086300-9f79-11ea-8a42-177fb54833f4.png)

The unanswered questions in this example are:
-  "When did something go wrong?"
-  "When was it marked as bad?"

In order to address that, we should:
- Add a `completedAt` (or similar) field to stateful constraints that keeps track of timing regardless of status (vs. `judgedAt/by`)
- Add a `vetoedAt` (or similar) field to the summary objects for artifact rollouts so we know when a veto happened

See also #939.

luispollo:
@erikmunson Finally added this, just FYI.

PR #1264, opened 2020-06-02
feat(mark-as-bad): add comment, show comment

emjburns:
Closes https://github.com/spinnaker/keel/issues/1233.

Adds a table to store 'mark as bad' comments + user + timestamp. 

`POST /applications/{application}/veto`
with body:

```kotlin
data class EnvironmentArtifactVeto(
  val reference: String,
  val version: String,
  val comment: String?
)
```
Marks an artifact as bad and saves the user who vetoed.

Unmark as bad is `DELETE /{application}/veto/{targetEnvironment}/{reference}/{version}`

(will make the corresponding updates in `gate` shortly)

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -11,6 +11,7 @@ import com.netflix.spinnaker.keel.api.id
 import com.netflix.spinnaker.keel.api.plugins.ResourceHandler
 import com.netflix.spinnaker.keel.api.plugins.supporting
 import com.netflix.spinnaker.keel.core.ResourceCurrentlyUnresolvable
+import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVeto
 import com.netflix.spinnaker.keel.diff.DefaultResourceDiff
 import com.netflix.spinnaker.keel.events.ResourceActuationLaunched
 import com.netflix.spinnaker.keel.events.ResourceActuationVetoed
@@ -94,7 +95,6 @@ class ResourceActuator(
            * containing [resource]. This ensures that the environment will be fully restored to
            * a prior good-state.
            */
-          // todo eb: limit this to not roll back forever
           if (response.vetoArtifact && resource.spec is VersionedArtifactProvider) {
             try {
               val versionedArtifact = when (desired) {
@@ -121,9 +121,14 @@ class ResourceActuator(
 
                   artifactRepository.markAsVetoedIn(
                     deliveryConfig = deliveryConfig,
-                    artifact = artifact,
-                    version = artifactVersion,
-                    targetEnvironment = environment)
+                    veto = EnvironmentArtifactVeto(
+                      reference = artifact.reference,
+                      version = artifactVersion,
+                      targetEnvironment = environment,
+                      vetoedBy = "Spinnaker",
+                      comment = "Automatically vetoed because deployment of this version failed."
+                    )
+                  )
                   // TODO: emit event + metric
                 }
               }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -126,7 +126,7 @@ class ResourceActuator(
                       version = artifactVersion,
                       targetEnvironment = environment,
                       vetoedBy = "Spinnaker",
-                      comment = "Automatically vetoed because deployment of this version failed."
+                      comment = "Automatically vetoed because multiple deployments of this version failed."
                     )
                   )
                   // TODO: emit event + metric
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -57,10 +57,10 @@ data class ArtifactSummaryInEnvironment(
   val deployedAt: Instant? = null,
   val replacedAt: Instant? = null,
   val replacedBy: String? = null,
-  val isPinned: Boolean = false,
   val pinned: ActionMetadata? = null,
-  val isVetoed: Boolean = false,
+  val isPinned: Boolean = pinned != null,
   val vetoed: ActionMetadata? = null,
+  val isVetoed: Boolean = vetoed != null,
   val statefulConstraints: List<StatefulConstraintSummary> = emptyList(),
   val statelessConstraints: List<StatelessConstraintSummary> = emptyList()
 )
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -58,12 +58,14 @@ data class ArtifactSummaryInEnvironment(
   val replacedAt: Instant? = null,
   val replacedBy: String? = null,
   val isPinned: Boolean = false,
-  val pinned: Pinned? = null,
+  val pinned: ActionMetadata? = null,
+  val isVetoed: Boolean = false,
+  val vetoed: ActionMetadata? = null,
   val statefulConstraints: List<StatefulConstraintSummary> = emptyList(),
   val statelessConstraints: List<StatelessConstraintSummary> = emptyList()
 )
 
-data class Pinned(
+data class ActionMetadata(
   val at: Instant,
   val by: String?,
   val comment: String?
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactVeto.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactVeto.kt
@@ -3,20 +3,27 @@ package com.netflix.spinnaker.keel.core.api
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import java.time.Instant
 
-data class EnvironmentArtifactPin(
+data class EnvironmentArtifactVeto(
   val targetEnvironment: String,
   val reference: String,
   val version: String,
-  val pinnedBy: String?,
+  val vetoedBy: String?,
   val comment: String?
 )
 
-data class PinnedEnvironment(
+data class VetoedArtifact(
   val deliveryConfigName: String,
   val targetEnvironment: String,
   val artifact: DeliveryArtifact,
   val version: String,
-  val pinnedBy: String?,
-  val pinnedAt: Instant?,
+  val vetoedBy: String?,
+  val vetoedAt: Instant?,
   val comment: String?
+)
+
+data class EnvironmentArtifactVetoes(
+  val deliveryConfigName: String,
+  val targetEnvironment: String,
+  val artifact: DeliveryArtifact,
+  val versions: MutableSet<String>
 )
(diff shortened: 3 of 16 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -121,9 +121,14 @@ class ResourceActuator(
...
-                    artifact = artifact,
-                    version = artifactVersion,
-                    targetEnvironment = environment)
+                    veto = EnvironmentArtifactVeto(
+                      reference = artifact.reference,
+                      version = artifactVersion,
+                      targetEnvironment = environment,
+                      vetoedBy = "Spinnaker",
+                      comment = "Automatically vetoed because deployment of this version failed."
+                    )
```
emjburns, line 130: This creates an automated user + reason for automatic rollbacks. Thoughts on wording here?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
```diff
@@ -58,12 +58,14 @@ data class ArtifactSummaryInEnvironment(
...
-  val pinned: Pinned? = null,
+  val pinned: ActionMetadata? = null,
+  val isVetoed: Boolean = false,
+  val vetoed: ActionMetadata? = null,
   val statefulConstraints: List<StatefulConstraintSummary> = emptyList(),
   val statelessConstraints: List<StatelessConstraintSummary> = emptyList()
 )
 
-data class Pinned(
+data class ActionMetadata(
```
emjburns, line 68: Reusing across pin + veto

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -952,15 +974,28 @@ class SqlArtifactRepository(
...
         .orderBy(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT.desc())
         .fetchOne { (version, deployedAt, promotionStatus, replacedBy, replacedAt) ->
-          val pinned: Pinned? = jooq
-            .select(ENVIRONMENT_ARTIFACT_PIN.PINNED_BY, ENVIRONMENT_ARTIFACT_PIN.PINNED_AT, ENVIRONMENT_ARTIFACT_PIN.COMMENT)
-            .from(ENVIRONMENT_ARTIFACT_PIN)
-            .where(ENVIRONMENT_ARTIFACT_PIN.ENVIRONMENT_UID.eq(deliveryConfig.getUidFor(environmentName)))
-            .and(ENVIRONMENT_ARTIFACT_PIN.ARTIFACT_UID.eq(artifact.uid))
-            .and(ENVIRONMENT_ARTIFACT_PIN.ARTIFACT_VERSION.eq(version))
-            .fetchOne { (pinnedBy, pinnedAt, comment) ->
-              Pinned(at = pinnedAt.toInstant(UTC), by = pinnedBy, comment = comment)
```
emjburns, line 962: I'm worried that gathering this data will make the api slower. I'm not going to fix that in this PR, I think we need a broader discussion about how to cache state here or generate api view that get get refreshed.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -198,8 +199,10 @@ class ApplicationController(
     @PathVariable("application") application: String,
     @RequestBody veto: EnvironmentArtifactVeto
   ) {
-
-    applicationService.markAsVetoedIn(application, veto, true)
+    val succeeded = applicationService.markAsVetoedIn(user, application, veto, true)
+    if (!succeeded) {
+      throw InvalidVetoException(application, veto.targetEnvironment, veto.reference, veto.version)
+    }
```
emjburns, line 205: a 'mark as bad' could fail if that version is pinned. in this case, we throw an error with a good user message.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
gal-yardeni: Maybe "this version is vetoed automatically as we tried to deploy it multiple times and it failed"?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
gal-yardeni: Love it!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactVeto.kt
```diff
@@ -0,0 +1,29 @@
+package com.netflix.spinnaker.keel.core.api
+
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import java.time.Instant
+
```
gal-yardeni, line 5: I'm confused about the 3 classes here -- can you please add some descriptions?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/InvalidVetoException.kt
```diff
@@ -0,0 +1,7 @@
+package com.netflix.spinnaker.keel.exceptions
+
+import org.omg.CORBA.UserException
+
+class InvalidVetoException(application: String, environment: String, reference: String, version: String) :
+  UserException("Unable to veto artifact version $version with reference $reference in environment " +
+  "$environment and application $application. Is this version pinned?")
```
gal-yardeni, line 7: Is pinning the only reason to get this exception?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
gal-yardeni: is it possible to throw that exception within `pplicationService.markAsVetoedIn`? it'll make the controller cleaner.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -86,9 +87,9 @@ class ExceptionHandler(
     return ApiError(CONFLICT, e)
   }
 
-  @ExceptionHandler(ValidationException::class)
+  @ExceptionHandler(ValidationException::class, InvalidVetoException::class)
   @ResponseStatus(BAD_REQUEST)
-  fun onInvalidDeliveryConfig(e: ValidationException): ApiError {
+  fun onInvalidDeliveryConfig(e: Exception): ApiError {
```
gal-yardeni, line 92: why was this changed?

emjburns: So that we could do the same thing for multiple exception classes (previously this was only handling the `ValidationException` as a `BAD_REQUEST`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactVeto.kt
emjburns: I can for the first one, i'll remove the middle one (i didn't need it but forgot about it), and the third is just a lift from another class to not be in with the pin stuff

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/InvalidVetoException.kt
emjburns: I think there might be another way (like something fails?) but this is the most probable way, and as far as I could tell the only way the 'veto' function would return false (false == veto didn't work). That's why I kept this like asking a question

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
emjburns: sure

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
```diff
@@ -58,12 +58,14 @@ data class ArtifactSummaryInEnvironment(
   val replacedAt: Instant? = null,
   val replacedBy: String? = null,
   val isPinned: Boolean = false,
-  val pinned: Pinned? = null,
+  val pinned: ActionMetadata? = null,
+  val isVetoed: Boolean = false,
```
lorin, line 62: Is it ever possible for `isVetoed` to be true and `vetoed` to be null? If not, I would just change `isVetoed` to a property that checks if `vetoed` is null. (Same with `pinned`)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
lorin: I like @gal-yardeni's idea of communicating that there were multiple attempts.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -102,18 +103,17 @@ class ApplicationService(
...
-      artifact = artifact,
-      version = veto.version,
-      targetEnvironment = veto.targetEnvironment,
+      veto = veto.copy(vetoedBy = user),
       force = force
     )
+    if (!succeeded) {
+      throw InvalidVetoException(application, veto.targetEnvironment, veto.reference, veto.version)
+    }
+    // TODO: publish ArtifactVetoedEvent
```
lorin, line 116: Is this `TODO` something that should be done for this PR, or can it come later?

emjburns: no, it's for later. the best kind of todo :D 

We need to surface events other than resource events, and we don't do that now. So I don't want to add any events

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -574,33 +576,46 @@ class SqlArtifactRepository(
...
+                ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.eq(veto.version)
               )
               .execute()
 
+            txn.insertInto(ENVIRONMENT_ARTIFACT_VETO)
+              .set(ENVIRONMENT_ARTIFACT_VETO.ENVIRONMENT_UID, envUid)
+              .set(ENVIRONMENT_ARTIFACT_VETO.ARTIFACT_UID, artUid)
+              .set(ENVIRONMENT_ARTIFACT_VETO.ARTIFACT_VERSION, veto.version)
+              .set(ENVIRONMENT_ARTIFACT_VETO.VETOED_AT, clock.timestamp())
+              .set(ENVIRONMENT_ARTIFACT_VETO.VETOED_BY, veto.vetoedBy)
```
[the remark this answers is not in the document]
emjburns: I didn't want to have a separate class just to capture that this is not null after it's submitted. I think the risk of this throwing an exception is pretty low. thanks for pointing it out though!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
emjburns: That's actually what we do when we create this class. I think you're right, it would be better to do it as a property

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: Updated to indicate that there were multiple deployments

PR #1272, opened 2020-06-04
fix(validation): disallow invalid artifact refs in delivery configs

### LINKED ISSUE #1271: keel incorrectly validates delivery configs with invalid references
A delivery config that has an invalid artifacts reference (i.e., refers to an artifact that is defined in the delivery config) isn't valid, but it passes keel validation.

Here's an example of a delivery config with an invalid reference (`this-ref-does-not-exist-in-this-config`) that passes validation:

```yaml
name: broken-example
application: foo
serviceAccount: keel@spinnaker
artifacts:
  - name: org/thing
    type: docker
    reference: org/thing
environments:
  - name: testing
    resources:
      - kind: titus/cluster@v1
        spec:
          moniker:
            app: foo
          locations:
            account: testing
            regions:
              - name: us-east-1
          container:
            reference: this-ref-does-not-exist-in-this-config
```

lorin:
Add a validation check for artifact references that refefrences artifacts that aren't defined in the delivery config.

Fixes #1271

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/NonexistentArtifactReferenceException.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/NonexistentArtifactReferenceException.kt
@@ -5,4 +5,3 @@ class NonexistentArtifactReferenceException(
 ) : ValidationException(
   "Config uses an artifact reference that does not correspond to any artifacts: $reference."
 )
-
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/NonexistentArtifactReferenceException.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/NonexistentArtifactReferenceException.kt
@@ -1,0 +1,8 @@
+package com.netflix.spinnaker.keel.exceptions
+
+class NonexistentArtifactReferenceException(
+  reference: String
+) : ValidationException(
+  "Config uses an artifact reference that does not correspond to any artifacts: $reference."
+)
+
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/NonexistentArtifactReferenceException.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/NonexistentArtifactReferenceException.kt
@@ -1,7 +1,8 @@
 package com.netflix.spinnaker.keel.exceptions
 
 class NonexistentArtifactReferenceException(
-  reference: String
+  invalidReference: String,
+  validReferences: List<String>
 ) : ValidationException(
-  "Config uses an artifact reference that does not correspond to any artifacts: $reference."
+  "Config uses an artifact reference ($invalidReference) that does not correspond to any artifacts. Valid references are: $validReferences."
 )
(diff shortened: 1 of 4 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/NonexistentArtifactReferenceException.kt
```diff
@@ -0,0 +1,7 @@
+package com.netflix.spinnaker.keel.exceptions
+
+class NonexistentArtifactReferenceException(
+  reference: String
+) : ValidationException(
+  "Config uses an artifact reference that does not correspond to any artifacts: $reference."
```
emjburns, line 6: It would be nice to also show existing references in the config to make the mistake easy to spot

lorin: @emjburns Do you mean augment the error message to include the list of artifact references that are properly defined?

emjburns: yes! exaclty.

emjburns: "Config uses an artifact reference ($reference) that does not correspond to any artifacts. Valid references are: $validReferences." ?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidator.kt
```diff
@@ -83,5 +86,20 @@ class DeliveryConfigValidator {
...
+    /**
+     * check: all referenced artifacts exist
+     */
+    config.environments.forEach { environment ->
+      environment.resources.forEach { resource ->
+        (resource.spec as? ArtifactReferenceProvider)
+          ?.artifactReference
+          ?.also {
+            if (!refs.contains(it)) {
+              throw NonexistentArtifactReferenceException(it)
```
luispollo, line 99: Consider fixing that name as you're touching this:
```suggestion
              throw NonExistentArtifactReferenceException(it)
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/exceptions/NonexistentArtifactReferenceException.kt
lorin: Addressed in b5d56502ade93d447134e5eea90406a5f3559183

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidator.kt
```diff
@@ -83,5 +86,20 @@ class DeliveryConfigValidator {
...
+      environment.resources.forEach { resource ->
+        (resource.spec as? ArtifactReferenceProvider)
+          ?.artifactReference
+          ?.also {
+            if (!refs.contains(it)) {
+              throw NonexistentArtifactReferenceException(it, refs)
+            }
+          }
+      }
+    }
```
robfletcher, line 103: I know in the vast majority of cases there will only be a single artifact, but is there a danger of us getting into whack-a-mole problems with this class in general? Fix one problem, a different exception pops up, fix that, yet another pops up. I wonder if there's a way we can group all the validation results at once (in a future PR -- I'm not suggesting doing it here).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/validators/DeliveryConfigValidator.kt
lorin: I ended up renaming this to `InvalidArtifactReferenceException` in a653ee5afe06df3356d24b31a683950182761c1f

PR #1196, opened 2020-05-19
feat(plugins): Expose `ConstraintEvaluator` as an extension point

robzienert:
Moves all of the core constraints code into `keel-api`.

Two new interfaces were added to help this effort:

- `com.netflix.spinnaker.keel.api.support.EventPublisher` - A subset interface
  of Spring's `ApplicationEventPublisher`. This is injected into `ConstraintEvaluator`
  classes as `SpringEventPublishedBridge`, which just delegates to Spring.
- `com.netflix.spinnaker.keel.api.constraints.ConstraintRepository` - This is
  just a subset of the god `KeelRepository` interface. Implemented by
  `ConstraintRepositoryBridge`, which just delegates to the `KeelRepository`
  implementation.

This also introduces the `ULID` library as an api dependency of `keel-api`.

I also did some non-functional cleanup: Single class per file rule.

⚠️ I didn't bother documenting all of the interfaces introduced to `keel-api` that already existed - it looks like the MD team hasn't really been doing many (any?) code docs -- I'm leaving this as an exercise for them to take on. If external contributions to Keel are desired, having docs on everything in `keel-api` is going to be needed rather than relying on tribal knowledge of contracts and behaviors.

PR #1275, opened 2020-06-05
fix(serialization): serialize constraint attributes

emjburns:
Fixes the problem that we can't correctly serialize the constraint attributes.

There are some just formatting changes in here, and I honestly don't know why. Maybe spotless doesn't have an opinion on them and my intellij does?


keel-api-jackson/src/main/kotlin/com/netflix/spinnaker/keel/json/KeelApiModule.kt
```diff
@@ -60,7 +62,9 @@ object KeelApiModule : SimpleModule("Keel API") {
         NamedType<DebianArtifact>(ArtifactType.deb.name),
         NamedType<DockerArtifact>(ArtifactType.docker.name),
         NamedType<DockerVersioningStrategy>(ArtifactType.docker.name),
-        NamedType<DebianSemVerVersioningStrategy>(ArtifactType.deb.name)
+        NamedType<DebianSemVerVersioningStrategy>(ArtifactType.deb.name),
+        NamedType<PipelineConstraintStateAttributes>("pipeline"),
+        NamedType<CanaryConstraintAttributes>("canary")
```
emjburns, line 67: This is how we know what the subtypes are, but in order to do this they had to be defined in the keel api module. This felt fine to me. I didn't do this in the `KeelConfigurationFinalizer` like resource spec things because the subtypes aren't part of a component right now, and I would have had to make them part of a component, then collect those in a list, then iterate over them like we do [here](https://github.com/spinnaker/keel/blob/master/keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt#L32).

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/constraints/CanaryConstraintAttributes.kt
```diff
@@ -0,0 +1,7 @@
+package com.netflix.spinnaker.keel.api.constraints
```
emjburns, line 1: all of these are lifted from one file and shifted into their own files in the api module

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintEvaluator.kt
```diff
@@ -3,24 +3,27 @@ package com.netflix.spinnaker.keel.constraints
...
-import java.time.Clock
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.runBlocking
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
+import java.time.Clock
 
 /**
  * An environment promotion constraint that deploys control (baseline) and experiment (canary)
```
emjburns, line 29: ugh thanks spotless for 0 changes except formatting here :(

keel-api-jackson/src/main/kotlin/com/netflix/spinnaker/keel/json/KeelApiModule.kt
luispollo: Hmm, so I thought the whole point of what @robzienert had done was to make it so that `keel-api` was the only dependency a plugin developer would need to write a new `ConstraintEvaluator`, but the moment you add implementation-specific types in `keel-api`, that signals to me that a developer outside our team would need to similarly make PRs for keel to add their constraint implementations... But I'll defer to Rob on this.

emjburns: So, agree 100% with your conclusion that this is not the best way to do things. At the moment I can't figure out how to do this in a different way (I feel like there is one though). I explored [this helper](https://github.com/spinnaker/kork/blob/master/kork-core/src/main/java/com/netflix/spinnaker/kork/jackson/ObjectMapperSubtypeConfigurer.java) but couldn't find an example of it in use.

luispollo: Good find! There are a few examples in clouddriver and kayenta: https://github.com/search?q=org%3Aspinnaker+ObjectMapperSubtypeConfigurer+-repo%3Aspinnaker%2Fkork&type=Code

luispollo: I wonder if a better option might be to revert #1196 temporarily, and ask @robzienert for guidance on how best to address the sub-type registration in a Spinnaker plugin world.

luispollo: Another idea: `ConstraintEvaluators`s are `@Component`s. We could add a method like `getConstraintStateAttributeType()` to the interface, which would return the sub-type of `ConstraintStateAttribute` used by that evaluator. Then, in `KeelConfigFinalizer`, you could iterate the list of evaluators and register those sub-types.

robfletcher: Sorry for weighing in late here, but isn't the right approach to have `ConstraintEvaluator` (implementations of which will be `@Component`) require implementations to specify sub-types of `Constraint` and `ConstraintStateAttribute` much like `ResourceHandler` implementations have to specify the `ResourceSpec` type? We can then easily register these on app start.

robfletcher: Yeah, in fact `ConstraintEvaluator` already does this for the constraint types themselves with the `supportedType` property. Could we not just do the same thing for the attributes?

luispollo: Yup, same suggestion I gave above, although you did a much better job describing it. 🙂

keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
```diff
@@ -50,15 +51,27 @@ class KeelConfigurationFinalizer(
       }
   }
```
emjburns, line 53: This auto-registers all stateful constraint attrs

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/constraints/StatefulConstraintEvaluator.kt
```diff
@@ -24,10 +24,16 @@ import com.netflix.spinnaker.keel.api.events.ConstraintStateChanged
 /**
  * TODO: Docs.
  */
```
emjburns, line 26: stateful constraint evaluators now have another property to surface the type of the attributes, so that we can auto register it for jackson

keel-api-jackson/src/main/kotlin/com/netflix/spinnaker/keel/json/KeelApiModule.kt
emjburns: Alright, I fixed it 🤞

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/constraints/StatefulConstraintEvaluator.kt
```diff
@@ -24,10 +24,16 @@ import com.netflix.spinnaker.keel.api.events.ConstraintStateChanged
...
-abstract class StatefulConstraintEvaluator<T : Constraint>(
+abstract class StatefulConstraintEvaluator<T : Constraint, A : ConstraintStateAttributes>(
```
gal-yardeni, line 27: Can you please explain this change? I don't really understand what is happening here

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluator.kt
```diff
@@ -18,9 +20,10 @@ class ManualJudgementConstraintEvaluator(
   repository: ConstraintRepository,
   private val clock: Clock,
   override val eventPublisher: EventPublisher
-) : StatefulConstraintEvaluator<ManualJudgementConstraint>(repository) {
+) : StatefulConstraintEvaluator<ManualJudgementConstraint, DefaultConstraintAttributes>(repository) {
```
gal-yardeni, line 23: So for every stateful constraint, we would have to pass a `constraintAttribute` object, or use this default object?

emjburns: Ya, every evaluator needs to tell us what type the attributes will be. I created this default one because not every stateful constraint needs attributes

PR #1283, opened 2020-06-11
ensure event timestamps have subsecond precision

### LINKED ISSUE #1158: Event rows have redundant,  possibly inconsistent data for timestamps
The `event` table has a discrete `timestamp` column, but all event objects also have a `timestamp` field which gets encoded in the JSON blob stored in the `json` column. This can and has led to inconsistencies and should be fixed.

robfletcher:
Because the `event.timestamp` column used to not store sub-second precision this change just ensures that the timestamps in the event JSON payload and the denormalized `timestamp` column are identical.

Fixes #1158

### DIFF
--- a/keel-sql/src/main/resources/db/changelog/20200611-fix-event-timestamps.yml
+++ b/keel-sql/src/main/resources/db/changelog/20200611-fix-event-timestamps.yml
@@ -1,0 +1,8 @@
+databaseChangeLog:
+- changeSet:
+    id: fix-event-timestamps
+    author: fletch
+    changes:
+    - sql:
+        sql: |
+          update event set timestamp = timestamp(json_unquote(json->'$.timestamp'));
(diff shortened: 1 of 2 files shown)


luispollo:
@robfletcher Thanks for doing this. How about the code change to ensure they're identical, though? As I pointed out before, my understanding is that two different `Instant`s are being used to fill in the event JSON (during Jackson serialization) and the timestamp column.

robfletcher:
> @robfletcher Thanks for doing this. How about the code change to ensure they're identical, though? As I pointed out before, my understanding is that two different `Instant`s are being used to fill in the event JSON (during Jackson serialization) and the timestamp column.

Nope ```txn
        .insertInto(EVENT)
        .set(EVENT.UID, ULID().nextULID(event.timestamp.toEpochMilli()))
        .set(EVENT.SCOPE, event.scope.name)
        .set(EVENT.REF, ref)
        .set(EVENT.TIMESTAMP, event.timestamp.toTimestamp())
        .set(EVENT.JSON, objectMapper.writeValueAsString(event))
        .execute()```

luispollo:
Ah, nice! Thanks for clarifying!

PR #1284, opened 2020-06-11
Derive resource metadata instead of storing it

### LINKED ISSUE #1097: Change in service account doesn't propagate to resources
If you change a service account on a delivery config, i don't think it propagates down into the resources. @michaelgalloway keeps hitting this

robfletcher:
Original intent here was a way to keep service account consistent between a delivery config and its resources. However, I think it makes sense to have all metadata be derived.

Fixes #1097.

### DIFF
--- a/keel-sql/src/main/resources/db/changelog/20200611-create-resource-metadata-view.yml
+++ b/keel-sql/src/main/resources/db/changelog/20200611-create-resource-metadata-view.yml
@@ -1,0 +1,30 @@
+databaseChangeLog:
+- changeSet:
+    id: create-resource-metadata-view
+    author: fletch
+    changes:
+    - createView:
+        viewName: resource_with_metadata
+        selectQuery: |
+          select
+            resource.uid,
+            resource.id,
+            resource.application,
+            resource.kind,
+            json_object(
+              'uid', resource.uid,
+              'id', resource.id,
+              'application', resource.application,
+              'environment', environment.uid,
+              'environmentName', environment.name,
+              'deliveryConfig', delivery_config.uid,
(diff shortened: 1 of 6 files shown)


keel-sql/src/main/resources/db/changelog/20200611-create-resource-metadata-view.yml
```diff
@@ -0,0 +1,30 @@
...
+              'serviceAccount', delivery_config.service_account
+            ) as metadata,
+            resource.spec
+          from resource
+          left outer join environment_resource on resource.uid = environment_resource.resource_uid
+          left outer join environment on environment.uid = environment_resource.environment_uid
+          left outer join delivery_config on delivery_config.uid = environment.delivery_config_uid
+    - dropColumn:
+        tableName: resource
+        columnName: metadata
```
luispollo, line 30: Instead of dropping it altogether, I think it might be prudent to allow resource-specific metadata to coexist with the metadata inherited from the delivery config and environment. I'd propose we merge the two together when reading. That would also avoid yet another discrepancy from the k8s format, which I get the feeling we're kinda regretting having diverged from.

robfletcher: The reason I went for dropping it altogether is that we currently don't use it, and I thought it might be confusing for us if we don't realize that not all the metadata is in the table. This doesn't have any impact on the in/out format of resources.

luispollo: Ah, right. Makes sense!

PR #1286, opened 2020-06-11
feat(artifacts): Introduce ArtifactPublisher plugin interface

luispollo:
Introduces a new keel plugin interface, `ArtifactPublisher`, that allows keel to support new types of artifacts via plugins. The reason to introduce this change now is that we want/need to enable other teams at Netflix to contribute plugins so that we can support other types of artifacts without having to implement it ourselves within keel. (We will soon need to do the same with resolvers and resource handlers, but that's a story for another time. 🙂)

This PR refactors the existing artifact detection support in keel, which is currently done exclusively via the `POST /artifacts/events` API (called by echo and igor to notify keel of Debian and Docker artifacts, respectively), such that keel itself implements the new plugin interface. This change was pretty straightforward. However, we also depend on finding the latest available version of an artifact when a delivery config is first registered into the system, and that turned out to require quite a lot of changes because we baked assumptions about Docker and Debian all over the code base... 😞 

💡  I know the number of modified files can be scary, but the vast majority of them have only trivial changes due to moving types between modules/packages. My suggestion for reviewing is to focus on the files where I've added comments.

TODO:
- [ ] Write tests for `DebianArtifactPublisher` and `DockerArtifactPublisher`.


keel-api-jackson/src/main/kotlin/com/netflix/spinnaker/keel/json/KeelApiModule.kt
```diff
@@ -55,13 +50,6 @@ object KeelApiModule : SimpleModule("Keel API") {
       setMixInAnnotations<Monikered, MonikeredMixin>()
       setMixInAnnotations<SubnetAwareRegionSpec, SubnetAwareRegionSpecMixin>()
       setMixInAnnotations<ConstraintState, ConstraintStateMixin>()
-
-      registerSubtypes(
```
luispollo, line 59: These are now registered in `KeelConfigurationFinalizer`.

keel-api-jackson/src/test/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifactTests.kt
```diff
@@ -83,4 +84,10 @@ private data class Fixture(
   val json: String
 ) {
   val mapper = configuredObjectMapper()
+    .also {
+      it.registerSubtypes(
```
luispollo, line 88: A small annoyance resulting from this change is that, because support for docker and debian is now provided by plugins, you have to register the artifact sub-types when you need them.

keel-api/keel-api.gradle.kts
```diff
@@ -12,6 +12,7 @@ plugins {
 dependencies {
   api("com.netflix.spinnaker.kork:kork-plugins-api:${property("korkVersion")}")
   api("com.netflix.spinnaker.kork:kork-exceptions:${property("korkVersion")}")
+  api("com.netflix.spinnaker.kork:kork-artifacts:${property("korkVersion")}")
```
luispollo, line 15: Required for the `Artifact` model used in `ArtifactEvent`. Unsure whether it's worth pulling the entire library or if we should just create a duplicate in keel.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/constraints/StatefulConstraintEvaluator.kt
```diff
@@ -20,6 +20,7 @@ import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.events.ConstraintStateChanged
+import com.netflix.spinnaker.keel.api.plugins.ConstraintEvaluator
```
luispollo, line 23: I moved `ConstraintEvaluator` to the plugins package so it sits along with all the other plugin interfaces.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/events/artifacts.kt
```diff
@@ -1,17 +1,27 @@
-package com.netflix.spinnaker.keel.events
+package com.netflix.spinnaker.keel.api.events
```
luispollo, line 1: Moved to `keel-api` as it's referenced in `ArtifactSupplier`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * The job of an [ArtifactSupplier] is to detect new versions of artifacts, using
+ * whatever mechanism they choose (e.g. they could receive events from another system,
+ * or poll an artifact repository for artifact versions), and notify keel via the [publishArtifact]
+ * method, so that the artifact versions can be persisted and evaluated for promotion.
+ *
+ * The default implementation of [publishArtifact] simply publishes the event via the [EventPublisher],
+ * and should *not* be overridden by implementors.
+ */
+interface ArtifactSupplier : SpinnakerExtensionPoint {
```
luispollo, line 18: The new plugin interface, hopefully self-explanatory from the code and documentation above.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/SupportedVersioningStrategy.kt
```diff
@@ -0,0 +1,13 @@
+package com.netflix.spinnaker.keel.api.plugins
+
+import com.netflix.spinnaker.keel.api.artifacts.VersioningStrategy
+
+/**
+ * Tuple of [VersioningStrategy] type name (e.g. "deb", "docker", etc.) and the
```
luispollo, line 6: I _think_ these names are right based on what I found, but would love confirmation.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -38,31 +40,21 @@ class ArtifactController(
   fun submitArtifact(@RequestBody echoArtifactEvent: EchoArtifactEvent) {
     echoArtifactEvent.payload.artifacts.forEach { artifact ->
       if (artifact.type.equals(deb.toString(), true) && artifact.isFromArtifactEvent()) {
-        publisher.publishEvent(ArtifactEvent(listOf(artifact), emptyMap()))
+        debianArtifactSupplier.publishArtifact(ArtifactEvent(listOf(artifact), emptyMap()))
```
luispollo, line 43: Here's where `ArtifactController` delegates publishing of the events coming in from echo and igor to the corresponding `ArtifactSuppliers`.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -38,31 +40,21 @@ class ArtifactController(
...
-        publisher.publishEvent(ArtifactEvent(listOf(artifact), emptyMap()))
+        dockerArtifactSupplier.publishArtifact(ArtifactEvent(listOf(artifact), emptyMap()))
       } else {
         log.debug("Ignoring artifact event with type {}: {}", artifact.type, artifact)
       }
     }
   }
 
-  @PostMapping(
-    path = ["/register"],
```
luispollo, line 51: Previously used to register artifacts before we had delivery configs. Artifacts are now registered via delivery configs.

keel-api-jackson/src/test/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifactTests.kt
emjburns: oof. Any way we could do this automatically in the `keel-test` module or something? Maybe that's a very ignorant question. But feels like we could have a `testConfiguredObjectMapper()` method that does that?

keel-api/keel-api.gradle.kts
robzienert: We can leave this as-is, but it's definitely not going to fly long-term. `kork-artifacts` brings in way too many transitive dependencies: https://github.com/spinnaker/kork/blob/master/kork-artifacts/kork-artifacts.gradle#L8-L9

Both of those deps are destined to ruin days.

emjburns: I think we should pull in this dependency, because it's how artifacts are represented in all of spinnaker. we don't want to miss updates to it and find out later we're breaking something I think.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
robzienert: nit: Would rename to `ArtifactPublisher`, since this interface isn't supplying anything (as "supplier" traditionally has implied in Java's std lib taxonomy).

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/SupportedVersioningStrategy.kt
emjburns: so i think that the "name" here is how to refer to the versioning strategy, not the artifact it goes with

keel-api/keel-api.gradle.kts
luispollo: My concern was that it may break the requirements for Spinnaker plugins. @robzienert @jonsie thoughts?

keel-api-jackson/src/test/kotlin/com/netflix/spinnaker/keel/api/DeliveryArtifactTests.kt
luispollo: Ah, no, that's a good idea. Let me work something out!

keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
```diff
@@ -63,6 +65,25 @@ class KeelConfigurationFinalizer(
       }
   }
```
emjburns, line 67: cool that we now have a way to register this stuff here instead of what we were doing before!

keel-api/keel-api.gradle.kts
robzienert: The way the module is written today is a definite no-go long-term. I agree that we should use those models, but the Jackson and Jinjava dependencies have got to go.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
robzienert: I still don't know why this project isn't using `ObjectMapperSubtypeConfigurer`... It's far less code to maintain from your side of things, and is a lot more flexible than this strategy.

keel-api/keel-api.gradle.kts
luispollo: I'm inclined to create a dupe of that one class we need in keel because it's not like we use shared model libraries for anything else anyway... @emjburns How do you feel about that?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
luispollo: Good point. Done!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/SupportedVersioningStrategy.kt
luispollo: Right, but this is how they were previously registered...
```kotlin
registerSubtypes(
  NamedType<DockerVersioningStrategy>(ArtifactType.docker.name),
  NamedType<DebianSemVerVersioningStrategy>(ArtifactType.deb.name)
)
```

luispollo: BTW, I think that `ArtifactType` enum will have to go away (we won't know artifact types until runtime).

keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
luispollo: I'll defer to @robfletcher on this one.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
gal-yardeni: can you explain a bit more about this obligation? why do we need 2 different artifact publishers?

luispollo: So the way I originally implemented this was to simply have `ArtifactController` implement `ArtifactPublisher` and register both Debian and Docker as supported artifact types. That works just fine, but I suspected I would get criticism for adding more "business logic" into a controller, and so my compromise was to move out the implementation of the interface and delegate. Once I did that, it seemed better to me to create two separate classes than a single `DebianAndDockerArtifactPublisher` class, but now that I think about it, I could have just named it `BasicArtifactPublisher` or `BuiltinArtifactPublisher` or something like that...

Thoughts?

jonsie: How about `DefaultArtifactPublisher` ?  That's the general pattern we use elsewhere in Spinnaker when we want to provide a default implementation of an interface.  In this case it makes sense because debs and docker images are most likely always going to be supported by default in Keel (no extensibility required).

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/DockerArtifactPublisher.kt
```diff
@@ -0,0 +1,28 @@
...
+  override val supportedArtifacts: List<SupportedArtifact<*>>
+    get() = listOf(
+      SupportedArtifact("docker", DockerArtifact::class.java)
+    )
+
+  override val supportedVersioningStrategies: List<SupportedVersioningStrategy<*>>
+    get() = listOf(
+      SupportedVersioningStrategy("docker", DockerVersioningStrategy::class.java)
+    )
+}
```
jonsie, line 28: Yeah, I think this and `DebianArtifactPublisher` can be squashed into one `DefaultArtifactPublisher`.

keel-api/keel-api.gradle.kts
jonsie: Thinking out loud here -- I think as a follow up item to this PR we could pull `Artifact` out to new module called `kork-api` that is basically vanilla Java + lombok.  We could then provide some Jackson Mixin classes to be used when configuring ObjectMapper to instruct Jackson how to handle `Artifact` fields -- not totally sure where we would do that configuration though.

luispollo: TBH, I didn’t realize `Artifact` even had Jackson annotations. I assumed it was just a POJO. At this point, I really think we should create a mirror class in keel, so we can remove the dependency on `kork-artifacts` and decouple the two conversations. If/when kork is refactored to be more plugin-friendly, then we can reintroduce the shared class. 

@emjburns do you object?

robfletcher: That sounds like the right approach. I'm concerned we might end up accidentally relying on some of those transitive dependencies.

emjburns: didn't realize kork-artifact had so many dependencies, and also didn't see @robzienert's comment when I commented (thanks github, why u so laggy?). Fine with creating a mirror class in keel, as long as it's quite flexible and doesn't barf when properties are added.

luispollo: Replaced use of kork's `Artifact` with a copycat called `SpinnakerArtifact` in keel-api.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/SupportedVersioningStrategy.kt
luispollo: `ArtifactType` is now simply an alias to `String`, since the artifact type name is really all we cared about.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -1,53 +1,32 @@
 package com.netflix.spinnaker.keel.api.artifacts
 
-import com.netflix.spinnaker.keel.api.artifacts.TagVersionStrategy.BRANCH_JOB_COMMIT_BY_JOB
-import com.netflix.spinnaker.keel.api.artifacts.TagVersionStrategy.SEMVER_JOB_COMMIT_BY_JOB
-import com.netflix.spinnaker.keel.api.artifacts.TagVersionStrategy.SEMVER_JOB_COMMIT_BY_SEMVER
-import com.netflix.spinnaker.keel.api.artifacts.TagVersionStrategy.SEMVER_TAG
```
luispollo, line 6: These have all been moved to `keel-artifact`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -1,53 +1,32 @@
...
-
-enum class ArtifactType {
-  deb, docker
-}
+typealias ArtifactType = String
```
luispollo, line 3: `ArtifactType` is now just an alias to `String` since we don't know artifact types until runtime and all we care about is the name.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -1,53 +1,32 @@
...
+abstract class DeliveryArtifact {
   abstract val name: String
   abstract val type: ArtifactType
   abstract val versioningStrategy: VersioningStrategy
   abstract val reference: String // friendly reference to use within a delivery config
   abstract val deliveryConfigName: String? // the delivery config this artifact is a part of
-  override fun toString() = "$type:$name" + if (reference != null) " (ref: $reference)" else ""
-}
-
-data class DebianArtifact(
```
luispollo, line 25: Moved to `keel-artifact`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -1,53 +1,32 @@
...
-  override val deliveryConfigName: String? = null,
-  override val reference: String = name,
-  val vmOptions: VirtualMachineOptions,
-  val statuses: Set<ArtifactStatus> = emptySet(),
-  override val versioningStrategy: VersioningStrategy = DebianSemVerVersioningStrategy
-) : DeliveryArtifact() {
-  override val type = ArtifactType.deb
-}
-
-data class DockerArtifact(
```
luispollo, line 36: Moved to `keel-artifact`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/VersioningStrategy.kt
```diff
@@ -3,7 +3,7 @@ package com.netflix.spinnaker.keel.api.artifacts
 /**
  * Strategy for how to sort versions of artifacts.
  */
-sealed class VersioningStrategy
+abstract class VersioningStrategy
```
luispollo, line 6: So that it can be implemented by plugins.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactPublisher.kt
```diff
@@ -0,0 +1,53 @@
...
+  suspend fun getLatestArtifact(artifact: DeliveryArtifact): SpinnakerArtifact?
+
+  // Utility functions to help convert a kork Artifact to a DeliveryArtifact
+  // Ideally these should be "static" and still overridden in sub-classes, but there's no such
+  // concept in Kotlin (or any other JVM language for that matter).
+  fun getFullVersionString(artifact: SpinnakerArtifact): String = artifact.version
+  fun getVersionDisplayName(artifact: SpinnakerArtifact): String = artifact.version
+  fun getReleaseStatus(artifact: SpinnakerArtifact): ArtifactStatus? = null
+  fun getBuildMetadata(artifact: SpinnakerArtifact, versioningStrategy: VersioningStrategy): BuildMetadata? = null
+  fun getGitMetadata(artifact: SpinnakerArtifact, versioningStrategy: VersioningStrategy): GitMetadata? = null
```
luispollo, line 48: This is the best I could come up with so far to extract behavior that was previously baked into the algorithms in `ArtifactListener` into interfaces that the artifact plugins can implement.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactPublisher.kt
```diff
@@ -0,0 +1,53 @@
...
+  fun publishArtifact(artifactEvent: ArtifactEvent) =
+    eventPublisher.publishEvent(artifactEvent)
+
+  /**
+   * Returns the latest available version for the given [DeliveryArtifact], represented
+   * as a [SpinnakerArtifact].
+   *
+   * This function may interact with external systems to retrieve artifact information as needed.
+   */
+  suspend fun getLatestArtifact(artifact: DeliveryArtifact): SpinnakerArtifact?
```
luispollo, line 39: This was the biggest addition to the interface. In addition to notifying keel when new artifact versions are available, plugins must also be able to return the latest version of the artifact. This is used when a new artifact is first registered into the system (i.e. a new delivery config with an artifact is created). Due to this change, I went back-and-forth on changing the name of the interface back to `ArtifactSupplier`, but couldn't come to a conclusion. Let me know if you have better suggestions for a name.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactPublisher.kt
```diff
@@ -0,0 +1,53 @@
...
+
+  // Utility functions to help convert a kork Artifact to a DeliveryArtifact
```
luispollo, line 41: Note to self: document each funtion properly.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactPublisher.kt
```diff
@@ -0,0 +1,53 @@
...
+ * whatever mechanism they choose (e.g. they could receive events from another system,
+ * or poll an artifact repository for artifact versions), and notify keel via the [publishArtifact]
+ * method, so that the artifact versions can be persisted and evaluated for promotion.
+ *
+ * The default implementation of [publishArtifact] simply publishes the event via the [EventPublisher],
+ * and should *not* be overridden by implementors.
+ */
+interface ArtifactPublisher<T : DeliveryArtifact> : SpinnakerExtensionPoint {
+  val eventPublisher: EventPublisher
+  val supportedArtifact: SupportedArtifact<T>
```
luispollo, line 27: I changed my mind on this and each `ArtifactPublisher` now supports a single type of artifact.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -57,23 +49,15 @@ class ArtifactListener(
...
-          when (artifact.type()) {
-            deb -> {
-              version = "${artifact.name}-${artifact.version}"
-              status = debStatus(artifact)
-            }
-            docker -> {
-              version = artifact.version
-            }
-          }
+        if (repository.isRegistered(artifact.name, artifact.artifactType)) {
```
luispollo, line 52: `ArtifactListener` now has no artifact type-specific logic -- every piece of information needed is retrieved from the `ArtifactPublisher` matching the artifact type via the newly defined functions on the interface.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -146,59 +133,14 @@ class ArtifactListener(
   private fun getLatestStoredVersion(artifact: DeliveryArtifact): String? =
     repository.artifactVersions(artifact).sortedWith(artifact.versioningStrategy.comparator).firstOrNull()
 
-  private suspend fun getLatestDeb(artifact: DebianArtifact): String? =
```
luispollo, line 149: All the code from here on down has been moved to `DebianArtifactPublisher` and `DockerArtifactPublisher`.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/DebianArtifactPublisher.kt
```diff
@@ -0,0 +1,91 @@
...
+        val normalizedVersion = if (version.startsWith("${artifact.name}-")) {
+          version.removePrefix("${artifact.name}-")
+        } else {
+          version
+        }
+        artifactService.getArtifact(artifact.name, normalizedVersion)
+      }
+
+  override fun getFullVersionString(artifact: SpinnakerArtifact): String =
+    "${artifact.name}-${artifact.version}"
```
luispollo, line 51: I'm not sure why, but for Debian we prefix the version with the package name.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/DockerArtifactPublisher.kt
```diff
@@ -0,0 +1,104 @@
...
+      .findDockerTagsForImage("*", artifact.name, serviceAccount)
+      .distinct()
+      .sortedWith(artifact.versioningStrategy.comparator)
+      .firstOrNull()
+
+    return if (latestTag != null) {
+      // TODO: do we need to look in a specific account? if so, then DockerArtifact should have an
+      // `account` property. In retrospect, it feels weird that we specify images without a repository
+      // URL (from which we could infer the account), or an account.
+      cloudDriverService.findDockerImages(account = "*", repository = artifact.name, tag = latestTag)
```
luispollo, line 58: @emjburns Would love your thoughts on this. It feels to me like we should add an `account` property in `DockerArtifact`. I understand we resolve the account automatically when finding Docker images to deploy, but the way we do it is to look at the Titus cluster spec, which means we can't really make sense of a Docker artifact without a cluster. I think that should not be the case.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/VersioningStrategyComparators.kt
```diff
@@ -31,9 +31,11 @@ import org.slf4j.LoggerFactory
 import org.springframework.util.comparator.NullSafeComparator
 
 val VersioningStrategy.comparator: Comparator<String>
+  // TODO: this needs to support versioning strategies provided by artifact plugins
```
luispollo, line 34: I guess we could just make it a property in `VersioningStrategy`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -192,10 +193,9 @@ class CombinedRepository(
 
   private fun Set<DeliveryArtifact>.transform(deliveryConfigName: String) =
     map { artifact ->
-      when (artifact) {
-        is DockerArtifact -> artifact.copy(deliveryConfigName = deliveryConfigName)
-        is DebianArtifact -> artifact.copy(deliveryConfigName = deliveryConfigName)
-      }
+      val asMap: MutableMap<String, Any?> = objectMapper.convertValue(artifact)
+      asMap["deliveryConfigName"] = deliveryConfigName
+      objectMapper.convertValue(asMap, artifact.javaClass)
```
luispollo, line 198: This is a bit hacky, but at least we don't have to make assumptions about the types. 🤷‍♂️

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -319,61 +319,32 @@ class ApplicationService(
...
+  ): ArtifactVersionSummary {
+    val artifactPublisher = artifactPublishers.supporting(artifact.type)
+    val spinArtifact = artifact.toSpinnakerArtifact(version)
+    return ArtifactVersionSummary(
+      version = version,
+      environments = environments,
+      displayName = artifactPublisher.getVersionDisplayName(spinArtifact),
+      build = artifactPublisher.getBuildMetadata(spinArtifact, artifact.versioningStrategy),
+      git = artifactPublisher.getGitMetadata(spinArtifact, artifact.versioningStrategy)
+    )
```
luispollo, line 331: This is much simplified now that we can delegate to `ArtifactPublisher`s.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionCheckerTests.kt
```diff
@@ -0,0 +1,567 @@
+package com.netflix.spinnaker.keel.actuation
```
luispollo, line 1: Re-adding this file was my mistake. I got confused with the conflict. Will remove.

keel-sql/keel-sql.gradle.kts
```diff
@@ -45,6 +45,7 @@ tasks.getByName<Delete>("clean") {
 
 dependencies {
   implementation(project(":keel-core"))
+  implementation(project(":keel-artifact"))
```
luispollo, line 48: This is very undesirable but is here because we make assumptions about Debian and Docker even when reading from the database.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
```diff
@@ -24,18 +24,20 @@ import com.fasterxml.jackson.module.kotlin.readValue
...
 import com.netflix.spinnaker.keel.serialization.configuredObjectMapper
 
 private val objectMapper: ObjectMapper = configuredObjectMapper()
 
 /**
  * A helper function to construct the proper artifact type from the serialized json.
+ * FIXME: this needs to go away in favor of [ArtifactPublisher] functions to keep
+ * artifact contracts generic.
  */
 fun mapToArtifact(
```
luispollo, line 42: @emjburns Would love some context on why we needed custom deserialization for artifacts in the first place.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/DeliveryArtifactModelConverter.kt
```diff
@@ -17,7 +18,7 @@ class DeliveryArtifactModelConverter : SubtypesModelConverter<DeliveryArtifact>(
 
   // TODO: can we just work this out automatically?
```
luispollo, line 19: Now we can. I just need to wire in the `ArtifactPublisher`s.

luispollo:
@robzienert @jonsie We have a couple tests that start up Spring with the full config of the app. They started failing after I added `@Import(PluginsAutoConfiguration::class)`, with an NPE instantiating the PF4J `VersionManager`. Any ideas?

lorin:
To make the PRs smaller and therefore easier to review, I'd recommend breaking it up into three PRs by doing:

1. A PR that just adds new code, but doesn't wire it up to existing code yet. (This is a behavioral no-op, since no new code is being invoked yet)
2. A PR that just moves classes around, but doesn't change any behavior. (This is a behavioral no-op, since it's just moving)).
3. A PR that wires up the new code.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/DebianArtifactPublisher.kt
emjburns: There's no real reason. It's a leftover inconsistency from when we only had debian packages and didn't think about saving the whole thing vs extracting just the version.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/DockerArtifactPublisher.kt
emjburns: At netflix we have images defaulted to only one place: 
- amis are _only_ in the `test` account, then shared with other accounts
- docker images are in both the test and prod docker registries
So for our artifacts, we don't need an account specified. That's why we didn't put it in earlier.

keel-sql/keel-sql.gradle.kts
emjburns: How else would we deserialize things?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
emjburns: IIRC because otherwise we couldn't read the artifacts from the database into their proper types.

luispollo:
Thanks for the suggestion, @lorin. I'll close this and open separate PRs.

keel-sql/keel-sql.gradle.kts
luispollo: Using polymorphism support in Jackson like we do for all other types with sub-classes.

Issue #1254, opened 2020-06-01
Pinning: does not "work" for unpin user flow (not a bug, a missing feature)


emjburns:
Pinning has two primary use cases:
(1) roll forward or roll back to a specific version because you are remediating a problem (primary)
(2) "freeze" an environment (i.e. something funky is going on I don't want this environment to change). (secondary)

Right now, the primary use case _does not work_ completely. You can never successfully unpin: we will always try to roll out a previously approved version, not the new one which you've created that fixes whatever was broken. This is a big problem.

We need to fix this in a way that allows users to unpin and know that they are going to deploy the new and non broken version.

@erikmunson is proposing leaning towards the more risky side - there might be a small percentage of cases where this is unsafe (maybe when a canary is run and that canary is bad), but the bigger problem is that you cannot actually complete the unpin successfully without some manual tomfoolery. Keel will always pick an older version to unpin to (not the newest one, because it won't have been approved yet).

erikmunson:
It was pointed out in another communication forum that this issue does not explicitly capture the currently proposed solution to the problem: at this time we think the best immediate term solution to this issue might be to make the _promotion_ process (evaluating constraints + approving versions whose constraints are met) continue running as normal when an environment is pinned. Today the behavior is that all constraint evaluation and version approval is completely halted while pinned, which leads to the un-usable behavior we currently have when folks want to unpin.

Some additional context on impact I shared in our internal Slack:
> just to really underscore the situation though — and clarify the customer impact — in my view the choices are:
> - wait and leave things as they are, ~99.5% of the time pinning is completely broken and requires us manually intervening — **even pausing does not solve this**
> - start evaluating things the way people expect, ~0.05% of the time people need to pause temporarily to respond to incidents and unpause when things have settled, same as we've told them to do for any unexpected situations

emjburns:
Group decision is to make constraints evaluate while pinned, but not deploy the version until unpinned. I'll be working on this shortly

PR #1291, opened 2020-06-12
feat(pin): eval constraints while pinned, but don't approve anything

emjburns:
fixes https://github.com/spinnaker/keel/issues/1254

omg this was so easy to do now that i re-wrote everything.

Note: this pr removes the old environment promotion checker tests as well, in favor of the new ones. I didn't rename the new ones because it makes the pr super hard to understand.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -40,24 +40,22 @@ class EnvironmentPromotionChecker(
...
-              constraintRunner.checkEnvironment(
-                EnvironmentContext(
-                  deliveryConfig = deliveryConfig,
-                  environment = environment,
-                  artifact = artifact,
-                  versions = versions,
-                  vetoedVersions = (vetoedArtifacts[envPinKey(environment.name, artifact)]?.versions) ?: emptySet()
-                )
-              )
+              constraintRunner.checkEnvironment(envContext)
```
gal-yardeni, line 58: I think you can remove the `else` here, as constraintRunner.checkEnvironment(envContext) is being called in both cases?

emjburns: The important thing is in the pin case, it's _only_ checking the environments and not doing anything else. So the `else` needs to stay

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/NewEnvironmentPromotionCheckerTests.kt
```diff
@@ -201,19 +201,24 @@ internal class NewEnvironmentPromotionCheckerTests : JUnit5Minutests {
...
-          test("the artifact is approved") {
+          test("the pinned artifact is approved") {
             verify(exactly = 1) {
               repository.approveVersionFor(deliveryConfig, artifact, "1.0", environment.name)
             }
           }
+
+          test("queued constraints aren't looked at") {
+            verify(exactly = 0) {
+              environmentConstraintRunner.checkStatelessConstraints(any(), any(), any(), any())
```
lorin, line 218: This is a very small nit, but it looks to me like this test is testing two distinct things, stateless constraints and queued constraints. I would pull out the stateless constraint check into a separate test for clarity, e.g.:

``` kotlin
          test("stateless constraints aren't looked at") {
            verify(exactly = 0) {
              environmentConstraintRunner.checkStatelessConstraints(any(), any(), any(), any())
            }
          }

```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
gal-yardeni: I guess reading code via GH is hard lol. thanks for clarifying :)

PR #1293, opened 2020-06-13
Support all protocols / all ports ingress rules

robfletcher:
Clouddriver can return a security group with ingress rules defined like this:  

```json
"inboundRules": [
    {
        "class": "com.netflix.spinnaker.clouddriver.model.securitygroups.IpRangeRule",
        "range": {
            "ip": "100.64.0.0",
            "cidr": "/10"
        },
        "protocol": "-1",
        "portRanges": [
            {}
        ],
        "description": "0AIr0"
    }
]
```

A protocol of `-1` means "all protocols" and an empty port range object means "all ports".

This change supports those in `ec2/security-group@v1` resources with the notation 

```yaml
protocol: ALL
portRange: ALL
```

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
@@ -26,6 +26,7 @@ import com.netflix.spinnaker.keel.api.actuation.TaskLauncher
 import com.netflix.spinnaker.keel.api.ec2.AllPorts
 import com.netflix.spinnaker.keel.api.ec2.CidrRule
 import com.netflix.spinnaker.keel.api.ec2.CrossAccountReferenceRule
+import com.netflix.spinnaker.keel.api.ec2.IngressPorts
 import com.netflix.spinnaker.keel.api.ec2.PortRange
 import com.netflix.spinnaker.keel.api.ec2.ReferenceRule
 import com.netflix.spinnaker.keel.api.ec2.SecurityGroup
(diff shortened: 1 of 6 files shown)


lorin:
@robfletcher Text of the PR says
> an empty list of port ranges means "all ports".

but the example provided shows a `portRanges` list that contains one element, which is an empty object:

```
"portRanges": [
            {}
        ]
```

Which one is it?

robfletcher:
Good spot. That is a real Clouddriver response. I’ll update the code to look for that.

robfletcher:
@lorin fixed

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -306,6 +308,14 @@ class SecurityGroupHandler(
         .toSet()
     )
 
+  private fun SecurityGroupModel.SecurityGroupRulePortRange.toPortRange(): IngressPorts =
+    (startPort to endPort).let { (start, end) ->
+      if (start == null || end == null) AllPorts else PortRange(start, end)
```
lorin, line 313: I'm impressed that the Kotlin compiler is smart enough to know that the arguments to `PortRange` are not nullable.

robfletcher: That's why I had to use that funky `let` construction. It wouldn't do it directly with the properties of the `SecurityGroupRulePortRange` class because it can't tell if there's anything side-effecty going on, or another thread messing with the object concurrently.

robfletcher: Not just for null handling either: https://kotlinlang.org/docs/reference/typecasts.html#smart-casts

PR #1303, opened 2020-06-16
fix(deploys): active server groups are not always healthy

emjburns:
We had a longstanding incorrect assumption that an active asg was healthy. this is not true. 

This pr takes advantage of the `instanceCounts` object on all server group responses:
```
instanceCounts": {
        "total": 1,
        "up": 1,
        "down": 0,
        "unknown": 0,
        "outOfService": 0,
        "starting": 0
      }
```

Now, we will only publish a "successfully deployed" event if the server groups are all healthy (`total == up`).

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
@@ -7,7 +7,7 @@ import com.netflix.spinnaker.keel.clouddriver.CloudDriverCache
 import com.netflix.spinnaker.keel.core.api.Capacity
 import com.netflix.spinnaker.kork.exceptions.SystemException
 
-data class ServerGroupCollection<T : BaseServerGroup> (
+data class ServerGroupCollection<T : BaseServerGroup>(
   val accountName: String,
   val serverGroups: Set<T>
 )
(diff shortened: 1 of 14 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -26,6 +26,20 @@ interface BaseServerGroup {
...
+  val total: Int,
+  val up: Int,
+  val down: Int,
+  val unknown: Int,
+  val outOfService: Int,
+  val starting: Int
+) {
+  // active asg is healthy if all instances are up
+  fun isHealthy(): Boolean =
+    up == total
```
lorin, line 42: A couple of things I worry about for using 100% instances up as the criteria for healthy:

* large clusters, where they may be stragglers that take longer to come up, which slows down deployments when gating on the asg being healthy
* one bad instance that doesn't come up (e.g., some transient or hardware problem unrelated to the build) can block the entire asg

I think using something less strict than 100% would be more robust, although I don't have a good intuition about what it should be (>50%? 75%? Configurable with a sensible default?).

How does orca decide when to disable the old asg when doing red-black deployments?

emjburns: Good point. For deployments there's an option : `Consider deployment successful when X percent of instances are healthy.`. 

I think that we should allow this to be overridden. In practice, I don't see it overridden very much, but if it is it's usually set to around 95%. 

I'll add an issue to work that in. I'd say we only want to do that if the cluster is larger than like 10 instances. It's not an easy change for this PR to add a configurable option

PR #1304, opened 2020-06-16
feat(artifacts): Introduce ArtifactPublisher plugin interface (1 of 3)

luispollo:
First part of breaking up #1286 into more manageable reviews.

This PR introduces only new code, which will be latent until part 3 of this series when I plan to refactor keel's existing artifact detection implementation to use the new plugin interface.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/SpinnakerArtifact.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/SpinnakerArtifact.kt
@@ -1,0 +1,20 @@
+package com.netflix.spinnaker.keel.api.artifacts
+
+/**
+ * An immutable data class that mirrors com.netflix.spinnaker.kork.artifacts.model.Artifact, but without
+ * all the Jackson baggage. One notable difference from the kork counterpart is that this class enforces
+ * non-nullability of a few key fields without which it doesn't make sense for an artifact to exist in
+ * Managed Delivery terms.
+ */
+data class SpinnakerArtifact(
+  val name: String,
+  val type: String,
+  val reference: String,
+  val version: String,
+  val customKind: Boolean? = null,
+  val location: String? = null,
+  val artifactAccount: String? = null,
+  val provenance: String? = null,
+  val uuid: String? = null,
+  val metadata: Map<String, Any?> = emptyMap()
+)
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/events/artifacts.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/events/artifacts.kt
@@ -1,0 +1,27 @@
+package com.netflix.spinnaker.keel.api.events
+
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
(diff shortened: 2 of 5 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/SpinnakerArtifact.kt
```diff
@@ -0,0 +1,20 @@
+package com.netflix.spinnaker.keel.api.artifacts
+
+/**
+ * An immutable data class that mirrors com.netflix.spinnaker.kork.artifacts.model.Artifact, but without
+ * all the Jackson baggage. One notable difference from the kork counterpart is that this class enforces
+ * non-nullability of a few key fields without which it doesn't make sense for an artifact to exist in
+ * Managed Delivery terms.
+ */
+data class SpinnakerArtifact(
```
luispollo, line 9: I've debated with myself about the name of this class. My second candidate would be `VersionedArtifact`, to convey the idea that the most valuable piece of information we're looking for in it is the artifact version (along with the type and name to correlate with delivery configs). Let me know what you prefer.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/events/artifacts.kt
```diff
@@ -0,0 +1,27 @@
...
+data class ArtifactEvent(
+  val artifacts: List<SpinnakerArtifact>,
+  val details: Map<String, Any>?
+)
+
+/**
+ * Event emitted with a new [DeliveryArtifact] is registered.
+ */
+data class ArtifactRegisteredEvent(
+  val artifact: DeliveryArtifact
```
gal-yardeni, line 19: What is the difference here between `SpinnakerArtifact` and `DeliveryArtifact`?

luispollo: `SpinnakerArtifact`, as the code docs in that class explain, is just a copycat of `com.netflix.spinnaker.core.models.Artifact`, i.e. the object that other Spinnaker services use to refer to a (versioned) artifact. `DeliveryArtifact` is a keel-specific model which describes high-level details about an artifact (as specified in the delivery config), so we can go find the actual artifacts in other Spinnaker services or other systems.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/SpinnakerArtifact.kt
lorin: I went back and forth on this as I read this PR. I think I prefer `VersionedArtifact`, but possibly even `VersionedSpinnakerArtifact` to differentiate it from, say, keel's `CompleteVersionedArtifact` .

emjburns: I think `SpinnakerArtifact` is fine, because it's a class that is pulled from the rest of spinnaker and is just used to receive data.

luispollo: Thank you both. I'll leave it as is for now, and we can revisit later.

Issue #1263, opened 2020-06-02
allowed-time constraint broken (ui + keel)


emjburns:
The allowed-times constraint seems to not be evaluating correctly

```yaml
---
windows:
- days: mon-fri
  hours: 7-16
type: allowed-times
```

This did not "pass" on Tuesday at 11 am. This also did not show in the UI at all. cc @erikmunson

gal-yardeni:
Adding to this bug, can be taken as a separate issue: we don't have validations today for `allowed-times` constraint when submitting the config. The only validation happening today is when evaluating the constraint.

emjburns:
Fun fact. The default timezone is UTC.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

osi:
I'm awaiting resolution of this to add the constraint.

@spinnakerbot remove-label to-be-closed
@spinnakerbot remove-label stale

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

osi:
@spinnakerbot remove-label stale

(still tracking the issue to know when to safely use the constraint)

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

osi:
@spinnakerbot remove-label stale

emjburns:
This problem has been fixed, but we don't have a UI for this (yet).

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

[...]

PR #1313, opened 2020-06-17
fix(allowed-times): use correct zoneId, switch default to PST

emjburns:
Addresses the keel side of https://github.com/spinnaker/keel/issues/1263

`UTC` (the short code) wasn't valid. You have to use the long code. Switched to PST while i was at it because... most of us are in California.


emjburns:
@lorin oops, I'll fix the file comment.

The property is actually not reading from the keel.yml, it's reading from a fast property. I agree that it's probably better to set a default of UTC for oss and then override it in a config, however I was being lazy

PR #1315, opened 2020-06-17
fix(deploy): we don't need to rollback on failure

emjburns:
The automatic rollback feature of deploys doesn't play nicely with keel's management. I'm switching this to default to off because keel will take care of rolling back if the deploy fails


luispollo:
We should probably get rid of this property altogether later. I can't think of a case where this would be used now that we have pinning.

emjburns:
@luispollo agreed! I'll create an issue for that - we probably need to make sure no one is actually using it, which won't be too hard.

PR #1318, opened 2020-06-17
Fix incorrect field names on ALB upsert task

robfletcher:


robfletcher:
> This is what we get for having the parsers configured to be lenient with unknown property names! 😝

In fairness, in this case that's in Orca.

Issue #1310, opened 2020-06-17
Remove deprecated ArtifactProvider implementations


luispollo:
The following have been deprecated for months and removed from all customer delivery configs. We should take a minute to remove the code from keel.
```
Warning:(11, 43) Kotlin: 'ArtifactImageProvider' is deprecated. Non-reference-based artifact providers are no longer supported.
Warning:(12, 43) Kotlin: 'JenkinsImageProvider' is deprecated. Non-reference-based artifact providers are no longer supported.
Warning:(16, 42) Kotlin: 'DigestProvider' is deprecated. Non-reference-based artifact providers are no longer supported.
Warning:(18, 42) Kotlin: 'VersionedTagProvider' is deprecated. Non-reference-based artifact providers are no longer supported.
```

emjburns:
This might require a bigger refactor than you're anticipating. 

The `DigestProvider` is used in titus code as a way to represent a fully specified container. There's probably a better way to do this (like whatever we do for ec2) but it's just not done yet. So, this is not as simple as just deleting these classes.

luispollo:
Thanks for the heads-up @emjburns. Removing the `good first issue` label. 🙂

luispollo:
@emjburns I removed the deprecation warning for `DigestProvider` per discussion above. I think we can remove the remaining classes after making sure no one is using those formats in their delivery configs.

gal-yardeni:
@luispollo didn't you handle this already?

luispollo:
@gal-yardeni Sorry for the delay. Somehow I missed the notification email. No, the deprecated classes are still there. You can see the deprecation warnings when you build.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

luispollo:
Closed by #1612

PR #1321, opened 2020-06-18
feat(artifacts): Introduce ArtifactSupplier plugin interface (3 of 3)

luispollo:
### Intro
Part 3 of 3 of introducing the ~`ArtifactPublisher`~ `ArtifactSupplier` plugin interface.

This PR:
- Introduces implementations of the plugin for Debian and Docker, respectively `DebianArtifactSupplier` and `DockerArtifactSupplier`, which encapsulate the implementation previously hard-coded in classes like `ArtifactListener`.
- Removes assumptions about `DeliveryArtifact` sub-types throughout the code (except in certain test classes) in favor of dynamically determining types from the corresponding `ArtifactSupplier`.
- Does the same as above for sub-types of `VersioningStrategy`.
- Changes `ArtifactType` from an enum to an alias to `String`, since we can't determine artifact types until runtime.
- Renames `SpinnakerArtifact` to `PublishedArtifact`, which I think is a much better name.
- Adds tests for the new plugins.

The PR touches many files, but most of them are trivial changes. Hopefully it's not too tiring to review.

### Testing
My hope is that the end result of this change has zero behavior impact. I've tested various APIs, with both Docker and Debian delivery configs, and watched the background processes for exceptions. I also looked at the database since I changed a bit of how serialization/deserialization of `DeliveryArtifact` sub-classes works. Things seem OK on my machine, but I definitely plan to do a much more thorough test in prestaging.

Additional tests:
- [x] Test updating an artifact in a delivery config? (like changing status)
- [x] Verify that versions are fetched on startup locally when you have no versions in the database
- [x] Test a whole deploy with titus submitting a delivery config and making sure the server group is actually deployed
- [x] Test a whole deploy with EC2 submitting a delivery config and making sure the server group is actually deployed
- [x] Test that different docker versioning strategies serialize ok
- [x] Test that fetching the delivery config back from the API works + has the correctly formatted information
- [x] Test upgrade path from `master` to branch (submit delivery config against `master` code then start the app from the branch and check everything works)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -1,20 +1,28 @@
...
+ * (i.e. the output of a build that is published to an artifact repository), but rather the high-level properties
+ * that allow keel and [ArtifactPublisher] plugins to find/process the actual artifacts.
+ */
 abstract class DeliveryArtifact {
   abstract val name: String
   abstract val type: ArtifactType
   abstract val versioningStrategy: VersioningStrategy
   abstract val reference: String // friendly reference to use within a delivery config
   abstract val deliveryConfigName: String? // the delivery config this artifact is a part of
+  open val statuses: Set<ArtifactStatus> = emptySet()
```
emjburns, line 26: Why did you move this to the top level object?

luispollo: Because I figured release status may not forever be a Debian-specific concept.

keel-api-jackson/src/main/kotlin/com/netflix/spinnaker/keel/json/KeelApiModule.kt
```diff
@@ -55,13 +50,6 @@ object KeelApiModule : SimpleModule("Keel API") {
       setMixInAnnotations<Monikered, MonikeredMixin>()
       setMixInAnnotations<SubnetAwareRegionSpec, SubnetAwareRegionSpecMixin>()
       setMixInAnnotations<ConstraintState, ConstraintStateMixin>()
-
-      registerSubtypes(
-        NamedType<DebianArtifact>(ArtifactType.deb.name),
-        NamedType<DockerArtifact>(ArtifactType.docker.name),
-        NamedType<DockerVersioningStrategy>(ArtifactType.docker.name),
-        NamedType<DebianSemVerVersioningStrategy>(ArtifactType.deb.name)
-      )
```
luispollo, line 64: These are now registered at startup in `KeelConfigurationFinalizer`.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -86,17 +69,28 @@ class ArtifactListener(
   @EventListener(ArtifactRegisteredEvent::class)
   fun onArtifactRegisteredEvent(event: ArtifactRegisteredEvent) {
     val artifact = event.artifact
+    val deliveryConfig = artifact.deliveryConfigName
+      ?.let { repository.getDeliveryConfig(it) }
+      ?: error("Delivery config name missing in artifact")
```
emjburns, line 74: Should we be throwing spinnaker exceptions instead of using `IllegalStateException` here?

emjburns: (comment applies for every use of `error()` in this PR

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -114,11 +108,12 @@ class ArtifactListener(
         repository.getAllArtifacts().forEach { artifact ->
           launch {
             val lastRecordedVersion = getLatestStoredVersion(artifact)
-            val latestVersion = when (artifact) {
-              is DebianArtifact -> getLatestDeb(artifact)?.let { "${artifact.name}-$it" }
-              is DockerArtifact -> getLatestDockerTag(artifact)
-              else -> error("Unrecognized artifact type: ${artifact.type}")
-            }
+            val artifactPublisher = artifactPublishers.supporting(artifact.type)
+            val deliveryConfig = artifact.deliveryConfigName
```
emjburns, line 112: This bit is repeated, it's probably worth pulling it into a function since it also involves error handling

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/DebianArtifactPublisher.kt
```diff
@@ -0,0 +1,88 @@
...
+ */
+@Component
+class DebianArtifactPublisher(
+  override val eventPublisher: EventPublisher,
+  private val artifactService: ArtifactService
+) : ArtifactPublisher<DebianArtifact> {
+  override val supportedArtifact = SupportedArtifact("deb", DebianArtifact::class.java)
+
+  override val supportedVersioningStrategies: List<SupportedVersioningStrategy<*>>
+    get() = listOf(
```
emjburns, line 33: this can just be `override val supportedVersioningStrategies = listOf(..)` just like the one above, no need to do the `get()`

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/ContainerProvider.kt
```diff
@@ -29,10 +29,8 @@ data class ReferenceProvider(
   val reference: String
 ) : ContainerProvider()
 
+// used in titus handler as a way to represent a fully specified container
 @JsonDeserialize(using = JsonDeserializer.None::class)
-@Deprecated("Non-reference-based artifact providers are no longer supported.",
```
luispollo, line 33: @emjburns I took this opportunity to remove this deprecation warning based on your comments in #1310.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/DockerArtifactPublisher.kt
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    val latestTag = cloudDriverService
+      .findDockerTagsForImage("*", artifact.name, deliveryConfig.serviceAccount)
+      .distinct()
+      .sortedWith(artifact.versioningStrategy.comparator)
+      .firstOrNull()
+
+    return if (latestTag != null) {
+      // TODO: do we need to look in a specific account? if so, then DockerArtifact should have an
```
emjburns, line 50: we're looking in all known repositories here I believe (`"*"`). I'd probably suggest changing this comment to indicate what we do?

keel-web/keel-web.gradle.kts
```diff
@@ -31,6 +31,7 @@ dependencies {
   implementation("org.springdoc:springdoc-openapi-webmvc-core:1.2.32")
   implementation("org.springdoc:springdoc-openapi-kotlin:1.2.32")
   implementation("org.apache.maven:maven-artifact:3.6.3")
+  implementation("com.netflix.spinnaker.kork:kork-plugins")
```
luispollo, line 34: This enables Keel to run plugins provided by other parties.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/DeliveryArtifactModelConverter.kt
```diff
@@ -1,23 +1,23 @@
...
-    DebianArtifact::class.java,
-    DockerArtifact::class.java
-  )
+class DeliveryArtifactModelConverter(
+  artifactPublishers: List<ArtifactPublisher<*>>
+) : SubtypesModelConverter<DeliveryArtifact>(DeliveryArtifact::class.java) {
 
   override val discriminator: String? = DeliveryArtifact::type.name
 
-  // TODO: can we just work this out automatically?
```
luispollo, line 18: Now we can. 🙂

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/Main.kt
```diff
@@ -41,7 +42,10 @@ private val DEFAULT_PROPS = mapOf(
     "com.netflix.spinnaker.keel"
   ]
 )
-@Import(PlatformComponents::class)
+@Import(
+  PlatformComponents::class,
+  PluginsAutoConfiguration::class
```
luispollo, line 47: Second bit required to enable Keel to run external plugins.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
luispollo: Good point. Let me create a sub-class of `SystemException` for these cases, because I couldn't find a good one in kork.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/DockerArtifactPublisher.kt
luispollo: OK, will do. I left the `TODO` there because this was a change from the previous implementation (we were just getting the version before), so I wasn't sure what needed to go here, but followed the example of the other call to find versions/tags. Hopefully it works OK.

luispollo:
Not sure why the tests are failing, they pass locally. Will rebase and check again. 🤔

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
emjburns: I think we could move those at that future time. I don't think status should be lifted at this moment because it doesn't apply to docker and will be a bit confusing

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
```diff
@@ -25,36 +27,40 @@ import io.mockk.verify
...
-    val event: KorkArtifactEvent,
+    val event: ArtifactPublishedEvent,
     val artifact: DeliveryArtifact,
     val repository: KeelRepository = mockk(relaxUnitFun = true),
     val artifactService: ArtifactService = mockk(relaxUnitFun = true),
     val clouddriverService: CloudDriverService = mockk(relaxUnitFun = true),
     val publisher: ApplicationEventPublisher = mockk(relaxUnitFun = true)
   ) {
-    val listener: ArtifactListener = ArtifactListener(repository, artifactService, clouddriverService, publisher)
+    private val eventBridge = SpringEventPublisherBridge(publisher)
```
emjburns, line 52: What is this eventBridge?

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
```diff
@@ -148,7 +162,7 @@ internal class ArtifactListenerTests : JUnit5Minutests {
 
     context("artifact is already registered") {
       before {
-        every { repository.isRegistered("fnord", deb) } returns true
+        every { repository.isRegistered("fnord", DEBIAN) } returns true
```
emjburns, line 165: Is this going to break serialization/deserialization anywhere?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -1,20 +1,28 @@
...
-
-  override fun toString() = name
-}
+typealias ArtifactType = String
 
+/**
+ * The release status of an artifact. This may not necessarily be applicable to all
+ * [DeliveryArtifact] sub-classes.
+ */
 enum class ArtifactStatus {
```
gal-yardeni, line 9: Can those statuses by applicable for other types other than `debians`?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
gal-yardeni: +1 to @emjburns's point.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
```diff
@@ -88,15 +73,24 @@ class ArtifactListener(
     val artifact = event.artifact
 
     if (repository.artifactVersions(artifact).isEmpty()) {
-      when (artifact) {
-        is DebianArtifact -> storeLatestDebVersion(artifact)
-        is DockerArtifact -> storeLatestDockerVersion(artifact)
+      val artifactPublisher = artifactPublishers.supporting(artifact.type)
+      val latestArtifact = runBlocking {
+        log.debug("Retrieving latest version of registered artifact {}", artifact)
+        artifactPublisher.getLatestArtifact(artifact.deliveryConfig, artifact)
```
gal-yardeni, line 79: something is not adding up for me -- how come `artifactPublisher` (which should only support `publish`), is getting the latest artifact?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactPublisher.kt
```diff
@@ -21,24 +26,63 @@ interface ArtifactPublisher<T : DeliveryArtifact> : SpinnakerExtensionPoint {
   val supportedVersioningStrategies: List<SupportedVersioningStrategy<*>>
```
gal-yardeni, line 27: I'm conflicting with the name. As this interface is not only responsible for publishing, I think we can rename it to `artifactHandler` or something silimar

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
luispollo: I added a `@JsonIgnore` to `DockerArtifact`, but will keep it in the base class.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
luispollo: This is needed because we cannot have a dependency on any Spring libraries in `keel-api`, since it's the library that plugin implementors will use (and it also causes problems with the PF4J stuff, I think -- @robzienert would know more). So, the solution Rob came up with, which I just replicated for artifact plugins, was to define a new interface called `EventPublisher` that looks-and-feels like Spring's `ApplicationEventPublisher`, which we can use in `keel-api` and plugins specifically. In `keel`, we just wire in this bean which  is a "bridge" between the generic interface and the Spring implementation.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
luispollo: Nope. Same values, just different variable names.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
luispollo: I don't know, so I'm trying to future-proof the interfaces in case we do.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
luispollo: I mentioned this in other PRs. I went back-and-forth on the name for this class. My first name was `ArtifactSource`, then `ArtifactSupplier`, and finally (after a suggestion from @robzienert), `ArtifactPublisher`. I don't mind changing it again. Which one of those do you think is best? (Note, `ArtifactProvider` is already taken elsewhere in keel).

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
emjburns: hmm, fair point. Thanks for adding a `@JsonIgnore` to the docker artifact.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
emjburns: Ah, interesting! Thanks.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
emjburns: I don't care about the names, just confirming that it serializes the same way

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -197,11 +198,9 @@ class CombinedRepository(
...
   private fun Set<DeliveryArtifact>.transform(deliveryConfigName: String) =
     map { artifact ->
-      when (artifact) {
-        is DockerArtifact -> artifact.copy(deliveryConfigName = deliveryConfigName)
-        is DebianArtifact -> artifact.copy(deliveryConfigName = deliveryConfigName)
-        else -> error("Unrecognized artifact type: ${artifact.type}")
-      }
+      val artifacAsMap: MutableMap<String, Any?> = objectMapper.convertValue(artifact)
+      artifacAsMap["deliveryConfigName"] = deliveryConfigName
+      objectMapper.convertValue(artifacAsMap, artifact.javaClass)
```
emjburns, line 203: Is this necessary because we no longer guarantee that all artifacts are data classes?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifact.kt
```diff
@@ -12,8 +10,9 @@ data class DebianArtifact(
   override val deliveryConfigName: String? = null,
   override val reference: String = name,
   val vmOptions: VirtualMachineOptions,
-  val statuses: Set<ArtifactStatus> = emptySet(),
+  override val statuses: Set<ArtifactStatus> = emptySet(),
   override val versioningStrategy: VersioningStrategy = DebianSemVerVersioningStrategy
 ) : DeliveryArtifact() {
-  override val type = ArtifactType.deb
+  override val type = DEBIAN
+  override fun toString(): String = super.toString()
```
emjburns, line 17: isn't this the default behavior?

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/ContainerProvider.kt
emjburns: Thanks

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
```diff
@@ -19,65 +19,36 @@ package com.netflix.spinnaker.keel.sql
...
-        DockerArtifact(
-          name = name,
-          tagVersionStrategy = objectMapper.convertValue(tagVersionStrategy),
-          captureGroupRegex = details["captureGroupRegex"]?.toString(),
-          reference = reference,
-          deliveryConfigName = deliveryConfigName
-        )
-      }
-    }
+    return objectMapper.convertValue(artifactAsMap, artifactPublisher.supportedArtifact.artifactClass)
```
emjburns, line 51: Does this correctly convert a `null` status to an empty list for docker artifacts?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListener.kt
emjburns: What about `ArtifactPlugin`? (I don't care too much. Just throwing out an idea)

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/dockerArtifacts.kt
```diff
@@ -0,0 +1 @@
+
```
emjburns, line 1: Should this file be deleted?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
luispollo: It won't be null, it just won't be present in the JSON because of the `@JsonIgnore`.

Example for Debian:
```json
{"statuses": ["SNAPSHOT"], "vmOptions": {"baseOs": "xenial", "regions": ["us-west-2"], "baseLabel": "RELEASE", "storeType": "EBS"}}
```

Example for Docker:
```json
{"tagVersionStrategy": "branch-job-commit-by-job"}
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifact.kt
luispollo: That's what I thought, but it turns out `data class` overrides `toString`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
luispollo: Exactly. It's not super pretty, but we can't make assumptions about how plugin developers will implement sub-classes of `DeliveryArtifact`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactPublisher.kt
luispollo: I went with `ArtifactSupplier` in the end! I think that's better!

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifact/ArtifactListenerTests.kt
luispollo: Yep, all good.

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/dockerArtifacts.kt
luispollo: Yes! Good spot, thanks!

Issue #1311, opened 2020-06-17
Add "healthy percentage" for clusters


emjburns:
As brought up by @lorin, we should expose an option for configuring what percentage of up instances == a "healthy" cluster. This is important because sometimes large clusters never have all instances up and healthy, and sometimes a hardware issue can prevent a cluster from being healthy. It's up to a user to decide if they want their health to be a little less sensitive 

https://github.com/spinnaker/keel/pull/1303#discussion_r441711236

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

Issue #1324, opened 2020-06-19
Support using cloud provider health for managed delivery


emjburns:
Right now if your instances never become healthy we won't mark something as deployed (because of https://github.com/spinnaker/keel/pull/1303)

We should support that.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

emjburns:
rob fixed this

PR #1329, opened 2020-06-22
feat(cluster): allow no health and healthy percentage

emjburns:
closes https://github.com/spinnaker/keel/issues/1311 and https://github.com/spinnaker/keel/issues/1324.

Recently we had an adopter that didn't use cloud provider health, and that meant that his environments view never showed "deployed". That is a v bad experience. 

This is a first stab at including some health information in the resources. I know that @robfletcher wants to pull this style info out of the resource entirely, but I think it's more important to support this now than it is to move these things out.

Thoughts on the implementation here?

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactInterfaces.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactInterfaces.kt
@@ -52,7 +52,27 @@ interface ArtifactReferenceProvider {
     }
 }
 
-interface ComputeResourceSpec : ResourceSpec, VersionedArtifactProvider, ArtifactReferenceProvider
+interface ComputeResourceSpec : ResourceSpec, VersionedArtifactProvider, ArtifactReferenceProvider {
+  val clusterHealth: ClusterHealth?
+}
+
+/**
+ * Contains user-configured health information about a server group
+ * [ignoreHealthForDeployments] if true, considers deployments healthy when instances are not in a
+ *  [InstanceCounts] down, out of service, or starting status. This is the setting for apps which
+ *  "only consider provider health"
+ * [healthyPercentage] the percentage of instances that need to be in the "up" status before
+ *  we consider a cluster fully deployed
+ */
+data class ClusterHealth(
+  val ignoreHealthForDeployments: Boolean = false,
+  val healthyPercentage: Double = 100.0
+) {
+  init {
+    require(healthyPercentage > 0.0) { "healthyPercentage must be > 0" }
+    require(healthyPercentage <= 100.0) { "healthyPercentage must be <= 100" }
+  }
+}
 
 /**
  * Simple container of the information defined in [ArtifactProvider] that ensures non-nullability of the fields.
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/HealthyHelper.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/HealthyHelper.kt
@@ -1,0 +1,29 @@
+package com.netflix.spinnaker.keel.clouddriver.model
+
+import com.netflix.spinnaker.keel.api.ClusterHealth
+
+/**
+ * Considers user defined health information to determine if a cluster is healthy or not
+ */
+fun isHealthy(clusterHealth: ClusterHealth?, instanceCounts: InstanceCounts?): Boolean {
+  // if we have no information about cluster health, default to not healthy
+  clusterHealth ?: return false
+  // if we have no information about the instance health, default to not healthy
+  instanceCounts ?: return false
+
+  val healthyCount: Double = if (clusterHealth.ignoreHealthForDeployments) {
+    // a deploy may not come up healthy, so use the total of up and and unknown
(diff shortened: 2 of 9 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactInterfaces.kt
```diff
@@ -52,7 +52,27 @@ interface ArtifactReferenceProvider {
...
+ * Contains user-configured health information about a server group
+ * [ignoreHealthForDeployments] if true, considers deployments healthy when instances are not in a
+ *  [InstanceCounts] down, out of service, or starting status. This is the setting for apps which
+ *  "only consider provider health"
+ * [healthyPercentage] the percentage of instances that need to be in the "up" status before
+ *  we consider a cluster fully deployed
+ */
+data class ClusterHealth(
+  val ignoreHealthForDeployments: Boolean = false,
+  val healthyPercentage: Double = 100.0
```
robfletcher, line 69: Do we really need to support fractional percentages?

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/HealthyHelper.kt
```diff
@@ -0,0 +1,29 @@
+package com.netflix.spinnaker.keel.clouddriver.model
```
robfletcher, line 1: This doesn't seem to be the logical package for this.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactInterfaces.kt
```diff
@@ -52,7 +52,27 @@ interface ArtifactReferenceProvider {
...
+/**
+ * Contains user-configured health information about a server group
+ * [ignoreHealthForDeployments] if true, considers deployments healthy when instances are not in a
+ *  [InstanceCounts] down, out of service, or starting status. This is the setting for apps which
+ *  "only consider provider health"
+ * [healthyPercentage] the percentage of instances that need to be in the "up" status before
+ *  we consider a cluster fully deployed
+ */
+data class ClusterHealth(
+  val ignoreHealthForDeployments: Boolean = false,
```
robfletcher, line 68: Could this name be clearer for users?

emjburns:
@robfletcher is your concern re: putting deployment info in a resource spec major enough that you wouldn't want to merge this PR?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactInterfaces.kt
luispollo: +1. In fact, this is even a bit paradoxical given the name of the class. Can't a user just set `healthyPercentage` to zero if they don't care about it?

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/HealthyHelperTests.kt
```diff
@@ -0,0 +1,79 @@
...
+    context("no info provided") {
+      test("returns false") {
+        expectThat(isHealthy(null, null)).isFalse()
+      }
+    }
+
+    context("info provided") {
+      test("100% and ignoring health, healthy case") {
+        expectThat(
+          isHealthy(ignoreHealth, InstanceCounts(100, 0, 0, 100, 0, 0))
```
luispollo, line 33: Please add the parameter names to these instantiations for readability.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
```diff
@@ -127,7 +128,8 @@ data class ClusterSpec(
   override val maxDiffCount: Int? = 2,
   @JsonIgnore
   // Once clusters go unhappy, only retry when the diff changes, or if manually unvetoed
-  override val unhappyWaitTime: Duration? = null
+  override val unhappyWaitTime: Duration? = null,
+  override val clusterHealth: ClusterHealth? = null
```
luispollo, line 132: I think it will be pretty confusing to have both a `health` and a `clusterHealth` keys at the same level in the spec. Any chance we could merge the two together?

robfletcher:
> @robfletcher is your concern re: putting deployment info in a resource spec major enough that you wouldn't want to merge this PR?

Not necessarily. Depends on our appetite for migrating specs later if we decide to tackle that.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt
robfletcher: That's a good point. Thinking about it further, should this be moved under the deploy strategy? It's fundamentally related to that (i.e. when do we consider the deploy successful)?

luispollo:
@emjburns Are you planning to get back to this, or should we close it?

emjburns:
oof, forgot about this. going to close.

PR #1331, opened 2020-06-23
fix(orcaError): catch errors from orca when failing to get info on tasks

gal-yardeni:
If we are trying to query orca for information about some task and orca returns an error, log it, and remove that task from `task_tracking` table.

Question: should we publish an event for that unknown task, something like `ResourceTaskUnknown`?

### DIFF
--- a/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
+++ b/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
@@ -43,36 +43,46 @@ class OrcaTaskMonitorAgent(
 
   // 1. Get active tasks from task tracking table
   // 2. For each task, call orca and ask for details
-  // 3. For each completed task, will emit an event for success/failure
+  // 3. For each completed task, will emit an event for success/failure and delete from the table
   override suspend fun invokeAgent() {
     coroutineScope {
       taskTrackingRepository.getTasks()
         .associate {
           it.subject to
             async {
-              orcaService.getOrchestrationExecution(it.id, DEFAULT_SERVICE_ACCOUNT)
-            }
-        }
-        .mapValues { it.value.await() }
-        .filterValues { it.status.isComplete() }
-        .map { (resourceId, taskDetails) ->
-          // only resource events are currently supported
-          if (resourceId.startsWith(RESOURCE.toString())) {
-            val id = resourceId.substringAfter(":")
-            try {
-              when (taskDetails.status.isSuccess()) {
-                true -> publisher.publishEvent(
-                  ResourceTaskSucceeded(
-                    resourceRepository.get(id), listOf(Task(taskDetails.id, taskDetails.name)), clock))
-                false -> publisher.publishEvent(
-                  ResourceTaskFailed(
-                    resourceRepository.get(id), taskDetails.execution.stages.getFailureMessage() ?: "", listOf(Task(taskDetails.id, taskDetails.name)), clock))
+              try {
+                orcaService.getOrchestrationExecution(it.id, DEFAULT_SERVICE_ACCOUNT)
+              } catch (e: Exception) {
+                log.warn("Exception ${e.message} has caught while calling orca to fetch status for execution id: ${it.id}" +
+                  " Possible reason: orca is saving info for 2000 tasks/app and this task is older.")
+                // when we get an exception from orca, we shouldn't try to get the status anymore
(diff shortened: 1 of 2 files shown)


keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
```diff
@@ -43,36 +43,46 @@ class OrcaTaskMonitorAgent(
...
             async {
-              orcaService.getOrchestrationExecution(it.id, DEFAULT_SERVICE_ACCOUNT)
-            }
+              try {
+                orcaService.getOrchestrationExecution(it.id, DEFAULT_SERVICE_ACCOUNT)
+              } catch (e: Exception) {
+                log.warn("Exception ${e.message} has caught while calling orca to fetch status for execution id: ${it.id}" +
+                  " Possible reason: orca is saving info for 2000 tasks/app and this task is older.")
+                // when we get an exception from orca, we shouldn't try to get the status anymore
+                taskTrackingRepository.delete(it.id)
```
luispollo, line 59: I think we should only delete the record from our table for very specific errors from orca. Perhaps only on `404`. The majority of other errors are simply indicating either a problem on our end (we messed up making the call, authenticating, etc.) or on orca's end (any `5xx` error). You can do that by catching a `RetrofitError` instead and looking at the HTTP status code.

lorin: In addition, when logging exceptions, it's good practice to pass the exception object as the last argument to the logging statement, so that the full stack trace appears in the log file. For example:

```
log.warn("Exception ${e.message} has caught while calling orca to fetch status for execution id: ${it.id}" +
                  " Possible reason: orca is saving info for 2000 tasks/app and this task is older.", e)
```

gal-yardeni: @luispollo I thought about it, but I'm not sure that it worth keeping this task record if orca returned _any_ kind of error, as it probably indicates we cannot determinate the task's status. Not quite sure what is the right thing to do here

lorin: @gal-yardeni I think that for any other type of error, we should just rethrow the exception. We don't really ever expect these to happen, so if they get thrown, we'll probably need to go in manually and fix something.

gal-yardeni: @lorin how should we differentiate between those cases? should we only delete the record from the DB if ocra returns 404, and for other status code we will throw an error?

lorin: @gal-yardeni Yes, that's exactly what I think we should do.

gal-yardeni: done!

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
```diff
@@ -43,36 +45,51 @@ class OrcaTaskMonitorAgent(
...
             async {
-              orcaService.getOrchestrationExecution(it.id, DEFAULT_SERVICE_ACCOUNT)
-            }
+              try {
+                orcaService.getOrchestrationExecution(it.id, DEFAULT_SERVICE_ACCOUNT)
+              } catch (e: HttpException) {
+                when (e.isNotFound) {
+                  true -> {
+                    log.warn("Exception ${e.message} has caught while calling orca to fetch status for execution id: ${it.id}" +
+                      " Possible reason: orca is saving info for 2000 tasks/app and this task is older. $e")
```
lorin, line 61: To show the stack trace in the log, the exception needs to be passed explicitly as a final argument, not interpolated into the string:

```suggestion
                      " Possible reason: orca is saving info for 2000 tasks/app and this task is older.", e)
```

gal-yardeni: @lorin thank you for the explanation! I didn't know that :)

PR #1340, opened 2020-06-24
refactor(db): refactor markAsVetoedIn

lorin:
Refactor the `SqlArtifactRepository.markAsVetoedIn` method to (hopefully) make the code easier to understand. Mostly this is pulling out chunks of code into separate methods. I also added some header docs with preconditions and postconditions.

This is preparation work for tackling #1086. I wanted to first submit a refactoring change that had no (intended) behavioral changes, to make it easier to review.

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
@@ -52,7 +52,6 @@ import java.time.ZoneOffset.UTC
 import javax.xml.bind.DatatypeConverter
 import org.jooq.DSLContext
 import org.jooq.Record1
-import org.jooq.Record2
 import org.jooq.Select
 import org.jooq.SelectConditionStep
 import org.jooq.impl.DSL
@@ -593,7 +592,7 @@ class SqlArtifactRepository(
     }
 
     return selectArtifactDetailsFromEnvironment(envUid, artUid, veto.version)
-      .fetchOne { (_, ref) ->
+      .fetchOne { ref ->
         ref?.let { reference ->
           /**
            * If there's a promotion reference, that means this artifact version was deployed as a result of
@@ -683,11 +682,9 @@ class SqlArtifactRepository(
     envUid: String,
     artUid: String,
     version: String
-  ): SelectConditionStep<Record2<String, String>> {
+  ): SelectConditionStep<Record1<String>> {
     return jooq
-      .select(
-        ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS,
-        ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_REFERENCE)
+      .select(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_REFERENCE)
       .from(ENVIRONMENT_ARTIFACT_VERSIONS)
       .where(
         ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(envUid),
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
@@ -592,7 +592,7 @@ class SqlArtifactRepository(
     }
 
     return selectPromotionReference(envUid, artUid, veto.version)
-      .fetchOne { (ref : String?) ->
+      .fetchOne { (ref: String?) ->
         ref?.let { reference ->
           /**
            * If there's a promotion reference, that means this artifact version was deployed as a result of
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
@@ -592,7 +592,7 @@ class SqlArtifactRepository(
       return false
     }
 
-    return getArtifactDetailsFromEnvironment(envUid, artUid, veto.version)
+    return selectArtifactDetailsFromEnvironment(envUid, artUid, veto.version)
       .fetchOne { (_, ref) ->
         ref?.let { reference ->
           /**
@@ -643,7 +643,11 @@ class SqlArtifactRepository(
       .execute()
   }
 
-  private fun DSLContext.addRecordToEnvironmentArtifactVetoTable(envUid: String, artUid: String, veto: EnvironmentArtifactVeto) {
+  private fun DSLContext.addRecordToEnvironmentArtifactVetoTable(
+    envUid: String,
+    artUid: String,
+    veto: EnvironmentArtifactVeto
+  ) {
     insertInto(ENVIRONMENT_ARTIFACT_VETO)
       .set(ENVIRONMENT_ARTIFACT_VETO.ENVIRONMENT_UID, envUid)
       .set(ENVIRONMENT_ARTIFACT_VETO.ARTIFACT_UID, artUid)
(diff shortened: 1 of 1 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -532,6 +534,45 @@ class SqlArtifactRepository(
...
+   *  @param force if true, record the version as vetoed even if it is marked as the promotion reference
+   *               (automated rollback target) of a previously vetoed version
+   *
+   *  @return true on success
+   *
+   * Preconditions for success:
+   *
+   *  1. [veto.version] is not currently pinned in the target environment
+   *  2. There exists a record R in the environment_artifacts_versions table such that:
+   *     a. R references the artifact version to veto and the target environment encoded in [veto]
```
lorin, line 551: Note that precondition 2 (specifically, 2a) is what's leading to #1086, which happens when there is no record in the `environment_artifacts_versions` table that corresponds to the vetoed version.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -565,53 +592,31 @@ class SqlArtifactRepository(
...
-      .fetchOne { (_, reference) ->
-        if (reference != null && !force) {
-          log.warn(
-            "Not vetoing artifact version as it appears to have already been an automated rollback target: " +
-              "deliveryConfig=${deliveryConfig.name}, " +
-              "environment=${veto.targetEnvironment}, " +
-              "artifactVersion=${veto.version}, " +
-              "priorVersionReference=$reference")
-          return@fetchOne false
+    return selectArtifactStatusAndPromotionReferenceInTargetEnvironment(envUid, artUid, veto.version)
```
gal-yardeni, line 595: omg that's a long name :) how about `getArtifactDetailsFromEnvironment`?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -532,6 +534,45 @@ class SqlArtifactRepository(
...
+   *
+   *  1. [veto.version] is not currently pinned in the target environment
+   *  2. There exists a record R in the environment_artifacts_versions table such that:
+   *     a. R references the artifact version to veto and the target environment encoded in [veto]
+   *     b. R.promotion_reference is NULL, or [force] is true
+   *
+   *  Note: 2b is a precondition to avoid cascading veto-triggered deployments. If R contains a promotion reference, then
+   *  [veto.version] was originally deployed as a result of another artifact version being vetoed (see postcondition 3a).
+   *
+   *
```
emjburns, line 557: Here's a dumb question. What's the difference between a "Pre" and "post" condition in this context?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -620,22 +625,83 @@ class SqlArtifactRepository(
...
+      .set(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_REFERENCE, prior ?: veto.version)
+      .where(
+        ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(envUid),
+        ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artUid),
+        ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.eq(veto.version)
+      )
+      .execute()
+  }
+
+  private fun selectArtifactStatusAndPromotionReferenceInTargetEnvironment(envUid: String, artUid: String, version: String): SelectConditionStep<Record2<String, String>> {
```
emjburns, line 673: nit: bump the vars and return type down on other lines so this isn't so long

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -565,53 +592,31 @@ class SqlArtifactRepository(
...
-        if (reference != null && !force) {
-          log.warn(
-            "Not vetoing artifact version as it appears to have already been an automated rollback target: " +
-              "deliveryConfig=${deliveryConfig.name}, " +
-              "environment=${veto.targetEnvironment}, " +
-              "artifactVersion=${veto.version}, " +
-              "priorVersionReference=$reference")
-          return@fetchOne false
+    return selectArtifactStatusAndPromotionReferenceInTargetEnvironment(envUid, artUid, veto.version)
+      .fetchOne { (_, ref) ->
```
emjburns, line 596: if we're not using status (as `(_,ref)` indicates), can we just not fetch it?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -565,53 +592,31 @@ class SqlArtifactRepository(
...
-              .set(ENVIRONMENT_ARTIFACT_VETO.ARTIFACT_VERSION, veto.version)
-              .set(ENVIRONMENT_ARTIFACT_VETO.VETOED_AT, clock.timestamp())
-              .set(ENVIRONMENT_ARTIFACT_VETO.VETOED_BY, veto.vetoedBy)
-              .set(ENVIRONMENT_ARTIFACT_VETO.COMMENT, veto.comment)
-              .onDuplicateKeyUpdate()
-              .set(ENVIRONMENT_ARTIFACT_VETO.VETOED_AT, clock.timestamp())
-              .set(ENVIRONMENT_ARTIFACT_VETO.VETOED_BY, veto.vetoedBy)
-              .set(ENVIRONMENT_ARTIFACT_VETO.COMMENT, veto.comment)
-              .execute()
+            txn.updateAsVetoedInEnvironmentArtifactVersionsTable(prior, veto, envUid, artUid)
```
emjburns, line 618: I like what you did here with these functions.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
lorin: Preconditions are predicates which have to be true _before_ you call the function. 

Postconditions are predicates which are supposed to be true _after_ the function completes. 

If the preconditions are true before the function is invoked, and the code is correct, then the postconditions will be true once the function completes.

Basically, if you call the function and the preconditions aren't true, then you can blame the person who wrote the code that calls the function.

If you call the function, and the preconditions are true, but the postconditions aren't true, then you can blame the person who implemented the function.

You can think of the argument types in a Kotlin function declaration as an example of a precondition, and the return type as an example of a postcondition. (In those cases, the preconditions and postconditions are enforced by the compiler).

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
lorin: The embarrassing truth is that I tried, but I couldn't get it to compile. Because `fetchOne` is an overloaded function, if you pass in only one argument, the Kotlin compiler doesn't know which method to use, so you need to disambguate it with explicit type information. I couldn't figure out the right incantation to get it to work. I may give it another shot.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
lorin: Fixed in 49a07702f

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
lorin: fixed in d6d12b1

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
lorin: It's working now. :shrug:

Fixed in 946564cce

lorin: Ugh, so it compiles with this change but it breaks the tests. If I can't figure. out why, I'll probably just revert to fetching the status and not using it.

PR #1343, opened 2020-06-24
chore(build): disable new Gradle flag until we can upgrade Nebula

robfletcher:


settings.gradle.kts
```diff
@@ -22,7 +22,7 @@ listOf("fiat", "kork").forEach { prj ->
   }
 }
 
-enableFeaturePreview("VERSION_ORDERING_V2")
+// enableFeaturePreview("VERSION_ORDERING_V2")
```
lorin, line 25: Can you add a comment here that mentions what we're waiting on to re-enable this?

robfletcher:
No longer relevant

PR #1349, opened 2020-07-02
Gradle 6.5.1

robfletcher:


cfieber:
I've been waiting until we adopt the version internally before updating all the Spinnaker projects so that we hopefully get less crazy composite build configuration

I've also been updating all the services when I bump our gradle versions as well as our plugin

Unless there is something you are specifically hoping to get out of this I'd suggest waiting

PR #1360, opened 2020-07-09
refactor(eureka): Bump kork; remove keel-eureka

robzienert:
Depends on https://github.com/spinnaker/kork/pull/710


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/activation/DiscoveryActivator.kt
```diff
@@ -1,13 +1,13 @@
...
-import com.netflix.spinnaker.keel.activation.ApplicationUp
-import com.netflix.spinnaker.kork.eureka.RemoteStatusChangedEvent
+import com.netflix.spinnaker.kork.discovery.InstanceStatus.UP
+import com.netflix.spinnaker.kork.discovery.RemoteStatusChangedEvent
 import org.slf4j.LoggerFactory
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.context.ApplicationListener
+import org.springframework.stereotype.Component
 
+@Component
```
robzienert, line 10: This class actually is redundant, but since Keel has defined its own objects for representing discovery up/down status, I decided not to change it. This class is probably unnecessary, though.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/activation/DiscoveryActivator.kt
```diff
@@ -1,13 +1,13 @@
...
 class DiscoveryActivator(
```
robfletcher, line 11: How does this continue to work? What happens when running locally and there's no Eureka?

robzienert: https://github.com/spinnaker/kork/blob/master/kork-core/src/main/java/com/netflix/spinnaker/kork/discovery/NoDiscoveryStatusPublisher.java

robzienert: All of the boilerplate code that does discovery stuff that used to be in services was promoted into kork.

robfletcher: Oh sweet. I think we just created our own event types to decouple things from directly depending on Eureka. I'll see if I can get rid of this class.

PR #1361, opened 2020-07-09
feat(docker): don't allow a/a/a format

emjburns:
Docker convention is to have images of the form `organization/image-name:tag`. Adding some validation to enforce that because it's tricky when images have a slash in the `image-name` format.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifact.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifact.kt
@@ -16,11 +16,20 @@ data class DockerArtifact(
   val captureGroupRegex: String? = null,
   override val versioningStrategy: VersioningStrategy = DockerVersioningStrategy(tagVersionStrategy, captureGroupRegex)
 ) : DeliveryArtifact() {
+  init {
+    require(name.count { it == '/' } <= 1) {
+      "Docker image name has more than one slash, which is not Docker convention. Please convert to `organization/image-name` format."
+    }
+  }
+
   override val type = DOCKER
+
   @JsonIgnore
   val organization: String = name.substringBefore('/')
+
   @JsonIgnore
   val image: String = name.substringAfter('/')
+
   @JsonIgnore
   override val statuses: Set<ArtifactStatus> = emptySet()
   override fun toString(): String = super.toString()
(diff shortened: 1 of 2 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifact.kt
```diff
@@ -16,11 +16,20 @@ data class DockerArtifact(
   val captureGroupRegex: String? = null,
   override val versioningStrategy: VersioningStrategy = DockerVersioningStrategy(tagVersionStrategy, captureGroupRegex)
 ) : DeliveryArtifact() {
+  init {
+    require(name.count { it == '/' } <= 1) {
+      "Docker image name has more than one slash, which is not Docker convention. Please convert to `organization/image-name` format."
```
gal-yardeni, line 21: does this throw an exception when reading the config? or after processing it?

emjburns: it throws an exception when it tries to turn the config into kotlin data classes

gal-yardeni: I see -- it might be worth to add this check to the `DeliveryConfigValidator`, as we can throw a bad request while validating the file. The artifact section is not being modified after submitting the config, correct?

emjburns: We do throw a bad request with this style. When you submit, you get a 400 that looks like:

```---
message: "Cannot construct instance of `com.netflix.spinnaker.keel.artifacts.DockerArtifact`,\
  \ problem: Docker image name has more than one slash, which is not Docker convention.\
  \ Please convert to `organization/image-name` format.\n at [Source: (PushbackInputStream);\
  \ line: 8, column: 1] (through reference chain: com.netflix.spinnaker.keel.core.api.SubmittedDeliveryConfig[\"\
  artifacts\"]->java.util.HashSet[0])"
details:
  error: "invalid_value"
  message: "Docker image name has more than one slash, which is not Docker convention.\
    \ Please convert to `organization/image-name` format."
  location:
    line: 8
    column: 1
  path:
    field: "artifacts"
  - type: "array"
    index: 0
  type: "invalid_value"
  pathExpression: ".artifacts[0]"
timestamp: "2020-07-09T18:37:44.764Z"
status: 400
error: "Bad Request"```

gal-yardeni: oh I see... ok, makes sense to me!

PR #1364, opened 2020-07-09
Fix environment level locations for clusters

robfletcher:
Using a custom deserializer for `ClusterSpec` broke assignment of `locations` based on the environment.

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/ClusterSpecDeserializer.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/ClusterSpecDeserializer.kt
@@ -1,11 +1,13 @@
 package com.netflix.spinnaker.keel.ec2.jackson
 
 import com.fasterxml.jackson.core.TreeNode
+import com.fasterxml.jackson.databind.BeanProperty
 import com.fasterxml.jackson.databind.DeserializationContext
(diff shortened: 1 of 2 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/ClusterSpecDeserializer.kt
```diff
@@ -35,6 +37,9 @@ class ClusterSpecDeserializer : StdNodeBasedDeserializer<ClusterSpec>(ClusterSpe
       )
     }
 
+  private fun DeserializationContext.findInjectableLocations() =
+    findInjectableValue("locations", BeanProperty.Bogus(), null) as SubnetAwareLocations?
```
luispollo, line 41: What does this do?

robfletcher: In `SubmittedEnvironmentDeserializer` we store `locations` (if it is defined there) in this injectable values thing. This retrieves it. For most resource specs we do this with `@JacksonInject` (see `LocatableMixin`). The problem @emjburns discovered yesterday where our delivery config was not being imported was because that annotation wasn't working for `ClusterSpec` resources because we're now using a custom deserializer.

Long term, I'd like to figure out how to not need to use the custom deserializer but that's going to take a little longer.

PR #1365, opened 2020-07-10
fix(delete): delete without serialization

emjburns:
Recently we had an incident where a rogue delivery config caused resource checks to drop off. To remediate the issue we wanted to delete the delivery config, but we couldn't because our current delete relies on serialization of resources.

This PR changes delete to delete without serialization so that it is more bester and able to be used when we have serialization problems

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
@@ -64,7 +64,7 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
     }
 
     fun delete() {
-      repository.deleteByApplication(deliveryConfig.application)
+      repository.deleteWithoutSerialization(deliveryConfig.application)
     }
 
     fun store() {
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
@@ -64,7 +64,7 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
     }
 
     fun delete() {
-      repository.deleteWithoutSerialization(deliveryConfig.application)
+      repository.delete(deliveryConfig.application)
     }
 
     fun store() {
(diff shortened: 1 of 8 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -79,66 +82,6 @@ class SqlDeliveryConfigRepository(
         }
     } ?: throw NoDeliveryConfigForApplication(application)
```
emjburns, line 84: removing this method, because I've re-written it to be in a transaction and to also delete any new bits that were added since this was written

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -248,13 +191,97 @@ class SqlDeliveryConfigRepository(
...
+    val application = sqlRetry.withRetry(READ) {
+      jooq
+        .select(DELIVERY_CONFIG.NAME)
+        .from(DELIVERY_CONFIG)
+        .where(DELIVERY_CONFIG.NAME.eq(name))
+        .fetchOne(DELIVERY_CONFIG.NAME)
+    }
+    deleteWithoutSerialization(application)
+  }
+
```
emjburns, line 204: Delete all things associated with a delivery config in a transaction.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -83,12 +83,8 @@ class DeliveryConfigController(
...
     and @authorizationSupport.hasServiceAccountAccess('DELIVERY_CONFIG', #name)"""
   )
-  fun delete(@PathVariable("name") name: String): DeliveryConfig {
-    val deliveryConfig = repository.getDeliveryConfig(name)
-    log.info("Deleting delivery config $name: $deliveryConfig")
-    repository.deleteDeliveryConfig(name)
-    return deliveryConfig
-  }
+  fun delete(@PathVariable("name") name: String) =
+    repository.deleteDeliveryConfigByName(name)
```
gal-yardeni, line 87: should we add the log message here?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -64,7 +64,7 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
     }
 
     fun delete() {
-      repository.deleteByApplication(deliveryConfig.application)
+      repository.deleteWithoutSerialization(deliveryConfig.application)
```
robfletcher, line 67: Shouldn't we just remove this side effect without necessarily calling it out in the name of the function?

emjburns: Sure, I'll rename. It was leftover from when I was going to leave both methods. Good call

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
emjburns: This method isn't really called - we use the `ApplicationController` mostly now - but I can add a log

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
emjburns: i have such a hard time remembering which is which with serialization/deserialization lol

PR #1374, opened 2020-07-13
chore(dependencies): replace kork dependencies with kork-runtime

cfieber:


cfieber:
no, the intention is that kork-core lives on as a place for shared utility code and compile time interfaces for things. we are splitting out more optional runtime only configurations into separate modules (and then including them back into the final service runtime classpath via kork-runtime)

PR #1377, opened 2020-07-14
Fix serialization of pinned and vetoed fields

robfletcher:
Fixes serialization of fields in artifact/environment summary. Instead of boolean fields and objects now there are just the objects.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
@@ -566,8 +566,8 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
           )
           expectThat(envArtifactSummary)
             .isNotNull()
-            .get { isPinned }
-            .isFalse()
+            .get { pinned }
+            .isNull()
         }
 
         context("once pinned") {
@@ -610,7 +610,6 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
             )
             expect {
               that(envArtifactSummary).isNotNull()
-              that(envArtifactSummary?.isPinned).isTrue()
               that(envArtifactSummary?.pinned).isEqualTo(ActionMetadata(by = pin1.pinnedBy, at = clock.instant(), comment = pin1.comment))
             }
           }
@@ -695,7 +694,6 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
         )
         expect {
           that(envArtifactSummary).isNotNull()
-          that(envArtifactSummary?.isVetoed).isTrue()
           that(envArtifactSummary?.vetoed).isEqualTo(ActionMetadata(by = "tester", at = clock.instant(), comment = "you bad"))
         }
       }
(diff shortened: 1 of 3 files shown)


keel-core/src/test/kotlin/com/netflix/spinnaker/keel/serialization/BooleanPropertySerializationTests.kt
```diff
@@ -0,0 +1,34 @@
...
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import java.time.Instant.now
+import strikt.api.expectThat
+import strikt.jackson.at
+import strikt.jackson.has
+import strikt.jackson.isObject
+import strikt.jackson.path
+
+class BooleanPropertySerializationTests : JUnit5Minutests {
```
gal-yardeni, line 15: Should this be renamed to something else? `Boolean` with no boolean is a bit confusing :)

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -610,7 +610,6 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
             )
             expect {
               that(envArtifactSummary).isNotNull()
-              that(envArtifactSummary?.isPinned).isTrue()
```
gal-yardeni, line 613: In order to keep the same functionality, should we add here (and below, if I understand the change correctly):
` that(envArtifactSummary?.pinned).isNotNull()` ?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/serialization/BooleanPropertySerializationTests.kt
robfletcher: Yes, I'll do that. Good point.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
robfletcher: The next line does an `isEqualTo` on the same field so testing it's not null explicitly doesn't really add anything.

robfletcher:
There was a change made in a recent version of Jackson so that those field definitions conflict. It started serializing `pinned` as `isPinned` and because there was already an `isPinned` property, one overrode the other. 

The change in behavior seems to have be prompted by this issue: https://github.com/FasterXML/jackson-module-kotlin/issues/80 unfortunately the "fix" was a breaking change. The only way I could successfully get _both_ properties to serialize was to annotate each _twice_ with `@JsonProperty` (once on the property itself, once on the getter, i.e. using `@get:JsonProperty` syntax). Anything less that 2 annotations on each property either results in one property overriding the other in the serialized output, or Jackson throwing an exception because it detects the ambiguity.

lorin:
> The only way I could successfully get _both_ properties to serialize was to annotate each _twice_ with `@JsonProperty` (once on the property itself, once on the getter, i.e. using `@get:JsonProperty` syntax). Anything less that 2 annotations on each property either results in one property overriding the other in the serialized output, or Jackson throwing an exception because it detects the ambiguity.

🤮

robfletcher:
@erikmunson does this look like the format that will un-fuck the environments view?

erikmunson:
Yep looks right to me - thanks for getting it resolved! I'm hopeful there is a way to prevent backend folks from having to deal with this property overriding business in the future 🤞

robfletcher:
> Yep looks right to me - thanks for getting it resolved! I'm hopeful there is a way to prevent backend folks from having to deal with this property overriding business in the future 🤞

Yeah, I think I'm going to raise an issue on the jackson-module-kotlin repo

PR #1378, opened 2020-07-14
Use a generated application column on event table

robfletcher:
@robzienert observed recently that we were doing some quite expensive queries selecting for `json -> '$.application'` on the event table. This PR introduces a generated column to avoid the need to do that.

See #1359

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
@@ -36,7 +36,6 @@ import java.time.Instant.EPOCH
 import java.time.LocalDateTime
 import java.time.ZoneOffset
 import org.jooq.DSLContext
-import org.jooq.Field
 import org.jooq.impl.DSL
 import org.slf4j.LoggerFactory
 
@@ -48,10 +47,6 @@ open class SqlResourceRepository(
   private val objectMapper: ObjectMapper,
   private val sqlRetry: SqlRetry
 ) : ResourceRepository {
-
-  companion object {
-    val EVENT_JSON_APPLICATION: Field<String> = field("json->'$.application'")
-  }
 
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
@@ -209,7 +204,7 @@ open class SqlResourceRepository(
         )
         // ...or application events that match the application as they apply to all resources
         .or(EVENT.SCOPE.eq(Scope.APPLICATION.name)
-          .and(EVENT_JSON_APPLICATION.eq(resource.application))
+          .and(EVENT.APPLICATION.eq(resource.application))
         )
         .orderBy(EVENT.TIMESTAMP.desc())
         .limit(limit)
@@ -248,7 +243,7 @@ open class SqlResourceRepository(
           )
           // ...or application events that match the application as they apply to all resources
           .or(EVENT.SCOPE.eq(Scope.APPLICATION.name)
-            .and(EVENT_JSON_APPLICATION.eq(event.application))
+            .and(EVENT.APPLICATION.eq(event.application))
           )
           .orderBy(EVENT.TIMESTAMP.desc())
           .limit(1)
--- a/keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
+++ b/keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
@@ -8,14 +8,23 @@ databaseChangeLog:
           alter table event add column application varchar(255) generated always as (json ->> '$.application');
     - createIndex:
         tableName: event
-        indexName: event_ref_scope_application_timestamp_idx
+        indexName: event_scope_application_timestamp_idx
         columns:
-        - column:
-            name: ref
         - column:
             name: scope
         - column:
             name: application
+        - column:
+            name: timestamp
+            descending: true
+    - createIndex:
+        tableName: event
+        indexName: event_scope_ref_timestamp_idx
+        columns:
+        - column:
+            name: scope
+        - column:
+            name: ref
         - column:
             name: timestamp
             descending: true
--- a/keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
+++ b/keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
@@ -1,0 +1,28 @@
+databaseChangeLog:
+- changeSet:
+    id: add-event-application-column
+    author: fletch
+    changes:
+    - sql:
+        sql: |
+          alter table event add column application varchar(255) generated always as (json ->> '$.application') stored;
+    - createIndex:
+        tableName: event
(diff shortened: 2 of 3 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -49,10 +48,6 @@ open class SqlResourceRepository(
   private val sqlRetry: SqlRetry
 ) : ResourceRepository {
 
-  companion object {
-    val EVENT_JSON_APPLICATION: Field<String> = field("json->'$.application'")
-  }
```
robfletcher, line 54: I'm basically replacing this with an actual column

keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
```diff
@@ -0,0 +1,28 @@
+databaseChangeLog:
+- changeSet:
+    id: add-event-application-column
+    author: fletch
+    changes:
+    - sql:
+        sql: |
+          alter table event add column application varchar(255) generated always as (json ->> '$.application') stored;
```
robfletcher, line 8: See https://www.compose.com/articles/mysql-for-json-generated-columns-and-indexing/

I'm not sure we _really_ need to use `stored` since the column will be part of an index. Any opinions?

keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
```diff
@@ -0,0 +1,28 @@
...
+        columns:
+        - column:
+            name: ref
+        - column:
+            name: scope
+        - column:
+            name: application
+        - column:
+            name: timestamp
+            descending: true
```
robfletcher, line 21: Create a new index the relevant queries can use

keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
```diff
@@ -0,0 +1,28 @@
...
+    - dropIndex:
+        tableName: event
+        indexName: event_uid_scope_timestamp_idx
```
robfletcher, line 24: Drop the old index that has one fewer columns

keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
lorin: That link is currently borked for some reason, wayback has the article:

https://web.archive.org/web/20190718135741/https://www.compose.com/articles/mysql-for-json-generated-columns-and-indexing/

lorin, line 8: Do you have a sense as to:
* how long this migration will take against main? (There are about 162K records in `event` right now)
* what will the impact be on the behavior of keel while the migration is happening? (Will this prevent keel from writing or reading the event table while it's happening?)

robfletcher: Good question. Running a query to extract that data from the event table takes under a second so I don't think there should be any significant impact.

keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
```diff
@@ -0,0 +1,28 @@
...
+    id: add-event-application-column
+    author: fletch
+    changes:
+    - sql:
+        sql: |
+          alter table event add column application varchar(255) generated always as (json ->> '$.application') stored;
+    - createIndex:
+        tableName: event
+        indexName: event_ref_scope_application_timestamp_idx
+        columns:
```
lorin, line 12: If I'm reading the where clause of the `eventHistory` query correctly, it looks like this:

```sql
select json from event
where 
  (scope="resource" and ref=uid) or
  (scope="application" and application=application)
...
```
This index is defined as `(ref,scope,application,timestamp)`.

I think the event history query will still require a full table scan, because the second part of the where clause can't take full advantage of the index. I think you need to define two indexes to avoid the table scan:

1. `(scope,ref,timestamp)` 
2. `(scope,application,timestamp)`

Hopefully, someone with more database experience than I can chime in here.

keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
```diff
@@ -0,0 +1,28 @@
...
+            name: scope
+        - column:
+            name: application
+        - column:
+            name: timestamp
+            descending: true
+    - dropIndex:
+        tableName: event
+        indexName: event_uid_scope_timestamp_idx
+    rollback:
```
lorin, line 25: The rollback currently doesn't handle the added and dropped indexes.

robfletcher: You get auto-rollback when you use Liquibase's defined migration types

keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
robfletcher: That's how _I_ always thought indices worked but @marchello2000 suggested that queries can benefit from using a subset of the columns in an index. Would be interested in his take on this.

marchello2000: For the index to be used (other than query optimizer deciding to use it, etc...) the columns used in the where clause have to be "contiguous" in the index. so if index is `col1, col2, col3` and you do a where clause for `col1=x or col3=y` you will get the benefit of index for `col1=x` but not for `col3=y`

So I think here you'd want to the `scope` to be first column of the index since it's used in both parts of the `or`. But after that only one of them will benefit from the index, so pick the one that has the most cardinality

you can make sure it uses the index with
```sql
explain format=json select json from event...
```

marchello2000: i am not sure what happens with the `or` it (MySQL) might just decide full table scan is more efficient.. You'd have to try out - I don't know for sure

lorin: I think it's worth doing a mini-benchmark test with one index and two indexes and see what the difference in performance is.

keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
robfletcher: I've changed it to virtual because we never actually select the value from that column, thus there's no point storing the value outside of the index. Testing locally with a dump of real data I could not see any significant difference in time to run the migration. Either way it was sub-second.

keel-sql/src/main/resources/db/changelog/20200714-add-event-application-column.yml
robfletcher: With my original index: `"query_cost": "53658.80"`
With revised 2 indices: `"query_cost": "146.21"`

I think you were right @lorin !

PR #1387, opened 2020-07-21
Add support for NPM artifacts

luispollo:
Adds native support for NPM artifacts through `igor` (sync) and `echo` (async). The NPM implementation leverages the versioning strategy used for Debian, which is conventionally used for both at Netflix (the class is now renamed to `NetflixSemverVersioningStrategy` for that reason).

Sample API response with a test app/artifact:
```yaml
---
applicationPaused: false
hasManagedResources: false
currentEnvironmentConstraints: []
artifacts:
- name: "@netflix-treasure/treasure-example-ui"
  type: "npm"
  reference: "treasure-example"
  versions:
  - version: "1.0.0-10-5d44dc4890536dbb9888b6b6e22f3250734db082"
    displayName: "1.0.0"
    environments: []
    build:
      id: 10
    git:
      commit: "5d44dc4890536dbb9888b6b6e22f3250734db082"
```

Depends on https://github.com/spinnaker/igor/pull/821


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApproveOldVersionTests.kt
```diff
@@ -137,6 +137,10 @@ abstract class ApproveOldVersionTests<T : KeelRepository> : JUnit5Minutests {
       )
     }
 
+    after {
+      flush()
```
luispollo, line 141: This was missing and causing the DB records to leak into tests following this one. Thanks @robfletcher for the help debugging this!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NetflixSemVerVersioningStrategy.kt
```diff
@@ -0,0 +1,51 @@
...
+  override val comparator: Comparator<String> =
+    NETFLIX_SEMVER_COMPARATOR
+
+  private val NPM_VERSION_REGEX = Regex("(\\d+\\.\\d+\\.\\d+(-rc(\\.\\d+)?)?)(-[h]?\\w+)?(-\\w+)?")
+  private val BUILD_REGEX = Regex("-[h]?")
+
+  /**
+   * Extracts a version display name from the longer version string, leaving out build and git details if present.
+   */
+  fun getVersionDisplayName(artifact: PublishedArtifact): String {
```
emjburns, line 22: I think it's unclear why you have NPM specific stuff in a comparator that's used for both debians and npm artifacts. Can you make this a little cleaner?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
```diff
@@ -34,11 +35,11 @@ interface ArtifactSupplier<A : DeliveryArtifact, V : VersioningStrategy> : Spinn
    * persisted and evaluated for promotion into deployment environments.
    *
    * The default implementation of [publishArtifact] simply publishes the event via the [EventPublisher],
-   * and should *not* be overridden by implementors.
+   * and should typically *not* be overridden by implementors.
    */
   @JvmDefault
-  fun publishArtifact(artifactPublishedEvent: ArtifactPublishedEvent) =
-    eventPublisher.publishEvent(artifactPublishedEvent)
+  fun publishArtifact(artifact: PublishedArtifact) =
```
gal-yardeni, line 41: why was the event name changed?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -26,21 +26,30 @@ import org.springframework.stereotype.Component
...
     runWithIoContext {
-      artifactService
-        .getVersions(artifact.name)
+      artifactService.getVersions(artifact.name, artifactType = DEBIAN)
         .map { version -> "${artifact.name}-$version" }
         .sortedWith(artifact.versioningStrategy.comparator)
         .firstOrNull() // versioning strategies return descending by default... ¯\_(ツ)_/¯
         ?.let { version ->
-          artifactService.getArtifact(artifact.name, version.removePrefix("${artifact.name}-"))
+          artifactService.getArtifact(artifact.name, version.removePrefix("${artifact.name}-"), DEBIAN)
```
gal-yardeni, line 52: hmm why was `deb` added here? is it working only for debians?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NetflixSemVerVersioningStrategy.kt
luispollo: Oops, yeah, that was a poor choice of a variable name. Should be something like `NETFLIX_VERSION_REGEX`.

luispollo: BTW, I know for a fact that this regex is not matching _all_ the patterns we use, because the "convention" is somewhat loose. I expect to have to revisit the expression later if we find cases that don't match and that we want to support.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
gal-yardeni: oh ooops I didn't notice I'm in DebianArtifactSupplier class 😅

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
luispollo: I didn't change the event name, just the type that's passed into the method. Instead of passing an event with multiple artifacts, I'm now passing a single artifact, which is what this method is intended to process.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
luispollo: See the change to `ArtifactService.getArtifact()`. You now have to pass in the artifact type as a parameter because the service supports querying multiple types.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NetflixSemVerVersioningStrategy.kt
```diff
@@ -0,0 +1,51 @@
...
+    return NPM_VERSION_REGEX.find(artifact.version)?.groups?.get(1)?.value
+      ?: artifact.version
+  }
+
+  /**
+   * Extracts the build number from the version string, if available.
+   */
+  fun getBuildNumber(artifact: PublishedArtifact): Int? {
```
gal-yardeni, line 30: aren't we get the build number straight from the event? like in the example you mentioned yesterday: ` build:
      id: 10` ?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NetflixSemVerVersioningStrategy.kt
```diff
@@ -0,0 +1,51 @@
...
+    return NPM_VERSION_REGEX.find(artifact.version)?.groups?.get(4)?.value
+      ?.let {
+        it.replace(BUILD_REGEX, "")
+      }?.toInt()
+  }
+
+  /**
+   * Extracts the commit hash from the version string, if available.
+   */
+  fun getCommitHash(artifact: PublishedArtifact): String? {
```
gal-yardeni, line 40: same for commit hash ^

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NetflixSemVerVersioningStrategy.kt
luispollo: Fixed.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NetflixSemVerVersioningStrategy.kt
luispollo: No, we're extracting it from the artifact version string.

luispollo: Confirmed: echo is not adding those fields to the events. We can add them later, though.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -26,21 +26,30 @@ import org.springframework.stereotype.Component
...
+    } else {
+      log.debug("Ignoring artifact event without release status: $artifact")
+    }
+  }
 
   override fun getLatestArtifact(deliveryConfig: DeliveryConfig, artifact: DeliveryArtifact): PublishedArtifact? =
     runWithIoContext {
-      artifactService
-        .getVersions(artifact.name)
+      artifactService.getVersions(artifact.name, artifact.statusesForQuery, DEBIAN)
```
luispollo, line 47: This was previously looking up versions for all statuses, not only the ones configured in the delivery artifact.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NetflixSemVerVersioningStrategy.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import com.netflix.spinnaker.keel.api.artifacts.PublishedArtifact
+import com.netflix.spinnaker.keel.api.artifacts.VersioningStrategy
+import com.netflix.spinnaker.keel.core.NETFLIX_SEMVER_COMPARATOR
+
+/**
+ * A [VersioningStrategy] that codifies the versioning scheme conventions at Netflix.
+ */
+object NetflixSemVerVersioningStrategy : VersioningStrategy {
+  override val type: String = "netflix-semver"
```
luispollo, line 11: Technically, this is used as the discriminator to determine the sub-type of `VersioningStrategy` for deserialization, but we don't expose that in the delivery config as a property (it's hard-coded with each `DeliveryArtifact` sub-class), so this type name shouldn't matter.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/apidocs/ApiDocCompatibilityTests.kt
```diff
@@ -52,7 +54,7 @@ class ApiDocCompatibilityTests : JUnit5Minutests {
       .let { jacksonObjectMapper().readTree(it) }
   }
 
-  fun tests() = rootContext<SchemaValidator> {
+  fun tests() = SKIP - rootContext<SchemaValidator> {
```
emjburns, line 57: is this just commenting out this test essentially?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/apidocs/ApiDocTests.kt
```diff
@@ -105,7 +112,7 @@ class ApiDocTests : JUnit5Minutests {
       .let { jacksonObjectMapper().readTree(it) }
   }
 
-  fun tests() = rootContext<Assertion.Builder<JsonNode>> {
+  fun tests() = SKIP - rootContext<Assertion.Builder<JsonNode>> {
```
emjburns, line 115: why did you need to skip this? I'd suggest putting a comment here..

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/apidocs/ApiDocCompatibilityTests.kt
emjburns: I'd add a comment as to why you needed to do this

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/apidocs/ApiDocTests.kt
luispollo: Because this is causing all PRs to fail and there's no one consuming the API docs. We had a discussion in Slack this morning, but the TL;DR is that it has to do with the ordering of `@Bean`s that customize the handling of types to produce the API docs, which is causing a `StackOverflowError`. In discussing with @robfletcher, we agreed to disable these tests temporarily until we can figure that out, since it's unrelated to changes in the PRs.

emjburns: Sure, can you add that as a comment please?

luispollo:
> I'd prefer you pull the build and commit extraction out of the comparator. I think it should live in its own class specifically for that purpose, not be lumped into a comparitor.

They're not in the comparator, they're in the versioning strategy class...

Issue #1380, opened 2020-07-15
Better exception when exceeding max comment size


lorin:
If the user specifies a comment in "mark as bad" (or "pin", I think) and it exceeds the maximum size, the error returned isn't meaningful, it's just a 500 error. We should return a more meaningful exception with a proper message so that the UI will communicate back to the end user what the problem was.

PR #1389, opened 2020-07-23
fix(Comment length): throw bad request if comment length > 256

gal-yardeni:
fixes https://github.com/spinnaker/keel/issues/1380 .

Applies for `pinned` and `veto` endpoints: if `comment` field length is > 256 (the max in the DB), throw `bad request` instead of `internal error`.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
```diff
@@ -9,7 +9,15 @@ data class EnvironmentArtifactPin(
   val version: String,
   val pinnedBy: String?,
   val comment: String?
-)
+) {
+  init {
+    if (comment != null) {
+      require(comment.length <= 255) {
+        "Comment length should be 255 characters max."
```
luispollo, line 16: This can be displayed in the UI, correct? In that case, I'd suggest tweaking the message a bit:
```suggestion
        "Comments should have a maximum size of 255 characters."
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactPin.kt
```diff
@@ -19,4 +27,12 @@ data class PinnedEnvironment(
   val pinnedBy: String?,
   val pinnedAt: Instant?,
   val comment: String?
-)
+) {
+  init {
+    if (comment != null) {
+      require(comment.length <= 255) {
+        "Comment length should be 255 characters max."
```
luispollo, line 34:
```suggestion
        "Comments should have a maximum size of 255 characters."
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactVeto.kt
```diff
@@ -11,7 +11,15 @@ data class EnvironmentArtifactVeto(
   val version: String,
   val vetoedBy: String?,
   val comment: String?
-)
+) {
+  init {
+    if (comment != null) {
```
luispollo, line 16: Instead of repeating this code in 3 different places, how about declaring a `validateComment` function in a shared `validations.kt` module that you can call from all places?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
```diff
@@ -61,6 +61,15 @@ internal class ApplicationControllerTests : JUnit5Minutests {
     const val user = "keel@keel.io"
   }
 
+  val payloadWithLongComment =
+    """{
+        |  "version": "master-h22.0e0310f",
+        |  "reference": "my-artifact",
+        |  "targetEnvironment": "testing",
+        |  "comment": "ccdkclkvdlf;kgvfl;dkdfl;,d.,md,vmfdkljfkgm,mdbkfgjkfgmb,fbmfkbnfgkbnfgkbnfgbnfgbnfgnbfgbnfgbmf,.bmfgklbhjgkbmfg,bmkfbkvmlvmka;ldvkdflmd,mcas;dflkfjdflkgvfd,mvldfkgjbkm,m,dvmdfljvkdmvd,vfm,lfdvjfdkmvfdlvmfkdvmfddgfksdl;gkfdlgkfdl;gkdfl;gkdflgkfdlgkdlfgkldfgkfdllg",
```
luispollo, line 69: This is a fun one to generate random text... 😄 
https://baconipsum.com/api/?type=all-meat&paras=2&start-with-lorem=1

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentArtifactVeto.kt
gal-yardeni: honestly, I think it's more complicated and less readable, for this simple use case, as it's a simple data class with no dependencies on other classes (validators), but it's only my opinion..

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
```diff
@@ -61,6 +61,16 @@ internal class ApplicationControllerTests : JUnit5Minutests {
     const val user = "keel@keel.io"
   }
 
+  val payloadWithLongComment =
+    """{
+        |  "version": "master-h22.0e0310f",
+        |  "reference": "my-artifact",
+        |  "targetEnvironment": "testing",
+        |  "comment": "Bacon ipsum dolor amet turducken prosciutto shoulder ground round hamburger, flank frankfurter rump ham hock sirloin leberkas meatloaf shankle landjaeger pig.  Shoulder shankle doner ball tip burgdoggen kevin alcatra bresaola.  Leberkas alcatra cow, sausage picanha chislic tongue hamburger turkey tail chicken flank.  Chislic tri-tip ribeye meatloaf swine biltong rump.  Bacon salami ball tip, kevin turkey cow short loin cupim drumstick landjaeger rump meatball flank.",
+  "Burgdoggen leberkas pork loin salami ribeye pig, ham meatloaf sirloin beef prosciutto chuck cupim pork swine.  Bacon cupim doner pig drumstick prosciutto pork loin picanha beef ribs pork.  Pig beef strip steak buffalo short ribs swine jowl short loin alcatra.  Frankfurter spare ribs cupim leberkas.  Ham hock cow venison boudin ham tenderloin chuck chislic porchetta kevin leberkas.  Tri-tip buffalo meatball leberkas tail.",
```
luispollo, line 70: I think this might break the test because of the comma before. You probably only need the first paragraph?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
```diff
@@ -61,6 +61,16 @@ internal class ApplicationControllerTests : JUnit5Minutests {
     const val user = "keel@keel.io"
   }
 
+  val payloadWithLongComment =
+    """{
+        |  "version": "master-h22.0e0310f",
+        |  "reference": "my-artifact",
+        |  "targetEnvironment": "testing",
+        |  "comment": "Bacon ipsum dolor amet turducken prosciutto shoulder ground round hamburger, flank frankfurter rump ham hock sirloin leberkas meatloaf shankle landjaeger pig.  Shoulder shankle doner ball tip burgdoggen kevin alcatra bresaola.  Leberkas alcatra cow, sausage picanha chislic tongue hamburger turkey tail chicken flank.  Chislic tri-tip ribeye meatloaf swine biltong rump.  Bacon salami ball tip, kevin turkey cow short loin cupim drumstick landjaeger rump meatball flank.",
```
luispollo, line 69: Luv it! 😆  🥓

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/Validations.kt
```diff
@@ -0,0 +1,10 @@
+package com.netflix.spinnaker.keel.core
```
luispollo, line 1: nit: Generally I think we name kotlin modules with a lower-case first letter.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationControllerTests.kt
gal-yardeni: shortened the comment 👍

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/Validations.kt
gal-yardeni: fixed

PR #1393, opened 2020-07-23
debug(artifacts): Add more debug logs

luispollo:
It's hard to troubleshoot things without logs...


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -35,9 +35,12 @@ class ArtifactController(
   )
   @ResponseStatus(ACCEPTED)
   fun submitArtifact(@RequestBody echoArtifactEvent: EchoArtifactEvent) {
+    log.debug("Received artifact event")
```
gal-yardeni, line 38: should we log here which artifact event is it?

luispollo: I went back and forth on that because the payloads may be large, but I guess you're right. I'll add it.

luispollo: Done!

PR #1394, opened 2020-07-24
fix(artifacts): Fix artifact event publishing

luispollo:
Silly, silly me...


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
```diff
@@ -39,7 +39,7 @@ interface ArtifactSupplier<A : DeliveryArtifact, V : VersioningStrategy> : Spinn
    */
   @JvmDefault
   fun publishArtifact(artifact: PublishedArtifact) =
-    eventPublisher.publishEvent(artifact)
+    eventPublisher.publishEvent(ArtifactPublishedEvent(listOf(artifact)))
```
luispollo, line 42: @gal-yardeni you kinda spotted this in the other PR but I didn't pay enough attention... 🤦

PR #1396, opened 2020-07-27
chore(spotless): upgrade spinnaker gradle plugin

plumpy:
This includes updates to google-java-format and ktlint, so requires some
formatting changes. These were made with a simple `./gradlew spotlessApply`.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@ kapt.use.worker.api=true
 korkVersion=7.59.0
 liquibaseTaskPrefix=liquibase
 org.gradle.parallel=true
-spinnakerGradleVersion=8.3.0
+spinnakerGradleVersion=8.4.0
 
 # To enable a composite reference to a project, set the
 #  project property `'<projectName>Composite=true'`.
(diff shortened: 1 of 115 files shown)


luispollo:
@plumpy Thanks for the PR, and sorry for the long delay providing any feedback. Would you be willing to solve the conflicts?
Cc: @robfletcher since I think this relates to issues you ran into with the formatting lib last week.

plumpy:
The latest version of the plugin uses Google Java Format 1.8, which requires JDK 11. So you aren't going to be able to build with a Java 8 JDK anymore.

The rest of the projects all use cross-compilation to compile with the Java 11 JDK against the Java 8 API. I tried to do that with keel but ran into issues with Jooq. I believe it generates Java 11 code under a Java 11 JDK, even if the target is Java 8. Would you like me to look into fixing that? You won't be able to upgrade the plugin until that works.

plumpy:
FYI, You can see the errors yourself by doing this:

```
export JAVA_HOME /path/to/jdk11
export JDK_18 /path/to/jdk18
./gradlew -PenableCrossCompilation=true build
```

robfletcher:
> FYI, You can see the errors yourself by doing this:
> 
> ```
> export JAVA_HOME /path/to/jdk11
> export JDK_18 /path/to/jdk18
> ./gradlew -PenableCrossCompilerPlugin=true build
> ```

Why is that not breaking in #1433 ?

robfletcher:
That failure on the build is confusing. It says the class version is 55 (which means compiled with Java 11) but that's on the JDK 8 build job. I don't understand.

plumpy:
_That_ failure is because google-java-format 1.8 was compiled with Java 11, so the Java 8 `javac` can't run it. Hence why we need to switch to the Java 11 `javac`

PR #1397, opened 2020-07-29
Fix overrides for Titus cluster map properties

robfletcher:
There was a subtle bug where the default took precedence over the override in map properties of Titus clusters. This fixes it.


keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterDesiredStateResolutionTests.kt
```diff
@@ -0,0 +1,148 @@
...
+          test("override value applies in overridden region") {
+            expectThat(desired)
+              .getValue("af-south-1")
+              .get(desiredProperty)
+              .isEqualTo(specProperty.get(spec.overrides.getValue("af-south-1")))
+          }
+        }
+      }
+//
+//    context("the `env` property") {
```
gal-yardeni, line 58: are those not valid tests?

Issue #1270, opened 2020-06-04
Add artifact (reference or uid) to constraint state & approvals, or prove this isn't a problem


emjburns:
Right now we store constraint state ([here](https://github.com/spinnaker/keel/blob/master/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt#L450-L460)) with the identifiers as: environment UID, constraint type, and artifact version. This is missing the artifact (reference or uid). Why? Not sure.

This feels like it's encoding the assumption that you'll only ever have one artifact in each environment, so the version is fine to identify it. That doesn't seem like a good assumption.

The artifact uid or reference is entirely omitted from both env+constraint tables (`environment_artifact_constraint`, `environment_artifact_queued_approval`).

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

PR #1399, opened 2020-07-31
fix(artifactType): resolve mixing between artifact types and version

gal-yardeni:
Recently we came across a delivery config that had both docker and Debian artifacts, and keel didn't manage it correctly.
A log message helped us discovered some mixup between artifact types and versions when queuing versions for approval. 

 -- for example:
docker artifact with version 1.0
Debian artifact with version 2.0
both were approved for `testing` environment
our current flow will fetch all `queuedForApproval` versions and returned both 1.0 and 2.0, and then will `checkStatelessConstraints` with each artifact type (like docker with version 2.0), which will potentially be an error.

The problem was that we don't take into consideration a unique artifact identifier like `reference` when fetching all `queuedForApproval` versions.
This fix adds a new column for artifact reference to the `environment_artifact_queued_approval` table, which will enforce considering the right artifact per version.

[Update] This PR is also addressing issue #1270 by adding a `reference` column to `environment_artifact_constraint` table as well, and ensuring a unique and correct fetch by artifact version + artifact reference combination. 

[Update #2] at first I was using `artifact_type` as an identifier, but after a discussion with the team we've realized a single config can have multiple artifact types with the same version, BUT their reference cannot be the same. So I converted `type` into `reference`.


keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionCheckerTests.kt
```diff
@@ -333,5 +351,102 @@ internal class NewEnvironmentPromotionCheckerTests : JUnit5Minutests {
...
+      }
+
+      test("verify the right artifact type and version approved") {
+        verify {
+          repository.approveVersionFor(deliveryConfigWith2ArtifactTypes, dockerArtifact, "2.0", environment.name)
+          repository.approveVersionFor(deliveryConfigWith2ArtifactTypes, debianArtifact, "3.0", environment.name)
+        }
+      }
+
+      test("approved versions of other artifact types are not getting mixed up and are not checked") {
```
gal-yardeni, line 409: Prior to this change, this test failed as different artifact types and versions got checked.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -529,6 +530,7 @@ class SqlDeliveryConfigRepository(
                 .set(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ENVIRONMENT_UID, envUid)
```
emjburns, line 530: you can delete the comment on line 528! :)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -897,14 +917,15 @@ class SqlDeliveryConfigRepository(
   }
 
   // todo eb: add link to artifact https://github.com/spinnaker/keel/issues/1270
```
emjburns, line 919: You can delete this comment now! :)

keel-sql/src/main/resources/db/changelog/20200730-add-artifact-type-column.yml
```diff
@@ -0,0 +1,25 @@
+databaseChangeLog:
+  - changeSet:
+      id: add-artifact-type-column
+      author: lpollo
```
emjburns, line 4: this should be you! not that it really matters

keel-sql/src/main/resources/db/changelog/20200730-add-artifact-type-column.yml
```diff
@@ -0,0 +1,25 @@
...
+      changes:
+        - addColumn:
+            tableName: environment_artifact_queued_approval
+            columns:
+              - name: artifact_type
+                type: varchar(255)
```
emjburns, line 10: I think this should have:
`afterColumn: artifact_version` so it's not at the end of the table.

Also, do you ever query using artifact type? we might need an index on it if you do anything like "get all pending debians for this environment"

keel-sql/src/main/resources/db/changelog/20200730-add-artifact-type-column.yml
```diff
@@ -0,0 +1,25 @@
...
+                constraints:
+                  nullable: false
+        - addColumn:
+            tableName: environment_artifact_constraint
+            columns:
+              - name: artifact_type
+                type: varchar(255)
```
emjburns, line 17: same `afterColumn` comment + question about indexing here as well

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -897,14 +917,15 @@ class SqlDeliveryConfigRepository(
...
-  override fun getQueuedConstraintApprovals(deliveryConfigName: String, environmentName: String): Set<String> {
+  override fun getQueuedConstraintApprovals(deliveryConfigName: String, environmentName: String, artifactType: String): Set<String> {
     val environmentUID = environmentUidByName(deliveryConfigName, environmentName)
       ?: return emptySet()
 
     return sqlRetry.withRetry(READ) {
       jooq.select(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ARTIFACT_VERSION)
         .from(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL)
         .where(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ENVIRONMENT_UID.eq(environmentUID))
+        .and(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ARTIFACT_TYPE.eq(artifactType))
```
emjburns, line 928: adding this here means that when you roll this out we won't consider older versions with `artifactType == null` (all the old data). It's probably worth allowing null here as well

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -937,15 +960,17 @@ class SqlDeliveryConfigRepository(
...
   ) {
     sqlRetry.withRetry(WRITE) {
       environmentUidByName(deliveryConfigName, environmentName)
         ?.also { envUid ->
           jooq.deleteFrom(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL)
             .where(
               ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ENVIRONMENT_UID.eq(envUid),
-              ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ARTIFACT_VERSION.eq(artifactVersion))
+              ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ARTIFACT_VERSION.eq(artifactVersion),
+              ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ARTIFACT_TYPE.eq(artifactType))
```
emjburns, line 973: I think you'll also have to allow `null` here to be able to clean up old data

keel-sql/src/main/resources/db/changelog/20200730-add-artifact-type-column.yml
```diff
@@ -0,0 +1,25 @@
...
+                constraints:
+                  nullable: false
+
+    rollback:
```
emjburns, line 21: You can do:
```      rollback:
        - dropColumn:
            tableName: environment_artifact_constraint
            columnName: artifact_type```
for both, you don't have to write the raw sql

**Comment by emjburns** [comment]
Some small comments for you. My main question is - have you made sure this works in all forms on old data? I don't want there to be a problem rolling this out because we fail when `artifact_type` is `null`.

File: keel-sql/src/main/resources/db/changelog/20200730-add-artifact-type-column.yml (hunk 5 above)
Comment (gal-yardeni) [reply keel-sql/src/main/resources/db/changelog/20200730-add-artifact-type-column.yml:10]: oh i didn't know about `afterColumn` trick :) 
I'm not querying by type only, we query by type + version + artifact type+ environment uid from what I can tell.

Comment (robfletcher) [reply keel-sql/src/main/resources/db/changelog/20200730-add-artifact-type-column.yml:10]: Sounds like it should be added into that index then, no?

Comment (gal-yardeni) [reply keel-sql/src/main/resources/db/changelog/20200730-add-artifact-type-column.yml]: added.

**Comment by gal-yardeni** [comment]
@emjburns @robfletcher I've changed `type` into `reference` as we discussed on slack. I added the index and made `artifact_referance` nullable for backwards compatibility. Let me know if anything else is missing! thanks

File: keel-sql/src/main/resources/db/changelog/20200731-add-artifact-reference-column.yml [hunk 10]
```diff
@@ -0,0 +1,71 @@
...
+                  nullable: true
+      rollback:
+        - dropColumn:
+            tableName: environment_artifact_constraint
+            columnName: artifact_reference
+        - dropColumn:
+            tableName: environment_artifact_queued_approval
+            columnName: artifact_reference
+
+  - changeSet:
```
gal-yardeni, line 30: for some reason, the second part `create-artifact-reference-indicies` is not running. What did I miss?

keel-sql/src/main/resources/db/changelog/20200731-add-artifact-reference-column.yml
```diff
@@ -0,0 +1,71 @@
...
+                afterColumn: artifact_version
+                constraints:
+                  nullable: true
+      rollback:
+        - dropColumn:
+            tableName: environment_artifact_constraint
+            columnName: artifact_reference
+        - dropColumn:
+            tableName: environment_artifact_queued_approval
+            columnName: artifact_reference
```
robfletcher, line 28: I don't think you need this. Most Liquibase migrations have auto-rollback. It's only really when you're doing something unusual (e.g. with `sql`) that you need an explicit rollback.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunnerTests.kt
```diff
@@ -68,6 +68,7 @@ internal class EnvironmentConstraintRunnerTests : JUnit5Minutests {
       deliveryConfig.name,
       environment.name,
       "2.0",
+      artifact.type,
```
robfletcher, line 71:
```suggestion
      artifact.reference,
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunnerTests.kt
```diff
@@ -76,6 +77,7 @@ internal class EnvironmentConstraintRunnerTests : JUnit5Minutests {
       deliveryConfig.name,
       environment.name,
       "1.2",
+      artifact.type,
```
robfletcher, line 80:
```suggestion
      artifact.reference,
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluatorTests.kt
```diff
@@ -50,6 +50,7 @@ internal class ManualJudgementConstraintEvaluatorTests : JUnit5Minutests {
       deliveryConfigName = configName,
       environmentName = env.name,
       artifactVersion = version,
+      artifactReference = artifact.type,
```
robfletcher, line 53:
```suggestion
      artifactReference = artifact.reference,
```

keel-sql/src/main/resources/db/changelog/20200731-add-artifact-reference-column.yml
gal-yardeni: thanks! is this true for the index rollback below as well?

PR #1400, opened 2020-07-31
chore(dockerfiles): add Dockerfiles for container images

nimakaviani:
The number of Dockerfiles introduced in this PR matches one to one to what we have for other spinnaker services. 

I have made some swapping of layers in the Dockerfiles for improved efficiency.

/cc @plumpy 
/cc @link108 
/cc @queueburt

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -1,7 +1,8 @@
 FROM alpine:3.11
 MAINTAINER sig-platform@spinnaker.io
 RUN apk --no-cache add --update bash openjdk11-jre
-RUN adduser -D -S spinnaker
+RUN addgroup -S -g 10111 spinnaker
+RUN adduser -D -S spinnaker -u 10111 -G spinnaker
 COPY keel-web/build/install/keel-web /opt/keel
 RUN mkdir -p /opt/keel/plugins && chown -R spinnaker:nogroup /opt/keel/plugins
 USER spinnaker
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -3,7 +3,7 @@ MAINTAINER sig-platform@spinnaker.io
 RUN apk --no-cache add --update bash openjdk11-jre
 RUN addgroup -S -g 10111 spinnaker
 RUN adduser -D -S spinnaker -u 10111 -G spinnaker
-COPY keel-web/build/install/keel-web /opt/keel
+COPY keel-web/build/install/keel /opt/keel
 RUN mkdir -p /opt/keel/plugins && chown -R spinnaker:nogroup /opt/keel/plugins
 USER spinnaker
-CMD ["/opt/keel/bin/keel-web"]
+CMD ["/opt/keel/bin/keel"]
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -1,0 +1,8 @@
+FROM alpine:3.11
+MAINTAINER sig-platform@spinnaker.io
+RUN apk --no-cache add --update bash openjdk11-jre
+RUN adduser -D -S spinnaker
+COPY keel-web/build/install/keel-web /opt/keel
+RUN mkdir -p /opt/keel/plugins && chown -R spinnaker:nogroup /opt/keel/plugins
+USER spinnaker
+CMD ["/opt/keel/bin/keel-web"]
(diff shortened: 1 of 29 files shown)


Dockerfile.slim
```diff
@@ -0,0 +1,8 @@
+FROM alpine:3.11
+MAINTAINER sig-platform@spinnaker.io
+RUN apk --no-cache add --update bash openjdk11-jre
+RUN adduser -D -S spinnaker
```
plumpy, line 4:
```suggestion
RUN addgroup -S -g 10111 spinnaker
RUN adduser -D -S spinnaker -u 10111 -G spinnaker
```

We're going to add this to all the other files in the next week or two (because [a Kubernetes securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) requires a constant uid), so we might as well get ahead of things.

Dockerfile.ubuntu
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:bionic
+MAINTAINER sig-platform@spinnaker.io
+RUN apt-get update && apt-get -y install openjdk-11-jre-headless wget
+RUN adduser --disabled-login --system spinnaker
```
plumpy, line 4:
```suggestion
RUN adduser --system --uid 10111 --group spinnaker
```

(same uid change, plus `--disabled-login` is redundant, it's implied by `--system`)

nimakaviani:
thanks, done!

plumpy:
I can't get the compile step to work, can you? I get this:

```
Execution failed for task ':keel-sql:generateJooqMetamodel'.
> com.spotify.docker.client.exceptions.DockerException: java.util.concurrent.ExecutionException: javax.ws.rs.ProcessingException: java.io.IOException: No such file or directory
```

nimakaviani:
Good catch! It breaks for me too when trying to run the compile container.

The reason is, `keel-sql` uses [ayedo/jooq-modelator](https://github.com/ayedo/jooq-modelator) to dynamically generate the database metamodel code for keel. And `jooq-modelator` has an unfortunate dependency to docker, to start up a mysql container and build / verify the metamodel against it.

the error you saw occurs if `/var/run/docker.sock` is missing inside the compile container. 

`docker run -v/var/run/docker.sock:/var/run/docker.sock ...` will allow jOOQ to start up a mysql container. But then, how to enable communication between the two containers for the dynamic code generation to work requires extra network wiring, probably as part of the CI setup.

I don't have a good answer as to how to go about it right now, but let me take it to the platform SIG next Tuesday, and we can discuss options after.

nimakaviani:
depends on #1410 because of the binary file renaming.

Dockerfile.slim
```diff
@@ -0,0 +1,9 @@
+FROM alpine:3.11
+MAINTAINER sig-platform@spinnaker.io
+RUN apk --no-cache add --update bash openjdk11-jre
+RUN addgroup -S -g 10111 spinnaker
+RUN adduser -D -S spinnaker -u 10111 -G spinnaker
+COPY keel-web/build/install/keel /opt/keel
```
luispollo, line 6: I get a `keel-web` dir, not `keel`, from running `keel-web:installDist`. Does Docker build this OK for you?

nimakaviani: sorry I should have been clearer in my last comment.

I only changed `keel-web` to `keel` in the last commit of this PR. reason is, once #1410 is merged, the `io.spinnaker.package` plugin will create the install script as `keel` and not `keel-web` anymore. I didnt want to have to fix up references in the dockerfiles after $1410 is merged, hence [this comment](https://github.com/spinnaker/keel/pull/1400#issuecomment-670227267) on this PR depending on #1410.

I guess merging this line of work should go as #1409, #1410, and then #1400.

luispollo: Got it. The Dockerfiles look good to me, and I merged #1410 before seeing this comment, so I guess I'll watch for kork changes so that we can get this merged. I didn't see a PR linked in the issue -- were you planning to make that change in kork as well?

luispollo: Actually, I don't think there's any harm in getting this merged before #1409 is fixed. Let me know.

nimakaviani: there isn't any harm except that it wont build properly in the meantime. so if there is no dependency to it, then we are good.

nimakaviani: > I didn't see a PR linked in the issue -- were you planning to make that change in kork as well?

I can PR the kork change too but then a new release needs to be cut for it and keel will have to bump the kork version. I think it is easier if as part of #1409, someone with permissions to do all three tasks above can do all of this in one go.

PR #1402, opened 2020-08-03
Extension registry

robfletcher:
Currently identification of sub-types for generating API docs is completely decoupled from registering them with Jackson in order to support deserialization. This is particularly awkward for plugins as there is no convenient way to augment API doc generation. This PR gives us a simple central registry of extensions to things like `ResourceSpec` that can be used to drive both Jackson and API doc configuration.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/support/ExtensionRegistry.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+interface ExtensionRegistry {
+  fun <BASE : Any> register(
+    baseType: Class<BASE>,
+    extensionType: Class<out BASE>,
+    discriminator: String
+  )
+
+  fun <BASE : Any> extensionsOf(baseType: Class<BASE>): Map<String, Class<out BASE>>
+}
```
robfletcher, line 14: This interface is part of `keel-api` so plugins can use it to add additional sub-types that will get properly registered with Jackson and springdoc-openapi without requiring the plugin to have either of those as dependencies.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/extensions/DefaultExtensionRegistry.kt
```diff
@@ -0,0 +1,34 @@
...
+    discriminator: String
+  ) {
+    baseToExtensionTypes
+      .getOrPut(baseType, ::mutableMapOf)
+      .also { it[discriminator] = extensionType }
+    log.info("Registering extension \"$discriminator\" for ${baseType.simpleName} using ${extensionType.simpleName}")
+    mappers.forEach {
+      it.registerSubtypes(NamedType(extensionType, discriminator))
+    }
+  }
```
robfletcher, line 27: Default implementation does the Jackson registration right away and stores the subtype mapping for later use.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
```diff
@@ -45,8 +47,7 @@ class KeelConfigurationFinalizer(
     (resourceHandlers.map { it.supportedKind } + specMigrators.map { it.input })
       .forEach { (kind, specClass) ->
         log.info("Registering ResourceSpec sub-type {}: {}", kind, specClass.simpleName)
-        val namedType = NamedType(specClass, kind.toString())
-        objectMappers.forEach { it.registerSubtypes(namedType) }
+        extensionRegistry.register(specClass, kind.toString())
```
robfletcher, line 50: Example use, here replacing explicit Jackson configuration

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/ConstraintModelConverter.kt
```diff
@@ -11,14 +12,14 @@ import org.springframework.stereotype.Component
...
+    get() = mapping.values.toList()
 
   override val discriminator: String? = Constraint::type.name
 
-  override val mapping: Map<String, Class<out Constraint>> = evaluators.associate {
-    it.supportedType.name to it.supportedType.type
-  }
+  override val mapping: Map<String, Class<out Constraint>>
+    get() = extensionRegistry.extensionsOf()
 }
```
robfletcher, line 25: I'd like to unify all these implementations as they're basically all the same. I'll probably do another commit to this PR.

emjburns:
Very cool! I like

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/RegisteredExtensionModelConverter.kt
```diff
@@ -0,0 +1,32 @@
...
+        .also { schema ->
+            schema.discriminator = Discriminator()
+              .propertyName("type") // TODO: is this a broken assumption?
+              .mapping(extensionTypes.mapValues { (_, v) -> constructRef(v.simpleName) })
+        }
+      ref(annotatedType.rawClass)
+    } else {
+      super.resolve(annotatedType, context, chain)
+    }
+}
```
robfletcher, line 32: This replaces several very similar model converters for other types with a single class that will pick up anything registered in `ExtensionRegistry`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/support/ExtensionRegistry.kt
```diff
@@ -0,0 +1,26 @@
...
+    extensionType: Class<out BASE>,
+    discriminator: String
+  )
+
+  fun <BASE : Any> extensionsOf(baseType: Class<BASE>): Map<String, Class<out BASE>>
+
+  fun baseTypes(): Collection<Class<*>>
+}
+
+inline fun <reified BASE : Any> ExtensionRegistry.extensionsOf(): Map<String, Class<out BASE>> =
```
luispollo, line 18: Just curious: is there any advantage to making these extension functions instead of `default` functions inside the `interface` body?

robfletcher: You can't inline a member function, and if it's not inline it can't use reified generics.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/GenericWildcardTypeModelConverter.kt
```diff
@@ -22,12 +21,12 @@ import org.springframework.stereotype.Component
  *
  * @see https://github.com/springdoc/springdoc-openapi/issues/434#issuecomment-586700760
  */
-@Component
+// @Component
```
luispollo, line 24: Was this intentional?

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/RegisteredExtensionModelConverter.kt
```diff
@@ -0,0 +1,32 @@
...
+@Component
+class RegisteredExtensionModelConverter(
+  private val extensionRegistry: ExtensionRegistry
+) : BaseModelConverter() {
+  override fun resolve(
+    annotatedType: AnnotatedType,
+    context: ModelConverterContext,
+    chain: MutableIterator<ModelConverter>
+  ): Schema<*>? =
+    if (annotatedType.rawClass in extensionRegistry.baseTypes()) { val extensionTypes = extensionRegistry.extensionsOf(annotatedType.rawClass)
```
luispollo, line 21: nit:
```suggestion
    if (annotatedType.rawClass in extensionRegistry.baseTypes()) {
      val extensionTypes = extensionRegistry.extensionsOf(annotatedType.rawClass)
```

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/GenericWildcardTypeModelConverter.kt
robfletcher: Yeah, trying to see if we actually need it. I'm going to do some inspection of the API docs & ensure they look correct then either put this back or delete the whole class.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/RegisteredExtensionModelConverter.kt
```diff
@@ -0,0 +1,32 @@
...
+      context.defineSchemaAsOneOf(annotatedType.rawClass, extensionTypes.values.toList())
+        .also { schema ->
+            schema.discriminator = Discriminator()
+              .propertyName("type") // TODO: is this a broken assumption?
```
luispollo, line 26: I have a faint recollection that we may have used a different property name somewhere... In the events class tree maybe? I can try and take a closer look.

robfletcher: I checked all the ones that are actually registered using the `ExtensionRegistry` and the only outlier is `ResourceSpec` which is handled separately anyway and never hits this bit of code.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ErrorSimulationControllerTests.kt
```diff
@@ -3,23 +3,23 @@ package com.netflix.spinnaker.keel.rest
...
 
-@ExtendWith(SpringExtension::class)
 @SpringBootTest(
   classes = [KeelApplication::class, MockEurekaConfiguration::class],
   properties = ["tests.error-simulation=true"],
   webEnvironment = SpringBootTest.WebEnvironment.MOCK
 )
 
 @AutoConfigureMockMvc
+@EnableAutoConfiguration(exclude = [TaskSchedulingAutoConfiguration::class])
```
luispollo, line 22: I wonder if we should make a `DefaultTestConfiguration` class with these common annotations for Spring tests...

robfletcher: Yeah, I was considering making an annotation that groups some things.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/apidocs/ApiDocTests.kt
```diff
@@ -95,26 +93,20 @@ class ApiDocTests : JUnit5Minutests {
...
-      .andExpect(status().isOk)
-      .andReturn()
-      .response
-      .contentAsString
-      .also(::println)
-      .let { jacksonObjectMapper().readTree(it) }
-  }
+    get() = extensionRegistry.extensionsOf<DeliveryArtifact>().values.toList()
 
   // FIXME: the order of the @Beans used to customize type handling for the OpenAPI stuff is causing a stack overflow
```
luispollo, line 99: If you got this to pass, please remove the `FIXME` comment.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/support/ExtensionRegistry.kt
luispollo: Ah. Never mind, then. 🙂

robfletcher:
> This looks great to me, @robfletcher. Thanks for simplifying this stuff. I have a gut feeling that we might even end up using this registry for other purposes with external plugins.

That's really what the intention is. Because for the instance profiles plugin I will be adding a new implementation of `InstanceProvider` and I want it to show up in the API docs without the plugin having to do anything specifically to customize them.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/apidocs/ApiDocTests.kt
robfletcher: I bet if I do the tests will fail again :sob:

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/apidocs/GenericWildcardTypeModelConverter.kt
robfletcher: I diffed the API doc generated with and without this component and there is no meaningful difference

PR #1404, opened 2020-08-06
refactor(plugins): Make ResourceHandler an interface for Spinnaker plugin compatibility

luispollo:
It turns out that abstract classes don't work with the current implementation of the Spinnaker plugin framework because it uses proxy classes which can implement interfaces marked as `SpinnakerExtensionPoint`, but cannot extend an abstract class. This PR refactors `ResourceHandler` into an interface and pulls the common behavior around using resolvers into an abstract `ResolvableResourceHandler` base-class that the native plugin implementations now extend.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
@@ -7,11 +7,8 @@ import com.netflix.spinnaker.keel.api.ResourceKind
 import com.netflix.spinnaker.keel.api.ResourceSpec
 import com.netflix.spinnaker.keel.api.actuation.Task
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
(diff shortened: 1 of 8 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
```diff
@@ -33,32 +30,14 @@ import org.slf4j.LoggerFactory
...
  */
-abstract class ResourceHandler<S : ResourceSpec, R : Any>(
-  private val resolvers: List<Resolver<*>>
-) : SpinnakerExtensionPoint {
+interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
   val name: String
-    get() = javaClass.simpleName
-
-  protected val log: Logger by lazy { LoggerFactory.getLogger(javaClass) }
+    get() = extensionClass.simpleName
```
luispollo, line 35: Recommendation from @jonsie.

PR #1405, opened 2020-08-06
Java 11 Compatibility

robfletcher:
Fixes a couple of things that didn't work with Java 11, and makes GH Actions use Java 11.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,11 +14,14 @@ jobs:
     # Only run this on repositories in the 'spinnaker' org, not on forks.
     if: startsWith(github.repository, 'spinnaker/')
     runs-on: ubuntu-latest
-    steps:
+    strategy:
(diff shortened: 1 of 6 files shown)


luispollo:
Neat! Are you intending to actually switch our version in the gradle scripts?

gal-yardeni:
LGTM 👍  Does it mean we'll need to upgrade the java version to have keel running locally?

robfletcher:
> Neat! Are you intending to actually switch our version in the gradle scripts?

The only reference in our Gradle scripts is the `jvmTarget` for Kotlin compilation. I don't believe the base AMI is using Java 11 yet so it makes sense to keep that at 8.

robfletcher:
> LGTM 👍 Does it mean we'll need to upgrade the java version to have keel running locally?

It should be backward compatible.

Issue #1409, opened 2020-08-06
Springboot 2.2.5 used as part of Kork 7.57 has a bug in loading properties.


nimakaviani:
Kork 7.57 uses SpringBoot 2.2.5 [here](https://github.com/spinnaker/kork/blob/63ff1c7e5133c975519acb1ce70980090581d784/spinnaker-dependencies/spinnaker-dependencies.gradle#L22) which has a [known bug](https://github.com/spring-projects/spring-boot/issues/21050) when loading properties under `additional-location`.

This breaks the defaulting behavior we intend to introduce to keel . A potential solution would be to bump springBoot to 2.2.7 or higher and bump Kork accordingly.

luispollo:
I have a vague memory of us hitting issues with a newer version of Spring recently. Looping in @robfletcher and @cfieber in case that rings a bell for them.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

luispollo:
The issue we ran into when trying to upgrade last time was this: https://github.com/spinnaker/kork/pull/665

I tried again today and ran into the same problem.

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 60 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

PR #1410, opened 2020-08-06
chore(defaulting): set reasonable defaults for services

nimakaviani:
- add reasonable keel configuration defaults under `keel-web/config/keel.yml`
- apply `io.spinnaker.package` to keel-web
- use `additional-location` for additive properties

This work depends on fixing https://github.com/spinnaker/keel/issues/1409 , so let's address that first.

/cc @ezimanyi 
/cc @emjburns

### DIFF
--- a/keel-web/config/keel.yml
+++ b/keel-web/config/keel.yml
@@ -1,0 +1,93 @@
+eureka:
+  enabled: false
+
+keel:
+  resourceCheck:
+    minAgeDuration: 10s
+  artifact-refresh:
+    frequency: PT1M
+  plugins:
+    bakery:
+      enabled: false
+      baseImages:
+        bionic:
+          candidate: bionicbase-x86_64-201904232145-ebs
+          unstable: bionicbase-unstable-x86_64-201904252133-ebs
+          release: bionicbase-x86_64-201904041959-ebs
+        xenial:
+          candidate: xenialbase-x86_64-201904232145-ebs
+          unstable: xenialbase-unstable-x86_64-201904252133-ebs
+          release: xenialbase-x86_64-201904041959-ebs
+          previous: xenialbase-x86_64-201902202219-ebs
+    ec2:
+      enabled: true
+    deliveryConfig:
+      enabled: true
+    titus:
+      enabled: false
+  constraints:
+    manual-judgement:
+      interactive-notifications:
+        enabled: true
+
+front50:
+  enabled: ${services.front50.enabled}
+  baseUrl: ${services.front50.baseUrl:http://localhost:8080}
+clouddriver:
+  enabled: ${services.clouddriver.enabled}
+  baseUrl: ${services.clouddriver.baseUrl:http://localhost:7002}
+echo:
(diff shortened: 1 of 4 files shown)


ezimanyi:
This looks great, definitely on board with having reasonable defaults defined in the service so that users only need to set properties when they want to override them.  (This will also mean that `kleat` doesn't need to have knowledge of defaults, keeping this knowledge within `keel`.)

Of course, will defer to @emjburns as to the actual implementation here and the values chosen for these defaults as I am not very knowledgable about `keel`.

keel-web/config/keel.yml
```diff
@@ -0,0 +1,93 @@
...
+    baseUrl: ${services.fiat.baseUrl:http://localhost:7003}
+
+sql:
+  enabled: true
+  connectionPools:
+    default:
+      jdbcUrl: jdbc:mysql://127.0.0.1:3306/keel?useSSL=false&serverTimezone=UTC
+      user: root
+      password:
+  migration:
```
luispollo, line 63: I'm 99.9% sure we don't use this key.

keel-web/config/keel.yml
```diff
@@ -0,0 +1,93 @@
...
+    ec2:
+      enabled: true
+    deliveryConfig:
+      enabled: true
+    titus:
+      enabled: false
+  constraints:
+    manual-judgement:
+      interactive-notifications:
+        enabled: true
```
luispollo, line 31: This will only work with additional setup in echo and Slack. I'd suggest defaulting to `false`.

keel-web/config/keel.yml
```diff
@@ -0,0 +1,93 @@
...
+  # moved to spinnaker-common
+  #keyStore: /apps/keel/config/all-services-client.p12
+  #trustStore: /apps/keel/config/services-truststore.p12
+  propagateSpinnakerHeaders: true
+  connectTimeoutMs: 30000
+  readTimeoutMs: 59000
+  interceptor.skipHeaderCheck: true
+
+server:
+  port: 7002
```
luispollo, line 78: We use `8087` at Netflix, which I just realized was already claimed by Rosco here: https://spinnaker.io/reference/architecture/

`7002` is also already used by Clouddriver. I'd suggest using a new port and updating the docs.

luispollo:
@nimakaviani Thanks for contributing this! I've no idea why we never created the file in the first place. 🤷‍♂️ 
Overall, looks OK to me, but left you a couple of comments. Let me know what you think.

keel-web/config/keel.yml
nimakaviani: I use 7010 for my local developments and it doesn't seem to collide with other component ports. so will switch to that instead.

keel-web/config/keel.yml
nimakaviani: hmm I remember I had to set it for keel to start up, but that is not the case anymore. maybe I was on an older library version. fixed. thx.

nimakaviani:
fixed, thanks!

PR #1413, opened 2020-08-07
fix(plugins): Fix discovery of supported resource kinds

luispollo:
Replaces previous reliance on a Spring-injected list of `SupportedKind<*>` with querying the `ExtensionRegistry`, such that external plugin-provided resource kinds are included.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
@@ -14,22 +14,16 @@ import org.springframework.stereotype.Component
  * things like reading resources from the database, or parsing JSON.
  */
 @Component
-class ResourceSpecIdentifier @Autowired constructor (
-  private val extensionRegistry: ExtensionRegistry?
-) {
-  private var _kinds: List<SupportedKind<*>>? = null
-
+class ResourceSpecIdentifier(
   private val kinds: List<SupportedKind<*>>
-    get() {
-      return when {
-        _kinds != null -> _kinds!!
-        extensionRegistry != null -> extensionRegistry
-          .extensionsOf<ResourceSpec>()
-          .entries
-          .map { SupportedKind(ResourceKind.parseKind(it.key), it.value) }
-        else -> error("Keel is misconfigured. No resource kinds registered.")
-      }
-    }
+) {
+  @Autowired
+  constructor(extensionRegistry: ExtensionRegistry) :
+    this(extensionRegistry
+      .extensionsOf<ResourceSpec>()
+      .entries
+      .map { SupportedKind(ResourceKind.parseKind(it.key), it.value) }
+    )
 
   fun identify(kind: ResourceKind): Class<out ResourceSpec> =
     kinds.find { it.kind == kind }?.specClass ?: throw UnsupportedKind(kind)
@@ -37,7 +31,5 @@ class ResourceSpecIdentifier @Autowired constructor (
   /**
    * Constructor useful for tests so they can just wire in using varargs.
    */
-  constructor(vararg kinds: SupportedKind<*>) : this(null) {
-    _kinds = kinds.toList()
-  }
+  constructor(vararg kinds: SupportedKind<*>) : this(kinds.toList())
 }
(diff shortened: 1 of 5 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
```diff
@@ -12,14 +14,30 @@ import org.springframework.stereotype.Component
...
+
+  private val kinds: List<SupportedKind<*>>
+    get() {
+      return when {
+        _kinds != null -> _kinds!!
+        extensionRegistry != null -> extensionRegistry
+          .extensionsOf<ResourceSpec>()
+          .entries
+          .map { SupportedKind(ResourceKind.parseKind(it.key), it.value) }
+        else -> error("Keel is misconfigured. No resource kinds registered.")
```
luispollo, line 30: @robfletcher this was the best I could come up with, but it kinda sucks. I didn't want to rewrite all the tests that used the `vararg` constructor... Let me know if you have a better idea.

robfletcher: I'd go with keeping the original constructor but moving the `@Autowired` annotation to a new one that takes `ExtensionRegistry` as a parameter.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
```diff
@@ -34,6 +33,8 @@ class KeelConfigurationFinalizer(
 
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
+  private val kinds: List<SupportedKind<*>> by lazy { resourceHandlers.map { it.supportedKind } }
```
robfletcher, line 36: Shouldn't we get this from the `ExtensionRegistry`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
luispollo: I thought of that, but wasn't sure that the `@Component` annotation on the class/default constructor wouldn't give it priority when auto-wiring...

keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
luispollo: We can. It's just used to print the list of kinds at the very end, and in this class in particular the registry is populated off the same list of resource handlers, so I didn't bother.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
robfletcher: Easy enough to test

keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
robfletcher: Yeah, it's not a big deal. I just feel like if we want to treat `ExtensionRegistry` as the source of truth for this kind of thing, we might as well use it everywhere. Avoids people copy 'n pasting this if nothing else.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
luispollo: It works. Sorry for being lazy. 😄

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/config/EC2Config.kt
```diff
@@ -47,22 +42,6 @@ import org.springframework.context.annotation.Configuration
 @EnableConfigurationProperties(CanaryConstraintConfigurationProperties::class)
 @ConditionalOnProperty("keel.plugins.ec2.enabled")
 class EC2Config {
-
-  @Bean
```
luispollo, line 51: These are no longer needed now that we pull the supported kinds from the `ExtensionRegistry`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
```diff
@@ -12,9 +14,17 @@ import org.springframework.stereotype.Component
...
+class ResourceSpecIdentifier(
   private val kinds: List<SupportedKind<*>>
 ) {
+  @Autowired
+  constructor(extensionRegistry: ExtensionRegistry) :
+    this(extensionRegistry
+      .extensionsOf<ResourceSpec>()
+      .entries
+      .map { SupportedKind(ResourceKind.parseKind(it.key), it.value) }
+    )
```
robfletcher, line 26: Yeah, I like the simplicity of that

PR #1414, opened 2020-08-07
fix(constraints): Fix query to load constraint state

luispollo:
Fixes the query to get constraint state such that the artifact reference is only used when actually passed in.

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
@@ -568,11 +568,18 @@ class SqlDeliveryConfigRepository(
         .where(
           ENVIRONMENT_ARTIFACT_CONSTRAINT.ENVIRONMENT_UID.eq(environmentUID),
           ENVIRONMENT_ARTIFACT_CONSTRAINT.ARTIFACT_VERSION.eq(artifactVersion),
-          (ENVIRONMENT_ARTIFACT_CONSTRAINT.ARTIFACT_REFERENCE.eq(artifactReference)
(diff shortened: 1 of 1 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -568,11 +568,17 @@ class SqlDeliveryConfigRepository(
...
-            .or(ENVIRONMENT_ARTIFACT_CONSTRAINT.ARTIFACT_REFERENCE.isNull)),
           ENVIRONMENT_ARTIFACT_CONSTRAINT.TYPE.eq(type)
         )
+        .let {
+          if (artifactReference != null) {
+            it.and(
+              ENVIRONMENT_ARTIFACT_CONSTRAINT.ARTIFACT_REFERENCE.eq(artifactReference)
+            )
+          } else {
+            it
```
gal-yardeni, line 579: few things:
1. can you please add a comment here if possible? 
2. will else {it} handle the `null` case?
3. this fix should be added to line 929 as well :)

luispollo: 1. Sure.
2. If the artifact reference is `null`, we simply won't include it in the query, so it won't matter what's in the DB (same as previous behavior, i.e. backwards compatible).
3. Thanks, will do!

luispollo: Actually, for number 3, I don't think we can do the same thing because otherwise there's no filtering based on artifact in the query. In fact, I think that argument should not be nullable in that method.

PR #1416, opened 2020-08-07
feat(plugins): Introduce KeelServiceSdk for external plugins

luispollo:
The goal is to expose a few of the core services provided by keel such that plugins can do things like look up versions of artifacts and launch Orca tasks.

### DIFF
--- a/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/plugins/KeelServiceSdk.kt
+++ b/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/plugins/KeelServiceSdk.kt
@@ -1,0 +1,12 @@
+package com.netflix.spinnaker.keel.plugins
+
+import com.netflix.spinnaker.keel.api.actuation.TaskLauncher
+import com.netflix.spinnaker.keel.persistence.KeelRepository
+
+/**
+ * A simple SDK that can be consumed by external plugins to access core Keel functionality.
+ */
+interface KeelServiceSdk {
+  val repository: KeelRepository
+  val taskLauncher: TaskLauncher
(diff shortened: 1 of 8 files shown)


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/plugins/KeelServiceSdk.kt
```diff
@@ -0,0 +1,12 @@
+package com.netflix.spinnaker.keel.plugins
+
+import com.netflix.spinnaker.keel.api.actuation.TaskLauncher
+import com.netflix.spinnaker.keel.persistence.KeelRepository
+
+/**
+ * A simple SDK that can be consumed by external plugins to access core Keel functionality.
+ */
+interface KeelServiceSdk {
```
luispollo, line 9: @robzienert I just realized this needs to go in `keel-api` to be accessible to plugins, right?

luispollo: @robfletcher With that, I'll need to move `KeelRepository` from `keel-core` to `keel-api` as well. It's an interface, but it's got a bunch of default method implementations... 😕 How do you feel about that?

luispollo: Actually, I think I'll define a read-only interface for plugin access.

luispollo:
@robfletcher @robzienert Sorry for the forced merge. I just wanted to check if this will work in the plugin and there's no impact anywhere in keel.

PR #1417, opened 2020-08-08
fix(constraints): Fix the fix for reading/updating constraints

luispollo:
Follow-up to #1414. This should do it.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/constraints/ConstraintState.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/constraints/ConstraintState.kt
@@ -26,6 +26,8 @@ data class ConstraintState(
   val deliveryConfigName: String,
   val environmentName: String,
   val artifactVersion: String,
+  // FIXME: this should not be nullable as it's the only field uniquely identifying the artifact,
+  //  but would need a DB migration.
   val artifactReference: String? = null,
   val type: String,
   val status: ConstraintStatus,
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
@@ -507,6 +507,7 @@ class SqlDeliveryConfigRepository(
               .set(ENVIRONMENT_ARTIFACT_CONSTRAINT.JUDGED_AT, MySQLDSL.values(ENVIRONMENT_ARTIFACT_CONSTRAINT.JUDGED_AT))
               .set(ENVIRONMENT_ARTIFACT_CONSTRAINT.COMMENT, MySQLDSL.values(ENVIRONMENT_ARTIFACT_CONSTRAINT.COMMENT))
               .set(ENVIRONMENT_ARTIFACT_CONSTRAINT.ATTRIBUTES, MySQLDSL.values(ENVIRONMENT_ARTIFACT_CONSTRAINT.ATTRIBUTES))
+              .set(ENVIRONMENT_ARTIFACT_CONSTRAINT.ARTIFACT_REFERENCE, state.artifactReference)
               .execute()
 
             txn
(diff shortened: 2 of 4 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -507,6 +507,7 @@ class SqlDeliveryConfigRepository(
               .set(ENVIRONMENT_ARTIFACT_CONSTRAINT.JUDGED_AT, MySQLDSL.values(ENVIRONMENT_ARTIFACT_CONSTRAINT.JUDGED_AT))
               .set(ENVIRONMENT_ARTIFACT_CONSTRAINT.COMMENT, MySQLDSL.values(ENVIRONMENT_ARTIFACT_CONSTRAINT.COMMENT))
               .set(ENVIRONMENT_ARTIFACT_CONSTRAINT.ATTRIBUTES, MySQLDSL.values(ENVIRONMENT_ARTIFACT_CONSTRAINT.ATTRIBUTES))
+              .set(ENVIRONMENT_ARTIFACT_CONSTRAINT.ARTIFACT_REFERENCE, state.artifactReference)
```
luispollo, line 510: This was missing and so the column was never updated.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -571,10 +572,11 @@ class SqlDeliveryConfigRepository(
...
         )
         .let {
-          // For backwards-compatibility, only use the artifact reference in the query when passed in
+          // For backwards-compatibility, only use the artifact reference in the query when passed in, but also match
+          // rows with a missing reference.
           if (artifactReference != null) {
-            it.and(
-              ENVIRONMENT_ARTIFACT_CONSTRAINT.ARTIFACT_REFERENCE.eq(artifactReference)
+            it.and(ENVIRONMENT_ARTIFACT_CONSTRAINT.ARTIFACT_REFERENCE.eq(artifactReference)
+              .or(ENVIRONMENT_ARTIFACT_CONSTRAINT.ARTIFACT_REFERENCE.isNull)
```
luispollo, line 579: This should allow updating those rows in the DB which currently have a `NULL` in the column.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
gal-yardeni: Can you elaborate on why this needs to be set here? the setting is done at line 503, and the artifact reference should not be changed when updating the state.

gal-yardeni:
Thanks for taking a look at this! I'm still not sure it's the right solution though. I suggest we will add some tests maybe to test this scenario if it's possible. I'll be back on Tuesday and continue to investigate it then.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: Because the `UNIQUE KEY` on the table is on `environment_uid`,`type` and `artifact_version` only, and with the other change below the record will load if `artifact_reference` is `NULL` (even though we actually have a reference for the artifact) and so will cause a duplicate exception here, so we want to take this opportunity to update the column with the right value.

luispollo:
> Thanks for taking a look at this! I'm still not sure it's the right solution though. I suggest we will add some tests maybe to test this scenario if it's possible. I'll be back on Tuesday and continue to investigate it then.

Sounds good. I'll try to write some tests.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: P.S. The alternative to this would have been to write a DB migration to fill in the missing reference, since it's mandatory in practice, which I feel should have been done in the original PR where this change was introduced.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/constraints/ConstraintState.kt
```diff
@@ -26,6 +26,8 @@ data class ConstraintState(
   val deliveryConfigName: String,
   val environmentName: String,
   val artifactVersion: String,
```
emjburns, line 28: FYI this was nullable to allow artifact reference to be rolled out originally. Do any recent artifacts have this, or is it only a couple of old ones? The code should set reference to name if it's not passed in upon submission of the delivery config

PR #1421, opened 2020-08-10
fix(core): Delay usage of extension registry when identifying resource types

luispollo:
We were reading the supported resource kinds too early in the constructor. This PR delays that until the list is actually used to give a chance for the configuration finalizer to register the types.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
@@ -6,7 +6,6 @@ import com.netflix.spinnaker.keel.api.plugins.SupportedKind
 import com.netflix.spinnaker.keel.api.plugins.UnsupportedKind
 import com.netflix.spinnaker.keel.api.support.ExtensionRegistry
 import com.netflix.spinnaker.keel.api.support.extensionsOf
-import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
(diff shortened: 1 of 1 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/resources/ResourceSpecIdentifier.kt
```diff
@@ -16,22 +15,19 @@ import org.springframework.stereotype.Component
...
+    get() = extensionsOf<ResourceSpec>()
       .entries
       .map { SupportedKind(ResourceKind.parseKind(it.key), it.value) }
-    ) {
-    log.debug("Registered the following kinds from extension registry: ${kinds.map { it.kind }.joinToString()}")
-  }
 
   fun identify(kind: ResourceKind): Class<out ResourceSpec> =
-    kinds.find { it.kind == kind }?.specClass ?: throw UnsupportedKind(kind)
+    (extensionRegistry?.supportedKinds ?: kinds).find { it.kind == kind }?.specClass ?: throw UnsupportedKind(kind)
```
luispollo, line 30: This gives priority to the extension registry which is auto-wired by default, and falls back on the list of explicit supported kinds which is only used in tests.

lorin: I recommend adding this information as a comment

luispollo: Done!

PR #1426, opened 2020-08-12
Fix JOOQ generation under Java 11

robfletcher:
When using the Gradle dæmon the JOOQ generation task generates its .java files to the wrong place. This fixes it by ensuring `jooqConfig.xml` contains a fully-qualified path.


keel-sql/keel-sql.gradle.kts
```diff
@@ -88,6 +89,18 @@ jooqModelator {
   dockerContainerPort = 3306
 }
 
+// expand properties in jooqConfig.xml so it gets a fully-qualified directory to generate into
+tasks.withType<ProcessResources> {
+  filesMatching("jooqConfig.xml") {
+    expand(project.properties)
```
luispollo, line 95: Smart! 👍

PR #1427, opened 2020-08-12
feat(api): Enable parsing of YAML anchors and aliases

### LINKED ISSUE #1398: jackson doesn't properly parse anchors and aliases in the managed delivery manifest
I am creating this issue as a followup to our conversation during the SIG meeting. We noticed that managed delivery manifests that use anchors and aliases are not properly parsed even though the yaml document is totally valid.

Here is an example manifest.

```name: my-manifest
application: nimakspin
serviceAccount: sekret-service-account@spinnaker-test.net
artifacts: []
hasManagedResources: true
environments:
  - name: testing
    hasManagedResources: true
    resources:
    - &secgrp 
      kind: "ec2/security-group@v1"
      spec:
        moniker:
          app: "nimakspin"
        locations:
          account: "my-aws-account"
          vpc: "eksctl-demo-fargate-cluster/VPC"
          regions:
          - name: "us-west-2"
        description: "test LoadBalancer securityGroup by ALB Ingress Controller"
        inboundRules:
        - protocol: "TCP"
          portRange:
            startPort: 80
            endPort: 80
          blockRange: "0.0.0.0/0"
        - protocol: "TCP"
          portRange:
            startPort: 22
            endPort: 22
          blockRange: "0.0.0.0/0"
  - name: prod
    hasManagedResources: true
    resources:
    - *secgrp
```

The problem appears to be with Jackson failing to parse aliases and anchors where SnakeYAML parses the yaml document properly.

@luispollo suggested double passing of the yaml document, first into SnakeYAML and then into Jackson in order for the parsing of the manifest to work properly. That can be a relatively easy fix for this.

luispollo:
Fixes #1398.

Cc: @nimakaviani

### DIFF
--- a/keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigYamlParsingFilterTests.kt
+++ b/keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigYamlParsingFilterTests.kt
@@ -1,0 +1,126 @@
+package com.netflix.spinnaker.keel.rest
+
+import com.fasterxml.jackson.databind.jsontype.NamedType
+import com.fasterxml.jackson.module.kotlin.readValue
+import com.netflix.spinnaker.keel.api.ec2.EC2_SECURITY_GROUP_V1
+import com.netflix.spinnaker.keel.api.ec2.SecurityGroupSpec
+import com.netflix.spinnaker.keel.api.plugins.SupportedKind
+import com.netflix.spinnaker.keel.core.api.SubmittedDeliveryConfig
+import com.netflix.spinnaker.keel.core.api.SubmittedResource
+import com.netflix.spinnaker.keel.ec2.jackson.registerKeelEc2ApiModule
+import com.netflix.spinnaker.keel.test.configuredTestObjectMapper
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import io.mockk.Runs
+import io.mockk.every
+import io.mockk.just
+import io.mockk.mockk
+import io.mockk.slot
+import io.mockk.verify
+import javax.servlet.FilterChain
+import javax.servlet.ServletRequest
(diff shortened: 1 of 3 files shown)


keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigYamlParsingFilterTests.kt
```diff
@@ -0,0 +1,126 @@
...
+                  portRange:
+                    startPort: 22
+                    endPort: 22
+                  blockRange: "0.0.0.0/0"
+          - name: prod
+            hasManagedResources: true
+            resources:
+            - kind: ec2/security-group@v1
+              spec:
+                << : *secgrp
```
luispollo, line 77: @nimakaviani Note that this is slightly different than your example because we do additional validation of the spec and verify that resource IDs (which derive from the `moniker` field) are unique. IMO this is a good-enough compromise, but wanted to give you a heads-up.

nimakaviani: looks great thanks! 

agreed. parsing is independent of the validation. As long as parsing works as expected, plugins can to decide at what level to use anchors and aliases and how to do the validations. so good enough IMO too. thanks again.

robfletcher:
I'm kind of stumped as to why this doesn't work out of the box given Jackson's YAML module uses SnakeYAML under the hood. Have we raised an issue on that repo?

robfletcher:
This seems relevant: https://github.com/FasterXML/jackson-dataformats-text/issues/98

luispollo:
> This seems relevant: [FasterXML/jackson-dataformats-text#98](https://github.com/FasterXML/jackson-dataformats-text/issues/98)

Yup. I referenced that exact same link in #1398 🙂

PR #1429, opened 2020-08-13
feat(artifactMetadata): initial implementation for fetching artifact metadata

gal-yardeni:
This is an initial implementation for fetching additional artifact metadata, like build and git details.

**_How does it work?_**
- For each type of artifact, after being saved in the DB in `ARTIFACT_VERSIONS` table, we will call `boost` endpoint (powered by `rocket`), with the build number and commit hash (which all artifacts should include).
- Build number + commit hash will be added to all (Netflix) artifact types (for npm/debians, it'll be as a part of the metadata, for docker -- we will fetch it from Titus registry via clouddriver cache).
- The response from `boost` will be a generic `Build` type, and we will convert it into `BuildMetadata` and `GitMetadata` when fetching the needed fields.
- The `BuildMetadata` and `GitMetadata` will be stored in the `ARTIFACT_VERSIONS` table.
- When retrieving the artifact information (to the UI or API), we will look up the table and returned the data saved there (or return the `hacky` way we currently have).

TODO:
- [x] resolve the circular dependency: when wiring the `repository` into the individual implementations of `ArtifactSupplier`, I'm getting a circular dependency error as `deliveryConfigRepository` has `artifactSuppliers: List<ArtifactSupplier<*, *>>` wire into it.
- [x] add tests
- [x] implement `/ci/builds` to actually call `boost`, and parse the response as we need in gate
- [ ] add `buildNumber` and `commitId` into debian (and npm?) artifact metadata in igor
- [x] once titus push their change to add metadata into docker, add it to clouddriver + echo + keel
- [ ] UI: figure out which fields we want to show to users


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/GitMetadata.kt
```diff
@@ -1,9 +1,13 @@
...
+ * The git metadata of an artifact.
  */
 data class GitMetadata(
-  val commit: String
+  val commit: String, // commit hash, can be short or long sha
+  val author: String? = null,
+  val linkToCommit: String? = null,
+  val repoName: String? = null, // the repository name, like "myApp"
+  val projectName: String? = null, // the project name, like SPKR
+  val commitMessage: String? = null
```
emjburns, line 12: this could get long, we will need a limit to what we store here

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -82,7 +88,10 @@ class ArtifactListener(
       if (latestVersion != null) {
         var status = artifactSupplier.getReleaseStatus(latestArtifact)
         log.debug("Storing latest version {} (status={}) for registered artifact {}", latestVersion, status, artifact)
-        repository.storeArtifact(artifact.name, artifact.type, latestVersion, status)
+
+        val artifactMetadata = artifactMetadata(latestArtifact)
```
emjburns, line 92: I think we should save the artifact, then fetch the metadata, then then re-save. I think this because theres a potential that those metadata calls could fail, and that shouldn’t lead to a failure to save the artifact (which would mean no one would be able to deploy that version == bad). it should just lead to less details (and we could probably show this in the UI somehow - maybe by recording on the artifact version if it’s pre or post metadata fetch?)

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -104,21 +104,21 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
     with(subject) {
       register(artifact1)
       setOf(version1, version2, version3).forEach {
-        store(artifact1, it, SNAPSHOT)
+        store(artifact1, it, SNAPSHOT, null)
```
emjburns, line 107: is it worth making the default value for this last parameter `null` so you don't have to update these calls?

keel-igor/src/main/kotlin/com/netflix/spinnaker/igor/BuildService.kt
```diff
@@ -0,0 +1,25 @@
...
+   * Retrieves build information by commit id and build number from a ci service.
+   *
+   * @param projectKey The "project" within the SCM system where the repository exists, which can be a user's personal
+   *        area (e.g. "SPKR", "~<username>")
+   * @param repoSlug The repository name (e.g. "myapp")
+   * @param commitId The commit id.
+   * @param buildNumber the build number .
+   */
+  // TODO: implement this call in igor
+  @GET("/ci/builds")
```
emjburns, line 18: I think you'll need to put in retries and handle error from this call. For example, if the build is not found the convention in spinnaker services is to throw a 404. you'll need to catch that and do the appropriate thing

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
```diff
@@ -0,0 +1,36 @@
...
+  fun getArtifactMetadata(
+    commitId: String?,
+    buildNumber: String?
+  ): ArtifactMetadata? {
+
+    if (commitId == null || buildNumber == null) {
+      return null
+    }
+      val builds = runBlocking {
+        buildService.getArtifactMetadata(commitId = commitId, buildNumber = buildNumber)
```
emjburns, line 27: retries might be good here, as well as error handling

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -117,6 +121,18 @@ class SqlArtifactRepository(
     return objectMapper.writeValueAsString(details)
   }
 
+  private fun BuildMetadata.detailsAsJson(): String {
+    val details = objectMapper.convertValue<Map<String, Any?>>(this)
```
emjburns, line 125: Do you need to convert to a map first? I think write as string will do that for you.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -76,23 +78,31 @@ class DebianArtifactSupplier(
   }
```
emjburns, line 79: I feel like instead of reading from the repository, this is where the calls to outside services should be?

Then, when this is called the artifact details get returned, and then saved in the database for use elsewhere. Then you don't have to have keelrepository dependency here

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
gal-yardeni: Done!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: Done!

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
gal-yardeni: Done!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/Build.kt
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * An immutable data class that represents a published software artifact in the Spinnaker ecosystem.
+ *
+ * This class mirrors [com.netflix.spinnaker.igor.build.model.GenericBuild], but without all the Jackson baggage.
+ */
+
+data class Build(
+  val building: Boolean = false,
+  val fullDisplayName: String? = null,
+  val name: String? = null,
```
luispollo, line 12: Does it makes sense to have an object of this class without at least one of these two fields present?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/Build.kt
```diff
@@ -0,0 +1,25 @@
...
+  val number: Int = 0,
+  val duration: Int? = null,
+  /** String representation of time in nanoseconds since Unix epoch  */
+   val timestamp: String? = null,
+
+  val result: Any? = null,
+  val artifacts: List<Any>? = null,
+  val testResults: List<Any>? = null,
```
luispollo, line 20: Curious about the use of `Any` here. Do we not know the structure of these objects?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
```diff
@@ -1,9 +1,14 @@
...
  */
 data class BuildMetadata(
-  val id: Int
+  val id: Int, // should be converted to number. Needs to be removed
+  val number: String? = null, // build number, currently duplicate value to id
+  val uid: String? = null, // a unique id for the build, which is generated by the CI service
+  val jobUrl: String? = null,
+  val jobName: String? = null,
+  val completedAt: String? = null,
+  val startedAt: String? = null
```
luispollo, line 13: I'd think like none of these fields should be nullable...

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
```diff
@@ -84,6 +85,14 @@ interface ArtifactSupplier<A : DeliveryArtifact, V : VersioningStrategy> : Spinn
...
    */
   fun getGitMetadata(artifact: PublishedArtifact, versioningStrategy: VersioningStrategy): GitMetadata? = null
+
+  /**
+   * Given a [PublishedArtifact] supported by this [ArtifactSupplier],
+   * return the [ArtifactMetadata] for the artifact, if available.
+   *
+   * This function is currently expected to make calls to CI systems.
+   */
+  fun getArtifactMetadata(artifact: PublishedArtifact): ArtifactMetadata? = null
```
luispollo, line 95: My feeling is that this should probably replace `getGitMetadata` and `getBuildMetadata` on this interface... In case what prevented you from doing that was the passing of `versioningStrategy` as parameter above, the only reason I did that was because we were extracting the metadata from the artifact version string, but once we expect artifacts to carry better metadata, we could probably remove that parameter.

luispollo: P.S. Given the expectation that the call will call other services, it should probably be a `suspend` function.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -59,6 +60,8 @@ class ArtifactListener(
             .also { wasAdded ->
               if (wasAdded) {
                 publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.artifactType))
+                // send an event when the artifact is stored, so we can fetch its metadata
+                publisher.publishEvent(ArtifactSaved(artifact.copy(version = version), status))
```
luispollo, line 64: Any reason not to just have the event listener function subscribe to both events, instead of adding a second event here?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactStoredListener.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.context.event.EventListener
+import org.springframework.stereotype.Component
+
+@Component
+class ArtifactStoredListener(
+  private val repository: KeelRepository,
+  private val artifactSuppliers: List<ArtifactSupplier<*, *>>
+) {
+
+  @EventListener(ArtifactSaved::class)
```
luispollo, line 16: Here you could add `ArtifactVersionUpdated::class` as I mentioned above.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifactSupplier.kt
```diff
@@ -71,10 +74,16 @@ class NpmArtifactSupplier(
    * Extracts the commit hash from the version string using the Netflix semver convention.
    */
   override fun getGitMetadata(artifact: PublishedArtifact, versioningStrategy: VersioningStrategy): GitMetadata? {
+
     return NetflixSemVerVersioningStrategy.getCommitHash(artifact)
       ?.let { GitMetadata(it) }
   }
 
+  override fun getArtifactMetadata(artifact: PublishedArtifact): ArtifactMetadata? {
```
luispollo, line 82: The implementation of all 3 `ArtifactSupplier` classes for this function is the same. Could we make it the default implementation in the interface? You can define the dependency on `ArtifactMetadataService` there as well.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -323,12 +323,23 @@ class ApplicationService(
   ): ArtifactVersionSummary {
     val artifactSupplier = artifactSuppliers.supporting(artifact.type)
     val publishedArtifact = artifact.toSpinnakerArtifact(version)
+    val artifactStatus = artifactSupplier.getReleaseStatus(publishedArtifact)
+    val buildMetadata = repository.getArtifactBuildMetadata(artifact.name, artifact.type, version, artifactStatus)
+    val gitMetadata = repository.getArtifactGitMetadata(artifact.name, artifact.type, version, artifactStatus)
```
luispollo, line 328: We could simplify this a bit by having a single method `repository.getArtifactMetadata()` and unpacking the build and git metadata here instead of in all repository implementations.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -323,12 +323,23 @@ class ApplicationService(
...
     return ArtifactVersionSummary(
       version = version,
       environments = environments,
       displayName = artifactSupplier.getVersionDisplayName(publishedArtifact),
-      build = artifactSupplier.getBuildMetadata(publishedArtifact, artifact.versioningStrategy),
-      git = artifactSupplier.getGitMetadata(publishedArtifact, artifact.versioningStrategy)
+
+      // first attempt to fetch the artifact metadata from the DB, then fallback to the default if not found
```
luispollo, line 334: Ah, now I see why you kept those functions. In this case, I'd suggest renaming them so it's clear that they're not really going to retrieve additional data from anywhere, but instead just try to parse data out of the artifact. Maybe `ArtifactSupplier.parseBuildMetadata()` and `parseGitMetadata()` (or `extract` if you like that better)?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
luispollo: Update: after reading more of the PR I understand why you kept the other 2 functions. I suggested new names in another comment.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
luispollo: +1. Retries on `50x` and fail gracefully on `404`.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+@Component
+class ArtifactMetadataService(
+  private val buildService: BuildService
+) {
+
+  fun getArtifactMetadata(
+    commitId: String?,
+    buildNumber: String?
+  ): ArtifactMetadata? {
```
luispollo, line 21: Generally the pattern I see in other services is to have non-nullable parameters to these calls, and have the caller handle nullability, as it doesn't really make sense to make the call in the first place if you don't have those parameters available.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Provides functionality to convert build metadata, which is coming from internal service, to artfiact metadata (via igor).
+ */
+@Component
+class ArtifactMetadataService(
+  private val buildService: BuildService
+) {
+
+  fun getArtifactMetadata(
```
luispollo, line 18: Should also be `suspend`, I think. Then you don't have to call `runBlocking` below.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
```diff
@@ -0,0 +1,36 @@
...
+    if (commitId == null || buildNumber == null) {
+      return null
+    }
+      val builds = runBlocking {
+        buildService.getArtifactMetadata(commitId = commitId, buildNumber = buildNumber)
+    }
+
+    // TODO: implement the exact conversion between build service response to the objects we are expecting
+    return ArtifactMetadata(
+      buildMetadata = BuildMetadata(id = builds[0].number),
```
luispollo, line 32: `TODO`: check size of `builds` array before indexing into the first element.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/Build.kt
```diff
@@ -0,0 +1,25 @@
+package com.netflix.spinnaker.keel.api.artifacts
+
+/**
+ * An immutable data class that represents a published software artifact in the Spinnaker ecosystem.
+ *
+ * This class mirrors [com.netflix.spinnaker.igor.build.model.GenericBuild], but without all the Jackson baggage.
+ */
+
+data class Build(
```
luispollo, line 9: This could probably go in `keel-igor`...

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/Build.kt
gal-yardeni: we don't. This object is a copy from Igor `GenericBuild` object. My goal is to have this object to suit different CI providers, so I don't want to enforce any kind of structure here.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/Build.kt
gal-yardeni: done

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/Build.kt
gal-yardeni: what do you mean? which 2 fields?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
gal-yardeni: Done - added suspend to the `getArtifactMetadata` + rename the build/git functions to `getDefaultXXXmetadata`

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
gal-yardeni: I don't understand this comment - lets talk about it offline

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifactSupplier.kt
gal-yardeni: lets talk about this offline

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: changed to `getDefaultXXMetadata` -- let me know if it makes sense.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
gal-yardeni: done!

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
gal-yardeni: Do you think it'll be better to check the nullability prior to making this call? I can move it to the actual suppliers.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/Build.kt
luispollo: Sorry, the ones on the lines selected for the comment, `fullDisplayName` and `name`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: I think `getDefaultXyzMetadata` doesn't really give any hints about what the function is expected to do, as "default" doesn't have any intrinsic meaning here, to the best of my knowledge. I'd prefer something like `extractXyzMetadata` as suggested earlier, but up to you.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
luispollo: Yes, that's what I was trying to say.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
```diff
@@ -1,9 +1,14 @@
 package com.netflix.spinnaker.keel.api.artifacts
 
 /**
- * todo eb: other information should go here, like a link to the jenkins build. But that needs to be done
- * in a scalable way. For now, this is just a minimal container for information we can parse from the version.
+ * The build metadata of an artifact.
  */
 data class BuildMetadata(
```
gal-yardeni, line 6: @erikmunson I'm planning to extend `BuildMetadata` and `GitMetadata` below with those fields. I'm sure we will not be showing all of them -- but please let me know if you think something is missing here. The `uid` in this class is the boost build id we are using to fetch the logs -- it will be the path for linking the logs, potentially in the bottom part of the tab when we are showing CI information for the artifact.

keel-igor/src/main/kotlin/com/netflix/spinnaker/igor/BuildService.kt
gal-yardeni: I thought about it, but I _think_ for any kind of exception, I will want to retry as the data might be there later on from some reason.. let me know if it makes sense.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
gal-yardeni: done!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/GitMetadata.kt
emjburns: Ya, I think we should limit because we don't want any SQL errors like "overflow" or whatever it gives us. We can just truncate the commitMessage before we save it

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Provides functionality to convert build metadata, which is coming from internal service, to artifact metadata (via igor).
+ */
+@Component
+class ArtifactMetadataService(
+  private val buildService: BuildService
+) {
+
+  @Retry(name = "getArtifactMetadata", fallbackMethod = "getFallback")
```
gal-yardeni, line 21: @luispollo need your input here regarding the `retry` -- I've configured it in `keel.yml` but haven't tested it yet. let me know if you think something is missing.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
```diff
@@ -64,26 +67,28 @@ class DockerArtifactSupplier(
...
-    }
+  override fun parseDefaultGitMetadata(artifact: PublishedArtifact, versioningStrategy: VersioningStrategy): GitMetadata? {
+      if (versioningStrategy.hasCommit()) {
+        return GitMetadata(commit = artifact.version.substringAfterLast("."))
+      }
     return null
   }
 
+  override suspend fun getArtifactMetadata(artifact: PublishedArtifact): ArtifactMetadata? {
+    return super.getArtifactMetadataInternal(artifact)
```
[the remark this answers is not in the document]
gal-yardeni: I couldn't manage to get it to work without overriding `getArtifactMetadata` in each class.. @luispollo do you might have any ideas here?

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
gal-yardeni: @luispollo introduced it to me! it supposed to save you a bunch of code lines

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -75,26 +78,29 @@ class DebianArtifactSupplier(
     }
   }
 
-  override fun getBuildMetadata(artifact: PublishedArtifact, versioningStrategy: VersioningStrategy): BuildMetadata? {
+  override fun parseDefaultBuildMetadata(artifact: PublishedArtifact, versioningStrategy: VersioningStrategy): BuildMetadata? {
```
luispollo, line 81: I like the new names.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -75,26 +78,29 @@ class DebianArtifactSupplier(
...
     // attempt to parse helpful info from the appversion.
-    // todo(eb): replace, this is brittle
     val appversion = AppVersion.parseName(artifact.version)
     if (appversion?.commit != null) {
       return GitMetadata(commit = appversion.commit)
     }
     return null
   }
 
+  override suspend fun getArtifactMetadata(artifact: PublishedArtifact): ArtifactMetadata? {
```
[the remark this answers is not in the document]
luispollo: Doesn't look like it.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -75,26 +78,29 @@ class DebianArtifactSupplier(
...
+    return super.getArtifactMetadataInternal(artifact)
+  }
+
```
luispollo, line 103:
```suggestion
```

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BaseArtifactSupplier.kt
```diff
@@ -0,0 +1,23 @@
+package com.netflix.spinnaker.keel.artifacts
+
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactMetadata
+import com.netflix.spinnaker.keel.api.artifacts.PublishedArtifact
+import com.netflix.spinnaker.keel.services.ArtifactMetadataService
+import org.slf4j.LoggerFactory
+
+abstract class BaseArtifactSupplier (
```
luispollo, line 8: Consider documenting the class.

luispollo: Actually, I think this class may be unnecessary. It's a bit misleading in that it doesn't implement the `ArtifactSupplier` interface despite the name. It looks like all you wanted to do was to reuse the `getArtifactMetadataInternal` function, so you could define that as an extension function in a kotlin module (not a class), and import that function in the `ArtifactSupplier` implementations. I don't think this is a blocker to merging, but in my understanding it would be a more idiomatic way to do it in kotlin.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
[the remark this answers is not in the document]
luispollo: The reason it doesn't work is that `BaseArtifactSupplier` does not implement `ArtifactSupplier` as I pointed out above. The function called on the base class is not the same as the interface override (note the `Internal` suffix). The overrides are expected in these implementations of the interface, but I suggested a different way of reusing that "internal" function above.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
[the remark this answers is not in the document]
luispollo: Yes, because it's not the same function it's calling on `super` (note the `Internal` suffix in the name).

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BaseArtifactSupplier.kt
luispollo: The other, perhaps better alternative would be to make it an actual base class that implements the interface. Then you could have a default implementation of `getArtifactMetadata` in the abstract class and remove the implementations from all the sub-classes.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplierTests.kt
```diff
@@ -98,18 +102,27 @@ internal class DockerArtifactSupplierTests : JUnit5Minutests {
...
       test("returns build metadata based on tag when available") {
-        expectThat(dockerArtifactSupplier.getBuildMetadata(latestArtifact, DockerVersioningStrategy(SEMVER_JOB_COMMIT_BY_SEMVER)))
+        expectThat(dockerArtifactSupplier.parseDefaultBuildMetadata(latestArtifact, DockerVersioningStrategy(SEMVER_JOB_COMMIT_BY_SEMVER)))
           .isEqualTo(BuildMetadata(id = 1182))
-        expectThat(dockerArtifactSupplier.getBuildMetadata(latestArtifact, DockerVersioningStrategy(INCREASING_TAG)))
+        expectThat(dockerArtifactSupplier.parseDefaultBuildMetadata(latestArtifact, DockerVersioningStrategy(INCREASING_TAG)))
           .isNull()
       }
+
+      //TODO[gyardeni]: enable this test once will have docker build number + commit id
```
luispollo, line 118: I think a better way to do this would be to remove the comments and add a `SKIP -` in front of `test()`.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
luispollo: +1!

keel-core/src/main/java/com/netflix/rocket/semver/shaded/NetflixVersionComparator.java
```diff
@@ -4,7 +4,7 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-public class DebianVersionComparator implements Comparator<String> {
+public class NetflixVersionComparator implements Comparator<String> {
```
luispollo, line 7: Good call!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
```diff
@@ -50,3 +52,8 @@ data class ArtifactCheckTimedOut(
   val name: String,
   val deliveryConfigName: String?
 ) : TelemetryEvent()
+
+data class ArtifactSaved(
```
luispollo, line 56: Is this really a `TelemetryEvent`? My guess was that those were only events that we were using to collect metrics.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
luispollo: @gal-yardeni This should be relatively straightforward to test if you configure your igor endpoint to point to a mock server that returns error codes. There's examples of using `MockWebServer` in the tests (see [here](https://github.com/spinnaker/keel/blob/master/keel-web/src/test/kotlin/com/netflix/spinnaker/keel/integration/AuthPropagationTests.kt)). The only thing that stands out to me in the annotation is that you've configured your `fallbackMethod` to a method named `getFallback` but I don't see that method in your class...

About the configuration, you mentioned you configured retry properties in `keel.yml`. I assume you meant our internal config file. It would be good to add that to https://github.com/spinnaker/keel/blob/master/keel-web/config/keel.yml so that it can be reviewed and used by OSS deployments.

keel-web/config/keel.yml
```diff
@@ -70,6 +70,13 @@ okHttpClient:
   readTimeoutMs: 59000
   interceptor.skipHeaderCheck: true
 
+resilience4j.retry:
```
luispollo, line 73: nit: I think `retry` should go on the next line.

keel-web/config/keel.yml
```diff
@@ -70,6 +70,13 @@ okHttpClient:
...
+  configs:
+    default:
```
luispollo, line 75: Not a blocker, but you might want to define specific retry properties for each service and wire those into the respective classes. Can be done in a future PR.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
luispollo, line 21: I think it would be good to add a comment here explaining why we need retries. Something like:
```suggestion
  /**
    * Returns additional metadata about the specified build and commit, if available. This call is configured
    * to auto-retry as it's not on a code path where any external retries would happen.
    */
  @Retry(name = "getArtifactMetadata", fallbackMethod = "getFallback")
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -181,11 +185,62 @@ class SqlArtifactRepository(
...
         .set(ARTIFACT_VERSIONS.VERSION, version)
         .set(ARTIFACT_VERSIONS.RELEASE_STATUS, status?.toString())
+        .set(ARTIFACT_VERSIONS.BUILD_METADATA, objectMapper.writeValueAsString(artifactMetadata?.buildMetadata))
+        .set(ARTIFACT_VERSIONS.GIT_METADATA, objectMapper.writeValueAsString(artifactMetadata?.gitMetadata))
         .onDuplicateKeyIgnore()
         .execute()
     } == 1
   }
 
+  override fun updateArtifactMetadata(name: String, type: ArtifactType, version: String, status: ArtifactStatus?, artifactMetadata: ArtifactMetadata?) {
```
luispollo, line 195: Does it make sense to even call this function if `artifactMetadata` is `null`? It feels to me like that parameter should be non-nullable.

gal-yardeni: It does, because there can be a variety of reasons for the call not to work or some missing params, something is wrong with the artifact structure... I don't want it to fail because it's not nullable

luispollo: Sorry, I don't follow. The function is called `updateArtifactMetadata`, so supposedly you'd only call this when you have some metadata to set, and `artifactMetadata` is not `null`, right? Do you ever expect to call this function to set the metadata to `null`? That's what I was trying to say...

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplierTests.kt
gal-yardeni: done!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
gal-yardeni: it's not. I just follow the pattern but you're right, it not a `TelemetryEvent`. removed

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: ohhh i see what you are saying. will change it.

gal-yardeni: done!

keel-web/config/keel.yml
gal-yardeni: I saw examples for either option - but will change :)

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
gal-yardeni: added the fallback method, let me know if it makes sense.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
gal-yardeni: done!

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BaseArtifactSupplier.kt
gal-yardeni: I tried to implement it and faced an obstacle - my current logic is:
1. whenever a new version of an artifact is being saved, publish an `ArtifactSaved` event
2. in `ArtifactSaved` listener, call `artifactSupplier.getArtifactMetadata(event.artifact)`
3. store this information in the DB.
4. in `ApplicationService` call `repository.getArtifactBuildMetadata(..)` to fetch information, or call fallback.

Changing it to the logic you've proposed for the concrete implementations (debians, npms etc) will not have the option to call the DB to fetch this information since we don't want to include DB calls in those classes. Another option is to store the fallback build and git data in the DB, but I'm not sure what is the right way to do it.
Does it make sense to you? Am I missing something?

luispollo: > Changing it to the logic you've proposed for the concrete implementations (debians, npms etc) will not have the option to call the DB to fetch this information since we don't want to include DB calls in those classes.

Ah, right. I may have mixed those up, sorry. You can still use the idea of the abstract base class for common behavior, though for the event flow, and just keep `ApplicationService` as is.

luispollo (approved):
I still think that `DockerArtifactSupplier` and `DebianArtifactSupplier` should extend `BaseArtifactSupplier`, but we can refactor later.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
```diff
@@ -0,0 +1,99 @@
...
+          link = ""
+        ),
+        branch = scm?.first()?.branch,
+        project = properties?.get("projectKey") as String?,
+      )
+    )
+
+
+  // this method will be invoked whenever the retry will fail
+  protected fun fallback( buildNumber: String, commitId: String, e: Exception) {
```
luispollo, line 93: Just curious if you've tested this and how?

gal-yardeni:
> I still think that `DockerArtifactSupplier` and `DebianArtifactSupplier` should extend `BaseArtifactSupplier`, but we can refactor later.

Done!

PR #1431, opened 2020-08-13
chore(logs): Improve logging for artifact approval flow

luispollo:
Just adding some more logs to key spots in the artifact approval flow to help troubleshoot issues.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
```diff
@@ -77,37 +77,46 @@ class EnvironmentConstraintRunner(
     envContext: EnvironmentContext,
     pendingVersionsToCheck: MutableSet<String>
   ) {
-    var version: String? = null
```
emjburns, line 80: feels like it wasn't super necessary to change the variable names, only pointing that out because it makes it a bit harder to review the logging changes and confirm that it's logging only.

luispollo: I can't help but make opportunistic improvements when I have a chance... 🤷‍♂️

PR #1433, opened 2020-08-14
chore(build): use cross-compilation in the CI builds

plumpy:
(This is what all the other open-source projects have been doing since April. I made a half-hearted effort to get this working then, but obviously I gave up way too quickly.)


.mergify.yml
```diff
@@ -2,8 +2,8 @@ pull_request_rules:
   - name: Automatically merge on CI success and review
     conditions:
       - base=master
-      - status-success=build (8)
-      - status-success=build (11)
+      - status-success=build (true)
+      - status-success=build (false)
```
robfletcher, line 6: I'm not clear what this means

robfletcher: oh, is it because the job name will have `true`/`false` appended from the matrix?

robfletcher: I think I need to go read up about the cross compilation option

plumpy: Yeah, it comes from the matrix. Here's the plugin: https://github.com/nebula-plugins/gradle-java-cross-compile-plugin

Basically it means you can use the Java 11 `javac` but compile against the Java 8 JDK to make sure you don't use new APIs. If you just use `-target 8`, it'll compile to Java 8 bytecodes but successfully compile with methods like `Optional#stream`.

The problem with the jooq is that the `@Generated` annotation moved between Java 8 and Java 11 and using `DETECT_FROM_JDK` uses the Java 11 annotation, so compiling against the Java 8 APIs doesn't work.

robfletcher: So this is good to merge?

plumpy: Yep.

PR #1443, opened 2020-08-18
Kotlin 1.4

robfletcher:
Update to Kotlin 1.4


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt
```diff
@@ -133,6 +133,7 @@ class ExceptionHandler(
         is MissingKotlinParameterException -> ParsingError.MISSING_PROPERTY
         else -> when (rootCause) {
           is MissingKotlinParameterException -> ParsingError.MISSING_PROPERTY
+          is NullPointerException -> ParsingError.INVALID_VALUE
```
robfletcher, line 136: I guess some underlying library started throwing a different exception, because I had to add this to pass tests

PR #1445, opened 2020-08-19
feat(build): Replace jooqModelator with native jooq tooling + docker

luispollo:
This PR introduces the ability to build keel inside a Docker container by installing MySQL in the container, then using liquibase and the standard jOOQ code generation tool in place of the jooqModelator plugin (which itself uses Docker).

Cc: @nimakaviani

### DIFF
--- a/Dockerfile.compile
+++ b/Dockerfile.compile
@@ -23,6 +23,4 @@ RUN usermod -d /var/lib/mysql/ mysql && \
 
 CMD /etc/init.d/mysql restart && \
     ./gradlew --no-daemon -PenableCrossCompilerPlugin=true -PbuildingInDocker=true \
-        keel-sql:liquibaseUpdate \
-        keel-sql:jooqGenerate \
         keel-web:installDist -x test
--- a/Dockerfile.compile
+++ b/Dockerfile.compile
@@ -1,11 +1,28 @@
 FROM ubuntu:bionic
 LABEL maintainer="sig-platform@spinnaker.io"
+RUN echo "mysql-server mysql-server/root_password password sa" | debconf-set-selections
+RUN echo "mysql-server mysql-server/root_password_again password sa" | debconf-set-selections
 RUN apt-get update && apt-get install -y \
     openjdk-8-jdk \
     openjdk-11-jdk \
+    mysql-server \
  && rm -rf /var/lib/apt/lists/*
 ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk-amd64
 ENV JDK_18 /usr/lib/jvm/java-8-openjdk-amd64
 ENV GRADLE_USER_HOME /workspace/.gradle
 ENV GRADLE_OPTS -Xmx6g
-CMD ./gradlew --no-daemon -PenableCrossCompilerPlugin=true keel-web:installDist -x test
+
+RUN usermod -d /var/lib/mysql/ mysql && \
+    sed -i "s/bind-address\s*=\s*127.0.0.1/bind-address = 0.0.0.0/" /etc/mysql/mysql.conf.d/mysqld.cnf && \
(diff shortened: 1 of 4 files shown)


keel-sql/keel-sql.gradle.kts
```diff
@@ -1,16 +1,22 @@
 import ch.ayedo.jooqmodelator.gradle.JooqModelatorTask
 import com.diffplug.gradle.spotless.SpotlessExtension
 
+val buildingInDocker = project.properties["buildingInDocker"]?.toString().let { it == "true" }
+
 plugins {
   `java-library`
   id("kotlin-spring")
   id("ch.ayedo.jooqmodelator") version "3.9.0"
+  id("nu.studer.jooq") version "5.0.1"
```
luispollo, line 10: We didn't strictly need this plugin, but I used it to generate the list of runtime dependencies for the `jooqGenerate` custom task below because I suck at gradle. I tried creating a vanilla configuration instead but got stuck. Would be happy to remove the plugin if anyone can show me the way.

nimakaviani:
awesome, thanks for doing this!

keel-sql/keel-sql.gradle.kts
```diff
@@ -1,16 +1,22 @@
...
 }
 
 afterEvaluate {
-  tasks.getByName("compileKotlin") {
-    dependsOn("generateJooqMetamodel")
+  // When not running in Docker, we can use jooqModelator (which itself uses Docker) to generate the jOOQ meta-model
+  if (!buildingInDocker) {
+    tasks.getByName("compileKotlin") {
+      dependsOn("generateJooqMetamodel")
+    }
```
robfletcher, line 19: Why don't we just switch over completely?

luispollo: We can... I was just not looking to spend too much more time on this at this point as jooqModelator works fine for our use. We'd need to add another task to spin up the docker container, etc.

luispollo: Let me give it a try -- if I can use the vanilla mysql container, then it should be relatively straightforward.

keel-sql/keel-sql.gradle.kts
```diff
@@ -1,16 +1,21 @@
...
 
 afterEvaluate {
-  tasks.getByName("compileKotlin") {
-    dependsOn("generateJooqMetamodel")
+  // When not running in Docker, we can use jooqModelator (which itself uses Docker) to generate the jOOQ meta-model
+  if (!buildingInDocker) {
+    tasks.getByName("compileKotlin") {
+      dependsOn("jooqGenerate")
+    }
   }
```
luispollo, line 19:
```suggestion
  tasks.getByName("compileKotlin") {
    dependsOn("jooqGenerate")
  }
```

Dockerfile.compile
```diff
@@ -1,11 +1,28 @@
...
+    /etc/init.d/mysql start && \
+    mysql -u root --password=sa -e "GRANT ALL PRIVILEGES ON *.* TO 'root'@'%' IDENTIFIED BY 'sa' WITH GRANT OPTION;" && \
+    mysql -u root --password=sa -e "GRANT PROXY ON ''@'' TO 'root'@'%' WITH GRANT OPTION;" && \
+    /etc/init.d/mysql restart && \
+    mysql -u root --password=sa -e 'CREATE DATABASE `keel` DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;'
+
+CMD /etc/init.d/mysql restart && \
+    ./gradlew --no-daemon -PenableCrossCompilerPlugin=true -PbuildingInDocker=true \
+        keel-sql:liquibaseUpdate \
+        keel-sql:jooqGenerate \
```
luispollo, line 27:
```suggestion
```

PR #1446, opened 2020-08-20
feat(plugins): Enablement for Java plugins

luispollo:
This PR addresses the following:
- Make `Resource` metadata accessor fields accessible from Java-based plugins
- Add a version of `TaskLauncher.submitJob()` that is asynchronous but not `suspend` so it can be called from Java

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
@@ -22,9 +22,6 @@ data class Resource<out T : ResourceSpec>(
   val application: String
     get() = metadata.getValue("application").toString()
 
-  val environment: String
-    get() = metadata.getValue("environmentName").toString()
-
   /**
    * Attempts to find an artifact in the delivery config based on information in this resource's spec.
    */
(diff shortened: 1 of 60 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Resource.kt
```diff
@@ -13,6 +13,18 @@ data class Resource<out T : ResourceSpec>(
...
+  val id: String
+    get() = metadata.getValue("id").toString()
+
+  val serviceAccount: String
+    get() = metadata.getValue("serviceAccount").toString()
+
+  val application: String
+    get() = metadata.getValue("application").toString()
+
+  val environment: String
```
luispollo, line 25: This is new. I added it as it's needed to find things like notification config in a `ResourceHandler`.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskLauncher.kt
```diff
@@ -63,6 +63,15 @@ class OrcaTaskLauncher(
       type = SubjectType.RESOURCE
     )
 
+  override fun submitJobAsync(
```
luispollo, line 66: @robfletcher This was the best I could come up with for a Java-friendly version of the method. It relies on `kotlin-coroutines-jdk8`, which I'm not sure how that fits with our plans to move to Java 11. Also, I found indications that the use of `GlobalScope` is discouraged, but wasn't sure what the alternative could be in this case.

Issue #1401, opened 2020-08-03
Keel gets out of sync with reality when instances fail to launch during deployment


erikmunson:
tl;dr In scenarios where a deployment happens but instances fail to come up, Keel marks versions as deployed when they are not from a traffic/health perspective. Real life scenario from a recent event:

1. Keel tells Orca to deploy a new server group, but instances aren't coming up due to a capacity shortage
2. Orca sits and waits for instances to come up until it times out, failing the task
3. There are now two active server groups: the old one with healthy instances serving traffic, and the new one which has zero instances of any kind. From a user's perspective this deployment has failed and wasn't cleaned up
4. Keel comes back around, sees there's an "active" server group in the cluster with the right version and the right capacity settings, and marks the version it was trying to deploy as deployed successfully. It does not take into account that there is an old server group serving traffic, or the fact that no instances are up in the new server group

Things were left in a state where the new version was not deployed and the task itself failed, but Keel said everything was great and the new version replaced the old version. Seems very related to the [two active server groups GH issue](https://github.com/spinnaker/keel/issues/1084) we're also aware of.

In the best case, this is unexpected and folks are confused/annoyed that they're seeing incorrect data. In the worst case this could theoretically result in outages/incidents due to either old code running without people realizing _or_ new code coming online as instances are launched later (in this example when the capacity shortage is over), effectively causing multiple versions to run simultaneously without anyone being alerted that things are changing/deploying.

emjburns:
This is done. We now only publish a "deployed" event when the server group is actually healthy - that means that it has all the correct capacity and that instances are "up" if you want them to be up.

PR #1448, opened 2020-08-20
feat(titus): only allow one active server group

emjburns:
Fist stab at addressing https://github.com/spinnaker/keel/issues/1084 for titus only (titus is less complicated). Feedback desired.

This PR adds a new property to the server group response (`onlyEnabledServerGroup`) which defaults to true. When the desired state is resolved for a server group, it is also true. When we fetch the current state of clusters we calculate whether that property is true or false for each "active server group" (aka the one we're looking at to see if it's in the right state or not).

If `onlyEnabledServerGroup == false`, and that's the only diff, then we disable the oldest active server group (heuristic alert! feedback desired!). The heuristic is: disable the oldest active server group with the _wrong version_ until there are none left. Then, do the same with clusters with the right version until there is only one active server group left.

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
@@ -26,6 +26,7 @@ interface BaseServerGroup {
   val disabled: Boolean
     get() = false
   val instanceCounts: InstanceCounts
+  val createdTime: Long
 }
 
 data class InstanceCounts(
@@ -74,7 +75,8 @@ data class ServerGroup(
   override val moniker: Moniker,
   override val buildInfo: BuildInfo? = null,
   override val disabled: Boolean,
-  override val instanceCounts: InstanceCounts
+  override val instanceCounts: InstanceCounts,
+  override val createdTime: Long
 ) : BaseEc2ServerGroup
 
 fun ServerGroup.toActive(accountName: String) =
@@ -95,7 +97,8 @@ fun ServerGroup.toActive(accountName: String) =
     accountName = accountName,
     moniker = moniker,
     buildInfo = buildInfo,
-    instanceCounts = instanceCounts
+    instanceCounts = instanceCounts,
+    createdTime = createdTime
   )
 
 // todo eb: this should be more general so that it works for all server groups, not just ec2
@@ -116,7 +119,8 @@ data class ActiveServerGroup(
   val accountName: String,
   override val moniker: Moniker,
   override val buildInfo: BuildInfo? = null,
-  override val instanceCounts: InstanceCounts
+  override val instanceCounts: InstanceCounts,
+  override val createdTime: Long
 ) : BaseEc2ServerGroup
 
 fun ActiveServerGroup.subnet(cloudDriverCache: CloudDriverCache): String =
--- a/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
+++ b/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
@@ -75,7 +75,6 @@ import java.time.Clock
 import java.time.Instant
 import java.time.ZoneId
 import java.time.format.DateTimeFormatter
-import java.util.concurrent.atomic.AtomicInteger
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.runBlocking
@@ -460,12 +459,12 @@ class TitusClusterHandler(
     current != null && affectedRootPropertyTypes.all { it == Capacity::class.java }
 
   /**
-   * @return true if the only difference is in the onlyActiveServerGroup property
+   * @return true if the only difference is in the onlyEnabledServerGroup property
    */
   private fun ResourceDiff<TitusServerGroup>.isEnabledOnly(): Boolean =
     current != null &&
-      affectedRootPropertyNames.all { it == "onlyActiveServerGroup" } &&
-      current!!.onlyActiveServerGroup != desired.onlyActiveServerGroup
+      affectedRootPropertyNames.all { it == "onlyEnabledServerGroup" } &&
+      current!!.onlyEnabledServerGroup != desired.onlyEnabledServerGroup
 
   private fun TitusClusterSpec.generateOverrides(serverGroups: Map<String, TitusServerGroup>) =
     serverGroups.forEach { (region, serverGroup) ->
@@ -490,8 +489,8 @@ class TitusClusterHandler(
         .size
 
       when (numEnabled) {
-        1 -> activeServerGroup.copy(onlyActiveServerGroup = true)
-        else -> activeServerGroup.copy(onlyActiveServerGroup = false)
+        1 -> activeServerGroup.copy(onlyEnabledServerGroup = true)
+        else -> activeServerGroup.copy(onlyEnabledServerGroup = false)
       }
     }
 
--- a/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
+++ b/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
@@ -40,6 +40,7 @@ import com.netflix.spinnaker.keel.api.plugins.ResolvableResourceHandler
 import com.netflix.spinnaker.keel.api.plugins.Resolver
 import com.netflix.spinnaker.keel.api.titus.CLOUD_PROVIDER
 import com.netflix.spinnaker.keel.api.titus.TITUS_CLUSTER_V1
+import com.netflix.spinnaker.keel.api.titus.exceptions.ActiveServerGroupsException
 import com.netflix.spinnaker.keel.api.titus.exceptions.RegistryNotFoundException
 import com.netflix.spinnaker.keel.api.titus.exceptions.TitusAccountConfigurationException
 import com.netflix.spinnaker.keel.api.withDefaultsOmitted
@@ -51,7 +52,9 @@ import com.netflix.spinnaker.keel.clouddriver.model.Constraints
 import com.netflix.spinnaker.keel.clouddriver.model.DockerImage
 import com.netflix.spinnaker.keel.clouddriver.model.MigrationPolicy
 import com.netflix.spinnaker.keel.clouddriver.model.Resources
+import com.netflix.spinnaker.keel.clouddriver.model.ServerGroup
 import com.netflix.spinnaker.keel.clouddriver.model.TitusActiveServerGroup
+import com.netflix.spinnaker.keel.clouddriver.model.TitusServerGroup as ClouddriverTitusServerGroup
 import com.netflix.spinnaker.keel.core.api.DEFAULT_SERVICE_ACCOUNT
 import com.netflix.spinnaker.keel.core.orcaClusterMoniker
 import com.netflix.spinnaker.keel.core.serverGroup
@@ -72,6 +75,7 @@ import java.time.Clock
 import java.time.Instant
 import java.time.ZoneId
 import java.time.format.DateTimeFormatter
+import java.util.concurrent.atomic.AtomicInteger
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.runBlocking
@@ -101,7 +105,7 @@ class TitusClusterHandler(
 
   override suspend fun current(resource: Resource<TitusClusterSpec>): Map<String, TitusServerGroup> =
     cloudDriverService
-      .getServerGroups(resource)
+      .getActiveServerGroups(resource)
       .byRegion()
 
   override suspend fun actuationInProgress(resource: Resource<TitusClusterSpec>): Boolean =
@@ -147,11 +151,13 @@ class TitusClusterHandler(
 
           val job = when {
             diff.isCapacityOnly() -> diff.resizeServerGroupJob()
+            diff.isEnabledOnly() -> diff.disableOtherServerGroupJob(resource, version.toString())
             else -> diff.upsertServerGroupJob(tagToUse) + resource.spec.deployWith.toOrcaJobProperties("Amazon")
           }
 
-          val description = when (version) {
-            null -> "Resize server group ${desired.moniker} in ${desired.location.account}/${desired.location.region}"
+          val description = when {
+            diff.isCapacityOnly() -> "Resize server group ${desired.moniker} in ${desired.location.account}/${desired.location.region}"
+            diff.isEnabledOnly() -> "Disable extra active server group ${job["asgName"]} in ${desired.location.account}/${desired.location.region}"
             else -> "Deploy $version to server group ${desired.moniker} in ${desired.location.account}/${desired.location.region}"
           }
           log.info("Upserting server group using task: {}", job)
@@ -165,13 +171,14 @@ class TitusClusterHandler(
             )
           }
 
-          tags.forEach { tag ->
-            publisher.publishEvent(
-              ArtifactVersionDeploying(
+          if (!diff.isEnabledOnly() && !diff.isCapacityOnly()) {
+            // only publish deploying if we're deploying a new version
+            tags.forEach { tag ->
+              publisher.publishEvent(ArtifactVersionDeploying(
                 resourceId = resource.id,
                 artifactVersion = tag
-              )
-            )
+              ))
+            }
           }
           return@map result
         }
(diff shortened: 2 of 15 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -26,6 +26,7 @@ interface BaseServerGroup {
   val disabled: Boolean
     get() = false
   val instanceCounts: InstanceCounts
+  val createdTime: Long
```
emjburns, line 29: adding created time so we can determine the age of server groups

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -313,6 +320,52 @@ class TitusClusterHandler(
       false
     }
```
emjburns, line 322: This is the heuristic part, where we handle if the only diff is that there are too many enabled server groups

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -413,35 +474,73 @@ class TitusClusterHandler(
...
+    val activeServerGroups = getActiveServerGroups(
       resource.spec.locations.account,
       resource.spec.moniker,
       resource.spec.locations.regions.map { it.name }.toSet(),
       resource.serviceAccount
-    )
-      .also { them ->
-        val sameContainer: Boolean = them.distinctBy { it.container.digest }.size == 1
-        val healthy: Boolean = them.all {
-          it.instanceCounts?.isHealthy(resource.spec.deployWith.noHealth) == true
```
emjburns, line 426: here is where we calculate and set the property `onlyActiveServerGroup`. I had to refactor this to not be a chain of kotlin functions because I couldn't get it to work otherwise.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -413,35 +474,73 @@ class TitusClusterHandler(
...
-                )
-              )
-            }
+      when (numEnabled.get()) {
+        1 -> activeServerGroup.copy(onlyActiveServerGroup = true)
+        else -> activeServerGroup.copy(onlyActiveServerGroup = false)
+      }
+    }
+
+    // publish health events
```
emjburns, line 501: no change in logic here, just moved.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -313,6 +320,52 @@ class TitusClusterHandler(
...
+
+    return mapOf(
+      "type" to "disableServerGroup",
+      "cloudProvider" to CLOUD_PROVIDER,
+      "credentials" to desired.location.account,
+      "moniker" to sgToDisable.moniker.orcaClusterMoniker,
+      "region" to sgToDisable.region,
+      "serverGroupName" to sgToDisable.name,
+      "asgName" to sgToDisable.name
+    )
```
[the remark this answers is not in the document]
robfletcher: Thinking about it I feel like > 1 excess server group is probably quite an edge case. Let's go with this as a first iteration.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -313,6 +320,52 @@ class TitusClusterHandler(
...
+    val current = requireNotNull(current) {
+      "Current server group must not be null when generating a disable job"
+    }
+    // todo eb: figure out what server group to disable
+    val existingServerGroups: Map<String, List<ClouddriverTitusServerGroup>> = runBlocking { getExistingServerGroupsByRegion(resource) }
+    val sgInRegion = existingServerGroups.getOrDefault(current.location.region, emptyList()).filterNot { it.disabled }
+
+    if (sgInRegion.size < 2) {
+      log.error("Diff says this is not the only active server group, but now we say otherwise. " +
+        "What is going on? Existing server groups: {}", existingServerGroups)
```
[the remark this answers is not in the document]
robfletcher: So this would maybe be a transient condition due to something else acting on the server group we initially detected. I think the right thing to do in that case is pretty much always to just stop and check again next cycle.

erikmunson:
This feels like a sensible algorithm/behavior overall, I'll share a concern and a thought:

1. Based on my reading, it appears this behavior will worsen the outcome of the bug described in https://github.com/spinnaker/keel/issues/1401. Today, that scenario leads to keel thinking the right version is running when the only active and healthy server group is really the _previous version_. Bad, but at least traffic is still being served. With this new logic, my understanding is that keel would notice there are two server groups and disable the _only server group with healthy instances_ (the one running the 'wrong' version), effectively causing an outage. Is that assessment accurate?

2. While I can't currently think of any other scenarios where this behavior would become problematic, it _might_ be useful to bounce the idea around with folks that are deeply familiar with Orca's previous troubles around getting clusters into unexpected states. Mark, Daniel, and Cameron come to mind. No idea whether it would actually be useful in this case but I know in the past we've realized there are potential gaps in our approach after running through imaginary scenarios with those folks (and others).

Thanks for aiming for simplicity with this approach, and finding a behavior that doesn't require direct input or involvement from app owners!

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -74,7 +75,8 @@ data class ServerGroup(
   override val moniker: Moniker,
   override val buildInfo: BuildInfo? = null,
   override val disabled: Boolean,
-  override val instanceCounts: InstanceCounts
+  override val instanceCounts: InstanceCounts,
+  override val createdTime: Long
```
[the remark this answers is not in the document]
emjburns: FWIW Instant does not "just work", and since this is just in the clouddriver response I'm inclined to leave it as a long for ease of use. It doesn't feel worth it to transform it to another data type since that wouldn't give us anything.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -147,11 +151,13 @@ class TitusClusterHandler(
...
             diff.isCapacityOnly() -> diff.resizeServerGroupJob()
+            diff.isEnabledOnly() -> diff.disableOtherServerGroupJob(resource, version.toString())
             else -> diff.upsertServerGroupJob(tagToUse) + resource.spec.deployWith.toOrcaJobProperties("Amazon")
           }
 
-          val description = when (version) {
-            null -> "Resize server group ${desired.moniker} in ${desired.location.account}/${desired.location.region}"
+          val description = when {
+            diff.isCapacityOnly() -> "Resize server group ${desired.moniker} in ${desired.location.account}/${desired.location.region}"
+            diff.isEnabledOnly() -> "Disable extra active server group ${job["asgName"]} in ${desired.location.account}/${desired.location.region}"
```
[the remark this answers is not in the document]
emjburns: Yep, it will. These other things are more of an optimization when there's something non-version related going on. On the AWS side there is a check for if it's capacity and scaling policies together. I don't know if it makes sense to do that optimization or not.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -313,6 +320,52 @@ class TitusClusterHandler(
...
+    val sgInRegion = existingServerGroups.getOrDefault(current.location.region, emptyList()).filterNot { it.disabled }
+
+    if (sgInRegion.size < 2) {
+      log.error("Diff says this is not the only active server group, but now we say otherwise. " +
+        "What is going on? Existing server groups: {}", existingServerGroups)
+      throw ActiveServerGroupsException(resource.id, "No other active server group found to disable.")
+    }
+
+    val wrongImageSGs = sgInRegion.filterNot { it.image.dockerImageVersion == desiredVersion }.sortedBy { it.createdTime }
+    val rightImageSGs = sgInRegion.filter { it.image.dockerImageVersion == desiredVersion }.sortedBy { it.createdTime }
```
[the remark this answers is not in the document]
emjburns: woah TIL. Thanks

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -413,35 +474,73 @@ class TitusClusterHandler(
...
-        val healthy: Boolean = them.all {
-          it.instanceCounts?.isHealthy(resource.spec.deployWith.noHealth) == true
+    ).map { activeServerGroup ->
+      // get all server groups in the cluster, split by region,
+      val existingServerGroups: Map<String, List<ClouddriverTitusServerGroup>> = getExistingServerGroupsByRegion(resource)
+      val numEnabled = AtomicInteger()
+      existingServerGroups
+        .getOrDefault(activeServerGroup.location.region, emptyList<ServerGroup>())
+        .forEach { serverGroup ->
+          log.info(">> serverGroup is disabled: {}", serverGroup.disabled)
```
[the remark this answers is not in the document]
emjburns: oops debugging statement left in, i'll remove! I always start my local debugging logs with ">>" so it's easy to search for them :D  good catch.

emjburns:
@erikmunson Yes this will make that bug worse. I'll be tackling that one next. Feels like an ok tradeoff since they're both not super common.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -313,6 +319,52 @@ class TitusClusterHandler(
...
+    }
+
+    val (rightImageASGs, wrongImageASGs) = sgInRegion
+      .sortedBy { it.createdTime }
+      .partition { it.image.dockerImageVersion == desiredVersion }
+
+    val sgToDisable = when {
+      wrongImageASGs.isNotEmpty() -> {
+        log.debug("Disabling oldest server group with incorrect docker image version for {}", resource.id)
+        wrongImageASGs.first()
```
luispollo, line 343: Any particular reason why we would disable just the oldest and not every server group with the wrong image?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -313,6 +319,52 @@ class TitusClusterHandler(
...
+      }
+      rightImageASGs.size > 1 -> {
+        log.debug("Disabling oldest server group with correct docker image version " +
```
luispollo, line 346: Same question as above.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
luispollo: Ah, I see @robfletcher asked the same question below and you already answered. Fine to address in an upcoming PR.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -398,12 +450,23 @@ class TitusClusterHandler(
...
   /**
    * @return `true` if the only changes in the diff are to capacity.
    */
   private fun ResourceDiff<TitusServerGroup>.isCapacityOnly(): Boolean =
     current != null && affectedRootPropertyTypes.all { it == Capacity::class.java }
 
+  /**
+   * @return true if the only difference is in the onlyActiveServerGroup property
+   */
+  private fun ResourceDiff<TitusServerGroup>.isEnabledOnly(): Boolean =
```
luispollo, line 465: nit: consider naming it `isActiveOnly()` for consistency with the naming elsewhere.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
emjburns: It feels like an unneeded optimization at this point. I likely won't do it in a future PR unless we get feedback about it

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -413,35 +476,71 @@ class TitusClusterHandler(
...
-              // and we don't care about some of them.
-              publisher.publishEvent(
-                ArtifactVersionDeployed(
-                  resourceId = resource.id,
-                  artifactVersion = tag
-                )
-              )
-            }
+      when (numEnabled) {
+        1 -> activeServerGroup.copy(onlyActiveServerGroup = true)
```
luispollo, line 497: Shouldn't this be `> 1`?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
emjburns: Active has a weird meaning that I don't really want to conflate here.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
emjburns: No, if there are more than on enabled server groups that it's not the only enabled one

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TestUtils.kt
```diff
@@ -40,6 +41,105 @@ fun TitusServerGroup.toClouddriverResponse(
       tags = emptyMap(),
       resources = resources,
       capacityGroup = moniker.app,
-      instanceCounts = instanceCounts.run { com.netflix.spinnaker.keel.clouddriver.model.InstanceCounts(total, up, down, unknown, outOfService, starting) }
+      instanceCounts = instanceCounts.run { com.netflix.spinnaker.keel.clouddriver.model.InstanceCounts(total, up, down, unknown, outOfService, starting) },
```
luispollo, line 44: nit: fix import

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
luispollo: OK. In that case I don't know what "enabled" means in this context, haha. 🤷‍♂️

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
luispollo: Oof, that's confusing, sorry... Why do we loop and count if all we care about is if there's only one, then? Wouldn't that just be a `.filter` and a `.size`?

emjburns: ha ya that would be better! let me make that change

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
emjburns: Ok, let me explain the underlying weirdness:
- clouddriver has an endpoint for getting the "active" server group, which for us is basically the latest enabled server group.
- that meaning of "active" is used throughout here
- this property I guess could be called "onlyEnabledServerGroup" - maybe that would be clearer? probably would be a lot clearer

If I rename to `onlyEnabledServerGroup` for the property, does that make more sense?

luispollo: It does, but don't bother, it's alright.

PR #1449, opened 2020-08-21
fix(filter): make servlet filter pattern match request path

nimakaviani:
`/delivery-configs/` does not match requests to `/delivery-configs`.
We either want `/delivery-configs` or `/delivery-configs/*`. Since the
former is more restrictive, I settled on the former.


nimakaviani:
I am also tempted to say that any request with `Content-Type: application/x-yaml` could benefit from this double passing and proper parsing of anchors and aliases. so maybe we can relax filtering by url pattern?

PR #1454, opened 2020-08-24
feat(plugins): Provide injection of SLF4J loggers for plugin beans

luispollo:
Over the last few days, I've tried a handful of different approaches in an attempt to inject an SLF4J logger instantiated by keel into a plugin, all of which failed. Namely:
- Defining a `@Bean` of type `org.slf4j.ILoggerFactory` and making that a constructor argument in the plugin class. The plugin bean is instantiated before the logger factory, despite the dependency (I don't understand why, since the other two constructor args which also specify dependencies cause the corresponding beans to be created in keel early).
- Defining a `@Bean` with `@Scope("prototype")` of type `org.slf4j.Logger` and making that a constructor argument in the plugin class. Fails with the same behavior as above.
- Implementing a `BeanDefinitionRegistryPostProcessor` that registers a logger or logger factory _before_ the plugin even loads. Plugin class still fails to instantiate with failed dependency.

After doing some research, I came across this approach which explicitly injects a logger instance into beans based on an annotation, which bypasses Spring auto-wiring and gives us much more control. I'd like to cut a new release of keel with this support such that I can test that with the plugin. 🤞 

TODO:
- [x] Tests

P.S. If this works, we could entertain the idea of supporting this for all plugins via kork.


robfletcher:
What's the context here? You don't want to depend on SLF4J in the plugin? But you will have to in order to define the plugin parameter to the constructor no?

luispollo:
> What's the context here? You don't want to depend on SLF4J in the plugin? But you will have to in order to define the plugin parameter to the constructor no?

It's OK to depend on `slf4j-api` in the plugin. The problem is that the plugin is very likely to pull in a transitive dependency on an actual binding to SLF4J (e.g. logback), and just as likely to pull a version that conflicts with the version that's in keel. That's what I ran into, and after going 3 levels deep trying to pin down transitive dependencies on the plugin side, I decided to try an alternative.

My hope with this is that we'll avoid that by having keel, not the plugin, instantiate the logger.

robfletcher:
Couldn't we still end up with a version conflict on `slf4j-api`?

luispollo:
> Couldn't we still end up with a version conflict on `slf4j-api`?

Yes, but my hope is that it stops there, and not 3 or 4 levels deep into the dependency graph from whatever implementation gets pulled in, like I ran into. Also, I _think_ `slf4j-api` should get its versioned aligned with `kork-bom` in the plugin. At least that's been my observation.

luispollo:
I've tested this locally and the injector fails to set the logger field with an `IllegalArgumentException` on an `isAssignableFrom()` check deep in Spring:
```
Caused by: java.lang.IllegalArgumentException: Can not set org.slf4j.Logger field com.netflix.spinnaker.plugins.treasure.TreasureStaticWebAppHandler.LOG to ch.qos.logback.classic.Logger
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:167)
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:171)
	at java.base/jdk.internal.reflect.UnsafeObjectFieldAccessorImpl.set(UnsafeObjectFieldAccessorImpl.java:81)
	at java.base/java.lang.reflect.Field.set(Field.java:780)
	at com.netflix.spinnaker.config.PluginLoggerInjector$postProcessBeforeInitialization$1.doWith(PluginLoggerInjector.kt:24)
	at org.springframework.util.ReflectionUtils.doWithFields(ReflectionUtils.java:705)
	at org.springframework.util.ReflectionUtils.doWithFields(ReflectionUtils.java:684)
	at com.netflix.spinnaker.config.PluginLoggerInjector.postProcessBeforeInitialization(PluginLoggerInjector.kt:19)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.applyBeanPostProcessorsBeforeInitialization(AbstractAutowireCapableBeanFactory.java:416)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1788)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:595)
	... 31 common frames omitted
```
In chatting with @robzienert, we guessed it's a class loader issue, but not sure. I'm throwing in the towel on plugin logging for now, will see how things behave once we switch to the V2 framework.

PR #1456, opened 2020-08-24
feat(resources): Add standardized display name for resources

luispollo:
Adds a `displayName` property to `ResourceSpec` and `ResourceSummary` such that we have a consistent way of figuring out what to display in the UI for all resource types.


luispollo:
Cc:@erikmunson

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -148,11 +148,6 @@ class ApplicationService(
     ResourceSummary(
       resource = this,
       status = resourceStatusService.getStatus(id),
-      moniker = if (spec is Monikered) {
```
luispollo, line 151: Moved to `Monikered`.

erikmunson:
lovely ❤️

PR #1459, opened 2020-08-25
fix(core): Don't make assumptions about which resources link to artifacts

luispollo:
Instead of checking for `ComputeResourceSpec` to figure out which resources have links to artifacts, just check the artifact-provider interfaces it wraps, allowing for other types of resources to provide that information. This is required to enable our current work on internal resource plugins.

Also, annotates a few methods on those interfaces with `@JvmDefault` so they can be used in Java plugins.

Cc: @erikmunson

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactInterfaces.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactInterfaces.kt
@@ -13,6 +13,7 @@ interface ArtifactProvider {
   val artifactName: String?
   val artifactType: ArtifactType?
 
+  @JvmDefault
   fun completeArtifactOrNull() =
     if (artifactName != null && artifactType != null) {
       CompleteArtifact(artifactName!!, artifactType!!)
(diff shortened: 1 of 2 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactInterfaces.kt
```diff
@@ -13,6 +13,7 @@ interface ArtifactProvider {
   val artifactName: String?
   val artifactType: ArtifactType?
 
+  @JvmDefault
```
luispollo, line 16: For Java plugins.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactInterfaces.kt
```diff
@@ -28,6 +29,7 @@ interface ArtifactProvider {
 interface VersionedArtifactProvider : ArtifactProvider {
   val artifactVersion: String?
 
+  @JvmDefault
```
luispollo, line 32: For Java plugins.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ArtifactInterfaces.kt
```diff
@@ -44,6 +46,7 @@ interface ArtifactReferenceProvider {
   val artifactReference: String?
   val artifactType: ArtifactType?
 
+  @JvmDefault
```
luispollo, line 49: For Java plugins.

PR #1461, opened 2020-08-27
fix(titus): Handle 404s when getting server groups

luispollo:
Follow-up to #1448. Determining current state was failing when there's no existing server group (when clouddriver returns 404).


keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -521,19 +521,28 @@ class TitusClusterHandler(
...
+          account = resource.spec.locations.account,
+          cluster = resource.spec.moniker.toString()
+        )
+        .serverGroups
+        .forEach { sg ->
+          val existing = existingServerGroups.getOrPut(sg.region, { mutableListOf() })
+          existing.add(sg)
+          existingServerGroups[sg.region] = existing
+        }
+    } catch (e: HttpException) {
```
gal-yardeni, line 538: should we log the error here?

emjburns: We don't need to log since if it's a 404 we expect that (well, we NOW expect it), and otherwise we throw the exception

PR #1463, opened 2020-08-27
feat(ec2): don't allow multiple enabled ASGs

emjburns:
Follow up to https://github.com/spinnaker/keel/pull/1448: same functionality, but for EC2!

Also, fixing the titus tests to correctly throw a not found when there are no server groups, instead of returning the wrapper object with an empty list like I incorrectly expected it to.

Closes https://github.com/spinnaker/keel/issues/1084

### DIFF
--- a/keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ServerGroup.kt
+++ b/keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ServerGroup.kt
@@ -45,6 +45,7 @@ data class ServerGroup(
   val health: Health = Health(),
   val scaling: Scaling = Scaling(),
   val tags: Map<String, String> = emptyMap(),
+  val onlyEnabledServerGroup: Boolean = true,
   @get:ExcludedFromDiff
   val image: ActiveServerGroupImage? = null,
   @get:ExcludedFromDiff
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -54,6 +54,7 @@ import com.netflix.spinnaker.keel.clouddriver.model.CustomizedMetricSpecificatio
 import com.netflix.spinnaker.keel.clouddriver.model.MetricDimensionModel
 import com.netflix.spinnaker.keel.clouddriver.model.PredefinedMetricSpecificationModel
 import com.netflix.spinnaker.keel.clouddriver.model.ScalingPolicy
+import com.netflix.spinnaker.keel.clouddriver.model.ServerGroup as ClouddriverServerGroup
 import com.netflix.spinnaker.keel.clouddriver.model.StepAdjustmentModel
 import com.netflix.spinnaker.keel.clouddriver.model.Tag
 import com.netflix.spinnaker.keel.clouddriver.model.subnet
@@ -65,6 +66,7 @@ import com.netflix.spinnaker.keel.ec2.MissingAppVersionException
 import com.netflix.spinnaker.keel.ec2.toEc2Api
 import com.netflix.spinnaker.keel.events.ArtifactVersionDeployed
 import com.netflix.spinnaker.keel.events.ArtifactVersionDeploying
+import com.netflix.spinnaker.keel.exceptions.ActiveServerGroupsException
 import com.netflix.spinnaker.keel.exceptions.ExportError
 import com.netflix.spinnaker.keel.orca.ClusterExportHelper
 import com.netflix.spinnaker.keel.orca.OrcaService
@@ -83,6 +85,7 @@ import java.time.format.DateTimeFormatter
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.runBlocking
 import org.springframework.context.ApplicationEventPublisher
 import retrofit2.HttpException
 
@@ -110,7 +113,7 @@ class ClusterHandler(
 
   override suspend fun current(resource: Resource<ClusterSpec>): Map<String, ServerGroup> =
     cloudDriverService
-      .getServerGroups(resource)
+      .getActiveServerGroups(resource)
       .byRegion()
 
   override suspend fun upsert(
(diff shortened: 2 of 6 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -546,6 +565,51 @@ class ClusterHandler(
           current!!.scaling.stepScalingPolicies != desired.scaling.stepScalingPolicies
         )
 
+  private fun ResourceDiff<ServerGroup>.disableOtherServerGroupJob(resource: Resource<ClusterSpec>, desiredVersion: String): Map<String, Any?> {
```
emjburns, line 568: This is the exact same logic as titus. Like, the exact same. This is what prompted me to think about how much of the cluster stuff can be common between cloud providers.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -796,33 +860,72 @@ class ClusterHandler(
...
+          account = resource.spec.locations.account,
+          cluster = resource.spec.moniker.toString()
+        )
+        .serverGroups
+        .forEach { sg ->
+          val existing = existingServerGroups.getOrPut(sg.region, { mutableListOf() })
+          existing.add(sg)
+          existingServerGroups[sg.region] = existing
         }
+    } catch (e: HttpException) {
```
emjburns, line 920: @luispollo added error handling for ec2 :) thanks again for the fix for titus!

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterHandler.kt
```diff
@@ -536,13 +536,10 @@ class TitusClusterHandler(
           existingServerGroups[sg.region] = existing
         }
     } catch (e: HttpException) {
```
emjburns, line 538: slightly simplifying this logic

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusClusterHandlerTests.kt
```diff
@@ -229,7 +229,7 @@ class TitusClusterHandlerTests : JUnit5Minutests {
         coEvery { cloudDriverService.titusActiveServerGroup(any(), "us-west-2") } throws RETROFIT_NOT_FOUND
         coEvery { cloudDriverService.findDockerImages("testregistry", "spinnaker/keel", any(), any()) } returns
           listOf(DockerImage("testregistry", "spinnaker/keel", "master-h2.blah", "sha:1111"))
-        coEvery { cloudDriverService.listTitusServerGroups(any(), any(), any(), any()) } returns ServerGroupCollection(titusAccount, emptySet())
+        coEvery { cloudDriverService.listTitusServerGroups(any(), any(), any(), any()) } throws RETROFIT_NOT_FOUND
```
emjburns, line 232: thx clouddriver you are awesome

keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ServerGroup.kt
```diff
@@ -45,6 +45,7 @@ data class ServerGroup(
   val health: Health = Health(),
   val scaling: Scaling = Scaling(),
   val tags: Map<String, String> = emptyMap(),
+  val onlyEnabledServerGroup: Boolean = true,
```
luispollo, line 48: Suggestion to make it even clearer:
```suggestion
  val isOnlyEnabledServerGroup: Boolean = true,
```

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
luispollo: You could always extract this into a common function in a shared module like `keel-core` (or even a new shared module between EC2 and Titus)... Also, FWIW, both those specs inherit from `ComputeResourceSpec`, so you could even think about extracting common relevant fields to that interface if it helps.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -796,33 +860,72 @@ class ClusterHandler(
...
-          if (appVersion != null) {
-            publisher.publishEvent(
-              ArtifactVersionDeployed(
-                resourceId = resource.id,
-                artifactVersion = appVersion
-              )
-            )
-          }
+    ).map { activeServerGroup ->
+      // get all server groups in the cluster, split by region,
```
luispollo, line 871: This comment doesn't seem to match the code below, since the result is just a flat list. Maybe "get all active server groups in the cluster and tag the ones that are the only enabled ones in their region"? Also, maybe move the comment to line 865?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
emjburns: Hmm good point. There might be some weirdness with typing for a common function (I'm not the best at generics). Let me think about this.

keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ServerGroup.kt
emjburns: I'm gonna keep this as is to match titus, and because the suggested name is pretty close. It doesn't seem worth adding the "is"

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
emjburns: removing the comment! good call. it was left over / out of place

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
emjburns: Given that you need two types, and they're both abstract types on the resource handler, I don't think it's worth it to make a separate function. I'd rather think about how we can abstract some of the cluster logic into a generic cluster handler because I think that's where the functionality actually should live.

PR #1466, opened 2020-08-28
feat(api): Standardize mechanism for ResourceHandlers to notify artifact deployments

luispollo:
This PR introduces two new methods to `ResourceHandler`, `notifyArtifactDeploying` and `notifyArtifactDeployed` which standardize the implementation of notifying core Keel of resource handlers beginning and successfully completing artifact deployments, allowing for external plugins to leverage the same mechanism.


emjburns:
😍  I like this implementation so much better! Just confirming - things are wired up correctly for the existing titus and ec2 plugins, right? Just want to make sure you've tested that the events still fire when running locally

luispollo:
> 😍 I like this implementation so much better! Just confirming - things are wired up correctly for the existing titus and ec2 plugins, right? Just want to make sure you've tested that the events still fire when running locally

Yup! Plus, I didn't touch any part of the tests that confirm the behavior around those events, so if the tests pass it means the events are fired! 🙂

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -200,7 +200,7 @@ class ResourceActuator(
           }
         }
       } catch (e: ResourceCurrentlyUnresolvable) {
-        log.warn("Resource check for {} failed (hopefully temporarily) due to {}", id, e.message, e)
+        log.warn("Resource check for {} failed (hopefully temporarily) due to {}", id, e.message)
```
luispollo, line 203: This was causing stack traces to be printed for a case that's an expected condition. It was just noise in the logs.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
```diff
@@ -137,6 +153,26 @@ interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
...
   suspend fun actuationInProgress(resource: Resource<S>): Boolean = false
+
+  /**
+   * Notifies core Keel that a new artifact version is being deployed to the specified resource.
+   *
+   * The default implementation achieves this by publishing an [ArtifactVersionDeploying] event via the
+   * [EventPublisher], and should *not* be overridden by plugin implementations.
+   */
+  @JvmDefault
+  fun notifyArtifactDeploying(resource: Resource<S>, artifactVersion: String) =
```
lorin, line 164: I recommend marking these new methods as `protected` unless you foresee a scenario where they'll be called by anyone other than subclasses.

emjburns:
LGTM!

emjburns:
LGTM!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
luispollo: Can't because it's an interface...

PR #1471, opened 2020-09-02
fix(build): Parity with jooqModelator

luispollo:
Fixes the issue with caching/incremental builds, and the issue starting docker when the container is already running.


keel-sql/keel-sql.gradle.kts
```diff
@@ -65,16 +67,37 @@ dependencies {
...
 
-tasks.register<Exec>("runMysql") {
-  commandLine("docker run --name mysqlJooq -d --rm -e MYSQL_ROOT_PASSWORD=sa -e MYSQL_DATABASE=keel -p 6603:3306 mysql:5.7".split(" "))
-  doLast {
-    // Wait for the DB server to come up...
-    sleep(15 * 1000)
-  }
+tasks.register<Exec>("startMysql") {
+  commandLine("sh", "-c",
+    "docker stop mysqlJooq || " +
```
luispollo, line 72: Hacky, but does the trick -- always try to stop the container just in case before starting.

keel-sql/keel-sql.gradle.kts
```diff
@@ -65,16 +67,37 @@ dependencies {
...
+  dependsOn("mysqlSwitch")
 }
 
 tasks.register<Exec>("stopMysql") {
   commandLine("docker stop mysqlJooq".split(" "))
+  enabled = false
+  dependsOn("mysqlSwitch")
+}
+
+tasks.register("mysqlSwitch") {
```
luispollo, line 86: This was the only trick I could find to make this work. This task has the same caching config as `jooqGenerate`, which means it will only run if there's changes to relevant jooq files, and when it does, it enables the start/stop tasks to run MySQL.

luispollo:
@robfletcher I'm running into an issue where this runs fine if I do `./gradlew keel-sql:build` but not if I run the full build for keel. Any thoughts?

luispollo:
Closing in favor of #1519

PR #1472, opened 2020-09-03
fix(migrator): make migrator registration conditional

nimakaviani:
The migrator is ec2 specific and should only be registered if
the ec2 plugin is enabled.


robfletcher (approved):
We should probably change the way our component scanning works so we don't need to re-specify this.

PR #1474, opened 2020-09-04
tests(tracing): Improve tests for resource tracing

luispollo:
This PR adds additional tests for `X-SPINNAKER-RESOURCE-ID` propagation across threads and coroutines, but unfortunately I'm still unable to reproduce the mismatches we see in production. I'll keep looking into it as time permits, but think it wouldn't hurt to merge this.


lorin:
@luispollo Do you have a sense of how much latency these add to the test suite? I'm generally a fan of adding tests, but I don't know if the benefit is worth the cost here if the additional increase in latency is significant, especially since these tests don't catch the problem we're trying to solve.

luispollo:
> @luispollo Do you have a sense of how much latency these add to the test suite? I'm generally a fan of adding tests, but I don't know if the benefit is worth the cost here if the additional increase in latency is significant, especially since these tests don't catch the problem we're trying to solve.

About 15 seconds:
```
TracingSupportTests injects X-SPINNAKER-RESOURCE-ID to MDC in the coroutine context from resource PASSED
TracingSupportTests injects X-SPINNAKER-RESOURCE-ID to MDC in the coroutine context from exportable PASSED
TracingSupportTests removes X-SPINNAKER-RESOURCE-ID from MDC after block executes PASSED
TracingSupportTests propagates X-SPINNAKER-RESOURCE-ID through nested coroutine contexts PASSED
TracingSupportTests does not mix up X-SPINNAKER-RESOURCE-ID between parallel coroutines PASSED (13.2s)
```

lorin:
I don't think that committing these tests is worth the additional 15s test latency.

PR #1475, opened 2020-09-04
Fix packaging in Titus plugin

robfletcher:
We component scan in the `com.netflix.spinnaker.keel` package but everything in the Titus plugin was in `com.netflix.spinnaker.titus` so any `@Component` annotated beans are ignored. This resulted in the `TitusImageResolver` not being used causing resolution of images to fail.


keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
```diff
@@ -111,5 +113,6 @@ class KeelConfigurationFinalizer(
     log.info("Supported artifacts: {}", artifactSuppliers.joinToString { it.supportedArtifact.name })
     log.info("Using resource handlers: {}", resourceHandlers.joinToString { it.name })
     log.info("Using artifact handlers: {}", artifactHandlers.joinToString { it.name })
+    log.info("Using resolvers: {}", resolvers.joinToString { it.name })
```
robfletcher, line 116: The problem was pretty apparent when I added this bit of logging

luispollo (approved):
Great sleuthing!

PR #1488, opened 2020-09-10
fix(orcaException): fix deserialization issue while parsing errors from orca

gal-yardeni:
keel assumed that failed orca tasks always had certain fields present in the context about the exception. This turned out not to be true. This change makes all exception fields optional so keel doesn't throw a deserialization exception.

In addition, if keel encountered an error when deserializing a task, that would prevent it from processing other tasks that are valid. This PR adds a warning log instead of throwing an exception.

### DIFF
--- a/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
+++ b/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
@@ -65,7 +65,10 @@ class OrcaTaskMonitorAgent(
                     // when we get not found exception from orca, we shouldn't try to get the status anymore
                     taskTrackingRepository.delete(it.id)
                   }
-                  else -> throw e
+                  else -> log.warn(
+                    "Exception ${e.message} has caught while calling orca to fetch status for execution id: ${it.id}" ,
+                    e
+                  )
                 }
                 null
               }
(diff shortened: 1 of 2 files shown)


keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskMonitorAgent.kt
```diff
@@ -65,7 +65,10 @@ class OrcaTaskMonitorAgent(
                     // when we get not found exception from orca, we shouldn't try to get the status anymore
                     taskTrackingRepository.delete(it.id)
                   }
-                  else -> throw e
+                  else -> log.warn(
```
luispollo, line 68: I understand the goal here, but I don't think we can just ignore these errors as those tasks will remain "stuck". We should set the resource status to an error status and/or clean-up the task tracking record, IMO.

gal-yardeni: agree - I think we need to have an additional mechanism for clean up tasks. we should address it in a different PR.

lorin: I think this is strictly better than we have right now, so I'm inclined to get this out for now, and circle back to it.

lorin (approved):

PR #1489, opened 2020-09-10
fix(artifacts): Fix parsing of Netflix semantic versions

luispollo:
Fixes issues parsing variants of semantic version strings used at Netflix.


emjburns (approved):
Thanks!

PR #1492, opened 2020-09-11
Fix deserialization of Titus cluster resources

robfletcher:
This was an error caused when moving from a `@JsonCreator` factory method see https://github.com/spinnaker/keel/blob/v0.173.0/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/cluster/TitusClusterSpec.kt#L95-L112 to a custom deserializer (in order to remove Jackson annotations from the API types).


luispollo (approved):
LGTM.

PR #1493, opened 2020-09-11
fix(db): narrow scope of database query

lorin:
In the `markAsSuccessfullyDeployedTo` function, the query to determine the list of approved artifact versions was too broad. This change strengthens the WHERE clause.


luispollo (approved):
LGTM.

luispollo (approved):

PR #1494, opened 2020-09-11
fix(updateMetadata): fix storing artifact metadata in the DB

gal-yardeni:
We are storing artifact versions differently for each type of artifact:
For example, for debian is `deb-sample-app-server-0.180.0-h410.8cd06f9` (name+version)
for docker is `master-h10.62bbbd6`
and for npm is `1.0.0-10-5d44dc4890536dbb9888b6b6e22f3250734db082`.

Since we are storing the artifact `version` in multiple ways, when we query the `artifact_version` table, we cannot do `artifact_version.version.eq(version)` --> we need to check`artifact_version.version.contains(version)` since `version` is usually just the version (without the name). This PR fixes this issue and adding some logs around it.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
@@ -41,7 +41,7 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
 
   fun store(artifact: DeliveryArtifact, version: String, status: ArtifactStatus?): Boolean
 
-  fun updateArtifactMetadata(name: String, type: ArtifactType, version: String, status: ArtifactStatus?, artifactMetadata: ArtifactMetadata)
+  fun updateArtifactMetadata(name: String, type: ArtifactType, version: String, status: ArtifactStatus?, artifactMetadata: ArtifactMetadata) : Boolean
 
     /**
    * @return Build and Git metadata for a given artifact version
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
@@ -41,7 +41,7 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
 
   fun store(artifact: DeliveryArtifact, version: String, status: ArtifactStatus?): Boolean
 
-  fun updateArtifactMetadata(name: String, type: ArtifactType, version: String, status: ArtifactStatus?, artifactMetadata: ArtifactMetadata) : Boolean
+  fun updateArtifactMetadata(name: String, type: ArtifactType, version: String, status: ArtifactStatus?, artifactMetadata: ArtifactMetadata)
 
     /**
    * @return Build and Git metadata for a given artifact version
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
@@ -190,21 +190,25 @@ class SqlArtifactRepository(
     } == 1
   }
 
-  override fun updateArtifactMetadata(name: String, type: ArtifactType, version: String, status: ArtifactStatus?, artifactMetadata: ArtifactMetadata) {
+  override fun updateArtifactMetadata(name: String, type: ArtifactType, version: String, status: ArtifactStatus?, artifactMetadata: ArtifactMetadata) : Boolean {
     if (!isRegistered(name, type)) {
       throw NoSuchArtifactException(name, type)
     }
 
-    return sqlRetry.withRetry(WRITE) {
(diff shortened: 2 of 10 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -190,21 +190,25 @@ class SqlArtifactRepository(
...
         .set(ARTIFACT_VERSIONS.BUILD_METADATA, objectMapper.writeValueAsString(artifactMetadata.buildMetadata))
         .set(ARTIFACT_VERSIONS.GIT_METADATA, objectMapper.writeValueAsString(artifactMetadata.gitMetadata))
-        .where(ARTIFACT_VERSIONS.NAME.eq(name))
-        .and(ARTIFACT_VERSIONS.TYPE.eq(type))
-        .and(ARTIFACT_VERSIONS.VERSION.eq(version))
+
+        .where(
+          ARTIFACT_VERSIONS.NAME.eq(name),
+          ARTIFACT_VERSIONS.TYPE.eq(type),
+          ARTIFACT_VERSIONS.VERSION.contains(version))
```
gal-yardeni, line 206: this is the major change ^. let me know if there's a better way for doing this...

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -41,7 +41,7 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
 
   fun store(artifact: DeliveryArtifact, version: String, status: ArtifactStatus?): Boolean
 
-  fun updateArtifactMetadata(name: String, type: ArtifactType, version: String, status: ArtifactStatus?, artifactMetadata: ArtifactMetadata)
+  fun updateArtifactMetadata(name: String, type: ArtifactType, version: String, status: ArtifactStatus?, artifactMetadata: ArtifactMetadata) : Boolean
```
luispollo, line 44: You added the return type but it doesn't look like you're using it anywhere?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: AFAIK, this is only a problem for Debians for historical reasons. The way we've approached handling it elsewhere is slightly different, for example:
https://github.com/spinnaker/keel/blob/9c83039ce03194841c27d6665cfb1304d3e77b60/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt#L71-L78

So, another way of doing this would be something like:
```suggestion
          ARTIFACT_VERSIONS.VERSION.eq(version).or(ARTIFACT_VERSIONS.VERSION.eq("$name-$version"))
```

robfletcher (commented):
Would it make sense to fix the inconsistency in the way we store versions rather than working round it?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
gal-yardeni: Will remove it!

gal-yardeni:
@robfletcher I agree it should be aligned. I'm currently not in the right mindset to make the change since I'm dealing with lots of other edge cases that need to be debugged and I'm afraid it can cause some migrations, but I will open an issue for it and try to handle it asap. cc @emjburns

gal-yardeni:
opened https://github.com/spinnaker/keel/issues/1497 to track artifacts versions alignment issue.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -190,18 +190,20 @@ class SqlArtifactRepository(
...
         .set(ARTIFACT_VERSIONS.BUILD_METADATA, objectMapper.writeValueAsString(artifactMetadata.buildMetadata))
         .set(ARTIFACT_VERSIONS.GIT_METADATA, objectMapper.writeValueAsString(artifactMetadata.gitMetadata))
-        .where(ARTIFACT_VERSIONS.NAME.eq(name))
-        .and(ARTIFACT_VERSIONS.TYPE.eq(type))
-        .and(ARTIFACT_VERSIONS.VERSION.eq(version))
+
+        .where(
+          ARTIFACT_VERSIONS.NAME.eq(name),
+          ARTIFACT_VERSIONS.TYPE.eq(type),
+          ARTIFACT_VERSIONS.VERSION.eq(version).or(ARTIFACT_VERSIONS.VERSION.eq("$name-$version")))
```
luispollo, line 206: It would be good to add a test to cover this change in the query.

luispollo (approved):
It'd be nice to add a test, otherwise LGTM.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: added one test :)

PR #1500, opened 2020-09-12
fix(artifacts): More fixes for NPM version parsing

luispollo:
In this PR:
- Update for shaded Rocket's `DebianVersionComparator` which fixes the `NumberFormatException`
- Additional updates to the regular expression used in `NpmArtifactSupplier` such that it matches the format used for Debian packages at Netflix and allows us to continue to use the same comparator.

As noted in the `TODOs` added in this PR, in my opinion we should consolidate/simplify our parsing of semantic version strings, ideally by using an off-the-shelf library like we're already using for Docker tags, but at the very least consolidating between our usage of the `frigga` library's `AppVersion` class and the Rocket code when comparing versions for Debian and NPM.

### DIFF
--- a/keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifactSupplierTests.kt
+++ b/keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifactSupplierTests.kt
@@ -18,12 +18,12 @@ import com.netflix.spinnaker.keel.services.ArtifactMetadataService
 import com.netflix.spinnaker.keel.test.deliveryConfig
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
-import io.mockk.coEvery as every
-import io.mockk.coVerify as verify
 import io.mockk.mockk
 import kotlinx.coroutines.runBlocking
 import strikt.api.expectThat
 import strikt.assertions.isEqualTo
+import io.mockk.coEvery as every
+import io.mockk.coVerify as verify
 
 internal class NpmArtifactSupplierTests : JUnit5Minutests {
   object Fixture {
@@ -36,14 +36,14 @@ internal class NpmArtifactSupplierTests : JUnit5Minutests {
       deliveryConfigName = deliveryConfig.name,
       statuses = setOf(CANDIDATE)
     )
-    val versions = listOf("1.0.0-rc", "1.0.0", "1.0.1-5", "1.0.2-h6", "1.0.3-h7-gc0c60369")
+    val versions = listOf("1.0.0-rc", "1.0.0-rc.1", "1.0.0", "1.0.1-5", "1.0.2-h6", "1.0.3-rc-h7.gc0c60369")
     val latestArtifact = PublishedArtifact(
       name = npmArtifact.name,
       type = npmArtifact.type,
       reference = npmArtifact.reference,
-      version = "${npmArtifact.name}-${versions.last()}",
-      metadata = mapOf("releaseStatus" to CANDIDATE, "buildNumber" to "1", "commitId" to "a15p0")
-      )
+      version = versions.last(),
+      metadata = mapOf("releaseStatus" to CANDIDATE, "buildNumber" to "7", "commitId" to "gc0c60369")
+    )
     val npmArtifactSupplier = NpmArtifactSupplier(eventBridge, artifactService, artifactMetadataService)
 
     val artifactMetadata = ArtifactMetadata(
@@ -56,10 +56,10 @@ internal class NpmArtifactSupplierTests : JUnit5Minutests {
           name = "job bla bla",
           link = "enkins.com"
         ),
-        number = "1"
+        number = "7"
       ),
       GitMetadata(
-        commit = "a15p0",
+        commit = "gc0c60369",
         author = "keel-user",
         repo = Repo(
           name = "keel",
@@ -70,7 +70,7 @@ internal class NpmArtifactSupplierTests : JUnit5Minutests {
           url = "www.github.com/pr/111"
         ),
         commitInfo = Commit(
-          sha = "a15p0",
+          sha = "gc0c60369",
           message = "this is a commit message",
           link = ""
         ),
@@ -91,7 +91,7 @@ internal class NpmArtifactSupplierTests : JUnit5Minutests {
           artifactService.getArtifact(npmArtifact.name, versions.last(), NPM)
         } returns latestArtifact
         every {
-          artifactMetadataService.getArtifactMetadata("1", "a15p0")
+          artifactMetadataService.getArtifactMetadata("7", "gc0c60369")
         } returns artifactMetadata
       }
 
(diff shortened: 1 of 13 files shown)


keel-core/src/main/java/com/netflix/rocket/semver/shaded/DebianVersionComparator.java
```diff
@@ -93,9 +93,17 @@ private int compareMainPart(String s0, String s1) {
       if (i >= mainParts1.length) {
         return 1;
       }
-      int diff = Integer.parseInt(mainParts0[i]) - Integer.parseInt(mainParts1[i]);
-      if (diff != 0) {
-        return diff;
+      boolean isNumeric0 = isNumeric(mainParts0[i]);
```
luispollo, line 96: This fixes the `NumberFormatException` we were seeing in the logs.

keel-core/src/main/java/com/netflix/rocket/semver/shaded/DebianVersionComparator.java
```diff
@@ -4,7 +4,7 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-public class NetflixVersionComparator implements Comparator<String> {
+public class DebianVersionComparator implements Comparator<String> {
```
luispollo, line 7: I initially thought when @gal-yardeni renamed this that it was a good idea, but since we're just copying/shading the code here, it's easier to keep the name intact.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NetflixSemVerVersioningStrategy.kt
```diff
@@ -15,13 +15,13 @@ object NetflixSemVerVersioningStrategy : VersioningStrategy {
...
     // version digits (capturing group 1)
-    "(\\d+\\.\\d+\\.\\d+)" +
+  "(\\d+\\.\\d+\\.\\d+)" +
     // release qualifier (capturing group 3)
-    "([-~](dev|snapshot|rc|rel|final))?" +
-    // build number (capturing group 5)
-    "([-\\.][h]?(\\d+\\b))?" +
-    // commit hash (capturing group 7)
-    "(-(\\w{7,}))?"
+    "([-~](dev|snapshot|rc)(\\.\\d+)?)?" +
```
luispollo, line 20: Pre-release qualifiers can include a counter separated by a dot.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NetflixSemVerVersioningStrategy.kt
```diff
@@ -15,13 +15,13 @@ object NetflixSemVerVersioningStrategy : VersioningStrategy {
...
+    // build number (capturing group 6)
+    "(-[h]?(\\d+\\b))?" +
```
luispollo, line 22: The separator here must be a hyphen, always.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NetflixSemVerVersioningStrategy.kt
```diff
@@ -15,13 +15,13 @@ object NetflixSemVerVersioningStrategy : VersioningStrategy {
...
+    // commit hash (capturing group 8)
+    "(\\.(\\w{7,}))?"
```
luispollo, line 24: The separator here must be a dot, always.

luispollo:
@emjburns I decided not to change the comparator used in `NetflixSemverVersioningStrategy` because it's used by both Debian and NPM. Instead, I'm aligning NPM to match Debian, both in Rocket and keel.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifactSupplierTests.kt
```diff
@@ -36,14 +36,14 @@ internal class NpmArtifactSupplierTests : JUnit5Minutests {
...
+    val versions = listOf("1.0.0-rc", "1.0.0-rc.1", "1.0.0", "1.0.1-5", "1.0.2-h6", "1.0.3-rc-h7.gc0c60369")
     val latestArtifact = PublishedArtifact(
       name = npmArtifact.name,
       type = npmArtifact.type,
       reference = npmArtifact.reference,
-      version = "${npmArtifact.name}-${versions.last()}",
-      metadata = mapOf("releaseStatus" to CANDIDATE, "buildNumber" to "1", "commitId" to "a15p0")
-      )
+      version = versions.last(),
+      metadata = mapOf("releaseStatus" to CANDIDATE, "buildNumber" to "7", "commitId" to "gc0c60369")
```
gal-yardeni, line 45: As far as I can see, the commit id we present is 7 chars (I'm also fixing it in my PR)

keel-core/src/main/java/com/netflix/rocket/semver/shaded/DebianVersionComparator.java
gal-yardeni: So this comparator is not applicable for NPMs?

luispollo: It is, and we're still using it for NPMs, but as I said this class is copied "as is" from a Rocket code repo (hence the `shaded` in the package name), and so it's confusing to rename the class here. Any time we want to update the class (like I did in this PR), you'll have to fix the name. I thought it'd be better to keep the original name.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifactSupplierTests.kt
luispollo: Yes, you're right, that's the size of a "short hash". I'm not sure why this example had a longer hash, I'll fix it. My reason for changing things here was simply to have consistency between the metadata and the version string.

keel-core/src/main/java/com/netflix/rocket/semver/shaded/DebianVersionComparator.java
lorin: Just to confirm my understanding, the behavior here is:

* if exactly one argument is numeric, the numeric argument is greater
* If neither is numeric: s1 is greater

If neither is numeric, should we do an equality check?

lorin (approved):
I had one question, otherwise LGTM!

keel-core/src/main/java/com/netflix/rocket/semver/shaded/DebianVersionComparator.java
luispollo: TBH, I didn't spend time reviewing this other than it would avoid `NumberFormatException`, because the code is literally copied/pasted as is from Rocket.

PR #1504, opened 2020-09-13
fix(titus): Revert ignoring serialization of container field in TitusClusterSpec

luispollo:
We were ignoring the `container` field when serializing, but requiring it when deserializing:
https://github.com/spinnaker/keel/blob/9e944c253820818464d92e38a84be98c474671a3/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/jackson/TitusClusterSpecDeserializer.kt#L48-L49

This just removes the unnecessary annotated field from the mixin.

### DIFF
--- a/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/jackson/mixins/TitusClusterSpecMixin.kt
+++ b/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/jackson/mixins/TitusClusterSpecMixin.kt
@@ -29,9 +29,6 @@ interface TitusClusterSpecMixin {
   val artifactVersion: String?
 
   @get:JsonIgnore
-  val container: ContainerProvider
-
-  @get:JsonIgnore
   val maxDiffCount: Int?
 
   @get:JsonIgnore


luispollo:
Cc:@lorin it would be nice to add coverage for this in ballast.

robfletcher (commented):
Should it be write only? Trying to think why we ever thought this field should be ignored.

luispollo:
> Should it be write only? Trying to think why we ever thought this field should be ignored.

I know, right? I remember thinking "why are we ignoring `container`?" when I reviewed, but got a bit confused between that field and the `containerAttributes` you added, and thought I was just being paranoid. We can make it write-only, but I think it doesn't matter because we're using a custom deserializer...

robfletcher:
I'm planning on removing the custom deserializer tomorrow!

robfletcher:
I think this may go back to a weird situation we used to have where the top level `TitusClusterSpec` had a `containerProvider` property, then `TitusServerGroupSpec` had a `container` property. It didn't really make any sense so I removed the duplication but I guess I missed that we no longer needed to be ignoring the property.

robfletcher (approved):

PR #1507, opened 2020-09-15
fix(artifacts): Fix Netflix semver version comparisons

luispollo:
We were still failing to parse valid versions of NPM artifacts with Frigga in `NETFLIX_SEMVER_COMPARATOR`. This PR fixes that by replacing the use of Frigga in this case with our internal parser, using a regular expression we know matches Rocket and accommodates for version strings with and without the package name as a prefix.


lorin (approved):
LGTM. 

Do you happen to know what the behavior is if the version strings are parseable but we can't determine an ordering?

For example, different packages entirely, or two versions that differ only in the sha

```
foo-0.368.0-h473.dfc46fb <> bar-0.367.0-h472.abc53cd

0.368.0-h473.dfc46fb <> 0.368.0-h473.0493add
```

luispollo:
> LGTM.
> 
> Do you happen to know what the behavior is if the version strings are parseable but we can't determine an ordering?
> 
> For example, different packages entirely, or two versions that differ only in the sha
> 
> ```
> foo-0.368.0-h473.dfc46fb <> bar-0.367.0-h472.abc53cd
> 
> 0.368.0-h473.dfc46fb <> 0.368.0-h473.0493add
> ```

For the first case, with the change in this PR, we'd ignore the package name and look only at the version. This would only happen if we messed up storing artifacts, though. I believe the second case is just comparing the commit portion lexicographically.

PR #1510, opened 2020-09-16
fix(artifacts): Another attempt at fixing NPM versioning

luispollo:
Recognizing that the versioning scheme we use at Netflix for Debian packages is not fully compliant with the SemVer spec, whereas NPM (specifically the `npm version` command) actually enforces valid SemVer version strings, this PR splits out the comparators for Debian and NPM, using our existing chosen library for SemVer for the latter.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianVersioningStrategy.kt
```diff
@@ -0,0 +1,20 @@
+package com.netflix.spinnaker.keel.artifacts
+
+import com.netflix.spinnaker.keel.api.artifacts.VersioningStrategy
+
+/**
+ * A [VersioningStrategy] that for Debian packages..
+ */
+object DebianVersioningStrategy : VersioningStrategy {
+  override val type: String = "debian"
```
luispollo, line 9: The fact that this has a different name than the version strategy we were using previously shouldn't have any impact as it's not present in delivery configs or persisted to the database.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NpmVersioningStrategy.kt
```diff
@@ -0,0 +1,20 @@
+package com.netflix.spinnaker.keel.artifacts
+
+import com.netflix.spinnaker.keel.api.artifacts.VersioningStrategy
+
+/**
+ * A [VersioningStrategy] that for NPM packages.
+ */
+object NpmVersioningStrategy : VersioningStrategy {
+  override val type: String = "npm"
```
luispollo, line 9: The fact that this has a different name than the version strategy we were using previously shouldn't have any impact as it's not present in delivery configs or persisted to the database.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -20,21 +20,23 @@ import org.slf4j.LoggerFactory
...
+ * Note: this implementation currently makes some Netflix-specific assumptions with regards to artifact
+ * versions so that it can extract build and commit metadata.
  */
 @Component
 class DebianArtifactSupplier(
   override val eventPublisher: EventPublisher,
   private val artifactService: ArtifactService,
   override val artifactMetadataService: ArtifactMetadataService
-) : BaseArtifactSupplier<DebianArtifact, NetflixSemVerVersioningStrategy>(artifactMetadataService) {
+) : BaseArtifactSupplier<DebianArtifact, DebianVersioningStrategy>(artifactMetadataService) {
```
luispollo, line 33: Debian artifacts now use a Debian-specific versioning strategy.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifactSupplier.kt
```diff
@@ -18,20 +18,22 @@ import com.netflix.spinnaker.kork.exceptions.IntegrationException
...
+ * Note: this implementation currently makes some Netflix-specific assumptions with regards to artifact
+ * versions so that it can extract build and commit metadata.
  */
 @Component
 class NpmArtifactSupplier(
   override val eventPublisher: EventPublisher,
   private val artifactService: ArtifactService,
   override val artifactMetadataService: ArtifactMetadataService
-) : BaseArtifactSupplier<NpmArtifact, NetflixSemVerVersioningStrategy>(artifactMetadataService) {
+) : BaseArtifactSupplier<NpmArtifact, NpmVersioningStrategy>(artifactMetadataService) {
```
luispollo, line 31: Debian artifacts now use an NPM-specific versioning strategy.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/VersioningStrategyComparators.kt
```diff
@@ -106,3 +93,37 @@ val NETFLIX_SEMVER_COMPARATOR: Comparator<String> = object : Comparator<String>
 
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 }
+
+// descending by default
+val NPM_VERSION_COMPARATOR: Comparator<String> = object : Comparator<String> {
+  override fun compare(s1: String, s2: String) =
+    semverComparator.compare(s1.toVersion(), s2.toVersion())
```
luispollo, line 100: The NPM comparator now just uses SemVer.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifactSupplier.kt
robfletcher: NPM artifacts, right?

luispollo: Darn copy/paste! 😆

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NpmVersioningStrategy.kt
robfletcher: Are you sure about that? It appears in the JSON schema. If it's not in the delivery config that's a bug

luispollo: I was surprised by this initially when I changed the name from `deb` to `netflix-semver` a couple months back and it didn't break, but I found the reason is that the versioning strategies for debian and npm are hard-coded as defaults and we don't really support anything else for those artifacts (see `supportedVersioningStrategy` below), so if a user would specify something different, _then_ things would break. I understand this is not "correct" per the schema, but would prefer to address that separately if you think it's worth the trouble.

Debian: https://github.com/spinnaker/keel/pull/1510/files#diff-ed217978125ec043314d9063f798b04fR18

NPM: https://github.com/spinnaker/keel/pull/1510/files#diff-dbe8bb0ebf9249a6504795a1236a1d61R16

lorin (approved):
LGTM

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NpmVersioningStrategy.kt
luispollo: @robfletcher I'm merging this so I can test it with live data, but feel free to comment if you disagree with the approach.

luispollo: P.S. About the fact that it appears in the schema, it should still be marked optional since it has a default, correct? I don't expect any users to be specifying the versioning strategy for Debian or NPM, and given that the schema just went live (meaning no one would have seen the old discriminator), I hope this will be OK.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/VersioningStrategyComparators.kt
```diff
@@ -106,3 +93,37 @@ val NETFLIX_SEMVER_COMPARATOR: Comparator<String> = object : Comparator<String>
...
+
+/**
+ * Trims a leading "v" off of the semver if present
+ */
+private fun parseSemver(input: String?): SemVer? {
+  input ?: return null
+  return try {
+    SemVer.parse(input.removePrefix("v"))
+  } catch (e: IllegalArgumentException) {
+    null
```
gal-yardeni, line 126: I know it's already merged -- but should we log those exceptions?

PR #1513, opened 2020-09-17
more test cases for schema validation

robfletcher:
Pulled and anonymized some example data from the real database as some of our examples were a little stale.

### DIFF
--- a/keel-web/src/test/resources/examples/titus-cluster-with-artifact-example.yml
+++ b/keel-web/src/test/resources/examples/titus-cluster-with-artifact-example.yml
@@ -1,30 +1,36 @@
 ---
-kind: titus/cluster@v1
-metadata:
-  serviceAccount: my-email@spinnaker.io
-spec:
-  moniker:
-    app: keeldemo
-    stack: examples
-    detail: ec2v1
-  container:
-    organization: emburns
-    image: spin-titus-demo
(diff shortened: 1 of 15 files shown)


luispollo (approved):
Luv it! Thanks for taking the time to do this! ❤️

keel-web/src/test/resources/examples/titus-cluster-with-artifact-example.yml
```diff
@@ -1,30 +1,33 @@
...
-  dependencies:
-    loadBalancerNames: []
-    securityGroupNames:
-      - keeldemo
-      - nf-datacenter
-      - nf-infrastructure
-    targetGroups: []
+application: fnord
+serviceAccount: delivery-engineering@netflix.com
+environments:
```
gal-yardeni, line 4: where is the artifact here? :)

gal-yardeni (approved):
one minor comment, other than that LGTM!

keel-web/src/test/resources/examples/titus-cluster-with-artifact-example.yml
robfletcher: Good point. These were previously just resources rather than full delivery configs, so I'm not sure why the file was called that originally.

PR #1522, opened 2020-09-18
feat(notifications): Improve manual judgement notifications

### LINKED ISSUE #1266: Add more valuable information into Slack notifications
It would be useful to add the environment being operated upon into the slack notifications, similar to what RADAR is doing:

image (5)

luispollo:
Improves the manual judgement (Slack) notification by adding a link to the corresponding artifact version in the Environments view and restructuring the format to make it more easily readable.

Sample:
image

Closes #1266

### DIFF
--- a/keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ManualJudgementNotifier.kt
+++ b/keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ManualJudgementNotifier.kt
@@ -6,8 +6,12 @@ import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus
 import com.netflix.spinnaker.keel.api.events.ConstraintStateChanged
 import com.netflix.spinnaker.keel.core.api.ManualJudgementConstraint
 import com.netflix.spinnaker.keel.echo.model.EchoNotification
+import com.netflix.spinnaker.keel.persistence.KeelRepository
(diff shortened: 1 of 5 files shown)


luispollo:
Cc: @erikmunson

erikmunson:
yes

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Environment.kt
```diff
@@ -1,6 +1,7 @@
 package com.netflix.spinnaker.keel.api
 
 data class Environment(
+  val application: String,
```
luispollo, line 4: I added this here because we were doing `environment.resources.first()?.application`, which was hacky and could return `null`.

gal-yardeni:
Thanks for doing this!
A have a general comment - should we ditch the `version` completely at this point? 
By looking at the code, in ` ConstraintStateChanged.toEchoNotification` you can call to get the artifact metadata by artifact version and type and get much more useful information :) (That was my original plan actually)

luispollo:
> Thanks for doing this!
> A have a general comment - should we ditch the `version` completely at this point?
> By looking at the code, in ` ConstraintStateChanged.toEchoNotification` you can call to get the artifact metadata by artifact version and type and get much more useful information :) (That was my original plan actually)

Good point, @gal-yardeni! I'll add that and post a sample! Thanks for the feedback.

lorin (approved):
LGTM

PR #1523, opened 2020-09-20
feat(UpdateOldVersions): admin endpoint for updating old artifact versions

gal-yardeni:
Adding a new admin endpoint for updating old artifact version records with their metadata. 
Each artifact supplier had a different implementation for getting an artifact by its version, so added support for it as well (very similar to `getLatestArtifact`).
I'm sure we can optimize this code, but my goal, for now, is to push it though the environments and run it one time per env, then we can decide whether to remove it completely or optimize it more. Appriciate your thoughts!

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
@@ -9,6 +9,7 @@ import com.netflix.spinnaker.keel.pause.ActuationPauser
 import com.netflix.spinnaker.keel.persistence.DiffFingerprintRepository
 import com.netflix.spinnaker.keel.persistence.KeelRepository
 import com.netflix.spinnaker.keel.telemetry.ArtifactSaved
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import org.slf4j.LoggerFactory
 import org.springframework.context.ApplicationEventPublisher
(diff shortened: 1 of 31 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
```diff
@@ -56,4 +63,38 @@ class AdminService(
...
+    //1. get all registered artifacts
+    val deliveryArtifacts = repository.getAllArtifacts(type)
+    //2. for each artifact, fetch all versions
+    deliveryArtifacts.forEach { artifact ->
+      val versions = repository.artifactVersions(artifact)
+      versions.forEach { version ->
+        val status = repository.getReleaseStatus(artifact, version)
+        //don't update if metadata is already exists
+        if (repository.getArtifactBuildMetadata(artifact.name, type, version, status) == null) {
+          runBlocking {
```
luispollo, line 82: Consider running this `async` so that you'll do this in parallel instead of sequentially.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
```diff
@@ -56,4 +63,38 @@ class AdminService(
...
+            try {
+              //3. for each version, get information commit and build number from artifact supplier
+              val publishedArtifact = artifactSupplier.getArtifactByVersion(artifact, version)
+              //4. send an ArtifactSaved event, which is responsible for getting artifact metadata and storing it
+              if (publishedArtifact != null) {
+                publisher.publishEvent(ArtifactSaved(publishedArtifact, status))
+              }
+            } catch (ex: Exception) {
+              log.error("error has accrued while trying to get artifact by version or it's metadata for artifact ${artifact.name}" +
```
luispollo, line 91:
```suggestion
              log.error("error trying to get artifact by version or its metadata for artifact ${artifact.name}" +
```

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
```diff
@@ -61,4 +61,15 @@ class AdminController(
   ) {
     adminService.forceConstraintReevaluation(application, environment, type)
   }
+
+  //This endpoint will update old artifact version records with their metadata, if exists, by type [deb/docker/npm].
+  //Please note: it's an admin endpoint and is not intent to be used more than once per environment for now.
+  @PostMapping(
+    path = ["/backfillArtifacts/{type}"]
```
luispollo, line 68: Suggestion:
```suggestion
    path = ["/artifacts/{type}/metadata/backfill"]
```

luispollo (approved):
Nice! Minor suggestions, but LGTM!

PR #1525, opened 2020-09-21
Better representation of resource kinds in schema

robfletcher:
Now each type of resource specifies the kind alongside the matching spec format in a way JSON schema validation can pick up.

Instead of being: 
```yaml
$defs:
  SubmittedResource:
    properties:
      metadata:
        type: object
      kind:
        enum:
        - ec2/cluster@v1
        - titus/cluster@v1

  ClusterSpecSubmittedResource:
    allOf:
    - $ref: #/$defs/SubmittedResource
    - type: object
      properties:
        spec:
          $ref: #/$defs/ClusterSpec

  TitusClusterSpecSubmittedResource:
    allOf:
    - $ref: #/$defs/SubmittedResource
    - type: object
      properties:
        spec:
          $ref: #/$defs/TitusClusterSpec
```

It will now be:
```yaml
$defs:
  SubmittedResource:
    properties:
      kind:
        enum:
        - ec2/cluster@v1
        - titus/cluster@v1
    allOf:
    - if:
        properties:
          kind:
            const: ec2/cluster@v1
      then:
        properties:
          spec:
            $ref: #/$defs/ClusterSpec
    - if:
        properties:
          kind:
            const: titus/cluster@v1
      then:
        properties:
          spec:
            $ref: #/$defs/TitusClusterSpec
```

This means that the JSON schema can enforce that the correct `kind` goes with the correct type of `spec`. Note how there's not actually any association between `kind` and the type of `spec` beforehand. There's a good explanation of this format [here](http://json-schema.org/understanding-json-schema/reference/conditionals.html).

This also let me drop the `discriminator` and `mapping` fields that are a convention from Open API and not actually enforced by JSON schema tools.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ClusterDeployStrategy.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ClusterDeployStrategy.kt
@@ -1,10 +1,12 @@
 package com.netflix.spinnaker.keel.api
 
 import com.netflix.spinnaker.keel.api.DeployHealth.AUTO
+import com.netflix.spinnaker.keel.api.schema.Discriminator
 import java.time.Duration
 import java.time.Duration.ZERO
 
-sealed class ClusterDeployStrategy {
+abstract class ClusterDeployStrategy {
+  @Discriminator abstract val strategy: String
   open val isStaggered: Boolean = false
   open val stagger: List<StaggeredRegion> = emptyList()
   abstract val health: DeployHealth
@@ -26,13 +28,17 @@ data class RedBlack(
   // The order of this list is important for pauseTime based staggers
   override val stagger: List<StaggeredRegion> = emptyList()
 ) : ClusterDeployStrategy() {
+  override val strategy = "red-black"
+
   override val isStaggered: Boolean
     get() = stagger.isNotEmpty()
 }
 
 data class Highlander(
   override val health: DeployHealth = AUTO
-) : ClusterDeployStrategy()
+) : ClusterDeployStrategy() {
+  override val strategy = "highlander"
+}
 
 enum class DeployHealth {
   /** Use Orca's default (Discovery and ELB/Target group if attached). */
--- a/keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ApplicationLoadBalancerSpec.kt
+++ b/keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ApplicationLoadBalancerSpec.kt
@@ -7,11 +7,12 @@ import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec.Action
 import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec.Listener
 import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec.TargetGroup
 import com.netflix.spinnaker.keel.api.ec2.LoadBalancerType.APPLICATION
+import com.netflix.spinnaker.keel.api.schema.Optional
 import java.time.Duration
 
 data class ApplicationLoadBalancerSpec(
   override val moniker: Moniker,
-  override val locations: SubnetAwareLocations,
+  @Optional override val locations: SubnetAwareLocations,
   override val internal: Boolean = true,
   override val dependencies: LoadBalancerDependencies = LoadBalancerDependencies(),
   override val idleTimeout: Duration = Duration.ofSeconds(60),
--- a/keel-schema-generator/src/main/kotlin/com/netflix/spinnaker/keel/schema/Generator.kt
+++ b/keel-schema-generator/src/main/kotlin/com/netflix/spinnaker/keel/schema/Generator.kt
@@ -21,8 +21,6 @@ import kotlin.reflect.KFunction
 import kotlin.reflect.KParameter
 import kotlin.reflect.KProperty
 import kotlin.reflect.KType
-import kotlin.reflect.KTypeProjection.Companion.invariant
-import kotlin.reflect.full.createType
 import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.full.hasAnnotation
 import kotlin.reflect.full.isSubclassOf
@@ -78,6 +76,7 @@ class Generator(
       properties = schema.properties,
       required = schema.required,
       discriminator = schema.discriminator,
+      allOf = schema.allOf,
       `$defs` = context.definitions.toSortedMap(String.CASE_INSENSITIVE_ORDER)
     )
   }
@@ -93,7 +92,7 @@ class Generator(
       type.isSingleton -> buildSchemaForKotlinSingleton(type)
       type.isSealed -> buildSchemaForSealedClass(type)
       extensionRegistry.baseTypes().contains(type.java) -> buildSchemaForTypeHierarchy(type)
-      type.typeParameters.isNotEmpty() -> buildSchemaForGenericTypeHierarchy(type, discriminator)
+      type.typeParameters.isNotEmpty() -> buildSchemaForGenericTypeHierarchy(type)
       // Otherwise this is just a regular object schema
       else -> buildSchemaForRegularClass(type, discriminator)
     }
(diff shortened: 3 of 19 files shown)


keel-schema-generator/src/test/kotlin/com/netflix/spinnaker/keel/schema/GeneratorTests.kt
```diff
@@ -688,14 +687,46 @@ internal class GeneratorTests {
...
+                .one {
+                  isA<ObjectSchema>().get { properties.keys }
+                    .containsExactlyInAnyOrder(
+                      Foo<*>::type.name,
+                      Foo<*>::genericProperty.name,
+                    )
+                }
+            },
+
+            dynamicTest("generic sub-type discriminator in $subTypeName is a single value enums") {
```
luispollo, line 717: If it's a single value, why does it need to be an `enum`?

robfletcher: Because that's how JSON schema represents that. You can't specify a string literal like you can in Cue.

robfletcher: Huh, actually draft 6 introduced a `const` keyword for this. I'm making some changes to how this works anyway so I'll work that in.

keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ApplicationLoadBalancerSpec.kt
```diff
@@ -7,11 +7,12 @@ import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec.Action
 import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec.Listener
 import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec.TargetGroup
 import com.netflix.spinnaker.keel.api.ec2.LoadBalancerType.APPLICATION
+import com.netflix.spinnaker.keel.api.schema.Optional
 import java.time.Duration
 
 data class ApplicationLoadBalancerSpec(
   override val moniker: Moniker,
-  override val locations: SubnetAwareLocations,
+  @Optional override val locations: SubnetAwareLocations,
```
robfletcher, line 15: Once I better enforced the schema I discovered a few missing things, such as this field that needs to be `@Optional` (because the `locations` value can get inherited from the environment)

keel-schema-generator/src/main/kotlin/com/netflix/spinnaker/keel/schema/Generator.kt
```diff
@@ -93,7 +92,7 @@ class Generator(
       type.isSingleton -> buildSchemaForKotlinSingleton(type)
       type.isSealed -> buildSchemaForSealedClass(type)
       extensionRegistry.baseTypes().contains(type.java) -> buildSchemaForTypeHierarchy(type)
-      type.typeParameters.isNotEmpty() -> buildSchemaForGenericTypeHierarchy(type, discriminator)
+      type.typeParameters.isNotEmpty() -> buildSchemaForGenericTypeHierarchy(type)
```
robfletcher, line 95: Parameter was never used and confused things inside the method

keel-schema-generator/src/main/kotlin/com/netflix/spinnaker/keel/schema/Schema.kt
```diff
@@ -94,6 +101,20 @@ data class AllOf(
...
+)
+
+data class Condition(
+  val properties: Map<String, ConstSchema>
+)
+
+data class Subschema(
+  val properties: Map<String, Schema>,
+  val required: SortedSet<String> = emptySet<String>().toSortedSet()
+)
```
robfletcher, line 116: New bits of model for representing the conditional sub-schema

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/apidocs/ApiDocCompatibilityTests.kt
```diff
@@ -60,11 +75,24 @@ class ApiDocCompatibilityTests {
...
+      "/examples/wrong-resource-kind.yml"
+    ).map {
+      dynamicTest("example delivery config ${it.substringAfterLast("/")} is invalid") {
+        val messages = schema.validate(loadExample(it))
+        expectThat(messages)
+          .describedAs("validation messages for ${it.substringAfterLast("/")}")
+          .isNotValid()
+      }
+    }
+
```
robfletcher, line 95: A test that checks that a resource where the `kind` and `spec` don't match is rejected. This failed before these changes.

keel-web/src/test/resources/examples/cluster-example.yml
```diff
@@ -22,7 +22,7 @@ environments:
     spec:
       health:
         terminationPolicies:
-        - Default
+        - default
```
robfletcher, line 25: I had to change a bunch of things in these example files to lowercase because JSON schema has no way of specifying an enum as case-insensitive. We'll still actually parse case-insensitively but it's not really practical to list all the possible variations in the schema. Still thinking about ways to improve this but I don't want it to just be free-form text because that's not very helpful for users.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ClusterDeployStrategy.kt
```diff
@@ -1,10 +1,12 @@
 package com.netflix.spinnaker.keel.api
 
 import com.netflix.spinnaker.keel.api.DeployHealth.AUTO
+import com.netflix.spinnaker.keel.api.schema.Discriminator
 import java.time.Duration
 import java.time.Duration.ZERO
 
-sealed class ClusterDeployStrategy {
+abstract class ClusterDeployStrategy {
+  @Discriminator abstract val strategy: String
```
robfletcher, line 9: I had to make this change in order that the schema generator outputs something where a validator can tell the difference between `RedBlack` and `Highlander`. It makes no difference to the YAML format.

keel-schema-generator/src/main/kotlin/com/netflix/spinnaker/keel/schema/Schema.kt
```diff
@@ -18,11 +18,11 @@ data class RootSchema(
   val description: String?,
   val properties: Map<String, Schema>,
   val required: SortedSet<String>,
-  val discriminator: OneOf.Discriminator? = null,
+  val allOf: List<ConditionalSubschema>? = null,
   val `$defs`: SortedMap<String, Schema>
 ) {
   @Suppress("unused", "PropertyName")
-  val `$schema`: String = "https://json-schema.org/draft/2019-09/schema"
+  val `$schema`: String = "http://json-schema.org/draft/2019-09/schema#"
```
lorin, line 25: What's the motiviation for this URL change?

keel-schema-generator/src/main/kotlin/com/netflix/spinnaker/keel/schema/Schema.kt
```diff
@@ -79,21 +84,29 @@ data class Reference(
...
+  val properties: Map<String, ConstSchema>
+)
+
+data class Subschema(
+  val properties: Map<String, Schema>,
+  val required: SortedSet<String> = emptySet<String>().toSortedSet()
+)
+
 /**
  * Yes, I really had to implement an either monad to get this all to work.
```
lorin, line 111: Never apologize for implementing monads.

keel-schema-generator/src/main/kotlin/com/netflix/spinnaker/keel/schema/Schema.kt
robfletcher: I had just gotten it wrong (I think because I copied & pasted from my URL bar). Update was based on this documentation: http://json-schema.org/understanding-json-schema/reference/schema.html#schema

keel-schema-generator/src/main/kotlin/com/netflix/spinnaker/keel/schema/Schema.kt
robfletcher: I'm kind of amazed Kotlin doesn't have one by default

keel-web/src/test/resources/examples/cluster-example.yml
lorin: It's kinda sucky, but could you enumerate lower case, upper case, and all caps as separate options as a workaround, e.g. (I don't know the actual syntax) `[default, Default, DEFAULT]`?

lorin (approved):
LGTM, although I'm not familiar enough with this code to do a meaningful review of some of these changes.

PR #1530, opened 2020-09-23
refactor(artifacts): Simplify processing of published artifacts

luispollo:
Refactors processing of published artifacts, in preparation to add artifact timestamps, as follows:
- Introduces the use of a proper model when dealing with published artifact versions in the repositories
- Replaces the use of two separate operations to write the `ARTIFACT_VERSIONS` record for basic published artifact information (in `ArtifactListener`) and metadata (in `ArtifactStoredListener`), consolidates both in a single `storeArtifactVersion(...)` function that takes a `PublishedArtifact` with all the properties
- Removes `ArtifactSaved` event and `ArtifactStoredListener` -- they were unnecessary as the only code path where retrieval of artifact metadata happens was already on event processing functions in `ArtifactListener`
- Removes methods from `ArtifactSupplier` that were moved to `PublishedArtifact` properties


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
```diff
@@ -59,24 +58,12 @@ interface ArtifactSupplier<A : DeliveryArtifact, V : VersioningStrategy> : Spinn
    */
   fun getArtifactByVersion(artifact: DeliveryArtifact, version: String): PublishedArtifact?
 
-  /**
-   * Given a [PublishedArtifact] supported by this [ArtifactSupplier], return the full representation of
-   * a version string, if different from [PublishedArtifact.version].
-   */
-  fun getFullVersionString(artifact: PublishedArtifact): String = artifact.version
```
luispollo, line 66: We can just use the `version` property in `PublishedArtifact`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
```diff
@@ -59,24 +58,12 @@ interface ArtifactSupplier<A : DeliveryArtifact, V : VersioningStrategy> : Spinn
...
    * Given a [PublishedArtifact] supported by this [ArtifactSupplier], return the display name for the
    * artifact version, if different from [PublishedArtifact.version].
    */
   fun getVersionDisplayName(artifact: PublishedArtifact): String = artifact.version
 
-  /**
-   * Given a [PublishedArtifact] supported by this [ArtifactSupplier], return the [ArtifactStatus] for
-   * the artifact, if applicable.
-   */
-  fun getReleaseStatus(artifact: PublishedArtifact): ArtifactStatus? = null
```
luispollo, line 78: We can just use the `status` property in `PublishedArtifact`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
```diff
@@ -17,5 +19,40 @@ data class PublishedArtifact(
...
+      "createdAt" to createdAt
+    ),
+    gitMetadata = gitMetadata,
+    buildMetadata = buildMetadata
+  )
+
+  val status: ArtifactStatus? = metadata["releaseStatus"]?.toString()
+    ?.let { ArtifactStatus.valueOf(it) }
+
+  val createdAt = (metadata["createdAt"] as? Long)
```
luispollo, line 53: Will add a column to `ARTIFACT_VERSIONS` to store this in a follow-up PR.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -53,15 +54,15 @@ class ArtifactListener(
...
-          val version = artifactSupplier.getFullVersionString(artifact)
-          var status = artifactSupplier.getReleaseStatus(artifact)
-          log.info("Registering version {} ({}) of {} {}", version, status, artifact.name, artifact.type)
-          repository.storeArtifact(artifact.name, artifact.artifactType, version, status)
+          val enrichedArtifact = artifactSupplier.addMetadata(artifact.normalized())
+
+          log.info("Registering version {} (status={}) of {} artifact {}",
+            artifact.version, artifact.status, artifact.type, artifact.name)
+
+          repository.storeArtifactVersion(enrichedArtifact)
```
luispollo, line 62: Instead of doing one write to `ARTIFACT_VERSIONS` to store the basics here, and another as a reaction to the `ArtifactSaved` event below, consolidate in a single call to `storeArtifactVersion()`.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -77,17 +78,16 @@ class ArtifactListener(
...
+
       if (latestVersion != null) {
-        var status = artifactSupplier.getReleaseStatus(latestArtifact)
-        log.debug("Storing latest version {} (status={}) for registered artifact {}", latestVersion, status, artifact)
-        repository.storeArtifact(artifact.name, artifact.type, latestVersion, status)
-        // send an event when the artifact is stored, so we can fetch its metadata
-        publisher.publishEvent(ArtifactSaved(latestArtifact, status))
+        log.debug("Storing latest version {} (status={}) for registered artifact {}", latestVersion.version, latestVersion.status, artifact)
+        val enrichedArtifact = artifactSupplier.addMetadata(latestVersion.normalized())
+        repository.storeArtifactVersion(enrichedArtifact)
```
luispollo, line 90: Instead of doing one write to `ARTIFACT_VERSIONS` to store the basics here, and another as a reaction to the `ArtifactSaved` event just below, consolidate in a single call to `storeArtifactVersion()`.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -115,26 +115,24 @@ class ArtifactListener(
...
 
               if (hasNew) {
-                log.debug("$artifact has a missing version $latestVersion, persisting.")
-                val status = artifactSupplier.getReleaseStatus(latestArtifact)
-                repository.storeArtifact(artifact.name, artifact.type, latestVersion, status)
-                // send an event when the artifact is stored, so we can fetch its metadata
-                publisher.publishEvent(ArtifactSaved(latestArtifact, status))
+                log.debug("$artifact has a missing version ${latestVersion.version}, persisting.")
+                val enrichedArtifact = artifactSupplier.addMetadata(latestVersion.normalized())
+                repository.storeArtifactVersion(enrichedArtifact)
```
luispollo, line 135: Instead of doing one write to `ARTIFACT_VERSIONS` to store the basics here, and another as a reaction to the `ArtifactSaved` event just below, consolidate in a single call to `storeArtifactVersion()`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -37,18 +36,12 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
    * @return `true` if a new version is persisted, `false` if the specified version was already
    * known (in which case this method is a no-op).
    */
-  fun store(name: String, type: ArtifactType, version: String, status: ArtifactStatus?): Boolean
+  fun storeVersion(artifact: PublishedArtifact): Boolean
```
luispollo, line 39: We now use a model to store and retrieve published artifact versions (`PublishedArtifact`).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
```diff
@@ -64,39 +56,4 @@ class AdminService(
       }
     }
   }
-
-  /**
-   * updates old artifact version records with their metadata, if exists, by type [deb/docker/npm]
-   */
-  fun backfillOldArtifactVersions(type: String) {
```
luispollo, line 71: Not used anymore.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -306,51 +304,33 @@ class ApplicationService(
...
   fun getApplicationEventHistory(application: String, limit: Int) =
     repository.applicationEventHistory(application, limit)
 
-  private val ArtifactVersions.key: String
-    get() = "$type:$name"
-
   private fun ConstraintState.toConstraintSummary() =
     StatefulConstraintSummary(type, status, createdAt, judgedBy, judgedAt, comment, attributes)
-
-  private fun DeliveryArtifact.toSpinnakerArtifact(version: String): PublishedArtifact =
```
luispollo, line 348: Moved to `DeliveryArtifact` without a dependency on Jackson.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
```diff
@@ -52,8 +52,3 @@ data class ArtifactCheckTimedOut(
   val name: String,
   val deliveryConfigName: String?
 ) : TelemetryEvent()
-
-data class ArtifactSaved(
```
luispollo, line 56: No longer needed.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
```diff
@@ -61,15 +61,4 @@ class AdminController(
   ) {
     adminService.forceConstraintReevaluation(application, environment, type)
   }
-
-  //This endpoint will update old artifact version records with their metadata, if exists, by type [deb/docker/npm].
-  //Please note: this is an admin endpoint and is not intent to be used more than once per environment for now.
-  @PostMapping(
-    path = ["/artifacts/{type}/metadata/backfill"]
```
luispollo, line 68: No longer used.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
```diff
@@ -17,5 +19,40 @@ data class PublishedArtifact(
...
+    ?.let { Instant.ofEpochMilli(it) }
+
+  // FIXME: it's silly that we're prepending the artifact name for Debian only...
```
emjburns, line 56: agreed, this is an annoying inconsistency

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
```diff
@@ -17,5 +19,40 @@ data class PublishedArtifact(
...
+    ),
+    gitMetadata = gitMetadata,
+    buildMetadata = buildMetadata
+  )
+
+  val status: ArtifactStatus? = metadata["releaseStatus"]?.toString()
+    ?.let { ArtifactStatus.valueOf(it) }
+
+  val createdAt = (metadata["createdAt"] as? Long)
+    ?.let { Instant.ofEpochMilli(it) }
```
emjburns, line 54: should this be an instant, or the localdatetime format that we use in other tables?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
emjburns: What if the call to get metadata fails? Then we won't write the version of the artifact at all. It seems more reliable to first write the artifact, then add the metadata to it if it exists

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -77,17 +78,16 @@ class ArtifactListener(
 
     if (repository.artifactVersions(artifact).isEmpty()) {
       val artifactSupplier = artifactSuppliers.supporting(artifact.type)
-      val latestArtifact = runBlocking {
```
emjburns, line 80: I dislike the renaming here, because it's confusing. below we have `latestVersion.version`. Could you please revert this name switching?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -148,6 +146,21 @@ class ArtifactListener(
...
+  private fun ArtifactSupplier<*,*>.addMetadata(artifact: PublishedArtifact): PublishedArtifact {
+    val artifactMetadata = runBlocking {
+      try {
+        getArtifactMetadata(artifact)
+      } catch (ex: Exception) {
+        log.error("Could not fetch artifact metadata for name ${artifact.name} and version ${artifact.version}", ex)
+        null
+      }
+    }
+    return artifact.copy(gitMetadata = artifactMetadata?.gitMetadata, buildMetadata = artifactMetadata?.buildMetadata)
```
emjburns, line 161: we should only add the metadata if it's not null. otherwise, let's just return the original artifact

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
emjburns: Ah, I see, you've wrapped the get metadata in exception handling and we will save the original artifact. That seems fine.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -148,6 +146,21 @@ class ArtifactListener(
   private fun getLatestStoredVersion(artifact: DeliveryArtifact): String? =
     repository.artifactVersions(artifact).sortedWith(artifact.versioningStrategy.comparator).firstOrNull()
 
+  /**
+   * Returns a copy of the [PublishedArtifact] with the git and build metadata populated, if available.
+   */
+  private fun ArtifactSupplier<*,*>.addMetadata(artifact: PublishedArtifact): PublishedArtifact {
+    val artifactMetadata = runBlocking {
+      try {
+        getArtifactMetadata(artifact)
```
emjburns, line 155: Let's add retries around this, please

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
luispollo: Yep. There should be no change in behavior there. That's what we were doing in the other listener already.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
luispollo: We use `Instant` in all the models. `LocalDateTime` is only used in the JOOQ metamodels.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
luispollo: Ok.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
luispollo: I didn't think it was worth the extra `if` statement because the original object will also have those fields set to `null` by default...

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
luispollo: The retries are already there, further down into the call chain, in `ArtifactMetadataService`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
emjburns: Ah okay, it's only if there is a timestamp column? Just curious, because you mentioned adding it to a column in a future pr

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
emjburns: GREAT!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
luispollo: I can write a migration in a follow-up PR to fix this.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
emjburns: Are you sure those fields will be set to null?

emjburns (approved):

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -30,5 +30,15 @@ abstract class DeliveryArtifact {
...
   open val statuses: Set<ArtifactStatus> = emptySet()
+
+  fun toPublishedArtifact(version: String, status: ArtifactStatus? = null) =
+    PublishedArtifact(
+      name = name,
+      type = type,
+      reference = reference,
+      version = version,
+      metadata = mapOf("releaseStatus" to status)
+    ).normalized()
```
robfletcher, line 41: I'm catching up on recent artifact work. Can you explain to me like I'm 5 the difference between a `DeliveryArtifact` and a `PublishedArtifact`? Does the latter represent a specific instance of the former with a version and associated metadata? That seems to be the case, but I feel like `PublishedArtifact` is maybe not the best name. Neither is `DeliveryArtifact` for that matter, but one thing at a time.

robfletcher (approved):
This looks like a good simplification to me.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
luispollo: Yeah, if the value is not present in the `metadata` map (which I plan to fill from the column in a future PR), then it will just get set to `null`. As for using a `Long` here, it's because that's what we'll get from Rocket  (where we use this same class to represent the published artifacts). We can still store it in a `TIMESTAMP` column in the DB, though.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
luispollo: Yep, that's exactly it. `DeliveryArtifact` really just has some metadata (type, name, etc.) that allows us to find and sort the actual artifacts. I've been struggling with these names too. A couple of variants I thought of for `PublishedArtifact` were `PublishedArtifactVersion` or just `ArtifactVersion`. Does any of those sound better to you?

luispollo: @robfletcher I'm merging this as is, but let me know. I can change the name in a follow-up PR.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
gal-yardeni: why? I was going to refactor and reuse it in another way.

luispollo: As I mentioned in a related comment (https://github.com/spinnaker/keel/pull/1530#discussion_r493838264), my understanding was that we'd no longer use this endpoint. If we decide to reintroduce (parts of) this refactoring, I'll restore it.

PR #1532, opened 2020-09-23
feat(artifacts): Add artifact creation timestamp

luispollo:
Adds a `CREATED_AT` column to the `ARTIFACT_VERSIONS` table to store the artifact creation timestamp, as reported through artifact events via `POST /artifacts/events`.

The ultimate purpose of this field/column, alongside git branch information obtained from CI integration, is to replace the use of artifact versions for sorting. This puts us one little step closer. 🙂 

This PR also renames `PublishedArtifact` to `ArtifactVersion` as discussed here: https://github.com/spinnaker/keel/pull/1530#discussion_r493864519

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/ArtifactVersion.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/ArtifactVersion.kt
@@ -9,7 +9,7 @@ import java.time.Instant
  * One notable difference from the kork counterpart is that this class enforces non-nullability of a few
  * key fields without which it doesn't make sense for an artifact to exist in Managed Delivery terms.
  */
-data class PublishedArtifact(
+data class ArtifactVersion(
   val name: String,
   val type: String,
   val reference: String,
@@ -59,5 +59,6 @@ data class PublishedArtifact(
       }
     }
 
+  // FIXME: it's silly that we're prepending the artifact name for Debian only...
   fun normalized() = copy(version = if (type == DEBIAN && !version.startsWith(name)) "$name-$version" else version)
 }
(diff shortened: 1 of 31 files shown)


emjburns (approved):

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/ArtifactVersion.kt
```diff
@@ -9,7 +9,7 @@ import java.time.Instant
  * One notable difference from the kork counterpart is that this class enforces non-nullability of a few
  * key fields without which it doesn't make sense for an artifact to exist in Managed Delivery terms.
  */
-data class PublishedArtifact(
+data class ArtifactVersion(
```
luispollo, line 12: @robfletcher For you. 🙂

keel-sql/src/main/resources/db/changelog/20200923-add-created-at-to-artifact-versions.yml
```diff
@@ -0,0 +1,14 @@
...
+      author: lpollo
+      changes:
+        - addColumn:
+            tableName: artifact_versions
+            columns:
+              - column:
+                  name: created_at
+                  type: timestamp(6)
+                  constraints:
+                    nullable: true
```
robfletcher, line 13: Is this nullable because we have old data and no way to back-fill the value, or because it comes from metadata after the initial record is written?

luispollo: The former. We could potentially infer the value from `buildMetadata.completedAt`, but I'm guessing we'll miss some. Could we use epoch time as a default? We could also write a component to fetch the data from Rocket and make the column not nullable in a follow-up PR.

PR #1536, opened 2020-09-24
feat(artifacts): Published artifact handling refactoring and timestamps

luispollo:
This my second take at #1530 and #1532. This PR includes the following changes:
- Renaming `PublishedArtifact` to `ArtifactVersion` based on discussion with @robfletcher. 
- Promoting the usage of this model class to encapsulate the properties of artifact versions stored in the `artifact_versions` table through the `storeArtifactVersion()` and `getArtifactVersion()` repository models. The rationale for this is that we did not have a model in place to represent those table rows, which I realized when I tried to add a `createdAt` property for the artifact versions.
- Once this model and methods were in place, it became an obvious "home" for the git and build metadata also stored in that table, and so I removed the associated repository methods in favor of the methods above (which already return the metadata as part of the `ArtifactVersion` instance).
- Replaces the usage of the `ArtifactSaved` event and corresponding `ArtifactStoredListener` with code in `ArtifactListener` to fetch and store the metadata. The rationale for this change is that I noticed the only code paths where the event was fired were in `ArtifactListener`, which is itself already an event handler processing artifact events asynchronously, so there was no benefit in adding another event, and it was also adding unnecessary complexity. The refactored methods in `ArtifactListener` are pretty straightforward to read after the change, IMHO.
- Finally, introduces a field and matching column to store artifact creation timestamp in the `ArtifactVersion` class and `artifact_versions` table, respectively.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/ArtifactVersion.kt
```diff
@@ -0,0 +1,64 @@
...
+import java.time.Instant
+
+/**
+ * An immutable data class that represents a published software artifact in the Spinnaker ecosystem.
+ *
+ * This class mirrors [com.netflix.spinnaker.kork.artifacts.model.Artifact], but without all the Jackson baggage.
+ * One notable difference from the kork counterpart is that this class enforces non-nullability of a few
+ * key fields without which it doesn't make sense for an artifact to exist in Managed Delivery terms.
+ */
+data class ArtifactVersion(
```
luispollo, line 12: This is just the renamed `PublishedArtifact` class with a few additions. Not sure why GitHub didn't detect that this time around.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -30,5 +31,18 @@ abstract class DeliveryArtifact {
   abstract val reference: String // friendly reference to use within a delivery config
   abstract val deliveryConfigName: String? // the delivery config this artifact is a part of
   open val statuses: Set<ArtifactStatus> = emptySet()
+
+  fun toArtifactVersion(version: String, status: ArtifactStatus? = null, createdAt: Instant? = null) =
```
luispollo, line 35: New implementation of this convenience function that no longer depends on Jackson.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -53,17 +54,19 @@ class ArtifactListener(
...
-          val version = artifactSupplier.getFullVersionString(artifact)
-          var status = artifactSupplier.getReleaseStatus(artifact)
-          log.info("Registering version {} ({}) of {} {}", version, status, artifact.name, artifact.type)
-          repository.storeArtifact(artifact.name, artifact.artifactType, version, status)
+          val enrichedArtifact = artifactSupplier.addMetadata(artifact.normalized())
+
+          log.info("Registering version {} (status={}) of {} artifact {}",
+            artifact.version, artifact.status, artifact.type, artifact.name)
+
+          repository.storeArtifactVersion(enrichedArtifact)
```
luispollo, line 62: This method now persists both the "basic" properties of the artifact and the metadata in one database write.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -53,17 +54,19 @@ class ArtifactListener(
       .forEach { artifact ->
         if (repository.isRegistered(artifact.name, artifact.artifactType)) {
           val artifactSupplier = artifactSuppliers.supporting(artifact.artifactType)
-          val version = artifactSupplier.getFullVersionString(artifact)
-          var status = artifactSupplier.getReleaseStatus(artifact)
-          log.info("Registering version {} ({}) of {} {}", version, status, artifact.name, artifact.type)
-          repository.storeArtifact(artifact.name, artifact.artifactType, version, status)
+          val enrichedArtifact = artifactSupplier.addMetadata(artifact.normalized())
```
luispollo, line 57: This method fetches metadata from the external CI service.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -57,25 +55,12 @@ class DebianArtifactSupplier(
         }
     }
 
-  override fun getArtifactByVersion(artifact: DeliveryArtifact, version: String): PublishedArtifact? =
+  override fun getArtifactByVersion(artifact: DeliveryArtifact, version: String): ArtifactVersion? =
     runWithIoContext {
       artifactService.getArtifact(artifact.name, version.removePrefix("${artifact.name}-"), DEBIAN)
     }
 
-  override fun getFullVersionString(artifact: PublishedArtifact): String =
```
luispollo, line 65: These methods are now getters in `ArtifactVersion` and no longer need to be replicated in different `ArtifactSupplier` implementations.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -75,6 +57,23 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
     type: ArtifactType
   ): List<String>
 
+  /**
```
luispollo, line 60: I moved these down because they were in-between some `DeliveryArtifact`-related methods, but it makes the diff a bit harder to read. If you prefer, I can move them back where they were.

keel-sql/src/main/resources/db/changelog/20200923-add-created-at-to-artifact-versions.yml
```diff
@@ -0,0 +1,14 @@
...
+      author: lpollo
+      changes:
+        - addColumn:
+            tableName: artifact_versions
+            columns:
+              - column:
+                  name: created_at
+                  type: timestamp(6)
+                  constraints:
+                    nullable: true
```
luispollo, line 13: This is nullable because we'll start with a bunch of records that don't have the timestamp when we migrate, and also there's no sensible default we can use. We can look into populating this after the fact with a batch job or something.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
```diff
@@ -66,37 +65,38 @@ class AdminService(
...
     deliveryArtifacts.forEach { artifact ->
       val versions = repository.artifactVersions(artifact)
       versions.forEach { version ->
+        sleep(5000) // sleep a little in-between versions to cut the instance where this runs some slack
+        log.debug("Evaluating $artifact version $version as candidate to back-fill metadata")
         val status = repository.getReleaseStatus(artifact, version)
-        //don't update if metadata is already exists
-        if (repository.getArtifactBuildMetadata(artifact.name, type, version, status) == null) {
-            runBlocking {
-              launch {
```
luispollo, line 84: I removed the `launch` here as it wasn't doing anything inside this `runBlocking`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
```diff
@@ -66,37 +65,38 @@ class AdminService(
...
-    //1. get all registered artifacts
+    log.debug("Starting to back-fill old artifacts versions with artifact metadata...")
+    // 1. get all registered artifacts
     val deliveryArtifacts = repository.getAllArtifacts(type)
-    //2. for each artifact, fetch all versions
+    // 2. for each artifact, fetch all versions
     deliveryArtifacts.forEach { artifact ->
       val versions = repository.artifactVersions(artifact)
       versions.forEach { version ->
+        sleep(5000) // sleep a little in-between versions to cut the instance where this runs some slack
```
luispollo, line 79: Added a little delay here to allow some breathing room for the instance where this runs...

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
gal-yardeni: fwiw, I wanted to implement it like this from the first place, but was concerned with external CI service failures :)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -313,20 +312,14 @@ class CombinedRepository(
...
     artifactRepository.getAll(type)
 
-  override fun storeArtifact(name: String, type: ArtifactType, version: String, status: ArtifactStatus?): Boolean =
-    artifactRepository.store(name, type, version, status)
+  override fun storeArtifactVersion(artifactVersion: PublishedArtifact): Boolean =
+    artifactRepository.storeArtifactVersion(artifactVersion)
 
-  override fun storeArtifact(artifact: DeliveryArtifact, version: String, status: ArtifactStatus?): Boolean =
-    artifactRepository.store(artifact, version, status)
+  override fun getArtifactVersion(name: String, type: ArtifactType, version: String, status: ArtifactStatus?): PublishedArtifact? =
```
gal-yardeni, line 318: the name `getArtifactVersion` is a bit confusing to me, as it's indicated we are getting the artifact version and not `PublishedArtifact` object -- should we maybe remane it to `getArtifact` or `getArtifactByxxx`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
```diff
@@ -144,15 +143,11 @@ interface KeelRepository : KeelReadOnlyRepository {
 
   fun getAllArtifacts(type: ArtifactType? = null): List<DeliveryArtifact>
 
-  fun storeArtifact(name: String, type: ArtifactType, version: String, status: ArtifactStatus?): Boolean
+  fun storeArtifactVersion(artifactVersion: PublishedArtifact): Boolean
 
-  fun storeArtifact(artifact: DeliveryArtifact, version: String, status: ArtifactStatus?): Boolean
+  fun getArtifactVersion(name: String, type: ArtifactType, version: String, status: ArtifactStatus?): PublishedArtifact?
```
gal-yardeni, line 148: same comment as above

gal-yardeni (approved):
LGTM, thanks for simplifing the implementation. I have some concerns with some method names, but I don't have any better suggestions 🤷

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
luispollo: Yeah, I know... We've been having a conversation about the naming of the classes used to represent artifacts. I think the last proposal we landed on would be to rename `DeliveryArtifact` to `ArtifactSpec` and `PublishedArtifact` to just `Artifact`, when it would then make total sense to rename these methods. I'd like to tackle that in a separate PR though, if that's OK.

luispollo: Cc: @robfletcher

luispollo: @gal-yardeni Update: I went ahead and renamed the methods per your suggestion.

PR #1538, opened 2020-09-24
Disambiguate ingress rule types in schema

robfletcher:
Without specifying `additionalProperties: false` some ingress rules can look like other types. That caused errors for users because the schema couldn't distinguish between the ingress rule types. I've made `additionalProperties: false` be the default for objects except where it's creating parameterized types.

Also, fixed values, such as singleton objects were being rendered as an `enum` with a single value when they can just be a `const`.

Addresses one of the things in #1535

### DIFF
--- a/keel-web/src/test/resources/examples/security-group-with-cidr-rule-example.yml
+++ b/keel-web/src/test/resources/examples/security-group-with-cidr-rule-example.yml
@@ -1,0 +1,25 @@
+---
+application: fnord
+serviceAccount: delivery-engineering@netflix.com
+environments:
+- name: test
+  resources:
+  - kind: ec2/security-group@v1
+    spec:
+      moniker:
+        app: keeldemo
+        stack: example
+        detail: ec2v1
+      locations:
+        account: test
+        vpc: vpc0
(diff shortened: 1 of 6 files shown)


keel-web/src/test/resources/examples/security-group-with-cidr-rule-example.yml
```diff
@@ -0,0 +1,25 @@
...
+        regions:
+        - name: us-west-2
+        - name: us-east-1
+      description: Managed Security Group for keeldemo example
+      inboundRules:
+      - protocol: tcp
+        blockRange: "2600:1f1f:4b2:3a00::/56"
+        portRange:
+          startPort: 7001
+          endPort: 7001
```
robfletcher, line 25: This is an example of the problem. Although `ReferenceRule` does not have a `blockRange` property, because it also did not forbid additional properties this data could match either `ReferenceRule` or `CIDRRule` schemas. If `ReferenceRule` specifies `additionalProperties: false` the ambiguity is removed.

luispollo (approved):
LGTM.

PR #1539, opened 2020-09-25
should be able to specify additionalProperties: false on root

robfletcher:
D'oh, missed this in #1538 

Also added some more tests.

### DIFF
--- a/keel-schema-generator/src/test/kotlin/com/netflix/spinnaker/keel/schema/GeneratorTests.kt
+++ b/keel-schema-generator/src/test/kotlin/com/netflix/spinnaker/keel/schema/GeneratorTests.kt
@@ -27,14 +27,11 @@ import strikt.assertions.containsExactlyInAnyOrder
 import strikt.assertions.containsKey
 import strikt.assertions.containsKeys
 import strikt.assertions.doesNotContain
-import strikt.assertions.filterIsInstance
-import strikt.assertions.first
 import strikt.assertions.get
-import strikt.assertions.hasEntry
 import strikt.assertions.hasSize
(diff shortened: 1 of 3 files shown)


keel-schema-generator/src/test/kotlin/com/netflix/spinnaker/keel/schema/GeneratorTests.kt
```diff
@@ -618,43 +639,36 @@ internal class GeneratorTests {
...
-        .map { (discriminatorValue, type) ->
-          dynamicTest("the discriminator property of ${type.simpleName} is an enum with the single value \"$discriminatorValue\"") {
-            expectThat(schema.`$defs`[type.simpleName])
-              .isA<ObjectSchema>()
-              .get { properties[Wrapper<*>::type.name] }
-              .isA<EnumSchema>()
-              .get(EnumSchema::enum)
-              .containsExactly(discriminatorValue)
-          }
+        .flatMap { (discriminatorValue, type) ->
```
robfletcher, line 647: oops. I had `map` instead of `flatMap` so none of the contained tests were running

keel-schema-generator/src/test/kotlin/com/netflix/spinnaker/keel/schema/GeneratorTests.kt
```diff
@@ -618,43 +639,36 @@ internal class GeneratorTests {
     }
 
     @TestFactory
-    fun subTypeDiscriminatorConst() =
```
robfletcher, line 621: Removed as this is just a dupe of one of the tests below

PR #1542, opened 2020-09-25
fix(artifacts): Fix handling of upper-case artifact types

luispollo:
I forgot that artifact events coming from echo have upper-case artifact type strings... 😛

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
@@ -61,6 +61,9 @@ data class PublishedArtifact(
       }
     }
 
-  // FIXME: it's silly that we're prepending the artifact name for Debian only...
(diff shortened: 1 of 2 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
```diff
@@ -61,6 +61,9 @@ data class PublishedArtifact(
       }
     }
 
-  // FIXME: it's silly that we're prepending the artifact name for Debian only...
-  fun normalized() = copy(version = if (type == DEBIAN && !version.startsWith(name)) "$name-$version" else version)
+  fun normalized() = copy(
+    type = type.toLowerCase(),
+    // FIXME: it's silly that we're prepending the artifact name for Debian only...
+    version = if (type.toLowerCase() == DEBIAN && !version.startsWith(name)) "$name-$version" else version
```
lorin, line 67: Looks like you're calling `type.toLowerCase()` twice.

lorin (approved):
LGTM

PR #1540, opened 2020-09-25
Additional indices to help environment check performance

robfletcher:

PR #1544, opened 2020-09-28
Fix DB index size

robfletcher:
PR #1540 adds an index composed of 4 `varchar(255)` columns. Therefore the size of the index you'd think would be 1020 bytes (4 * 255). The max index size in MySQL InnoDB tables is 3072 bytes. However, when using utf8mb4 character data the columns use 4 bytes per character so the index is actually 4080 bytes -- larger than the 3072 byte limit.

This was exacerbated by the fact that the dockerized mySQL database we use to run the `jooqGenerate` task in Gradle was not set up to use utf8 so the problem was not caught until deployment time.

### DIFF
--- a/keel-sql/keel-sql.gradle.kts
+++ b/keel-sql/keel-sql.gradle.kts
@@ -1,6 +1,5 @@
 import com.diffplug.gradle.spotless.SpotlessExtension
 import org.liquibase.gradle.LiquibaseTask
-import java.lang.Thread.sleep
 
 val buildingInDocker = project.properties["buildingInDocker"]?.toString().let { it == "true" }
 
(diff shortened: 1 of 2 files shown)


keel-sql/keel-sql.gradle.kts
```diff
@@ -83,7 +82,7 @@ tasks.getByName<LiquibaseTask>("liquibaseUpdate") {
         commandLine("sh", "-c", "docker stop mysqlJooq >/dev/null 2>&1 || true")
       }
       exec {
-        commandLine("sh", "-c", "docker run --name mysqlJooq --health-cmd='mysqladmin ping -s' -d --rm -e MYSQL_ROOT_PASSWORD=sa -e MYSQL_DATABASE=keel -p 6603:3306 mysql:5.7 >/dev/null 2>&1; while STATUS=\$(docker inspect --format \"{{.State.Health.Status}}\" mysqlJooq); [ \$STATUS != \"healthy\" ]; do if [ \$STATUS = \"unhealthy\" ]; then echo \"Docker failed to start\"; exit -1; fi; sleep 1; done")
+        commandLine("sh", "-c", "docker run --name mysqlJooq --health-cmd='mysqladmin ping -s' -d --rm -e MYSQL_ROOT_PASSWORD=sa -e MYSQL_DATABASE=keel -p 6603:3306 mysql:5.7 --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci >/dev/null 2>&1; while STATUS=\$(docker inspect --format \"{{.State.Health.Status}}\" mysqlJooq); [ \$STATUS != \"healthy\" ]; do if [ \$STATUS = \"unhealthy\" ]; then echo \"Docker failed to start\"; exit -1; fi; sleep 1; done")
```
robfletcher, line 85: Make the dockerized DB use utf8

luispollo: Good find!

luispollo:
> However, when using utf8mb4 character data the columns use 4 bytes per character so the index is actually 4080 bytes -- larger than the 3072 byte limit.

So don't we have to fix the index?

robfletcher:
> > However, when using utf8mb4 character data the columns use 4 bytes per character so the index is actually 4080 bytes -- larger than the 3072 byte limit.
> 
> So don't we have to fix the index?

Yes we do, it's too big.

keel-sql/src/main/resources/db/changelog/20200925-environment-related-indices.yml
```diff
@@ -11,8 +11,6 @@ databaseChangeLog:
             name: name
         - column:
             name: type
-        - column:
-            name: version
```
robfletcher, line 15: Will be good to find a solution to this, but since this is a new query and the original PR was designed to tweak performance of existing queries we can live without it for now

PR #1546, opened 2020-09-29
fix(artifacts): Fix summaries for same artifact with different release statuses

luispollo:
We got a bug report for a scenario where the user had a delivery config with two artifacts pointing to the same Debian package but with different release statuses. The `/application` API would return an incorrect artifact summary (with empty environments) for one of the artifacts. It turns out we were looking up the corresponding environment summary by artifact type and name only (as opposed to by reference), which was causing results not to be found.

I recommend viewing the diff with "ignore whitespace" enabled.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
@@ -34,7 +34,7 @@ data class EnvironmentSummary(
   }
 
   fun getArtifactPromotionStatus(artifact: DeliveryArtifact, version: String) =
-    artifacts.find { it.name == artifact.name && it.type == artifact.type }
+    artifacts.find { it.reference == artifact.reference && it.type == artifact.type }
       ?.let {
         when (version) {
           it.versions.current -> CURRENT
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -23,6 +23,7 @@ import com.netflix.spinnaker.keel.core.api.DependsOnConstraint
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactPin
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVeto
 import com.netflix.spinnaker.keel.core.api.EnvironmentSummary
+import com.netflix.spinnaker.keel.core.api.PromotionStatus
 import com.netflix.spinnaker.keel.core.api.PromotionStatus.PENDING
 import com.netflix.spinnaker.keel.core.api.PromotionStatus.SKIPPED
 import com.netflix.spinnaker.keel.core.api.ResourceArtifactSummary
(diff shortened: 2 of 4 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/EnvironmentSummary.kt
```diff
@@ -34,7 +34,7 @@ data class EnvironmentSummary(
   }
 
   fun getArtifactPromotionStatus(artifact: DeliveryArtifact, version: String) =
-    artifacts.find { it.name == artifact.name && it.type == artifact.type }
+    artifacts.find { it.reference == artifact.reference && it.type == artifact.type }
```
luispollo, line 37: This was the fix.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -183,70 +184,78 @@ class ApplicationService(
...
-                  environmentName = environmentSummary.name,
-                  artifactReference = artifact.reference,
-                  version = version
-                )
-                if (potentialSummary == null || potentialSummary.state == "pending") {
-                  ArtifactSummaryInEnvironment(
-                    environment = environmentSummary.name,
-                    version = version,
-                    state = status.name.toLowerCase()
+          environmentSummary.getArtifactPromotionStatus(artifact, version)
```
luispollo, line 193: The logic in this file is unchanged. I just extracted some of it to a new function to make the loops easier to reason about.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -44,6 +45,7 @@ import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
 
 @RestController
+@Timed
```
luispollo, line 48: Unrelated, will revert.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -786,12 +786,12 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
       }
 
       test("retrieves successfully") {
-        val artifactVersion = subject.getArtifactInstance(artifact1.name, artifact1.type, version1, SNAPSHOT)!!
```
emjburns, line 789: I don't think renaming this adds a ton of value, and it makes the PR longer / a bit harder to review. You don't have to revert this, but please keep this in mind going forward. What I typically do is change names if I'm trying to figure something out, and then when I figure it out change them back unless I think it's very miss-named. This is more of a style thing, but I'm commenting because I've seen you do this in several recent PRs.

luispollo: Yeah, I know... I kinda regretted it after making the change. I just have a hard time looking past inconsistent naming like this. But I could totally have made a separate PR. Happy to undo these if you like.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -183,70 +184,78 @@ class ApplicationService(
...
-              artEnvSummary = addStatelessConstraintSummaries(artEnvSummary, deliveryConfig, environment, version, artifact)
-              artifactSummariesInEnvironments.add(artEnvSummary)
-            }
-          }
         }
-        return@map versionToSummary(artifact, version, artifactSummariesInEnvironments.toSet())
+
+        buildArtifactVersionSummary(artifact, version, artifactSummariesInEnvironments)
       }
-      return@map ArtifactSummary(
```
emjburns, line 234: I personally think the `return@map` makes it more clear, because I dislike having the last line returned by default for anything more than ~2 lines. It just makes it easier to scan and see what's going on

emjburns (approved):

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
luispollo: Reverted. I'll open a separate one to rename

emjburns: I'd just like you to weigh the cost of renaming the variables in every file which you think they are a little bit wrong. It's a tax on the people working on the codebase in general. Sometimes it's worth it, I agree! But I think sometimes it's smoother (when working on a shared codebase) to live with a little bit of idiosyncrasy. Up to you here on what you'd like to do.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: The compiler was issuing warnings for both statements because there are two `.map` calls. That's why I decided to refactor so that the loops are much more concise and fit in a single screen. Hope that's OK.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
luispollo: I hear you. I try to keep that in mind, but I recognize I'm a bit obsessed with detail. Will try and tune that down going forward (or at least ask first if it's worth making the change).

PR #1548, opened 2020-09-29
fix(artifacts): Increase number/duration of retries for artifact metadata

luispollo:
Doubles the number of retries for artifact metadata and spaces them out over a longer period to improve our chances. Also logs retries dues to empty results from Igor/Rocket.

### DIFF
--- a/keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
+++ b/keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
@@ -95,10 +95,14 @@ class ArtifactMetadataService(
     val retry = Retry.of(
       "get artifact metadata",
       RetryConfig.custom<List<Build>?>()
-        .maxAttempts(10)
-        .waitDuration(Duration.ofMillis(1000))
-        .retryOnResult{
-          result -> result.isNullOrEmpty()
+        // retry 20 times over a total of 90 seconds
+        .maxAttempts(20)
+        .waitDuration(Duration.ofMillis(4500))
+        .retryOnResult{ result ->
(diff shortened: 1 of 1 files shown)


lorin (approved):
Change looks good to me, but could you say a little more about the motivation for this change? What's the issue that the retries are working around?

luispollo:
> Change looks good to me, but could you say a little more about the motivation for this change? What's the issue that the retries are working around?

Yes, sorry, I should have included more context in the PR description. We currently attempt to retrieve metadata only once, when a new version of an artifact is detected, which is why we have retries in this code path. However, we've found that if there's ever anything that causes the downstream CI system (Rocket) to take longer to index the build metadata, or become unavailable for any reason, for longer than we're retrying, than we end up with missing metadata. This change is not a fail-proof way of ensuring we do, but is an attempt at improving the odds. If we still fail, @gal-yardeni is working on a scheduled batch job to backfill the missing metadata (see #1549).

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/ArtifactMetadataService.kt
```diff
@@ -95,18 +95,23 @@ class ArtifactMetadataService(
     val retry = Retry.of(
       "get artifact metadata",
       RetryConfig.custom<List<Build>?>()
-        .maxAttempts(10)
-        .waitDuration(Duration.ofMillis(1000))
-        .retryOnResult{
-          result -> result.isNullOrEmpty()
+        // retry 20 times over a total of 90 seconds
```
luispollo, line 98: @emjburns Does a minute and a half worth of additional waiting sound acceptable to you (worst case scenario)?

gal-yardeni (approved):
LGTM!

Issue #1149, opened 2020-05-11
Artifact Versions: should we care about them? Let's think about it


emjburns:
Right now, we only store artifact versions in our database if there is a config that cares about them. Seems to make sense, right? Well, lets consider this excellent case which @gal-yardeni discovered. 

Here is your environment view, take a look! All the dark grey bars have status "Skipped", and no other info.
![Screen Shot 2020-05-11 at 11 51 31 AM](https://user-images.githubusercontent.com/8454927/81600016-d4e20a00-937d-11ea-91dd-867daa547f2b.png)
Super sleuths will also notice that something else is going on with all the red and the version stuck in deploying. Ignore that! Not relevant for this issue. It involves two dueling orcas, and lots of errors.

### Explanation of how things got to this state:

Say you have a delivery config that cares about a docker image. You have it in `keel` while versions 6-13 are created. We don't load versions before 6, because we don't particularly care about the long history of an artifact. That seems reasonable.

Then, you delete your config from main. Maybe on purpose! Maybe on accident! 

Days pass.

Meanwhile, versions 14, 15, and 16 of the docker image were published.

Then, you decide you want your app to be managed again. When you submit the config, we fetch the latest version of the artifact (16) and start deploying it. It went fine.

❓ Is the artifact list how you would expect? ❓ 

_It's not what I would expect. I'd expect us to show the middle versions (14 and 15)._

Two possible solutions for this: 
- we store all new versions of the artifact if _either_ an artifact is registered _or_ we have already recorded versions of the artifact. ❗ side effect alert ❗ for versions where there was no active delivery config we will have no data on the versions, and they'll show up as skipped w/o more info.
- when you delete your delivery config we also delete all the artifact versions we've stored if you're the last person to care about the version ❗ side effect alert ❗ if you immediately re-submit the delivery config you'll essentially remove the ability to pin backwards more than 1 versions, because we no longer know about those versions

Thoughts?

erikmunson:
oh yah this makes sense. I worry a lot about deleting historical version data – that's some of the most important data we have. I would actually feel way more comfortable blowing away history events vs. dropping version data.

> we store all new versions of the artifact if either an artifact is registered or we have already recorded versions of the artifact. 

I have no idea how this works... is there anything to consider in terms of query performance if we need to look for any previously recorded versions going all the way back in time (i.e. near-unbounded queries)? Maybe I'm misunderstanding what you're suggesting in terms of logic.

> they'll show up as skipped w/o more info.

[...]

Issue #1545, opened 2020-09-28
Retries to get artifact metadata are not sufficient


luispollo:
We recently received a support request at Netflix where a new version of an artifact which had metadata included for all prior versions suddenly was missing it. Upon investigating, my theory is that we failed to retrieve non-empty metadata from Igor after the 10 retries. I'll be adding logs to catch/confirm that case, but we'll need to think of a more fail-proof way of getting the metadata. This was for a Docker artifact, which means Rocket wasn't involved.

The next version had the metadata, which makes it an intermittent issue, I'm guessing somewhere between Igor and Clouddriver/Titus.

luispollo:
Cc: @gal-yardeni

PR #1549, opened 2020-09-29
chore(artifactVersion): scheduled script to backfill metadata, limit versions fetching to 30

gal-yardeni:
closes https://github.com/spinnaker/keel/issues/1149 and https://github.com/spinnaker/keel/issues/1545

- Adding a scheduled script to run every 6 hours (we can debate about the specific time :) )
- Limiting artifact version fetching to the last 30 versions (also debatable...).


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/persistence/KeelReadOnlyRepository.kt
```diff
@@ -49,6 +49,8 @@ interface KeelReadOnlyRepository {
 
   fun isRegistered(name: String, type: ArtifactType): Boolean
 
+  fun artifactVersions(artifact: DeliveryArtifact, limit: Int): List<String>
```
luispollo, line 52: Why not add the limit to the existing function below, with a default value?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -327,6 +327,9 @@ class CombinedRepository(
   override fun artifactVersions(artifact: DeliveryArtifact): List<String> =
     artifactRepository.versions(artifact)
 
+  override fun artifactVersions(artifact: DeliveryArtifact, limit: Int): List<String> =
```
luispollo, line 330: Same comment on adding the `limit` parameter to the existing function instead of defining a new one.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
```diff
@@ -63,7 +67,18 @@ class AdminService(
   }
 
   /**
-   * Updates artifact version records with the corresponding metadata, if available, by type [deb/docker/npm]
+   * For each artifact type [deb/docker/npm], run back-fill script to in case there's some missing data
+   */
+  @Scheduled(fixedDelayString = "\${keel.artifact-metadata-backfill.frequency:PT6H}")
+  fun updateBackFillArtifactMetadataForAllTypes(){
```
luispollo, line 73: Just a suggestion on naming:
```suggestion
  fun periodicallyBackFillArtifactMetadata(){
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -237,9 +237,9 @@ class SqlArtifactRepository(
         .also { versions ->
           // FIXME: remove special handling for Docker
           return if (artifact is DockerArtifact) {
-            filterDockerVersions(artifact, versions)
+            filterDockerVersions(artifact, versions, limit)
           } else {
-            versions
+            versions.subList(0, Math.min(versions.size, limit))
```
luispollo, line 242: If we ever convert _all_ apps to using branch+timestamp for sorting, then this limit could be imposed in the SQL query, which would be much more efficient. Might be worth adding a comment to that effect here.

keel-web/config/keel.yml
```diff
@@ -6,6 +6,8 @@ keel:
     minAgeDuration: 10s
   artifact-refresh:
     frequency: PT1M
+  metadata-backfill:
```
luispollo, line 9:
```suggestion
  artifact-metadata-backfill:
```

erikmunson:
💯 Nice. Not sure how to feel about the 6 hour timing — I would say if we continue to see issues with missing data after @luispollo's PR to increase the retries, we might want to consider making the timing on this more aggressive to catch when things slip through.

luispollo (commented):
This definitely won't hurt, but as discussed in chat, I think a more aggressive approach like attempting to backfill missing metadata on-demand in the environment promotion loop will be needed for sorting based on branch to work.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -297,6 +297,11 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
           test("querying with only release returns correct versions") {
             expectThat(subject.versions(artifact2)).containsExactly(version5, version4)
           }
+
+          test("querying for limit returns limit") {
+            val artifactWithAll = artifact1.copy(statuses = emptySet())
+            expectThat(subject.versions(artifactWithAll, 2)).containsExactly(version5, version4)
+          }
```
robfletcher, line 304: Are there more versions than that in the database? It looks like the test right above that does not use the limit also gets 2 results back.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
robfletcher, line 242: Do we expect any significant performance benefit if we don't actually limit the query? I get that it would be difficult to do that given the sort method.

robfletcher: In theory we could always assign a sort weight and persist that in its own column, making the query itself sortable.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/persistence/KeelReadOnlyRepository.kt
gal-yardeni: tried it and it asked me to change all the other places where this function is being called... did I miss something? 😅

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: thanks @luispollo , good point - i'll add a comment for it. it's much better using `orderBy` or something similar and uses native SQL functions :)

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
gal-yardeni: this test is calling `persist`, which is storing 5 versions for `artifact1` and we expecting to get only 2 back

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: I'm not sure how to measure the performance here.. I was struggling with the best way to limit that list, we can do it at the API level as well if you think it's better?

robfletcher: I think it's good to have the limit here because it makes subsequent optimization easier. I really hope we can get to a place where we could use a limit on the query itself. In this case we're reading all the entries and hydrating objects, then discarding all but `limit`. I'd expect some kind of improvement just based on the size of the response from the API endpoint, but limiting the query would be optimal.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/persistence/KeelReadOnlyRepository.kt
luispollo: Did you assign a default value for the parameter?

keel-web/config/keel.yml
lorin: Rather than having two configuration options with `artifact-` prefixes, I suggest doing `artifact.` instead, that way you could do them as subsections in the yaml if desired.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
lorin: Just piping in here to agree with @robfletcher. We really do need to get in the habit of limiting the size all of our queries at the SQL level or we're going to see performance degradation over time.

The sooner we are able to implement version sorting in the SQL query itself, the better off we're going to be.

lorin (approved):
LGTM

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/persistence/KeelReadOnlyRepository.kt
emjburns: Ya, you'll need to assign a default value to the parameter ( `..., limit: Int = 30)`)

gal-yardeni: 🤦

PR #1553, opened 2020-09-30
refactor(ec2): extract method

lorin:
Extract common code into a method. This is code that generates the payload for the orca job that creates new server groups.

No (intentional)) behavior changes in this PR.

This is prep work for adding support for different [EBS volume types](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html). That future work should only require modifying the new function.

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -207,11 +207,11 @@ class ClusterHandler(
         }
         when {
           diff.shouldDeployAndModifyScalingPolicies() -> {
-            stages.add(diff.createServerGroupJob(refId) + resource.spec.deployWith.toOrcaJobProperties("Amazon"))
+            stages.add(createServerGroupOrcaJob(diff, refId, resource))
             refId++
             stages.addAll(diff.modifyScalingPolicyJob(refId))
           }
-          else -> stages.add(diff.createServerGroupJob(refId) + resource.spec.deployWith.toOrcaJobProperties("Amazon"))
+          else -> stages.add(createServerGroupOrcaJob(diff, refId, resource))
         }
 
         if (stages.isEmpty()) {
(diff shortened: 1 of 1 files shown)


robfletcher (approved):

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -330,6 +329,9 @@ class ClusterHandler(
       return@coroutineScope tasks
     }
 
+  private fun createServerGroupOrcaJob(diff: ResourceDiff<ServerGroup>, refId: Int, resource: Resource<ClusterSpec>): Map<String, Any?> =
+    diff.createServerGroupJob(refId) + resource.spec.deployWith.toOrcaJobProperties("Amazon")
```
luispollo, line 333: Looks like there was already an extension function defined on the `ResourceDiff` object with almost the same purpose. Wouldn't it make sense to just add the additional parameters to that method instead of adding another level of indirection?

lorin: @luispollo My concern with adding to the existing function is that I felt like it would violate the "does one thing" aspect of it. But I did end up changing the additional function so it's also an extension function.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -1172,7 +1174,6 @@ class ClusterHandler(
     )
 
     private const val REGION_PLACEHOLDER = "{{region}}"
-    private const val REGION_PATTERN = "([a-z]{2}(-gov)?)-([a-z]+)-\\d"
```
lorin, line 1175: This was unused so I zapped it.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -375,7 +374,7 @@ class ClusterHandler(
       .toSet()
 
     // let's assume that the largest server group is the most important and should be the base
-    val base = serverGroups.values.maxBy { it.capacity.desired ?: it.capacity.max }
+    val base = serverGroups.values.maxByOrNull { it.capacity.desired ?: it.capacity.max }
```
lorin, line 377: `maxBy` is deprecated, so I changed it to `maxByOrNull`.

PR #1557, opened 2020-09-30
refactor(environments): delay delivery config last checked at write

robfletcher:
Instead of continuing to check environment promotions for paused apps, just ensure that resource checks wait until environment checks are reasonably recent. This is to allow for environment checks to catch up if the application has been paused, and we'll therefore deploy an up-to-date artifact rather than whatever was approved last time the app was un-paused. If manual intervention has happened during the pause time we could roll back to an unwanted older version.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
@@ -8,10 +8,6 @@ import com.netflix.spinnaker.keel.telemetry.ArtifactCheckTimedOut
 import com.netflix.spinnaker.keel.telemetry.EnvironmentsCheckTimedOut
 import com.netflix.spinnaker.keel.telemetry.ResourceCheckTimedOut
 import com.netflix.spinnaker.keel.telemetry.ResourceLoadFailed
-import java.time.Duration
-import java.util.concurrent.atomic.AtomicBoolean
-import kotlin.coroutines.CoroutineContext
-import kotlin.math.max
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.TimeoutCancellationException
@@ -25,6 +21,10 @@ import org.springframework.context.ApplicationEventPublisher
 import org.springframework.context.event.EventListener
 import org.springframework.scheduling.annotation.Scheduled
 import org.springframework.stereotype.Component
+import java.time.Duration
+import java.util.concurrent.atomic.AtomicBoolean
+import kotlin.coroutines.CoroutineContext
+import kotlin.math.max
 
 @Component
 class CheckScheduler(
@@ -121,6 +121,8 @@ class CheckScheduler(
               } catch (e: TimeoutCancellationException) {
                 log.error("Timed out checking environments for ${it.application}/${it.name}", e)
                 publisher.publishEvent(EnvironmentsCheckTimedOut(it.application, it.name))
+              } finally {
+                repository.markDeliveryConfigCheckComplete(it)
               }
             }
         }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -37,12 +37,13 @@ import com.netflix.spinnaker.keel.veto.VetoResponse
 import com.netflix.spinnaker.kork.exceptions.SpinnakerException
 import com.netflix.spinnaker.kork.exceptions.SystemException
 import com.netflix.spinnaker.kork.exceptions.UserException
-import java.time.Clock
 import kotlinx.coroutines.async
 import kotlinx.coroutines.supervisorScope
 import org.slf4j.LoggerFactory
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Component
+import java.time.Clock
+import java.time.Duration
 
 /**
  * The core component in keel responsible for resource state monitoring and actuation.
@@ -85,6 +86,17 @@ class ResourceActuator(
         return@withTracingContext
       }
 
+      val deliveryConfig = deliveryConfigRepository.deliveryConfigFor(resource.id)
+      val environment = checkNotNull(deliveryConfig.environmentFor(resource)) {
+        "Failed to find environment for ${resource.id} in deliveryConfig ${deliveryConfig.name}"
+      }
+
+      if (deliveryConfig.isPromotionCheckStale()) {
+        log.debug("Environment check for {} is {} old, skipping checks", id, deliveryConfig.isPromotionCheckStale())
+        publisher.publishEvent(ResourceCheckSkipped(resource.kind, id, "PromotionCheckStale"))
+        return@withTracingContext
+      }
+
       try {
         val (desired, current) = plugin.resolve(resource)
         val diff = DefaultResourceDiff(desired, current)
@@ -120,12 +132,6 @@ class ResourceActuator(
 
               if (versionedArtifact != null) {
                 with(versionedArtifact) {
-                  val deliveryConfig = deliveryConfigRepository.deliveryConfigFor(resource.id)
-                  val environment = deliveryConfig.environmentFor(resource)?.name
-                    ?: error(
-                      "Failed to find environment for ${resource.id} in deliveryConfig ${deliveryConfig.name} " +
-                        "while attempting to veto artifact $artifactType:$artifactName version $artifactVersion"
-                    )
                   val artifact = deliveryConfig.matchingArtifactByName(versionedArtifact.artifactName, artifactType)
                     ?: error("Artifact $artifactType:$artifactName not found in delivery config ${deliveryConfig.name}")
 
@@ -134,7 +140,7 @@ class ResourceActuator(
                     veto = EnvironmentArtifactVeto(
                       reference = artifact.reference,
                       version = artifactVersion,
-                      targetEnvironment = environment,
+                      targetEnvironment = environment.name,
                       vetoedBy = "Spinnaker",
                       comment = "Automatically marked as bad because multiple deployments of this version failed."
                     )
@@ -212,6 +218,14 @@ class ResourceActuator(
         publisher.publishEvent(ResourceCheckError(resource, e.toSpinnakerException(), clock))
       }
     }
+  }
+
+  private fun DeliveryConfig.isPromotionCheckStale(): Boolean {
+    val age = Duration.between(
+      deliveryConfigRepository.deliveryConfigLastChecked(this),
+      clock.instant()
+    )
+    return age > Duration.ofMinutes(5)
   }
 
   private fun Exception.toSpinnakerException(): SpinnakerException =
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -92,7 +92,7 @@ class ResourceActuator(
       }
 
       if (deliveryConfig.isPromotionCheckStale()) {
-        log.debug("Environment check for {} is {} old, skipping checks", id, deliveryConfig.isPromotionCheckStale())
+        log.debug("Environments check for {} is stale, skipping checks", deliveryConfig.name)
         publisher.publishEvent(ResourceCheckSkipped(resource.kind, id, "PromotionCheckStale"))
         return@withTracingContext
       }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/PeriodicallyCheckedRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/PeriodicallyCheckedRepository.kt
@@ -12,4 +12,10 @@ interface PeriodicallyCheckedRepository<T : Any> {
    * different values.
    */
   fun itemsDueForCheck(minTimeSinceLastCheck: Duration, limit: Int): Collection<T>
+
+  /**
+   * Optional operation to mark a check as complete. If this is not implemented it's assumed that
+   * [itemsDueForCheck] takes care of that.
+   */
+  fun markCheckComplete(deliveryConfig: T) {}
 }
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
@@ -16,7 +16,6 @@ import com.netflix.spinnaker.keel.core.api.ApplicationSummary
 import com.netflix.spinnaker.keel.core.api.UID
 import com.netflix.spinnaker.keel.core.api.parseUID
 import com.netflix.spinnaker.keel.core.api.randomUID
-import com.netflix.spinnaker.keel.pause.PauseScope
 import com.netflix.spinnaker.keel.pause.PauseScope.APPLICATION
 import com.netflix.spinnaker.keel.persistence.DeliveryConfigRepository
 import com.netflix.spinnaker.keel.persistence.NoDeliveryConfigForApplication
@@ -41,10 +40,6 @@ import com.netflix.spinnaker.keel.persistence.metamodel.Tables.RESOURCE_WITH_MET
 import com.netflix.spinnaker.keel.resources.ResourceSpecIdentifier
 import com.netflix.spinnaker.keel.sql.RetryCategory.READ
 import com.netflix.spinnaker.keel.sql.RetryCategory.WRITE
-import java.time.Clock
-import java.time.Duration
-import java.time.Instant.EPOCH
-import java.time.ZoneOffset
 import org.jooq.DSLContext
 import org.jooq.Record1
 import org.jooq.Select
@@ -52,10 +47,14 @@ import org.jooq.impl.DSL
 import org.jooq.impl.DSL.inline
 import org.jooq.impl.DSL.row
 import org.jooq.impl.DSL.select
-import org.jooq.impl.DSL.update
 import org.jooq.util.mysql.MySQLDSL
 import org.slf4j.LoggerFactory
 import java.net.InetAddress
+import java.time.Clock
+import java.time.Duration
+import java.time.Instant
+import java.time.Instant.EPOCH
+import java.time.ZoneOffset.UTC
 
 class SqlDeliveryConfigRepository(
   private val jooq: DSLContext,
@@ -616,11 +615,11 @@ class SqlDeliveryConfigRepository(
             artifactReference,
             constraintType,
             ConstraintStatus.valueOf(status),
-            createdAt.toInstant(ZoneOffset.UTC),
+            createdAt.toInstant(UTC),
             judgedBy,
             when (judgedAt) {
               null -> null
-              else -> judgedAt.toInstant(ZoneOffset.UTC)
+              else -> judgedAt.toInstant(UTC)
             },
             comment,
             mapper.readValue(attributes)
@@ -669,11 +668,11 @@ class SqlDeliveryConfigRepository(
             artifactReference,
             constraintType,
             ConstraintStatus.valueOf(status),
-            createdAt.toInstant(ZoneOffset.UTC),
+            createdAt.toInstant(UTC),
             judgedBy,
             when (judgedAt) {
               null -> null
-              else -> judgedAt.toInstant(ZoneOffset.UTC)
+              else -> judgedAt.toInstant(UTC)
             },
             comment,
             mapper.readValue(attributes)
@@ -786,11 +785,11 @@ class SqlDeliveryConfigRepository(
           artifactReference,
           type,
           ConstraintStatus.valueOf(status),
-          createdAt.toInstant(ZoneOffset.UTC),
+          createdAt.toInstant(UTC),
           judgedBy,
           when (judgedAt) {
             null -> null
-            else -> judgedAt.toInstant(ZoneOffset.UTC)
+            else -> judgedAt.toInstant(UTC)
           },
           comment,
           mapper.readValue(attributes)
@@ -854,11 +853,11 @@ class SqlDeliveryConfigRepository(
             artifactReference,
             constraintType,
             ConstraintStatus.valueOf(status),
-            createdAt.toInstant(ZoneOffset.UTC),
+            createdAt.toInstant(UTC),
             judgedBy,
             when (judgedAt) {
               null -> null
-              else -> judgedAt.toInstant(ZoneOffset.UTC)
+              else -> judgedAt.toInstant(UTC)
             },
             comment,
             mapper.readValue(attributes)
@@ -924,11 +923,11 @@ class SqlDeliveryConfigRepository(
             artifactReference,
             constraintType,
             ConstraintStatus.valueOf(status),
-            createdAt.toInstant(ZoneOffset.UTC),
+            createdAt.toInstant(UTC),
             judgedBy,
             when (judgedAt) {
               null -> null
-              else -> judgedAt.toInstant(ZoneOffset.UTC)
+              else -> judgedAt.toInstant(UTC)
             },
             comment,
             mapper.readValue(attributes)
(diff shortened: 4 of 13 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/PeriodicallyCheckedRepository.kt
```diff
@@ -12,4 +12,10 @@ interface PeriodicallyCheckedRepository<T : Any> {
    * different values.
    */
   fun itemsDueForCheck(minTimeSinceLastCheck: Duration, limit: Int): Collection<T>
+
+  /**
+   * Optional operation to mark a check as complete. If this is not implemented it's assumed that
+   * [itemsDueForCheck] takes care of that.
+   */
+  fun markCheckComplete(deliveryConfig: T) {}
```
robfletcher, line 20: New method that allows repository implementations to have async check cycles

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -594,18 +596,18 @@ class SqlDeliveryConfigRepository(
...
+                      artifactVersion,
+                      artifactReference,
+                      constraintType,
+                      status,
+                      createdAt,
+                      judgedBy,
+                      judgedAt,
+                      comment,
+                      attributes
+                    ) ->
```
robfletcher, line 610: I don't know why spotless did this but I'm not going to waste time fighting it

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -1042,18 +1044,17 @@ class SqlDeliveryConfigRepository(
         select(DELIVERY_CONFIG.UID, DELIVERY_CONFIG.NAME)
           .from(DELIVERY_CONFIG, DELIVERY_CONFIG_LAST_CHECKED)
           .where(DELIVERY_CONFIG.UID.eq(DELIVERY_CONFIG_LAST_CHECKED.DELIVERY_CONFIG_UID))
+          .and(DELIVERY_CONFIG_LAST_CHECKED.LOCKED_BY.isNull)
           .and(DELIVERY_CONFIG_LAST_CHECKED.AT.lessOrEqual(cutoff))
           .orderBy(DELIVERY_CONFIG_LAST_CHECKED.AT)
```
robfletcher, line 1049: Only get delivery configs that aren't already being checked by another instance

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -1042,18 +1044,17 @@ class SqlDeliveryConfigRepository(
...
             it.forEach { (uid, _) ->
-              insertInto(DELIVERY_CONFIG_LAST_CHECKED)
-                .set(DELIVERY_CONFIG_LAST_CHECKED.DELIVERY_CONFIG_UID, uid)
-                .set(DELIVERY_CONFIG_LAST_CHECKED.AT, now.toTimestamp())
-                .onDuplicateKeyUpdate()
-                .set(DELIVERY_CONFIG_LAST_CHECKED.AT, now.toTimestamp())
+              update(DELIVERY_CONFIG_LAST_CHECKED)
+                .set(DELIVERY_CONFIG_LAST_CHECKED.LOCKED_BY, InetAddress.getLocalHost().hostName)
+                .where(DELIVERY_CONFIG_LAST_CHECKED.DELIVERY_CONFIG_UID.eq(uid))
                 .execute()
```
robfletcher, line 1058: Mark that this instance is working on this delivery config so others won't pick it up (the value in the `locked_by` column is arbitrary but might help with debugging)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -1064,6 +1065,21 @@ class SqlDeliveryConfigRepository(
...
+        .setNull(DELIVERY_CONFIG_LAST_CHECKED.LOCKED_BY)
+        .set(DELIVERY_CONFIG_LAST_CHECKED.AT, clock.instant().toTimestamp())
+        .where(DELIVERY_CONFIG_LAST_CHECKED.DELIVERY_CONFIG_UID.eq(
+          select(DELIVERY_CONFIG.UID)
+            .from(DELIVERY_CONFIG)
+            .where(DELIVERY_CONFIG.NAME.eq(deliveryConfig.name))
+        ))
+        .execute()
+    }
+  }
```
robfletcher, line 1081: Clear the `locked_by` column and update the last check timestamp

keel-sql/src/main/resources/db/changelog/20200930-environment-check-lock.yml
```diff
@@ -0,0 +1,24 @@
...
+    author: fletch
+    changes:
+    - addColumn:
+        tableName: delivery_config_last_checked
+        columns:
+        - column:
+            name: locked_by
+            type: varchar(64)
+            constraints:
+              nullable: true
```
robfletcher, line 13: Add the new column to the table

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -85,6 +86,17 @@ class ResourceActuator(
...
+      val deliveryConfig = deliveryConfigRepository.deliveryConfigFor(resource.id)
+      val environment = checkNotNull(deliveryConfig.environmentFor(resource)) {
+        "Failed to find environment for ${resource.id} in deliveryConfig ${deliveryConfig.name}"
+      }
+
+      if (deliveryConfig.isPromotionCheckStale()) {
+        log.debug("Environment check for {} is {} old, skipping checks", id, deliveryConfig.isPromotionCheckStale())
+        publisher.publishEvent(ResourceCheckSkipped(resource.kind, id, "PromotionCheckStale"))
+        return@withTracingContext
+      }
```
robfletcher, line 98: Skip resource check if environment has not been checked recently. This is to allow for environment checks to catch up if the application has been paused, and we'll therefore deploy an up-to-date artifact rather than whatever was approved last time the app was un-paused. If manual intervention has happened during the pause time we could roll back to an unwanted older version.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -120,12 +132,6 @@ class ResourceActuator(
 
               if (versionedArtifact != null) {
                 with(versionedArtifact) {
-                  val deliveryConfig = deliveryConfigRepository.deliveryConfigFor(resource.id)
-                  val environment = deliveryConfig.environmentFor(resource)?.name
-                    ?: error(
-                      "Failed to find environment for ${resource.id} in deliveryConfig ${deliveryConfig.name} " +
-                        "while attempting to veto artifact $artifactType:$artifactName version $artifactVersion"
-                    )
```
robfletcher, line 128: Just pulled this up earlier in the method since I needed to get the delivery config

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -209,6 +215,14 @@ class ResourceActuator(
     }
   }
 
+  private fun DeliveryConfig.isPromotionCheckStale(): Boolean {
+    val age = Duration.between(
+      deliveryConfigRepository.deliveryConfigLastChecked(this),
+      clock.instant()
+    )
+    return age > Duration.ofMinutes(5)
+  }
```
robfletcher, line 224: Seems a reasonable threshold to me 🤷

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -1125,6 +1147,18 @@ class SqlDeliveryConfigRepository(
...
+    return sqlRetry.withRetry(READ) {
+      jooq
+        .select(DELIVERY_CONFIG_LAST_CHECKED.AT)
+        .from(DELIVERY_CONFIG_LAST_CHECKED, DELIVERY_CONFIG)
+        .where(DELIVERY_CONFIG_LAST_CHECKED.DELIVERY_CONFIG_UID.eq(DELIVERY_CONFIG.UID))
+        .and(DELIVERY_CONFIG.NAME.eq(deliveryConfig.name))
+        .fetchOne(DELIVERY_CONFIG_LAST_CHECKED.AT)
+        .toInstant(UTC)
+    }
+  }
```
robfletcher, line 1160: New method to just get the timestamp when the delivery config was last check for environment promotions

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepositoryPeriodicallyCheckedTests.kt
```diff
@@ -24,12 +28,34 @@ internal object SqlDeliveryConfigRepositoryPeriodicallyCheckedTests :
...
+            SqlPausedRepository(jooq, sqlRetry, clock)
+              .pauseApplication(it.first().application, "fzlem@netflix.com")
+        }
+      }
+
+      test("delivery config is ignored") {
+        expectThat(nextResults()).hasSize(1)
+      }
+    }
+  }
```
robfletcher, line 60: Tests that paused apps' delivery configs do not get considered for checking

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -1042,18 +1044,27 @@ class SqlDeliveryConfigRepository(
         select(DELIVERY_CONFIG.UID, DELIVERY_CONFIG.NAME)
           .from(DELIVERY_CONFIG, DELIVERY_CONFIG_LAST_CHECKED)
           .where(DELIVERY_CONFIG.UID.eq(DELIVERY_CONFIG_LAST_CHECKED.DELIVERY_CONFIG_UID))
+          .and(
+            DELIVERY_CONFIG_LAST_CHECKED.LOCKED_BY.isNull
+              .or(DELIVERY_CONFIG_LAST_CHECKED.LOCKED_AT.lessOrEqual(cutoff))
+          )
```
robfletcher, line 1050: This clause allows the delivery config to be reconsidered for checking if the lock is stale. I figured using the same cutoff made sense rather than overcomplicating things.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -1042,18 +1044,27 @@ class SqlDeliveryConfigRepository(
...
           .and(DELIVERY_CONFIG_LAST_CHECKED.AT.lessOrEqual(cutoff))
+          .andNotExists(
+            selectOne()
+              .from(PAUSED)
+              .where(PAUSED.NAME.eq(DELIVERY_CONFIG.APPLICATION))
+              .and(PAUSED.SCOPE.eq(APPLICATION.name))
+          )
```
robfletcher, line 1057: Don't allow paused apps delivery configs to be checked

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -1042,18 +1044,27 @@ class SqlDeliveryConfigRepository(
...
             it.forEach { (uid, _) ->
-              insertInto(DELIVERY_CONFIG_LAST_CHECKED)
-                .set(DELIVERY_CONFIG_LAST_CHECKED.DELIVERY_CONFIG_UID, uid)
-                .set(DELIVERY_CONFIG_LAST_CHECKED.AT, now.toTimestamp())
-                .onDuplicateKeyUpdate()
-                .set(DELIVERY_CONFIG_LAST_CHECKED.AT, now.toTimestamp())
+              update(DELIVERY_CONFIG_LAST_CHECKED)
+                .set(DELIVERY_CONFIG_LAST_CHECKED.LOCKED_BY, InetAddress.getLocalHost().hostName)
+                .set(DELIVERY_CONFIG_LAST_CHECKED.LOCKED_AT, now.toTimestamp())
+                .where(DELIVERY_CONFIG_LAST_CHECKED.DELIVERY_CONFIG_UID.eq(uid))
```
robfletcher, line 1067: Instead of updating the checked at timestamp right away, mark the delivery config as locked (using local host name is just convenient for debugging, the value is arbitrary). Also added a `locked_at` time so we can ignore stale locks.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -1064,6 +1075,22 @@ class SqlDeliveryConfigRepository(
...
+        .setNull(DELIVERY_CONFIG_LAST_CHECKED.LOCKED_AT)
+        .set(DELIVERY_CONFIG_LAST_CHECKED.AT, clock.instant().toTimestamp())
+        .where(DELIVERY_CONFIG_LAST_CHECKED.DELIVERY_CONFIG_UID.eq(
+          select(DELIVERY_CONFIG.UID)
+            .from(DELIVERY_CONFIG)
+            .where(DELIVERY_CONFIG.NAME.eq(deliveryConfig.name))
+        ))
+        .execute()
+    }
+  }
```
robfletcher, line 1092: On completion of the check cycle clear the lock and update the last checked timestamp. Only doing this _after_ the check itself is complete means we can delay resource checks until environment checks are up-to-date.

keel-sql/src/main/resources/db/changelog/20200930-environment-check-lock.yml
```diff
@@ -0,0 +1,29 @@
...
+        - column:
+            name: locked_at
+            type: timestamp(3)
+            constraints:
+              nullable: true
```
robfletcher, line 18: 2 new columns

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepositoryPeriodicallyCheckedTests.kt
```diff
@@ -24,12 +30,63 @@ internal object SqlDeliveryConfigRepositoryPeriodicallyCheckedTests :
...
+            SqlPausedRepository(jooq, sqlRetry, clock)
+              .pauseApplication(it.first().application, "fzlem@netflix.com")
+          }
+      }
+
+      test("delivery config is ignored") {
+        expectThat(nextResults()).hasSize(1)
+      }
+    }
+  }
```
robfletcher, line 62: Test we don't check delivery configs for paused apps

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepositoryPeriodicallyCheckedTests.kt
```diff
@@ -24,12 +30,63 @@ internal object SqlDeliveryConfigRepositoryPeriodicallyCheckedTests :
...
+        before {
+          clock.incrementBy(Duration.ofMinutes(35))
+        }
+
+        test("the lock expires") {
+          expectThat(nextResults()).hasSize(1)
+        }
+      }
+    }
+  }
```
robfletcher, line 91: Test we don't check delivery configs already being checked by another instance, and that we can ignore a stale lock and re-check after a certain amount of time.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
lorin: This means a five-minute latency between un-pausing an app and seeing actuation start, right? That feels a little long-ish to me. (Maybe two minutes instead?) 

The problem is, I don't have a sense of the worst-case delay for the environment check to happen, so I don't know what the threshold is for too small a waiting time. At some point, we should do some benchmarking to get a sense of how long it typically takes keel to do a full sweep of environment checks.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
lorin: What's the hazard (dangerous scenario) that the lock prevents? (i.e., what is the bad thing that might happen if this code just set the last checked timestamp and didn't do any locking?)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
lorin: It looks like a precondition of this function is that there exists a record in the delivery_config_last_checked table for that delivery config. Will that always be true? 

If not, I think it's safer to have this return `Instant?` and have the caller handle the case when there's no record yet.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
lorin: As an aside, I really like the term [leases](http://martin.kleppmann.com/2016/02/08/how-to-do-distributed-locking.html) for the kinds of lock implemented here, that expires after a certain amount of time.

lorin (approved):
Some comments and questions, but nothing that would block merge.

lorin:
One thing I forgot to mention: there are a lot of good comments you left as PR annotations that document your rationale. I recommend you add these as comments in the code, to aid future readers.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
robfletcher: I should probably make this configurable but, it's not a 5 minute delay. If the environment check is older than 5 minutes, we just put the resource check to the back of the line. The theory being the environment check is likely happening _right now_ or is about to. Next time the resource is checked the environment check will likely be up to date. That should be a lot sooner than 5 minutes.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
robfletcher: If you look in the `store` method, we always insert a row when we initially create the delivery config

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
robfletcher: Yeah, that's good. I'll change it

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
robfletcher: Multiple Keel instances trying to check the same set of environments at once. Unlikely to cause an issue, but especially if the app has been paused for a while, it's probably going to be a slower check and may delay other work.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -25,6 +21,10 @@ import org.springframework.context.ApplicationEventPublisher
 import org.springframework.context.event.EventListener
 import org.springframework.scheduling.annotation.Scheduled
 import org.springframework.stereotype.Component
+import java.time.Duration
+import java.util.concurrent.atomic.AtomicBoolean
+import kotlin.coroutines.CoroutineContext
+import kotlin.math.max
```
luispollo, line 27: Looks like none of these are used below?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -85,6 +86,17 @@ class ResourceActuator(
         return@withTracingContext
       }
 
+      val deliveryConfig = deliveryConfigRepository.deliveryConfigFor(resource.id)
+      val environment = checkNotNull(deliveryConfig.environmentFor(resource)) {
+        "Failed to find environment for ${resource.id} in deliveryConfig ${deliveryConfig.name}"
+      }
+
+      if (deliveryConfig.isPromotionCheckStale()) {
+        log.debug("Environment check for {} is {} old, skipping checks", id, deliveryConfig.isPromotionCheckStale())
```
luispollo, line 95:
```suggestion
        log.debug("Promotion check for environment {} in delivery config {} is old, skipping checks for resource {}", environment.name, deliveryConfig.name, id)
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: My guess is because of the parenthesis on the same line as the curly brace.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: Just to confirm: in the unlikely event that the instance holding the lock crashes just before reaching that `finally` block where the call to this method is made, the lock will eventually expire, which will cause the environment to be rechecked, correct?

keel-sql/src/main/resources/db/changelog/20200930-environment-check-lock.yml
```diff
@@ -0,0 +1,29 @@
...
+        tableName: delivery_config_last_checked
+        columns:
+        - column:
+            name: locked_by
+            type: varchar(64)
+            constraints:
+              nullable: true
+        - column:
+            name: locked_at
+            type: timestamp(3)
```
luispollo, line 16: I think we've been using `timestamp(6)` as a standard?

luispollo (approved):
Looks good, @robfletcher! Just a minor suggestion on a log message and a question, nothing major.

keel-sql/src/main/resources/db/changelog/20200930-environment-check-lock.yml
robfletcher: No, that would be nanosecond precision. `timestamp(3)` is millisecond.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
robfletcher: Done

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
robfletcher: Right, that's why the select is predicated on `locked_by` being `null` _or_ `locked_at` being <= some cutoff.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
robfletcher: Ugh, I think I originally wrote the log line when I was using a `StopWatch` to time the execution. I'll fix this up

PR #1560, opened 2020-10-01
fix(bakery): better exception message on bad AMI version string

lorin:
If the app version associated with an AMI does not conform to the Frigga library convention, the exception thrown is:

```
java.lang.NullPointerException: AppVersion.parseName(version) must not be null
	at com.netflix.spinnaker.keel.bakery.constraint.ImageExistsConstraintEvaluator$findMatchingImage$1.invokeSuspend(ImageExistsConstraintEvaluator.kt:57)
```

This change displays the problematic app version string to the exception message.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
@@ -52,7 +52,7 @@ class ImageExistsConstraintEvaluator(
 
   private fun findMatchingImage(version: String, vmOptions: VirtualMachineOptions): NamedImage? {
     log.debug("Searching for baked image for {} in {}", version, vmOptions.regions.joinToString())
-    val appVersion = AppVersion.parseName(version) ?: throw SystemException("Invalid ami app version: $version")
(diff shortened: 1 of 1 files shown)


keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
```diff
@@ -51,10 +52,11 @@ class ImageExistsConstraintEvaluator(
 
   private fun findMatchingImage(version: String, vmOptions: VirtualMachineOptions): NamedImage? {
     log.debug("Searching for baked image for {} in {}", version, vmOptions.regions.joinToString())
+    val appVersion = AppVersion.parseName(version) ?: throw SystemException("Invalid ami app version: $version")
```
lorin, line 55: Not sure whether `SystemException` was the right one to use here.

emjburns (approved):

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
luispollo, line 55:
```suggestion
    val appVersion = AppVersion.parseName(version) ?: throw SystemException("Invalid AMI app version: $version")
```

luispollo (approved):

PR #1563, opened 2020-10-02
feat(notifications): resource notifications, unhealthy notifications, action count 

emjburns:
This PR does a couple of things

(1) updates how we detect that we should veto something by looking at not just how many times we've seen a diff, but how many times we have taken action on it. This is a prerequisite change for some future changes I'm going to make but I think it's important enough to do here (so we can roll it out separately and because it's top of my mind)
(2) add a generalized notifier that allows repeat notifications for things (just resources for now) by pulling the environment config for a resource and using that to send the notification. This can be generalized a lot more, but no need right now
(3) adds unhealthy notifications that will notify you when we detect a resource has become unhealthy. This notification is cleared when the resource becomes healthy

The next thing I'm doing will actually address doing the right thing for unhealthy resources. Stay tuned.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -165,6 +165,9 @@ class ResourceActuator(
 
         log.debug("Checking resource {}", id)
```
emjburns, line 167: I'll be doing this todo next

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListener.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.netflix.spinnaker.keel.notifications.NotificationScope.RESOURCE
+import com.netflix.spinnaker.keel.notifications.Notifier
+import com.netflix.spinnaker.keel.notifications.Notifier.UNHEALTHY
+import com.netflix.spinnaker.keel.veto.unhealthy.UnsupportedResourceTypeException
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.context.event.EventListener
+import org.springframework.stereotype.Component
+
+
```
emjburns, line 21: This class creates the nicely formatted message to send.

@erikmunson I had to remove the duration unhealthy. We can bring it back, but I want to think more deeply about how to save that because we won't be recording it anymore. It would just be recorded for the notification at this point

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -173,6 +176,7 @@ class ResourceActuator(
             plugin.create(resource, diff)
               .also { tasks ->
                 publisher.publishEvent(ResourceActuationLaunched(resource, plugin.name, tasks, clock))
+                diffFingerprintRepository.markActionTaken(id)
```
emjburns, line 179: this is a change. previously, we cared only about whether we saw a diff. now, we're marking that we are taking action on a diff. this is used in the UnhappyVeto, so that we only veto an artifact if we've taken action on it the allotted number of times.

no behavior change yet

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/friendlyTime.kt
```diff
@@ -0,0 +1,29 @@
+package com.netflix.spinnaker.keel.notifications
+
+import java.time.Duration
+
```
emjburns, line 4: I worked hard on this and am keeping it in the code base even though I'm not using it yet. YET.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
```diff
@@ -49,7 +49,8 @@ class UnhappyVeto(
     val resourceId = resource.id
     val application = resource.application
 
-    if (diffFingerprintRepository.diffCount(resourceId) <= maxDiffCount(resource)) {
+    // maxDiff count represents the number of times we're allowed to see a diff and take action to try and fix it.
+    if (diffFingerprintRepository.actionTakenCount(resourceId) <= maxDiffCount(resource)) {
```
emjburns, line 53: using action taken count instead of diff count here to prepare for future changes

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/friendlyTime.kt
erikmunson: i feel this to my core.

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ResourceNotifier.kt
```diff
@@ -0,0 +1,149 @@
...
+  private val echoService: EchoService,
+  private val keelRepository: KeelRepository,
+  private val notifierRepository: NotifierRepository,
+  private val keelNotificationConfig: KeelNotificationConfig,
+  private val springEnv: Environment,
+  private val spectator: Registry
+) {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+  private val NOTIFICATION_SENT_ID = "keel.notification.sent"
+
```
emjburns, line 55: sending notifications is behind a fast property so we can shut it off

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/model/EchoNotification.kt
```diff
@@ -13,8 +13,9 @@ data class EchoNotification(
   val templateGroup: String? = null,
   val severity: Severity,
   val source: Source? = null,
-  val interactiveActions: InteractiveActions?,
-  val additionalContext: Map<String, Any?>? = null
+  val interactiveActions: InteractiveActions? = null,
+  val additionalContext: Map<String, Any?>? = null,
+  val useInteractiveBot: Boolean = true
```
emjburns, line 18: @luispollo here is where we default to using the interactive bot for all of keel notifications (not task notifications, just ones sent like this by keel)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: This feels like the wrong repository/table to use to store this information. IMHO, this should go in the `resources` table, as we're counting action taken to resolve a diff from desired to current state of the resource. It's got nothing to do with diff fingerprints.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/NotifierRepository.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.netflix.spinnaker.keel.notifications.NotificationScope
+import com.netflix.spinnaker.keel.notifications.Notifier
+import org.springframework.beans.factory.annotation.Value
+import java.time.Clock
+
+/**
+ * A repository for storing a list of ongoing notifications, and calculating whether
+ * they should be resent
+ */
+abstract class NotifierRepository(
```
luispollo, line 12: As I pointed out in the previous PR, I think we should fix the discrepancy between the terminology and comments and the class names. If we're storing a "list of ongoing notifications", then this should be a `NotificationRepository`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/Notifier.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.keel.notifications
+
+/**
+ * All valid notifiers
+ */
+enum class Notifier {
```
luispollo, line 6: I am sure I wrote more comments than I can now find in the previous PR. I don't know what happened to them... 😞

Anyway, I had made a suggestion to refactor these models slightly to improve the naming and encapsulation. Something like this:
```kotlin
enum class NotificationType {
  RESOURCE_UNHEALTHY
}

abstract class Notification(
  abstract val scope: NotificationScope,
  abstract val ref: String,
  val subject: String,
  val body: String,
  val color: String = "#cccccc"
)

data class ResourceNotification(
  val resource: Resource<*>,
  val subject: String,
  val body: String,
  val color: String = "#cccccc"
) : Notification(NotificationScope.RESOURCE, resource.id, subject, body, color)

data class ResourceUnhealthyNotification(
  override val resource: Resource<*>
) : ResourceNotification(resource, "Managed Delivery", "The resource ${resource.id} is unhealthy.")
```
That way, your repository methods could just take the instance of the `Notification` object, which would encapsulate all the information you need to persist in the database. It would also remove the "notifier" term from these classes and leave that term for the classes which actually do the work of sending the notifications.

luispollo: P.S. This is not an original idea by any means. We use the same approach with the `PersistentEvent` and `ResourceEvent` class hierarchy. 🙂

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
luispollo: Can you say more about what those future changes will be?

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ResourceNotifier.kt
```diff
@@ -0,0 +1,149 @@
...
+    get() = springEnv.getProperty("keel.notifications.resource", Boolean::class.java, true)
+
+  @EventListener(ResourceNotificationEvent::class)
+  fun onResourceNotificationEvent(event: ResourceNotificationEvent) {
+    if (notificationsEnabled){
+      val shouldNotify = notifierRepository.addNotification(event.scope, event.identifier, event.notifier)
+      if (shouldNotify) {
+        notify(event)
+        notifierRepository.markSent(event.scope, event.identifier, event.notifier)
+        spectator.counter(
```
luispollo, line 66: Nice! ❤️

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListener.kt
```diff
@@ -0,0 +1,74 @@
...
+      q = resource.spec.displayName,
+      stack = spec.moniker.stack ?: "(none)",
+      detail = spec.moniker.detail ?: "(none)"
+    )
+    val resourceUrl = "$spinnakerBaseUrl/#/applications/${resource.application}/clusters?${params.toURL()}"
+
+    return NotificationMessage(
+      subject = "${resource.spec.displayName} is unhealthy",
+      body = "<$resourceUrl|${resource.id}> is unhealthy and " +
+        "Spinnaker does not detect any changes to this resource. " +
```
luispollo, line 68:
```suggestion
        "Spinnaker did not detect any changes to this resource. " +
```

luispollo, line 68: Suggestion:
```suggestion
      body = "Spinnaker has detected that <$resourceUrl|${resource.id}> is unhealthy, but " +
        "there have been no changes to its configuration. " +
```

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/model/EchoNotification.kt
luispollo: Cool. We might want to fix the model in echo such that you don't have to specify a whole `interactiveActions` block just to specify the color, but that's icing on the cake. 🙂

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ResourceNotifier.kt
```diff
@@ -0,0 +1,149 @@
...
+      additionalContext = mapOf(
+        "formatter" to "MARKDOWN",
+        "subject" to event.message.subject,
+        "body" to event.message.body
+      ),
+      interactiveActions = generateInteractiveConfig(event.identifier, event.notifier.name, event.message.color)
+    )
+  }
+
+  private fun generateInteractiveConfig(resourceId: String, notifier: String, color: String): InteractiveActions? =
```
luispollo, line 127: We might want to fix the model in echo such that you don't have to specify a whole `interactiveActions` block just to specify the color, but that's icing on the cake. 🙂

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
gal-yardeni: @emjburns doesn't the column `count` in `diff_fingerprint` table is the counter for each time we are taking action on a diff? or is it just when we are seeing a diff?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/NotificationMessage.kt
```diff
@@ -0,0 +1,7 @@
+package com.netflix.spinnaker.keel.notifications
+
+data class NotificationMessage(
```
gal-yardeni, line 3: is this structure limited to a specific type of notification?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/friendlyTime.kt
gal-yardeni: omg I sware I saw this as an interview question once 😃

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ResourceNotifier.kt
```diff
@@ -0,0 +1,149 @@
...
+      if (shouldNotify) {
+        notify(event)
+        notifierRepository.markSent(event.scope, event.identifier, event.notifier)
+        spectator.counter(
+          NOTIFICATION_SENT_ID,
+          listOf(BasicTag("notifier", event.notifier.name))
+        )
+          .runCatching { increment() }
+          .onFailure {
+            log.error("Exception incrementing {} counter: {}", NOTIFICATION_SENT_ID, it.message)
```
lorin, line 72: I recommend passing the exception itself (`it`) instead of just the message (`it.message`) as the last argument to `log.error`. That way, both the message and the stack trace will get captured to the log.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/TitusClusterHandler.kt
```diff
@@ -497,6 +497,9 @@ class TitusClusterHandler(
     val healthy: Boolean = activeServerGroups.all {
       it.instanceCounts?.isHealthy(resource.spec.deployWith.health) == true
     }
+
+    eventPublisher.publishEvent(ResourceHealthEvent(resource, healthy))
```
lorin, line 501: Do we have documentation somewhere that describes the contract that we expect a resource handler to conform to? This PR looks like it modifies the existing contract by now expecting resource handlers to publish resource health events.

lorin (approved):
Just a couple of minor questions/comments.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: The existing column is just when we see a diff

emjburns: I think you could make a case for either place. It is super related to diff fingerprints because we take action on the diff that that fingerprint represents. I think you could also make the case that we should store the fingerprint with the resource. But, it seems much simpler to store them in a different table. To me the resource table is about the config not about what we're doing in the moment.

I'd like to leave it the way it is.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/NotificationMessage.kt
emjburns: No, it's the same for both email and slack

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/NotificationScope.kt
```diff
@@ -0,0 +1,5 @@
+package com.netflix.spinnaker.keel.notifications
+
+enum class NotificationScope {
+  RESOURCE
```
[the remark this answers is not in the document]
emjburns: Correct, I think we're safe here. We're only reading the time out of the database. That's a good point though. I tend to go with enum because it feels safer than strings, but it could end up being "too safe".

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/NotifierRepository.kt
emjburns: Okay, I can change it to be `NotificationRepository`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
emjburns: https://github.com/spinnaker/keel/pull/1563/files#diff-43eecffdd336c15071ef33731c885473R168-R169

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ResourceNotifier.kt
emjburns: Sure, we could do that in a follow up PR. It doesn't feel that important though

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListener.kt
emjburns: I don't think this is any clearer, and I don't like the use of "but" (a configuration change is not the only reason something could be unhealthy). I'm going to keep it as is for now

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/TitusClusterHandler.kt
emjburns: No, I don't think we do.

Longer term I really want to create a common base for cluster handlers, because so much of the code is doing the same transforms. I think that would help alleviate the need for making the expectations clear.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/Notifier.kt
emjburns: @luispollo I'm going to take some of these suggestions, but not all of them. I think what you're trying to convey is that some refactoring to the structures used here can help them be a lot more easily extendable. 

In the class description, I commented that this class will need work to make it super generic and that I don't think it's worth it at this moment since we only have one type of notification. It took me a long time to understand why this refactoring would be helpful, so in the future if you could add the "why" before suggesting a big change that would help me understand where you're coming from

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/TitusClusterHandler.kt
luispollo: I tried to document the contract not too long ago: https://github.com/spinnaker/keel/blob/c3b40de05e19562fadd8ec3d6636e4f84f568e6e/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt#L16-L38

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/Notifier.kt
luispollo: Thanks for being open to the suggestions, @emjburns. While extensibility comes as a nice side-effect of the suggested refactoring, my primary motivation as I tried to explain above was to improve encapsulation and naming. I've observed that we have occurrences in other places in keel as well where we tend to pass in discrete fields to a repository method when the goal is to persist an entire object, and so encapsulating those fields in the corresponding object is just good OO design, IMHO. This is why I suggested moving the notification type and the entity ID (`ref`) into the `Notification` class. That way your `NotificationEvent` could just carry a `Notification`, instead of the discrete fields, and similarly your repository methods would just take a `Notification` as a parameter.

The changes you made were great already. Up to you whether you think the additional bit described above should be done in this PR. I'd be happy to make those changes later as well if it helps.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEvent.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.netflix.spinnaker.keel.notifications.Notification
+import com.netflix.spinnaker.keel.notifications.NotificationScope
+import com.netflix.spinnaker.keel.notifications.NotificationType
+
+data class NotificationEvent(
+  val scope: NotificationScope,
+  val identifier: String,
+  val notificationType: NotificationType,
+  val message: Notification
```
luispollo, line 11: This is what I meant when I brought up encapsulation: these 3 fields could be collapsed into one, the `Notification` object, because those are part of what that object is intended to represent (the notification doesn't make sense without an associated type and identifier).

keel-sql/src/main/resources/db/changelog/20200928-add-notifier-repo.yml
```diff
@@ -0,0 +1,40 @@
+databaseChangeLog:
+  - changeSet:
+      id: add-notifier-repo
+      author: emjburns
+      changes:
+        - createTable:
+            tableName: notifier
```
luispollo, line 7: Please change the table name to match the class:
```suggestion
            tableName: notifications
```

keel-sql/src/main/resources/db/changelog/20200928-add-notifier-repo.yml
```diff
@@ -0,0 +1,40 @@
+databaseChangeLog:
+  - changeSet:
+      id: add-notifier-repo
```
luispollo, line 3:
```suggestion
      id: add-notifications-repo
```

keel-sql/src/main/resources/db/changelog/20200928-add-notifier-repo.yml
```diff
@@ -0,0 +1,40 @@
...
+            tableName: notifier
+            columns:
+              - column:
+                  name: scope
+                  type: varchar(255)
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: identifier
```
luispollo, line 16: To avoid confusing this with an actual identifier for this table (which this is not), we've used the name `ref` in the `events` table, for instance. I'd recommend doing the same here.
```suggestion
                  name: ref
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEvent.kt
```diff
@@ -0,0 +1,13 @@
+package com.netflix.spinnaker.keel.events
+
+import com.netflix.spinnaker.keel.notifications.Notification
+import com.netflix.spinnaker.keel.notifications.NotificationScope
+import com.netflix.spinnaker.keel.notifications.NotificationType
+
+data class NotificationEvent(
+  val scope: NotificationScope,
+  val identifier: String,
```
luispollo, line 9: To avoid confusing this with an actual identifier for this object (which this is not), we've used the name `ref` in the `PersistentEvent` class, which has a similar model. I'd recommend following that pattern here as well.
```suggestion
  val ref: String, // the identifier of the persistent entity (resource, environment, etc.) this notification refers to
```

luispollo (changes requested):
@emjburns Thank you so much for working on this. I ❤️ the new notification capability and the friendly time stuff you've added. I'm marking this as unapproved temporarily just so that you don't forget to rename the `notifier` table to `notifications` to match the renaming of the classes, will approve as soon as that's fixed. Thanks again for taking my suggestions into consideration! 🙏

keel-sql/src/main/resources/db/changelog/20200928-add-notifier-repo.yml
emjburns: ahhhh I missed that one, thanks!

keel-sql/src/main/resources/db/changelog/20200928-add-notifier-repo.yml
emjburns: fyi for database stuff the suggest changes isn't useful because I need to change all the auto generated jooq stuff. just pointing it out is fine!

luispollo: Yeah, but you need to change this file so that jooq will re-generate the classes... 🙂

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/Notifier.kt
emjburns: Discussed offline, for now keeping the three fields since we don't persist the message itself. In the future if we decide to save the message we will rethink this

luispollo (approved):
Thanks for the open and productive discussion. LGTM!

keel-sql/src/main/resources/db/changelog/20200928-add-notifier-repo.yml
emjburns: Ya of course. Pointing it out is very useful. But the "auto commit" stuff that git does is not the complete picture here. I tend to not do the suggestions because I like to just do the change locally to make sure I get everything involved

gal-yardeni (approved):
very cool!

PR #1564, opened 2020-10-02
feat(artifacts): Source-driven artifact filtering and sorting by timestamp

### LINKED ISSUE #1480: Source-driven artifact filtering
Today, we ask people to provide identifying information about their artifact (organization/name, type) and allow a little bit of filtering for different types of artifacts — filtering on release status for debians, and providing a custom regex (😱) for docker image tags.

While this was a reasonable first stab at being able to locate relevant artifacts, it's very quickly becoming clear that our customers have _zero_ interest in thinking about the minutiae of artifacts — how they're built, the way they're categorized, or really even that they exist to some degree. What we've heard in feedback sessions is that folks are really interested in _source control_ — is this a PR? Which branch is it from? That information is what our customers use as their mental model for their versions. Because dealing with these minutiae is tedious and opaque to most customers, they often get scared off from adopting MD when they realize something about their build configuration needs to change (many people have never even read or touched their build config).

Moving to a source-centric way of finding the right artifact has been a non-starter in the past because we haven't had access to that data when looking at a version. With the recent work to [start keeping track of git + build metadata](https://github.com/spinnaker/keel/pull/1429), however, modeling artifacts more closely to the way our customers think about them is now possible.

The three use cases I'm currently aware of that this new metadata could help us address are roughly:
- I want stuff from this branch (e.g. master , dev)
- I want stuff from this set of branches (e.g. starts with feature- but can end with anything)
- I want any pull request build/artifact

If we can give our customers a way to satisfy those needs, my hypothesis is that they'll be more likely to start _and complete_ adoption — and feel that MD is delivering on the value of worrying about fewer details once they do.

luispollo:
This PR introduces the ability to filter artifacts based on source metadata (such as the originating branch, or whether the commit came from a pull request), as well as the ability to sort artifacts without relying on versions, by using the `createdAt` timestamp associated with each artifact in the database.

A new `from` property is introduced in the artifact specification to specify filters based on the origin of the artifact in source control (branches, pull requests, or a combination of both). The filtering and sorting strategies are auto-detected based on the presence of the details under this field.

Sample artifact definitions:
```yaml
# Artifacts from the "master" branch
name: myapp
application: myapp
serviceAccount: myapp@netflix.com
artifacts:
- name: "@mycompany/myapp"
  type: npm
  reference: master
  from:
    branch:
      name: master
```

```yaml
# Artifacts from any branch starting with "feature-"
name: myapp
application: myapp
serviceAccount: myapp@netflix.com
artifacts:
- name: "@mycompany/myapp"
  type: npm
  reference: feature-branches
  from:
    branch:
      startsWith: feature-
```

```yaml
# Artifacts from any branch containing with "feature"
name: myapp
application: myapp
serviceAccount: myapp@netflix.com
artifacts:
- name: "@mycompany/myapp"
  type: npm
  reference: feature-branches
  from:
    branch:
      regex: ".*feature.*"
```

```yaml
# Artifacts from any pull request
name: myapp
application: myapp
serviceAccount: myapp@netflix.com
artifacts:
- name: "@mycompany/myapp"
  type: npm
  reference: pr
  from:
    pullRequestsOnly: true
```

```yaml
# Artifacts from pull requests whose branches start with "feature-"
name: myapp
application: myapp
serviceAccount: myapp@netflix.com
artifacts:
- name: "@mycompany/myapp"
  type: npm
  reference: feature-prs
  from:
    branch:
      startsWith: feature-
    pullRequestsOnly: true
```

More implementation details in inline comments. I recommend viewing the diff ignoring whitespace.

Closes #1480 

Cc: @erikmunson 

TO-DO:
- [x] Allow using regular expressions for branches
- [x] Allow specifying artifacts originated from PRs
- [x] Remove `sortBy` field in favor of auto-detecting sort strategy based on presence of other fields

TO-DO (future PRs):
- [ ] (Maybe) add generated `branch` and `pr_number` columns to the `artifact_versions` table so we can avoid querying into the JSON field and also add an index 
- [x] Add support for a "startsWith" branch spec, and make regex an explicit advanced option

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
@@ -1,5 +1,6 @@
 package com.netflix.spinnaker.keel.api.artifacts
 
+import com.netflix.spinnaker.keel.api.artifacts.SortStrategy.BRANCH_AND_TIMESTAMP
 import com.netflix.spinnaker.keel.api.artifacts.SortStrategy.VERSION
 import com.netflix.spinnaker.keel.api.schema.Discriminator
 import java.time.Instant
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
@@ -1,5 +1,6 @@
 package com.netflix.spinnaker.keel.api.artifacts
 
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactSortByMethod.VERSION
 import com.netflix.spinnaker.keel.api.schema.Discriminator
 import java.time.Instant
 
@@ -18,6 +19,14 @@ enum class ArtifactStatus {
 }
 
 /**
+ * The ways in which artifacts can be sorted.
+ */
+enum class ArtifactSortByMethod(val type: String) {
+  VERSION("version"),
+  BRANCH_AND_TIMESTAMP("branch-and-timestamp")
+}
+
+/**
  * An artifact as defined in a [DeliveryConfig].
  *
  * Unlike other places within Spinnaker, this class does not describe a specific instance of a software artifact
@@ -32,6 +41,8 @@ abstract class DeliveryArtifact {
   abstract val reference: String // friendly reference to use within a delivery config
   abstract val deliveryConfigName: String? // the delivery config this artifact is a part of
   open val statuses: Set<ArtifactStatus> = emptySet()
+  open val branch: String? = null
+  open val sortBy: ArtifactSortByMethod = VERSION
 
   fun toArtifactInstance(version: String, status: ArtifactStatus? = null, createdAt: Instant? = null) =
     PublishedArtifact(
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
@@ -1,7 +1,5 @@
 package com.netflix.spinnaker.keel.api.artifacts
 
-import com.netflix.spinnaker.keel.api.artifacts.SortStrategy.BRANCH_AND_TIMESTAMP
-import com.netflix.spinnaker.keel.api.artifacts.SortStrategy.VERSION
 import com.netflix.spinnaker.keel.api.schema.Discriminator
 import java.time.Instant
 
@@ -20,14 +18,6 @@ enum class ArtifactStatus {
 }
 
 /**
- * The ways in which artifacts can be sorted.
- */
-enum class SortStrategy(val type: String) {
-  VERSION("version"),
-  BRANCH_AND_TIMESTAMP("branch-and-timestamp")
-}
-
-/**
  * An artifact as defined in a [DeliveryConfig].
  *
  * Unlike other places within Spinnaker, this class does not describe a specific instance of a software artifact
@@ -39,11 +29,25 @@ abstract class DeliveryArtifact {
   abstract val name: String
   @Discriminator abstract val type: ArtifactType
   abstract val versioningStrategy: VersioningStrategy
-  abstract val reference: String // friendly reference to use within a delivery config
-  abstract val deliveryConfigName: String? // the delivery config this artifact is a part of
+  /** A friendly reference to use within a delivery config. */
+  abstract val reference: String
+  /** The delivery config this artifact is a part of. */
+  abstract val deliveryConfigName: String?
+  /** A set of release statuses to filter by. Mutually exclusive with [fromBranch] and [fromPullRequest]. */
   open val statuses: Set<ArtifactStatus> = emptySet()
-  open val branch: String? = null
-  open val sortBy: SortStrategy = VERSION
+  /** A specific branch to filter by, or a regular expression like "feature-.*" */
+  open val fromBranch: String? = null
+  /** Whether to include only artifacts from pull requests. */
+  open val fromPullRequest: Boolean? = false
+
+  val filteredByBranch: Boolean
+    get() = fromBranch != null
+
+  val filteredByPullRequest: Boolean
+    get() = fromPullRequest == true
+
+  val filteredByReleaseStatus: Boolean
+    get() = statuses.isNotEmpty()
 
   fun toArtifactInstance(version: String, status: ArtifactStatus? = null, createdAt: Instant? = null) =
     PublishedArtifact(
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
@@ -1,6 +1,6 @@
 package com.netflix.spinnaker.keel.api.artifacts
 
-import com.netflix.spinnaker.keel.api.artifacts.ArtifactSortByMethod.VERSION
+import com.netflix.spinnaker.keel.api.artifacts.SortStrategy.VERSION
 import com.netflix.spinnaker.keel.api.schema.Discriminator
 import java.time.Instant
 
@@ -21,7 +21,7 @@ enum class ArtifactStatus {
 /**
  * The ways in which artifacts can be sorted.
  */
-enum class ArtifactSortByMethod(val type: String) {
+enum class SortStrategy(val type: String) {
   VERSION("version"),
   BRANCH_AND_TIMESTAMP("branch-and-timestamp")
 }
@@ -42,7 +42,7 @@ abstract class DeliveryArtifact {
   abstract val deliveryConfigName: String? // the delivery config this artifact is a part of
   open val statuses: Set<ArtifactStatus> = emptySet()
   open val branch: String? = null
-  open val sortBy: ArtifactSortByMethod = VERSION
+  open val sortBy: SortStrategy = VERSION
 
   fun toArtifactInstance(version: String, status: ArtifactStatus? = null, createdAt: Instant? = null) =
     PublishedArtifact(
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
@@ -18,6 +18,40 @@ enum class ArtifactStatus {
 }
 
 /**
+ * Filters for source code branches. The fields within this class are mutually-exclusive.
+ *
+ * @param name A specific branch to match against.
+ * @param startsWith Match branches starting with this string.
+ * @param regex A regular expression to match against (e.g. "feature-.*").
+ */
+data class BranchFilterSpec(
+  val name: String? = null,
+  val startsWith: String? = null,
+  val regex: String? = null
+) {
+  init {
+    require(
+      (name != null && startsWith == null && regex == null)
+        || (name == null && startsWith != null && regex == null)
+        || (name == null && startsWith == null && regex != null)
+    ) {
+      "Please specify only one of 'name', 'startsWith' or 'regex'."
+    }
+  }
+}
+
+/**
+ * Filters for the origin of an artifact in source control.
+ *
+ * @param branch A [BranchFilterSpec] with branch filters.
+ * @param pullRequestOnly Whether to include only artifacts built from pull requests.
+ */
+data class ArtifactOriginFilterSpec(
+  val branch: BranchFilterSpec? = null,
+  val pullRequestOnly: Boolean? = false
+)
+
+/**
  * An artifact as defined in a [DeliveryConfig].
  *
  * Unlike other places within Spinnaker, this class does not describe a specific instance of a software artifact
@@ -33,18 +67,16 @@ abstract class DeliveryArtifact {
   abstract val reference: String
   /** The delivery config this artifact is a part of. */
   abstract val deliveryConfigName: String?
-  /** A set of release statuses to filter by. Mutually exclusive with [fromBranch] and [fromPullRequest]. */
+  /** A set of release statuses to filter by. Mutually exclusive with [from] filters. */
   open val statuses: Set<ArtifactStatus> = emptySet()
-  /** A specific branch to filter by, or a regular expression like "feature-.*" */
-  open val fromBranch: String? = null
-  /** Whether to include only artifacts from pull requests. */
-  open val fromPullRequest: Boolean? = false
+  /** Filters for the artifact origin in source control. */
+  open val from: ArtifactOriginFilterSpec? = null
 
   val filteredByBranch: Boolean
-    get() = fromBranch != null
+    get() = from?.branch != null
 
   val filteredByPullRequest: Boolean
-    get() = fromPullRequest == true
+    get() = from?.pullRequestOnly == true
 
   val filteredByReleaseStatus: Boolean
     get() = statuses.isNotEmpty()
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
@@ -3,6 +3,7 @@ package com.netflix.spinnaker.keel.persistence
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactMetadata
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactSortByMethod.BRANCH_AND_TIMESTAMP
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactStatus.FINAL
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactStatus.RELEASE
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactStatus.SNAPSHOT
@@ -27,6 +28,7 @@ import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactPin
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVeto
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVetoes
 import com.netflix.spinnaker.keel.core.api.PinnedEnvironment
+import com.netflix.spinnaker.keel.exceptions.InvalidArtifactSpecException
 import com.netflix.spinnaker.time.MutableClock
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
@@ -62,7 +64,7 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
     val subject: T
   ) {
     // the artifact built off a feature branch
-    val artifact1 = DebianArtifact(
+    val versionedSnapshotDebian = DebianArtifact(
       name = "keeldemo",
       deliveryConfigName = "my-manifest",
       reference = "candidate",
@@ -71,27 +73,39 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
     )
 
     // the artifact built off of master
-    val artifact2 = DebianArtifact(
+    val versionedReleaseDebian = DebianArtifact(
       name = "keeldemo",
       deliveryConfigName = "my-manifest",
       reference = "master",
       vmOptions = VirtualMachineOptions(baseOs = "bionic", regions = setOf("us-west-2")),
       statuses = setOf(RELEASE)
     )
-    val artifact3 = DockerArtifact(
+    
+    val versionedDockerArtifact = DockerArtifact(
       name = "docker",
       deliveryConfigName = "my-manifest",
       reference = "docker-artifact",
       tagVersionStrategy = BRANCH_JOB_COMMIT_BY_JOB
     )
-    val environment1 = Environment("test")
-    val environment2 = Environment("staging")
+
+    val debianWithSortingByTimestamp =  DebianArtifact(
+      name = "keeldemo",
+      deliveryConfigName = "my-manifest",
+      reference = "feature-branch",
+      vmOptions = VirtualMachineOptions(baseOs = "bionic", regions = setOf("us-west-2")),
+      sortBy = BRANCH_AND_TIMESTAMP,
+      branch = "my-feature-branch"
+    )
+
+    val featureBranchEnvironment = Environment("feature-branch")
+    val testEnvironment = Environment("test")
+    val stagingEnvironment = Environment("staging")
     val manifest = DeliveryConfig(
       name = "my-manifest",
       application = "fnord",
       serviceAccount = "keel@spinnaker",
-      artifacts = setOf(artifact1, artifact2, artifact3),
-      environments = setOf(environment1, environment2)
+      artifacts = setOf(versionedSnapshotDebian, versionedReleaseDebian, versionedDockerArtifact, debianWithSortingByTimestamp),
+      environments = setOf(featureBranchEnvironment, testEnvironment, stagingEnvironment)
     )
     val version1 = "keeldemo-0.0.1~dev.8-h8.41595c4" // snapshot
     val version2 = "keeldemo-0.0.1~dev.9-h9.3d2c8ff" // snapshot
@@ -103,8 +117,8 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
     val versionOnly = "0.0.1~dev.8-h8.41595c4"
 
     val pin1 = EnvironmentArtifactPin(
-      targetEnvironment = environment2.name, // staging
-      reference = artifact2.reference,
+      targetEnvironment = stagingEnvironment.name, // staging
+      reference = versionedReleaseDebian.reference,
       version = version4, // the older release build
       pinnedBy = "keel@spinnaker",
       comment = "fnord"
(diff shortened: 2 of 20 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -32,6 +41,8 @@ abstract class DeliveryArtifact {
   abstract val reference: String // friendly reference to use within a delivery config
   abstract val deliveryConfigName: String? // the delivery config this artifact is a part of
   open val statuses: Set<ArtifactStatus> = emptySet()
+  open val branch: String? = null
+  open val sortBy: ArtifactSortByMethod = VERSION
```
luispollo, line 45: Default sorting method is still by version for backwards compatibility.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -221,31 +231,53 @@ class SqlArtifactRepository(
...
+        .select(ARTIFACT_VERSIONS.VERSION, ARTIFACT_VERSIONS.RELEASE_STATUS)
+        .from(ARTIFACT_VERSIONS)
+        .where(ARTIFACT_VERSIONS.NAME.eq(artifact.name))
+        .and(ARTIFACT_VERSIONS.TYPE.eq(artifact.type))
+        .apply {
+          if (artifact.statuses.isNotEmpty()) {
+            and(ARTIFACT_VERSIONS.RELEASE_STATUS.`in`(*artifact.statuses.map { it.toString() }.toTypedArray()))
+          }
+
+          // When sorting "natively" by branch and timestamp, delegate sorting and limiting to the database
```
luispollo, line 256: Just calling attention to this comment.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -17,6 +18,14 @@ enum class ArtifactStatus {
   FINAL, CANDIDATE, SNAPSHOT, RELEASE, UNKNOWN
 }
 
+/**
+ * The ways in which artifacts can be sorted.
+ */
+enum class ArtifactSortByMethod(val type: String) {
```
luispollo, line 24: I couldn't come up with a better name for this enum... `ArtifactSortingMethod` didn't sound right either. Let me know if you have any suggestions.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -70,6 +74,12 @@ class SqlArtifactRepository(
   private val sqlRetry: SqlRetry,
   private val artifactSuppliers: List<ArtifactSupplier<*, *>> = emptyList()
 ) : ArtifactRepository {
+
+  companion object {
+    private val ARTIFACT_VERSIONS_BRANCH =
+      field<String?>("keel.artifact_versions.git_metadata->'$.branch'")
```
luispollo, line 80: @robfletcher I know we ran into performance issues with JSON queries before. Should we look into creating a new view for artifacts like we did for resources?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -71,27 +73,39 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
...
     )
-    val environment1 = Environment("test")
-    val environment2 = Environment("staging")
+
+    val debianWithSortingByTimestamp =  DebianArtifact(
+      name = "keeldemo",
+      deliveryConfigName = "my-manifest",
+      reference = "feature-branch",
+      vmOptions = VirtualMachineOptions(baseOs = "bionic", regions = setOf("us-west-2")),
+      sortBy = BRANCH_AND_TIMESTAMP,
```
luispollo, line 96: This is how you say you want sorting by timestamps.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
lorin: Suggestion: `ArtifactSortStrategy` or even just `SortStrategy`, since this is in the `artifacts` module already.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -32,6 +41,8 @@ abstract class DeliveryArtifact {
   abstract val reference: String // friendly reference to use within a delivery config
   abstract val deliveryConfigName: String? // the delivery config this artifact is a part of
   open val statuses: Set<ArtifactStatus> = emptySet()
+  open val branch: String? = null
```
lorin, line 44: When the `BRANCH_AND_TIMESTAMP` sort method is used, is an artifact always associated with exactly one branch? Does that mean you need a separate artifact in the delivery config if you want to deploy a feature branch for testing?

(Not criticizing this approach, just asking for clarification. I don't actually know how deploying feature branches works currently).

lorin (approved):
LGTM

Are there any scenarios where the `createdAt` timestamp could be inconsistent with the ordering that the user expects?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
emjburns: Yes, and it is always true that you need to define a new artifact if you want to deploy a new feature branch.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
emjburns: I like `SortStrategy`

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -221,31 +231,53 @@ class SqlArtifactRepository(
...
-        .sortedWith(artifact.versioningStrategy.comparator)
-        .also { versions ->
-          // FIXME: remove special handling for Docker
-          // FIXME: limit the sql query, not the returned list (once we'll have a native sorting mechanism)
-          return if (artifact is DockerArtifact) {
-            filterDockerVersions(artifact, versions, limit)
-          } else {
-            versions.subList(0, Math.min(versions.size, limit))
+  override fun versions(artifact: DeliveryArtifact, limit: Int): List<String> {
+    // sanity checks
```
emjburns, line 235: nit: this comment isn't needed, it's very clear that this is a validation check

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: What area of performance are you concerned about?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -221,31 +231,53 @@ class SqlArtifactRepository(
...
+    if (!isRegistered(artifact.name, artifact.type)) {
+      throw NoSuchArtifactException(artifact)
+    }
+
+    if (artifact.sortBy == BRANCH_AND_TIMESTAMP && artifact.branch == null) {
+      throw InvalidArtifactSpecException(
+        "The ${artifact.type} artifact ${artifact.name} specifies sorting by branch and timestamp, but no branch is defined in the config.")
```
emjburns, line 242:
```suggestion
        "Artifact ${artifact.reference} missing required branch for branch and timestamp sorting: ($artifact)")
```
I think using reference is better, and I like adding the config in the message so that it's clear what the current config is (to help with parsing issues/etc)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -221,31 +231,53 @@ class SqlArtifactRepository(
...
+        .apply {
+          if (artifact.statuses.isNotEmpty()) {
+            and(ARTIFACT_VERSIONS.RELEASE_STATUS.`in`(*artifact.statuses.map { it.toString() }.toTypedArray()))
+          }
+
+          // When sorting "natively" by branch and timestamp, delegate sorting and limiting to the database
+          if (artifact.sortBy == BRANCH_AND_TIMESTAMP) {
+            // TODO: should we also be comparing the repo with what's configured for the app in front50?
+            and(ARTIFACT_VERSIONS_BRANCH.eq(artifact.branch))
+              .and(ARTIFACT_VERSIONS.CREATED_AT.isNotNull)
```
emjburns, line 260: so here, we are excluding versions which don't have a created at time yet, right? I'd love if there was some way to know that we are excluding some here, but it doesn't seem possible to add a log message telling us that. I'd like us to easily know that a version exists, but doesn't have enough metadata to be used in case a user complains of slowness

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -32,6 +41,8 @@ abstract class DeliveryArtifact {
...
+  open val sortBy: ArtifactSortByMethod = VERSION
```
emjburns, line 46: In order to validate when branch-and-timestamp is specified, you can do something like: https://github.com/spinnaker/keel/blob/master/keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ClusterSpec.kt#L201-L211

it's super easy

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: The last time I added a query for a JSON field, it ended up slowing queries down, which @robfletcher caught and corrected here: https://github.com/spinnaker/keel/pull/1378

This is why I pinged him. I should have looked for that PR earlier. My specific question is: should I add a generated column and an index, like we did for the `events` table, as part of this PR (or a follow-up)?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
luispollo: That's right. Here's an example with release statuses when sorting by versions:
```yaml
artifacts:
- name: "@mycompany/myapp"
  reference: feature-branch
  type: npm
  statuses:
    - SNAPSHOT
- name: "@mycompany/myapp"
  reference: master
  type: npm
  statuses:
    - RELEASE
```
With branch and timestamp:
```yaml
artifacts:
- name: "@mycompany/myapp"
  reference: feature-branch
  type: npm
  branch: my-feature-branch
  sortBy: branch-and-timestamp
- name: "@mycompany/myapp"
  reference: master
  type: npm
  branch: master
  sortBy: branch-and-timestamp
```

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
luispollo: Nice! Will try that.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: Yeah, I wanted to print the delivery config name but wasn't 100% positive it would be (always) filled.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: I mean... We could add a separate query in this function to check for that and log, if we think it's worth the extra cost. Thoughts?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
gal-yardeni: shouldn't `branch` is enough? is there another way, other than `timestamp` to sort a branch's artifacts?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
gal-yardeni: back to my question above - do we even need to specify the `sortBy?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/KeelApiModule.kt
```diff
@@ -63,6 +65,7 @@ object KeelApiModule : SimpleModule("Keel API") {
       setMixInAnnotations<SubnetAwareRegionSpec, SubnetAwareRegionSpecMixin>()
       setMixInAnnotations<Resource<*>, ResourceMixin>()
       setMixInAnnotations<ResourceSpec, ResourceSpecMixin>()
+      setMixInAnnotations<ArtifactSortByMethod, ArtifactSortByMethodMixin>()
```
gal-yardeni, line 68: for my knowledge, what is this doing?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: good call Emily.. another potential way to test it out would be to count the # of versions which returning from this call and compare it to the limit to see the differences if any.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
luispollo: My thinking was that, especially at this point, where this is very new, it was important to make the selection explicit. I've also had OSS in mind: they won't have support for sorting by branch and timestamp out of the box.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/KeelApiModule.kt
luispollo: Adding the mixin to customize Jackson serialization of the enum, so that we can use use lower-case values.

luispollo:
> Are there any scenarios where the `createdAt` timestamp could be inconsistent with the ordering that the user expects?

That's a good question that's hard for me to answer at this point. I think the answer will come as we get users using this option and provide feedback. I know we've already received some internal feedback that there might be corner cases this method does not address. Happy to share that with you privately.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
luispollo: @emjburns I tried that, but ran into issues with class initialization order: https://kotlinlang.org/docs/reference/classes.html#derived-class-initialization-order

I spent some time trying to get it to work, but it felt brittle and so I'd rather add a validator separately in a follow-up PR. Hope that's OK.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
luispollo: See response to previous comment.

erikmunson:
VERY EXCITING. This is awesome. I have a couple thoughts on the spec design/experience:

1. Is there a functional requirement for coupling together the sort (timestamp) with the filtering (branch)? Based on [the original use cases for source-based filtering](https://github.com/spinnaker/keel/issues/1480) we mapped out, encoding a specific filtering strategy into the sort configuration seems like it will become brittle/misnamed quickly, and seeing it in the name likely also gives an implicit signal that branch is the only thing you can filter on with that option (vs. PRs, etc)
2. Given the additional filtering options covered in #1480, I encourage you to think about how to model the `branch` part of the spec in a way that leaves you future room to add a 'branch starts with xyz' and 'is from a pull request' option. That may mean keeping things as-is, but it may not. I often hop into my editor and sketch out what having all those options available would look like to understand the end state I'm going after so all the pieces fit together nicely when I'm done
3. I understand and support making this opt-in for this initial PR while we try it out. At the same time, keep in mind that if this experiment is looking successful then our customers will want (and need) it to be the default so they don't have to fight against us to get their app working the way they want. I suggest coming up with a plan now for how you'd make this the default (at least at Netflix) should it prove successful so you can make decisions now that work with that instead of against it. Backwards compatibility is great, but is primarily a means to an end (the end being "not breaking people")

luispollo:
@erikmunson Thanks for the reality check. Very much appreciated. 🙏 

This is what happens when I try to multitask... 😞 I swear I read the issue a couple times before starting, and had all those other requirements in mind, but as I started working on it and switched context a couple of times, they totally faded away. I've put this back in a draft state and will try and come up with a more sensible model.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
emjburns: eww yuck. that's fine

luispollo: The good news is we don't need any validation anymore as I'm now auto-detecting the strategy. 🙂

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -221,31 +231,54 @@ class SqlArtifactRepository(
...
+              and(ARTIFACT_VERSIONS_PR_NUMBER.isNotNull).and(ARTIFACT_VERSIONS_PR_NUMBER.ne(EMPTY_PR_NUMBER))
+            }
+
+            if (artifact.filteredByBranch) {
+              // TODO: should we also be comparing the repo with what's configured for the app in front50?
+              and(ARTIFACT_VERSIONS_BRANCH.likeRegex(artifact.fromBranch))
+            }
+
+            // With branches or pull requests, delegate sorting and limiting to the database
+            if (artifact.filteredByPullRequest || artifact.filteredByBranch) {
```
luispollo, line 259: This check is needed here because apparently we are passing an empty set of statuses to indicate "ignore statuses" (there are tests for that which break if this check is not here).

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -29,9 +29,25 @@ abstract class DeliveryArtifact {
   abstract val name: String
   @Discriminator abstract val type: ArtifactType
   abstract val versioningStrategy: VersioningStrategy
-  abstract val reference: String // friendly reference to use within a delivery config
-  abstract val deliveryConfigName: String? // the delivery config this artifact is a part of
```
emjburns, line 33: These comments would be better as kdoc comments and not inline

gal-yardeni:
I love it so much better without the sortingWith field! @erikmunson comment was spot on!

luispollo:
> I love it so much better without the sortingWith field! @erikmunson comment was spot on!

Do you love it enough to approve it? 😄

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
luispollo: Uhh, that's exactly what I did on the right side of the diff? 🤔 😄

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: @emjburns I'm planning to follow up with a PR to address https://github.com/spinnaker/keel/issues/1565 so we get metrics when we're missing artifact metadata.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -221,31 +230,54 @@ class SqlArtifactRepository(
...
+              and(ARTIFACT_VERSIONS.CREATED_AT.isNotNull)
+                .orderBy(ARTIFACT_VERSIONS.CREATED_AT.desc())
```
gal-yardeni, line 260: ❤️  for default sorting + limit

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: I find all of these `if` conditions a bit confusing.. I worry that someone else that will see it can mess it up pretty quickly 🤷

erikmunson:
Spec looks awesome! I must admit I didn't even _think_ of automatically detecting the sort type, that's a cool idea and I bet it will work really well.

I have one nit: is there a way we can satisfy the 'starts with feature-' use case without pulling our regular expression toolbox out of the garage? Based on what I've heard so far from customers I believe that a prefix is the only common case and it'd be cool to support that without going all-in on regex. Could we bake that in as an option? Maybe as a separate spec field (e.g. `fromBranchStartingWith`)?

Looking forward to this so much!

gal-yardeni (approved):
LGTM overall! We need really good docs around it so people could understand exactly their options here, but we have time to worry about it :)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: Any suggestions?

luispollo:
> I have one nit: is there a way we can satisfy the 'starts with feature-' use case without pulling our regular expression toolbox out of the garage? Based on what I've heard so far from customers I believe that a prefix is the only common case and it'd be cool to support that without going all-in on regex. Could we bake that in as an option? Maybe as a separate spec field (e.g. `fromBranchStartingWith`)?

Hey @erikmunson. How about we add an optional prefix to the field value, like so?

Examples:
```yaml
# Matches anything starting with "feature-"
name: myapp
application: myapp
serviceAccount: myapp@netflix.com
artifacts:
- name: "@mycompany/myapp"
  type: npm
  reference: feature-branches
  fromBranch: "startsWith: feature-"
```
Regexp version for advanced users:
```yaml
# Matches anything starting with "feature-"
name: myapp
application: myapp
serviceAccount: myapp@netflix.com
artifacts:
- name: "@mycompany/myapp"
  type: npm
  reference: feature-branches
  fromBranch: "regex: feature-.*"
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: not really :(

erikmunson:
@luispollo I think the naming/options in those snippets makes sense. We've generally tried to avoid encoding structured data into strings in specs (unfortunately we had to compromise on that for `kind` :/), is it possible to take that exact same naming scheme and make `fromBranch` accept either a string _or_ an object that has a `startsWith` / `regex` field?

luispollo:
@erikmunson I thought about that as well but was trying to avoid the extra work of writing custom (de-)serialization for the object. How about we get this out there to test and I work on that in a separate PR?

erikmunson:
I think if the plan is to test it out ourselves and give interested customers a chance to try the simpler configuration option (i.e. just branch name) that's great — I'd be concerned about giving those more advanced options to customers ahead of the UX getting sorted out

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
emjburns: OH I mean at the top of the method like with 
/**
* @param reference 
*/

emjburns: wow formatting there did not work

Just a style preference. But I prefer docs outside the method if there is more than one or two comments

luispollo: But this is not a method, it's the body of the class...

luispollo:
> I think if the plan is to test it out ourselves and give interested customers a chance to try the simpler configuration option (i.e. just branch name) that's great — I'd be concerned about giving those more advanced options to customers ahead of the UX getting sorted out

@erikmunson Absolutely. I wasn't expecting us to roll this out to anyone until we've had a chance to dog-food it ourselves and tweak whatever needs tweaking. We both know we won't get it all right the first time around. I plan to quickly follow up with additional PRs as things pop up.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
emjburns: 🤦  everything is great over here Luis. I am doing great. I'm going to blame this one on wishing github would show me a couple more surrounding lines (but obviously only when I want them) (but also without clicking that, cause how could I know I wanted them)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
emjburns: I think these could be good candidates for a new column in the `artifact_versions` table. It seems like we might be querying for those things, so they should probably be in their own separate field. I do think that adding the JSON field here will slow things down

luispollo: Sounds good. I'll add the generated columns in a follow-up PR.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -70,6 +70,15 @@ class SqlArtifactRepository(
   private val sqlRetry: SqlRetry,
   private val artifactSuppliers: List<ArtifactSupplier<*, *>> = emptyList()
 ) : ArtifactRepository {
+
```
emjburns, line 73: You're going to add columns for these in another PR, right?

luispollo: Yup! Generated columns so there's no risk of things getting out-of-sync.

emjburns: ah yes, i see your earlier comment now

emjburns (approved):

luispollo:
Updated samples in the description with the format we agreed upon in offline discussion.

gal-yardeni:
@luispollo It's so great that this PR is pretty small! I have a question - if you just want to say `branch: master`, what is the right configuration for it?

luispollo:
> @luispollo It's so great that this PR is pretty small! I have a question - if you just want to say `branch: master`, what is the right configuration for it?

That's the first example in the PR description above. Here it is again:
```yaml
# Artifacts from the "master" branch
name: myapp
application: myapp
serviceAccount: myapp@netflix.com
artifacts:
- name: "@mycompany/myapp"
  type: npm
  reference: master
  from:
    branch:
      name: master
```

PR #1566, opened 2020-10-05
feature(ec2): Add support for launch templates

lorin:
## Summary

Add support for managing EC2 apps that have been configured to use AWS launch templates.

## Background

Clouddriver now support AWS [launch templates](https://docs.aws.amazon.com/autoscaling/ec2/userguide/LaunchTemplates.html). Clouddriver opts in apps to use launch templates, on a per-app basis.

## What changed from keel's perspective

For apps that are configured to use launch templates, the response payload from clouddriver changes when querying for server group information. For these apps, there is no longer a `launchConfig` field. Instead, there is a `launchTemplate` field. The `launchTemplate` field contains similar information, but the data has a different shape.

## PR details

With this PR, we can now handle response payloads from clouddriver where there is a `launchTemplate` field instead of a `launchConfig` field.

Note that this PR does not change:

* the delivery config schema (the section called `launchConfiguration` remains unchanged)
* the shape of the data that we send to orca when actuating EC2 clusters

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
@@ -105,10 +105,8 @@ data class ServerGroup(
  * It is intended to be called from the init block of classes that implement BaseEc2ServerGroup
  */
 fun ensureLaunchConfigInfoIsPresent(sg : BaseEc2ServerGroup) {
-  sg.run {
-    if(launchConfig == null && launchTemplate == null) {
-      throw SystemException("Server group info contains neither launchConfig nor launchTemplate fields: ${sg.name}")
-    }
+  if(sg.launchConfig == null && sg.launchTemplate == null) {
+    throw SystemException("Server group info contains neither launchConfig nor launchTemplate fields: ${sg.name}")
   }
 }
 
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
@@ -44,7 +44,20 @@ data class InstanceCounts(
 interface BaseEc2ServerGroup : BaseServerGroup {
   val zones: Set<String>
   val image: ActiveServerGroupImage
-  val launchConfig: LaunchConfig
+
+  /**
+   * One of these two fields is always null, and the other is always
+   * non-null.
+   *
+   * If AWS launch templates are enabled in clouddriver for this app,
+   * associated with this server group, then the launchTemplate
+   * field will be populated.
+   *
+   * Otherwise, the launchConfig field will be populated.
+   */
+  val launchConfig: LaunchConfig?
+  val launchTemplate: LaunchTemplate?
+
   val asg: AutoScalingGroup
   val scalingPolicies: List<ScalingPolicy>
   val vpcId: String
@@ -63,7 +76,8 @@ data class ServerGroup(
   override val region: String,
   override val zones: Set<String>,
   override val image: ActiveServerGroupImage,
-  override val launchConfig: LaunchConfig,
+  override val launchConfig: LaunchConfig? = null,
+  override val launchTemplate: LaunchTemplate? = null,
   override val asg: AutoScalingGroup,
   override val scalingPolicies: List<ScalingPolicy>,
   override val vpcId: String,
@@ -77,7 +91,26 @@ data class ServerGroup(
   override val disabled: Boolean,
   override val instanceCounts: InstanceCounts,
   override val createdTime: Long
-) : BaseEc2ServerGroup
+) : BaseEc2ServerGroup {
+  init {
+    ensureLaunchConfigInfoIsPresent(this)
+  }
+}
+
+/**
+ * Clouddriver should always return either a launchConfig field or a launchTemplate
+ * field for an EC2 server group.
+ *
+ * This is a helper function that asserts that one of these fields must be present.
+ * It is intended to be called from the init block of classes that implement BaseEc2ServerGroup
+ */
+fun ensureLaunchConfigInfoIsPresent(sg : BaseEc2ServerGroup) {
+  sg.run {
+    if(launchConfig == null && launchTemplate == null) {
+      throw SystemException("Server group info contains neither launchConfig nor launchTemplate fields: ${sg.name}")
+    }
+  }
+}
 
 fun ServerGroup.toActive(accountName: String) =
   ActiveServerGroup(
@@ -86,6 +119,7 @@ fun ServerGroup.toActive(accountName: String) =
     zones = zones,
     image = image,
     launchConfig = launchConfig,
+    launchTemplate = launchTemplate,
     asg = asg,
     scalingPolicies = scalingPolicies,
     vpcId = vpcId,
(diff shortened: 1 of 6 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -173,6 +212,23 @@ data class LaunchConfig(
   val instanceMonitoring: InstanceMonitoring
 )
 
+data class LaunchTemplate(
+  val launchTemplateData: LaunchTemplateData
+)
+
+data class LaunchTemplateData(
```
caseyhebebrand, line 219: I am not sure what other `LaunchTemplateData` keel cares about, but here is a [link ](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata.html)to the other properties just in case.

lorin: I wasn't sure whether to populate all of the fields or just the ones we are crrently consuming, so I started with just the ones we are currently consuming.

Also, I see it supports ram disk id! I didn't see that in the payloads that I was playing with, I'm going to add that in for parity.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -77,7 +91,26 @@ data class ServerGroup(
...
+
+/**
+ * Clouddriver should always return either a launchConfig field or a launchTemplate
+ * field for an EC2 server group.
+ *
+ * This is a helper function that asserts that one of these fields must be present.
+ * It is intended to be called from the init block of classes that implement BaseEc2ServerGroup
+ */
+fun ensureLaunchConfigInfoIsPresent(sg : BaseEc2ServerGroup) {
+  sg.run {
```
gal-yardeni, line 108: why do we need here `sb.run`?

lorin: To be honest, I don't remember why I put it there to begin with, but we don't need it. I removed it.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -77,7 +91,24 @@ data class ServerGroup(
...
+
+/**
+ * Clouddriver should always return either a launchConfig field or a launchTemplate
+ * field for an EC2 server group.
+ *
+ * This is a helper function that asserts that one of these fields must be present.
+ * It is intended to be called from the init block of classes that implement BaseEc2ServerGroup
+ */
+fun ensureLaunchConfigInfoIsPresent(sg : BaseEc2ServerGroup) {
+  if(sg.launchConfig == null && sg.launchTemplate == null) {
```
gal-yardeni, line 108: what will happen if both are not null? does clouddriver guarantees that only one of those exists?

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -173,6 +210,25 @@ data class LaunchConfig(
   val instanceMonitoring: InstanceMonitoring
 )
 
+data class LaunchTemplate(
```
gal-yardeni, line 213: question: can `LaunchTemplateData` be at the top level, meaning just move all of its fields to `LaunchTemplate`?

gal-yardeni (approved):
minor comments, but lgtm overall!

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
lorin: That's a good question. @caseyhebebrand or @jeyrschabu, does clouddriver guarantee that only one of `launchConfig` and `launchTemplate` will be returned for an EC2 server group?

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
lorin: The challenge is the shape of the clouddriver response is:

```json
{
 "launchTemplate": {
   "launchTemplateData": {

   }
  }
}
```

While we probably could write some Jackson-related code so that these classes don't have to exactly match the shape of the response, my inclination is to keep the Jackson mapping logic as simple as possible.

gal-yardeni: oh I see, makes sense!

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
caseyhebebrand: Jeyrs is the best person to answer this, but there shouldn't be case where both are present. The main reason being that a server group can only have one or the other in AWS, and we never have a state where a server group is truly modified without creating a new launch config/template.

caseyhebebrand (approved):

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
jeyrschabu: Both `launchConfig` and `launchTemplate` (top level props on a server group) are eventually populated as we index them. Important to note that a server group can't have both.

If a server group doesn't have a launch configuration, `launchConfigurationName` will be `null`.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
jeyrschabu: This models the `aws` `launchTemplateData`,  are the other properties dropped? 
I guess you only need properties that you are specifically accessing on the object.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -954,19 +955,20 @@ class ClusterHandler(
         subnet = subnet(cloudDriverCache),
         availabilityZones = zones
       ),
-      launchConfiguration = launchConfig.run {
+      launchConfiguration =
         LaunchConfiguration(
-          imageId = imageId,
+          imageId = launchConfig?.imageId ?: launchTemplateData!!.imageId,
```
jeyrschabu, line 960: There is a top-level `Image` field on a server group. You could look the image id that way and avoid the conditional here.

lorin: Fixed in bf496872

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
lorin: @jeyrschabu Yeah, this initial implementation just parses out the data that keel consumes, and ignores the other fields.

PR #1577, opened 2020-10-07
fix(heath): consider 0 instances unhealthy when you've asked for > 0

emjburns:
my titus cluster was "healthy" even though it had 0 instances (I asked for one). That's not healthy.


lorin (approved):
LGTM

PR #1582, opened 2020-10-08
feat(dockerData): adding branch + date information to docker images

gal-yardeni:


keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
```diff
@@ -29,10 +30,28 @@ class DockerArtifactSupplier(
...
   override val supportedVersioningStrategy =
     SupportedVersioningStrategy("docker", DockerVersioningStrategy::class.java)
 
+  override fun publishArtifact(artifact: PublishedArtifact) {
+    if (artifact.hasDate()) {
+      super.publishArtifact(artifact.copy(
+        //changing date to createdAt, to match other artifact types
+        metadata = artifact.metadata.toMutableMap().also {
+         it["createdAt"] = artifact.metadata["date"]
+          it.remove("date")
```
luispollo, line 45: I think you can simplify this with:
```suggestion
         it["createdAt"] = it.remove("date")
```

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
```diff
@@ -29,10 +30,28 @@ class DockerArtifactSupplier(
...
+        }
+        ))
+
+    } else {
+      log.debug("Docker artifact: ${artifact.name} does not contains a date.")
```
luispollo, line 50:
```suggestion
      log.warn("Docker artifact: ${artifact.name} does not contains a date.")
```

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
luispollo: You know, I was just thinking... Instead of renaming the map key here, we could just consider that key as a fall-back in `PublishedArtifact.createdAt`. That way you only have a single place to change, and don't have to change the artifact supplier!

Like so:
```kotlin
  val createdAt = metadata["createdAt"]
    ?: metadata["date"]
    ?.let {
      when (it) {
        is Long -> Instant.ofEpochMilli(it) // to accommodate for artifact events from CI integration
        is Instant -> it
        else -> null
      }
    }
```

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
gal-yardeni: done

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
gal-yardeni: done

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
```diff
@@ -52,7 +52,14 @@ data class PublishedArtifact(
...
 
-  val createdAt = metadata["createdAt"]
+  val createdAt = let {
+    if (metadata.containsKey("createdAt") && metadata["createdAt"] != null)
+      metadata["createdAt"]
+    else {
+      //docker artifact createdAt time is under date field
+      metadata["date"]
+    }
+  }
```
luispollo, line 62: This could be simplified as:
```suggestion
    metadata["createdAt"]
      //docker artifact createdAt time is under date field
      ?: metadata["date"]
```

luispollo (approved):
LGTM. Just a suggestion for a simplification inline.

PR #1589, opened 2020-10-09
refactor(model): Rename artifact models for clarity

luispollo:
Renames `DeliveryArtifact` to `ArtifactSpec` and `PublishedArtifact` to `ArtifactInstance` to make evident that the former is not actually an artifact, but a specification to find artifact instances, and that the latter represents such instances. This PR also renames the `delivery_artifact` table to `artifact_spec` to match the new class name. The `artifact_versions` table is untouched as I didn't feel the name is confusing.

The goal of this PR is to resolve the confusion this naming has caused in the past. It's an internal change only: there is no impact to the format of delivery configs.


keel-sql/src/main/resources/db/changelog/20201008-rename-delivery-artifact.yml
```diff
@@ -0,0 +1,8 @@
+databaseChangeLog:
+  - changeSet:
+      id: rename-delivery-artifact
+      author: lpollo
+      changes:
+        - renameTable:
+            oldTableName: delivery_artifact
+            newTableName: artifact_spec
```
emjburns, line 8: What about just `artifact`? that matches the other tables (`resource`, `environment`, etc)

robfletcher: I think we should consider renaming the `artifact_uid` columns on `artifact_last_checked`, `delivery_config_artifact`, `environment_artifact_pin`, `environment_artifact_versions`, and `environment_artifact_veto` to `artifact_spec_uid` to avoid confusion

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/ArtifactInstance.kt
```diff
@@ -10,8 +10,7 @@ import java.time.Instant
  * key fields without which it doesn't make sense for an artifact to exist in Managed Delivery terms. It also adds
  * a couple of keel-specific fields to store artifact metadata.
  */
-// TODO: rename to `Artifact` or `ArtifactInstance`
-data class PublishedArtifact(
+data class ArtifactInstance(
```
robfletcher, line 13: Even just… `Artifact`?

keel-sql/src/main/resources/db/changelog/20201008-rename-delivery-artifact.yml
gal-yardeni: +1 to renaming to `artifact`

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/testSupport.kt
```diff
@@ -0,0 +1,19 @@
+package com.netflix.spinnaker.keel.persistence
+
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactInstance
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactSpec
+import com.netflix.spinnaker.keel.api.artifacts.ArtifactStatus
+import java.time.Instant
+
+fun ArtifactSpec.toArtifactInstance(version: String, status: ArtifactStatus? = null, createdAt: Instant? = null) =
```
gal-yardeni, line 8: thank you for pulling it out!

keel-sql/src/main/resources/db/changelog/20201008-rename-delivery-artifact.yml
luispollo: @emjburns @gal-yardeni But it's really _not_ an "artifact". That was the whole point... Another name I discussed with @robfletcher before was `ArtifactDescriptor`, but I think it's uglier than `ArtifactSpec`.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/ArtifactInstance.kt
luispollo: I think historically we avoided just `Artifact` because that's the name of the class in `kork` (which incidentally we also use in `keel`). I don't mind, personally, if others agree.

keel-sql/src/main/resources/db/changelog/20201008-rename-delivery-artifact.yml
emjburns: Can't you make the same argument about the resource table? It's a resource specification, not the actual resource. I'm just talking about the table name, not the actual class name. For the class I think `ArtifactSpec` makes sense

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/ArtifactInstance.kt
robfletcher: Yeah, that's a reasonable point

keel-sql/src/main/resources/db/changelog/20201008-rename-delivery-artifact.yml
```diff
@@ -0,0 +1,8 @@
+databaseChangeLog:
+  - changeSet:
+      id: rename-delivery-artifact
+      author: lpollo
+      changes:
+        - renameTable:
```
lorin, line 6: I'd like us to get in the happy of doing database migrations that don't generate errors. 

As implemented, when this migration takes effect during the red-black deployment, it's going to cause the old ASG to throw errors because it will be querying a database that no longer exists.

Here's one possible strategy:

Step 1: Create the new table, modify the write logic to write to both tables, modify the read the logic to read from the new table first, then read from the old one if the data is not in the new one.

Step 2:  Once the code is no longer running, backfill the new table with data from the old one.

Step 3: Modify the code to only read/write the new table, write a migration to drop the old table.

lorin: Also, if you were going to rename the database table, I would start a new PR for step 1 rather than rolling it into this one.

luispollo:
Closing this as it feels like a distraction at the moment. I'll leave the branch in place if anyone is interested in picking this up at any point.

PR #1592, opened 2020-10-09
feat(handlers): allow handler to pass on taking action

emjburns:
Introducing a new behavior for resource handlers: the ability to pass on doing anything.

Why do we want this? Sometimes we can detect that there is a diff but we can't act on it. When might this be? Well, we might have situations where there is a diff (say, instances are unhealthy or there are too many active server groups) but we don't have enough information to fix the problem. Right now we get that way when there are two active server groups and the most recent is unhealthy (*note* I will work towards making this not the case, but I don't think I can make progress unless we have a way to opt out of taking action).

We want to communicate that we can't resolve the diff to the user. I'm doing this by adding a new state called `DIFF_NOT_ACTIONABLE`. 

This is not a great state for the user to see. Basically, it says "hey a human needs to step in". I built this so that we can slowly handle more and more cases that might at first fall into this category. But having this state means we don't need to address them all at once.

Looking for feedback on the approach:
I opted to create a new method (`willTakeAction(..)`) that happens before we call any `upsert` methods so the plugin can do a little check. I did this because I thought it was decently clean to have one method determining this. It's easier to test the various diff scenarios where we want to not take action. It was also pretty easy to publish an event if the plugin passes.

The other option I thought of was to modify the `upsert` method so it could return no tasks and a message, and then if a plugin did this emit the event. That seemed much more complicated. 

_please view with ignore whitespace set_

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ActionResponse.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ActionResponse.kt
@@ -1,0 +1,6 @@
+package com.netflix.spinnaker.keel.api.plugins
+
+data class ActionResponse(
+  val willTakeAction: Boolean = true,
+  val message: String? = null
+)
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
@@ -84,7 +84,7 @@ interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
   suspend fun willTakeAction(
     resource: Resource<S>,
     resourceDiff: ResourceDiff<R>
-  ): ActionDecision = ActionDecision()
+  ): ActionDecision = ActionDecision(willAct = true)
 
   /**
    * Create a resource so that it matches the desired state represented by [resource].
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
@@ -84,7 +84,7 @@ interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
   suspend fun willTakeAction(
     resource: Resource<S>,
     resourceDiff: ResourceDiff<R>
-  ): ActionResponse = ActionResponse()
+  ): ActionDecision = ActionDecision()
 
   /**
    * Create a resource so that it matches the desired state represented by [resource].
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
@@ -76,6 +76,17 @@ interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
   suspend fun current(resource: Resource<S>): R?
 
   /**
+   * Evaluates the resource diff and determines whether it can take action to resolve the diff.
+   * @return a response indicating if it can take action, and a message explaining why if it can't.
+   *
+   * The default value is to take action on all changes. Only override if needed.
+   */
+  suspend fun willTakeAction(
+    resource: Resource<S>,
+    resourceDiff: ResourceDiff<R>
+  ): ActionResponse = ActionResponse()
+
+  /**
    * Create a resource so that it matches the desired state represented by [resource].
    *
    * By default this just delegates to [upsert].
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -7,7 +7,7 @@ import com.netflix.spinnaker.keel.api.ResourceDiff
 import com.netflix.spinnaker.keel.api.ResourceSpec
 import com.netflix.spinnaker.keel.api.VersionedArtifactProvider
 import com.netflix.spinnaker.keel.api.actuation.Task
-import com.netflix.spinnaker.keel.api.plugins.ActionResponse
+import com.netflix.spinnaker.keel.api.plugins.ActionDecision
 import com.netflix.spinnaker.keel.api.plugins.ResourceHandler
 import com.netflix.spinnaker.keel.api.plugins.supporting
 import com.netflix.spinnaker.keel.core.ResourceCurrentlyUnresolvable
@@ -173,8 +173,8 @@ class ResourceActuator(
 
         log.debug("Checking resource {}", id)
 
-        val actionResponse = plugin.willTakeAction(resource, diff)
-        if (actionResponse.willTakeAction) {
+        val decision = plugin.willTakeAction(resource, diff)
+        if (decision.willAct) {
           when {
             current == null -> {
               log.warn("Resource {} is missing", id)
@@ -212,8 +212,8 @@ class ResourceActuator(
             }
           }
         } else {
-          log.info("Resource {} skipped because it can't be fixed: ${actionResponse.message}", id)
-          publisher.publishEvent(ResourceDiffNotActionable(resource, actionResponse.message))
+          log.warn("Resource {} skipped because it can't be fixed: ${decision.message}", id)
+          publisher.publishEvent(ResourceDiffNotActionable(resource, decision.message))
         }
       } catch (e: ResourceCurrentlyUnresolvable) {
         log.warn("Resource check for {} failed (hopefully temporarily) due to {}", id, e.message)
@@ -294,7 +294,7 @@ class ResourceActuator(
   private suspend fun <S : ResourceSpec, R : Any> ResourceHandler<S, R>.willTakeAction(
     resource: Resource<*>,
     resourceDiff: ResourceDiff<*>
-  ): ActionResponse =
+  ): ActionDecision =
     willTakeAction(resource as Resource<S>, resourceDiff as ResourceDiff<R>)
 
   @Suppress("UNCHECKED_CAST")
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -7,12 +7,14 @@ import com.netflix.spinnaker.keel.api.ResourceDiff
 import com.netflix.spinnaker.keel.api.ResourceSpec
 import com.netflix.spinnaker.keel.api.VersionedArtifactProvider
 import com.netflix.spinnaker.keel.api.actuation.Task
+import com.netflix.spinnaker.keel.api.plugins.ActionResponse
 import com.netflix.spinnaker.keel.api.plugins.ResourceHandler
 import com.netflix.spinnaker.keel.api.plugins.supporting
 import com.netflix.spinnaker.keel.core.ResourceCurrentlyUnresolvable
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVeto
 import com.netflix.spinnaker.keel.diff.DefaultResourceDiff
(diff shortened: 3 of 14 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -116,6 +117,43 @@ class ClusterHandler(
...
+      activeServerGroups.forEach { serverGroup ->
+        val healthy = serverGroup.instanceCounts?.isHealthy(
+          resource.spec.deployWith.health,
+          resource.spec.resolveCapacity(serverGroup.location.region)
+        ) == true
+        if (!healthy && potentialInactionableRegions.contains(serverGroup.location.region)) {
+          inactionableRegions.add(serverGroup.location.region)
+        }
+      }
+      if (inactionableRegions.isNotEmpty()) {
```
emjburns, line 144: Here's the message we send for the case where we can't do anything.

Note: these methods are almost identical for the cluster handlers. It's on my todo list to create a nice base class for cluster handlers. I dream about it. But it hasn't happened yet.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ActionResponse.kt
```diff
@@ -0,0 +1,6 @@
+package com.netflix.spinnaker.keel.api.plugins
+
+data class ActionResponse(
```
gal-yardeni, line 3: Maybe rename to `ActionData`?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
```diff
@@ -75,6 +75,17 @@ interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
    */
   suspend fun current(resource: Resource<S>): R?
 
+  /**
+   * Evaluates the resource diff and determines whether it can take action to resolve the diff.
+   * @return a response indicating if it can take action, and a message explaining why if it can't.
+   *
+   * The default value is to take action on all changes. Only override if needed.
+   */
+  suspend fun willTakeAction(
```
gal-yardeni, line 84: it's a bit confusing to see `willTakeAction` as both a function and a property in `ActionResponse`. Is it possible to rename one of those?

emjburns: Ha yes very true. Do you have any ideas? I am very bad at naming

gal-yardeni: me too 😅  maybe `isActionRequired?`

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ActionResponse.kt
robfletcher: or `ActionDecision`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -171,44 +173,47 @@ class ResourceActuator(
...
+                  // if a task was launched and it completed, either successfully or not, the delta is resolved
+                  publisher.publishEvent(ResourceDeltaResolved(resource, clock))
+                }
+                else -> publisher.publishEvent(ResourceValid(resource, clock))
               }
-              else -> publisher.publishEvent(ResourceValid(resource, clock))
             }
           }
+        } else {
+          log.info("Resource {} skipped because it can't be fixed: ${actionResponse.message}", id)
```
robfletcher, line 215: Minor point but this might be a `warn` kinda thing, maybe?

robfletcher (approved):
Couple of super-minor comments but I think this fundamentally looks good

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ActionResponse.kt
emjburns: I like `ActionDecision`, I will switch to that

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: Good point, will update

lorin (approved):
The general approach looks good to me. (I like the idea of creating a new precondition check). I had a few stylistic comments, but nothing that would block merge.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
```diff
@@ -75,6 +75,17 @@ interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
...
+    resource: Resource<S>,
+    resourceDiff: ResourceDiff<R>
+  ): ActionResponse = ActionResponse()
```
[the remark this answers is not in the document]
emjburns: Do you think that's any cleaner? Personally, I'd rather have a clear value set to 'true' vs the assumption that null means something.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -171,44 +173,47 @@ class ResourceActuator(
 
         log.debug("Checking resource {}", id)
 
-        // todo eb: add support for plugins to look at a diff and say "I can't fix this"
-        // todo eb: emit event for ^ with custom message provided by the plugin
+        val actionResponse = plugin.willTakeAction(resource, diff)
+        if (actionResponse.willTakeAction) {
+          when {
```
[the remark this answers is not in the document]
emjburns: Ya, good point.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -171,44 +173,47 @@ class ResourceActuator(
 
         log.debug("Checking resource {}", id)
 
-        // todo eb: add support for plugins to look at a diff and say "I can't fix this"
```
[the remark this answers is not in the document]
emjburns: I did add this todo a week ago, so somehow it doesn't feel as nice. But it's still something!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
```diff
@@ -75,6 +75,17 @@ interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
...
+  /**
+   * Evaluates the resource diff and determines whether it can take action to resolve the diff.
+   * @return a response indicating if it can take action, and a message explaining why if it can't.
+   *
+   * The default value is to take action on all changes. Only override if needed.
+   */
+  suspend fun willTakeAction(
+    resource: Resource<S>,
+    resourceDiff: ResourceDiff<R>
+  ): ActionDecision = ActionDecision()
```
luispollo, line 87: I suggest making it explicit that the default is to take action here so you don't have to look at the other class:
```suggestion
  ): ActionDecision = ActionDecision(willAct = true)
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -171,44 +173,47 @@ class ResourceActuator(
...
+                  publisher.publishEvent(ResourceDeltaResolved(resource, clock))
+                }
+                else -> publisher.publishEvent(ResourceValid(resource, clock))
               }
-              else -> publisher.publishEvent(ResourceValid(resource, clock))
             }
           }
+        } else {
+          log.warn("Resource {} skipped because it can't be fixed: ${decision.message}", id)
+          publisher.publishEvent(ResourceDiffNotActionable(resource, decision.message))
```
luispollo, line 216: Just a thought: it might be useful to include the diff in the event.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -442,3 +443,22 @@ data class ResourceCheckError(
     exception.message
   )
 }
+
+data class ResourceDiffNotActionable(
+  override val kind: ResourceKind,
+  override val id: String,
+  override val application: String,
+  override val timestamp: Instant,
+  override val message: String?
```
luispollo, line 452: It doesn't look like we're using this constructor anywhere. Could just make the one below the default.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -116,6 +117,43 @@ class ClusterHandler(
...
+        return ActionDecision(
+          willAct = false,
+          message = "There is more than one server group enabled " +
+            "but the latest is not healthy in ${inactionableRegions.joinToString(" and ")}. " +
```
luispollo, line 148: This will look a bit weird if you have 3 regions, no? Maybe just use a comma instead?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -116,6 +117,43 @@ class ClusterHandler(
...
+            "Spinnaker cannot resolve the problem at this time, " +
+            "manual intervention might be required."
```
luispollo, line 150:
```suggestion
            "Spinnaker cannot resolve the problem at this time. " +
            "Manual intervention might be required."
```

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/TitusClusterHandler.kt
```diff
@@ -121,6 +122,43 @@ class TitusClusterHandler(
...
+          inactionableRegions.add(serverGroup.location.region)
+        }
+      }
+      if (inactionableRegions.isNotEmpty()) {
+        return ActionDecision(
+          willAct = false,
+          message = "There is more than one server group enabled " +
+            "but the latest is not healthy in ${inactionableRegions.joinToString(" and ")}. " +
+            "Spinnaker cannot resolve the problem at this time, " +
+            "manual intervention might be required."
```
luispollo, line 155: Same comments on message.

luispollo (approved):
Minor suggestions, otherwise LGTM. Good tests!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: I'll include the diff in the log, and publish a delta detected event here so the delta shows up. good call

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
emjburns: this is the data class definition, we need that

emjburns: oh i see what you're saying. I just want to leave it the same as the other events though

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
emjburns: It will look a bit weird, but I think ok for now (I just did some handling of this in the unhealthy notification which I can add if people think it's super weird).

It'll look like 
> `not healthy in us-east-1 and us-west-2 and eu-west-1`
Vs what we probably want is 
> `not healthy in us-east-1, us-west-2, and eu-west-1`
or
> `not healthy in all regions`

I'm going to leave this for a follow up PR because that change is small

PR #1593, opened 2020-10-09
feat(notifications): add region to unhealthy

emjburns:
Updating the subject of the notification to have a list of regions

Screen Shot 2020-10-09 at 4 15 28 PM

### DIFF
--- a/keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListener.kt
+++ b/keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListener.kt
@@ -4,7 +4,6 @@ import com.netflix.spinnaker.config.UnhealthyNotificationConfig
 import com.netflix.spinnaker.keel.api.AccountAwareLocations
 import com.netflix.spinnaker.keel.api.Locatable
 import com.netflix.spinnaker.keel.api.Monikered
-import com.netflix.spinnaker.keel.api.Resource
 import com.netflix.spinnaker.keel.events.ClearNotificationEvent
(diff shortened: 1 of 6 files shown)


gal-yardeni (approved):
so fancy 🎉

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListener.kt
```diff
@@ -82,7 +82,7 @@ class UnhealthyNotificationListener(
     val resourceUrl = "$spinnakerBaseUrl/#/applications/${resource.application}/clusters?${params.toURL()}"
 
     return Notification(
-      subject = "${resource.spec.displayName} is unhealthy",
+      subject = "${resource.spec.displayName} is unhealthy in ${regions.joinToString(" and ")}",
```
erikmunson, line 85: I suggest making this a little more human by checking the length of the list and using commas for all the but the last list item if the length exceeds 2. so like`us-east-1, us-west-2, and eu-west-1`. If you have access to the spec here as well, might say "all regions" if the length of this list is equal to the length of the regions in the spec.

it's stuff like this that makes UI codebases _so much fun_...

erikmunson (approved):
I love this ❤️ — had a suggestion around humanizing the text which is consistent with how we do things in the other parts of the product but either way it's great to see this info without even a click

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListener.kt
emjburns: Arg. I will attempt this.

PR #1594, opened 2020-10-12
chore(artifacts): Update DebianArtifactParser

luispollo:
Our current version of `DebianArtifactParser` is barfing on version strings like `lpollo-meme-lab-server-0.0.1~snapshot.202010092150-h1.8ae7d58`. This updates to the latest code from Rocket to fix it.


luispollo:
Turns out that parsing with this class in export, even with the updates, doesn't match the versions nebula is generating... 🤦  We'll need to figure something else out.

PR #1596, opened 2020-10-14
SSD by default

lorin:
Default to using SSD-backed EBS volumes instead of "standard magnetic". This will hopefully resolve a performance issue we've seen with keel where a node occasionally gets stuck with very high disk load.


lorin:
Closing this for now so it doesn't just sit here on draft mode. Will re-open when ready for review.

PR #1600, opened 2020-10-15
Support AMIs where regions were baked individually

robfletcher:
Keel has always only supported deploying a single "named image" to a cluster. However, there are situations where the correct app version may have different image names in each region. e.g.

* Bakery retries one region
* Multiple bakes happen targeting different individual regions

In this case we can fail to ever resolve the AMI because we only look at the latest image with a matching app version tag.

This led to an issue where a managed cluster in one region existed side-by-side with a pipeline deployed cluster in another.
Keel launched a bake, which succeeded, and it started deploying the cluster.
Meanwhile, the pipeline-initiated bake completed and the image it created was in the "wrong" region for the managed cluster.
Keel was no longer able to resolve the desired state of the cluster, resulting in errors.
Additionally, if anything in the cluster definition had changed, no update would have taken place because the AMI could no longer be found.

This PR allows for an app version to be split across multiple AMIs in different regions.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
@@ -11,7 +11,7 @@ import com.netflix.spinnaker.keel.api.support.EventPublisher
 import com.netflix.spinnaker.keel.artifacts.DebianArtifact
 import com.netflix.spinnaker.keel.bakery.api.ImageExistsConstraint
 import com.netflix.spinnaker.keel.clouddriver.ImageService
-import com.netflix.spinnaker.keel.clouddriver.getLatestNamedImageForAppVersionInRegions
+import com.netflix.spinnaker.keel.clouddriver.getLatestNamedImages
 import com.netflix.spinnaker.keel.getConfig
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
 import com.netflix.spinnaker.kork.exceptions.SystemException
@@ -46,7 +46,7 @@ class ImageExistsConstraintEvaluator(
 
   private fun imagesExistInAllRegions(version: String, vmOptions: VirtualMachineOptions): Boolean =
     runBlocking {
-      imageService.getLatestNamedImageForAppVersionInRegions(
+      imageService.getLatestNamedImages(
         appVersion = version.parseAppVersion(),
         account = defaultImageAccount,
         regions = vmOptions.regions
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
@@ -1,6 +1,5 @@
 package com.netflix.spinnaker.keel.bakery.constraint
 
-import com.netflix.frigga.ami.AppVersion
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
@@ -13,8 +12,8 @@ import com.netflix.spinnaker.keel.bakery.api.ImageExistsConstraint
 import com.netflix.spinnaker.keel.clouddriver.ImageService
 import com.netflix.spinnaker.keel.clouddriver.getLatestNamedImages
 import com.netflix.spinnaker.keel.getConfig
+import com.netflix.spinnaker.keel.parseAppVersion
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
-import com.netflix.spinnaker.kork.exceptions.SystemException
 import kotlinx.coroutines.runBlocking
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
@@ -53,9 +52,6 @@ class ImageExistsConstraintEvaluator(
       )
     }
       .keys.containsAll(vmOptions.regions)
-
-  private fun String.parseAppVersion() =
-    AppVersion.parseName(this) ?: throw SystemException("Invalid AMI app version: $this")
 
   private val defaultImageAccount: String
     get() = dynamicConfigService.getConfig("images.default-account", "test")
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
@@ -44,14 +44,8 @@ class ImageExistsConstraintEvaluator(
     version: String,
     deliveryConfig: DeliveryConfig,
     targetEnvironment: Environment
-  ): Boolean {
-    if (artifact !is DebianArtifact) {
-      return true
-    }
-
-    val image = findMatchingImage(version, artifact.vmOptions)
-    return image != null
-  }
+  ): Boolean =
+    artifact !is DebianArtifact || findMatchingImages(version, artifact.vmOptions).isNotEmpty()
 
   private data class Key(
     val account: String,
@@ -60,7 +54,7 @@ class ImageExistsConstraintEvaluator(
   )
 
   private val cache = cacheFactory
-    .asyncLoadingCache<Key, NamedImage>(cacheName = "namedImages") { key ->
+    .asyncLoadingCache<Key, Collection<NamedImage>>(cacheName = "namedImages") { key ->
       log.debug("Searching for baked image for {} in {}", key.version, key.regions.joinToString())
       imageService.getLatestNamedImageWithAllRegionsForAppVersion(
         // TODO: Frigga and Rocket version parsing are not aligned. We should consolidate.
@@ -71,7 +65,7 @@ class ImageExistsConstraintEvaluator(
       )
     }
 
-  private fun findMatchingImage(version: String, vmOptions: VirtualMachineOptions): NamedImage? =
+  private fun findMatchingImages(version: String, vmOptions: VirtualMachineOptions): Collection<NamedImage> =
     runBlocking {
       cache.get(Key(defaultImageAccount, version, vmOptions.regions)).await()
     }
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
@@ -10,13 +10,11 @@ import com.netflix.spinnaker.keel.api.plugins.ConstraintEvaluator
 import com.netflix.spinnaker.keel.api.support.EventPublisher
 import com.netflix.spinnaker.keel.artifacts.DebianArtifact
 import com.netflix.spinnaker.keel.bakery.api.ImageExistsConstraint
-import com.netflix.spinnaker.keel.caffeine.CacheFactory
 import com.netflix.spinnaker.keel.clouddriver.ImageService
-import com.netflix.spinnaker.keel.clouddriver.model.NamedImage
+import com.netflix.spinnaker.keel.clouddriver.getLatestNamedImageForAppVersionInRegions
 import com.netflix.spinnaker.keel.getConfig
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
 import com.netflix.spinnaker.kork.exceptions.SystemException
-import kotlinx.coroutines.future.await
 import kotlinx.coroutines.runBlocking
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
@@ -31,8 +29,7 @@ import org.springframework.stereotype.Component
 class ImageExistsConstraintEvaluator(
   private val imageService: ImageService,
   private val dynamicConfigService: DynamicConfigService,
-  override val eventPublisher: EventPublisher,
-  cacheFactory: CacheFactory
+  override val eventPublisher: EventPublisher
 ) : ConstraintEvaluator<ImageExistsConstraint> {
 
   override fun isImplicit(): Boolean = true
@@ -45,30 +42,20 @@ class ImageExistsConstraintEvaluator(
     deliveryConfig: DeliveryConfig,
     targetEnvironment: Environment
   ): Boolean =
-    artifact !is DebianArtifact || findMatchingImages(version, artifact.vmOptions).isNotEmpty()
+    artifact !is DebianArtifact || imagesExistInAllRegions(version, artifact.vmOptions)
 
-  private data class Key(
-    val account: String,
-    val version: String,
-    val regions: Set<String>
-  )
-
-  private val cache = cacheFactory
-    .asyncLoadingCache<Key, Collection<NamedImage>>(cacheName = "namedImages") { key ->
-      log.debug("Searching for baked image for {} in {}", key.version, key.regions.joinToString())
-      imageService.getLatestNamedImageWithAllRegionsForAppVersion(
-        // TODO: Frigga and Rocket version parsing are not aligned. We should consolidate.
-        appVersion = AppVersion.parseName(key.version)
-          ?: throw SystemException("Invalid AMI app version: ${key.version}"),
-        account = key.account,
-        regions = key.regions
+  private fun imagesExistInAllRegions(version: String, vmOptions: VirtualMachineOptions): Boolean =
+    runBlocking {
+      imageService.getLatestNamedImageForAppVersionInRegions(
+        appVersion = version.parseAppVersion(),
+        account = defaultImageAccount,
+        regions = vmOptions.regions
       )
     }
+      .keys.containsAll(vmOptions.regions)
 
-  private fun findMatchingImages(version: String, vmOptions: VirtualMachineOptions): Collection<NamedImage> =
-    runBlocking {
-      cache.get(Key(defaultImageAccount, version, vmOptions.regions)).await()
-    }
+  private fun String.parseAppVersion() =
+    AppVersion.parseName(this) ?: throw SystemException("Invalid AMI app version: $this")
 
   private val defaultImageAccount: String
     get() = dynamicConfigService.getConfig("images.default-account", "test")
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
@@ -128,7 +128,7 @@ class ImageService(
    * As a side effect this method will prime the cache for any additional regions where the image is
    * available.
    */
-  suspend fun getLatestNamedImageForAppVersionInRegion(
+  suspend fun getLatestNamedImage(
     appVersion: AppVersion,
     account: String,
     region: String
@@ -216,7 +216,7 @@ class ImageService(
 
 /**
  * Find the latest properly tagged images in [account] for each of [regions] using
- * [ImageService.getLatestNamedImageForAppVersionInRegion] in parallel for each region.
+ * [ImageService.getLatestNamedImage] in parallel for each region.
  *
  * In many cases all the values in the resulting map will be the same [NamedImage] instance, but
  * this may not be the case if images were baked separately in each region.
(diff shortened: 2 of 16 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -37,14 +38,6 @@ class ImageService(
...
     getLatestNamedImage(artifactName, account)?.toImage(artifactName)
 
-  /**
-   * If possible, return the latest image that's present in all regions and has tags.
-   * If that doesn't exist, just return the latest image.
-   */
-  suspend fun getLatestImageWithAllRegions(artifactName: String, account: String, regions: List<String>): Image? =
-    getLatestNamedImageWithAllRegions(artifactName, account, regions)?.toImage(artifactName)
-      ?: getLatestImage(artifactName, account)
-
```
robfletcher, line 47: Not used anywhere

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -83,110 +76,43 @@ class ImageService(
...
-      .firstOrNull {
-        // TODO: Frigga and Rocket version parsing are not aligned. We should consolidate.
-        try {
-          AppVersion.parseName(it.appVersion).run {
-            packageName == appVersion.packageName && version == appVersion.version && commit == appVersion.commit
-          }
-        } catch (ex: Exception) {
-          throw SystemException("trying to parse name for version ${it.appVersion} but got an exception", ex)
-        }
-      }
```
robfletcher, line 110: Not used anywhere

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -84,109 +77,42 @@ class ImageService(
...
-   * Get a specific image for an app version.
-   *
-   * @param region if supplied the latest image in this region is returned, if `null` the latest
-   * image regardless of region.
+   * Returns the latest image that is present in all regions.
+   * Each AMI must have tags.
    */
-  suspend fun getLatestNamedImage(appVersion: AppVersion, account: String, region: String? = null): NamedImage? =
-    cloudDriverService.namedImages(
+  suspend fun getLatestNamedImageWithAllRegionsForAppVersion(appVersion: AppVersion, account: String, regions: Collection<String>): Collection<NamedImage> {
```
lorin, line 83: I'd make this plural `getLatestNamedImages...` since it now returns a collection, and update the header comment to reflect the fact that there are scenarios where it can return multiple NamedImage objects.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -114,30 +117,34 @@ class ImageResolver(
...
+    return VersionedNamedImage(listOf(image), null, null)
   }
 
   private fun Resource<ClusterSpec>.withVirtualMachineImages(image: VersionedNamedImage): Resource<ClusterSpec> {
     val imageIdByRegion = image
-      .namedImage
-      .amis
+      .namedImages
+      .map { it.amis }
+      .reduceRight { acc, map -> acc + map }
```
lorin, line 127: Woo, foldr!

lorin (approved):
Picked a very small nit, otherwise LGTM

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
robfletcher: I'm getting rid of this complexity right now!

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
```diff
@@ -44,37 +40,21 @@ class ImageExistsConstraintEvaluator(
...
-        regions = key.regions
+  private fun imagesExistInAllRegions(version: String, vmOptions: VirtualMachineOptions): Boolean =
+    runBlocking {
+      imageService.getLatestNamedImageForAppVersionInRegions(
+        appVersion = version.parseAppVersion(),
+        account = defaultImageAccount,
+        regions = vmOptions.regions
       )
     }
+      .keys.containsAll(vmOptions.regions)
```
robfletcher, line 54: Look for images in all the regions, and pass the constraint if we have them all

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -83,110 +83,91 @@ class ImageService(
...
+            }
+        }
+        // filter to images in the correct account and the desired region
+        .filter { image ->
+          image.accounts.contains(account) && image.amis.containsKey(region)
+        }
+        // reduce to the newest images required to support all regions we want
+        .sortedByDescending { it.creationDate }
+        .firstOrNull()
+  }
```
robfletcher, line 123: This is the call to CloudDriver that backs the new cache. It looks for `appVersion` in the specified account and _single_ region. That filtering logic is the same as existed before but the region check is simpler now because it only cares about one.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -83,110 +83,91 @@ class ImageService(
...
+    region: String
+  ): NamedImage? =
+    namedImageCache
+      .get(NamedImageCacheKey(appVersion, account, region))
+      .await()
+      // prime the cache if the image is also in other regions
+      ?.also { image ->
+        (image.amis.keys - region).forEach { otherRegion ->
+          namedImageCache.put(NamedImageCacheKey(appVersion, account, otherRegion), completedFuture(image))
         }
```
robfletcher, line 143: Method to find a single region just refers to the cache that will load on demand from CloudDriver. If the resulting image actually supports other regions besides the desired one, it pre-seeds the cache for those other regions too which will make future lookups (or lookups happening in parallel) more efficient.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -83,110 +83,91 @@ class ImageService(
...
-      "Finding latest qualifying named image for $packageName in account $account and regions $regions:\n " +
-        "selected image=${image?.imageName}\n " +
-        "rejected images=${eliminatedImages.map { it.key + ": " + it.value + "\n" }.joinToString("")}"
-    )
-
-    return image
+    }
+      .mapValues { (_, it) -> it.await() }
+      .filterNotNullValues()
   }
```
robfletcher, line 171: Method to find images in multiple regions. It's just doing a nice little coroutine-ized async call to `getLatestNamedImageForAppVersionInRegion` for each region. Most often the result will be the same for all regions.

The advantage of doing it this way is that the calling code doesn't have to consider what regions are supported by what image, it can just key off the map and not care if they're all the same image, all different ones, or somewhere in between.

I decided to always return whatever regions we _do_ find, rather than returning nothing if we don't find them all as it's less confusing. This just means the calling code needs to check whether all regions are in the result map (if it cares).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/_maps.kt
```diff
@@ -0,0 +1,9 @@
+package com.netflix.spinnaker.keel
+
+/**
+ * Filters a map to retain only those entries that have non-null values. Also narrows the
+ * value type on the map.
+ */
+@Suppress("UNCHECKED_CAST")
+fun <K, V : Any> Map<out K, V?>.filterNotNullValues(): Map<K, V> =
+  filterValues { it != null } as Map<K, V>
```
robfletcher, line 9: Just moved this from a private method in `ImageResolver` as I'm using it somewhere else now too (should be in the Kotlin stdlib if you ask me)

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/constraints/Ec2CanaryConstraintDeployHandler.kt
```diff
@@ -54,17 +53,16 @@ class Ec2CanaryConstraintDeployHandler(
...
       account = imageResolver.defaultImageAccount,
-      regions = constraint.regions.toList()
+      regions = regions
     )
-      ?.imageName ?: error("Image not found for $version in all requested regions ($regions)")
+
+    val missingRegions = regions - images.keys
+    if (missingRegions.isNotEmpty()) {
+      error("Image not found for $version in all requested regions ($regions)")
+    }
```
robfletcher, line 65: Example of calling code checking that all the regions it asked for were found

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -146,7 +160,7 @@ class ImageResolver(
       spec = spec.copy(
         overrides = overrides,
         _artifactName = image.artifact?.name
-          ?: error("Artifact not found in image ${image.namedImage.imageName}"),
+          ?: error("Artifact not found in images ${image.namedImages.values.map { it.imageName }}"),
```
robfletcher, line 163: Interestingly the Jenkins image path always sets `image.artifact` to `null` which makes me believe it flat out doesn't work any more. Planning a further PR to remove the deprecated image providers

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -259,6 +213,35 @@ class ImageService(
...
+    async {
+      getLatestNamedImageForAppVersionInRegion(
+        appVersion = appVersion,
+        account = account,
+        region = region
+      )
+    }
+  }
+    .mapValues { (_, it) -> it.await() }
+    .filterNotNullValues()
```
robfletcher, line 242: Method to find images in multiple regions. It's just doing a nice little coroutine-ized async call to `getLatestNamedImageForAppVersionInRegion` for each region. Most often the result will be the same for all regions.

The advantage of doing it this way is that the calling code doesn't have to consider what regions are supported by what image, it can just key off the map and not care if they're all the same image, all different ones, or somewhere in between.

I decided to always return whatever regions we _do_ find, rather than returning nothing if we don't find them all as it's less confusing. This just means the calling code needs to check whether all regions are in the result map (if it cares).

I implemented this as an extension function because that way the tests can use stubs for the individual region calls. In some cases that simpler, but it also has the advantage that it's actually exercising this parallelization logic.

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageServiceTests.kt
```diff
@@ -64,7 +57,7 @@ class ImageServiceTests : JUnit5Minutests {
       tagsByImageId = mapOf(
         "ami-001" to mapOf(
           "build_host" to "https://jenkins/",
-          "appversion" to "my-package-0.0.1~rc.97-h98/JENKINS-job/98",
+          "appversion" to "my-package-0.0.1~rc.97-h98.1c684a9/JENKINS-job/98",
```
robfletcher, line 60: Took me a while to figure out why this wasn't working correctly. The appversion wouldn't parse

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
```diff
@@ -44,37 +41,21 @@ class ImageExistsConstraintEvaluator(
...
-      log.debug("Searching for baked image for {} in {}", key.version, key.regions.joinToString())
-      imageService.getLatestNamedImageWithAllRegionsForAppVersion(
-        // TODO: Frigga and Rocket version parsing are not aligned. We should consolidate.
-        appVersion = AppVersion.parseName(key.version)
-          ?: throw SystemException("Invalid AMI app version: ${key.version}"),
-        account = key.account,
-        regions = key.regions
+  private fun imagesExistInAllRegions(version: String, vmOptions: VirtualMachineOptions): Boolean =
+    runBlocking {
+      imageService.getLatestNamedImageForAppVersionInRegions(
```
luispollo, line 49: nit, and I know this wasn't changed in this PR, but I find this function name unnecessarily long/complicated. We're naming the function after the parameters it takes... It could just be `getLatestNamedImage` with the right params.

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageServiceTests.kt
luispollo: Welcome to version scheme hell... 😭

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -83,111 +83,65 @@ class ImageService(
...
-        val curAppVersion = try {
-          AppVersion.parseName(namedImage.appVersion)
-        } catch (ex: Exception) {
-          throw SystemException("trying to parse name for image ${namedImage.imageName} with version ${namedImage.appVersion} but got an exception", ex)
+        // filter to images with matching app version
+        .filter { image ->
+          // TODO: Frigga and Rocket version parsing are not aligned. We should consolidate.
+          runCatching { AppVersion.parseName(image.appVersion) }
+            .getOrElse { ex ->
+              throw SystemException("trying to parse name for image ${image.imageName} with version ${image.appVersion} but got an exception", ex)
```
luispollo, line 110: nit:
```suggestion
              throw SystemException("Error parsing name for image ${image.imageName} with version ${image.appVersion}: ${ex.message}", ex)
```

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -83,111 +83,65 @@ class ImageService(
...
-        }
-        if (!tagsExistForAllAmis(it.tagsByImageId)) {
-          errors.add("[image does not have tags for all regions: existing tags ${it.tagsByImageId}]")
-        }
-        if (errors.isEmpty()) {
-          true
-        } else {
-          eliminatedImages[it.imageName] = errors.joinToString(",")
-          false
+  suspend fun getLatestNamedImageForAppVersionInRegion(
```
luispollo, line 131: As discussed above, consider renaming to `getLatestNamedImage`.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -83,111 +83,65 @@ class ImageService(
...
+    appVersion: AppVersion,
+    account: String,
+    region: String
+  ): NamedImage? =
+    namedImageCache
+      .get(NamedImageCacheKey(appVersion, account, region))
+      .await()
+      // prime the cache if the image is also in other regions
```
luispollo, line 139: Nice!

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
luispollo: Nice! ❤️

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -259,6 +213,35 @@ class ImageService(
...
+ * Find the latest properly tagged images in [account] for each of [regions] using
+ * [ImageService.getLatestNamedImageForAppVersionInRegion] in parallel for each region.
+ *
+ * In many cases all the values in the resulting map will be the same [NamedImage] instance, but
+ * this may not be the case if images were baked separately in each region.
+ *
+ * The resulting map will contain no entry for regions where an image is not found. The calling
+ * code must check this if it requires all regions to be present.
+ */
+suspend fun ImageService.getLatestNamedImageForAppVersionInRegions(
```
luispollo, line 227: As discussed above, consider renaming to `getLatestNamedImagesByRegion`.

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageServiceTests.kt
```diff
@@ -378,48 +348,112 @@ class ImageServiceTests : JUnit5Minutests {
...
+      val appVersion = AppVersion.parseName("my-package-0.0.1~rc.97-h98.1c684a9")
+      val regions = setOf("us-west-1", "ap-south-1")
       before {
         every {
-          cloudDriver.namedImages(
-            user = DEFAULT_SERVICE_ACCOUNT,
-            imageName = packageName,
-            account = "test"
+          cloudDriver.namedImages(any(), any(), any())
+        } returns listOf(image1, image6)
```
luispollo, line 356: I find these tests pretty hard to reason about with the current naming for the image variables. Consider renaming them to something that helps understand why they're in "varying stages of completeness".

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/constraints/Ec2CanaryConstraintDeployHandler.kt
```diff
@@ -54,17 +54,16 @@ class Ec2CanaryConstraintDeployHandler(
...
 
-    val image = imageService.getLatestNamedImageWithAllRegionsForAppVersion(
-      // TODO: Frigga and Rocket version parsing are not aligned. We should consolidate.
-      appVersion = try {
-        AppVersion.parseName(version.replace("~", "_"))
-      } catch (ex: Exception) {
-        throw SystemException("trying to parse name for version $version but got an exception", ex)
-      },
+    val images = imageService.getLatestNamedImageForAppVersionInRegions(
+      appVersion = AppVersion.parseName(version.replace("~", "_")),
```
luispollo, line 58: Any particular reason why you removed the `try/catch` block here?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/constraints/Ec2CanaryConstraintDeployHandler.kt
```diff
@@ -54,17 +54,16 @@ class Ec2CanaryConstraintDeployHandler(
...
       account = imageResolver.defaultImageAccount,
-      regions = constraint.regions.toList()
+      regions = regions
     )
-      ?.imageName ?: error("Image not found for $version in all requested regions ($regions)")
+
+    val missingRegions = regions - images.keys
+    if (missingRegions.isNotEmpty()) {
+      error("Image not found for $version in all requested regions ($regions)")
```
luispollo, line 65:
```suggestion
      error("Image not found for $version in all requested regions ($regions.joinToString())")
```

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageServiceTests.kt
robfletcher: Yeah, I thought about completely rewriting this test class since the hardest thing in this entire PR was reasoning about what the fuck it was doing. I felt like the PR was large enough though and that might be best tackled separately.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
robfletcher: I changed it as suggested

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/constraints/Ec2CanaryConstraintDeployHandler.kt
robfletcher: I actually meant to go back and make a standard extension function for this. It appears like there are 2 error modes, one in which it throws an exception and one in which it returns `null`.

luispollo:
@robfletcher This PR is near and dear to my heart because the very first orca bug I ever tried to fix (a few weeks into the job) was to give at least a decent error message for users when we couldn't find images with matching names in the different regions. This was because no one had the energy to try and fix the real issue. Love it that you're addressing this in MD. ❤️

luispollo (approved):

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/caffeine/CacheFactory.kt
```diff
@@ -71,6 +70,12 @@ class CacheFactory(
     CaffeineCacheMetrics.monitor(meterRegistry, this, cacheName)
 }
 
+fun <K, V> (suspend (K) -> V?).toAsyncCacheLoader() : AsyncCacheLoader<K, V> =
```
lorin, line 73: I look at this code and think "I have no idea if it's correct or not, but I bet if it compiles then it'll work properly."

lorin (approved):
LGTM

PR #1602, opened 2020-10-15
chore(monitoring): Enable New Relic tracing in select methods

luispollo:
Annotates `ResourceActuator.checkResource` and `EnvironmentPromotionChecker.checkEnvironments` so that New Relic will record traces for them (which doesn't happen by default for Kotlin suspend functions).

This won't have any practical effects in any environment where the New Relic agent is not actually running.

Reference: https://docs.newrelic.com/docs/agents/java-agent/api-guides/java-agent-api-instrument-using-annotation


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -22,6 +23,7 @@ class EnvironmentPromotionChecker(
 ) {
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
+  @Trace(dispatcher=true)
```
luispollo, line 26: https://docs.newrelic.com/docs/agents/java-agent/api-guides/java-agent-api-instrument-using-annotation

PR #1605, opened 2020-10-16
fix(api): Convert crazy Slack markdown syntax

luispollo:
We've been using Slack markdown syntax for hyperlinks in commit messages because they render nicely in Slack, but not so much in the UI. This PR converts the link syntax so that it will render OK in the UI. 

Note: depends on a UI update to run the commit message text through the markdown parser.

Cc: @erikmunson


emjburns (approved):
Nice! That's pretty cool!

PR #1607, opened 2020-10-16
chore(frigga): handle frigga exceptions gracefully

gal-yardeni:
This PR is wrapping all calls to Frigga to parse a version with try/catch, adding logs around it, and throw `systemException` if needed.

### DIFF
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
@@ -14,7 +14,6 @@ import com.netflix.spinnaker.keel.api.plugins.SupportedArtifact
 import com.netflix.spinnaker.keel.api.plugins.SupportedVersioningStrategy
 import com.netflix.spinnaker.keel.api.support.EventPublisher
 import com.netflix.spinnaker.keel.services.ArtifactMetadataService
-import org.apache.logging.log4j.util.Strings
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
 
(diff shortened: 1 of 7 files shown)


keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -62,32 +63,47 @@ class DebianArtifactSupplier(
 
   override fun getVersionDisplayName(artifact: PublishedArtifact): String {
     // TODO: Frigga and Rocket version parsing are not aligned. We should consolidate.
-    val appversion = AppVersion.parseName(artifact.version)
-    return if (appversion?.version != null) {
-      appversion.version
-    } else {
-      artifact.version.removePrefix("${artifact.name}-")
+    return try {
+      val appversion = AppVersion.parseName(artifact.version)
```
lorin, line 67: Since the same try/catch/log.warn logic appears in three separate places in this class, I'd create a helper function here to call in place of `AppVersion.parseName`, something like:

```kotlin
fun parseName(artifact: PublishedArtifact): AppVersion? =
  try {
    appVersion.parseName(artifact.version)
 } catch (ex: Exception) {
    log.warn("trying to parse artifact ${artifact.name} with version ${artifact.version} but got an exception", ex)
    null
}
```

That way, it's easier to change the logging statement in the future since it's only in one place.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -62,32 +63,47 @@ class DebianArtifactSupplier(
...
+      if (appversion?.version != null) {
+        return appversion.version
+      } else {
+        return artifact.version.removePrefix("${artifact.name}-")
+      }
+    } catch (ex: Exception) {
+      log.warn("trying to parse artifact ${artifact.name} with version ${artifact.version} but got an exception", ex)
+      Strings.EMPTY
```
emjburns, line 75: I think we should return `artifact.version` here, not whatever this value is, because it's for the display name

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -122,7 +123,12 @@ class ImageHandler(
     diff: DefaultResourceDiff<Image>
   ): List<Task> {
     // TODO: Frigga and Rocket version parsing are not aligned. We should consolidate.
-    val appVersion = AppVersion.parseName(desiredVersion)
+    val appVersion = try {
+      AppVersion.parseName(desiredVersion)
+    } catch (ex: Exception) {
+      log.error("trying to parse desired version for artifact ${artifact.name} with version $desiredVersion but got an exception", ex)
+      throw SystemException("Invalid for artifact ${artifact.name} with version $desiredVersion")
```
lorin, line 130:
```suggestion
      throw SystemException("Invalid for artifact ${artifact.name} with version $desiredVersion", ex)
```

lorin: If you pass the original exception as the second argument to `SystemException`, then its stack trace will also appear on the log (it will show the SystemException stack trace, then it will say "caused by..." and show the original stack trace). If you do that, I think you can eliminate the `log.error` statement on line 129.

lorin (approved):
A couple of minor comments, nothing that would block merge.

emjburns:
I don't think this will fix the problem. We want to ignore an unparsable version, not throw an exception. There are two ways to do this I think. The first is in the artifact repository getVersions call, filter out anything that doesn't parse with frigga for debians. The second might be in the image handler. I'm wondering if you could modify [this call](https://github.com/spinnaker/keel/blob/master/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt#L44) to find the latest parsable version (i.e. keep looking at older versions if the version isn't parsable)

gal-yardeni:
@emjburns I see what you are saying. I think it's handling some part of the problem, as the environment view is crashing on not being able to parse the build info. Do you think it's worth merging this first and handle the other part in a follow-up PR? I'm not quite sure how to handle the other part yet.

emjburns:
@gal-yardeni won't we throw an exception anyways now? I don't feel like throwing a `SystemException` is really an improvement over the current exception we throw. Can you give me an example of how the call is failing and how this PR will fix it?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
gal-yardeni: done

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
gal-yardeni: done

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
gal-yardeni: oh nice! let me do that.

gal-yardeni:
after chatting with @emjburns , we decided this PR is addressing one aspect of the problem - which is handling frigga exceptions gracefully (i.e not crashing the UI when we have a bad version).

A follow-up PR will be addressing the other aspect - meaning, filtering out bad versions prior to storing those in the DB. For example, for debians we will ignore completely `local` versions. For dockers, we will ignore `latest` tag (different handling that we are doing today).
@lorin I addressed your feedback as well.

emjburns (approved):

PR #1608, opened 2020-10-17
feature(ec2): Add support for launch templates, take 2

lorin:
## Summary

Add support for managing EC2 apps that have been configured to use AWS launch templates. This is a second 
attempt at #1566, which was reverted in #1573 because of a bug.

## What was the original bug?

When resolving the current status of a cluster, if the current launch configuration contains a ramdisk id field that is
an empty string: `""`, keel is supposed to treat it as a `null`.

#1566 introduced a regression where it did not convert the blank string to a null. This led to false diffs where
the current state of an EC2 cluster had a ramdisk id of `""` and the desired state of an cluster had a ramdisk id of `null`.

A test has been added for this case. 

The rest of this description is identical to #1566.

## Background

Clouddriver now support AWS [launch templates](https://docs.aws.amazon.com/autoscaling/ec2/userguide/LaunchTemplates.html). Clouddriver opts in apps to use launch templates, on a per-app basis.

## What changed from keel's perspective

For apps that are configured to use launch templates, the response payload from clouddriver changes when querying for server group information. For these apps, there is no longer a `launchConfig` field. Instead, there is a `launchTemplate` field. The `launchTemplate` field contains similar information, but the data has a different shape.

## PR details

With this PR, we can now handle response payloads from clouddriver where there is a `launchTemplate` field instead of a `launchConfig` field.

Note that this PR does not change:

* the delivery config schema (the section called `launchConfiguration` remains unchanged)
* the shape of the data that we send to orca when actuating EC2 clusters

### DIFF
--- a/keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/TestUtils.kt
+++ b/keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/TestUtils.kt
@@ -10,9 +10,12 @@ import com.netflix.spinnaker.keel.clouddriver.model.ActiveServerGroupImage
 import com.netflix.spinnaker.keel.clouddriver.model.AutoScalingGroup
 import com.netflix.spinnaker.keel.clouddriver.model.Capacity
 import com.netflix.spinnaker.keel.clouddriver.model.CustomizedMetricSpecificationModel
+import com.netflix.spinnaker.keel.clouddriver.model.IamInstanceProfile
 import com.netflix.spinnaker.keel.clouddriver.model.InstanceCounts
 import com.netflix.spinnaker.keel.clouddriver.model.InstanceMonitoring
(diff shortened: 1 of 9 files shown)


keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/TestUtils.kt
```diff
@@ -26,14 +29,53 @@ import com.netflix.spinnaker.keel.core.parseMoniker
...
+ *
+ * The older launch configuration
+ * The newer launch templates
+ */
+enum class LaunchInfo {
+  LAUNCH_CONFIG,
+  LAUNCH_TEMPLATE
+}
+
 fun ServerGroup.toCloudDriverResponse(
```
lorin, line 44: This is test helper code that has been updated so it can return the two different types of launch info payloads

emjburns (approved):

PR #1610, opened 2020-10-20
feat(artifact): veto only if never deployed successfully

emjburns:
**Problem:**
One of the most helpful things we can do in MD is veto an artifact when the version isn't good (like, when the deploy times out because the instances never came up). 

One of the most _unhelpful_ things we can do is veto artifacts en mass when there is a problem (either with keel, with spinnaker, or with a cloud provider).

**Solution:**
This PR aims to narrow the scope of vetoing artifact versions to only veto versions that have **never** been successfully deployed. That means that we will no longer veto artifact versions that have been happy in an environment. 

We will not veto an artifact version if:
- there was a config change that is causing the instances to not start up
- there is a keel bug that causes the desired state of everything to change in a way that causes bad deploys (like what happened when we accidentally shrunk all the titus memory requirements)
- there is an AWS / titus bug that causes behavior like ^

In this case, we will just mark the cluster as unhappy and not do anything else. 

**Thoughts:**
I implemented this in the resource actuator because that's where we have information about the desired state as well as the resource. I wanted to implement this in the `UnhappyVeto` but we don't know what the desired state is there (it hasn't been calculated) so we can't do the right thing. Also, we don't have the artifact repository wired into the veto. Does that seem reasonable or is this a bad idea?

I also did a bit of refactoring of the veto stuff into a method or two since @lorin suggested this last time I touched this code. I think it makes it easier to read. But it does make it harder to tell what logic is new in this PR.

**I'm looking for:**
- Feedback on this general idea. Is it good? 
- Cases that I missed where we should actually veto the artifact

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -1,5 +1,7 @@
 package com.netflix.spinnaker.keel.actuation
 
+import com.netflix.spectator.api.Registry
+import com.netflix.spinnaker.keel.api.CompleteVersionedArtifact
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.Resource
@@ -33,6 +35,7 @@ import com.netflix.spinnaker.keel.persistence.ResourceRepository
 import com.netflix.spinnaker.keel.plugin.CannotResolveCurrentState
 import com.netflix.spinnaker.keel.plugin.CannotResolveDesiredState
 import com.netflix.spinnaker.keel.plugin.ResourceResolutionException
+import com.netflix.spinnaker.keel.telemetry.ArtifactVersionVetoed
 import com.netflix.spinnaker.keel.telemetry.ResourceCheckSkipped
 import com.netflix.spinnaker.keel.veto.VetoEnforcer
 import com.netflix.spinnaker.keel.veto.VetoResponse
@@ -69,7 +72,7 @@ class ResourceActuator(
   private val vetoEnforcer: VetoEnforcer,
   private val publisher: ApplicationEventPublisher,
   private val clock: Clock,
-  private val springEnv: SpringEnvironment,
+  private val springEnv: SpringEnvironment
 ) {
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
@@ -117,53 +120,9 @@ class ResourceActuator(
 
         val response = vetoEnforcer.canCheck(resource)
         if (!response.allowed) {
-          /**
-           * [VersionedArtifactProvider] is a special [resource] sub-type. When a veto response sets
-           * [VetoResponse.vetoArtifact] and the resource under evaluation is of type
-           * [VersionedArtifactProvider], disallow the desired artifact version from being deployed to
-           * the environment containing [resource]. This ensures that the environment will be fully restored to
-           * a prior good-state.
-           */
-          if (response.vetoArtifact && resource.spec is VersionedArtifactProvider) {
-            try {
-              val versionedArtifact = when (desired) {
-                is Map<*, *> -> {
-                  if (desired.size > 0) {
-                    @Suppress("UNCHECKED_CAST")
-                    (desired as Map<String, VersionedArtifactProvider>).values.first()
-                  } else {
-                    null
-                  }
-                }
-                is VersionedArtifactProvider -> desired
-                else -> null
-              }?.completeVersionedArtifactOrNull()
-
-              if (versionedArtifact != null) {
-                with(versionedArtifact) {
-                  val artifact = deliveryConfig.matchingArtifactByName(versionedArtifact.artifactName, artifactType)
-                    ?: error("Artifact $artifactType:$artifactName not found in delivery config ${deliveryConfig.name}")
-
-                  artifactRepository.markAsVetoedIn(
-                    deliveryConfig = deliveryConfig,
-                    veto = EnvironmentArtifactVeto(
-                      reference = artifact.reference,
-                      version = artifactVersion,
-                      targetEnvironment = environment.name,
-                      vetoedBy = "Spinnaker",
-                      comment = "Automatically marked as bad because multiple deployments of this version failed."
-                    )
-                  )
-                  // TODO: emit event + metric
-                }
-              }
-            } catch (e: Exception) {
-              log.warn("Failed to veto presumed bad artifact version for ${resource.id}", e)
-              // TODO: emit metric
-            }
-          }
-          log.debug("Skipping actuation for resource {} because it was vetoed: {}", id, response.message)
-          publisher.publishEvent(ResourceCheckSkipped(resource.kind, id, response.vetoName))
+          handleVetoingArtifact(response, resource, deliveryConfig, environment, desired)
+          log.debug("Skipping actuation for resource {} because it was vetoed: {}", resource.id, response.message)
+          publisher.publishEvent(ResourceCheckSkipped(resource.kind, resource.id, response.vetoName))
           publisher.publishEvent(
             ResourceActuationVetoed(
               resource.kind,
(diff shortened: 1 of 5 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -237,6 +196,91 @@ class ResourceActuator(
...
+            log.info(
+              "Vetoing artifact version {} of artifact {} for config {} and env {} because multiple deploys failed",
+              versionedArtifact,
+              artifact.reference,
+              deliveryConfig.name,
+              environment.name
+            )
+            publisher.publishEvent(ArtifactVersionVetoed(artifact.reference))
+          } else {
+            log.info(
```
gal-yardeni, line 253: Can you elaborate on how do you know, by `versionedArtifact == null`, that this artifact version was deployed successfully before?

gal-yardeni: oh i'm sorry, I was confused with `previousSuccess==null`. nevermind :)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
```diff
@@ -50,3 +50,7 @@ data class ArtifactCheckTimedOut(
   val name: String,
   val deliveryConfigName: String?
 ) : TelemetryEvent()
+
+data class ArtifactVersionVetoed(
+  val artifactReference: String,
```
luispollo, line 55: IMHO, we should stop using `artifactReference` alone as an identifier, and couple it with `artifactType`. Especially now that we can specify artifacts based on branches, I can easily see a case where you might have a frontend NPM package coming from master that you label `master`, and a backend Docker/Debian that you also label `master`. The difference is the type.

emjburns: @luispollo This is just for a telemetry event - because I want to add a tag here. I think you're right that the reference isn't useful (and it's not enough to identify as you pointed out). If an artifact is vetoed, what information do you want? Maybe the application is best here? We will have to dig into logs to see anything useful regardless.

luispollo: Well, even in Atlas, the problem I mentioned would come up, so I think it'd be worth adding the type here. Other than that, my comment was just because I see us propagating this pattern of using _just_ the reference to identify artifacts (even in repository methods), and it'll be harder to fix the more we add.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -237,6 +196,91 @@ class ResourceActuator(
...
+          }
+        }
+      } catch (e: Exception) {
+        log.warn("Failed to veto presumed bad artifact version for ${resource.id}", e)
+        // TODO: emit metric
+      }
+    }
+  }
+
+  private fun Any?.findArtifact() : CompleteVersionedArtifact? =
```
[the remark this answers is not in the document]
gal-yardeni: lol was thinking the same.

gal-yardeni (approved):
LGTM!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
emjburns: I switched to application because I think that at a glance we want to know which app is having artifacts vetoed. Then we can dive into the logs to find out more info

lorin (approved):
This approach looks good to me.

I can't think of a scenario where an artifact was successfully deployed previously but we still want to automatically veto it.

erikmunson:
OOOOO this is good. I'm sure _something_ will break this assumption at some point, but we won't find out until we try. Seems like a good/predictable way to behave in these situations.

congratulations

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
luispollo: You beat me to it! I was just about to suggest that. Good call!

Also, I take back what I said about artifact references. The whole point of those was to be able to use a single string value in a resource spec to point to an artifact, so it makes sense for them to be globally unique in the delivery config! So, in order to reason about an artifact in logs/metrics, we need the application and artifact reference as you did.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -237,6 +195,91 @@ class ResourceActuator(
...
+    when (this) {
+      is Map<*, *> -> {
+        if (this.size > 0) {
+          @Suppress("UNCHECKED_CAST")
+          (this as Map<String, VersionedArtifactProvider>).values.first()
```
robfletcher, line 273: Is this to handle the case where e.g. a cluster resolves to a map of region to server group?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationEventTests.kt
```diff
@@ -38,7 +38,7 @@ internal class ApplicationEventTests {
 
     publisher.publishEvent(TestEvent(this))
 
-    val eventThread = listener.awaitInvoked(Duration.ofMillis(100))
+    val eventThread = listener.awaitInvoked(Duration.ofMillis(300))
```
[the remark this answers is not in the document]
emjburns: It was timing out before the event got received. maybe github actions was being weird? So I bumped this up a little bit because 300 millis is not that much and I prefer it to the whole build failing

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -237,6 +195,91 @@ class ResourceActuator(
...
+   * [VersionedArtifactProvider] is a special [resource] sub-type. When a veto response sets
+   * [VetoResponse.vetoArtifact], the resource under evaluation is of type
+   * [VersionedArtifactProvider], and the version has never before been deployed successfully to an environment,
+   * disallow the desired artifact version from being deployed to
+   * the environment containing [resource]. This ensures that the environment will be fully restored to
+   * a prior good-state.
+   *
+   * If the version has previously been deployed successfully do not veto the artifact because
+   * we do not want to veto the artifact in the case of a bad config change or other downstream problem.
+   * Rolling back most likely will not help in these cases and it will probably cause confusion.
```
robfletcher, line 208: I'm fairly confused here. Is the `VetoResponse` in reference to the same artifact version that `desired` is referencing? That implies that something has already decided this version should be vetoed but we're now deciding whether to honor that decision? 

Or is the version `desired` references the one we're considering rolling back to?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationEventTests.kt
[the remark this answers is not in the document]
robfletcher: That's just the max time it will wait, right? It's not a `sleep` call.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: This is just refactoring code that already existed, so I'm not 100% sure

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: Yes, you're correct. The veto response doesn't specify the version, just whether the artifact version that is present should be vetoed. I agree that ideally the logic for vetoing an artifact should belong in the actual veto! Unfortunately, I don't feel we have enough data in the veto right now to make the right call (I describe this further in the "thoughts" section of my PR description). So, the way I implemented this was to decide whether or not to honor that veto in the `ResourceActuator`.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ApplicationEventTests.kt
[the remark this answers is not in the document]
emjburns: ya

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: Updated the method description to be a bit more detailed

PR #1611, opened 2020-10-21
fix(actuation): Fix handling of empty actuation task list

luispollo:
Ran into this bug with Ludo's app because the plugin does not override the `willTakeAction` method (yet), and was returning an empty list of tasks since there was nothing to do, but keel didn't check if the list was empty. We ended up recording 5 "actuations" and vetoing the resource for no reason.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -81,9 +81,14 @@ class ResourceActuator(
     withTracingContext(resource) {
       val id = resource.id
       try {
-        log.debug("checkResource $id")
         val plugin = handlers.supporting(resource.kind)
+        val deliveryConfig = deliveryConfigRepository.deliveryConfigFor(resource.id)
+        val environment = checkNotNull(deliveryConfig.environmentFor(resource)) {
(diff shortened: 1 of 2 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -81,9 +81,14 @@ class ResourceActuator(
     withTracingContext(resource) {
       val id = resource.id
       try {
-        log.debug("checkResource $id")
         val plugin = handlers.supporting(resource.kind)
+        val deliveryConfig = deliveryConfigRepository.deliveryConfigFor(resource.id)
```
luispollo, line 85: Moved these up here so we can log better context just below.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -200,8 +198,10 @@ class ResourceActuator(
 
               plugin.update(resource, diff)
                 .also { tasks ->
-                  publisher.publishEvent(ResourceActuationLaunched(resource, plugin.name, tasks, clock))
-                  diffFingerprintRepository.markActionTaken(id)
+                  if (tasks.isNotEmpty()) {
```
luispollo, line 201: This was the issue/fix.

emjburns (approved):

PR #1614, opened 2020-10-22
Detect AMIs split across multiple named images when checking for latest baked versions

robfletcher:
`ImageHandler` previously looked at a single `NamedImage` and could potentially think regions were missing if they were split across multiple image names.

This PR changes things so that the method in `ImageService` that finds the latest image for an artifact, looks for all named images that share the latest `appVersion` and `base_ami_version` tags.

This will stop metrics flagging that a region is missing in such a case.


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -42,33 +44,43 @@ class ImageService(
...
+    return if (latest == null) {
+      log.debug("No images found for {}", artifactName)
+      null
+    } else {
+      val regions = candidateImages
+        .filter {
+          it.appVersion == latest.appVersion && it.baseImageVersion == latest.baseImageVersion
         }
+        .flatMap { it.amis.keys }
+        .toSet()
```
robfletcher, line 74: Don't assume `latest` is the only image with that app version and base AMI version. Any others that may cover different regions are deployable as well.

emjburns (approved):

PR #1615, opened 2020-10-22
feat(processArtifacts): Move artifact processing decision to individual suppliers

gal-yardeni:
Until now, we had some filtering mechanism to filter artifacts at the repository level, or at the publishing time.

This PR move that responsibility to each artifact provider in a unified way, giving us an easy way to control the versions we want to filter at the API level.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
@@ -88,6 +88,11 @@ interface ArtifactSupplier<A : DeliveryArtifact, V : VersioningStrategy> : Spinn
    * This function is currently expected to make calls to CI systems.
    */
   suspend fun getArtifactMetadata(artifact: PublishedArtifact): ArtifactMetadata?
+
+  /**
+   * Given an artifact definition [DeliveryArtifact], and an actual artifact version as [PublishedArtifact], return whether this artifact should be processed and saved
+   */
+  fun shouldProcessArtifact(artifactDefinition: DeliveryArtifact, artifact: PublishedArtifact): Boolean
 }
 
 /**
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
@@ -52,19 +52,25 @@ class ArtifactListener(
       .artifacts
       .filter { it.type.toLowerCase() in artifactTypeNames }
       .forEach { artifact ->
-        if (repository.isRegistered(artifact.name, artifact.artifactType)) {
+        val deliveryArtifact = repository.getDeliveryArtifact(artifact.name, artifact.artifactType)
+        //check if the artifact is registered
+        if (deliveryArtifact != null) {
           val artifactSupplier = artifactSuppliers.supporting(artifact.artifactType)
-          val enrichedArtifact = artifactSupplier.addMetadata(artifact.normalized())
+          if (artifactSupplier.shouldProcessArtifact(deliveryArtifact, artifact)) {
+            val enrichedArtifact = artifactSupplier.addMetadata(artifact.normalized())
 
-          log.info("Registering version {} (status={}) of {} artifact {}",
-            artifact.version, artifact.status, artifact.type, artifact.name)
+            log.info("Registering version {} (status={}) of {} artifact {}",
+              artifact.version, artifact.status, artifact.type, artifact.name)
 
-          repository.storeArtifactInstance(enrichedArtifact)
-            .also { wasAdded ->
-              if (wasAdded) {
-                publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.artifactType))
+            repository.storeArtifactInstance(enrichedArtifact)
+              .also { wasAdded ->
+                if (wasAdded) {
+                  publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.artifactType))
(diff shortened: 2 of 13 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -279,12 +291,7 @@ class SqlArtifactRepository(
       versions
     } else {
       val sortedVersions = versions.sortedWith(artifact.versioningStrategy.comparator)
-      if (artifact is DockerArtifact) {
-        // FIXME: remove special handling for Docker
```
gal-yardeni, line 283: @luispollo say bye-bye to special handling for docker 😄

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -208,6 +208,18 @@ class SqlArtifactRepository(
         .value1()
     } > 0
```
emjburns, line 210: You can't do this this way - there are potentially many artifacts that will have the same name and type. You can only look up a delivery artifact given the artifact config.

For example:
Let's say several teams are deploying redis from the docker image `nflx/redis` provided by a centralized team. Each team is running their own redis cluster from that image, and wants to update it when a new version is rolled out. Each team has a redis artifact defined in their config.

Since there are several artifacts referring to redis by `name = nflx/redis` and `type = docker`, you can't find the delivery artifact by just knowing the name and type. You have to know the delivery config before you can find a delivery artifact.

emjburns (changes requested):
See my comment about the `getDeliveryArtifact` method

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -52,19 +52,25 @@ class ArtifactListener(
       .artifacts
       .filter { it.type.toLowerCase() in artifactTypeNames }
       .forEach { artifact ->
```
emjburns, line 54: "registered" means that we are saving versions for that artifact name + type. It's purposefully separate from the delivery config so that multiple apps can care about the same debian package or docker image

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ArtifactSupplier.kt
```diff
@@ -88,6 +88,11 @@ interface ArtifactSupplier<A : DeliveryArtifact, V : VersioningStrategy> : Spinn
    * This function is currently expected to make calls to CI systems.
    */
   suspend fun getArtifactMetadata(artifact: PublishedArtifact): ArtifactMetadata?
+
+  /**
+   * Given an artifact definition [DeliveryArtifact], and an actual artifact version as [PublishedArtifact], return whether this artifact should be processed and saved
+   */
+  fun shouldProcessArtifact(artifactDefinition: DeliveryArtifact, artifact: PublishedArtifact): Boolean
```
emjburns, line 95: I think for this we just need the published artifact.

The rules for processing should depend only on the type, and not someone's definition.

The rules we have are: 
- we ignore docker tags matching "latest"
- we ignore locally published debians, which we can tell by parsing the version and then trying to get the build number (if it is "local", we should ignore it)

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
```diff
@@ -125,6 +128,27 @@ class DockerArtifactSupplier(
...
+  /**
+   * Given a docker artifact and a docker tag, filters out all tags that don't produce exactly one capture
+   * group with the provided regex.
+   *
+   * This means that this will filter out tags like "latest" from the list.
+   */
+  override fun shouldProcessArtifact(artifactDefinition: DeliveryArtifact, artifact: PublishedArtifact): Boolean {
+    return try {
+      if (artifactDefinition is DockerArtifact) {
+        artifact.version != "latest" && TagComparator.parseWithRegex(artifact.version, artifactDefinition.tagVersionStrategy, artifactDefinition.captureGroupRegex) != null
```
emjburns, line 140: we don't want to exclude things here based on the comparator logic. All those tags are valid. The only tag that's not valid is "latest"

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: I see. I didn't really change the implementation in the caller function tho - we used to check if an artifact is registered based on name + type, and this function is just returning one of them. I don't understand why it's ok to use this logic in the `isRegistered` and not to fetch the actual artifact (although I understand your example).

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
gal-yardeni: ok, that changes things. Let me fix it.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: ok I was too quick to be happy about it :(

gal-yardeni:
closing this PR in favor of a new one (coming soon).

PR #1616, opened 2020-10-22
feat(titus): default container attrs resolver

emjburns:
The titus team recently started setting some default container attributes. This adds them to a user spec if they're not present.

Default values are pulled from the config.

You can turn it off via fast property in case we mess something up :)

### DIFF
--- a/keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/ContainerAttributesResolverTests.kt
+++ b/keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/ContainerAttributesResolverTests.kt
@@ -1,0 +1,130 @@
+package com.netflix.spinnaker.keel.titus
+
+import com.netflix.spinnaker.keel.api.Moniker
+import com.netflix.spinnaker.keel.api.Resource
+import com.netflix.spinnaker.keel.api.SimpleLocations
+import com.netflix.spinnaker.keel.api.SimpleRegionSpec
+import com.netflix.spinnaker.keel.api.plugins.supporting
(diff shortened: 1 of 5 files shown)


keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/ContainerAttributesResolverTests.kt
```diff
@@ -0,0 +1,130 @@
...
+      test("account and subnet are set") {
+        validateKeysSet(resolved)
+      }
+    }
+
+    context("account is set") {
+      fixture {
+        Fixture(
+          ContainerAttributesResolver(defaults, clouddriverService, springEnv),
+          baseSpec.copy(_defaults = baseSpec.defaults.copy(containerAttributes = mapOf(accountKey to awsAccountId)))
```
lorin, line 90: I would recommend using a different account id here (or adding another test with a different account id), to catch the bug where the user specifies a custom account id, and the resolver clobbers it with the default.

robfletcher (approved):

lorin (approved):
Small comment on the tests, otherwise LGTM!

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/ContainerAttributesResolverTests.kt
emjburns: ok i'll do that, good point

emjburns: oh wait i do this in the following test!

PR #1617, opened 2020-10-22
fix(veto): Fix never-expiring resource vetos

luispollo:
After much debugging, I found that, if a resource had been previously vetoed and a new diff detected, instead of rechecking the resource as expected after the wait period, we were just repeatedly vetoing it because we were always updating the `recheckTime` in that case, so it never expired.

This PR fixes that by introducing a check for only inserting a record when the veto is first created for the resource, and only updating it when the recheck time has expired, but _not_ in the in-between checks. I found it pretty hard to reason about this code because half of the business logic was in the `SqlUnhappyVetoRepository`, so this PR also refactors the code a bit to consolidate all the logic in `UnhappyVeto` which makes it easier to follow and to test.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepositoryTests.kt
@@ -17,21 +17,16 @@
  */
 package com.netflix.spinnaker.keel.persistence
 
-import com.netflix.spinnaker.keel.persistence.UnhappyVetoRepository.UnhappyVetoStatus
 import com.netflix.spinnaker.time.MutableClock
-import dev.minutest.experimental.minus
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
 import java.time.Clock
 import java.time.Duration
-import strikt.api.Assertion
-import strikt.api.expect
-import strikt.api.expectCatching
 import strikt.api.expectThat
 import strikt.assertions.containsExactlyInAnyOrder
 import strikt.assertions.hasSize
 import strikt.assertions.isEqualTo
-import strikt.assertions.isSuccess
+import strikt.assertions.isNull
 
 abstract class UnhappyVetoRepositoryTests<T : UnhappyVetoRepository> : JUnit5Minutests {
   abstract fun factory(clock: Clock): T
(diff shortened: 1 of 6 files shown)


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepositoryTests.kt
```diff
@@ -59,87 +54,33 @@ abstract class UnhappyVetoRepositoryTests<T : UnhappyVetoRepository> : JUnit5Min
...
+        expectThat(subject.getRecheckTime(resourceId)).isEqualTo(clock.instant() + waitDuration)
+      }
+
       test("marking happy works") {
         subject.delete(resourceId)
         expectThat(subject.getAll()).hasSize(0)
       }
     }
 
-    context("expiring the time") {
```
luispollo, line 85: Similar tests now exist in `UnhappyVetoTests`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/UnhappyVetoRepository.kt
```diff
@@ -55,11 +55,9 @@ abstract class UnhappyVetoRepository(
    *
    * @param wait the time to wait before re-checking the resource, `null` means "never re-check".
    */
-  abstract fun getOrCreateVetoStatus(
-    resourceId: String,
-    application: String,
-    wait: Duration? = Duration.parse(waitingTime)
-  ): UnhappyVetoStatus
+  abstract fun getRecheckTime(
```
luispollo, line 58: Refactored this to do just the repository part, which is to return data from the database.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlUnhappyVetoRepository.kt
```diff
@@ -60,21 +59,14 @@ class SqlUnhappyVetoRepository(
...
       jooq
         .select(UNHAPPY_VETO.RECHECK_TIME)
         .from(UNHAPPY_VETO)
         .where(UNHAPPY_VETO.RESOURCE_ID.eq(resourceId))
-        .fetchOne()
-        ?.value1()
-    }
-
-    return if (recheckTime == null || recheckTime > clock.timestamp()) {
-      markUnhappyForWaitingTime(resourceId, application, wait)
```
luispollo, line 74: This was the problematic line: we should not be updating the recheck time when it has not yet expired, because we'd just keep pushing it forward into the future forever.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
```diff
@@ -56,20 +60,29 @@ class UnhappyVeto(
...
-    // allow for a check every [waitingTime] even if the resource is unhappy
-    if (vetoStatus.shouldRecheck) {
-      log.debug("Marking resource $resourceId unhappy for $wait, but allowing resource check.")
-      unhappyVetoRepository.markUnhappyForWaitingTime(resourceId, application, wait)
-      return allowedResponse()
-    }
+    val recheckTime = unhappyVetoRepository.getRecheckTime(resourceId)
 
-    if (vetoStatus.shouldSkip) {
+    /**
```
luispollo, line 65: I consolidated all the business logic for unhappy vetos here so it's easy to reason about. I added a comment describing the algorithm which hopefully also helps others understand the code.

emjburns (approved):

PR #1619, opened 2020-10-23
feat(processArtifacts): Move artifact processing decision to individual suppliers

gal-yardeni:
This PR move that responsibility to each artifact provider in a unified way, giving us an easy way to control the versions we want to filter at the API level.

### DIFF
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
@@ -54,17 +54,21 @@ class ArtifactListener(
       .forEach { artifact ->
         if (repository.isRegistered(artifact.name, artifact.artifactType)) {
           val artifactSupplier = artifactSuppliers.supporting(artifact.artifactType)
-          val enrichedArtifact = artifactSupplier.addMetadata(artifact.normalized())
+          if (artifactSupplier.shouldProcessArtifact(artifact)) {
+            val enrichedArtifact = artifactSupplier.addMetadata(artifact.normalized())
 
-          log.info("Registering version {} (status={}) of {} artifact {}",
-            artifact.version, artifact.status, artifact.type, artifact.name)
+            log.info("Registering version {} (status={}) of {} artifact {}",
+              artifact.version, artifact.status, artifact.type, artifact.name)
 
-          repository.storeArtifactInstance(enrichedArtifact)
-            .also { wasAdded ->
-              if (wasAdded) {
-                publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.artifactType))
+            repository.storeArtifactInstance(enrichedArtifact)
+              .also { wasAdded ->
+                if (wasAdded) {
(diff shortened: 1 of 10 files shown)


keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -54,17 +54,21 @@ class ArtifactListener(
...
-                publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.artifactType))
+            repository.storeArtifactInstance(enrichedArtifact)
+              .also { wasAdded ->
+                if (wasAdded) {
+                  publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.artifactType))
+                }
               }
-            }
+          } else {
+            log.debug("Artifact $artifact shouldn't be processed due to supplier limitations. Ignoring this artifact version.")
```
emjburns, line 70: excellent log statement, this will be useful

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -101,10 +101,31 @@ class DebianArtifactSupplier(
...
-  private fun PublishedArtifact.hasReleaseStatus() =
-    this.metadata.containsKey("releaseStatus") && this.metadata["releaseStatus"] != null
 
   private val DeliveryArtifact.statusesForQuery: List<String>
     get() = statuses.map { it.name }
+
+  override fun shouldProcessArtifact(artifact: PublishedArtifact): Boolean =
+    artifact.hasReleaseStatus() && artifact.hasCorrectVersion()
+
+
```
emjburns, line 111: duplicate comment here

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -101,10 +101,31 @@ class DebianArtifactSupplier(
...
+      true
+    } else {
+      log.debug("Ignoring artifact event without release status: $this")
+      false
+    }
+  }
+
+  // Debian Artifacts should not have "local" as a part of their version string
+  private fun PublishedArtifact.hasCorrectVersion() : Boolean {
+    return if (!this.version.contains("local")) {
```
emjburns, line 124: I think we only care about "local" in a specific spot. I think it would be better to parse with frigga here and then get the build number to check if that is == local

emjburns:
Looks great except for my comment around local

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
gal-yardeni: ohhhh good point!!!!

emjburns (approved):

PR #1620, opened 2020-10-26
Allow resource updates to change version

robfletcher:
Previously on updating a resource the migrators were applied so the "current" resource compared with the request payload would be the latest spec version. If the only change in the payload was to update the version (and associated changes in spec format) the update would appear to be a no-op. This PR introduces a new repository method `getRaw` that retrieves the resource as currently stored in the database (without applying migrators). This is only intended for use in the update request flow.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
@@ -45,12 +45,23 @@ interface ResourceRepository : PeriodicallyCheckedRepository<Resource<ResourceSp
   fun allResources(callback: (ResourceHeader) -> Unit)
 
   /**
-   * Retrieves a single resource by its unique [id].
+   * Retrieves a single resource by its unique [id]. If the resource is of an obsolete kind, the
+   * result is automatically migrated to the latest version of that kind.
    *
    * @return The resource represented by [id] or `null` if [id] is unknown.
    * @throws NoSuchResourceException if [id] does not map to a resource in the repository.
    */
   fun get(id: String): Resource<ResourceSpec>
+
+  /**
+   * Retrieves a single resource by its unique [id]. No migration is performed if the resource is of
+   * an obsolete kind.
+   *
+   * This is only intended for use when updating resources as automatically migration can mask the
+   * difference between current and updated resource states when a user is trying to update to a
+   * newer kind.
+   */
+  fun getRaw(id: String): Resource<ResourceSpec>
 
   fun hasManagedResources(application: String): Boolean
 
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
@@ -61,18 +61,23 @@ open class SqlResourceRepository(
     }
   }
 
-  override fun get(id: String): Resource<ResourceSpec> {
-    return sqlRetry.withRetry(READ) {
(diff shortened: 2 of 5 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -45,13 +45,20 @@ interface ResourceRepository : PeriodicallyCheckedRepository<Resource<ResourceSp
...
 
   /**
-   * Retrieves a single resource by its unique [id].
+   * Retrieves a single resource by its unique [id]. If the resource is of an obsolete kind, the
+   * result is automatically migrated to the latest version of that kind.
    *
    * @return The resource represented by [id] or `null` if [id] is unknown.
    * @throws NoSuchResourceException if [id] does not map to a resource in the repository.
    */
   fun get(id: String): Resource<ResourceSpec>
```
emjburns, line 55: Can you add your comment about how this method should be used here as well?

emjburns (approved):

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
robfletcher: Is that better?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepository.kt
```diff
@@ -45,13 +45,24 @@ interface ResourceRepository : PeriodicallyCheckedRepository<Resource<ResourceSp
...
 
+  /**
+   * Retrieves a single resource by its unique [id]. No migration is performed if the resource is of
+   * an obsolete kind.
+   *
+   * This is only intended for use when updating resources as automatically migration can mask the
```
luispollo, line 60:
```suggestion
   * This is only intended for use when updating resources as automatically migrating them can mask the
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -82,23 +87,35 @@ open class SqlResourceRepository(
...
+    parseKind(kind).let {
+      Resource(
+        it,
+        objectMapper.readValue<Map<String, Any?>>(metadata).asResourceMetadata(),
+        objectMapper.readValue(spec, resourceSpecIdentifier.identify(it))
+      )
+    }
+
   /**
    * Constructs a resource object from its database representation
```
luispollo, line 105:
```suggestion
   * Constructs a resource object from its database representation, applying migrators in the process.
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -61,18 +61,23 @@ open class SqlResourceRepository(
     }
   }
 
-  override fun get(id: String): Resource<ResourceSpec> {
-    return sqlRetry.withRetry(READ) {
+  override fun get(id: String): Resource<ResourceSpec> =
+    readResource(id, this::parseAndMigrateResource)
+
+  override fun getRaw(id: String): Resource<ResourceSpec> =
```
luispollo, line 67: I find the use of the word "raw" here a bit misleading, since the return type is the same as the "non-raw" version. Perhaps `getOriginalResource` here and `getCanonicalResource` for the migrated version?

luispollo (approved):
LGTM, just a couple of suggestions on naming and doc comments.

PR #1621, opened 2020-10-27
refactor(plugins): Define Java-friendly versions of ResourceHandler methods

luispollo:
### What
Introduces Java-friendly counterparts to the `current` and `desired` methods in `ResourceHandler` that can be safely implemented in Java to avoid issues with shared threads with coroutines.

### Why
This is intended to address an issue we found where a Java-based plugin blocks in either of those calls, causing all of keel's resource actuation to also block since the Java code is entirely oblivious to coroutine management, but was blocking a coroutine thread.

### How
My approach was to define methods that take in an `Executor` passed in by keel and that are expected to return a `CompletableFuture` that runs in that executor to complete its work. This allows keel to convert the future to a coroutine-friendly `Deferred` object that respects coroutine-specific behaviors like timeouts and cancellations.

This should be 100% backwards-compatible with both Java- and Kotlin-based implementations, so I expect all the tests to pass.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
@@ -12,6 +12,8 @@ import com.netflix.spinnaker.keel.api.events.ArtifactVersionDeploying
 import com.netflix.spinnaker.keel.api.support.EventPublisher
 import com.netflix.spinnaker.kork.exceptions.SystemException
 import com.netflix.spinnaker.kork.plugins.api.internal.SpinnakerExtensionPoint
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.Executor
 
 /**
  * A [ResourceHandler] is a keel plugin that provides resource state monitoring and actuation capabilities
@@ -59,9 +61,24 @@ interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
    * The value returned by this method is used as the basis of the diff (with the result of
    * [current] in order to decide whether to call [create]/[update]/[upsert].
    *
+   * Note: Java-based plugins should override [desiredAsync] instead.
+   *
    * @param resource the resource as persisted in the Keel database.
    */
-  suspend fun desired(resource: Resource<S>): R
+  @JvmDefault
+  suspend fun desired(resource: Resource<S>): R =
+    TODO("Not implemented")
+
+  /**
+   * Asynchronous implementation of [desired] which returns a [CompletableFuture] encapsulating
+   * the execution of desired state retrieval for the [resource] using the specified [executor].
+   *
+   * This method should *only* be overridden by Java-based implementations. Kotlin-based implementations
+   * should override [desired] instead.
+   */
+  @JvmDefault
+  fun desiredAsync(resource: Resource<S>, executor: Executor): CompletableFuture<R> =
+    TODO("Not implemented")
 
   /**
    * Return the current _actual_ representation of what [resource] looks like in the cloud.
@@ -72,8 +89,23 @@ interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
    * [desired] in order to decide whether to call [create]/[update]/[upsert].
    *
    * Implementations of this method should not actuate any changes.
+   *
+   * Note: Java-based plugins should override [currentAsync] instead.
    */
-  suspend fun current(resource: Resource<S>): R?
+  @JvmDefault
+  suspend fun current(resource: Resource<S>): R? =
+    TODO("Not implemented")
+
+  /**
+   * Asynchronous implementation of [current] which returns a [CompletableFuture] encapsulating
+   * the execution of current state retrieval for the [resource] using the specified [executor].
+   *
+   * This method should *only* be overridden by Java-based implementations. Kotlin-based implementations
+   * should override [current] instead.
+   */
+  @JvmDefault
+  fun currentAsync(resource: Resource<S>, executor: Executor): CompletableFuture<R?> =
+    TODO("Not implemented")
 
   /**
    * Evaluates the resource diff and determines whether it can take action to resolve the diff.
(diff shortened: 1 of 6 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ResourceHandler.kt
```diff
@@ -52,16 +60,36 @@ interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
...
    *
+   * Note: the default implementation of this method delegates to [desiredAsync] for compatibility
+   * with Java-based plugins. Kotlin-based implementations should always override this method.
+   *
    * @param resource the resource as persisted in the Keel database.
    */
-  suspend fun desired(resource: Resource<S>): R
+  @JvmDefault
+  suspend fun desired(resource: Resource<S>): R =
+    // desiredAsync(resource, coroutineContext[CoroutineDispatcher]!!.asExecutor()).get()
```
luispollo, line 80: @robfletcher My initial attempt was to follow your suggestion and use the same dispatcher as the coroutine scope from within which this suspend function is called, but I ran into issues when the plugin blocked while executing in this context (unrelated coroutines got cancelled because the shared thread got interrupted, apparently). I then switched to a simple thread pool-backed executor below and things work as expected.

build.gradle.kts
```diff
@@ -74,6 +74,7 @@ subprojects {
         languageVersion = "1.4"
         jvmTarget = "11"
         freeCompilerArgs += "-Xjvm-default=enable"
+        freeCompilerArgs += "-Xopt-in=kotlin.ExperimentalStdlibApi"
```
luispollo, line 77: Required to access coroutine context elements, which surprisingly is an experimental API. 🤷‍♂️

luispollo:
@robfletcher One more data point: I ran a test where the plugin blocked for 60 seconds and I reduced the resource check timeout to 45 seconds, and while the coroutine scope did get a `TimeoutCancellationException`, it only happened after the full 60 seconds because the plugin code is running in an independent executor/thread pool which the coroutine management code is totally oblivious to. In other words, neither reusing the coroutine's dispatcher as an executor, nor using a separate executor, behave like what we'd like. 😞 

I'm checking with Jose on the option of converting the plugin to Kotlin for the time being as an alternative.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -304,31 +312,43 @@ class ResourceActuator(
...
+            desired(resource)
+          } catch (e: ResourceCurrentlyUnresolvable) {
+            throw e
+          } catch (e: Throwable) {
+            throw CannotResolveDesiredState(resource.id, e)
+          }
         }
+      } else {
+        // for Java compatibility
+        desiredAsync(resource, asyncExecutor).asDeferred()
```
luispollo, line 330: This works as expected with respect to coroutine timeouts from outer scopes. The only downside I've experienced is that cancelling the `CompletableFuture` does *not* interrupt the underlying running thread(s), which is by design since the interface is entirely decoupled from how the future is processed (may even be synchronous). I think that's acceptable, considering we're "enforcing" the use of a managed thread pool and that we're planning to implement timeouts in the Java plugin as well. Let me know what you think.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -304,31 +312,43 @@ class ResourceActuator(
...
-        try {
-          desired(resource)
-        } catch (e: ResourceCurrentlyUnresolvable) {
-          throw e
-        } catch (e: Throwable) {
-          throw CannotResolveDesiredState(resource.id, e)
+  private suspend fun <T : Any> ResourceHandler<*, T>.resolve(resource: Resource<ResourceSpec>): Pair<T, T?> {
+    val isKotlin = this.javaClass.isKotlinClass
+    return supervisorScope {
+      val desired = if (isKotlin) {
```
luispollo, line 318: This now checks if the plugin is a Kotlin class before using `async` to call `desired`, and falls back to calling `desiredAsync` (Java-friendly version) below if it's not. Same thing for `current` immediately below.

lorin: While this code works, I think the semantics would more clearly match the code if the `ResourceHandler` interface instead specified an enum that you could switch on to choose `desired` versus `desiredAsync`, something like this:

```kotlin
//
// ResourceHandler.kt
//
enum class Implementation {
 COROUTINE, ASYNC
}

interface ResourceHandler<S : ResourceSpec, R : Any> : SpinnakerExtensionPoint {
 val implementation : Implementation
}

//
// ResourceActuator.kt
//
private suspend fun <T : Any> ResourceHandler<*, T>.resolve(resource: Resource<ResourceSpec>) =
 supervisorScope {
  when(implementation) {
   COROUTINE -> { ... }
   ASYNC -> {...}
 }
}
```

lorin (approved):
I like this implementation, and it's strictly better than what we have right now. I made a stylistic comment but nothing that would block merge.

One thing I worry about more generally is a badly behaved plugin that ignores the passed executor and just blocks the main thread. (Note that the same risk exists for a badly behaved kotlin-based plugin that blocks the thread in the coroutine).

Solving the problem above would require changing the threading model in ResourceActuator. I'm not really worried about this risk today because Treasure is our only plugin and we have good control over the code, so I don't think it's worth working on right now.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: Oooh, that's good! I'll do that! Thanks for the suggestion.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: Good point on metrics. Not sure if there are any hooks available to instrument that (but now we have New Relic! 😉). BTW, if plugins are generally well-behaved, the cached thread pool should improve resource utilization, but they're not, we may end up with a lot of threads. We could switch that to a fixed thread pool instead. Thoughts?

lorin: I'd leave it as-is for now, we can change the implementation to a fixed thread pool later if we run into performance issues.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
lorin: Yeah, I agree that we should not have two separate checks. Ultimately we have to just trust that the plugin implementer is doing the right thing. I suggested the enum approach because I thought it would make the intent a little clearer in the ResourceActuator code, but it's really just an aesthetic judgement.

luispollo:
@robfletcher Are you still planning to look this over?

PR #1622, opened 2020-10-27
fix(titus): if we don't have a region be okay

emjburns:
Upon further reflection, let's not barf if we don't have a subnet entry. Let's just not add a key.

### DIFF
--- a/keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/ContainerAttributesResolverTests.kt
+++ b/keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/ContainerAttributesResolverTests.kt
@@ -20,6 +20,7 @@ import strikt.api.expect
 import strikt.api.expectThat
 import strikt.assertions.containsExactly
 import strikt.assertions.isEqualTo
+import strikt.assertions.isNull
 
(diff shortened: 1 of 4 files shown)


keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/ContainerAttributesResolverTests.kt
```diff
@@ -105,8 +109,8 @@ internal class ContainerAttributesResolverTests : JUnit5Minutests {
       }
 
       test("we leave the values that are set") {
-        val resolvedEast = resource.spec.resolveContainerAttributes("east")
-        val resolvedWest = resource.spec.resolveContainerAttributes("west")
+        val resolvedEast = resolved.spec.resolveContainerAttributes("east")
```
lorin, line 112: Were these tests incorrect before?

lorin (approved):
LGTM!

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/ContainerAttributesResolverTests.kt
emjburns: Yep!

PR #1623, opened 2020-10-27
chore(clusters): shared base cluster handler

emjburns:
This pr creates a shared cluster handler.

The `BaseClusterHandler` is an abstraction layer for clusters so that they don't have to implement some basic logic around orchestration, deciding things, etc. I created it because I keep writing really similar code for titus and ec2, and I'm tired of doing that. This pulls one function into the abstract base class to start. This also creates tests for that one function that are common.

I've chosen to use a spy of the handler so that we don't have to worry about dealing with the underlying bits of cloud that should be tested at the handler level (like, getting the active server groups). Instead, we can just craft the responses we want to test the right scenarios.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/AgentLockRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/AgentLockRepositoryTests.kt
@@ -1,6 +1,6 @@
 package com.netflix.spinnaker.keel.persistence
 
-import com.netflix.spinnaker.time.MutableClock
+import com.netflix.spinnaker.keel.time.MutableClock
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
 import java.time.Clock
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/AgentLockRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/AgentLockRepositoryTests.kt
@@ -1,6 +1,6 @@
 package com.netflix.spinnaker.keel.persistence
 
-import com.netflix.spinnaker.keel.time.MutableClock
+import com.netflix.spinnaker.time.MutableClock
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
 import java.time.Clock
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
@@ -1,0 +1,51 @@
+package com.netflix.spinnaker.keel.api.plugins
+
+import com.netflix.spinnaker.keel.api.Resource
+import com.netflix.spinnaker.keel.api.ResourceDiff
+import com.netflix.spinnaker.keel.api.ResourceSpec
+import com.netflix.spinnaker.keel.diff.toIndividualDiffs
+
+/**
+ * Common cluster functionality.
+ * This class contains abstracted logic where we want all clusters to make the
+ * same decision, or abstracted logic involving coordinating resources or
+ * publishing events.
+ */
+abstract class BaseClusterHandler<SPEC: ResourceSpec, RESOLVED: Any>(
+  resolvers: List<Resolver<*>>
+) : ResolvableResourceHandler<SPEC, Map<String, RESOLVED>>(resolvers) {
+
+  abstract fun getDesiredRegion(diff: ResourceDiff<RESOLVED>): String
+  abstract fun diffOnlyInEnabled(diff: ResourceDiff<RESOLVED>): Boolean
+  abstract fun getUnhealthyRegionsForActiveServerGroup(resource: Resource<SPEC>): List<String>
+
+  override suspend fun willTakeAction(
+    resource: Resource<SPEC>,
+    resourceDiff: ResourceDiff<Map<String, RESOLVED>>
+  ): ActionDecision {
+    // we can't take any action if there is more than one active server group
+    //  AND the current active server group is unhealthy
+    val potentialInactionableRegions = mutableListOf<String>()
+    val inactionableRegions = mutableListOf<String>()
+    resourceDiff.toIndividualDiffs().forEach { diff ->
+      if (diff.hasChanges() && diffOnlyInEnabled(diff)) {
+        potentialInactionableRegions.add(getDesiredRegion(diff))
+      }
+    }
+    if (potentialInactionableRegions.isNotEmpty()) {
+      val unhealthyRegions = getUnhealthyRegionsForActiveServerGroup(resource)
+      inactionableRegions.addAll(potentialInactionableRegions.intersect(unhealthyRegions))
+
+    }
+    if (inactionableRegions.isNotEmpty()) {
+      return ActionDecision(
+        willAct = false,
+        message = "There is more than one server group enabled " +
+          "but the latest is not healthy in ${inactionableRegions.joinToString(" and ")}. " +
+          "Spinnaker cannot resolve the problem at this time. " +
+          "Manual intervention might be required."
+      )
+    }
+    return ActionDecision(willAct = true)
+  }
+}
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
@@ -23,7 +23,7 @@ abstract class BaseClusterHandler<SPEC: ResourceSpec, RESOLVED: Any>(
   /**
    * returns true if the diff is only in whether there are too many clusters enabled
    */
-  abstract fun isDiffOnlyInEnabled(diff: ResourceDiff<RESOLVED>): Boolean
+  abstract fun ResourceDiff<RESOLVED>.isEnabledOnly(): Boolean
 
   /**
    * returns a list of regions where the active server group is unhealthy
@@ -39,7 +39,7 @@ abstract class BaseClusterHandler<SPEC: ResourceSpec, RESOLVED: Any>(
     val potentialInactionableRegions = mutableListOf<String>()
     val inactionableRegions = mutableListOf<String>()
     resourceDiff.toIndividualDiffs().forEach { diff ->
-      if (diff.hasChanges() && isDiffOnlyInEnabled(diff)) {
+      if (diff.hasChanges() && diff.isEnabledOnly()) {
         potentialInactionableRegions.add(getDesiredRegion(diff))
       }
     }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
@@ -15,8 +15,19 @@ abstract class BaseClusterHandler<SPEC: ResourceSpec, RESOLVED: Any>(
   resolvers: List<Resolver<*>>
 ) : ResolvableResourceHandler<SPEC, Map<String, RESOLVED>>(resolvers) {
 
+  /**
+   * parses the desired region from a resource diff
+   */
   abstract fun getDesiredRegion(diff: ResourceDiff<RESOLVED>): String
-  abstract fun diffOnlyInEnabled(diff: ResourceDiff<RESOLVED>): Boolean
+
+  /**
+   * returns true if the diff is only in whether there are too many clusters enabled
+   */
+  abstract fun isDiffOnlyInEnabled(diff: ResourceDiff<RESOLVED>): Boolean
+
+  /**
+   * returns a list of regions where the active server group is unhealthy
+   */
   abstract fun getUnhealthyRegionsForActiveServerGroup(resource: Resource<SPEC>): List<String>
 
   override suspend fun willTakeAction(
@@ -28,7 +39,7 @@ abstract class BaseClusterHandler<SPEC: ResourceSpec, RESOLVED: Any>(
     val potentialInactionableRegions = mutableListOf<String>()
     val inactionableRegions = mutableListOf<String>()
     resourceDiff.toIndividualDiffs().forEach { diff ->
-      if (diff.hasChanges() && diffOnlyInEnabled(diff)) {
+      if (diff.hasChanges() && isDiffOnlyInEnabled(diff)) {
         potentialInactionableRegions.add(getDesiredRegion(diff))
       }
     }
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -549,13 +549,10 @@ class ClusterHandler(
   /**
    * @return true if the only difference is in the onlyEnabledServerGroup property
    */
-  fun ResourceDiff<ServerGroup>.isEnabledOnly(): Boolean =
+  override fun ResourceDiff<ServerGroup>.isEnabledOnly(): Boolean =
     current != null &&
       affectedRootPropertyNames.all { it == "onlyEnabledServerGroup" } &&
       current!!.onlyEnabledServerGroup != desired.onlyEnabledServerGroup
-
-  override fun isDiffOnlyInEnabled(diff: ResourceDiff<ServerGroup>): Boolean =
-    diff.isEnabledOnly()
 
   /**
    * @return `true` if [current] doesn't exist and desired includes a scaling policy.
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
@@ -2,6 +2,7 @@ package com.netflix.spinnaker.keel.ec2.resource
 
 import com.fasterxml.jackson.module.kotlin.convertValue
 import com.netflix.rocket.api.artifact.internal.debian.DebianArtifactParser
+import com.netflix.spinnaker.keel.api.DeployHealth
 import com.netflix.spinnaker.keel.api.Exportable
 import com.netflix.spinnaker.keel.api.Moniker
 import com.netflix.spinnaker.keel.api.RedBlack
@@ -41,8 +42,7 @@ import com.netflix.spinnaker.keel.api.ec2.TerminationPolicy
 import com.netflix.spinnaker.keel.api.ec2.byRegion
 import com.netflix.spinnaker.keel.api.ec2.resolve
 import com.netflix.spinnaker.keel.api.ec2.resolveCapacity
-import com.netflix.spinnaker.keel.api.plugins.ActionDecision
-import com.netflix.spinnaker.keel.api.plugins.ResolvableResourceHandler
+import com.netflix.spinnaker.keel.api.plugins.BaseClusterHandler
 import com.netflix.spinnaker.keel.api.plugins.Resolver
 import com.netflix.spinnaker.keel.api.support.EventPublisher
 import com.netflix.spinnaker.keel.api.withDefaultsOmitted
@@ -99,7 +99,7 @@ class ClusterHandler(
   override val eventPublisher: EventPublisher,
   resolvers: List<Resolver<*>>,
   private val clusterExportHelper: ClusterExportHelper
-) : ResolvableResourceHandler<ClusterSpec, Map<String, ServerGroup>>(resolvers) {
+) : BaseClusterHandler<ClusterSpec, ServerGroup>(resolvers) {
 
   private val debianArtifactParser = DebianArtifactParser()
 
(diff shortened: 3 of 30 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Common cluster functionality.
+ * This class contains abstracted logic where we want all clusters to make the
+ * same decision, or abstracted logic involving coordinating resources or
+ * publishing events.
+ */
+abstract class BaseClusterHandler<SPEC: ResourceSpec, RESOLVED: Any>(
+  resolvers: List<Resolver<*>>
+) : ResolvableResourceHandler<SPEC, Map<String, RESOLVED>>(resolvers) {
```
luispollo, line 16: You can probably get rid of `ResolvableResourceHandler` now. The only purpose of that was to carry the reference to `resolvers` for both cluster handlers (which this class serves perfectly now).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
```diff
@@ -0,0 +1,51 @@
...
+
+  abstract fun getDesiredRegion(diff: ResourceDiff<RESOLVED>): String
+  abstract fun diffOnlyInEnabled(diff: ResourceDiff<RESOLVED>): Boolean
```
luispollo, line 19: Consider sticking with the convention of using a verb to start all method names, as in `isDiffOnlyInEnabled` or similar.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
```diff
@@ -0,0 +1,51 @@
...
+ * Common cluster functionality.
+ * This class contains abstracted logic where we want all clusters to make the
+ * same decision, or abstracted logic involving coordinating resources or
+ * publishing events.
+ */
+abstract class BaseClusterHandler<SPEC: ResourceSpec, RESOLVED: Any>(
+  resolvers: List<Resolver<*>>
+) : ResolvableResourceHandler<SPEC, Map<String, RESOLVED>>(resolvers) {
+
+  abstract fun getDesiredRegion(diff: ResourceDiff<RESOLVED>): String
```
luispollo, line 18: Consider adding doc comments for the functions in the base class since they'll be reused/overridden.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
emjburns: Well, this is only for clusters. I don't want to remove the resolvable resource handler because it's useful for other resource types I think

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/Ec2BaseClusterHandlerTests.kt
```diff
@@ -0,0 +1,125 @@
...
+      taskLauncher = taskLauncher,
+      eventPublisher = eventPublisher,
+      resolvers = resolvers,
+      clusterExportHelper = clusterExportHelper
+    ))
+
+  override fun getRegions(resource: Resource<ClusterSpec>): List<String> =
+    resource.spec.locations.regions.map { it.name }.toList()
+
+  override fun getResourceOneRegion(): Resource<ClusterSpec> {
```
luispollo, line 73: Suggestion:
```suggestion
  override fun getSingleRegionCluster(): Resource<ClusterSpec> {
```

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/Ec2BaseClusterHandlerTests.kt
```diff
@@ -0,0 +1,125 @@
...
+    return Resource(
+      kind = EC2_CLUSTER_V1_1.kind,
+      metadata = metadata,
+      spec = baseSpec
+    )
+  }
+
+  override fun getResourceTwoRegions(): Resource<ClusterSpec> {
```
luispollo, line 81: Suggestion:
```suggestion
  override fun getMultiRegionCluster(): Resource<ClusterSpec> {
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
emjburns: 👍  that's better

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandlerTests.kt
```diff
@@ -0,0 +1,96 @@
...
+        createSpyHandler(resolvers = resolvers, clock = clock, eventPublisher = eventPublisher),
+      )
+    }
+
+    after {
+      clearAllMocks()
+    }
+
+    context("testing whether handler will take action") {
+      context("diff in more then just enabled") {
```
luispollo, line 57: As you know, I've struggled a bit with the naming of this property and related code. Maybe try for a bit more self-explanatory description in the test at least? 🙂 

Perhaps `"diff in more than just server group enabled status"` or something along those lines?

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandlerTests.kt
```diff
@@ -0,0 +1,96 @@
...
+        test("handler will take action") {
+          val resource = getResourceOneRegion()
+          val diff = getDiffInMoreThanEnabled()
+          val response = runBlocking { handler.willTakeAction(resource, diff) }
+          expectThat(response.willAct).isTrue()
+        }
+      }
+
+      context("diff only in enabled") {
```
luispollo, line 66: Same comment as above on the test description.

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandlerTests.kt
```diff
@@ -0,0 +1,96 @@
...
+            val response = runBlocking { handler.willTakeAction(resource, diff) }
+            expectThat(response.willAct).isTrue()
+          }
+        }
+
+        context("one region unhealthy") {
+          before {
+            every { handler.getUnhealthyRegionsForActiveServerGroup(any()) } returns getRegions(getResourceOneRegion())
+          }
+          test("handler will not action") {
```
luispollo, line 83:
```suggestion
          test("handler will not take action") {
```

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusBaseClusterHandlerTests.kt
```diff
@@ -0,0 +1,117 @@
...
+import com.netflix.spinnaker.keel.orca.OrcaService
+import com.netflix.spinnaker.keel.titus.TITUS_CLUSTER_V1
+import com.netflix.spinnaker.keel.titus.TitusClusterHandler
+import com.netflix.spinnaker.keel.titus.byRegion
+import com.netflix.spinnaker.keel.titus.resolve
+import io.mockk.mockk
+import io.mockk.spyk
+import java.time.Clock
+
+class TitusBaseClusterHandlerTests : BaseClusterHandlerTests<TitusClusterSpec, TitusServerGroup, TitusClusterHandler>() {
```
luispollo, line 30: Why does this need to be a new test class? Can't we reuse `TitusClusterHandlerTests`?

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/Ec2BaseClusterHandlerTests.kt
```diff
@@ -0,0 +1,125 @@
...
+import com.netflix.spinnaker.keel.clouddriver.CloudDriverCache
+import com.netflix.spinnaker.keel.clouddriver.CloudDriverService
+import com.netflix.spinnaker.keel.diff.DefaultResourceDiff
+import com.netflix.spinnaker.keel.orca.ClusterExportHelper
+import com.netflix.spinnaker.keel.orca.OrcaService
+import io.mockk.mockk
+import io.mockk.spyk
+import java.time.Clock
+
+class Ec2BaseClusterHandlerTests : BaseClusterHandlerTests<ClusterSpec, ServerGroup, ClusterHandler>() {
```
luispollo, line 29: Why does this need to be a new test class? Can't we reuse `ClusterHandlerTests`?

luispollo (commented):
Awesome cleanup, @emjburns! Thanks for doing this. Got a couple of questions/suggestions, will take another look again later.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
luispollo: Fair enough. Was just pointing out that that property used to be in `ResourceHandler` (which was an abstract class before), and I moved it to `ResolvableResourceHandler` when working on the Treasure plugin since it doesn't use resolvers and I needed to convert `ResourceHandler` to an interface. But you're right -- it may be useful some day.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/Ec2BaseClusterHandlerTests.kt
emjburns: much better

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/Ec2BaseClusterHandlerTests.kt
emjburns: much better

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusBaseClusterHandlerTests.kt
emjburns: `TitusClusterHandlerTests` is super duper complicated, and I don't want to add to it. I want to test this specific functionality in its own test because it is the implementation of the base tests. Same comment with both classes

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandlerTests.kt
emjburns: will update wording. i agree that it's not the best

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusBaseClusterHandlerTests.kt
luispollo: I hear you. I think we have a tendency to structure test classes to match the scope of an entire target class that's being tested, but there's no rule around that. I'm not opposed to breaking up large/complex test classes into smaller ones that are easier to reason about. The only downside I can think of is that distributed is always harder to find than centralized, so people might get confused about what test class to add to, so we should be mindful about the names of the test classes to make that as obvious as possible. Carry on. 🙂

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandlerTests.kt
```diff
@@ -54,22 +54,22 @@ abstract class BaseClusterHandlerTests<
...
-          val resource = getResourceOneRegion()
+          val resource = getSingleRegionCluster()
           val diff = getDiffInMoreThanEnabled()
           val response = runBlocking { handler.willTakeAction(resource, diff) }
           expectThat(response.willAct).isTrue()
         }
       }
 
-      context("diff only in enabled") {
+      context("diff only in number of server groups enabled") {
```
luispollo, line 66: Much better!

luispollo (approved):
LGTM!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
```diff
@@ -0,0 +1,62 @@
...
+ * publishing events.
+ */
+abstract class BaseClusterHandler<SPEC: ResourceSpec, RESOLVED: Any>(
+  resolvers: List<Resolver<*>>
+) : ResolvableResourceHandler<SPEC, Map<String, RESOLVED>>(resolvers) {
+
+  /**
+   * parses the desired region from a resource diff
+   */
+  abstract fun getDesiredRegion(diff: ResourceDiff<RESOLVED>): String
```
robfletcher, line 21: Is that just returning the region name? Or is it some chunk of diff JSON?

Also both implementations are identical so I'm not sure why it's abstract.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -561,11 +549,14 @@ class ClusterHandler(
...
    * @return true if the only difference is in the onlyEnabledServerGroup property
    */
-  private fun ResourceDiff<ServerGroup>.isEnabledOnly(): Boolean =
+  fun ResourceDiff<ServerGroup>.isEnabledOnly(): Boolean =
     current != null &&
       affectedRootPropertyNames.all { it == "onlyEnabledServerGroup" } &&
       current!!.onlyEnabledServerGroup != desired.onlyEnabledServerGroup
 
+  override fun isDiffOnlyInEnabled(diff: ResourceDiff<ServerGroup>): Boolean =
+    diff.isEnabledOnly()
```
robfletcher, line 558: I'm confused why there's an extension function and then an abstract method whose implementations are identical and just call the extension function. Why not just call that extension function directly?

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandlerTests.kt
```diff
@@ -0,0 +1,96 @@
...
+import io.mockk.clearAllMocks
+import io.mockk.every
+import io.mockk.mockk
+import kotlinx.coroutines.runBlocking
+import strikt.api.expectThat
+import strikt.assertions.isFalse
+import strikt.assertions.isTrue
+import java.time.Clock
+
+abstract class BaseClusterHandlerTests<
```
robfletcher, line 19: It feels to me like this class is in the wrong module. The `keel-test` module was intended to be just common testing things, and `keel-core-test` would contain base classes for common tests (e.g. the repository tests).

keel-titus-plugin/keel-titus-plugin.gradle.kts
```diff
@@ -24,4 +24,5 @@ dependencies {
   testImplementation("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml")
   testImplementation("org.funktionale:funktionale-partials")
   testImplementation("org.apache.commons:commons-lang3")
+  testImplementation(project(":keel-core-test"))
```
robfletcher, line 27: Not sure why this dependency is added when it looks like you did that to get access to `MutableClock` but then moved `MutableClock` to `keel-test`

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/resource/TitusBaseClusterHandlerTests.kt
robfletcher: I'm all for breaking tests up around behavior. I'd like to also break the handlers up as I think they do not do "just one thing" but bloated tests cause more direct pain.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/AgentLockRepositoryTests.kt
```diff
@@ -1,6 +1,6 @@
 package com.netflix.spinnaker.keel.persistence
 
-import com.netflix.spinnaker.time.MutableClock
+import com.netflix.spinnaker.keel.time.MutableClock
```
robfletcher, line 3: Could we even move `MutableClock` to kork? The reason I initially put it outside the `keel` package hierarchy is that it's not something directly to do with Keel.

robfletcher (commented):
A few comments and questions about specifics but I like the general direction

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
emjburns: just region name. Implementations are identical, yes, but that's because the spec is structured these same for both classes. You need to have type information in order to actually get to the region

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
emjburns: Well, my thought was that I didn't know how to stub or mock extension functions. But, since I am not doing that for this guy I should do I the other way.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/AgentLockRepositoryTests.kt
emjburns: oh what do you know, it's already in kork: https://github.com/spinnaker/kork/blob/master/kork-test/src/main/java/com/netflix/spinnaker/kork/test/time/MutableClock.java

robfletcher: 😮

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
robfletcher: Is that something we could unify?

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandlerTests.kt
emjburns: i'll move it.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/BaseClusterHandler.kt
emjburns: Only by making a base cluster spec, which I don't want to do right now / in this PR

keel-titus-plugin/keel-titus-plugin.gradle.kts
emjburns: This module is needed now so that the titus plugin can implement the base cluster test

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/AgentLockRepositoryTests.kt
emjburns: However, I'm not going to update this PR to use the kork version :D sorry. I've backed out all the changes to the mutable clock class because the weren't needed

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/Ec2BaseClusterHandlerTests.kt
robfletcher, line 29: Shouldn't this just be called `Ec2ClusterHandlerTests`? Why `Base`?

emjburns: I'm just following the pattern you set for all the repository tests. Child class name is the thing + the parent class name

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/AgentLockRepositoryTests.kt
robfletcher: Oh lord, is it different?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -117,43 +117,31 @@ class ClusterHandler(
...
-    // we can't take any action if there is more than one active server group
-    //  AND the current active server group is unhealthy
-    val potentialInactionableRegions = mutableListOf<String>()
-    val inactionableRegions = mutableListOf<String>()
-    resourceDiff.toIndividualDiffs().forEach { diff ->
-      if (diff.hasChanges() && diff.isEnabledOnly()) {
-        potentialInactionableRegions.add(diff.desired.location.region)
-      }
+  override fun getDesiredRegion(diff: ResourceDiff<ServerGroup>): String =
+    diff.desired.location.region
```
gal-yardeni, line 121: can any of those be null or empty?

emjburns: no, other wise kotlin wouldn't let me do this

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/AgentLockRepositoryTests.kt
emjburns: I think there are some added extension functions in the keel one (like 'tick' vs increment)

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/Ec2BaseClusterHandlerTests.kt
emjburns: @robfletcher ohhhh. I see.

PR #1624, opened 2020-10-27
Default name for self-referential ingress rules

robfletcher:
Previously we required that users omit the `name` propertly in order to make a `ReferenceRule` self-referential. However, this results in spurious unresolvable diffs if the user _does_ specify the name. The cause is not obvious.

It makes more sense to default the name so that users _may_ omit it or supply it as they prefer.

### DIFF
--- a/keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupRule.kt
+++ b/keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupRule.kt
@@ -16,6 +16,7 @@
 package com.netflix.spinnaker.keel.api.ec2
 
 import com.netflix.spinnaker.keel.api.schema.Literal
+import com.netflix.spinnaker.keel.api.schema.Optional
 
 sealed class SecurityGroupRule {
   abstract val protocol: Protocol
@@ -24,18 +25,14 @@ sealed class SecurityGroupRule {
   enum class Protocol {
     ALL, TCP, UDP, ICMP
   }
-
-  open val isSelfReference: Boolean = false
 }
 
 data class ReferenceRule(
   override val protocol: Protocol,
-  val name: String? = null,
+  @Optional("defaults to the name of the security group the rule belongs to")
+  val name: String,
   override val portRange: IngressPorts
-) : SecurityGroupRule() {
-  override val isSelfReference: Boolean
-    get() = name == null
-}
+) : SecurityGroupRule()
 
 data class CrossAccountReferenceRule(
   override val protocol: Protocol,
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/SecurityGroupSpecDeserializer.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/SecurityGroupSpecDeserializer.kt
@@ -1,0 +1,34 @@
+package com.netflix.spinnaker.keel.ec2.jackson
+
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.InjectableValues.Std
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.deser.std.StdNodeBasedDeserializer
+import com.fasterxml.jackson.module.kotlin.convertValue
+import com.netflix.spinnaker.keel.api.Moniker
+import com.netflix.spinnaker.keel.api.ec2.SecurityGroupSpec
+import com.netflix.spinnaker.keel.core.name
+
+class SecurityGroupSpecDeserializer : StdNodeBasedDeserializer<SecurityGroupSpec>(SecurityGroupSpec::class.java) {
+  override fun convert(root: JsonNode, context: DeserializationContext) =
+    with(context.parser.codec as ObjectMapper) {
+      val moniker: Moniker = convertValue(root.path("moniker"))
+      SecurityGroupSpec(
+        moniker = moniker,
+        locations = convertValue(root.path("locations"))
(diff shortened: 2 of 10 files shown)


keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupRule.kt
```diff
@@ -24,18 +25,14 @@ sealed class SecurityGroupRule {
   enum class Protocol {
     ALL, TCP, UDP, ICMP
   }
-
-  open val isSelfReference: Boolean = false
```
robfletcher, line 28: Would need a backreference property for the `SecurityGroupSpec` in order to have a val like this and it was only used in one place, so I just inlined it

keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupRule.kt
```diff
@@ -24,18 +25,14 @@ sealed class SecurityGroupRule {
...
 }
 
 data class ReferenceRule(
   override val protocol: Protocol,
-  val name: String? = null,
+  @Optional("defaults to the name of the security group the rule belongs to")
+  val name: String,
```
robfletcher, line 33: `name` is non-nullable now. The `@Optional` annotation just tells the schema generator not to mark it as required.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/SecurityGroupSpecDeserializer.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.netflix.spinnaker.keel.core.name
+
+class SecurityGroupSpecDeserializer : StdNodeBasedDeserializer<SecurityGroupSpec>(SecurityGroupSpec::class.java) {
+  override fun convert(root: JsonNode, context: DeserializationContext) =
+    with(context.parser.codec as ObjectMapper) {
+      val moniker: Moniker = convertValue(root.path("moniker"))
+      SecurityGroupSpec(
+        moniker = moniker,
+        locations = convertValue(root.path("locations"))
+          ?: context.findInjectableValue("locations"),
```
robfletcher, line 20: Because custom deserialization ignores existing mixins we need to do this explicitly

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/SecurityGroupSpecDeserializer.kt
```diff
@@ -0,0 +1,34 @@
...
+        description = root.get("description")?.textValue(),
+        inboundRules = copy().run {
+          injectableValues = Std(mapOf("name" to moniker.name))
+          root.get("inboundRules")?.let { convertValue(it) } ?: emptySet()
+        },
```
robfletcher, line 25: This creates a unique mapper for the sub-tree with an injectable value based on the moniker we just parsed

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/mixins/ReferenceRuleMixin.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.keel.ec2.jackson.mixins
+
+import com.fasterxml.jackson.annotation.JacksonInject
+
+internal interface ReferenceRuleMixin {
+  @get:JacksonInject("name")
+  val name: String
```
robfletcher, line 7: This is how the parsed `ReferenceRule` will pick up the injected `name` if one is not supplied

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -289,7 +290,7 @@ class SecurityGroupHandler(
                   )
                   else -> ReferenceRule(
                     protocol,
-                    if (ingressGroup.name == name) null else ingressGroup.name, // if it's a self-referential rule keel models the name as null
+                    ingressGroup.name,
```
robfletcher, line 293: No longer necessary to null out the value coming back from River if it's the same as the current security group name

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -333,7 +334,7 @@ class SecurityGroupHandler(
           // we have to do a 2-phase create for self-referencing ingress rules as the referenced
           // security group must exist prior to the rule being applied. We filter then out here and
           // the subsequent diff will apply the additional group(s).
-          .filterNot { it.isSelfReference }
+          .filterNot { it is ReferenceRule && it.name == moniker.name }
```
robfletcher, line 337: Because I had to remove that property from `SecurityGroupRule`

emjburns (approved):

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
luispollo: I'm struggling a bit to understand how this would address the case we saw with Adam's delivery config, where the name he used did not seem to be derived from the moniker (name was just `spinnakerstatus`). Or was it just that he had no stack in the moniker?

robfletcher: correct

PR #1626, opened 2020-10-28
fix(deserialization): fix NPE in nested deserialization

robfletcher:
Jackson's fun API strikes again


luispollo (approved):
I don't think anyone will be able to truly review this, so here's my vote of confidence! 😄

PR #1627, opened 2020-10-28
fix(appversion): fix for wrong version parsing

gal-yardeni:
we actually need the artifact name to parse the version with Frigga.

### DIFF
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
@@ -36,14 +36,6 @@ class DebianArtifactSupplier(
   override val supportedVersioningStrategy =
     SupportedVersioningStrategy("deb", DebianVersioningStrategy::class.java)
 
-  override fun publishArtifact(artifact: PublishedArtifact) {
-    if (artifact.hasReleaseStatus()) {
-      super.publishArtifact(artifact)
-    } else {
-      log.debug("Ignoring artifact event without release status: $artifact")
-    }
-  }
-
   override fun getLatestArtifact(deliveryConfig: DeliveryConfig, artifact: DeliveryArtifact): PublishedArtifact? =
     runWithIoContext {
       artifactService.getVersions(artifact.name, artifact.statusesForQuery, DEBIAN)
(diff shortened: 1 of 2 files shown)


keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -36,14 +36,6 @@ class DebianArtifactSupplier(
   override val supportedVersioningStrategy =
     SupportedVersioningStrategy("deb", DebianVersioningStrategy::class.java)
 
-  override fun publishArtifact(artifact: PublishedArtifact) {
```
gal-yardeni, line 39: we don't need this (it's duplicated to the `shouldProcess` logic below).

emjburns (approved):

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
luispollo: This is pretty surprising to me, given the description of the PR. This is the function that gets called by `ArtifactController` to publish a new artifact. Has the base class been modified to call `shouldProcessArtifact`?

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplierTests.kt
```diff
@@ -174,6 +182,11 @@ internal class DebianArtifactSupplierTests : JUnit5Minutests {
           .isTrue()
       }
 
+      test ("should process artifact successfully2") {
```
luispollo, line 185:
```suggestion
      test ("should process artifact successfully") {
```

gal-yardeni: I have 2 tests with the same name, so this is 2 😄  will fix later on

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplierTests.kt
```diff
@@ -52,6 +52,14 @@ internal class DebianArtifactSupplierTests : JUnit5Minutests {
       metadata = mapOf("releaseStatus" to SNAPSHOT, "buildNumber" to "1", "commitId" to "a15p0")
     )
 
+    val actualVersionInPublishedArtifact = PublishedArtifact(
```
luispollo, line 55: If this is the actual format of a Debian published artifact (i.e. without the name part in the `version` string), then I think a better approach would be to modify `latestArtifact` above and just keep the existing test.

gal-yardeni: yeah good point.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
gal-yardeni: `artifactListener` is calling this. It was in previous PRs.

PR #1628, opened 2020-10-28
feat(ec2): EBS volumes: default to SSD 

lorin:
# Problem

Occasionally, we see a keel instances get "stuck": they stop actuating for several minutes, and stop reporting metrics. This is always accompanied by a spike in disk load.

One theory is that the problem is related to the older [magnetic][1]  EBS volume types (denoted `"standard"`) that clouddriver defaults to, as defined in [AwsConfiguration.groovy][2].

We'd like to run keel using an SSD-based EBS volume type, such as "gp2", which is now the default AWS volume type.

Unfortunately, we can't just specify volume type in keel's delivery configuration, because the current implementation of LaunchConfigurationSpec does not allow us to specify block device mappings.

# Proposed solution

Change the EBS volume default to "gp2", which is SSD-based, and is the current AWS default volume type.

Note: this change will configure all EC2 clusters to use "gp2" (SSD) for their EBS volume types instead of "standard" (magnetic). 

# Overview of changes

The only behavioral change is in ClusterHandler.kt, where we add a new `blockDevices` field to the payload for the orca job that creates a new server group.

The configuration info for EBS volumes is encapsulated in a new class: BlockDeviceConfig

# Test plan

1. Deploy keel to staging environment
1. Query the volume configuration of an instance from a managed EC2 test app
1. Trigger a deploy of the test app so keel creates a new ASG for the test app
1. Verify that:
   - The mounted EBS volume is of type `gp2`
   - The volume configuration is otherwise unchanged

We can check info of mounted volumes for an instance-id (say, `i-0123456789ABCDEF0`) using the following command:

```bash
$ aws ec2 describe-volumes --region us-east-1 \
 --filters Name=attachment.instance-id,Values=i-0123456789ABCDEF0
```

# FAQ

## Why does this diff include all of that information about volume sizes? The text above doesn't mention volume size at all.

If we want to change the EBS volume type, we need to pass [block device][3] objects. Each object has a "size" field that is required. I tested that the field is required by creating a pipeline where I edited the deploy stage to customize the volume type without specifying the size, like this:

```json
     "blockDevices": [
        {
          "deviceName": "/dev/sdb",
          "volumeType": "gp2"
        }
      ],
```

Alas, the pipeline failed with the following error:

```
EBS device /dev/sdb missing required value size
```

Clouddriver determines the default EBS volume size by instance type. I configured keel to use the same EBS volume sizes as clouddriver uses by copying over the table of values from clouddriver's [BlockDeviceConfig][4] class.

## Why not modify keel to allow users to specify a custom block device mapping instead of changing the default?

It felt more managed-delivery-ish to change an infrastructure detail behind the scenes than to introduce another low-level implementation detail. Everyone running on Managed Delivery now gets SSD-backed volumes! 

## Will this change trigger diffs/actuations for all EC2 clusters when it rolls out?

No, this change shouldn't result in any new "diffs", because we aren't encoding the volume type in the representation of a server group that gets diffed. We're just adding some new parameters to the payload of the CreateServerGroup orca job.

## Why is the volume type called "gp2"?

The "gp" stands for "general purpose", I believe. No idea where the "2" comes from.

# Where I'm looking for feedback

Is this a reasonable approach? Other approaches we could take are:

* Expose block device config in the delivery config instead
* Modify clouddriver so that we can specify a volume type without a size

# References

* [Amazon EBS volume types docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html)

[1]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html#EBSVolumeTypes_standard
[2]: https://github.com/spinnaker/clouddriver/blob/a54baac178fe7427ea8589334c84bfa136e0dac4/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/config/AwsConfiguration.groovy#L162
[3]: https://github.com/spinnaker/clouddriver/blob/29c5c1435ab0c750c4b491208f600647a3e98901/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonBlockDevice.groovy
[4]: https://github.com/spinnaker/clouddriver/blob/6338a60c2f6736570d71e8ae7072d3d266f52fff/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/BlockDeviceConfig.groovy


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/BlockDeviceConfig.kt
```diff
@@ -0,0 +1,195 @@
...
+)
+
+
+/**
+ * Default config info for EBS block devices
+ *
+ * Note: this class only has config data for a subset of AWS instance types. In particularly, keel only knows
+ * about instance types where the clouddriver default is to mount a single EBS volume at "/dev/sdb"
+ */
+@Component
```
luispollo, line 29: I think you need to put components in the `com.netflix.spinnaker.config` package for them to be automatically picked up by Spring while scanning?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/BlockDeviceConfig.kt
```diff
@@ -0,0 +1,195 @@
...
+          deviceName=config.deviceName,
+          size=size,
+          deleteOnTermination=config.deleteOnTermination)
+      )
+  }
+
+  /**
+   * Return the size of the EBS volume, in GB
+   *
+   * Copied over from:
```
luispollo, line 54: The redundancy here bothers me a bit. Did you consider exposing an API in clouddriver to return this data, which we could then just cache in keel?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/BlockDeviceConfig.kt
```diff
@@ -0,0 +1,195 @@
...
+  private val config : VolumeDefaultConfiguration
+)
+ {
+
+  /**
+   * Get the list of block devices for this instance type
+   * For instance types unsupported by this class, keel doesn't pass EBS config info, which results in falling back to
+   * clouddriver's defaults
+   */
+  fun get(instanceType: String): List<BlockDevice>? =
```
luispollo, line 41: Suggestion:
```suggestion
  fun getBlockDevicesFor(instanceType: String): List<BlockDevice>? =
```

luispollo:
> It felt more managed-delivery-ish to change an infrastructure detail behind the scenes than to introduce another low-level implementation detail. Everyone running on Managed Delivery now gets SSD-backed volumes!

I agree, but also think a principle we've historically tried to adhere to is that we abstract and default things, but still give the user the ability to configure meaningful details when it makes sense. This sees to be a pretty aggressive bet to make without discussing further with the Compute team and/or gathering more data on whether magnetic disks should remain an option. 

I'd lean towards a hybrid approach, where, yes, we default to SSD, but allow the user to override. But it doesn't need to be all the nitty gritty details of the EBS volume either! It could just be a `diskType: magnetic` or `diskType: ssd` property.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/BlockDeviceConfig.kt
luispollo: P.S. I'd be totally OK with making that an improvement for follow-up PRs.

luispollo:
@lorin Forgot to say: your PR descriptions are THE BEST! 🏆 👏

emjburns (approved):
Love this approach. I think that you're right, it's very MD to change this out from under people. It's a good first abstraction.

I think it would be nice to modify clouddriver so that it would auto-select the right size instead of copying that logic into keel. That can totally be a future pr though.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/BlockDeviceConfig.kt
lorin: Yeah, this change felt like the quickest way to evaluate whether this actually resolves the issue for us. If we end up keeping this functionality, I'll coordinate with the clouddriver devs to see if we can design a better long-term solution.

luispollo (approved):
Based on internal discussions, I don't object to merging. However, I still think we should make the volume type configurable to future-proof our implementation, but that can be a separate PR. Thanks for your due diligence researching and following through to fix the original issue that motivated this change.

PR #1629, opened 2020-10-29
Graph time spent doing environment check by app

robfletcher:
To help us pin down problematic apps.

### DIFF
--- a/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionCheckerTests.kt
+++ b/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionCheckerTests.kt
@@ -174,7 +174,7 @@ internal class NewEnvironmentPromotionCheckerTests : JUnit5Minutests {
 
           test("an event is not sent") {
             verify(exactly = 0) {
-              publisher.publishEvent(any<ArtifactVersionApproved>())
+              publisher.publishEvent(ofType<ArtifactVersionApproved>())
             }
           }
         }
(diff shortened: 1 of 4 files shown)


keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionCheckerTests.kt
```diff
@@ -174,7 +174,7 @@ internal class NewEnvironmentPromotionCheckerTests : JUnit5Minutests {
 
           test("an event is not sent") {
             verify(exactly = 0) {
-              publisher.publishEvent(any<ArtifactVersionApproved>())
+              publisher.publishEvent(ofType<ArtifactVersionApproved>())
```
robfletcher, line 177: Turns out that's not how you do it

emjburns: oh interesting

robfletcher: Yeah, I fixed a bunch more in https://github.com/spinnaker/keel/pull/1631

robfletcher: `any()` is fine in some situations, but if you want to strictly match the type of the parameter that actually gets passed, that's not what it does.

lorin (approved):
Nice and elegant.

PR #1631, opened 2020-10-29
Update inappropriate use of any() to ofType() in tests

robfletcher:
`any()` just satisfies the compile-time type check, it doesn't assert that the runtime type of the parameter matches the generic. For that you need `ofType()`. I just discovered this yesterday. I've updated tests that _are_ attempting to verify the parameter type to use the correct matcher function.


lorin (approved):
Nice catch. How did you find out about this behavior?

gal-yardeni (approved):

gal-yardeni:
Same question as @lorin mentioned :)

robfletcher:
@lorin @gal-yardeni When I made the changes for #1629 I added a new event and broke a couple of tests that were using `any()` instead of `ofType()`.

PR #1633, opened 2020-10-30
fix(dockerTimestamp): addressing docker timestamp as a string

gal-yardeni:


luispollo (approved):
LGTM!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
```diff
@@ -52,18 +52,18 @@ data class PublishedArtifact(
...
-      when (it) {
-        is Long -> Instant.ofEpochMilli(it) // to accommodate for artifact events from CI integration
-        is Instant -> it
-        else -> null
+      ?: metadata["date"])
+      ?.let {
+        when (it) {
+          is Long -> Instant.ofEpochMilli(it) // to accommodate for artifact events from CI integration
+          is Instant -> it
+          is String -> Instant.ofEpochMilli(it.toLong())
```
luispollo, line 63: You might want to wrap this in a try/catch in case there's a format issue in the string.

gal-yardeni: done

PR #1635, opened 2020-10-30
Ensure resources in environment check cycle are migrated

robfletcher:
Because `ClusterV1Spec` is not a `ArtifactReferenceProvider` the `ArtifactUsedConstraintEvaluator` did not find the artifacts they use. We were accidentally not migrating resources read via `deliveryConfigsDueForCheck` and attempting to check v1 clusters instead of always checking v1.1 clusters.


gal-yardeni (approved):
lgtm

PR #1636, opened 2020-11-03
tidy up migration fix

robfletcher:
Tidy up PR for #1635. Adds a test for migrating resources when delivery configs are checked, and unifies logic for doing so across both repositories that need it.

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ResourceFactory.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ResourceFactory.kt
@@ -1,0 +1,38 @@
+package com.netflix.spinnaker.keel.sql
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
+import com.netflix.spinnaker.keel.api.Resource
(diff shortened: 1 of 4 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ResourceFactory.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
+import com.netflix.spinnaker.keel.api.Resource
+import com.netflix.spinnaker.keel.api.ResourceKind
+import com.netflix.spinnaker.keel.resources.ResourceSpecIdentifier
+import com.netflix.spinnaker.keel.resources.SpecMigrator
+import com.netflix.spinnaker.keel.resources.migrate
+
+internal class ResourceFactory(
```
gal-yardeni, line 11: I like this approach. Can you add some comments here, like about when to use it?

robfletcher: Done!

gal-yardeni (approved):

PR #1639, opened 2020-11-04
Don't serialize resources at top level of delivery config

robfletcher:
I happened to notice the resources are included as an array at the top level of the delivery config JSON/YAML response as well as being under their respective environments. This is because `DeliveryConfig.resources` is a transient property not being ignored by Jackson.


luispollo (approved):
Good spot. There's a couple more synthetic properties like that in other places that I'd like to ignore too, will do that soon-ish.

PR #1640, opened 2020-11-04
add some logging so we can see what artifact used constraint is doing

robfletcher:
Seems like this behavior is vetoing artifact versions and it's not clear why


gal-yardeni (approved):
thanks for adding this!

luispollo (approved):
Thank you! 🙏

PR #1641, opened 2020-11-05
fix(veto): Do not mark never-deployed versions as bad

luispollo:
### Context
One of our customers reported the following problems:
> - The automatic ‘mark as bad’ logic can mark artifacts bad that have never been deployed and is what’s blocking me
> -  There’s no way to manually ‘mark as good’ an artifact that was actually fine, but deployment failed due to a transient failure
> - The candidate waiting for approval to production (build 16) is the build currently healthy in test, despite it being marked bad. Looks like it went healthy just a couple of minutes after the deployment timed out
> - The automatic ‘mark as bad’ doesn’t mark bad for the production environment, so I bet if I accidentally approved build 16 it’d try to deploy. I guess I’d expect it to mark both as bad to prevent accidental deployment, or at least have the approval workflow not fire
> - Spinnaker has tried to deploy the last version not marked bad (build 12) twice as a fallback after the other artifacts were taken out of consideration, but it itself hasn’t been marked as bad due to failing deployment

I believe these problems were caused by a combination of issues, a "perfect storm" where a resource was stuck in an unhappy state, and new artifact versions were approved for deployment into that resource. This PR is specifically intended to address the first problem on the list above.

### Problem
The logic that decides whether to mark an artifact version as bad as a result of the deployment target resource being unhappy was checking whether that particular artifact version was either currently or had been previously successfully deployed. However, this leaves out several other promotion statuses, such as when the artifact version has just been approved for deployment.

### Proposed solution
In my understanding, the only promotion status in which we should actually consider marking a version as bad is `DEPLOYING`, meaning we've been attempting to deploy the artifact to the corresponding resource, and the deployment has failed. All other promotion statuses should be ignored for the purpose of this check.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
@@ -13,14 +13,15 @@ import com.netflix.spinnaker.keel.api.plugins.ResourceHandler
 import com.netflix.spinnaker.keel.api.plugins.supporting
 import com.netflix.spinnaker.keel.core.ResourceCurrentlyUnresolvable
 import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactVeto
+import com.netflix.spinnaker.keel.core.api.PromotionStatus.DEPLOYING
 import com.netflix.spinnaker.keel.diff.DefaultResourceDiff
 import com.netflix.spinnaker.keel.events.ResourceActuationLaunched
-import com.netflix.spinnaker.keel.events.ResourceDiffNotActionable
 import com.netflix.spinnaker.keel.events.ResourceActuationVetoed
 import com.netflix.spinnaker.keel.events.ResourceCheckError
 import com.netflix.spinnaker.keel.events.ResourceCheckUnresolvable
 import com.netflix.spinnaker.keel.events.ResourceDeltaDetected
 import com.netflix.spinnaker.keel.events.ResourceDeltaResolved
+import com.netflix.spinnaker.keel.events.ResourceDiffNotActionable
 import com.netflix.spinnaker.keel.events.ResourceMissing
 import com.netflix.spinnaker.keel.events.ResourceTaskFailed
 import com.netflix.spinnaker.keel.events.ResourceTaskSucceeded
@@ -228,14 +229,10 @@ class ResourceActuator(
           val artifact = deliveryConfig.matchingArtifactByName(versionedArtifact.artifactName, versionedArtifact.artifactType)
             ?: error("Artifact ${versionedArtifact.artifactType}:${versionedArtifact.artifactName} not found in delivery config ${deliveryConfig.name}")
 
-          val previousSuccess = artifactRepository.wasSuccessfullyDeployedTo(
-            deliveryConfig = deliveryConfig,
-            artifact = artifact,
-            version = versionedArtifact.artifactVersion,
-            targetEnvironment = environment.name
-          )
+          val promotionStatus = artifactRepository.getArtifactPromotionStatus(
+            deliveryConfig, artifact, versionedArtifact.artifactVersion, environment.name)
 
-          if (!previousSuccess) {
+          if (promotionStatus == DEPLOYING) {
             artifactRepository.markAsVetoedIn(
               deliveryConfig = deliveryConfig,
               veto = EnvironmentArtifactVeto(
@@ -256,7 +253,7 @@ class ResourceActuator(
             publisher.publishEvent(ArtifactVersionVetoed(resource.application))
           } else {
             log.info(
-              "Not vetoing artifact version {} of artifact {} for config {} and env {} because it was previously successfully deployed",
+              "Not vetoing artifact version {} of artifact {} for config {} and env {} because it's not currently deploying",
               versionedArtifact,
               artifact.reference + ":" + artifact.type,
               deliveryConfig.name,
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
@@ -52,8 +52,8 @@ class UnhappyVeto(
     val resourceId = resource.id
     val application = resource.application
 
-    // maxDiff count represents the number of times we're allowed to see a diff and take action to try and fix it.
-    if (diffFingerprintRepository.actionTakenCount(resourceId) <= maxDiffCount(resource)) {
+    // maxActionCountPerDiff represents the number of times we're allowed to see a diff and take action to try and fix it.
+    if (diffFingerprintRepository.actionTakenCount(resourceId) <= maxActionCountPerDiff(resource)) {
       unhappyVetoRepository.delete(resourceId)
       return allowedResponse()
     }
@@ -90,7 +90,7 @@ class UnhappyVeto(
   override fun currentRejectionsByApp(application: String) =
     unhappyVetoRepository.getAllForApp(application).toList()
 
-  private fun maxDiffCount(resource: Resource<*>) =
+  private fun maxActionCountPerDiff(resource: Resource<*>) =
     when (resource.spec) {
       is UnhappyControl -> (resource.spec as UnhappyControl).maxDiffCount ?: maxDiffCount
       else -> maxDiffCount
@@ -124,7 +124,7 @@ class UnhappyVeto(
     }
 
   private fun unhappyMessage(resource: Resource<*>): String {
-    val maxDiffs = maxDiffCount(resource)
+    val maxDiffs = maxActionCountPerDiff(resource)
     val waitingTime = waitingTime(resource)
     if (waitingTime == Duration.ZERO) {
       return "Resource is unhappy and our $maxDiffs actions have not fixed it. " +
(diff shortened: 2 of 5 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/veto/unhappy/UnhappyVeto.kt
```diff
@@ -52,8 +52,8 @@ class UnhappyVeto(
     val resourceId = resource.id
     val application = resource.application
 
-    // maxDiff count represents the number of times we're allowed to see a diff and take action to try and fix it.
-    if (diffFingerprintRepository.actionTakenCount(resourceId) <= maxDiffCount(resource)) {
+    // maxActionCountPerDiff represents the number of times we're allowed to see a diff and take action to try and fix it.
+    if (diffFingerprintRepository.actionTakenCount(resourceId) <= maxActionCountPerDiff(resource)) {
```
luispollo, line 56: Tangentially related. This was a misnomer, IMO, especially since we actually have a column in the `diff_fingerprint` table with the diff count (not to be confused with `count_actions_taken`, which is what this code is about).

luispollo: P.S. We should consider changing the corresponding spec property as well in a future PR.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
```diff
@@ -228,14 +229,10 @@ class ResourceActuator(
...
-            deliveryConfig = deliveryConfig,
-            artifact = artifact,
-            version = versionedArtifact.artifactVersion,
-            targetEnvironment = environment.name
-          )
+          val promotionStatus = artifactRepository.getArtifactPromotionStatus(
+            deliveryConfig, artifact, versionedArtifact.artifactVersion, environment.name)
 
-          if (!previousSuccess) {
+          if (promotionStatus == DEPLOYING) {
```
luispollo, line 235: This was the problem... `wasSuccessfullyDeployedTo` checks for the presence of a record with a `CURRENT` or `PREVIOUS` promotion status, neither of which is the case with a just-approved version.

emjburns: @luispollo we also don't want to reject an artifact that has previously been deployed, so you'll also need to add in the CURRENT and PREVIOUS statuses

emjburns (changes requested):

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: The reason for this is that we don't want to mark something as bad that has been successfully deployed to an environment. In that case the artifact version is most likely totally fine. If we get to the point where the deployment fails of a version that was previously successful, we know it's probably an infra or env thing.

However, now that I'm thinking about it, is your point that artifacts go into the approved then deploying state always, and that we will never actually hit the case that the artifact is in a CURRENT or a PREVIOUS status? If so, I see what you're saying. That just means I think that we will have to track whether a version is successfully deployed in another place....

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -1261,6 +1261,26 @@ class SqlArtifactRepository(
...
   }
 
+  override fun getArtifactPromotionStatus(
+    deliveryConfig: DeliveryConfig,
+    artifact: DeliveryArtifact,
+    version: String,
+    environmentName: String
+  ): PromotionStatus? {
+    return sqlRetry.withRetry(READ) {
+
```
emjburns, line 1271: nit: extra newline

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -1261,6 +1261,26 @@ class SqlArtifactRepository(
...
+      jooq
+        .select(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS)
+        .from(ENVIRONMENT_ARTIFACT_VERSIONS)
+        .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(deliveryConfig.getUidFor(environmentName)))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifact.uid))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.eq(version))
+        .orderBy(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT.desc())
```
emjburns, line 1278: Why do you need to order by the deployed at time? I thought env + artifact + version should be unique in this table.

lorin:
I'm having some trouble understanding the problem, so I have some questions:

> was checking whether that particular artifact version was either currently or had been previously successfully deployed. However, this leaves out several other promotion statuses, such as when the artifact version has just been approved for deployment.

From the PR description, it sounds like we weren't vetoing an artifact in the case where the UnhappyVeto fired and the version was in the `DEPLOYING` state. Is that right? 

> In my understanding, the only promotion status in which we should actually consider marking a version as bad is DEPLOYING,

Based on this comment and the one before, it sounds like we weren't considering the only case we should actually consider. From that I infer that keel was never actually automatically marking as bad resources the way we would expect. Is that actually the case? If so, why wouldn't we have noticed before?

emjburns:
@lorin this is code that I changed a couple of weeks ago (https://github.com/spinnaker/keel/pull/1610), and I think what this PR is describing is that I didn't do it right

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: You're right. This was copy/paste from another function, will remove.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
luispollo: > @luispollo we also don't want to reject an artifact that has previously been deployed, so you'll also need to add in the CURRENT and PREVIOUS statuses

Right. But this is the check for marking the artifact as bad, so we'll _not_ mark it as bad if it's `CURRENT` or `PREVIOUS` (or any of the other statuses, for that matter), only when it's `DEPLOYING`.



> The reason for this is that we don't want to mark something as bad that has been successfully deployed to an environment. In that case the artifact version is most likely totally fine. If we get to the point where the deployment fails of a version that was previously successful, we know it's probably an infra or env thing.
> 
> However, now that I'm thinking about it, is your point that artifacts go into the approved then deploying state always, and that we will never actually hit the case that the artifact is in a CURRENT or a PREVIOUS status? If so, I see what you're saying. That just means I think that we will have to track whether a version is successfully deployed in another place....

Yes, we don't currently have any historical tracking of artifact status transitions (we should probably introduce artifact events at some point). I tried to think of the smallest change I could make to achieve a close-enough behavior to what you were attempting to model before, because I think the real solution here would be to take a step back and revisit the way we're modeling this, as I've brought up before, but that would take much longer. 

With my change, if you get to the resource check and the latest artifact version is not the one that's currently deploying, then we don't touch the artifact. Which includes if the artifact had been previously deployed successfully (`PREVIOUS`) or is currently deployed successfully (`CURRENT`) and for some other reason the cluster is unhappy.

luispollo:
> I'm having some trouble understanding the problem, so I have some questions:
> 
> > was checking whether that particular artifact version was either currently or had been previously successfully deployed. However, this leaves out several other promotion statuses, such as when the artifact version has just been approved for deployment.
> 
> From the PR description, it sounds like we weren't vetoing an artifact in the case where the UnhappyVeto fired and the version was in the `DEPLOYING` state. Is that right?

I hear you. This stuff is pretty hard to reason about, and it's entirely possible that I've misinterpreted the code and symptoms. But, yes, you're right. The original code was calling `artifactRepository.wasSuccessfullyDeployedTo`, which, in turn, looked for a record with a promotion status of `CURRENT` or `PREVIOUS` for that version.

> > In my understanding, the only promotion status in which we should actually consider marking a version as bad is DEPLOYING,
> 
> Based on this comment and the one before, it sounds like we weren't considering the only case we should actually consider. From that I infer that keel was never actually automatically marking as bad resources the way we would expect. Is that actually the case? If so, why wouldn't we have noticed before?

Not resources, but artifact versions. This is where it gets very confusing. One thing is the _resource_ veto (`UnhappyVeto`). Another thing entirely is this logic we have to veto the _artifact version_ if we've tried to deploy that version multiple times (above a configurable threshold, defaulting to 2 for clusters) and that's failed. The problem is with checking "we've tried to deploy _that version_ and failed" part. Given the current modeling of the relationship between resources and artifacts, and the data we keep in our database, that's not very straightforward. This is why I raised the question of modeling that relationship more explicitly in the past, but that's a conversation for another time. 🙂 

Let me know if this helps.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt
emjburns: 👍 sounds good, I definitely read through this a bit too quickly so turns out my concerns were unfounded! 

I'll make a note to think about storing whether a version has been successfully deployed or not

emjburns (approved):

PR #1646, opened 2020-11-06
Skip promotion checks for artifacts the environment doesn't use

robfletcher:
Instead of doing this with an implicit constraint (which will run for _every_ candidate artifact version), this is simply checked once before any checks at all. The problem was that the `ArtifactUsedConstraintEvaluator` would always return the same result for _any_ artifact version (because the determination does not depend on the version in any way). This meant that any _other_ constraints -- specifically `ImageUsedConstraint`, which can be expensive, would also get evaluated for each version. This PR just short-circuits checking for any environment that does not use the artifact.


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ApproveOldVersionTests.kt
```diff
@@ -59,7 +62,12 @@ abstract class ApproveOldVersionTests<T : KeelRepository> : JUnit5Minutests {
 
     val environment: Environment = Environment(
       name = "test",
-      constraints = setOf(ManualJudgementConstraint())
+      constraints = setOf(ManualJudgementConstraint()),
+      resources = setOf(
+        resource(
+          spec = DummyArtifactReferenceResourceSpec(artifactReference = "my-artifact")
+        )
+      )
```
robfletcher, line 70: had to add actual resources to a couple of tests that exercise constraint evaluation

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -27,79 +29,84 @@ class EnvironmentPromotionChecker(
...
+
+      deliveryConfig
+        .artifacts
+        .associateWith { repository.artifactVersions(it) }
+        .forEach { (artifact, versions) ->
+          if (versions.isEmpty()) {
+            log.warn("No versions for ${artifact.type} artifact name ${artifact.name} and reference ${artifact.reference} are known")
+          } else {
+            deliveryConfig.environments.forEach { environment ->
+              if (artifact.isUsedIn(environment)) {
```
robfletcher, line 52: This is the only new bit here, the diff looks big because indentation changed

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -111,6 +118,15 @@ class EnvironmentPromotionChecker(
...
   }
 
+  /**
+   * @return `true` if this artifact is used by any resource in [environment], `false` otherwise.
+   */
+  private fun DeliveryArtifact.isUsedIn(environment: Environment) =
+    environment
+      .resources
+      .map { (it.spec as? ArtifactReferenceProvider)?.artifactReference }
+      .contains(reference)
```
robfletcher, line 128: This is basically the logic from the old `ArtifactUsedConstraintEvaluator` but it's not necessary to check it for each version as the result will never vary by version

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionCheckerTests.kt
```diff
@@ -425,34 +445,61 @@ internal class NewEnvironmentPromotionCheckerTests : JUnit5Minutests {
...
+          subject.checkEnvironments(artifactNotUsedConfig)
+        }
+      }
+
+      test("constraint checks are skipped for that environment") {
+        verify(exactly = 0) {
+          environmentConstraintRunner.checkEnvironment(any())
+        }
+      }
+    }
```
robfletcher, line 503: Test for the new behavior

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -27,79 +29,84 @@ class EnvironmentPromotionChecker(
...
                   }
                 }
-
-              val versionSelected = queuedForApproval
-                .sortedWith(artifact.versioningStrategy.comparator.reversed())
-                .lastOrNull()
-              if (versionSelected == null) {
-                log.warn("No version of {} passes constraints for environment {}", artifact.name, environment.name)
+              } else {
+                log.warn("Skipping checks for {} as it is not used in environment {}", artifact.name, environment.name)
```
emjburns, line 105: Is this really warn level? Seems more like debug

emjburns (approved):
Even if this doesn't fix the slowness it seems like a great change

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
robfletcher: Fair

luispollo (approved):
Thanks for doing this. It will clean up our logs of noise considerably! 🙏

PR #1647, opened 2020-11-07
fix(compile): fixing gradle property for Dockerfile.compile

nimakaviani:
Fixing the typo in the compile script

@plumpy

### DIFF
--- a/Dockerfile.compile
+++ b/Dockerfile.compile
@@ -19,4 +19,4 @@ RUN usermod -d /var/lib/mysql/ mysql && \
     mysql -u root --password=sa -e 'CREATE DATABASE `keel` DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;'
 
 CMD /etc/init.d/mysql restart && \
-    ./gradlew Plugin=true -PbuildingInDocker=true keel-web:installDist -x test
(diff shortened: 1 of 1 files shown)


plumpy (approved):
🥳

Dockerfile.compile
```diff
@@ -19,4 +19,4 @@ RUN usermod -d /var/lib/mysql/ mysql && \
     mysql -u root --password=sa -e 'CREATE DATABASE `keel` DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;'
 
 CMD /etc/init.d/mysql restart && \
-    ./gradlew Plugin=true -PbuildingInDocker=true keel-web:installDist -x test
+    ./gradlew --no-daemon -PbuildingInDocker=true keel-web:installDist -x test
```
luispollo, line 22: I wonder what the original flag was intended for... Any ideas?

plumpy:
It was just an editing mistake. It originally looked like this:
`--no-daemon -PenableCrossCompilePlugin=true`

I hit `33x` in vim to delete 33 characters (the length of the second argument) but my cursor was in the wrong place, so it deleted the wrong 33 characters 😳  And I'd done that same change in like 50 different places, so I was just moving too quickly and screwed it up...

luispollo:
Haha! It happens! 😄 

Thanks for fixing it!

PR #1648, opened 2020-11-08
Fix NPE deleting config by name

luispollo:
We were not checking whether the delivery config existed when deleting by name... This PR adds those checks and the corresponding tests.

I've had to upgrade `testcontainers` in order to test this locally since the latest version of Docker broke it.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,6 +6,7 @@ org.gradle.parallel=true
 spinnakerGradleVersion=8.10.0
 buildingInDocker=false
 targetJava11=true
+testContainersVersion=1.15.0
 
 # To enable a composite reference to a project, set the
 #  project property `'<projectName>Composite=true'`.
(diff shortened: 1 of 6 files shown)


lorin:
This is strictly better than what we have now, and so I support merging this.   

However, I don't think we should throw an exception when attempting to delete a delivery config that doesn't exist. Rather, I think we should treat the delete operation as idempotent, and so attempting to delete a non-existent delivery config should be treated as a success rather than a failure.

lorin (approved):
This is strictly better than what we have now, and so I support merging this.   

However, I don't think we should throw an exception when attempting to delete a delivery config that doesn't exist. Rather, I think we should treat the delete operation as idempotent, and so attempting to delete a non-existent delivery config should be treated as a success rather than a failure.

luispollo:
> This is strictly better than what we have now, and so I support merging this.
> 
> However, I don't think we should throw an exception when attempting to delete a delivery config that doesn't exist. Rather, I think we should treat the delete operation as idempotent, and so attempting to delete a non-existent delivery config should be treated as a success rather than a failure.

How would an API client know that the call was "wrong", though?

lorin:
> How would an API client know that the call was "wrong", though? The behavior I implemented here is consistent with other APIs where you can pass an identifier for a thing that we don't find in the database.

If API clients need to ensure that a delivery config is present before they delete it, they can always do a GET before doing a DELETE.

luispollo:
Thanks for the feedback, @lorin. I'm gonna take the liberty of merging as is because of my second point above. If we agree as a team that we want to change the standard to ignore not-found records when deleting, then we should do that for all APIs for consistency, IMHO. Until then, I'll stick with the current convention.

PR #1649, opened 2020-11-09
fix moving artifacts between environments

robfletcher:
Discovered this by accident. If you move a resource from one environment to another the older entry in `environment_resource` remained there.

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
@@ -17,7 +17,6 @@ import com.netflix.spinnaker.keel.core.api.randomUID
 import com.netflix.spinnaker.keel.pause.PauseScope.APPLICATION
 import com.netflix.spinnaker.keel.persistence.DeliveryConfigRepository
 import com.netflix.spinnaker.keel.persistence.NoDeliveryConfigForApplication
-import com.netflix.spinnaker.keel.persistence.NoSuchDeliveryConfigException
 import com.netflix.spinnaker.keel.persistence.NoSuchDeliveryConfigName
 import com.netflix.spinnaker.keel.persistence.OrphanedResourceException
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.CURRENT_CONSTRAINT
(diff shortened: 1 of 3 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -366,6 +365,12 @@ class SqlDeliveryConfigRepository(
             .set(ENVIRONMENT_RESOURCE.RESOURCE_UID, select(RESOURCE.UID).from(RESOURCE).where(RESOURCE.ID.eq(resource.id)))
             .onDuplicateKeyIgnore()
             .execute()
+          // delete any other environment's link to this same resource (in case we are moving it
+          // from one environment to another)
+          jooq.deleteFrom(ENVIRONMENT_RESOURCE)
```
luispollo, line 370: Shouldn't this be in the scope of a transaction along with the other statements?

gal-yardeni (approved):

lorin (approved):
Question inspired by this PR: This change handles the case where a resource is moved from one environment to another. 

However, what happens if we just delete a resource from an environment. Will keel delete the corresponding record from the `ENVIRONMENT_RESOURCE` table?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
robfletcher: That entire function seems to be non-transactional for some reason. I vaguely remember there was some thinking behind that but I can't recall what.

robfletcher:
> Question inspired by this PR:
> 
> This change handles the case where a resource is moved from one environment to another. However, what happens if we just delete a resource from an environment? Will keel delete the corresponding record from the `ENVIRONMENT_RESOURCE` table?
> 
> It looks to me like it doesn't.

Happy to tackle that if you think it makes sense to do so as part of this PR

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
robfletcher: Okay, I figured out why. The only call path to this method is from `CombinedRepository.upsertDeliveryConfig` which uses Spring's annotation-bounded transactions.

luispollo: Right, but if I recall correctly, that does not automagically propagate to jooq. I may be wrong.

robfletcher: The entire point of `DeliveryConfigTransactionTests` is to prove that it does

PR #1650, opened 2020-11-09
fix(diff): reset actions taken count for new diff

emjburns:
We were not resetting the action taken count on a new diff. We hadn't seen that before because usually a diff is cleared before a new diff comes in.


luispollo (approved):
Good spot! Thanks for fixing it!

PR #1652, opened 2020-11-10
feat(artifacts): Source-driven artifacts, for real

luispollo:
### TL;DR
The gist of this PR is to refactor `VersioningStrategy` (which sorts version strings) into a `SortingStrategy` that can sort on arbitrary fields of an artifact version, and then introduce a new implementation of `SortingStrategy` that sorts artifact versions by creation timestamp. The `ArtifactSupplier` sub-classes were modified to auto-detect the right `SortingStrategy` to use based on whether the artifact is specified in terms of source code details or not. The vast majority of changes are due to renaming and type changes.

### Context
In #1564, I laid the foundation for supporting artifact definitions based on source properties (branches, PRs). While I had originally discussed with @emjburns, @gal-yardeni and others that we'd probably need to refactor the `VersioningStrategy` associated with a `DeliveryArtifact` into a more generic `SortingStrategy`, once I was able to implement artifact retrieval from the database based on branch/PR selectors, I overlooked the fact that there were many other places in keel where we were using the comparators encapsulated by the versioning strategies, which means that implementation was incomplete.

The goal if this PR is to fill in the gaps that were left unaddressed before.

Here's the inventory of places where we use (version) comparators:
* In `ArtifactListener`, when periodically syncing artifact versions, to determine whether the latest artifact available is newer than the last one recorded in the database.
* In `DebianArtifactSupplier` and `NpmArtifactSupplier`, to determine the latest version available from the external artifact repository (i.e. from Rocket, via Igor).
* In `DockerArtifactSupplier`, to determine the latest version of a docker image available from Clouddriver.
* In `EnvironmentPromotionChecker`, to sort pending versions in the database
* In `EnvironmentConstraintRunner`, to sort versions queued for approval in the database
* In `SqlArtifactRepository` and `SqlDeliveryConfigRepository`, to sort versions when reading from the database

Note: In the process of reviewing the code more thoroughly this time around, I've found that, beyond the necessary refactoring within Keel itself, the hardest places to fix are those spots where we've let the assumption about using versions for sorting leak outside of Keel, notably in our interaction with Igor/Rocket and Clouddriver to retrieve "versions" (which we then sort with the comparators). These can certainly be addressed, but may require work in other systems as well.

### Approach
As briefly mentioned above, my suggested approach is to refactor `VersioningStrategy` into a more generic `SortingStrategy` that can operate on any field available within a `PublishedArtifact` instance for sorting purposes, as opposed to version strings only. This can be relatively easily applied in places where we read data from our database, requiring basically to read additional columns from the `ARTIFACT_VERSIONS` table into corresponding `PublishedArtifact` objects, which then get passed to the comparators. The trickier part is when retrieving versions from _other_ places such as Clouddriver or Igor, as I pointed out above.

Once the model for sorting has been generalized per the above, my plan is to introduce a new implementation of `SortingStrategy` capable of sorting `PublishedArtifact` instances based on the `createdAt` timestamp. The choice of whether to use the legacy, version-based sorting strategies, or the source-driven one, would be done automatically based on whether source information was specified for the artifact in the delivery config, similar to what I did in #1564.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/SortingStrategy.kt
```diff
@@ -5,8 +5,8 @@ import com.netflix.spinnaker.keel.api.schema.Discriminator
 /**
  * Strategy for how to sort versions of artifacts.
  */
-interface VersioningStrategy {
+interface SortingStrategy {
   @Discriminator
   val type: String
-  val comparator: Comparator<String>
+  val comparator: Comparator<PublishedArtifact>
```
luispollo, line 11: This is what will allow us to sort artifact "versions" based on arbitrary fields (such as branch and creation timestamp).

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
```diff
@@ -70,10 +70,11 @@ class DockerArtifactSupplier(
     // Note: we currently don't have a way to derive account information from artifacts,
     // so, in the calls to clouddriver below, we look in all accounts.
     val latestTag = runWithIoContext {
+      // FIXME: in order to sort by properties other than version below, the service needs to return additional data
```
luispollo, line 73: Example of where we're calling an external system to retrieve artifact "versions", and will need more data in order to use the `SortingStrategy`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -42,37 +42,26 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
...
+  ): List<PublishedArtifact>
 
   /**
    * Persists the specified instance of the artifact.
    *
    * @return `true` if a new version is persisted, `false` if the specified version was already
    * known (in which case this method is a no-op).
    */
-  fun storeArtifactInstance(artifact: PublishedArtifact): Boolean
+  fun storeArtifactVersion(artifactVersion: PublishedArtifact): Boolean
```
luispollo, line 58: Renamed the methods here for consistency with the terminology used elsewhere.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
```diff
@@ -145,9 +145,9 @@ interface KeelRepository : KeelReadOnlyRepository {
 
   fun getAllArtifacts(type: ArtifactType? = null): List<DeliveryArtifact>
 
-  fun storeArtifactInstance(artifact: PublishedArtifact): Boolean
+  fun storeArtifactVersion(artifactVersion: PublishedArtifact): Boolean
```
luispollo, line 148: Renamed the methods here for consistency with the terminology elsewhere.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -1057,8 +1079,9 @@ class SqlArtifactRepository(
             ArtifactStatus.valueOf(releaseStatus)
           }
           .toSet()
+        // FIXME: in order to sort by properties other than version below, we need the full PublishedArtifact objects
```
luispollo, line 1082: This is the hardest place to change to use full artifact version objects instead of just version strings that I've come across so far. I think it's safe to say this is the most complex function in all of keel, and I'm scared of breaking it... 😬

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ArtifactController.kt
```diff
@@ -60,16 +60,6 @@ class ArtifactController(
     eventPublisher.publishEvent(ArtifactSyncEvent(true))
   }
 
-  @GetMapping(
-    path = ["/{name}/{type}"],
-    produces = [APPLICATION_JSON_VALUE, APPLICATION_YAML_VALUE]
-  )
-  fun listVersions(
```
luispollo, line 67: Just some fall clean-up since this API was not being used anywhere. 🍂 🗑️

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifactSupplier.kt
```diff
@@ -26,18 +26,19 @@ class NpmArtifactSupplier(
...
   override val supportedArtifact = SupportedArtifact(NPM, NpmArtifact::class.java)
 
-  override val supportedVersioningStrategy =
-    SupportedVersioningStrategy(NPM, NpmVersioningStrategy::class.java)
+  override val supportedSortingStrategy =
+    SupportedSortingStrategy(NPM, NpmVersionSortingStrategy::class.java)
 
   override fun getLatestArtifact(deliveryConfig: DeliveryConfig, artifact: DeliveryArtifact): PublishedArtifact? =
     runWithIoContext {
+      // FIXME: in order to sort by properties other than version below, the service needs to return additional data
```
luispollo, line 38: Another example of where we're calling an external system to retrieve artifact "versions", and will need more data in order to use the `SortingStrategy`. We could return more data in the list (bigger responses), or make N calls to fill in the details for each version (more calls).

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/SortingStrategy.kt
robfletcher: That seems like a great change

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -69,40 +70,42 @@ class EnvironmentPromotionChecker(
...
-                      if (constraintRunner.checkStatelessConstraints(artifact, deliveryConfig, v, environment)) {
-                        approveVersion(deliveryConfig, artifact, v, environment.name)
-                        repository.deleteQueuedConstraintApproval(deliveryConfig.name, environment.name, v, artifact.reference)
+                      if (constraintRunner.checkStatelessConstraints(artifact, deliveryConfig, artifactVersion.version, environment)) {
+                        approveVersion(deliveryConfig, artifact, artifactVersion.version, environment.name)
+                        repository.deleteArtifactVersionQueuedForApproval(
+                          deliveryConfig.name, environment.name, artifactVersion.version, artifact.reference)
+                      } else {
+                        log.debug("Version ${artifactVersion.version} of $artifact does not currently pass stateless constraints")
+                        queuedForApproval.remove(artifactVersion)
```
luispollo, line 98: This was missing in the code before, and would affect the logging statements down below. No effect on behavior, just that we might get misleading logs otherwise.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/persistence/KeelReadOnlyRepository.kt
```diff
@@ -30,9 +31,9 @@ interface KeelReadOnlyRepository {
 
   fun constraintStateFor(deliveryConfigName: String, environmentName: String, artifactVersion: String): List<ConstraintState>
 
-  fun pendingConstraintVersionsFor(deliveryConfigName: String, environmentName: String): List<String>
+  fun getPendingArtifactVersions(deliveryConfigName: String, environmentName: String, artifact: DeliveryArtifact): List<PublishedArtifact>
```
luispollo, line 34: Renamed for clarity and consistency.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/persistence/KeelReadOnlyRepository.kt
```diff
@@ -30,9 +31,9 @@ interface KeelReadOnlyRepository {
...
 
-  fun getQueuedConstraintApprovals(deliveryConfigName: String, environmentName: String, artifactReference: String?): Set<String>
+  fun getArtifactVersionsQueuedForApproval(deliveryConfigName: String, environmentName: String, artifact: DeliveryArtifact): List<PublishedArtifact>
```
luispollo, line 36: Renamed for clarity and consistency.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
```diff
@@ -28,23 +28,24 @@ class DebianArtifactSupplier(
...
 
   override fun getLatestArtifact(deliveryConfig: DeliveryConfig, artifact: DeliveryArtifact): PublishedArtifact? =
     runWithIoContext {
       artifactService.getVersions(artifact.name, artifact.statusesForQuery, DEBIAN)
-        .map { version -> "${artifact.name}-$version" }
-        .sortedWith(artifact.versioningStrategy.comparator)
-        .firstOrNull() // versioning strategies return descending by default... ¯\_(ツ)_/¯
-        ?.let { version ->
-          artifactService.getArtifact(artifact.name, version.removePrefix("${artifact.name}-"), DEBIAN)
+        // FIXME: this is making N calls to fill in data for each version so we can sort.
```
luispollo, line 42: For the time being, I've worked around the lack of support in the Rocket API for returning a list of artifact versions with details by adding an additional call to Igor/Rocket for each version returned. I realize this is expensive an technically unnecessary, but as discussed in chat, my focus at the moment is to get something working, which we can improve for performance later.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
```diff
@@ -69,46 +69,20 @@ class DockerArtifactSupplier(
 
     // Note: we currently don't have a way to derive account information from artifacts,
     // so, in the calls to clouddriver below, we look in all accounts.
-    val latestTag = runWithIoContext {
+    return runWithIoContext {
       cloudDriverService
         .findDockerTagsForImage("*", artifact.name, deliveryConfig.serviceAccount)
         .distinct()
-        .sortedWith(artifact.versioningStrategy.comparator)
+        // FIXME: this is making N calls to fill in data for each version so we can sort.
```
luispollo, line 76: Same comment on leaving performance optimizations for later.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifactSupplier.kt
```diff
@@ -26,22 +26,24 @@ class NpmArtifactSupplier(
...
 
   override fun getLatestArtifact(deliveryConfig: DeliveryConfig, artifact: DeliveryArtifact): PublishedArtifact? =
     runWithIoContext {
       artifactService
         .getVersions(artifact.nameForQuery, artifact.statusesForQuery, NPM)
-        .sortedWith(artifact.versioningStrategy.comparator)
-        .firstOrNull() // versioning strategies return descending by default... ¯\_(ツ)_/¯
-        ?.let { version ->
-          artifactService.getArtifact(artifact.nameForQuery, version, NPM)
+        // FIXME: this is making N calls to fill in data for each version so we can sort.
```
luispollo, line 40: Same comment on leaving performance optimizations for later.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -51,34 +56,46 @@ class ImageService(
...
+    val candidateImagesToVersions = cloudDriverService
+      .namedImages(DEFAULT_SERVICE_ACCOUNT, artifact.name, account)
       .filter { it.hasAppVersion && it.hasBaseImageVersion }
-      .sortedWith(NamedImageComparator)
-    val latest = candidateImages
-      .firstOrNull {
-        // TODO: Frigga and Rocket version parsing are not aligned. We should consolidate.
-        it.appVersion.parseAppVersion().packageName == artifactName
+      .associateBy { image ->
+        image.toArtifactVersion()
```
luispollo, line 64: @robfletcher Calling your attention to this change. This is now resorting to looking up the artifact version details from our database, and ignoring versions which we haven't stored. The only place this function is called is indirectly in `ImageHandler` to determine the latest baked AMI for the artifact. I'm afraid this might break that behavior because we'd ignore images for which we don't have an artifact version record in the database. Thoughts?

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -138,19 +155,6 @@ class ImageService(
         }
       }
 
-  suspend fun getNamedImageFromJenkinsInfo(packageName: String, account: String, buildHost: String, buildName: String, buildNumber: String): NamedImage? =
```
luispollo, line 141: Unused.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/NamedImage.kt
```diff
@@ -12,17 +12,6 @@ data class NamedImage(
   val amis: Map<String, List<String>?>
 )
 
-object NamedImageComparator : Comparator<NamedImage> {
```
luispollo, line 15: No longer used.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
```diff
@@ -133,18 +134,18 @@ class EnvironmentConstraintRunner(
...
   private fun handleOlderPendingVersions(
     envContext: EnvironmentContext,
-    pendingVersionsToCheck: MutableSet<String>
+    pendingVersionsToCheck: MutableList<PublishedArtifact>
   ) {
     log.debug("pendingVersionsToCheck: [$pendingVersionsToCheck] of artifact ${envContext.artifact.name} for environment ${envContext.environment.name} ")
     pendingVersionsToCheck
-      .sortedWith(envContext.artifact.versioningStrategy.comparator.reversed()) // oldest first
-      .forEach { version ->
+      .reversed() // oldest first
```
luispollo, line 141: These versions should already be sorted when we get here.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
```diff
@@ -158,10 +159,7 @@ class EnvironmentConstraintRunner(
     val artifact: DeliveryArtifact,
     val versions: List<String>,
     val vetoedVersions: Set<String>
-  ) {
-    fun hasNoConstraints(): Boolean =
```
luispollo, line 162: Unused.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -69,40 +69,42 @@ class EnvironmentPromotionChecker(
...
+                    .toMutableList()
 
                   /**
                    * Approve all constraints starting with oldest first so that the ordering is
                    * maintained.
                    */
                   queuedForApproval
-                    .sortedWith(artifact.versioningStrategy.comparator.reversed())
-                    .forEach { v ->
+                    .reversed()
```
luispollo, line 81: These versions should already be sorted by the time we get here.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -944,37 +947,58 @@ class SqlDeliveryConfigRepository(
     }
   }
 
-  override fun pendingConstraintVersionsFor(deliveryConfigName: String, environmentName: String): List<String> {
+  override fun getPendingArtifactVersions(deliveryConfigName: String, environmentName: String, artifact: DeliveryArtifact): List<PublishedArtifact> {
```
luispollo, line 950: All the functions that previously returned lists of version strings now return a list of `PublishedArtifact` so they match the comparator signature in `SortingStrategy`.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -944,37 +947,58 @@ class SqlDeliveryConfigRepository(
...
+          ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ARTIFACT_REFERENCE.eq(artifact.reference)
             // for backward comparability
             .or(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ARTIFACT_REFERENCE.isNull)
         )
-        .fetch(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ARTIFACT_VERSION)
-        .toSet()
+        .and(ARTIFACT_VERSIONS.NAME.eq(artifact.name))
+        .and(ARTIFACT_VERSIONS.TYPE.eq(artifact.type))
+        .and(ARTIFACT_VERSIONS.VERSION.eq(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ARTIFACT_VERSION))
+        .fetchSortedArtifactVersions(artifact)
```
luispollo, line 1001: This applies native SQL sorting and limiting when applicable, and falls back to the comparator from the sorting strategy when that doesn't apply. This means the return from this function is pre-sorted.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -944,37 +947,58 @@ class SqlDeliveryConfigRepository(
...
+        .from(ENVIRONMENT_ARTIFACT_CONSTRAINT, ARTIFACT_VERSIONS)
         .where(
           ENVIRONMENT_ARTIFACT_CONSTRAINT.ENVIRONMENT_UID.eq(environmentUID),
           ENVIRONMENT_ARTIFACT_CONSTRAINT.STATUS.eq(ConstraintStatus.PENDING.toString())
         )
-        .fetch(ENVIRONMENT_ARTIFACT_CONSTRAINT.ARTIFACT_VERSION)
+        .and(ARTIFACT_VERSIONS.NAME.eq(artifact.name))
+        .and(ARTIFACT_VERSIONS.TYPE.eq(artifact.type))
+        .and(ARTIFACT_VERSIONS.VERSION.eq(ENVIRONMENT_ARTIFACT_CONSTRAINT.ARTIFACT_VERSION))
+        .fetchSortedArtifactVersions(artifact)
```
luispollo, line 973: This applies native SQL sorting and limiting when applicable, and falls back to the comparator from the sorting strategy when that doesn't apply. This means the return from this function is pre-sorted.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
```diff
@@ -53,3 +74,99 @@ fun mapToArtifact(
...
+      type = type,
+      version = version,
+      status = status?.let { ArtifactStatus.valueOf(it) },
+      createdAt = createdAt?.toInstant(ZoneOffset.UTC),
+      gitMetadata = gitMetadata?.let { objectMapper.readValue(it) },
+      buildMetadata = buildMetadata?.let { objectMapper.readValue(it) },
+    )
+  }
+
+internal fun ArtifactVersionSelectStep.fetchSortedArtifactVersions(
```
luispollo, line 94: Utility function to avoid duplication when querying the database for artifact versions. This applies native SQL sorting and limiting when applicable, and falls back to the comparator from the sorting strategy if not.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -219,76 +210,31 @@ class SqlArtifactRepository(
...
-    return if (artifact.filteredByPullRequest || artifact.filteredByBranch) {
-      versions
-    } else {
-      val sortedVersions = versions.sortedWith(artifact.versioningStrategy.comparator)
-      if (artifact is DockerArtifact) {
-        filterDockerVersions(artifact, sortedVersions, limit)
-      } else {
-        sortedVersions.subList(0, Math.min(sortedVersions.size, limit))
-      }
+        .fetchSortedArtifactVersions(artifact, limit)
```
luispollo, line 232: This applies native SQL sorting and limiting when applicable, and falls back to the comparator from the sorting strategy when that doesn't apply. This means the return from this function is pre-sorted.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -377,30 +322,6 @@ class SqlArtifactRepository(
       throw NoSuchArtifactException(artifact)
     }
 
-  /**
-   * Given a docker artifact and a list of docker tags, filters out all tags that don't produce exactly one capture
-   * group with the provided regex.
-   *
-   */
-  private fun filterDockerVersions(artifact: DockerArtifact, versions: List<String>, limit: Int): List<String> =
```
luispollo, line 385: Moved to `ArtifactUtils`.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -71,14 +70,6 @@ class SqlArtifactRepository(
   private val artifactSuppliers: List<ArtifactSupplier<*, *>> = emptyList()
 ) : ArtifactRepository {
 
-  companion object {
```
luispollo, line 74: Moved to `ArtifactUtils`.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -980,8 +918,13 @@ class SqlArtifactRepository(
       val artifactVersions = deliveryConfig.artifacts.map { artifact ->
         val versionsInEnvironment = jooq
           .select(
-            ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION,
+            ARTIFACT_VERSIONS.NAME,
```
luispollo, line 921: @robfletcher This was the scariest function for me to change by far. I'd appreciate your attention here.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
luispollo: On second thought, I think it's safe to keep sorting the images by version, since the point here is just to find the latest AMI for a given artifact version. I'll undo this.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BranchAndTimestampSortingStrategy.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.netflix.spinnaker.keel.api.artifacts.SortingStrategy
+
+/**
+ * A [SortingStrategy] that compares artifact versions by branch and timestamp.
+ */
+object BranchAndTimestampSortingStrategy : SortingStrategy {
+  override val type: String = "branch-and-timestamp"
+
+  override val comparator: Comparator<PublishedArtifact> =
+    compareBy<PublishedArtifact> { it.gitMetadata?.branch }.thenByDescending { it.createdAt }
```
luispollo, line 13: This is it... All this work to prepare for this single line right here... 🙂

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifact.kt
```diff
@@ -16,9 +16,16 @@ data class DebianArtifact(
   override val reference: String = name,
   val vmOptions: VirtualMachineOptions,
   override val statuses: Set<ArtifactStatus> = emptySet(),
-  override val versioningStrategy: VersioningStrategy = DebianVersioningStrategy,
   override val from: ArtifactOriginFilterSpec? = null
 ) : DeliveryArtifact() {
   override val type = DEBIAN
+
+  override val sortingStrategy: SortingStrategy
```
luispollo, line 23: The sorting strategy is auto-selected based on whether the artifact spec is defined in terms of source details or not.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifact.kt
```diff
@@ -37,5 +36,13 @@ data class DockerArtifact(
 
   @JsonIgnore
   override val statuses: Set<ArtifactStatus> = emptySet()
+
+  override val sortingStrategy: SortingStrategy
```
luispollo, line 40: The sorting strategy is auto-selected based on whether the artifact spec is defined in terms of source details or not.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifact.kt
```diff
@@ -14,9 +14,16 @@ data class NpmArtifact(
   override val deliveryConfigName: String? = null,
   override val reference: String = name,
   override val statuses: Set<ArtifactStatus> = emptySet(),
-  override val versioningStrategy: VersioningStrategy = NpmVersioningStrategy,
   override val from: ArtifactOriginFilterSpec? = null
 ) : DeliveryArtifact() {
   override val type = NPM
+
+  override val sortingStrategy: SortingStrategy
```
luispollo, line 21: The sorting strategy is auto-selected based on whether the artifact spec is defined in terms of source details or not.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/mixins/DeliveryArtifactMixin.kt
```diff
@@ -1,13 +1,23 @@
...
+  val sortingStrategy: SortingStrategy
+
+  @get:JsonIgnore
+  val filteredByBranch : Boolean
+
+  @get:JsonIgnore
+  val filteredByPullRequest: Boolean
+
+  @get:JsonIgnore
+  val filteredByReleaseStatus: Boolean
```
luispollo, line 22: These were showing up in the serialized objects but are only meant for internal use.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/mixins/DeliveryArtifactMixin.kt
```diff
@@ -1,13 +1,23 @@
...
-import com.netflix.spinnaker.keel.api.artifacts.VersioningStrategy
+import com.netflix.spinnaker.keel.api.artifacts.SortingStrategy
 
 internal interface DeliveryArtifactMixin {
   @get:JsonProperty(access = WRITE_ONLY)
   val deliveryConfigName: String
 
-  @get:JsonProperty(access = WRITE_ONLY)
-  val versioningStrategy: VersioningStrategy
+  @get:JsonIgnore
```
luispollo, line 12: We were already skipping this property when serializing. This just makes it explicit that we're not expecting users to specify the sorting strategy in the delivery config for the time being.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
emjburns: 👍 I think this is fine for now until we have the correct API.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
emjburns: RIP jenkins support 👋

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
emjburns: Is there a test that verifies this?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/VersionComparators.kt
```diff
@@ -37,9 +38,9 @@ class TagComparator(
   private val semverComparator = NullSafeComparator(SEMVER_COMPARATOR, false)
   private val increasingComparator = NullSafeComparator(INCREASING_COMPARATOR, false)
```
emjburns, line 40: feels like all the naming changes in this file are not necessary

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/mixins/DeliveryArtifactMixin.kt
emjburns: I'm confused, I thought you had to specify write only in order for it to not serialize on read?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
luispollo: Good point. No, I'll try to add some.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/VersionComparators.kt
luispollo: You're right. I was just bugging me that similar implementations used different names for these variables so I changed them. Please indulge me this time? 🙂

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/mixins/DeliveryArtifactMixin.kt
luispollo: `READ_ONLY` is "ignore when deserializing" (i.e. "reading" the object), `WRITE_ONLY` is "ignore when serializing" (writing the object to JSON), and `@JsonIgnore` is both. Not confusing at all, right? 🤦‍♂️

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
luispollo: Added in https://github.com/spinnaker/keel/pull/1652/commits/c653644f2d8e68d445fd47f79977c902308d77c3

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BranchAndTimestampSortingStrategy.kt
luispollo: @erikmunson A question for you. Given this set of versions, branches and timestamps (in order of timestamp):

|version|branch|createdAt|
--|--|--
1.0.0|master|2020-11-12T15:20:00Z
1.0.1-rc.1|mybranch|2020-11-12T16:20:00Z
1.0.1|master|2020-11-12T17:20:00Z
1.0.2-rc.1|mybranch|2020-11-12T18:20:00Z

...this comparator will return the following list:

|version|branch|createdAt|
--|--|--
1.0.0|master|2020-11-12T15:20:00Z
1.0.1|master|2020-11-12T17:20:00Z
1.0.1-rc.1|mybranch|2020-11-12T16:20:00Z
1.0.2-rc.1|mybranch|2020-11-12T18:20:00Z

That is, we group by branch, _then_ sort by timestamp. When getting artifact summaries for the UI, given a size limit for the list, we may end up returning a bunch of versions for the first branch, and few or none for the second, in this example. From a UX perspective, I'm not sure this is ideal. We could ignore the branch altogether and sort only by timestamp. We could also roll this out and see what it looks like in staging, and then decide. Thoughts?

luispollo: P.S. To be clear, this is a compound problem between the sorting comparator here and the implementation of that API, so it's a bit tangential to this PR, but I was reminded of it when reviewing my own code, and thought I'd bring it up.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/NpmArtifactSupplierTests.kt
```diff
@@ -101,21 +111,23 @@ internal class NpmArtifactSupplierTests : JUnit5Minutests {
...
+            SupportedSortingStrategy(NPM, NpmVersionSortingStrategy::class.java)
           )
       }
 
       test("looks up latest artifact from igor") {
         val result = runBlocking {
           npmArtifactSupplier.getLatestArtifact(deliveryConfig, npmArtifact)
         }
-        expectThat(result).isEqualTo(latestArtifact)
+        expectThat(result!!.version).isEqualTo(latestArtifact.version)
```
emjburns, line 125: can you chain these togther with a `.isNotNull()` and `.isEqualTo`()`? I don't like using `!!`

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageServiceTests.kt
```diff
@@ -196,13 +199,6 @@ class ImageServiceTests : JUnit5Minutests {
         expectThat(newestImage.imageName).isEqualTo("my-package-0.0.1_rc.99-h100")
       }
```
emjburns, line 201: is this test no longer relevant?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
```diff
@@ -43,14 +44,14 @@ class EnvironmentConstraintRunner(
...
   ) {
-    val pendingVersionsToCheck: MutableSet<String> =
-      when (envContext.environment.constraints.anyStateful) {
-        true ->
-          repository
-            .pendingConstraintVersionsFor(envContext.deliveryConfig.name, envContext.environment.name)
-            .filter { envContext.versions.contains(it) }
-            .toMutableSet()
-        false -> mutableSetOf()
+    val pendingVersionsToCheck: MutableList<PublishedArtifact> =
```
emjburns, line 47: How does this now return a list of published artifacts? `getPendingArtifactVersions` returns a list of strings, and I don't think you changed that in this PR

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BranchAndTimestampSortingStrategy.kt
erikmunson: Is there any sort of mechanical/functional usefulness on the keel side in sorting by branch here? As a human using this I would expect things to be used strictly chronologically — i.e. if i have that `regex: ".*-branch"` filter, it would be pretty strange if I pushed a matching version and it got stuck in the middle of the list behind other things from days ago simply because of how the branch name happened to sort alphabetically against the older branch names. If there _is_ some kind of functional reason this branch sorting is useful for keel, we can talk about the tradeoffs/other alternatives.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
luispollo: I did. That's the function I wrote the test for in https://github.com/spinnaker/keel/pull/1652/commits/c653644f2d8e68d445fd47f79977c902308d77c3

luispollo: P.S. All the functions that used to return lists of version strings now return lists of `PublishedArtifact`.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
gal-yardeni: where are the calls are being made?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
```diff
@@ -69,46 +69,20 @@ class DockerArtifactSupplier(
...
+        //  Ideally, we'd make a single call to return the list with details for each version.
+        .mapNotNull { getArtifactByVersion(artifact, it) }
+        .sortedWith(artifact.sortingStrategy.comparator)
         .firstOrNull()
     }
-
-    return if (latestTag != null) {
```
gal-yardeni, line 80: why was this logic deleted? where are we creating the newly published artifact if its not the same?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
luispollo: On line 48:
```kotlin
artifactService.getArtifact(artifact.name, version, DEBIAN)
```

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
```diff
@@ -27,11 +27,11 @@ class DockerArtifactSupplier(
...
+) : BaseArtifactSupplier<DockerArtifact, DockerVersionSortingStrategy>(artifactMetadataService) {
   override val supportedArtifact = SupportedArtifact("docker", DockerArtifact::class.java)
 
-  override val supportedVersioningStrategy =
-    SupportedVersioningStrategy("docker", DockerVersioningStrategy::class.java)
+  override val supportedSortingStrategy =
+    SupportedSortingStrategy("docker", DockerVersionSortingStrategy::class.java)
 
   override fun getArtifactByVersion(artifact: DeliveryArtifact, version: String): PublishedArtifact? {
     return runWithIoContext {
```
luispollo, line 37: @gal-yardeni @emjburns Can the call to clouddriver here return images for multiple tags/versions of the same artifact? Something like a `tag = "*"`? If it does, then perhaps we could make a single call instead of N calls before sorting versions.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
luispollo: Because I changed the code immediately above to retrieve the details for all the versions, not just for the latest, so we can sort with the new `SortingStrategy`. The list now has the full objects, not just the version strings.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -262,62 +289,94 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
...
-          val updatedConstraintState = repository.constraintStateFor(deliveryConfig.name, environment.name)
-
-          expectThat(appConstraintState)
-            .contains(updatedConstraintState)
-          expectThat(updatedConstraintState)
-            .hasSize(1)
-          expectThat(updatedConstraintState.first().status)
-            .isEqualTo(ConstraintStatus.PASS)
-        }
+        context("artifact constraint flows") {
```
luispollo, line 292: Indentation changes will throw off the diff viewer here. I suggest ignoring whitespace when reviewing.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BranchAndTimestampSortingStrategy.kt
luispollo: In hindsight, I don't think there is. I think at some point I mixed up filtering by branch/PR with sorting by timestamp in my head, and it stuck. Let's agree to change it to `CreatedAtSortingStrategy` and sort by timestamp only, then?

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageServiceTests.kt
luispollo: I think I removed this by mistake. Will undo.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BranchAndTimestampSortingStrategy.kt
erikmunson: Love it! ✅

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -129,16 +129,16 @@ class ArtifactListener(
             if (latestArtifact != null) {
               val hasNew = when {
                 lastRecordedVersion == null -> true
-                latestArtifact.version != lastRecordedVersion -> {
-                  artifact.versioningStrategy.comparator.compare(lastRecordedVersion, latestArtifact.version) > 0
+                latestArtifact.version != lastRecordedVersion.version -> {
```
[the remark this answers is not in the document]
luispollo: I know. I struggled with these names a bit because the table is called `artifact_versions` and the methods are called `somethingArtifactVersion`, while the class is currently named `PublishedArtifact`, and we generally use "artifact" to refer to a `DeliveryArtifact`... 😓

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
gal-yardeni: maybe i'm missing something - where are you calling this N times? I see only 2 calls - one to fetch all versions. and one to get the actual artifact data.

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplierTests.kt
```diff
@@ -96,10 +99,10 @@ internal class DockerArtifactSupplierTests : JUnit5Minutests {
         )
       }
 
-      test("supports Docker versioning strategy") {
-        expectThat(dockerArtifactSupplier.supportedVersioningStrategy)
+      test("supports Docker version sorting strategy") {
```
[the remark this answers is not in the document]
luispollo: Actually, this is not being used at this point, because the sorting strategy is automatically determined (as opposed to being configured in the delivery config and deserialized), which is where `supportedSortingStrategy` would come into play). I stopped short of removing all serialization support because I'm not sure if this will stay this way forever, but was inclined to do so. This is a question I'd like @robfletcher to weigh in on, though.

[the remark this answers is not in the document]
gal-yardeni: oh I see... ok, thanks for explaining. Is there another place that you have tests for sorting by timestamp?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifactSupplier.kt
luispollo: Right, but the second call is in a loop.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplier.kt
luispollo: That worked! Please take a look: https://github.com/spinnaker/keel/pull/1652/commits/7acc44793f5a75866f14bbb5182af2ff435015f3

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -262,62 +289,94 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
...
+            )
+            expectThat(updatedState).isNotNull()
+            expectThat(updatedState!!.artifactReference).isNull()
+          }
 
-        test("can queue constraint approvals") {
-          queueConstraintApproval()
-          expectThat(repository.getQueuedConstraintApprovals(deliveryConfig.name, "staging", "keel"))
-            .isEqualTo(setOf("keel-1.0.0"))
+          // TODO: this should be removed when the artifactReference is made non-nullable in the repository methods
```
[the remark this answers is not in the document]
luispollo: I didn't touch the the `getConstraintState` function.

gal-yardeni (approved):
This looks good to me! thank you for taking the time and addressing this gap. I don't have any major concerns at the moment, at least until the CI team might be able to help us with performance :)

keel-artifact/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifactSupplierTests.kt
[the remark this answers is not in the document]
luispollo: Yes, here are the tests:
https://github.com/spinnaker/keel/blob/7acc44793f5a75866f14bbb5182af2ff435015f3/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt#L353-L404

https://github.com/spinnaker/keel/blob/7acc44793f5a75866f14bbb5182af2ff435015f3/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/artifacts/CreatedAtSortingStrategyTests.kt

And here are the tests for the SQL changes to sort by timestamp and filter by branch/pull request, from #1564:
https://github.com/luispollo/keel/blob/2fb289cecbd034e57bcca3e412b3850867a1e746/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt#L916-L1074

lorin (approved):
Some comments inline, but nothing that would block merge. I'm not familiar enough with this code to really do a great review, tho.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/persistence/KeelReadOnlyRepository.kt
luispollo: I agree. I can do that in a separate PR. The only reason I didn't do it here has to do with something that @gal-yardeni brought up, which is the fact that these objects have a `version` field, so following the default naming pattern you end up with things like `artifactVersion.version`, which looks a bit odd. But I couldn't think of a better alternative. 🤷‍♂️

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
luispollo: Thanks for the suggestion, @lorin. I'll break the function down. As for the last part of your comment, can you please explain what you mean by "side-effect style", maybe with an example?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -415,33 +336,38 @@ class SqlArtifactRepository(
...
+        .from(ARTIFACT_VERSIONS, ENVIRONMENT_ARTIFACT_VERSIONS)
+        .where(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION.eq(ARTIFACT_VERSIONS.VERSION))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(envUid))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifactId))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.APPROVED_AT.isNotNull)
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.ne(VETOED.name))
+        .orderBy(ENVIRONMENT_ARTIFACT_VERSIONS.APPROVED_AT.desc())
+        .limit(20)
+        .fetchArtifactVersions()
+        .sortedWith(artifact.sortingStrategy.comparator).firstOrNull()?.version
```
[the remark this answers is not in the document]
luispollo: Yes. It's sorting by a different column (`ENVIRONMENT_ARTIFACT_VERSIONS.APPROVED_AT`) and hard-capping at 20. I didn't want to mess around with that in this PR.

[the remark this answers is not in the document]
robfletcher: That worries me. We're limiting the result set based on one sorting mechanism and then re-sorting that truncated set based on a different one? Seems like a recipe for "why don't I see the version I expect?" type questions.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
robfletcher: Am I missing something? This function doesn't appear to be logically different.

luispollo:
> +1 on breaking this out into a utility function. I would recommend breaking this out even further, in order to make it easier for a reader to reason about.
> 
> In particular, it looks like this function has multiple responsibilities, including the following two, which involve multiple conditional statements:
> 
> narrow the query (and, limit stuff)
> apply the appropriate sorting strategy
> I would recommend breaking out those two into their own helper functions.
> 
> Also, I found it confusing how the and and limit calls are used here. I'm used to seeing them invoked functional style, this code uses them in side-effect style. I didn't even know they mutated the object that they were invoked in before reading this code and digging into the implementation.

@lorin For some reason I can no longer respond to that comment, so quoting it here. I've refactored the function, but took a different approach than you suggested because the sorting part is only separate for the flow where the artifact is not "source-driven", so I broke it out into two functions, one for each case. I think it's easier to read now, but let me know.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
luispollo: No, you're right. It's supposed to be functionally equivalent, it's just that it's pretty complex and so I was afraid I might fat-finger something, is all. Having said that, looking at the diff again now, it doesn't look that scary. Thanks for looking anyway!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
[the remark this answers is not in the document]
luispollo: @robfletcher So the context here is that this is the `lastVersionApprovedIn` function, so sorting doesn't really matter, as we're looking for the row with the latest `APPROVED_AT` value. Also, the result set here is not re-sorted anywhere (maybe you were looking at the changes for another function down below?).

PR #1653, opened 2020-11-10
Basic definition of verifyWith

robfletcher:
Proposed delivery config syntax

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Environment.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Environment.kt
@@ -4,7 +4,7 @@ data class Environment(
   val name: String,
   val resources: Set<Resource<*>> = emptySet(),
   val constraints: Set<Constraint> = emptySet(),
-  val verifyWith: Set<Validation> = emptySet(),
+  val verifyWith: Set<Verification> = emptySet(),
   val notifications: Set<NotificationConfig> = emptySet() // applies to each resource
 ) {
   override fun toString(): String = "Environment $name"
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Environment.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Environment.kt
@@ -4,6 +4,7 @@ data class Environment(
   val name: String,
   val resources: Set<Resource<*>> = emptySet(),
   val constraints: Set<Constraint> = emptySet(),
+  val verifyWith: Set<Validation> = emptySet(),
   val notifications: Set<NotificationConfig> = emptySet() // applies to each resource
 ) {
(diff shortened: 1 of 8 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Validation.kt
```diff
@@ -0,0 +1,8 @@
+package com.netflix.spinnaker.keel.api
+
+import com.netflix.spinnaker.keel.api.schema.Discriminator
+
+interface Validation {
```
robfletcher, line 5: Does this name work?

luispollo: Sounds good to me!

luispollo (approved):
LGTM!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Environment.kt
```diff
@@ -4,6 +4,7 @@ data class Environment(
   val name: String,
   val resources: Set<Resource<*>> = emptySet(),
   val constraints: Set<Constraint> = emptySet(),
+  val verifyWith: Set<Validation> = emptySet(),
```
robfletcher, line 7: @lorin we discussed the order of evaluation of things in here. I've made it a set for now given we use that almost everywhere else but do you think it would make sense to rather use a list?

lorin: I'm inclined to go with a list, given that the majority of the people who we've asked assume that the validations would run in the order defined in the delivery config.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Validation.kt
lorin: I'd be happier if the naming was consistent with the field in the delivery config. So, either "verifyWith / Verification" or "validateWith / Validation" (I don't have a preference as to which pair).

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Environment.kt
emjburns: @robfletcher Why did you choose "verifyWith" instead of "validateWith"?

lorin (approved):
I picked a couple of nits, but otherwise LGTM.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Environment.kt
robfletcher: Yeah,  I wondered because I started thinking about how we'll record verification state in the database.

PR #1655, opened 2020-11-11
Components for running verifications vs an environment

robfletcher:
Start of an implementation for executing verifications against an environment.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
@@ -1,0 +1,9 @@
+package com.netflix.spinnaker.keel.api.plugins
+
+import com.netflix.spinnaker.keel.api.Verification
+
+interface VerificationEvaluator<VERIFICATION: Verification> {
+  val supportedVerification: Pair<String, Class<out Verification>>
+
+  fun evaluate()
+}
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
@@ -3,10 +3,19 @@ package com.netflix.spinnaker.keel.api.plugins
 import com.netflix.spinnaker.keel.api.Verification
 import com.netflix.spinnaker.keel.api.verification.VerificationStatus
 
+/**
+ * A component responsible for performing verification of an [com.netflix.spinnaker.keel.api.Environment].
+ */
 interface VerificationEvaluator<VERIFICATION: Verification> {
-  val supportedVerification: Pair<String, Class<out Verification>>
+  val supportedVerification: Pair<String, Class<VERIFICATION>>
 
+  /**
+   * @return the current status of the verification.
+   */
   fun evaluate() : VerificationStatus
 
+  /**
+   * Start running [verification].
+   */
   fun start(verification: Verification)
 }
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
@@ -1,9 +1,12 @@
 package com.netflix.spinnaker.keel.api.plugins
 
 import com.netflix.spinnaker.keel.api.Verification
+import com.netflix.spinnaker.keel.api.verification.VerificationStatus
 
 interface VerificationEvaluator<VERIFICATION: Verification> {
   val supportedVerification: Pair<String, Class<out Verification>>
 
-  fun evaluate()
+  fun evaluate() : VerificationStatus
+
+  fun start(verification: Verification)
 }
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
@@ -1,5 +1,6 @@
 package com.netflix.spinnaker.keel.api.verification
 
+import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.Verification
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
@@ -7,17 +8,13 @@ import java.time.Instant
 
 interface VerificationRepository {
   fun getState(
-    verification: Verification,
-    environment: Environment,
-    deliveryArtifact: DeliveryArtifact,
-    version: String
+    context: VerificationContext,
+    verification: Verification
   ) : VerificationState?
 
   fun updateState(
+    context: VerificationContext,
     verification: Verification,
-    environment: Environment,
-    deliveryArtifact: DeliveryArtifact,
-    version: String,
     status: VerificationStatus
   )
 }
@@ -30,4 +27,16 @@ data class VerificationState(
 
 enum class VerificationStatus(val complete: Boolean) {
   RUNNING(false), PASSED(true), FAILED(true)
+}
+
+data class VerificationContext(
+  val deliveryConfig: DeliveryConfig,
+  val environmentName: String,
+  val version: String
+) {
+  val environment: Environment =
+    deliveryConfig.environments.first { it.name == environmentName }
+
+  val artifact: DeliveryArtifact =
+    deliveryConfig.artifacts.first()
 }
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
@@ -6,12 +6,20 @@ import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import java.time.Instant
 
 interface VerificationRepository {
-  fun getVerificationState(
+  fun getState(
     verification: Verification,
     environment: Environment,
     deliveryArtifact: DeliveryArtifact,
     version: String
   ) : VerificationState?
+
+  fun updateState(
+    verification: Verification,
+    environment: Environment,
+    deliveryArtifact: DeliveryArtifact,
+    version: String,
+    status: VerificationStatus
+  )
 }
 
 data class VerificationState(
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
@@ -1,0 +1,25 @@
+package com.netflix.spinnaker.keel.api.verification
+
+import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.api.Verification
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import java.time.Instant
+
+interface VerificationRepository {
+  fun getVerificationState(
+    verification: Verification,
+    environment: Environment,
+    deliveryArtifact: DeliveryArtifact,
+    version: String
+  ) : VerificationState?
+}
+
+data class VerificationState(
+  val status: VerificationStatus,
+  val startedAt: Instant,
+  val endedAt: Instant?
+)
+
+enum class VerificationStatus(val complete: Boolean) {
+  RUNNING(false), PASSED(true), FAILED(true)
+}
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
@@ -1,0 +1,104 @@
+package com.netflix.spinnaker.keel.persistence
+
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.api.Verification
+import com.netflix.spinnaker.keel.api.verification.VerificationContext
+import com.netflix.spinnaker.keel.api.verification.VerificationRepository
+import com.netflix.spinnaker.keel.api.verification.VerificationState
+import com.netflix.spinnaker.keel.api.verification.VerificationStatus
+import com.netflix.spinnaker.keel.artifacts.DockerArtifact
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.EnumSource
+import strikt.api.expectCatching
+import strikt.assertions.isEqualTo
+import strikt.assertions.isNotNull
+import strikt.assertions.isNull
+import strikt.assertions.isSuccess
+import java.io.Serializable
+
+abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationRepository> {
+
+  abstract fun createSubject(): IMPLEMENTATION
+
+  val subject: IMPLEMENTATION by lazy { createSubject() }
+
+  open fun VerificationContext.setup() {}
+
+  private data class DummyVerification(val id: Serializable) : Verification {
+    override val type = "dummy"
+  }
+
+  @Test
+  fun `an unknown verification has a null state`() {
+    val verification = DummyVerification(1)
+    val context = VerificationContext(
+      deliveryConfig = DeliveryConfig(
+        application = "fnord",
+        name = "fnord-manifest",
+        serviceAccount = "jamm@illuminati.org",
+        artifacts = setOf(
+          DockerArtifact(
+            name = "fnord",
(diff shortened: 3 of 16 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
```diff
@@ -127,4 +128,7 @@ class SqlConfiguration {
     clock: Clock,
     properties: SqlProperties
   ) = SqlUnhealthyRepository(clock, jooq, SqlRetry(sqlRetryProperties))
+
+  @Bean
+  fun verificationRepository() = SqlVerificationRepository()
```
robfletcher, line 133: Just so spring-based tests don't break

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
```diff
@@ -0,0 +1,9 @@
+package com.netflix.spinnaker.keel.api.plugins
+
+import com.netflix.spinnaker.keel.api.Verification
+
+interface VerificationEvaluator<VERIFICATION: Verification> {
+  val supportedVerification: Pair<String, Class<out Verification>>
+
+  fun evaluate()
```
robfletcher, line 8: Obviously this will need parameters

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -0,0 +1,38 @@
...
+) {
+  fun runVerificationsFor(environment: Environment, artifact: DeliveryArtifact, version: String) {
+    val statuses = environment
+      .verifyWith
+      .map { it to verificationRepository.getVerificationState(it, environment, artifact, version)?.status }
+
+    if (statuses.any { it.second == RUNNING }) {
+      log.debug("Verification already running for {}", environment.name)
+      return
+    }
```
robfletcher, line 24: No-op if anything is already running

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -0,0 +1,38 @@
...
+) {
+  fun runVerificationsFor(environment: Environment, artifact: DeliveryArtifact, version: String) {
+    val statuses = environment
+      .verifyWith
+      .map { it to verificationRepository.getVerificationState(it, environment, artifact, version)?.status }
+
+    if (statuses.any { (_, status) -> status == RUNNING }) {
+      log.debug("Verification already running for {}", environment.name)
+      return
+    }
```
robfletcher, line 24: no-op if anything is already running

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    val verificationToRun = statuses.firstOrNull { (_, status) -> status == null }?.first
+    if (verificationToRun != null) {
+      val evaluator = evaluators.first {
+        it.supportedVerification.first == verificationToRun.type
+      }
+      evaluator.evaluate()
+    } else {
+      log.debug("Verification complete for {}", environment.name)
+    }
```
robfletcher, line 34: start the first verification, or just exit if all are complete

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -0,0 +1,44 @@
...
+      .map {
+        it to verificationRepository
+          .getState(it, environment, artifact, version)
+          ?.status
+      }
+
+    if (statuses.any { (_, status) -> status == RUNNING }) {
+      log.debug("Verification already running for {}", environment.name)
+      return
+    }
```
robfletcher, line 28: I'm thinking we probably need to call another method on `VerificationEvaluator` here to see if it's done yet and store the updated state. Still thinking about how that should look.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
```diff
@@ -0,0 +1,33 @@
+package com.netflix.spinnaker.keel.api.verification
+
+import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.api.Verification
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import java.time.Instant
+
+interface VerificationRepository {
+  fun getState(
```
lorin, line 9: Note that a verification may have run multiple times (e.g., test suite failed, user triggered a re-run, which passed). This means that there could be multiple VerificationState objects associated with a (verification, environment, artifact, version) tuple.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
```diff
@@ -0,0 +1,33 @@
...
+}
+
+data class VerificationState(
+  val status: VerificationStatus,
+  val startedAt: Instant,
+  val endedAt: Instant?
+)
+
+enum class VerificationStatus(val complete: Boolean) {
+  RUNNING(false), PASSED(true), FAILED(true)
```
lorin, line 32: We might want an `ERROR` state to model cases where something unexpected happened that prevented keel from carrying out the verification.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
robfletcher: Not trying to solve for that yet

lorin (approved):

lorin (approved):
Still looks good to me

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -0,0 +1,84 @@
...
+        .select(
+          VERIFICATION_STATE.STATUS,
+          VERIFICATION_STATE.STARTED_AT,
+          VERIFICATION_STATE.ENDED_AT
+        )
+        .from(VERIFICATION_STATE)
+        .where(VERIFICATION_STATE.ENVIRONMENT_UID.eq(environmentUid))
+        .and(VERIFICATION_STATE.ARTIFACT_UID.eq(artifact.uid))
+        .and(VERIFICATION_STATE.ARTIFACT_VERSION.eq(version))
+        .fetchOneInto<VerificationState>()
```
robfletcher, line 39: Fetch state for a verification in an environment. Note there's no column for the verification yet. That's next.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -0,0 +1,84 @@
...
+        .insertInto(VERIFICATION_STATE)
+        .set(VERIFICATION_STATE.STATUS, status.name)
+        .set(status.timestampColumn, currentTimestamp())
+        .set(VERIFICATION_STATE.ENVIRONMENT_UID, environmentUid)
+        .set(VERIFICATION_STATE.ARTIFACT_UID, artifact.uid)
+        .set(VERIFICATION_STATE.ARTIFACT_VERSION, version)
+        .onDuplicateKeyUpdate()
+        .set(VERIFICATION_STATE.STATUS, status.name)
+        .set(status.timestampColumn, currentTimestamp())
+        .execute()
```
robfletcher, line 58: Insert / update verification state. Also no verification column yet.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
```diff
@@ -0,0 +1,12 @@
+package com.netflix.spinnaker.keel.api.plugins
+
+import com.netflix.spinnaker.keel.api.Verification
+import com.netflix.spinnaker.keel.api.verification.VerificationStatus
+
+interface VerificationEvaluator<VERIFICATION: Verification> {
+  val supportedVerification: Pair<String, Class<out Verification>>
```
luispollo, line 7: This should be using the generic type, right?
```suggestion
  val supportedVerification: Pair<String, Class<out VERIFICATION>>
```

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
```diff
@@ -0,0 +1,12 @@
...
+
+  fun evaluate() : VerificationStatus
+
+  fun start(verification: Verification)
```
luispollo, line 11: This should be using the generic type, right?
```suggestion
  fun start(verification: VERIFICATION)
```

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
```diff
@@ -0,0 +1,12 @@
+package com.netflix.spinnaker.keel.api.plugins
+
+import com.netflix.spinnaker.keel.api.Verification
+import com.netflix.spinnaker.keel.api.verification.VerificationStatus
+
+interface VerificationEvaluator<VERIFICATION: Verification> {
```
luispollo, line 6: Docs?

luispollo: Also, if we're planning to make this part of the exposed plugin interfaces, might as well mark it as `SpinnakerExtensionPoint`.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
```diff
@@ -0,0 +1,151 @@
...
+
+  val subject: IMPLEMENTATION by lazy { createSubject() }
+
+  open fun VerificationContext.setup() {}
+
+  private data class DummyVerification(override val id: String) : Verification {
+    override val type = "dummy"
+  }
+
+  @Test
```
luispollo, line 34: Just curious, since you mentioned this in chat recently: should we be discussing switching to JUnit as the new standard, in your opinion? Prons/cons compared to minutests?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.netflix.spinnaker.keel.api.verification.VerificationStatus.RUNNING
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+
+@Component
+class VerificationRunner(
+  private val verificationRepository: VerificationRepository,
+  private val evaluators: List<VerificationEvaluator<*>>
+) {
+  fun runVerificationsFor(context: VerificationContext) {
```
luispollo, line 17: Really nice use of extension functions in this method to abstract details and help the code tell a story. 👏

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunnerTests.kt
```diff
@@ -0,0 +1,234 @@
...
+      evaluator.start(any())
+    }
+  }
+
+  @ParameterizedTest(
+    name = "continues to the next if any verification was already running and has now completed with {0}"
+  )
+  @EnumSource(
+    value = VerificationStatus::class,
+    names = ["PASSED", "FAILED"]
```
luispollo, line 141: I assumed we would _not_ run a downstream verification if the previous one failed, i.e. that we would short-circuit in that case. What's the rationale for running it anyway?

keel-sql/src/main/resources/db/changelog/20201112-add-verification-repository.yml
```diff
@@ -0,0 +1,55 @@
+databaseChangeLog:
+- changeSet:
+    id: add-verification-repository
+    author: fletch
+    changes:
+    - createTable:
+        tableName: verification_state
+        columns:
+          - column:
+              name: environment_uid
```
luispollo, line 10: Maybe add FKs for these as we discussed?

luispollo (commented):
This looks great, @robfletcher, except for the generic type usage in the interface as I pointed out inline. Nice to be able to add some fresh new code to keel for a change, huh? 🙂

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunnerTests.kt
robfletcher: Avoiding playing whack-a-mole where you fix one thing only to realize the next is broken

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
robfletcher: Hmm, I think that should actually just be `Class<VERIFICATION>`

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
robfletcher: Hmm, it's actually annoying because then you'd have to make a cast at the call-site

keel-sql/src/main/resources/db/changelog/20201112-add-verification-repository.yml
robfletcher: Seperate PR

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
robfletcher: Eh, I just find performance in IntelliJ bad when there are many/large nested lambdas (which is the entire basis of Minutest's API). I don't know that I feel strongly enough to say "this is what we should do" but I thought I'd experiment with just using JUnit here

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunnerTests.kt
luispollo: In that case, shouldn't they run in parallel?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
luispollo: Ah, right. I guess just in the `supportedVerification` field above, then. I just noticed we were not using the generic type anywhere in the interface. 🙂

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
luispollo: I wish we could positively blame that on the lambdas, but IntelliJ has been slow for me for all projects lately, so I don't know if that's the only reason. 🤷‍♂️  

In any case, I've no objections to using JUnit. I'm interested in learning more about how to model tests using this approach, since I understand there's no equivalent to the "build-up" approach we use with minutest nested contexts, which helps with reuse (but adds complexity). I'd appreciate if you can point me to examples as you experiment with it. 🙏

luispollo (approved):

keel-sql/src/main/resources/db/changelog/20201112-add-verification-repository.yml
```diff
@@ -0,0 +1,55 @@
...
+              type: char(26)
+              constraints:
+                nullable: false
+          - column:
+              name: artifact_uid
```
lorin, line 15: Good opportunity for our first foreign key...

lorin (approved):

keel-sql/src/main/resources/db/changelog/20201112-add-verification-repository.yml
robfletcher: Could have at least 2 on this table, yes. I'll probably add it in a separate PR.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunnerTests.kt
robfletcher: Still the subject of debate. Lorin & I thought running sequentially was the simplest thing to start with.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
robfletcher: Yeah, I'd definitely like to refactor some of the common data setup here. There are some subtle differences but I think a couple of factory methods could probably go a long way.

PR #1656, opened 2020-11-11
feat(artifactDiff): Adding versions diffing link between two artifact versions

gal-yardeni:
This is a new addition to the application summary endpoint.
For each artifact version, in an environment, we are calculating the source and the target versions and generating a diff link in stash. 
Looking for feedback on the approach / any suggestions are welcomed! 
Still to do:
- [x] Add tests

An example of the result would be:
```
 "artifacts": [
    {
      "name": "gyardeni-app-server",
      "type": "deb",
      "reference": "gyardeni-deb",
      "versions": [
        {
          "version": "gyardeni-app-server-0.0.1~snapshot-h226.c25a357",
          "displayName": "0.0.1~snapshot",
          "environments": [
            {
              "name": "testing",
              "state": "current",
              "deployedAt": "2020-11-10T01:49:16.187Z",
              "compareLink": "https://stash.corp.netflix.com/projects/~gyardeni/repos/gyardeni-app/compare/commits?targetBranch=c25a3579d390226ac94620774e0c13ab6efabc63&sourceBranch=899feabba0516a815d06b4494cc868f4bf84ff4e"
            }
          ],
          "build": { [...] },
          "git": {[...]}
          },
          "createdAt": "2020-11-05T23:46:31.720Z"
        },
        {
          "version": "gyardeni-app-server-0.0.1~snapshot-h224.992bc68",
          "displayName": "0.0.1~snapshot",
          "environments": [
            {
              "name": "testing",
              "state": "previous",
              "deployedAt": "2020-11-10T01:05:01.036Z",
              "replacedAt": "2020-11-10T01:49:16.200Z",
              "replacedBy": "gyardeni-app-server-0.0.1~snapshot-h226.c25a357",
              "compareLink": "https://stash.corp.netflix.com/projects/~gyardeni/repos/gyardeni-app/compare/commits?targetBranch=c25a3579d390226ac94620774e0c13ab6efabc63&sourceBranch=899feabba0516a815d06b4494cc868f4bf84ff4e"
            }
          ],
          "build": {  [...]  },
            "completedAt": "2020-11-05T23:17:19.538Z",
            "startedAt": "2020-11-05T23:15:16.836Z",
            "status": "SUCCESS"
          },
          "git": { [...] },
          "createdAt": "2020-11-04T23:46:31.720Z"
        },

### CODE REVIEW

File: keel-core/src/main/java/com/netflix/rocket/semver/shaded/DebianVersionComparator.java [hunk 1]
```diff
@@ -38,44 +38,52 @@ private int compareVersionPart(String s0, String s1) {
...
-        String[] prerelease0 = mainPrerelease0[1].split(PRERELEASE_SEPARATOR);
-        String[] prerelease1 = mainPrerelease1[1].split(PRERELEASE_SEPARATOR);
-        for (int i = 0; i < prerelease0.length; i++) {
-          if (i >= prerelease1.length) {
-            if (prerelease0[i].equals("dev")) {
-              return -1;
-            } else {
-              return 1;
-            }
+    if (mainComparison != 0) {
```
luispollo, line 41: The changes in this file seem unrelated to your PR?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -233,6 +234,16 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
    */
   fun deletePin(deliveryConfig: DeliveryConfig, targetEnvironment: String, reference: String)
 
+  /**
+   * Given artifact details and promotion status, return its last deployed version
+   */
+  fun getGitMetadataByPromotionStatus(
+    deliveryConfig: DeliveryConfig,
+    environmentName: String,
+    artifactReference: String,
```
luispollo, line 243: Consider passing the `DeliveryArtifact` object here instead. I'm finding so many places in my PR where we need to do unnecessary joins between tables just because the methods take only an identifier and not the whole object. This would avoid adding to that problem.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -233,6 +234,16 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
    */
   fun deletePin(deliveryConfig: DeliveryConfig, targetEnvironment: String, reference: String)
 
+  /**
+   * Given artifact details and promotion status, return its last deployed version
```
luispollo, line 238: The description of the return here does not match the actual type of the return...

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -320,28 +344,54 @@ class ApplicationService(
...
+  private fun generateDiffLink (newerGitMetadata: GitMetadata?, olderGitMetadata: GitMetadata?) : String? {
+    return if (newerGitMetadata !=null && olderGitMetadata != null) {
+      "https://stash.corp.netflix.com/projects/${newerGitMetadata.project}/repos/${newerGitMetadata.repo?.name}/compare/diff?" +
+        "targetBranch=${olderGitMetadata.commitInfo?.sha}&sourceBranch=${newerGitMetadata.commitInfo?.sha}"
+    } else {
+      null
+    }
+  }
+
+  private fun getPotentialSummary (deliveryConfig: DeliveryConfig, environmentName: String, artifactReference: String, version: String) =
```
luispollo, line 389: This is just wrapping a single call with the same number of parameters. What's the goal of this function?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -220,21 +225,24 @@ class ApplicationService(
...
-        state = status.name.toLowerCase()
-      )
+      PENDING -> {
+        val newerGitMetadata = repository.getGitMetadataByPromotionStatus(deliveryConfig, environmentName,  artifact.reference, CURRENT.name)
+
+        ArtifactSummaryInEnvironment(
+          environment = environmentName,
+          version = version,
+          state = status.name.toLowerCase(),
+          // comparing CURRENT (new code) vs. PENDING (version is question, old code)
```
luispollo, line 239: Hmm, isn't that backwards? `PENDING` means it hasn't been approved/deployed yet. How can that be older than `CURRENT`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -245,12 +253,28 @@ class ApplicationService(
...
-      else -> repository.getArtifactSummaryInEnvironment(
-        deliveryConfig = deliveryConfig,
-        environmentName = environmentName,
-        artifactReference = artifact.reference,
-        version = version
-      )
+      DEPLOYING -> {
+        val olderGitMetadata = repository.getGitMetadataByPromotionStatus(deliveryConfig, environmentName, artifact.reference, CURRENT.name)
+        potentialSummary?.copy(
+          // comparing DEPLOYING (version is question, new code) vs. CURRENT (old code)
```
luispollo, line 259:
```suggestion
          // comparing DEPLOYING (version in question, new code) vs. CURRENT (old code)
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -220,21 +225,24 @@ class ApplicationService(
...
+        ArtifactSummaryInEnvironment(
+          environment = environmentName,
+          version = version,
+          state = status.name.toLowerCase(),
+          // comparing CURRENT (new code) vs. PENDING (version is question, old code)
+          diffLink = generateDiffLink(newerGitMetadata, artifactGitMetadata)
+        )
+      }
       SKIPPED -> {
         // some environments contain relevant info for skipped artifacts, so
```
luispollo, line 244: This comment was about the call immediately below, which you moved to the top of the function. Please move the comment as well.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -245,12 +253,28 @@ class ApplicationService(
...
+        val olderGitMetadata = repository.getGitMetadataByPromotionStatus(deliveryConfig, environmentName, artifact.reference, CURRENT.name)
+        potentialSummary?.copy(
+          // comparing DEPLOYING (version is question, new code) vs. CURRENT (old code)
+          diffLink = generateDiffLink(artifactGitMetadata, olderGitMetadata)
+        )
+      }
+      PREVIOUS -> {
+        val newerGitMetadata = potentialSummary?.replacedBy?.let { getArtifactInstance(artifact, it)?.gitMetadata }
+        potentialSummary?.copy(
+          //comparing PREVIOUS (version is question, old code) vs. the version which replaced it (new code)
```
luispollo, line 266:
```suggestion
          //comparing PREVIOUS (version in question, old code) vs. the version which replaced it (new code)
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -245,12 +253,28 @@ class ApplicationService(
...
+          diffLink = generateDiffLink(newerGitMetadata, artifactGitMetadata)
+        )
+      }
+      CURRENT -> {
+        val olderGitMetadata = repository.getGitMetadataByPromotionStatus(deliveryConfig, environmentName, artifact.reference, PREVIOUS.name)
```
luispollo, line 271: My understanding is that there may be potentially many versions with a `PREVIOUS` status. Which one will this pick?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -320,28 +344,54 @@ class ApplicationService(
...
+
+  private fun getArtifactInstance(
+    artifact: DeliveryArtifact,
+    version: String
+  ): PublishedArtifact? {
+    val releaseStatus = repository.getReleaseStatus(artifact, version)
+    return repository.getArtifactInstance(artifact.name, artifact.type, version, releaseStatus)
+  }
+
+  // Generating a stash diff link between source and target versions (the order does matter!)
```
luispollo, line 379: Consider removing specific mentions of Stash in this code, since it must work regardless of the SCM being used.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -320,28 +344,54 @@ class ApplicationService(
...
+  private fun generateDiffLink (newerGitMetadata: GitMetadata?, olderGitMetadata: GitMetadata?) : String? {
+    return if (newerGitMetadata !=null && olderGitMetadata != null) {
+      "https://stash.corp.netflix.com/projects/${newerGitMetadata.project}/repos/${newerGitMetadata.repo?.name}/compare/diff?" +
```
luispollo, line 382: This base URL cannot be hard-coded here, but instead retrieved at runtime to reflect what is configured in the Spinnaker installation for this particular artifact. 

Igor has an endpoint you can call to retrieve that information:
```
GET https://{{igor.host}}/scm/masters
{
  "stash": "<url>",
  "gitHub": "<url>"
}
```
The code might look something like this:
```kotlin
// The `scmType` field is currently missing from the model and is something you'll have to add/fill in
val baseUrl =  igorService.getScmUrl(newerGitMetadata.scmType)
when (newerGitMetadata.scmType) {
  "stash" -> "$baseUrl/projects/${newerGitMetadata.project}/repos/${newerGitMetadata.repo?.name}/compare/diff?" // ...
  "github" -> // ...
  else -> throw UnsupportedScmType(newerGitMetadata.scmType)
}
```
P.S. Whenever you find yourself using an internal Netflix URL like you have here, that's generally a good sign to take a pause and reevaluate, including to avoid leaking potentially sensitive information. 🙂

luispollo (changes requested):
@gal-yardeni This is pretty exciting! Thanks for working on it. I think the approach looks good, but there's currently an issue with hard-coding the base URL for the SCM service which is why I'm requesting changes.

keel-core/src/main/java/com/netflix/rocket/semver/shaded/DebianVersionComparator.java
gal-yardeni: This is the rocket fix for the snapshot versioning compression. It's a copy-paste from their file.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
gal-yardeni: true! changed.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: reverted it and deleted this function

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: you are right. It was the oposite. Thanks!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: done

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: the function `getGitMetadataByPromotionStatus` is querying `environemnt_artifact_version` by promotion status and other params, and will return the last `deployed_at` version. Using this version, we are querying to get its `gitMetadata` for the comparsion. Do you think it could be error-prone?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: Thanks for the detailed comment! I didn't think using a stash link is considered as sensitive information :) but I totally agree it can be better! will make the change.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: every SCM has a different comparison implementation. I didn't check how it was implemented in other SCM (like git etc.) as I think it's out of scope at the moment. I can address it in a follow-up PRs, if the team feels it's nessacery.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: @lpollo you’ve suggested that I’ll use igor to fetch the base url, but I’m having a dependencies issue as the applicationService code is in keel-core and igor is in keel-igor , so I can’t add igor to that module or I’ll have a circled dependency.. no?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
gal-yardeni: can you elaborate on why using `DeliveryArtifact` here will be better? for this function, I need the `deliveryConfig` abject as a whole, which I don't have in `DeliveryArtifact`.

erikmunson (commented):
SO COOL. I read through the logic for handling comparisons on each of the different promotion statuses which all looked great in terms of UX from my perspective.

My only question: since in this initial MVP we won't be showing commits in the UI itself, does it make sense to switch the Stash URL to point to the `commits` tab instead of the `diff` tab so folks can get a quick overview of changes before diving into the code diff itself? If you do choose to switch the tab to `commits`, it may also make sense to rename the field in the API data to `compareLink` or similar.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -1199,6 +1200,41 @@ class SqlArtifactRepository(
...
+        .select(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
+        .from(ENVIRONMENT_ARTIFACT_VERSIONS)
+        .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(deliveryConfig.getUidFor(environmentName)))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifact.uid))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.eq(promotionStatus))
+        .orderBy(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT.desc())
+        .fetch(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
+        .firstOrNull()
+
+      jooq.select(
```
emjburns, line 1224: weird spacing here, can you do
```jooq
  .select(blah)``` instead please?

File: keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt [hunk 13]
```diff
@@ -1199,6 +1200,41 @@ class SqlArtifactRepository(
     }
   }
```
emjburns, line 1202: It's probably worth adding a comment here that says "This will fetch the most recent artifact by the 'deployed_at" field"

Also, because you're sorting by deployed at, it won't work as you expect for artifacts that are APPROVED or SKIPPED or DEPLOYING. Do you think it's worth making that obvious by adding a:

```kotlin
require(promotionStatus in listOf(CURRENT, PREVIOUS)) { "Invalid promotion status used to query" }
```
(the require function is a kotlin thing that makes argument checking easier: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require.html)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
emjburns: I think luis is suggesting instead of artifact reference passing in the whole artifact object. However, since we need the artifact UUID, I don't think it matters here.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: fixed!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: @emjburns for `DEPLOYING`, I'm comparing against `CURRENT`. (`SKIPPED` and `APPROVED` are out of scope for this PR, for now, but I think it would make sense to compare them against the current as well). Do you still think it can cause issues?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
gal-yardeni: Going to keep it as is for now

keel-core/src/main/java/com/netflix/rocket/semver/shaded/DebianVersionComparator.java
luispollo: Right. So unrelated to diffing... 😄

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
luispollo: Yes, that is what I was suggesting.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: Sorting by `deployed_at` and getting the latest sounds OK to me.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: Regardless of whether you address the specifics of other SCM systems later, my point was about encapsulation and extensibility. Why hard-code comments and implementations when you can easily make it generic and future-proof?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -233,6 +234,16 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
    */
   fun deletePin(deliveryConfig: DeliveryConfig, targetEnvironment: String, reference: String)
 
+  /**
+   * Given artifact details and promotion status, return its last deployed version git metadata, sorted by deployed_at
```
luispollo, line 238:
```suggestion
   * Return the git metadata for the last deployed version of the specified artifact that matches the promotion status
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: added `require`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -344,4 +379,37 @@ class ApplicationService(
...
+
+  // Calling igor to fetch all base urls by SCM type, and returning the right one based on current commit link
+  private fun getScmBaseLink(commitLink: String): String? {
+    val scmInfo = runBlocking {
+      scmInfo.getSCMInfo()
+    }
+    //TODO[gyardeni]: replace this parsing when rocket will add scm type to gitMetadata
+    val currentScm = commitLink.substring(commitLink.indexOf("/") +2, commitLink.indexOf("."))
+    when(currentScm) {
+      "stash" ->
```
luispollo, line 409: Maybe a bit less brittle this way?
```suggestion
    when {
      "stash" in commitLink ->
```

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/SCMInfoService.kt
```diff
@@ -0,0 +1,14 @@
+package com.netflix.spinnaker.keel.services
+
+import com.netflix.spinnaker.igor.ScmService
+import com.netflix.spinnaker.keel.api.SCMInfo
+import org.springframework.stereotype.Component
+
+@Component
+class SCMInfoService(
```
luispollo, line 8: What's the purpose of this component? Any reason you can't just wire in `ScmService` and have it extend the new `SCMInfo` interface you created?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -1199,6 +1200,42 @@ class SqlArtifactRepository(
...
 
+  override fun getGitMetadataByPromotionStatus(
+    deliveryConfig: DeliveryConfig,
+    environmentName: String,
+    artifactReference: String,
+    promotionStatus: String
+  ): GitMetadata? {
+    return sqlRetry.withRetry(READ) {
+
+      val artifact = deliveryConfig.artifacts.firstOrNull { it.reference == artifactReference }
```
luispollo, line 1211: These lines here alone would be a good reason to replace the `artifactReference` parameter with `artifact: DeliveryArtifact`. There was not a single occurrence of other repository functions I'v refactored in this way where the call site did not already have an artifact object handy to pass down, and it avoids these kinds of unnecessary lookups.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -1199,6 +1200,42 @@ class SqlArtifactRepository(
...
+        .select(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
+        .from(ENVIRONMENT_ARTIFACT_VERSIONS)
+        .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(deliveryConfig.getUidFor(environmentName)))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifact.uid))
+        .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.eq(promotionStatus))
+        .orderBy(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT.desc())
+        .fetch(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
+        .firstOrNull()
+
+      jooq
```
luispollo, line 1224: You should only execute this select here if `version != null`, right?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: @luispollo there's a cost to making something "generic and future-proof", and I think a lot of times it's hard to know what the future will look like. A generic implementation that is a little bit wrong is often pretty hard to debug and to change. On the other side, it's fairly easy to refactor to a generic implementation later on when it's needed.

luispollo (commented):
Much better. Just left you a few other comments that would be nice to address, but, most importantly, I think you should write tests as part of this PR.

keel-core/src/main/java/com/netflix/rocket/semver/shaded/DebianVersionComparator.java
gal-yardeni: yes.. just my laziness for splitting PRs :)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: oh totally!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: yep, added that check

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: I now see what you are saying @luispollo ! Thanks for pointing it out. Changed to `DeliveryArtifact`.

erikmunson:
forgot to mention earlier: somewhere in the comments from you mentioned that it might make sense to add this link to the `APPROVED` promotion status as well — I'd +1 that idea (even if it comes in a later PR).

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/services/SCMInfoService.kt
gal-yardeni: done

gal-yardeni:
@erikmunson let's talk about `approved` and `skipped` tomorrow -- going to address it later

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/SCMInfo.kt
```diff
@@ -0,0 +1,7 @@
+package com.netflix.spinnaker.keel.api
+
+interface SCMInfo{
```
luispollo, line 3: nit:
```suggestion
interface ScmInfo{
```

luispollo (approved):

luispollo:
Thanks for addressing my feedback, @gal-yardeni! 🙏

PR #1657, opened 2020-11-13
feat(predeploy): pre-deploy events, bake implementation

emjburns:
PR for "pre-deployment" events, which at this moment include only bakes. This will expand to include builds.

**General**
Each of these events is tied to an artifact (a delivery artifact) and a version of that specific artifact.
I'm trying to make this flexible enough so that potentially in the future we could also migrate away from using the task view for deployments (even though we'd still be using tasks) and surface them in the environments view.

The key thing about these events: if you want to create multiple events in a series that roll up to the same "summary" (lifecycle step) you can provide an id. if you don't provide an id, it'll be a one-off event, and that's fine.

Right now I only have one scope (pre-deployment) and one type (bake). I think this gives us the flexibility to add other scopes and types in the future, but we don't need to over-index on that flexibility until we get something implemented. 

**Monitoring**
Lifecycle events can be monitored. I've implemented a bake monitor so that we get start and ending events for the bake, and so that we get a lovely clickable link, directly to the task. I've created a generic monitoring abstract class that gets what should be monitored, and then individual handlers actually do the checks.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
```diff
@@ -0,0 +1,49 @@
...
+  val text: String
+  val link: String
+  val startTime: Instant?
+  val endTime: Instant?
+}
+
+/**
+ * A specific lifecycle event with some of the details filled in.
+ */
+data class BakeEvent(
```
robfletcher, line 38: This should be in `keel-bakery`, no?

robfletcher (commented):
> The key thing about these events: if you want to update them, you can provide an id. If you don't want to update them, you don't have to provide an id.

Why would you update an event?

emjburns:
@robfletcher you'd update an event to give it more data. Like, recording that a bake finished successfully or failed with problems. Maybe this is a weird way to model it, or calling it "event" is not the right wording. Basically, we need to fill in the little card in the mocks with the right information which changes over time and then eventually stops changing (like when the bake is finished)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
emjburns: Hmm, ya probably.

robfletcher:
> @robfletcher you'd update an event to give it more data. Like, recording that a bake finished successfully or failed with problems. Maybe this is a weird way to model it, or calling it "event" is not the right wording. Basically, we need to fill in the little card in the mocks with the right information which changes over time and then eventually stops changing (like when the bake is finished)

I see. Yeah, I feel like "event" is maybe the wrong word then. To me, if it's modeled as events you'd have separate ones for start and end rather than updating the one.

lorin:
The only feedback I have is to echo @robfletcher about the event model. I also think of events as immutable values that occur at a single point in time (which is how things are modeled in our `events` table).

If we go with an immutable event model, there are basically two choices on how to handle a summary of the state, which is what the user typically cares about:

* have a separate table that models the current state, and update the record in addition toadding events
* compute the state each time from the stream of events

luispollo:
Hi @emjburns, I also echo @robfletcher's and @lorin's comments. In my mind, following the pattern we've been using for events elsewhere, you'd model this as a "start bake" event and a "end bake" event.

Speaking of patterns, I was also surprised to see that you defined an entirely new class hierarchy and repository for these events, since we already have a generic model for this in `PersistentEvent` (currently extended by `ResourceEvent` and `ApplicationEvent`. The table used to persist those has a `scope` column which you could use in the same fashion as you did in this PR, without having to add a separate repository. It also has a `ref` column to store the unique ID of the thing you're referencing in the event, a timestamp, and the body of the event as JSON. I couldn't think of anything you couldn't fit in this model, given your examples below.

Finally, I'd suggest using a scope of `ARTIFACT` for these events, with specific sub-types captured by the `type` field in the JSON (which is already modeled for you via `PersistentEvent`).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * [artifactVersion] is the version of the artifact that this event pertains to.
+ *
+ * [scope] and [type] control where the event is shown, right now there is only one of each.
+ * [id] is a caller-provided id: if the caller wants to be able to update the event,
+ *  it should not be random. If the event is fire and forget, the id can be null.
+ * [link] stores a link that will be surfaced in a clickable button.
+ * [text] is the text for the even that is shown to the user.
+ */
+interface LifecycleEvent {
```
gal-yardeni, line 22: should we add here `createdBy` or `author` or something that will tell us who initiated this event?

emjburns: These will all be generated in code, and I think the "type" should tell us who initiated them (like bake events will be initiated by the image handler)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -370,7 +373,8 @@ class ApplicationService(
       build = artifactInstance.buildMetadata
         ?: artifactSupplier.parseDefaultBuildMetadata(artifactInstance, artifact.versioningStrategy),
       git = artifactInstance.gitMetadata
-        ?: artifactSupplier.parseDefaultGitMetadata(artifactInstance, artifact.versioningStrategy)
+        ?: artifactSupplier.parseDefaultGitMetadata(artifactInstance, artifact.versioningStrategy),
+      lifecycleEvents = lifecycleEventService.getEventsForArtifactAndVersion(artifact, artifactInstance.version)
```
gal-yardeni, line 377: I'm curious to see how those events will be shown in the UI. @erikmunson , do you have any thoughts about it?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
gal-yardeni: for the bakes it makes sense, but for builds or other stuff? we might need this information? (I'm not sure, just trying to think if something is missing).

gal-yardeni:
+1 to have separate events for start and end instead of update a record. I'm curious to see how this will play out in the UI, and will work with other things that we've discussed, like having a link to see the task details for `deploying`. Will `deploying` be considered as a lifecycle event?

robfletcher:
FWIW I'm down with _not_ modeling it as an event stream and just using a word other than "event". Either way is fine.

luispollo:
> @luispollo I would much rather create a new event for this - I think it seems better to build things for the specific purpose than trying to jam something into the specific model we have. The `PersistentEvent` model is really simple and doesn't give any benefits as far as I see...

Sorry, I don't follow. You _would_ build new events that are purpose-fit for your use case. I'm just advocating you don't reinvent the wheel for how those are persisted... What benefit do you get from writing a separate repository just for this, and not extending the existing model, which was designed for exactly this purpose?

luispollo:
Specifically on this point:
> Right now I only have one scope (pre-deployment) and one type (bake). I think this gives us the flexibility to add other scopes and types in the future, but we don't need to over-index on that flexibility until we get something implemented. I am thinking of storing the specific event info as a blob, so then if we add more info for other types of events it's easier to deal with storing them.

...here's what this might look like in the current `events` table:
| uid	| ref | timestamp |scope |json | application |
-- | -- | -- | -- | -- | --
... | ...	| 2020-11-13 08:04:36.292	| ARTIFACT	| `{"artifactVersion": "mddemo-ec2-master-h30.1be073d", "type": "BakeStarted", "scope": "artifact", "status": "running", "link": "http://bakery/etc", "timestamp": "2020-11-13T00:04:36.292177Z" }`	| mddemoec2

I just can't see how that's different from what you've proposed in your PR description.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -169,6 +177,16 @@ class ImageHandler(
         parameters = mapOf("delta" to diff.toDebug())
       )
       publisher.publishEvent(BakeLaunched(desiredVersion))
```
emjburns, line 179: This event, kicked off in the image handler, is our cue to start monitoring the bake. It is the first event in the series of events for bakes. We assume that a `NOT_STARTED` event is always the first in the chain.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -0,0 +1,120 @@
...
+  private val orcaService: OrcaService,
+  @Value("\${spinnaker.baseUrl}") private val spinnakerBaseUrl: String
+): LifecycleMonitor(monitorRepository, publisher, lifecycleConfig) {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  override fun handles(event: LifecycleEvent): Boolean =
+    event.type == BAKE
+
+  override fun typeHandled(): LifecycleEventType = BAKE
+
```
emjburns, line 41: this function allows (by default) 5 exceptions in fetching the status. tasks are monitored every minute right now, so this means it's 5 minutes before we give up on something.

keel-core/src/main/kotlin/com/netflix/spinnaker/config/LifecycleConfig.kt
```diff
@@ -0,0 +1,13 @@
+package com.netflix.spinnaker.config
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+import java.time.Duration
+
```
emjburns, line 5: I've set the config here to have 30s between checks of the same task. Is that too long? Too short?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
```diff
@@ -0,0 +1,28 @@
+package com.netflix.spinnaker.keel.lifecycle
+
+import java.time.Instant
+
```
emjburns, line 4: I'm specifically using `artifactRef` here (a string) instead of the actual artifact because I had serialization problems with storing the whole artifact. The config name was getting saved as null and this was causing the first event to be disjoint from the other events. I know that using the whole object is better in general, but I felt like this was okay here.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
emjburns: Resolving for now - we may need to expand this event in the future, and that's okay.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventService.kt
```diff
@@ -0,0 +1,12 @@
...
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import org.springframework.stereotype.Component
+
+@Component
+class LifecycleEventService (
+  val lifecycleEventRepository: LifecycleEventRepository
+) {
+  fun getStepsForArtifactAndVersion(deliveryArtifact: DeliveryArtifact, version: String): List<LifecycleStep> =
+    lifecycleEventRepository.getSteps(deliveryArtifact, version)
+}
```
robfletcher, line 12: This is just passing right through. Does this class add any value?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventRepository.kt
```diff
@@ -0,0 +1,21 @@
+package com.netflix.spinnaker.keel.lifecycle
+
+import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+
+interface LifecycleEventRepository {
```
robfletcher, line 5: I'm not 100% clear why we have this repository as well as `LifecycleMonitorRepository`

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -0,0 +1,120 @@
...
+  override suspend fun monitor(task: MonitoredTask) {
+    kotlin.runCatching {
+       orcaService.getOrchestrationExecution(task.link, DEFAULT_SERVICE_ACCOUNT)
+    }
+      .onSuccess { execution ->
+        when {
+          execution.status == BUFFERED -> publishCorrectLink(task)
+          execution.status == RUNNING -> publishRunningEvent(task)
+          execution.status.isSuccess() -> publishSucceededEvent(task)
+          execution.status.isFailure() -> publishFailedEvent(task)
```
robfletcher, line 51: I like this

keel-sql/src/main/resources/db/changelog/20201016-create-lifecycle-events.yml
```diff
@@ -0,0 +1,151 @@
...
+              - column:
+                  name: scope
+                  type: varchar(50)
+                  constraints:
+                    nullable: false
+              - column:
+                  name: ref
+                  type: varchar(255)
+                  constraints:
+                    nullable: false
```
robfletcher, line 24: Would it make sense to model this as a foreign key to `delivery_artifact`?

keel-sql/src/main/resources/db/changelog/20201016-create-lifecycle-events.yml
```diff
@@ -0,0 +1,151 @@
...
+              - column:
+                  name: type
+                  type: varchar(50)
+                  constraints:
+                    nullable: false
+              - column:
+                  name: id
+                  type: varchar(100)
+                  constraints:
+                    nullable: false
```
robfletcher, line 39: I'm getting a little confused what these fields refer to. Is `type` the type of event? What is `id`? Is that the bakery id? Is the table structure too fine grained in that case?

keel-sql/src/main/resources/db/changelog/20201016-create-lifecycle-events.yml
```diff
@@ -0,0 +1,151 @@
...
+              - column:
+                  name: last_checked
+                  type: timestamp(3)
+                  constraints:
+                    nullable: false
+              - column:
+                  name: triggering_event
+                  type: json
+                  constraints:
+                    nullable: false
```
robfletcher, line 115: Would it make sense to make this a foreign key to the row on `lifecycle`?

robfletcher (commented):
A scattering of comments & questions but this is exciting stuff. I'm really glad we're moving toward actually monitoring the bake rather than just fire-n-forgetting it.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -169,6 +176,16 @@ class ImageHandler(
...
+      publisher.publishEvent(LifecycleEvent(
+        scope = PRE_DEPLOYMENT,
+        artifactRef = artifact.toLifecycleRef(),
+        artifactVersion = desiredVersion,
+        type = BAKE,
+        id = "bake-$desiredVersion",
+        status = NOT_STARTED,
+        text = "Launching bake for $version",
+        link = taskRef.id
+      ))
```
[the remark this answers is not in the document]
emjburns: Ya, we publish both the new lifecycle event and also a bake launched event. The `BakeLaunched` event is used for telemetry

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
emjburns: Ya, it would require a db look up. That's the other way I could probably do it - have a foreign key to the actual artifact. What would you recommend?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventRepository.kt
emjburns: This is a history of events that have happened. The `LifecycleMonitorRepository` is a working list of things to check.

I thought about doing both in the same repository (monitoring and recording events) but I felt like it was much more confusing. I'd be happy to talk about it with you more, and share what it looks like with actual data.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventService.kt
emjburns: Currently, it definitely does not. I thought that the pattern we wanted to use was that only "service" classes that abstract the repository get wired into the controllers / etc. So, I created this service. I can also remove it and just wire the repository into the `ApplicationService`. If I do that, and the logic gets more complicated, it's easy to put a service in place. 

I know @luispollo usually has opinions about this kind of thing. Thoughts?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventType.kt
```diff
@@ -0,0 +1,6 @@
+package com.netflix.spinnaker.keel.lifecycle
+
+enum class LifecycleEventType {
+  BAKE
```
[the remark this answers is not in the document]
emjburns: I think because it's _only_ the enum, and not anything else, it is okay. Or at least my thought was that strong typing here gave enough benefit that it deserved an enum and not a string. Other details about how the event is monitored live in `keel-bakery`.

keel-sql/src/main/resources/db/changelog/20201016-create-lifecycle-events.yml
emjburns: I think it might! Can we chat offline about foreign keys? I haven't worked with them and wanted to ask about the benefits of modeling things using them.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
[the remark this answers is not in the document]
robfletcher: In theory you could listen to the `LifecycleEvent` in 2 different places (it's not a destructive read from the event bus) but this is fine.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventService.kt
robfletcher: Yeah. I'd default to adding it if & when needed rather than adding a level of redirection before there's a good reason to

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventRepository.kt
robfletcher: I guess I'm wondering if we need to persist both. I can see the need for the `LifecycleMonitorRepository` but do we actually need to persist the events? They just represent state changes in the task, right?

keel-sql/src/main/resources/db/changelog/20201016-create-lifecycle-events.yml
emjburns: type is the type of event (`BAKE` for now). Id is a caller-generated string that ties multiple events together. It is what we group by to summarize. So for example, the bake "id" right now is `bake-$version`. That allows us to group all the events for that bake into a summary and allows you to have multiple of the same type of event in different summaries (like, for example, a rebake because the first bake failed).

I'm in general a bit confused about what to put in table columns vs what to shove in a json blob. I've been thinking about it like: everything I might want to query or filter on I should put in a column. Does more columns = slower/more overhead? Is there a downside to more columns?

keel-sql/src/main/resources/db/changelog/20201016-create-lifecycle-events.yml
emjburns: Ya I think so! This is the other place I was thinking would be a good candidate for a foreign key

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -93,5 +93,7 @@ abstract class DeliveryArtifact {
       )
     ).normalized()
 
+  fun toLifecycleRef() = "$deliveryConfigName:$reference"
```
gal-yardeni, line 96: what does `toLifecycleRef` stands for?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -0,0 +1,120 @@
...
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.boot.context.properties.EnableConfigurationProperties
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.stereotype.Component
+import java.time.Duration
+
+@Component
+@EnableConfigurationProperties(LifecycleConfig::class)
+class BakeryLifecycleMonitor(
```
gal-yardeni, line 28: can you add some docs here?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -0,0 +1,120 @@
...
+  @Value("\${spinnaker.baseUrl}") private val spinnakerBaseUrl: String
+): LifecycleMonitor(monitorRepository, publisher, lifecycleConfig) {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  override fun handles(event: LifecycleEvent): Boolean =
+    event.type == BAKE
+
+  override fun typeHandled(): LifecycleEventType = BAKE
+
+  override suspend fun monitor(task: MonitoredTask) {
```
gal-yardeni, line 42: @emjburns I think this is duplicate code -- we have an orca scheduled agent `OrcaTaskMonitorAgent` which does the exact same thing, I think. Can you check it, please?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -0,0 +1,120 @@
...
+  private fun orcaTaskIdToLink(task: MonitoredTask): String =
+    "$spinnakerBaseUrl/#/tasks/${task.link}"
+
+  private fun publishCorrectLink(task: MonitoredTask) {
+    publisher.publishEvent(task.triggeringEvent.copy(
+      link = orcaTaskIdToLink(task)
+    ))
+  }
+
+  private fun publishRunningEvent(task: MonitoredTask) {
```
gal-yardeni, line 81: consider extracting those functions to util class or something - I believe it will be useful in other places as well :)

keel-core/src/main/kotlin/com/netflix/spinnaker/config/LifecycleConfig.kt
gal-yardeni: Can you elaborate on what is the purpose of this class? (and maybe add docs)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventRepository.kt
gal-yardeni: I think that the `LifecycleMonitorRepository` can be replaced in favor of the `task_tracking` table.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
```diff
@@ -0,0 +1,133 @@
...
+  @EventListener(LifecycleEvent::class)
+  fun onLifecycleEvent(event: LifecycleEvent) {
+    if (event.status == NOT_STARTED && handles(event) && event.link != null) {
+      log.debug("${this.javaClass.simpleName} saving monitor event $event")
+      monitorRepository.save(MonitoredTask(event, event.link))
+    }
+  }
+
+  @Scheduled(fixedDelayString = "\${keel.lifecycle-monitor.frequency:PT1S}")
+  fun invokeMonitoring() {
```
gal-yardeni, line 81: I believe this code can be replaced with the current code in `OrcaTaskMonitorAgent`, which monitors tasks.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventService.kt
emjburns: Removing, discussed with @robfletcher offline. This just adds a layer of misdirection

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleMonitorRepository.kt
```diff
@@ -0,0 +1,106 @@
...
+
+class SqlLifecycleMonitorRepository(
+  private val jooq: DSLContext,
+  private val clock: Clock,
+  private val objectMapper: ObjectMapper,
+  private val sqlRetry: SqlRetry
+) : LifecycleMonitorRepository {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  override fun tasksDueForCheck(type: LifecycleEventType, minTimeSinceLastCheck: Duration, limit: Int): Collection<MonitoredTask> {
```
gal-yardeni, line 25: consolidate with `task_tracking` somehow? 🙏

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
```diff
@@ -0,0 +1,28 @@
...
+data class LifecycleEvent(
```
gal-yardeni, line 5: can `LifecycleEvent` and `LifecycleStep` live in the same place?

gal-yardeni:
@emjburns your approach looks good to me. I worry that we have a couple of duplications here -- as we already have a tracking and monitoring mechanism for orca tasks, I wish we can consolidate those to work together. Another thing that is missing for me is more comments -- It's not super clear what are we defining as a `Lifecycle` event/step and so on. If you can add more, it'll be great!

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
robfletcher: Can we do that when we need to? I think private functions / extensions like this are a good way to factor code. They can be moved to a common location once we find a need for them elsewhere. It doesn't seem likely to me that these would be used in another context.

robfletcher:
Summarizing a conversation I had with Emily this morning…

There are 2 models for the data needed here that we discussed. The first is the event-based model that is used in this PR where the progress of the task is modeled as a sequence of events. This is optimal for write-heavy use cases as it's very easy to add more events. It's also ideal when there are many sources contributing to the event stream (e.g. if there were multiple Keel instances checking the status of the bake simultaneously). The downside is you need to synthesize the current state snapshot on read.

The second approach is using a single row on a table to model the task and updating it every time its state changes (e.g. setting last check time regularly, updating status & end time when it completes). This is slightly more complex at write time since you need to `update… where` but it makes life easier for read-heavy use cases as the table directly models the current state snapshot. You just have to read that one row.

My opinion is that the use case here is the latter: once a bake completes it never changes again. It will generate 5-20 minutes worth of events and then remain completely static. In addition there are really only 2 interesting events. The first tells us when we started baking, and the last tells us when we finished and whether it worked or not. However, we will _read_ the final state snapshot every 30 seconds so long as someone has the environments view open.

That said, the data model here is not _wrong_, and it would be straightforward to refactor to the other approach if necessary / desirable without changing the API.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
```diff
@@ -0,0 +1,28 @@
...
+  val scope: LifecycleEventScope,
+  val artifactRef: String,
+  val artifactVersion: String,
+  val type: LifecycleEventType,
+  val id: String? = null,
```
lorin, line 10: Does the code contain any events yet with id of null? I didn't see any. If not, I'd suggest making this field required for now, and not nullable in the database, and when we hit a use case with null id events, we can add it then.

keel-sql/src/main/resources/db/changelog/20201016-create-lifecycle-events.yml
```diff
@@ -0,0 +1,151 @@
+databaseChangeLog:
+  - changeSet:
+      id: create-lifecycle-events
+      author: emjburns
+      changes:
+        - createTable:
+            tableName: lifecycle
```
lorin, line 7: I'd suggest being more explicit with the table name and calling it `lifecycle_event`

keel-sql/src/main/resources/db/changelog/20201016-create-lifecycle-events.yml
```diff
@@ -0,0 +1,151 @@
...
+            columns:
+              - column:
+                  name: scope
+              - column:
+                  name: ref
+              - column:
+                  name: artifact_version
+              - column:
+                  name: id
```
lorin, line 82: I think there's a query/index mismatch:

The only select query on this table right now is the `getEvents` query, which looks like this:

```sql
select json, timestamp from lifecylce where 
ref=.<REF> and
artifact_version=<VER>
order by timestamp asc
```

This index `(scope, ref, artifact_version, id)` isn't well-suited to execute the query efficiently, because it has `scope` as the first entry in the index, and getEvents doesn't match on scope at all.

lorin (commented):
*Note: I'm trying an experiment where I start by explaining my understanding of the PR. This ends up being longer than my actual feedback, and probably won't be useful to you, but writing it forced me to get a better understanding of the PR.*

# Summary of my understanding

Today, the environments view does not provide the user with any feedback about the status of a bake, which is a prereq to deploying an EC2 app to the first environment. (Has the bake even started yet? Did it succeed?). 

This PR is initial work to supporting exposing this information in the UI. The PR adds logic to check and persist the status of tasks such as bakes. In this initial implementation, bake is the only supported type of task.

## Concepts introduced

The primary concepts introdubed in this PR are 

* the lifecycle step, modeled by the LifecycleStep class
* the monitored task, modeled by the MonitoredTask class
* the lifecycle event, modeled by the LifecycleEvent class. 


### Lifecycle step

A *lifecycle step* is the generic term for the thing that happens over time that's required for deployment, that the user cares about. A bake is currently our
only example of a lifecycle steps.

A step has:
 - scope (e.g., pre-deployment)
 - type (e.g., bake)
 - status (e.g., not started, running, scuceeded, failed, unknown)
 - start and end times 
 - additional metadata
   * user-meaningful text
   * url that links to an external service that is actually executing the task (e.g., orca, jenkins)
 
Keel creates a LifecycleStep by rolling up a set of lifecycle events associated with the same step. 

### Lifecycle event

Examples of a lifecycle event:
* bake requested (bake task submitted to orca)
* bake has started
* bake completed successfully
* bake failed

Every lifecycle event is associated with an artifact version (e.g., keel-0.809.0-h967.9b9aaab), and has the following information

 - scope (e.g., pre-deployment)
 - type (e.g., bake)
 - id (e.g., bake-keel-0.809.0-h967.9b9aaab)
 - status (e.g., not started, running, scuceeded, failed, unknown)
 - timestamp
 - additional metadata 
   *  user-meaningful text
   *  url that links to an external service that is actually executing the task (e.g., orca, jenkins)


The initial (not started) event is generated by the keel class responsible for kicking off the job, which is modeled as a *MonitoredTask*
(see below). A corresponding *LifecycleMonitor* sub-class (e.g., *BakeryLifecycleMonitor*) periodically checks on the state of the monitored task and emits lifecycle
events on each check based on the state of the monitredtask.

Keel persists lifecycle events in the LifecycleEventRespository

### Monitored task

This is the task that accomplishes the job (e.g., bake) associated with the lifecyle step. A bake job is an example of a monitored task. 

Keel persists monitor tasks in the MonitorTaskRepository.


# Feedback

Overall approach looks sound.

The only issue I saw that I really want to see addressed before merge is what looks like a mismatch between query and index, I left a comment inline.

## Things that tripped my radar

This is stuff that caught my attention to the level that I felt it was worth mentioning, but shouldn't block merge.

### Event generation

We might generate a lot of events for long-running tasks. I'm not so worried about database performance (especially at the current configuraiton of 30s between checks) as I am about the scenario of: "I want to take a dump of the database, but it's big because of these events that I don't actually care about". 

At some point in the future, we may want to write some code that cleans up old events we don't care about.

### Individual scheduling of lifecycle monitors

For resources and environments, we have a single scheduled loop. For lifecycle monitoring, each LifecycleMonitor subclass has its own `invokeMonitoring` schedule loop. I can't articulate a specific problem with this approach, but it just felt different from the architecture of the rest of the system.

### Alternate reality of similar concepts

This PR introduces new types of events and tasks.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
emjburns: it's the key i'm using to store events, basically a reference to the artifact for lifecycle events

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -0,0 +1,120 @@
...
+            "fetching the task status. Giving up.")
+          publishExceptionEvent(task)
+          monitorRepository.delete(task)
+        } else {
+          monitorRepository.markFailureGettingStatus(task)
+        }
+      }
+  }
+
+  private fun orcaTaskIdToLink(task: MonitoredTask): String =
```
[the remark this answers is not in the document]
emjburns: Ha! Well. Good question. Both "link" fields are ambiguous. The thought is the link contains something that a monitor knows how to deal with. Constructing the spinnaker "link" for the user requires reading from the config to get the base URL. I thought that fit better in the handlers rather than the place we emit events.

It might be better to have this function in the place where we summarize all the events for the API, and that way we could just take anything that looks like an orca task id and make it into a user link. but, what if something looks like a uid but isn't for orca? the monitor for that thing would know about it, and would be able to construct the right link for the user. So that's why I decided to allow link to be any text and rely on the monitor to update it to a user friendly link.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -0,0 +1,120 @@
...
+    "$spinnakerBaseUrl/#/tasks/${task.link}"
+
+  private fun publishCorrectLink(task: MonitoredTask) {
```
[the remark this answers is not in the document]
emjburns: Here I am prepending the "www.spinnaker.com/" bit, which we need to read from the config file. because of my answer to your above question, the link starts out as a non-clickable task id, but then gets made into a clickable link by the bakery monitor

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
emjburns: I don't want to because the thing we are actually doing is changing the text - which will be specific to each monitor.  Also, these are the tiniest baby functions. Feels ok to repeat them in other monitors (eventually) with other text.

As I build the build monitoring, though, I will see what can be extracted.

keel-core/src/main/kotlin/com/netflix/spinnaker/config/LifecycleConfig.kt
emjburns: @lorin it might be possible, but to be honest I have not seen things quickly transition from baking -> deploying. i think it takes the clouddriver cache > 30s to update. so, I don't think it'll be a problem.

@gal-yardeni this is a config file - what kind of information would you be looking for? an explanation of how the config options will be used?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
```diff
@@ -0,0 +1,28 @@
...
+  val status: LifecycleEventStatus,
+  val text: String? = null,
+  val link: String? = null,
+  val timestamp: Instant? = null
```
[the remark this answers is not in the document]
emjburns: It is added at the database saving level, so that consumers don't have to send it. but when I read the object I want to get it back. Do you think this is the best way to model that?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
emjburns: Seems reasonable. I'll do it.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
emjburns: I think it's best practice to keep each thing to its own class file

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventRepository.kt
emjburns: I don't think so - because (1) I want to monitor tasks a bit differently (read one at a time instead of all of them) and (2) because each monitor knows what it's monitoring and how to respond to the changes. 

When I add a build monitor, it will monitor jenkins and emit events based on that. So it will have a different implementation.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
emjburns: they are making the same orca call, yes, but doing different things with the response. I don't think that this can be replaced with the existing orca monitoring

keel-sql/src/main/resources/db/changelog/20201016-create-lifecycle-events.yml
emjburns: discussed offline: not doing this, because serializing artifacts from their db records requires special handling. since we're not using any fields on the artifact, I don't think this makes sense.

keel-sql/src/main/resources/db/changelog/20201016-create-lifecycle-events.yml
emjburns: ohhh good point! will update.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleStep.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+data class LifecycleStep(
+  val scope: LifecycleEventScope,
+  val type: LifecycleEventType,
+  val id: String?,
+  val status: LifecycleEventStatus,
+  val text: String?,
+  val link: String?,
+  val startTime: Instant? = null,
+  val endTime: Instant? = null,
```
erikmunson, line 27: very sorry i didn't catch this earlier — is it strenuous to rename these to `startedAt` and `completedAt`? I don't have a personal preference between the two, but we use that naming convention across the rest of the API wherever we surface timestamp fields.

emjburns: 👍 no problem at all

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
emjburns: kept the ref because serializing artifacts is way harder than normal and we don't need any fields from it.

keel-sql/src/main/resources/db/changelog/20201016-create-lifecycle-events.yml
emjburns: done

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
emjburns: done

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
emjburns: added a todo to consolidate `OrcaTaskMonitorAgent` to be a `LifecycleMonitor` to take advantage of batch sizing.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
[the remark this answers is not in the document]
lorin: Ohhhh... That makes sense.

PR #1659, opened 2020-11-16
basic metrics for start / end of verifications

robfletcher:
Not sure if we really need timing metrics since performance of verifications is not down to Keel. I'll add the usual metrics for scheduled check lag when I add the scheduled check itself.


lorin (approved):
LGTM

luispollo (approved):
LGTM

PR #1660, opened 2020-11-16
Foreign keys

robfletcher:
Added foreign keys for tables that don't need schema changes in order to accommodate them.

Still to-do (because they need changes to the table structure):

- [x] ~artifact_versions -> delivery_artifact~
- [x] unhappy_veto -> resource
- [x] unhealthy -> resource

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
@@ -40,7 +40,6 @@ import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ARTIFACT_LAST_CHE
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ARTIFACT_VERSIONS
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.DELIVERY_ARTIFACT
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.DELIVERY_CONFIG
-import com.netflix.spinnaker.keel.persistence.metamodel.Tables.DELIVERY_CONFIG_ARTIFACT
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ENVIRONMENT
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ENVIRONMENT_ARTIFACT_PIN
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ENVIRONMENT_ARTIFACT_VERSIONS
@@ -173,16 +172,8 @@ class SqlArtifactRepository(
 
   override fun delete(artifact: DeliveryArtifact) {
     requireNotNull(artifact.deliveryConfigName) { "Error removing artifact - it has no delivery config!" }
-    val deliveryConfigId = select(DELIVERY_CONFIG.UID)
-      .from(DELIVERY_CONFIG)
-      .where(DELIVERY_CONFIG.NAME.eq(artifact.deliveryConfigName))
-
     jooq.transaction { config ->
       val txn = DSL.using(config)
-      txn.deleteFrom(DELIVERY_CONFIG_ARTIFACT)
-        .where(DELIVERY_CONFIG_ARTIFACT.DELIVERY_CONFIG_UID.eq(deliveryConfigId))
-        .and(DELIVERY_CONFIG_ARTIFACT.ARTIFACT_UID.eq(artifact.uid))
-        .execute()
       txn.deleteFrom(DELIVERY_ARTIFACT)
         .where(DELIVERY_ARTIFACT.UID.eq(artifact.uid))
         .execute()
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
@@ -29,14 +29,10 @@ import com.netflix.spinnaker.keel.persistence.metamodel.Tables.DELIVERY_CONFIG_A
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.DELIVERY_CONFIG_LAST_CHECKED
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ENVIRONMENT
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ENVIRONMENT_ARTIFACT_CONSTRAINT
-import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ENVIRONMENT_ARTIFACT_PIN
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL
-import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ENVIRONMENT_ARTIFACT_VERSIONS
-import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ENVIRONMENT_ARTIFACT_VETO
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.ENVIRONMENT_RESOURCE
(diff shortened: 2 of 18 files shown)


keel-sql/src/main/resources/db/changelog/20201116-foreign-keys.yml
```diff
@@ -0,0 +1,213 @@
...
+        referencedTableName: environment_artifact_constraint
+        referencedColumnNames: uid
+        referencesUniqueColumn: true
+        onDelete: CASCADE
+- changeSet:
+    id: foreign-keys-on-delivery_config_artifact
+    author: fletch
+    changes:
+    - sql:
+        sql: delete from delivery_config_artifact where not exists (select 1 from delivery_config where delivery_config_artifact.delivery_config_uid = delivery_config.uid)
```
luispollo, line 41: Ha! First time I noticed we even had this table! We should be able to get rid of it when we replace the delivery config name in `delivery_artifact` with the uid as an FK.

keel-sql/src/main/resources/db/changelog/20201116-foreign-keys.yml
```diff
@@ -0,0 +1,213 @@
...
+    author: fletch
+    changes:
+    - addForeignKeyConstraint:
+        baseTableName: environment
+        baseColumnNames: delivery_config_uid
+        constraintName: fk_environment_delivery_config
+        referencedTableName: delivery_config
+        referencedColumnNames: uid
+        referencesUniqueColumn: true
+        onDelete: CASCADE
```
luispollo, line 85: Just checking: the explicit deletes we have in the code at the moment should continue to work after this, right?

keel-sql/src/main/resources/db/changelog/20201116-foreign-keys.yml
```diff
@@ -0,0 +1,213 @@
...
+        referencedTableName: environment
+        referencedColumnNames: uid
+        referencesUniqueColumn: true
+        onDelete: CASCADE
+- changeSet:
+    id: foreign-keys-on-environment_artifact_versions
+    author: fletch
+    changes:
+    - sql:
+        sql: delete from environment_artifact_versions where not exists (select 1 from delivery_artifact where delivery_artifact.uid = environment_artifact_versions.artifact_uid)
```
luispollo, line 139: Should we look for orphans based on `environment_uid` as well?

luispollo (approved):
Thanks so much for taking the time to do this, @robfletcher! ❤️ 
Couple questions/comments, nothing blocking.

keel-sql/src/main/resources/db/changelog/20201116-foreign-keys.yml
robfletcher: Yes. I don't think this table is strictly necessary since it's not a many-to-many relationship

keel-sql/src/main/resources/db/changelog/20201116-foreign-keys.yml
robfletcher: I added these delete statements where the FK could not get added on a clone of production data so where there's no delete, there's no orphan data

keel-sql/src/main/resources/db/changelog/20201116-foreign-keys.yml
robfletcher: Yes, although they're probably unnecessary

lorin (approved):
All of the proposed FK additions look good to me. 

I tried to think through the risks, here's what I came up with:

1. Unexpected system behavior during the migration (e.g., migrations lock the database, halting resource checks, for longer than expected).
1. One of the migrations fails (e.g., because of an overlooked error in the migration code). It puts us into a state where the we have to fix the database schema by hand.
1. an overlooked error in one of the delete statements results in lost data.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -173,16 +172,8 @@ class SqlArtifactRepository(
...
-
     jooq.transaction { config ->
       val txn = DSL.using(config)
-      txn.deleteFrom(DELIVERY_CONFIG_ARTIFACT)
-        .where(DELIVERY_CONFIG_ARTIFACT.DELIVERY_CONFIG_UID.eq(deliveryConfigId))
-        .and(DELIVERY_CONFIG_ARTIFACT.ARTIFACT_UID.eq(artifact.uid))
-        .execute()
       txn.deleteFrom(DELIVERY_ARTIFACT)
         .where(DELIVERY_ARTIFACT.UID.eq(artifact.uid))
         .execute()
```
robfletcher, line 179: Only need to do one delete as the cascade will clean up `delivery_config_artifact`

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -110,68 +105,6 @@ class SqlDeliveryConfigRepository(
...
-                    row(ENVIRONMENT_RESOURCE.ENVIRONMENT_UID, ENVIRONMENT_RESOURCE.RESOURCE_UID)
-                      .`in`(
-                        it.map { resourceId -> row(envUid, resourceId) }
-                      )
-                  )
-                  .execute()
-              }
             txn
               .deleteFrom(ENVIRONMENT)
               .where(ENVIRONMENT.UID.eq(envUid))
```
robfletcher, line 110: All these things will cascade delete from `environment`

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -202,56 +135,10 @@ class SqlDeliveryConfigRepository(
...
-        txn.deleteFrom(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL)
-          .where(ENVIRONMENT_ARTIFACT_QUEUED_APPROVAL.ENVIRONMENT_UID.`in`(envUids))
-          .execute()
-        // remove artifact from delivery config
-        txn.deleteFrom(DELIVERY_CONFIG_ARTIFACT)
-          .where(
-            DELIVERY_CONFIG_ARTIFACT.DELIVERY_CONFIG_UID.eq(configUid),
-            DELIVERY_CONFIG_ARTIFACT.ARTIFACT_UID.`in`(artifactUids)
-          )
-          .execute()
```
robfletcher, line 254: all handled by cascade

keel-sql/src/main/resources/db/changelog/20201116-foreign-keys.yml
```diff
@@ -0,0 +1,290 @@
...
+    - addColumn:
+        tableName: unhealthy
+        columns:
+        - column:
+            name: resource_uid
+            type: char(26)
+    - modifySql:
+        dbms: mysql
+        append:
+          value: " first"
```
robfletcher, line 267: Note: this is how you get a new column in the first spot in a table in mySQL. It does not support Liquibase's `position` or `beforeColumn` properties, only `afterColumn`.

emjburns (commented):
All the code changes look good to me here. I don't have any experience with the msql or liqibase stuff you're doing so I can't comment on that

robfletcher:
> All the code changes look good to me here. I don't have any experience with the msql or liqibase stuff you're doing so I can't comment on that

Apart from the last 2 tables it should be straightforward enough. For those I'm just creating a new `resource_uid` column, populating it based on the old `resource_id`, setting up the foreign key, then dropping the old column.

luispollo:
@robfletcher Sorry for the late question, but judging by `DeliveryConfigRepositoryTests`, we don't have any test coverage for cascading of deletes when deleting "parent" records. Not trying to write tests for MySQL `ON DELETE CASCADE` here, but should we have some tests as a sanity check that we've configured all the tables correctly?

robfletcher:
@luispollo oh man, I could have sworn there were some, but I see what you mean. I'll add that. It's a totally valid test to ensure we don't end up with orphaned rows, and not difficult to write.

PR #1664, opened 2020-11-17
Fix missing primary keys following foreign key changes

robfletcher:
The `unhappy_veto` and `unhealthy` tables had a primary key of `resource_id`. I dropped that column and replaced it with `resource_uid` but didn't add a new primary key. Becuase `SqlUnhappyRepository` relies on `onDuplicateKeyIgnore` to avoid getting duplicate rows, this resulted in data corruption on that table.


lorin (approved):
LGTM

luispollo (approved):

PR #1665, opened 2020-11-18
fix(pinning): return the right diffing link when a version is pinned

gal-yardeni:
We recently added a `see changes` button to compare between artifact versions.

**The current solution**
For each artifact version and status, we checked which corresponding status we need to compare against.
For example:
current --> will be compared against the latest previous version
deploying/approved --> will be compared against the current version
pending --> will be compared against the current version.

**The problem** 
We didn't take into account that if we have a pinned version, we have to include that version when making the comparison in most cases. Since we _**have**_ to provide the right order when creating the SCM compare link, we needed to handle those differently.

**The solution** 
We first check to see if there is a pinned version for that artifact, in the environment. If so, we fetch that artifact version and compare against it first.
We also added the usage of the comparator for generating the SCM link, in order to make sure we are generating the link in the right order.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -420,7 +420,7 @@ class ApplicationService(
 
   private fun getUrl(version1: PublishedArtifact?, version2: PublishedArtifact?, artifact: DeliveryArtifact): String? {
     return if (version1 != null && version2 != null) {
-      return if (artifact.sortingStrategy.comparator.compare(version1, version2) < 0) {
+      return if (artifact.sortingStrategy.comparator.compare(version1, version2) > 0) { //these comparators sort in dec order, so condition is flipped
         //version2 is newer than version1
         generateCompareLink(version2.gitMetadata, version1.gitMetadata)
       } else {
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -295,13 +295,13 @@ class ApplicationService(
 
   // Pinning is a special case when is coming to creating a compare link between versions.
   // If there is a pinned version, which is not the same as the current version, we need
-  // to make sure we are creating the comparable link with referance to the pinned version.
+  // to make sure we are creating the comparable link with reference to the pinned version.
   private fun getPinnedArtifact(deliveryConfig: DeliveryConfig, environmentName: String, artifact: DeliveryArtifact, version: String): PublishedArtifact? {
     val pinnedVersion = repository.getPinnedVersion(deliveryConfig, environmentName, artifact.reference)
      return if (pinnedVersion != version)
       pinnedVersion?.let { getArtifactInstance(artifact, it) }
-    else {
-       null
+    else { //if pinnedVersion == current version, fetch the version which that the pinned version replaced
+       repository.getArtifactVersionByPromotionStatus(deliveryConfig, environmentName, artifact, PREVIOUS.name, pinnedVersion)
      }
   }
 
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -246,13 +246,13 @@ class ApplicationService(
 
     return when (status) {
       PENDING -> {
-        val olderGitMetadata = pinnedArtifact?: repository.getArtifactVersionByPromotionStatus(deliveryConfig, environmentName, artifact, CURRENT.name)
+        val olderArtifactVersion = pinnedArtifact?: repository.getArtifactVersionByPromotionStatus(deliveryConfig, environmentName, artifact, CURRENT.name)
         ArtifactSummaryInEnvironment(
           environment = environmentName,
           version = version,
           state = status.name.toLowerCase(),
           // comparing PENDING (version in question, new code) vs. CURRENT (old code)
-          compareLink = getUrl(currentArtifact, olderGitMetadata, artifact)
+          compareLink = getUrl(currentArtifact, olderArtifactVersion, artifact)
         )
       }
       SKIPPED -> {
@@ -268,25 +268,25 @@ class ApplicationService(
       }
 
       DEPLOYING, APPROVED -> {
-        val olderGitMetadata = pinnedArtifact?: repository.getArtifactVersionByPromotionStatus(deliveryConfig, environmentName, artifact, CURRENT.name)
+        val olderArtifactVersion = pinnedArtifact?: repository.getArtifactVersionByPromotionStatus(deliveryConfig, environmentName, artifact, CURRENT.name)
         potentialSummary?.copy(
           // comparing DEPLOYING/APPROVED (version in question, new code) vs. CURRENT (old code)
-          compareLink = getUrl(currentArtifact, olderGitMetadata, artifact)
+          compareLink = getUrl(currentArtifact, olderArtifactVersion, artifact)
         )
       }
       PREVIOUS -> {
-        val newerGitMetadata = potentialSummary?.replacedBy?.let { getArtifactInstance(artifact, it) }
+        val newerArtifactVersion = potentialSummary?.replacedBy?.let { getArtifactInstance(artifact, it) }
         potentialSummary?.copy(
           //comparing PREVIOUS (version in question, old code) vs. the version which replaced it (new code)
           //pinned artifact should not be consider here, as we know exactly which version replace the current one
-          compareLink = getUrl(currentArtifact, newerGitMetadata, artifact)
+          compareLink = getUrl(currentArtifact, newerArtifactVersion, artifact)
         )
       }
       CURRENT -> {
-        val olderGitMetadata = pinnedArtifact?: repository.getArtifactVersionByPromotionStatus(deliveryConfig, environmentName, artifact, PREVIOUS.name)
+        val olderArtifactVersion = pinnedArtifact?: repository.getArtifactVersionByPromotionStatus(deliveryConfig, environmentName, artifact, PREVIOUS.name)
         potentialSummary?.copy(
           // comparing CURRENT (version in question, new code) vs. PREVIOUS (old code)
-          compareLink = getUrl(currentArtifact, olderGitMetadata, artifact)
+          compareLink = getUrl(currentArtifact, olderArtifactVersion, artifact)
         )
       }
       else -> potentialSummary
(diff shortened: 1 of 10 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -242,16 +242,17 @@ class ApplicationService(
         version = version
       )
 
+    val pinnedArtifact = getPinnedArtifact(deliveryConfig, environmentName, artifact, version)
+
     return when (status) {
       PENDING -> {
-        val olderGitMetadata = repository.getGitMetadataByPromotionStatus(deliveryConfig, environmentName, artifact, CURRENT.name)
-
+        val olderGitMetadata = pinnedArtifact?: repository.getArtifactVersionByPromotionStatus(deliveryConfig, environmentName, artifact, CURRENT.name)
```
luispollo, line 249: So a pinned version doesn't get the status updated to `CURRENT`? That seems counter-intuitive...

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -265,31 +266,45 @@ class ApplicationService(
...
-      DEPLOYING -> {
-        val olderGitMetadata = repository.getGitMetadataByPromotionStatus(deliveryConfig, environmentName, artifact, CURRENT.name)
+
+      DEPLOYING, APPROVED -> {
+        val olderGitMetadata = pinnedArtifact?: repository.getArtifactVersionByPromotionStatus(deliveryConfig, environmentName, artifact, CURRENT.name)
         potentialSummary?.copy(
-          // comparing DEPLOYING (version in question, new code) vs. CURRENT (old code)
-          compareLink = generateDiffLink(baseScmUrl, artifactGitMetadata, olderGitMetadata)
+          // comparing DEPLOYING/APPROVED (version in question, new code) vs. CURRENT (old code)
+          compareLink = getUrl(currentArtifact, olderGitMetadata, artifact)
```
luispollo, line 274: IMHO, the original name of the function was more self-explanatory than the new one.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -411,4 +417,29 @@ class ApplicationService(
         throw UnsupportedScmType(message = "Stash is currently the only supported SCM type")
     }
   }
+
+  private fun getUrl(version1: PublishedArtifact?, version2: PublishedArtifact?, artifact: DeliveryArtifact): String? {
+    return if (version1 != null && version2 != null) {
+      return if (artifact.sortingStrategy.comparator.compare(version1, version2) < 0) {
```
luispollo, line 423: There are a couple of utility functions you could use here in place of this code to make it more readable. Sorry I didn't think of that earlier: https://github.com/spinnaker/keel/blob/e6f6e6c7824a68164f32887d61d7800c2310c647/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt#L267-L279

It would be nice to `ArtifactUtils`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: what do you mean? I didn't change the logic of the status updates, only for which version to compare against.

luispollo: If I'm reading this correctly, you'll give precedence to the pinned version instead of the version with a `CURRENT` status. I would expect that, whenever we deploy a version (regardless of whether it was as a result of a normal promotion or pinning), that version would become `CURRENT`. If that was the case, then the change in this specific line would be unnecessary because pinned == current.

gal-yardeni: I see what you are saying. The current logic gives priority for pinned versions, and if we find one, we won't call to get the `current` version as well. I think we could have some edge cases where pinned != current, so I kinda think it's safer to keep it that way 🤷

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/artifacts.kt
```diff
@@ -25,7 +25,7 @@ class DummyArtifact(
 }
 
 object DummySortingStrategy : SortingStrategy {
-  override val comparator: Comparator<PublishedArtifact> = compareBy { it.version }
+  override val comparator: Comparator<PublishedArtifact> = compareByDescending { it.version }
```
gal-yardeni, line 28: This was the pain in all of this comparator mess!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: I started doing it but it was more harm than good. I ended up using only one line, and it didn't make sense to me to export it via a bunch of files and classes..

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -242,16 +242,17 @@ class ApplicationService(
         version = version
       )
 
+    val pinnedArtifact = getPinnedArtifact(deliveryConfig, environmentName, artifact, version)
+
     return when (status) {
       PENDING -> {
-        val olderGitMetadata = repository.getGitMetadataByPromotionStatus(deliveryConfig, environmentName, artifact, CURRENT.name)
-
```
emjburns, line 248: is this still git metadata, or is it now an artifact instance?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -265,31 +266,45 @@ class ApplicationService(
...
       else -> potentialSummary
     }
   }
 
+  // Pinning is a special case when is coming to creating a compare link between versions.
+  // If there is a pinned version, which is not the same as the current version, we need
+  // to make sure we are creating the comparable link with reference to the pinned version.
+  private fun getPinnedArtifact(deliveryConfig: DeliveryConfig, environmentName: String, artifact: DeliveryArtifact, version: String): PublishedArtifact? {
+    val pinnedVersion = repository.getPinnedVersion(deliveryConfig, environmentName, artifact.reference)
+     return if (pinnedVersion != version)
```
emjburns, line 301: I'd check for null in this conditional so you don't have to do the `?` below. I think it's a bit more readable

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -1154,41 +1177,60 @@ class SqlArtifactRepository(
...
 
     return sqlRetry.withRetry(READ) {
-      val version = jooq
+      val fetchedVersion = jooq
         .select(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
         .from(ENVIRONMENT_ARTIFACT_VERSIONS)
         .where(ENVIRONMENT_ARTIFACT_VERSIONS.ENVIRONMENT_UID.eq(deliveryConfig.getUidFor(environmentName)))
         .and(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID.eq(artifact.uid))
         .and(ENVIRONMENT_ARTIFACT_VERSIONS.PROMOTION_STATUS.eq(promotionStatus))
+
```
emjburns, line 1197: nit: remove newlines before and after since it's part of the jooq statement

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -1154,41 +1177,60 @@ class SqlArtifactRepository(
...
+        //special case for pinning
+        .apply { if (version != null) and(ENVIRONMENT_ARTIFACT_VERSIONS.REPLACED_BY.eq(version)) }
+
         .orderBy(ENVIRONMENT_ARTIFACT_VERSIONS.DEPLOYED_AT.desc())
         .fetch(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_VERSION)
         .firstOrNull()
 
-      if (version != null) {
+      if (fetchedVersion != null) {
```
emjburns, line 1205: Do you think it's worth it to extract this into a function? `getPublishishedArtifact(artifact: DeliveryArtifact, version: String): PublishedArtifact?`?

That way it encapsulates the logic a bit nicer.

keel-test/src/main/kotlin/com/netflix/spinnaker/keel/test/artifacts.kt
emjburns: :( :( :( :( :( im so glad you figured this out

emjburns (approved):
I have a couple of formatting/style comments, but overall this looks good

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: you're right -- renamed

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: done

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: done

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
gal-yardeni: nice! by doing so, I've discovered we already have a function like this so I removed it completely!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
gal-yardeni: I ended up not doing it as it got this method to be much longer (if `pinnedVersion` is null, I need to return null. Here its happening automatically, and if I were to add your suggestion, I needed to add a special case for it and it would have gotten less readable, in my opinion).

PR #1666, opened 2020-11-19
prevent unintentionally extending cache TTL

robfletcher:
An issue arose recently where a new base AMI caused a new version of an image to get baked. Because on some instances the older image was cached Keel could arbitrarily pick either of the two available images (the one with the older base AMI, and the one with the newer one). This caused the desired state to flip-flop depending on which Keel instance checked it.

Exacerbating this is the fact that the cache is aggressively primed with any other regions found in the image. This meant that although the cache has a TTL of 1 hour it kept getting refreshed with the same value, extending the TTL.

This PR attempts to ensure we only prime the cache with additional regions if it currently doesn't contain a mapping for the app version / account / region combination, or the cached image is older than the one just retrieved from CloudDriver.

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
@@ -17,6 +17,7 @@
  */
 package com.netflix.spinnaker.keel.clouddriver
 
+import com.github.benmanes.caffeine.cache.AsyncCache
 import com.netflix.frigga.ami.AppVersion
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.caffeine.CacheFactory
@@ -37,7 +38,6 @@ import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.future.await
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
-import java.util.concurrent.CompletableFuture.completedFuture
 
 class ImageService(
   private val cloudDriverService: CloudDriverService,
@@ -128,16 +128,30 @@ class ImageService(
     appVersion: AppVersion,
     account: String,
     region: String
-  ): NamedImage? =
-    namedImageCache
-      .get(NamedImageCacheKey(appVersion, account, region))
-      .await()
+  ): NamedImage? {
+    val key = NamedImageCacheKey(appVersion, account, region)
+    return namedImageCache.get(key).await()
       // prime the cache if the image is also in other regions
       ?.also { image ->
         (image.amis.keys - region).forEach { otherRegion ->
-          namedImageCache.put(NamedImageCacheKey(appVersion, account, otherRegion), completedFuture(image))
+          namedImageCache.putIfMissingOrOlderThan(key.copy(region = otherRegion), image)
         }
       }
+  }
+
+  private fun AsyncCache<NamedImageCacheKey, NamedImage>.putIfMissingOrOlderThan(
+    key: NamedImageCacheKey,
+    image: NamedImage
+  ) {
+    synchronous().apply {
+      val updateCache = getIfPresent(key)
+        ?.let { it.creationDate < image.creationDate }
+        ?: true
+      if (updateCache) {
+        put(key, image)
+      }
+    }
+  }
 
   suspend fun findBaseAmiVersion(baseImageName: String): String {
     return cloudDriverService.namedImages(DEFAULT_SERVICE_ACCOUNT, baseImageName, "test")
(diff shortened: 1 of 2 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -128,16 +128,30 @@ class ImageService(
...
-      .get(NamedImageCacheKey(appVersion, account, region))
-      .await()
+  ): NamedImage? {
+    val key = NamedImageCacheKey(appVersion, account, region)
+    return namedImageCache.get(key).await()
       // prime the cache if the image is also in other regions
       ?.also { image ->
         (image.amis.keys - region).forEach { otherRegion ->
-          namedImageCache.put(NamedImageCacheKey(appVersion, account, otherRegion), completedFuture(image))
+          namedImageCache.putIfMissingOrOlderThan(key.copy(region = otherRegion), image)
```
robfletcher, line 137: Instead of _always_ adding the image to the cache for any other regions, only do so if it's not cached currently, or the current cached image is older than this one

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -128,16 +128,30 @@ class ImageService(
...
+  ) {
+    synchronous().apply {
+      val updateCache = getIfPresent(key)
+        ?.let { it.creationDate < image.creationDate }
+        ?: true
+      if (updateCache) {
+        put(key, image)
+      }
+    }
+  }
```
robfletcher, line 154: This is the implementation of that logic. It operates on a `synchronous` view of the cache as it's using `getIfPresent` to avoid calling CloudDriver so there's no need to handle the result asynchronously.

keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ServerGroup.kt
```diff
@@ -69,6 +69,7 @@ data class ServerGroup(
   }
 
   data class ActiveServerGroupImage(
+    @get:ExcludedFromDiff
     val imageId: String,
     val appVersion: String?,
     val baseImageVersion: String?,
```
robfletcher, line 75: There seems to be no compelling reason to include `imageId` in the diff we generate. It's meaningless to users and the information we need to decide whether to redeploy is contained in `appVersion` and `baseImageVersion`. If either of those changes, we'll re-deploy. The `imageId` field is necessary when we launch the Orca task but not in the diff.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -152,35 +166,6 @@ class ImageService(
...
-
-    if (appversion.size != 3) {
-      return false
-    }
-
-    if (appversion[1] != buildName || appversion[2] != buildNumber) {
-      return false
-    }
-    return true
-  }
```
robfletcher, line 182: Both these methods were unused

emjburns (approved):
Seems reasonable to me

lorin (approved):
## My understanding

The ImageResolver uses an in-memory cache to cache the AMI (image ids) associated with the most recent version of a baked image for an artifact version. Because of AWS's architecture, the same logical version of an (artifact version, base os) will have a different AMI in region, so we need a separate cache entry for each region that an artifact will be deployed to as specified in the delivery config. (Presumably, the same happens for account).

This means that an AMI is uniquely identified by a (version, account, region) triple. However, the value stored in the cache is a `NamedImage` object which contains the AMIs for all of the accounts/regions associated with the image name.

Because's the cache is in-memory, the state of the cache can vary by instance. For example, after a baseos update, one keel instance may have a stale version of the latest baseos AMI, and another instance may have the newer version of the baseos update.

If two different keel instances have different views of the latest version of baseos, then we can get flapping behavior where the ImageResolvers on the different instances keep computing different values for the `desired` state of the image id of the cluster.

The namedImageCache of the ImageService has a TTL of 60 minutes, so in theory it should settle down after at most an hour. However, there was logic in the ImageService to handle the case where the image was baked in multiple regions. In that case, the retrieved value was re-added to the cache for the other (version, account, region) triples. By re-adding them to the cache, this had the effect of resetting the TTLs for those other regions. If the image resolver queried the cache for two different regions often enough, the cache would never expire.

The proposed fix is to only re-add data to the cache in cases where:

* it's not there yet
* the cached data is stale

Here's the decision table. Note that we only care about cases where there's multi-region data, since it only updates the cache 
for regions other than the one that was just retrieves.

| Multi-reigon data | (version, account, region) entry exists in cache | cached image is older than river data | **prime cache?** |
| ----------------- | ------------------------------------------------ | ------------------------------------- | ---------------- |
| T                 | T                                                | T                                     | T                |
| T                 | T                                                | F                                     | F                |
| T                 | F                                                | \*                                    | T                |
| F                 | \*                                               | \*                                    | F                |


## Feedback

Looks good to me!

While this fix doesn't prevent the problem of inconsistent cached image data across keel nodes, it's definitely an improvement over the current situation.

Issue #1668, opened 2020-11-19
Building is failing because of new environment variable


saimadineni:
./gradlew build causes this issue
: Unexpected exception during bean creation; nested exception is java.lang.IllegalArgumentException: Could not resolve placeholder 'spinnaker.baseUrl' in value "${spinnaker.baseUrl}
This PR https://github.com/spinnaker/keel/pull/1669 to solve the issue.  
a Is this supposed to point to a particular service gate/orca/clouddriver ? With trial and error I was able to figure out that this could possibly be either deck or gate. What's the intention behind naming it as spinnaker.baseURL ?
Edit: If it is deck, why can't we just call it the deck/base url itself instead of calling it the spinnaker baseUrl
Also, I believe there needs to be a section for spinnaker base url in the keel.yml if we need to use spinnkerr base url as the service was failing to run as it couldn't resolve spinnaker base url
spinnaker:
  baseUrl: ${services.deck.baseUrl}

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

PR #1669, opened 2020-11-19
Fixing the environment variable

saimadineni:
This fix will help with resolving the issues related to failures while running keel locally
Issue: https://github.com/spinnaker/keel/issues/1668
A new environment variable called  "${spinnaker.baseUrl} was introduced, which doesn't seem to be necessary. It could have just been gate.baseUrl
 and all the changes related to spinnaker.baseURl? Is this supposed to point to a particular service gate/orca/clouddriver ? With trial and error I was able to figure out that this could possibly be gate. What's the intention behind naming it as spinnaker.baseURL ?
Edit: If it is deck, why can't we just call it the deck url itself instead of calling it the spinnaker baseUrl
Also, I believe there needs to be a section for spinnaker base url in the keel.yml if we need to use spinnkaer base url as the service was failing to run as it couldn't resolve spinnaker base url
spinnaker:
  baseUrl: ${services.deck.baseUrl}
Updated the comment @robfletcher

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
@@ -7,11 +7,18 @@ import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import java.time.Instant
 
 interface VerificationRepository {
+  /**
+   * @return the current state of [verification] as run against [context], or `null` if it has not
+   * yet been run.
+   */
   fun getState(
     context: VerificationContext,
     verification: Verification
-  ) : VerificationState?
+  ): VerificationState?
 
+  /**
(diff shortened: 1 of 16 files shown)


robfletcher:
What issues are you seeing?

saimadineni:
Providing more info:
Spinnaker.baseURL environment variable was introduced in this change
https://github.com/spinnaker/keel/blob/master/keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/ManualJudgementNotifier.kt#L30

emjburns:
So this variable represents the deck URL (or whatever URL users go to to interact with the UI). We use it for creating contextual links for notifications to help users get directed to the right place. Since it's only used for creating links, we could definitely give it a default garbage value so it doesn't impact local running. Let me put out a pr for that. Thanks for reporting and sorry for the delay in reviewing!

emjburns:
https://github.com/spinnaker/keel/pull/1685

PR #1671, opened 2020-11-20
test for cascading deletes from delivery config

robfletcher:
Luis pointed out in a comment on #1660 that there wasn't a test to confirm cascade delete behavior. I had thought there was but when I went looking I couldn't find it. This PR adds one.


gal-yardeni (approved):
nice

PR #1672, opened 2020-11-20
foreign keys for verification_state table

robfletcher:
I hadn't merged the PR for adding the `verification_state` table before opening #1660 so I missed these


lorin (approved):
LGTM

PR #1673, opened 2020-11-20
support clusters that depend on NLB target groups

robfletcher:
Even though we can't manage NLBs we can detect their target groups and not reject clusters that depend on them.


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/AmazonLoadBalancer.kt
```diff
@@ -24,6 +25,4 @@ interface AmazonLoadBalancer {
   val vpcId: String
   val subnets: Set<String>
   val scheme: String?
-  val idleTimeout: Int
-  val securityGroups: Set<String>
```
robfletcher, line 28: Had to move these fields off the interface as NLBs don't have them. Didn't require any other changes.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/NetworkLoadBalancerModel.kt
```diff
@@ -0,0 +1,18 @@
...
+  override val vpcId: String,
+  override val subnets: Set<String>,
+  override val scheme: String?,
+  override val availabilityZones: Set<String>,
+  val targetGroups: List<TargetGroup>
+) : AmazonLoadBalancer {
+  data class TargetGroup(
+    val targetGroupName: String
+  )
+}
```
robfletcher, line 18: Basic NLB model. We just want to check that the thing exists and has the target group we want

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
```diff
@@ -73,7 +74,13 @@ class RequiredLoadBalancerVeto(
     }
     spec.targetGroupsByRegion.forEach { (name, account, regions) ->
       val missingRegions = regions.filter { region ->
-        none { lb -> lb is ApplicationLoadBalancerModel && lb.account == account && lb.region == region && lb.targetGroups.any { it.targetGroupName == name } }
+        none { lb ->
+          when (lb) {
+            is ApplicationLoadBalancerModel -> lb.account == account && lb.region == region && lb.targetGroups.any { it.targetGroupName == name }
+            is NetworkLoadBalancerModel -> lb.account == account && lb.region == region && lb.targetGroups.any { it.targetGroupName == name }
+            else -> false
+          }
```
robfletcher, line 82: Once again I wish Kotlin had intersection types

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/NetworkLoadBalancerModel.kt
gal-yardeni: It's probably better to add here a comment that we don't support it yet? So we / anyone else looking at the code will be confused 🤷

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/vetos/RequiredLoadBalancerVeto.kt
lorin: if you were really bothered by the duplication here, you could derive an interface from `AmazonLoadBalancer` that has target groups on it (e.g., `TargetGroupedAmazonLoadBalancer`).

lorin (approved):
## My understanding

### Context

For reasons internal to Netflix, keel isn't able to automatically manage NLBs, so keel has never supported them.

### Problem

An internal user exported resources associated with an app. One of those resources was an EC2 cluster fronted by an NLB. The EC2 cluster got stuck in `MISSING_DEPENDENCY` state.

### Proposed Solution

Make keel aware of NLB resources by defining a `NetworkLoadBalancerModel` class. This class is deserialized from the `CloudDriverService.loadBalancersForApplication` RPC call for objects that have `name = "network"`

With this solution, even though keel can't actually manage NLB resources, keel` won't throw an error if an NLB is associated with a managed cluster.

## Feedback

The change looks good overall, but I don't understand the original failure mode well enough to judge whether this will fix the problem.  How was keel behaving when an EC2 cluster was fronted by NLB, and we did not support NLB model objects?

PR #1674, opened 2020-11-24
feat(compareLink): generate a comparable link in MJ slack notification

gal-yardeni:
Adding a "see changes" link whenever we send a manual judgment notification in slack.
Preview:
Screen Shot 2020-11-24 at 2 00 42 PM

Looking for feedback mainly around:
- naming (I called it "show full diff" and then the full diff, for now)
- order (where should we show it within the notification?)

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
@@ -25,7 +25,7 @@ fun generateCompareLink(scmInfo: ScmInfo, version1: PublishedArtifact?, version2
 //Generating a SCM compare link between source (new version) and target (old version) versions (the order matter!)
 private fun generateCompareLink(scmInfo: ScmInfo, newerGitMetadata: GitMetadata?, olderGitMetadata: GitMetadata?): String? {
   val baseScmUrl = newerGitMetadata?.commitInfo?.link?.let { getScmBaseLink(scmInfo, it) }
-  return if (baseScmUrl != null && olderGitMetadata != null) {
+  return if (baseScmUrl != null && olderGitMetadata != null && !(olderGitMetadata.commitInfo?.sha.isNullOrEmpty())) {
     "$baseScmUrl/projects/${newerGitMetadata.project}/repos/${newerGitMetadata.repo?.name}/compare/commits?" +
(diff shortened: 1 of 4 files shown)


erikmunson:
Exciting! Only feedback would be that the link itself is quite large/long and makes the notification even noisier than before as a result. How difficult/time consuming is it to add another button to the message next to 'approve' and 'reject'? Wondering if maybe a "See changes" button at the bottom of the message might be more user friendly since it's ideally such a common action?

gal-yardeni:
@erikmunson interesting idea! let me explore it!

emjburns:
I agree with Erik, I love the idea of a button! 

Also, I'm so excited that this functionality is happening!

lorin:
Another option (if a button ends up being too much work) is to replcae the explicit link with a short bit of hyperlinked text (something like "click for diff").

lorin (commented):
## My understanding

### Problem

When an MD user is informed via Slack that a new artifact version is waiting for approval in an environment, the user has to context-switch to the environments view to identify what code has changed in the new artifact version compared to what's currently deployed.

### Proposed solution

In the Slack notification, add a Stash link to the diff between the commit associated with the new artifact version and the commit associated with the currently deployed artifact version, in the relevant environment.

## Feedback

Overall: looks good to me.  I do agree with the comment that @erikmunson made that it would be better if we could reduce the size of the link (e.g., with a button).

### Naming: Consistent terminology

I think we should use the same terminology in Slack and in the environments view. Currently:

* Environment views uses "See changes"
* Slack notification uses "see full diff"

I don't have a strong opinion about which one we use, though.

### Ordering

I think the proposed location (Under the Message) is a good place.

### Possible behavior difference in Slack vs. enviroment view

The link from the Slack notification is not guaranteed to show the same diff as the link from the environments view:

* the environments view always diffs against the `CURRENT` artifact version in the environment
* the Slack notification diffs against what was the `CURRENT` artifaction version in the environment **when the Slack message was generated**

At the time that the Slack event appears, the two diffs are the same. But, if someone pins another version to the environment,
the two views will show different diffs.

Note: I'm not suggesting a behavior change, just noting a potential source of confusion for users.

lorin (approved):

gal-yardeni:
@lorin thanks for your feedback! I changed it to "See changes" text with a hyperlink to stash, when available.

You are making a valid point about pinning - my hope we can educate our customers to understand what they are comparing against. Will ask for feedback in the next few weeks.

emjburns (approved):

PR #1675, opened 2020-11-24
fix(view): don't summarize artifact when unused

emjburns:
This PR fixes the issue where we create summaries for environments where an artifact will never go. that is distracting and confusing in the UI. it does not spark joy.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
@@ -1,5 +1,7 @@
 package com.netflix.spinnaker.keel.api.artifacts
 
+import com.netflix.spinnaker.keel.api.ArtifactReferenceProvider
+import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.schema.Discriminator
 import java.time.Instant
 
@@ -92,6 +94,15 @@ abstract class DeliveryArtifact {
         "createdAt" to createdAt
       )
     ).normalized()
+
+  /**
+   * @return `true` if this artifact is used by any resource in [environment], `false` otherwise.
+   */
+  fun isUsedIn(environment: Environment) =
+    environment
+      .resources
+      .map { (it.spec as? ArtifactReferenceProvider)?.artifactReference }
+      .contains(reference)
 
   fun toLifecycleRef() = "$deliveryConfigName:$reference"
 
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -141,6 +141,15 @@ class ApplicationService(
     )
   }
 
+  fun getSummariesAllEntities(application: String): Map<String, Any> {
+    val summaries: MutableMap<String, Any> = mutableMapOf()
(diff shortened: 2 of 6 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -93,6 +95,15 @@ abstract class DeliveryArtifact {
       )
     ).normalized()
```
emjburns, line 97: Pulled this function from the `EnvironmentPromotionChecker`, it's useful.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -192,29 +201,38 @@ class ApplicationService(
...
-                }
+              if ( artifact.isUsedIn(environment)) { // only add a summary if the artifact is used in the environment
+                buildArtifactSummaryInEnvironment(deliveryConfig, environment.name, artifact, artifactVersion.version, status)
+                  ?.also {
+                    artifactSummariesInEnvironments.add(
+                      it.addStatefulConstraintSummaries(deliveryConfig, environment, artifactVersion.version)
+                        .addStatelessConstraintSummaries(deliveryConfig, environment, artifactVersion.version, artifact)
+                    )
+                  }
+              }
```
emjburns, line 235: ^ is the big fix, specifically line 227.

emjburns: Now, it's true that we're still calculating the summary in the SQL. We're just filtering it out here, and this is the view that the UI uses. We can surely optimize this later if needed but focusing on stopping the really annoying user facing behavior for now

erikmunson:
beautiful

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
```diff
@@ -75,13 +75,17 @@ class ApplicationController(
...
+        results.putAll(applicationService.getSummariesAllEntities(application))
+      } else {
+        entities.forEach { entity ->
+          results[entity] = when (entity) {
+            "resources" -> applicationService.getResourceSummariesFor(application)
+            "environments" -> applicationService.getEnvironmentSummariesFor(application)
+            "artifacts" -> applicationService.getArtifactSummariesFor(application,
+              maxArtifactVersions ?: DEFAULT_MAX_ARTIFACT_VERSIONS)
+            else -> throw InvalidRequestException("Unknown entity type: $entity")
+          }
```
emjburns, line 88: the environments summary was pulled twice - once for `environments`, once for `artifacts`. Adding a function that takes care of the way deck is calling us so that we don't have to do the expensive database calculation twice. Leaving the old function in here for one-off calls, or when deck just calls with `resources`

erikmunson: 🥳 no idea how much you expect here in terms tangible impact on latency, but if you think it'll be material it'd be cool if we could watch a response latency graph of some kind before/after deployment

robfletcher (approved):

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ApplicationController.kt
emjburns: ooooo. let me think.

PR #1678, opened 2020-11-30
prefer constructor autowiring to use of lateinit var in Spring tests

robfletcher:
Unfortunately this doesn't work with uses of `@MockkBean` but generally I think this is preferable.


lorin (approved):
LGTM

PR #1679, opened 2020-12-01
feat(lifecycle): build events

emjburns:
Adding the ability to monitor builds and have them as part of the pre-deployment steps.

The funny thing here is that the artifact is built during the build, and published before the build finishes. So, we won't "see" this build until the artifact is published. 

Also, linking to the CI view is complicated. We have to do some digging through the app config to figure out if the ci view is wired in and working.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
@@ -1,8 +1,7 @@
 package com.netflix.spinnaker.keel.api.artifacts
 
 import java.time.Instant
-import java.time.LocalDateTime
-import java.time.ZoneOffset.UTC
+import java.time.ZonedDateTime
 
 /**
  * The build metadata of an artifact.
@@ -18,21 +17,14 @@ data class BuildMetadata(
   val completedTs: String? = null,
   val status: String? = null
 ) {
-  fun getStartedAtInstant(): Instant? =
-    startedAt
-      ?.let {
-        LocalDateTime.parse(it.removeSuffix("Z"))
-          .atZone(UTC)
-          .toInstant()
-      }
+  val startedAtInstant: Instant?
+    get() = startedAt?.let { ZonedDateTime.parse(it).toInstant() }
 
-  fun getCompletedAtInstant(): Instant? =
-    completedAt
-      ?.let {
-        LocalDateTime.parse(it.removeSuffix("Z"))
-          .atZone(UTC)
-          .toInstant()
-      }
+  val completedAtInstant: Instant?
+    get() = completedAt?.let { ZonedDateTime.parse(it).toInstant() }
+
+  fun isComplete(): Boolean =
+    status in listOf("SUCCESS", "FAILED", "ABORTED", "UNSTABLE")
 }
 
 data class Job(
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
@@ -1,5 +1,9 @@
 package com.netflix.spinnaker.keel.api.artifacts
 
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneOffset.UTC
+
 /**
  * The build metadata of an artifact.
  */
@@ -13,7 +17,23 @@ data class BuildMetadata(
   val startedTs: String? = null,
   val completedTs: String? = null,
   val status: String? = null
-)
+) {
+  fun getStartedAtInstant(): Instant? =
+    startedAt
+      ?.let {
+        LocalDateTime.parse(it.removeSuffix("Z"))
+          .atZone(UTC)
+          .toInstant()
+      }
+
+  fun getCompletedAtInstant(): Instant? =
+    completedAt
+      ?.let {
+        LocalDateTime.parse(it.removeSuffix("Z"))
+          .atZone(UTC)
+          .toInstant()
+      }
+}
 
 data class Job(
   val link: String? = null,
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
@@ -89,6 +89,7 @@ class ArtifactListener(
         val data: MutableMap<String,String?> = mutableMapOf()
         data["buildNumber"] = artifact.metadata["buildNumber"]?.toString()
         data["commitId"] = artifact.metadata["commitId"]?.toString()
+        data["fallbackLink"] = buildMetadata.job?.link
 
         repository
           .getAllArtifacts(artifact.artifactType, artifact.name)
@@ -106,7 +107,8 @@ class ArtifactListener(
                 status = NOT_STARTED,
                 text = "Monitoring build for ${artifact.version}",
                 link = buildMetadata.uid,
-                data = data
+                data = data,
+                timestamp = buildMetadata.getStartedAtInstant()
               ))
             }
           }
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
@@ -62,7 +62,7 @@ class ArtifactListener(
             log.info("Registering version {} (status={}) of {} artifact {}",
               artifact.version, artifact.status, artifact.type, artifact.name)
 
-            enrichAndStore(artifact.normalized(), artifactSupplier)
+            enrichAndStore(artifact, artifactSupplier)
               .also { wasAdded ->
                 if (wasAdded) {
                   publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.artifactType))
@@ -80,16 +80,20 @@ class ArtifactListener(
   /**
    * Finds the delivery configs that are using an artifact,
    * and publishes a build lifecycle event for them.
+   *
+   * todo eb: If the status is complete, publish a NOT_STARTED and a completed event
+   *  maybe add a "monitor me" flag? that seems better
    */
   fun createBuildLifecycleEvent(artifact: PublishedArtifact) {
     log.debug("Publishing build lifecycle event for artifact $artifact")
     artifact.buildMetadata
       ?.let { buildMetadata ->
 
-        val data: MutableMap<String,String?> = mutableMapOf()
-        data["buildNumber"] = artifact.metadata["buildNumber"]?.toString()
-        data["commitId"] = artifact.metadata["commitId"]?.toString()
-        data["fallbackLink"] = buildMetadata.job?.link
+        val data = mutableMapOf(
+          "buildNumber" to artifact.metadata["buildNumber"]?.toString(),
+          "commitId" to artifact.metadata["commitId"]?.toString(),
+          "fallbackLink" to buildMetadata.job?.link
+        )
 
         repository
           .getAllArtifacts(artifact.artifactType, artifact.name)
@@ -108,7 +112,7 @@ class ArtifactListener(
                 text = "Monitoring build for ${artifact.version}",
                 link = buildMetadata.uid,
                 data = data,
-                timestamp = buildMetadata.getStartedAtInstant()
+                timestamp = buildMetadata.startedAtInstant
               ))
             }
           }
@@ -132,7 +136,7 @@ class ArtifactListener(
 
       if (latestArtifact != null) {
         log.debug("Storing latest version {} (status={}) for registered artifact {}", latestArtifact.version, latestArtifact.status, artifact)
-        enrichAndStore(latestArtifact.normalized(), artifactSupplier)
+        enrichAndStore(latestArtifact, artifactSupplier)
       } else {
         log.warn("No artifact versions found for ${artifact.type}:${artifact.name}")
       }
@@ -176,7 +180,7 @@ class ArtifactListener(
 
               if (hasNew) {
                 log.debug("$artifact has a missing version ${latestArtifact.version}, persisting.")
-                enrichAndStore(latestArtifact.normalized(), artifactSupplier)
+                enrichAndStore(latestArtifact, artifactSupplier)
               } else {
                 log.debug("No new versions to persist for $artifact")
               }
@@ -187,8 +191,14 @@ class ArtifactListener(
     }
   }
 
+  /**
+   * Normalizes an artifact by calling [PublishedArtifact.normalized],
+   * enriches it by adding metadata,
+   * creates the appropriate build lifecycle event,
+   * and stores in the database
+   */
   private fun enrichAndStore(artifact: PublishedArtifact, supplier: ArtifactSupplier<*,*>): Boolean {
-    val enrichedArtifact = supplier.addMetadata(artifact)
+    val enrichedArtifact = supplier.addMetadata(artifact.normalized())
     createBuildLifecycleEvent(enrichedArtifact)
     return repository.storeArtifactVersion(enrichedArtifact)
   }
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
@@ -12,6 +12,10 @@ import com.netflix.spinnaker.keel.api.events.ArtifactSyncEvent
 import com.netflix.spinnaker.keel.api.plugins.ArtifactSupplier
 import com.netflix.spinnaker.keel.api.plugins.supporting
 import com.netflix.spinnaker.keel.exceptions.InvalidSystemStateException
+import com.netflix.spinnaker.keel.lifecycle.LifecycleEvent
+import com.netflix.spinnaker.keel.lifecycle.LifecycleEventScope.PRE_DEPLOYMENT
+import com.netflix.spinnaker.keel.lifecycle.LifecycleEventStatus.NOT_STARTED
+import com.netflix.spinnaker.keel.lifecycle.LifecycleEventType.BUILD
 import com.netflix.spinnaker.keel.persistence.KeelRepository
 import com.netflix.spinnaker.keel.telemetry.ArtifactVersionUpdated
 import kotlinx.coroutines.launch
@@ -55,12 +59,10 @@ class ArtifactListener(
         if (repository.isRegistered(artifact.name, artifact.artifactType)) {
           val artifactSupplier = artifactSuppliers.supporting(artifact.artifactType)
           if (artifactSupplier.shouldProcessArtifact(artifact)) {
-            val enrichedArtifact = artifactSupplier.addMetadata(artifact.normalized())
-
             log.info("Registering version {} (status={}) of {} artifact {}",
               artifact.version, artifact.status, artifact.type, artifact.name)
 
-            repository.storeArtifactVersion(enrichedArtifact)
+            enrichAndStore(artifact.normalized(), artifactSupplier)
               .also { wasAdded ->
                 if (wasAdded) {
                   publisher.publishEvent(ArtifactVersionUpdated(artifact.name, artifact.artifactType))
@@ -72,6 +74,42 @@ class ArtifactListener(
         } else {
           log.debug("Artifact $artifact is not registered. Ignoring new artifact version.")
         }
+      }
+  }
+
+  /**
+   * Finds the delivery configs that are using an artifact,
+   * and publishes a build lifecycle event for them.
+   */
+  fun createBuildLifecycleEvent(artifact: PublishedArtifact) {
+    log.debug("Publishing build lifecycle event for artifact $artifact")
+    artifact.buildMetadata
+      ?.let { buildMetadata ->
+
+        val data: MutableMap<String,String?> = mutableMapOf()
+        data["buildNumber"] = artifact.metadata["buildNumber"]?.toString()
+        data["commitId"] = artifact.metadata["commitId"]?.toString()
+
+        repository
+          .getAllArtifacts(artifact.artifactType, artifact.name)
+          .forEach { deliveryArtifact ->
+            deliveryArtifact.deliveryConfigName?.let { configName ->
+
+              data["application"] = repository.getDeliveryConfig(configName).application
+
+              publisher.publishEvent(LifecycleEvent(
+                scope = PRE_DEPLOYMENT,
+                artifactRef = deliveryArtifact.toLifecycleRef(),
+                artifactVersion = artifact.version,
+                type = BUILD,
+                id = "build-${artifact.version}",
+                status = NOT_STARTED,
+                text = "Monitoring build for ${artifact.version}",
+                link = buildMetadata.uid,
+                data = data
+              ))
+            }
+          }
       }
   }
 
@@ -92,8 +130,7 @@ class ArtifactListener(
 
       if (latestArtifact != null) {
         log.debug("Storing latest version {} (status={}) for registered artifact {}", latestArtifact.version, latestArtifact.status, artifact)
-        val enrichedArtifact = artifactSupplier.addMetadata(latestArtifact.normalized())
-        repository.storeArtifactVersion(enrichedArtifact)
+        enrichAndStore(latestArtifact.normalized(), artifactSupplier)
       } else {
         log.warn("No artifact versions found for ${artifact.type}:${artifact.name}")
       }
@@ -137,8 +174,7 @@ class ArtifactListener(
 
               if (hasNew) {
                 log.debug("$artifact has a missing version ${latestArtifact.version}, persisting.")
-                val enrichedArtifact = artifactSupplier.addMetadata(latestArtifact.normalized())
-                repository.storeArtifactVersion(enrichedArtifact)
+                enrichAndStore(latestArtifact.normalized(), artifactSupplier)
               } else {
                 log.debug("No new versions to persist for $artifact")
               }
@@ -147,6 +183,12 @@ class ArtifactListener(
         }
       }
     }
+  }
+
+  private fun enrichAndStore(artifact: PublishedArtifact, supplier: ArtifactSupplier<*,*>): Boolean {
+    val enrichedArtifact = supplier.addMetadata(artifact)
+    createBuildLifecycleEvent(enrichedArtifact)
+    return repository.storeArtifactVersion(enrichedArtifact)
   }
 
   private fun getLatestStoredVersion(artifact: DeliveryArtifact): PublishedArtifact? =
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BaseArtifactSupplier.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BaseArtifactSupplier.kt
@@ -5,7 +5,7 @@ import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.artifacts.PublishedArtifact
 import com.netflix.spinnaker.keel.api.artifacts.SortingStrategy
 import com.netflix.spinnaker.keel.api.plugins.ArtifactSupplier
-import com.netflix.spinnaker.keel.services.ArtifactMetadataService
+import com.netflix.spinnaker.keel.igor.artifact.ArtifactMetadataService
 import org.slf4j.LoggerFactory
 
 abstract class BaseArtifactSupplier<A : DeliveryArtifact, V : SortingStrategy>(
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
@@ -4,6 +4,9 @@ import com.netflix.spinnaker.config.LifecycleConfig
 import com.netflix.spinnaker.keel.core.api.DEFAULT_SERVICE_ACCOUNT
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEvent
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEventStatus
+import com.netflix.spinnaker.keel.lifecycle.LifecycleEventStatus.FAILED
+import com.netflix.spinnaker.keel.lifecycle.LifecycleEventStatus.SUCCEEDED
+import com.netflix.spinnaker.keel.lifecycle.LifecycleEventStatus.UNKNOWN
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEventType
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEventType.BAKE
 import com.netflix.spinnaker.keel.lifecycle.LifecycleMonitor
@@ -57,25 +60,15 @@ class BakeryLifecycleMonitor(
           else -> publishUnknownEvent(task, execution)
         }
 
-        if (task.numFailures > 0) {
-          // we only care about consecutive failures, and we just had a success
-          monitorRepository.clearFailuresGettingStatus(task)
-        }
-
         if (execution.status.isComplete()) {
           endMonitoringOf(task)
+        } else {
+          markSuccessFetching(task)
         }
       }
       .onFailure { exception ->
         log.error("Error fetching status for $task: ", exception)
-        if (task.numFailures >= lifecycleConfig.numFailuresAllowed - 1) {
-          log.warn("Too many consecutive errors (${lifecycleConfig.numFailuresAllowed}) " +
-            "fetching the task status for $task. Giving up.")
-          publishExceptionEvent(task)
-          monitorRepository.delete(task)
-        } else {
-          monitorRepository.markFailureGettingStatus(task)
-        }
+        handleFailureFetching(task)
       }
   }
 
@@ -97,22 +90,22 @@ class BakeryLifecycleMonitor(
   }
   private fun publishSucceededEvent(task: MonitoredTask) {
     publisher.publishEvent(task.triggeringEvent.copy(
-      status = LifecycleEventStatus.SUCCEEDED,
+      status = SUCCEEDED,
       link = orcaTaskIdToLink(task),
       text = "Bake succeeded for version ${task.triggeringEvent.artifactVersion}"
     ))
   }
   private fun publishFailedEvent(task: MonitoredTask) {
     publisher.publishEvent(task.triggeringEvent.copy(
-      status = LifecycleEventStatus.FAILED,
+      status = FAILED,
       link = orcaTaskIdToLink(task),
       text = "Bake failed for version ${task.triggeringEvent.artifactVersion}"
     ))
   }
 
-  private fun publishExceptionEvent(task: MonitoredTask) {
+  override fun publishExceptionEvent(task: MonitoredTask) {
     publisher.publishEvent(task.triggeringEvent.copy(
-      status = LifecycleEventStatus.UNKNOWN,
+      status = UNKNOWN,
       link = orcaTaskIdToLink(task),
       text = "Failed to monitor bake of version ${task.triggeringEvent.artifactVersion}" +
         " because we could not get the status ${lifecycleConfig.numFailuresAllowed} times. Status unknown."
@@ -122,7 +115,7 @@ class BakeryLifecycleMonitor(
   private fun publishUnknownEvent(task: MonitoredTask, execution: ExecutionDetailResponse) {
     log.warn("Monitored bake ${task.triggeringEvent} in an unhandled status (${execution.status}")
     publisher.publishEvent(task.triggeringEvent.copy(
-      status = LifecycleEventStatus.UNKNOWN,
+      status = UNKNOWN,
       link = orcaTaskIdToLink(task),
       text = "Bake status unknown for version ${task.triggeringEvent.artifactVersion}"
     ))
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
@@ -11,7 +11,6 @@ import com.netflix.spinnaker.keel.lifecycle.LifecycleEventType
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEventType.BAKE
 import com.netflix.spinnaker.keel.lifecycle.LifecycleMonitor
 import com.netflix.spinnaker.keel.lifecycle.LifecycleMonitorRepository
-import com.netflix.spinnaker.keel.lifecycle.LinkNotFound
 import com.netflix.spinnaker.keel.lifecycle.MonitoredTask
 import com.netflix.spinnaker.keel.orca.ExecutionDetailResponse
 import com.netflix.spinnaker.keel.orca.OrcaExecutionStatus.BUFFERED
@@ -61,14 +60,14 @@ class BakeryLifecycleMonitor(
         }
 
         if (execution.status.isComplete()) {
-          endMonitoringOf(task)
+          endMonitoringOfTask(task)
         } else {
-          markSuccessFetching(task)
+          markSuccessFetchingStatus(task)
         }
       }
       .onFailure { exception ->
         log.error("Error fetching status for $task: ", exception)
-        handleFailureFetching(task)
+        handleFailureFetchingStatus(task)
       }
   }
 
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
@@ -178,8 +178,8 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
       BuildMetadata(
         id = 1,
         uid = "1234",
-        startedAt = "2020-11-24T04:44:04.000",
-        completedAt = "2020-11-25T03:04:02.259",
+        startedAt = "2020-11-24T04:44:04.000Z",
+        completedAt = "2020-11-25T03:04:02.259Z",
         job = Job(
           name = "job bla bla",
           link = "enkins.com"
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
@@ -178,8 +178,8 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
       BuildMetadata(
         id = 1,
         uid = "1234",
-        startedAt = "yesterday",
-        completedAt = "today",
+        startedAt = "2020-11-24T04:44:04.000",
+        completedAt = "2020-11-25T03:04:02.259",
         job = Job(
           name = "job bla bla",
           link = "enkins.com"
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/LifecycleEventRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/LifecycleEventRepositoryTests.kt
@@ -40,7 +40,7 @@ abstract class LifecycleEventRepositoryTests<T: LifecycleEventRepository> : JUni
     id = "bake-$version",
     text = "Submitting bake for version $version",
     link = "www.bake.com/$version",
-    data = mutableMapOf("hi" to "whatsup")
+    data = mapOf("hi" to "whatsup")
   )
   val anotherEvent = event.copy(id = "bake-$version-2")
 
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/LifecycleEventRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/LifecycleEventRepositoryTests.kt
@@ -14,6 +14,7 @@ import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
 import strikt.api.expect
 import strikt.assertions.isEqualTo
+import strikt.assertions.isNotEmpty
 import strikt.assertions.isNotNull
 import strikt.assertions.isNull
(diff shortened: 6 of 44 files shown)


keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -0,0 +1,197 @@
...
+  override fun publishExceptionEvent(task: MonitoredTask) {
+    publisher.publishEvent(task.triggeringEvent.copy(
+      status = UNKNOWN,
+      link = chooseLink(task),
+      text = "Failed to monitor build of version ${task.triggeringEvent.artifactVersion}" +
+        " because we could not get the status ${lifecycleConfig.numFailuresAllowed} times. Status unknown."
+    ))
+  }
+
+  fun chooseLink(task: MonitoredTask): String? {
```
erikmunson, line 158: this practically looks like UI code!

britneyapproved

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -55,12 +59,10 @@ class ArtifactListener(
         if (repository.isRegistered(artifact.name, artifact.artifactType)) {
           val artifactSupplier = artifactSuppliers.supporting(artifact.artifactType)
           if (artifactSupplier.shouldProcessArtifact(artifact)) {
-            val enrichedArtifact = artifactSupplier.addMetadata(artifact.normalized())
-
             log.info("Registering version {} (status={}) of {} artifact {}",
               artifact.version, artifact.status, artifact.type, artifact.name)
 
-            repository.storeArtifactVersion(enrichedArtifact)
+            enrichAndStore(artifact.normalized(), artifactSupplier)
```
emjburns, line 65: changed this to a funciton so we correctly do the chain of things (enrich, send event, and store) in the 3 places it's needed

erikmunson:
yessssssss

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BaseArtifactSupplier.kt
```diff
@@ -5,7 +5,7 @@ import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.artifacts.PublishedArtifact
 import com.netflix.spinnaker.keel.api.artifacts.SortingStrategy
 import com.netflix.spinnaker.keel.api.plugins.ArtifactSupplier
-import com.netflix.spinnaker.keel.services.ArtifactMetadataService
+import com.netflix.spinnaker.keel.igor.artifact.ArtifactMetadataService
```
emjburns, line 8: these are all moved for spring component shenanigans. and, because I don't like having things in a "services" package. things should be grouped based on what they accomplish as a whole, not what type of object they are. that's already in the name.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -57,25 +60,15 @@ class BakeryLifecycleMonitor(
           else -> publishUnknownEvent(task, execution)
         }
```
emjburns, line 62: abstracted some of this into the base.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
```diff
@@ -31,7 +32,8 @@ data class LifecycleEvent(
   val status: LifecycleEventStatus,
   val text: String? = null,
   val link: String? = null,
-  val timestamp: Instant? = null
+  val timestamp: Instant? = null,
+  val data: Map<String, Any?> = emptyMap()
```
emjburns, line 36: @robfletcher I added the ability to store extra data here that a monitor might need. For builds, we want to capture other stuff so we can smartly construct links. So the way I modeled it is that you can add things to a map, and then the data type is in the monitor. the monitor is responsible for parsing things the correct way.

Does this seem reasonable to you?

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/DeliveryConfigImporter.kt
```diff
@@ -10,7 +9,6 @@ import org.springframework.stereotype.Component
 /**
  * Provides functionality to import delivery config manifests from source control repositories (via igor).
  */
```
emjburns, line 11: this is initialized as a bean in the config, it doesn't need the component annotation. Previously we weren't even scanning these packages for components. I fixed that also.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: LOL great.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventType.kt
```diff
@@ -1,6 +1,6 @@
 package com.netflix.spinnaker.keel.lifecycle
 
 enum class LifecycleEventType {
-  BAKE
+  BAKE, BUILD
```
gal-yardeni, line 4: ❤️

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -0,0 +1,197 @@
...
+  @Value("\${spinnaker.baseUrl}") private val spinnakerBaseUrl: String
+) : LifecycleMonitor(monitorRepository, publisher, lifecycleConfig) {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  override fun handles(type: LifecycleEventType): Boolean =
+    type == BUILD
+
+  override suspend fun monitor(task: MonitoredTask) {
+    val buildData = parseAndValidate(task) ?: return
+    kotlin.runCatching {
```
gal-yardeni, line 46: is calling to `getArtifactMetadata` done only if the build data was missing?

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -0,0 +1,197 @@
...
+   }
+
+  private fun parseBuildData(task: MonitoredTask): BuildData =
+      objectMapper.convertValue(task.triggeringEvent.data)
+
+
+  private fun BuildMetadata.isComplete(): Boolean =
+    status == "SUCCESS" || status == "FAILED" || status == "ABORTED"
+
+  private fun publishRunningEvent(task: MonitoredTask) {
```
gal-yardeni, line 96: now that we have both bake and builds events, is it possible to extract those methods to a single place to make a cleaner code?

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -0,0 +1,197 @@
...
+        front50Service.applicationByName(buildData.application)
+      } catch (e: HttpException) {
+        if (e.isNotFound) {
+          null
+        } else {
+          throw e
+        }
+      }
+    }
+
```
gal-yardeni, line 171: nit: new line (and below)

gal-yardeni:
Love it! Can't wait to see it live!
Added small questions / comments but it looks great to me.

General question - are we handling here running builds, prior to producing an artifact?

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: No, it's how we update the status. so it's called every 30 seconds for any active builds. Likely it'll be called max 4 times for each build, since they're mostly done when we get them.

Should we call a different endpoint instead?

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: It wouldn't be any cleaner. youd still have to pass in the message, and the link, and the original event. so you'd save maybe 1 line of code.

and, it would make it more confusing I think.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -75,6 +77,44 @@ class ArtifactListener(
...
+   */
+  fun createBuildLifecycleEvent(artifact: PublishedArtifact) {
+    log.debug("Publishing build lifecycle event for artifact $artifact")
+    artifact.buildMetadata
+      ?.let { buildMetadata ->
+
+        val data: MutableMap<String,String?> = mutableMapOf()
+        data["buildNumber"] = artifact.metadata["buildNumber"]?.toString()
+        data["commitId"] = artifact.metadata["commitId"]?.toString()
+        data["fallbackLink"] = buildMetadata.job?.link
```
luispollo, line 92:
```suggestion
        val data = mutableMapOf(
          "buildNumber" to artifact.metadata["buildNumber"]?.toString(),
          "commitId" to artifact.metadata["commitId"]?.toString(),
          "fallbackLink" to buildMetadata.job?.link
        )
```

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
gal-yardeni: I'm thinking if we could utilize the info we are currently storing in the DB somehow like only call this endpoint if the build is in `building` status.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -75,6 +77,44 @@ class ArtifactListener(
       }
   }
 
+  /**
+   * Finds the delivery configs that are using an artifact,
+   * and publishes a build lifecycle event for them.
+   */
+  fun createBuildLifecycleEvent(artifact: PublishedArtifact) {
```
luispollo, line 84: Maybe `publishBuildNotStartedEvent`, since this is the only event this function will ever publish? Also, why `NOT_STARTED` here since this is a new artifact version and so a build has definitely already started? I understand that sometimes we get the artifact event _before_ the build finished, but it's definitely at least running by the time we get here, right?

luispollo: BTW, `BuildMetadata` has a `status` field that we get from CI.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -92,8 +132,7 @@ class ArtifactListener(
 
       if (latestArtifact != null) {
         log.debug("Storing latest version {} (status={}) for registered artifact {}", latestArtifact.version, latestArtifact.status, artifact)
-        val enrichedArtifact = artifactSupplier.addMetadata(latestArtifact.normalized())
-        repository.storeArtifactVersion(enrichedArtifact)
+        enrichAndStore(latestArtifact.normalized(), artifactSupplier)
```
luispollo, line 135: Looks like you're always calling `normalized()` before calling `enrichAndStore()` -- you could probably just move the call to the former there.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -57,25 +60,15 @@ class BakeryLifecycleMonitor(
...
 
-        if (task.numFailures > 0) {
-          // we only care about consecutive failures, and we just had a success
-          monitorRepository.clearFailuresGettingStatus(task)
-        }
-
         if (execution.status.isComplete()) {
           endMonitoringOf(task)
+        } else {
+          markSuccessFetching(task)
```
luispollo, line 66: I personally dislike method names that rely on good parameter names to understand what they do. I'd suggest `markSuccessFetchingTask` instead.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -57,25 +60,15 @@ class BakeryLifecycleMonitor(
...
         log.error("Error fetching status for $task: ", exception)
-        if (task.numFailures >= lifecycleConfig.numFailuresAllowed - 1) {
-          log.warn("Too many consecutive errors (${lifecycleConfig.numFailuresAllowed}) " +
-            "fetching the task status for $task. Giving up.")
-          publishExceptionEvent(task)
-          monitorRepository.delete(task)
-        } else {
-          monitorRepository.markFailureGettingStatus(task)
-        }
+        handleFailureFetching(task)
```
luispollo, line 71: Same comment on the method name here.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
luispollo: Just a quick comment here that you actually ended up with logic in a model class (`BuildMetadata`) that knows about and converts between types, not just in the monitor.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
```diff
@@ -61,4 +48,40 @@ abstract class LifecycleMonitor(
...
     monitorRepository.delete(task)
   }
+
+  /**
+   * Handles persisting failures and publishing exception events.
+   *
+   * Call when there was a failure getting status.
+   */
+  fun handleFailureFetching(task: MonitoredTask) {
+    if (task.numFailures >= lifecycleConfig.numFailuresAllowed - 1) {
```
luispollo, line 58: I know the original logic was already like this, but it strikes me as odd that we're comparing with `>=` and _also_ subtracting `1`. Let's say `lifecycleConfig.numFailuresAllowed` is 3. We'll consider it an error if `task.numFailures` is equal to or more than 2. That doesn't sound right, unless we're doing something weird when counting failures...

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: Hmm, good point. Let me do that!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
```diff
@@ -61,4 +48,40 @@ abstract class LifecycleMonitor(
...
+      monitorRepository.markFailureGettingStatus(task)
+    }
+  }
+
+  /**
+   * Handles clearing the failure status if there was one.
+   *
+   * Call when status has successfully been fetched.
+   */
+  fun markSuccessFetching(task: MonitoredTask) {
```
luispollo, line 73: Suggestion:
```suggestion
  fun markSuccessFetchingTaskStatus(task: MonitoredTask) {
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
```diff
@@ -61,4 +48,40 @@ abstract class LifecycleMonitor(
     log.debug("${this.javaClass.simpleName} has completed monitoring for $task")
     monitorRepository.delete(task)
   }
+
+  /**
+   * Handles persisting failures and publishing exception events.
+   *
+   * Call when there was a failure getting status.
+   */
+  fun handleFailureFetching(task: MonitoredTask) {
```
luispollo, line 57: Suggestion:
```suggestion
  fun handleFailureFetchingTaslStatus(task: MonitoredTask) {
```

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -0,0 +1,197 @@
...
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.boot.context.properties.EnableConfigurationProperties
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.stereotype.Component
+import retrofit2.HttpException
+
+@Component
+@EnableConfigurationProperties(LifecycleConfig::class)
+class BuildLifecycleMonitor(
```
luispollo, line 30: Consider adding doc comments.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -0,0 +1,197 @@
...
+      if (buildMetadata == null) {
+        log.error("Error fetching status for $task, response was null")
+        handleFailureFetching(task)
+      } else {
+        when (buildMetadata.status) {
+          "BUILDING" -> publishRunningEvent(task)
+          "SUCCESS" -> publishSucceededEvent(task, buildMetadata)
+          "FAILURE" -> publishFailedEvent(task, buildMetadata)
+          "ABORTED" -> publishAbortedEvent(task, buildMetadata)
+          "UNSTABLE" -> publishSucceededEvent(task, buildMetadata)
```
luispollo, line 59: Can you explain the mapping between `UNSTABLE` and `SUCCEEDED` here?

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -0,0 +1,197 @@
...
+          else -> publishUnknownStatusEvent(task, buildMetadata.status)
+        }
+
+        if (buildMetadata.isComplete()) {
+          endMonitoringOf(task)
```
luispollo, line 64: Same comment on making method names self-contained.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -0,0 +1,197 @@
...
+      }
+    }
+      .onFailure { exception ->
+        log.error("Error fetching status for $task: ", exception)
+        handleFailureFetching(task)
+      }
+  }
+
+  /**
+   * Parses build data if format is correct, otherwise publish an Unknown event
```
luispollo, line 77:
```suggestion
   * Parses build data if format is correct, otherwise publishes an Unknown event
```

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -0,0 +1,197 @@
...
+      endMonitoringOf(task)
+      null
+   }
+
+  private fun parseBuildData(task: MonitoredTask): BuildData =
+      objectMapper.convertValue(task.triggeringEvent.data)
+
+
+  private fun BuildMetadata.isComplete(): Boolean =
+    status == "SUCCESS" || status == "FAILED" || status == "ABORTED"
```
luispollo, line 94: How about `UNSTABLE`?

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -0,0 +1,197 @@
...
+    // ci not explicitly disabled, check and see if ci view is configured
+    return if (app.repoProjectKey != null && app.repoSlug != null && app.repoType != null) {
+      buildUidToLink(task)
+    } else {
+      jenkinsLink(buildData)
+    }
+  }
+
+  private fun buildUidToLink(task: MonitoredTask): String =
+    "$spinnakerBaseUrl/#/applications/${task.triggeringEvent.data["application"]}/builds/${task.link}/logs"
```
luispollo, line 186:
```suggestion
    "$spinnakerBaseUrl/#/applications/${app.name}/builds/${task.link}/logs"
```

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
luispollo, line 158: This function is 💯 !

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/DeliveryConfigImporter.kt
luispollo: Maybe we should stick with one of those approaches, though? Consider keeping the `@Component` annotation and removing the `@Bean` factory function.

keel-igor/src/test/kotlin/com/netflix/spinnaker/keel/services/BuildLifecycleMonitorTests.kt
```diff
@@ -0,0 +1,312 @@
...
+      }
+
+      context("UNSTABLE") {
+        before {
+          coEvery { artifactMetadataService.getArtifactMetadata(buildNumber, commitId) } returns
+            getArtifactMetadata("UNSTABLE")
+          runBlocking { subject.monitor(task) }
+        }
+
+        test("emits succeeded event") {
```
luispollo, line 172: Consider adding a comment here explaining why it is that we expect a succeeded event when the status is `UNSTABLE`.

keel-igor/src/test/kotlin/com/netflix/spinnaker/keel/services/BuildLifecycleMonitorTests.kt
```diff
@@ -0,0 +1,312 @@
...
+import io.mockk.slot
+import io.mockk.verify
+import kotlinx.coroutines.runBlocking
+import org.springframework.context.ApplicationEventPublisher
+import strikt.api.expectThat
+import strikt.assertions.isEqualTo
+import strikt.assertions.isTrue
+import java.time.Instant
+
+class BuildLifecycleMonitorTests : JUnit5Minutests {
```
luispollo, line 37: Love the tests! Great job! ❤️

luispollo (commented):
Overall, this looks great to me, @emjburns! I did have a couple of questions and will approve once they've been addressed.

robfletcher:
> Also, linking to the CI view is complicated. We have to do some digging through the app config to figure out if the ci view is wired in and working.

Shouldn't we let Deck worry about that? Why are we working out links server side?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
```diff
@@ -13,7 +17,23 @@ data class BuildMetadata(
...
   val completedTs: String? = null,
   val status: String? = null
-)
+) {
+  fun getStartedAtInstant(): Instant? =
+    startedAt
+      ?.let {
+        LocalDateTime.parse(it.removeSuffix("Z"))
+          .atZone(UTC)
+          .toInstant()
```
robfletcher, line 26: Pretty sure you can just parse the entire string as a `ZonedDateTime` rather than having to manipulate it like that.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
```diff
@@ -13,7 +17,23 @@ data class BuildMetadata(
   val startedTs: String? = null,
   val completedTs: String? = null,
   val status: String? = null
-)
+) {
+  fun getStartedAtInstant(): Instant? =
```
robfletcher, line 21: Rather than having this as a method I think the Kotlin-y way to do it would be:

```
val startedAtInstant: Instant?
  get() = // .. the rest of your implementation

File: keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt [hunk 29]
```diff
@@ -13,7 +17,23 @@ data class BuildMetadata(
   val startedTs: String? = null,
```
robfletcher, line 17: What is the difference between `startedAt` and `startedTs`? Why do we need both?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
```diff
@@ -13,7 +17,23 @@ data class BuildMetadata(
...
+          .toInstant()
+      }
+
+  fun getCompletedAtInstant(): Instant? =
+    completedAt
+      ?.let {
+        LocalDateTime.parse(it.removeSuffix("Z"))
+          .atZone(UTC)
+          .toInstant()
+      }
```
robfletcher, line 35:
```suggestion
  val completedAtInstant: Instant?
    get() = completedAt?.let { ZonedDateTime.parse(it).toInstant() }
```

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
robfletcher: Why does it need to be mutable?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -178,8 +178,8 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
       BuildMetadata(
         id = 1,
         uid = "1234",
-        startedAt = "yesterday",
-        completedAt = "today",
+        startedAt = "2020-11-24T04:44:04.000",
+        completedAt = "2020-11-25T03:04:02.259",
```
robfletcher, line 182: That doesn't seem to be the format in actual metadata I've looked at in the DB. They should include a zone marker.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/LifecycleEventRepositoryTests.kt
```diff
@@ -38,7 +39,8 @@ abstract class LifecycleEventRepositoryTests<T: LifecycleEventRepository> : JUni
     status = NOT_STARTED,
     id = "bake-$version",
     text = "Submitting bake for version $version",
-    link = "www.bake.com/$version"
+    link = "www.bake.com/$version",
+    data = mutableMapOf("hi" to "whatsup")
```
robfletcher, line 43: Why mutable?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
robfletcher: It seems reasonable if we don't know what fields may or may not be present

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
```diff
@@ -40,6 +23,10 @@ abstract class LifecycleMonitor(
 
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
+  init {
+    log.debug("Lifecycle monitor ${javaClass.simpleName} enabled")
+  }
```
robfletcher, line 28: Are you sure this will only get logged if it's enabled? (Assuming that's what you're trying to show)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
robfletcher: What's a `tasl`? :trollface:

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
robfletcher: That does seem pretty confusing. Isn't it the same as `if (task.numFailures > lifecycleConfig.numFailuresAllowed)`?

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
robfletcher, line 46:
```suggestion
    runCatching {
```
Dunno why that didn't auto-import for you

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
robfletcher: IIRC "unstable" means in Jenkins that it passed now, but it's failed recently

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
robfletcher: Or even make it an extension function so it reads as `task.endMonitoring()` 🤯

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
robfletcher: Is that a good thing?

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -0,0 +1,197 @@
...
+      status = UNKNOWN,
+      link = chooseLink(task),
+      text = "Failed to monitor build of version ${task.triggeringEvent.artifactVersion}" +
+        " because we could not get the status ${lifecycleConfig.numFailuresAllowed} times. Status unknown."
+    ))
+  }
+
+  fun chooseLink(task: MonitoredTask): String? {
+    val buildData = parseBuildData(task)
+    val app = runBlocking {
```
robfletcher, line 160: This is going to block the event dispatch thread while it makes a call to Fiddy? Seems like we could decouple that to avoid blocking.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
robfletcher, line 186: I'm confused about why we'd want to do this server side

robfletcher (commented):
Made a few comments. Still need to review that test at the end of the diff more thoroughly.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: @gal-yardeni that's a good idea, I'll do that.

emjburns: Actually, I want to leave the logic that encapsulates status in one place. Right now I can't think of a good way to publish the right events w/o checking the API, or wiring in the other repository (which isn't super easy because we don't have the full delivery artifact, cause saving that was hard). I think I'm going to leave it as is, and if the load gets too high on the API we're calling we can update it. does that sound okay?

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
emjburns: Because we add application in it for each delivery artifact (line 103). But I'll change to the more kotlin-y way of initializing. i was apparently stuck in java land

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
emjburns: So, right now, in order to kick of monitoring the event has to have a "NOT_STARTED" status. Then the build monitor updates to the correct status almost immediately. I'm going to change that in a different pr (to have a `monitor: Boolean` flag), but I don't want to wrap that change into this pr. But, I didn't know that I'd want to use these things this way until I had another type of thing to monitor! Fun discoveries using things that are freshly built :) 

In light of that,  I don't want to change the method name.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
emjburns: Good call

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
emjburns: We will compromise on `markSuccessFetchingStatus` and `handleFailureFetchingStatus`

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
emjburns: To me, handle implies it's doing more, which that method is doing.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
emjburns: So this is called when we have a failure, which means the number of failures that we have is actually one greater than what is saved on the task. If we only want to allow 3 failures, we shouldn't try a fourth time (I think). So this was how I wrote the logic to do that.

In the case that you wrote @robfletcher, we'd allow one more failure than the config states

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
gal-yardeni: Good point. I just looked at your code again, and another thing you can do is to pass the build status in `ArtifactListener`:`createBuildLifecycleEvent`. If it's already in a terminal status, you could save that API call.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: These come from jenkins. Apparently "unstable" here means that the build passed but the tests failed. Since the build passed, I've chosen to call that a "success". I'm expecting to change this if we get user feedback, but I thought this was a good start.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: good point

keel-igor/src/test/kotlin/com/netflix/spinnaker/keel/services/BuildLifecycleMonitorTests.kt
emjburns: I'll add a comment in the build monitor itself

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
emjburns: No idea, this is what comes from the API.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
emjburns: yep, you are right! good catch

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/LifecycleEventRepositoryTests.kt
emjburns: oh idk. not needed in this test.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
emjburns: Well, there's no way to disable it, so .... yes? :D I kind of just want a confirmation that it's being used. I like those comforting log messages.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: Because... erik asked me to. He said we could do it on the UI side, but it would be better to just have the data.

Also, we don't have support right now for an "alternate link". It makes sense to calculate the link shown on the back end because we can do more complex logic more easily.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: Ya. How would you recommend decoupling this? making it a suspend function?

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/DeliveryConfigImporter.kt
emjburns: Considering the `@Bean` approach is the one that worked for that class, I didn't want to make an unrelated change in my PR. We will need both methods anyways since the `@Bean` one is used for creating retrofit clients

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: @gal-yardeni ya, but then that does weird things with how I roll up the events into user-facing steps (like, how do you get both the start and end time?). I've added a todo to make this a little better.

gal-yardeni: you can pass the entire build metadata for it, I think. The reason I'm suggesting it is because if the build is terminal, the build metadata object you'll get from this call is duplicate to the one in `artifactListener`, and as we have some precentege of builds already in terminal states, we don't need to monitor them again, just send an event.
We should give rocket a heads up that we are adding another scheduled call to their services, just in case :)

luispollo: +1 to Gal's suggestion. Also, in my observation, we rarely ever go back to fix `TODOs` in the code...

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
luispollo: On naming, maybe just `publishBuildLifecycleEvent` as a compromise, then? At least you use the same verb as in other places.

On the `NOT_STARTED` status, I think what's confusing here is that you seem to be overloading that status field to capture state for both the act of monitoring the build (where this status would totally make sense), and the status of the build itself. I think these should be separate fields. Maybe that's what you meant by the `monitor: Boolean` flag, not sure.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
robfletcher: I can't math

luispollo: I think that just means you should be incrementing `task.numFailures` before you reach this line...

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
robfletcher: Yeah, but if we only ever use one of them, do we actually need the other on our model?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleMonitor.kt
luispollo: This was just a test to see who was paying attention... 😝

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
luispollo: I think it's what @robfletcher said. It was failing before, but now it passed.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -43,56 +47,53 @@ class BuildLifecycleMonitor(
...
         log.error("Error fetching status for $task, response was null")
-        handleFailureFetching(task)
+        handleFailureFetchingStatus(task)
       } else {
         when (buildMetadata.status) {
           "BUILDING" -> publishRunningEvent(task)
           "SUCCESS" -> publishSucceededEvent(task, buildMetadata)
           "FAILURE" -> publishFailedEvent(task, buildMetadata)
           "ABORTED" -> publishAbortedEvent(task, buildMetadata)
+          // UNSTABLE means build passed but tests failed (might need to reevaluate status in the future)
```
luispollo, line 63: The official definition is: 
> The Build had some errors but they were not fatal. A Build is unstable if it was built successfully and one or more publishers report it unstable. For example if the JUnit publisher is configured and a test fails then the Build will be marked unstable.

Source, if you want to add the link to the comment: https://www.jenkins.io/doc/book/glossary/

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
luispollo: @emjburns you were right! I added the "official" definition and a link in a comment above.

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: discussed offline, we can optimize later

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
```diff
@@ -33,7 +37,7 @@ class BuildLifecycleMonitor(
   override val lifecycleConfig: LifecycleConfig,
   val objectMapper: ObjectMapper,
   val artifactMetadataService: ArtifactMetadataService,
-  val front50Service: Front50Service,
+  val front50Service: Front50Service?,
```
luispollo, line 40: Why is `front50Service` optional?

luispollo (approved):
Changes look good to me! Thanks for compromising on the names. I do think you should address the `NOT_STARTED` event status comments, but up to you if you want to do that in a separate PR.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
emjburns: Fair point, but an unrelated change. I'll defer to @gal-yardeni on what we use

keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/BuildLifecycleMonitor.kt
emjburns: Because it's not actually a required dependency of keel, and I don't think this is important to make it so. If we can't look up app details in front50 here, because it's not wired in, then we will fallback to the jenkins URL. that seemed fine to me.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
emjburns: @luispollo yep, my plan is to decouple the status from whether we have to monitor or not so that the status of the event can be the status of the actual think. I'll do that in an immediately following pr. I'll address naming changes in that PR because I'll need to touch the method anyway.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/BuildMetadata.kt
gal-yardeni: We get this info from rocket. I'll take a pass later this week to see if it's still needed, thanks for pointing it out!

PR #1680, opened 2020-12-02
feat(orca): initial support for container jobs

lorin:
# Problem

To support the new verifyWith functionaly, keel needs the ability to launch container-based (batch) jobs on Titus.

# Proposed solution

Leverage the existing OrcaTaskLauncher so that we can run container jobs through orca. The code in this PR
enables us to create orchestrations that consist of a single *Run Job* stage.

This PR is a bottom-up approach: introducing some new low-level primitives that we'll be able to use. The
primitives aren't wired up to anything yet.

# Classes introduced

## ContainerJobConfig

This class is essentially a one-to-one mapping to the parameters supported by the Run Job stage. I did my
best to reverse-engineer the default parameters through experiments running jobs using deck and examining the
generated JSON.

## JobStatus

This class contains information about a job running on Titus. I've added it here because it contains information we 
need to get the URLs for stdout/stderr and other logs. (Calculation of those URLs is coming in a future PR).

# Other changes

## Parsing data from "variables" field of Orca response

Based on experiments, I saw that when we launch an orchestration in Orca, there's a top-level `variables` field
that gets populated based on the output of stages. It seemed easier to parse the job status info from there
than from the stage output, so that's how I implemented it.

# Examples of how to use this code

## Launching a job

```kotlin
val application = ...
val serviceAccount = ...

// Describing the job
val containerJobConfig = ContainerJobConfig(
  name = "run from keel",
  organization = "lhochstein",
  repository = "lhochstein/ducker",
  
  region = "us-east-1",
  account = "titustestvpc",
  credentials = "titustestvpc",
  registry = "testregistry",
  
  application = application,
  serviceAccount = serviceAccount
)

// Submitting the job to orca
val orcaTaskLauncher: OrcaTaskLauncher = ...

val response : ExecutionDetailResponse = orcaTaskLauncer.submitJob(
 stages = listOf(container.createRunJobStage()),

 subject = "container integration test",
 description = "testing running a container",
 correlationId = "keel-run-container-test",
 
 user = serviceAccount,
 
 application = application,
 notifications = emptySet()
)

val taskId = response.id
```

(I know this looks verbose, but I didn't want to abstract this any further until we start actually consuming it).

## Getting the job status

```kotlin
val jobStatus = orcaService
  .getOrchestrationExecution(taskId, serviceAccount)
  .getJobStatus()
```

# Feedback sought

Some questions I had working on this:

* Is the naming and package location reasonable?

* Is it reasonable to reuse the OrcaTaskLauncher for this? Currently, the launcher is only used for actuating resources.

* How complete should the data representation be? The ContainerJobConfig contains all of the values that seem to get set by default when launching a job from Deck, but it's not fully comprehensive: there are optional fields that currently aren't specified here. I assumed it would be better to just add them later. 

# FAQ

## Hey, where are the tests?

There isn't really any logic in this PR to unit tests. I wrote an end-to-end as a CLI, but I had to put that in the internal repo because I needed to use metatron-based auth for it to work. I'll submit an internal PR for that.

### DIFF
--- a/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaService.kt
+++ b/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaService.kt
@@ -75,6 +75,11 @@ data class TaskRefResponse(
   val taskId by lazy { ref.substringAfterLast("/") }
 }
 
+data class KeyValuePair(
+  val key: String,
+  val value: Any
+)
+
 data class ExecutionDetailResponse(
   val id: String,
   val name: String,
@@ -84,7 +89,8 @@ data class ExecutionDetailResponse(
   val endTime: Instant?,
   val status: OrcaExecutionStatus,
   val execution: OrcaExecutionStages = OrcaExecutionStages(emptyList()),
-  val stages: List<OrcaExecutionStage>? = emptyList() // for pipelines, stages are not encapsulated in `execution`
+  val stages: List<OrcaExecutionStage>? = emptyList(), // for pipelines, stages are not encapsulated in `execution`
+  val variables: List<KeyValuePair>? = null
 )
 
 typealias OrcaExecutionStage = Map<String, Any>
--- a/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/container/ContainerJobConfig.kt
+++ b/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/container/ContainerJobConfig.kt
@@ -23,20 +23,16 @@ data class ContainerJobConfig(
   val account: String,
   val region: String,
   val organization: String,
-
   /**
    * Repository name associated with the container, e.g.: "acme/widget"
    */
   val repository: String,
-
   val type: ImageVersionReference = TAG,
-
   /**
    * Either a tag (e.g., "latest") or a digest (e.g., "sha256:be93efc727ba59813adc896859bccf32fb0f02202fe0526a9cd76326b9729cb3),
    * depending on what [type] is
    */
   val versionIdentifier: String = "latest",
-
   /**
    * ID that identifies an image, e.g.:
    *
@@ -44,53 +40,29 @@ data class ContainerJobConfig(
    * acme/widget:sha256:780f11bfc03495da29f9e2d25bf55123330715fb494ac27f45c96f808fd2d4c5
    */
   val imageId: String = "$repository:$versionIdentifier",
-
   val desiredCapacity: Int= 1,
-
   val maxCapacity: Int = 1,
-
   val minCapacity: Int = 1,
-
   val cpus: Int = 2,
-
   val memoryMb: Int = 2048,
-
   val diskMb: Int = 256,
-
   val networkMbps: Int = 1024,
-
   val gpus: Int = 0,
-
   val entrypoint: String = "",
-
   val runtimeLimitSeconds: Int = 2700,
-
   val retries: Int = 0,
-
   val cloudProvider : String = "titus",
-
   val cloudProviderType: String = "aws",
-
   val serviceAccount: String,
-
   val iamInstanceProfile: String = application + "InstanceProfile",
-
   val securityGroups: List<String> = emptyList(),
-
   val capacityGroup: String = application,
-
   val credentials: String,
-
   val registry : String,
-
   val containerAttributes: Map<String, String> = emptyMap(),
-
   val environmentVariables: Map<String, String> = emptyMap(),
-
   val labels: Map<String, String> = emptyMap(),
-
   val deferredInitialization : Boolean = true,
-
   val waitForCompletion: Boolean = true
 ) {
 
--- a/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/container/ContainerJobConfig.kt
+++ b/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/container/ContainerJobConfig.kt
@@ -73,7 +73,7 @@ data class ContainerJobConfig(
 
   val serviceAccount: String,
 
-  val iamInstanceProfile: String = application + "instanceProfile",
+  val iamInstanceProfile: String = application + "InstanceProfile",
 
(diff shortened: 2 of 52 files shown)


keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/container/ContainerJobConfig.kt
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * Containers can be identified by either:
+ *  tag (e.g., "latest")
+ *  digest (e.g., "sha256:780f11bfc03495da29f9e2d25bf55123330715fb494ac27f45c96f808fd2d4c5")
+ */
+enum class ImageVersionReference {
+  TAG,
+  DIGEST
+}
+
```
emjburns, line 16: nit: It's weird that `ContainerJobConfig(` is a data class but the parameters have newlines in between them.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/container/ContainerJobConfig.kt
```diff
@@ -0,0 +1,151 @@
...
+  val retries: Int = 0,
+
+  val cloudProvider : String = "titus",
+
+  val cloudProviderType: String = "aws",
+
+  val serviceAccount: String,
+
+  val iamInstanceProfile: String = application + "instanceProfile",
```
emjburns, line 76: does "instance" need a capital i here?

lorin:
@emjburns and I had a discussion about what module this code should live in. Where we landed is that, since this code is coupled to the implementation of Titus, it makes more sense for this code to live in `keel-titus-plugin` rather than `keel-orca`.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/container/ContainerJobConfig.kt
lorin: > nit: It's weird that `ContainerJobConfig(` is a data class but the parameters have newlines in between them.

Fun fact: `ContainerJobConfig`: started out its life as an interface, and then morphed into a data class before I submitted the PR, but I forgot to clean up the spacing.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/container/ContainerJobConfig.kt
lorin: Indeed it does. Will fix.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/container/ContainerJobConfig.kt
lorin: Fixed.

robfletcher (approved):
This seems fine as far as it goes. I'm in two minds about using Orca but I think it _at least_ makes sense to leverage it initially.

robfletcher:
> @emjburns and I had a discussion about what module this code should live in. Where we landed is that, since this code is coupled to the implementation of Titus, it makes more sense for this code to live in `keel-titus-plugin` rather than `keel-orca`. I'm going to update this PR.

Given that depends on `keel-orca` that seems like a reasonable decision

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaService.kt
```diff
@@ -75,6 +75,11 @@ data class TaskRefResponse(
   val taskId by lazy { ref.substringAfterLast("/") }
 }
 
+data class KeyValuePair(
+  val key: String,
+  val value: Any
+)
```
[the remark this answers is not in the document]
lorin: > You could always just use Kotlin's `Pair<String, Any>` unless there's a compelling reason to have our own type.

I'm using the data class as a deserialization target to parse out JSON from Spinnaker that looks like this:

```json
{
  "key": "jobStatus",
  "value": {
    ...
   } 
}
```

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -0,0 +1,123 @@
...
+ */
+enum class ImageVersionReference {
+  TAG,
+  DIGEST
+}
+
+/**
+ * Holds the config info needed to execute a container via the Run Job stage
+ */
+data class ContainerJobConfig(
```
gal-yardeni, line 20: Is it possible to rearrange this data class, so that all mandatory fields without default values would be on top?

gal-yardeni (approved):
makes sense to me. `titus` is probably a good place to hold this logic.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
lorin: done!

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaService.kt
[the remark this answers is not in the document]
luispollo: Wouldn't `List<Pair<String, Any>>?` deserialize just fine below, though?

luispollo:
> @emjburns and I had a discussion about what module this code should live in. Where we landed is that, since this code is coupled to the implementation of Titus, it makes more sense for this code to live in `keel-titus-plugin` rather than `keel-orca`. I'm going to update this PR.

@lorin Sorry I was late to review, but I looked over the changes and couldn't spot where we depend on any types from `keel-titus-plugin`... Could you please point me to it? To me, `keel-orca` would have seemed like the right place for this, and if there were any common types between the run job implementation and the Titus plugin, they should be in `keel-titus-api`. Not a big deal, maybe something to improve in a separate PR.

PR #1681, opened 2020-12-02
Scheduler for running verifications

robfletcher:
This adds a new `@Scheduled` function to `CheckScheduler` that invokes verifications for environment/artifact combinations using the current version deployed in the environment.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
@@ -33,6 +33,7 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
 
   open fun VerificationContext.setup() {}
   open fun VerificationContext.setupCurrentArtifactVersion() {}
+  open fun VerificationContext.pauseApplication() {}
 
   private data class DummyVerification(override val id: String) : Verification {
     override val type = "dummy"
@@ -185,25 +186,25 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
       version = "fnord-0.190.0-h378.eacb135"
     )
 
+    private fun next() = expectCatching {
+      subject.nextEnvironmentsForVerification(minAge, limit)
+    }
+
     @Test
     fun `nothing is returned to check if there is no current artifact version for any environment`() {
       context.setup()
 
-      expectCatching {
-        subject.nextEnvironmentsForVerification(minAge, limit)
-      }
-        .isSuccess()
-        .isEmpty()
+      next().isSuccess().isEmpty()
     }
 
     @Test
     fun `returns the current version if it has yet to be verified`() {
-      context.setup()
-      context.setupCurrentArtifactVersion()
-
-      expectCatching {
-        subject.nextEnvironmentsForVerification(minAge, limit)
+      with(context) {
+        setup()
+        setupCurrentArtifactVersion()
       }
+
+      next()
         .isSuccess()
         .hasSize(1)
         .withFirst {
@@ -211,6 +212,17 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
           get { environmentName } isEqualTo context.environmentName
           get { deliveryConfig.name } isEqualTo context.deliveryConfig.name
         }
+    }
+
+    @Test
+    fun `returns nothing if the application is paused`() {
+      with(context) {
+        setup()
+        setupCurrentArtifactVersion()
+        pauseApplication()
+      }
+
+      next().isSuccess().isEmpty()
     }
   }
 }
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
@@ -305,5 +305,43 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
           get { version } isEqualTo context2.version
         }
     }
+
+    @Test
+    fun `can return both artifacts sequentially when an environment has more than one`() {
+      val artifact1 = DockerArtifact(
+        name = "artifact1",
+        deliveryConfigName = context.deliveryConfig.name,
+        reference = "ref-1"
+      )
+      val artifact2 = DockerArtifact(
+        name = "artifact2",
+        deliveryConfigName = context.deliveryConfig.name,
+        reference = "ref-2"
+      )
+      val deliveryConfig = context.deliveryConfig.copy(
+        artifacts = setOf(artifact1, artifact2)
+      )
+      val context1 = context.copy(
+        deliveryConfig = deliveryConfig,
+        artifactReference = artifact1.reference,
+        version = "artifact1-0.254.0-h645.4ce7392"
+      )
+      val context2 = context.copy(
+        deliveryConfig = deliveryConfig,
+        artifactReference = artifact2.reference,
+          version = "artifact2-0.390.0-h584.93a3040"
+      )
+      with(context1) {
+        setup()
+        setupCurrentArtifactVersion()
+      }
+      with(context2) {
+        setup()
+        setupCurrentArtifactVersion()
+      }
+
+      next().isSuccess().hasSize(1)
+      next().isSuccess().hasSize(1)
+    }
   }
 }
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
@@ -24,6 +24,7 @@ import strikt.assertions.isNotEmpty
 import strikt.assertions.isNotNull
 import strikt.assertions.isNull
 import strikt.assertions.isSuccess
+import strikt.assertions.one
 import strikt.assertions.withFirst
 import java.time.Duration
 
@@ -53,7 +54,8 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
         artifacts = setOf(
           DockerArtifact(
             name = "fnord",
-            deliveryConfigName = "fnord-manifest"
+            deliveryConfigName = "fnord-manifest",
+            reference = "fnord-docker"
           )
         ),
         environments = setOf(
@@ -64,6 +66,7 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
         )
       ),
       environmentName = "test",
+      artifactReference = "fnord-docker",
       version = "fnord-0.190.0-h378.eacb135"
     )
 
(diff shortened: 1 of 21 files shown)


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
```diff
@@ -148,4 +166,182 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
...
+      }
+      with(context2) {
+        setup()
+        setupCurrentArtifactVersion()
+      }
+
+      next().isSuccess().hasSize(1)
+      next().isSuccess().hasSize(1)
+    }
+  }
```
robfletcher, line 346: All the tests for the query selecting the next things to verify

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -167,6 +173,38 @@ class CheckScheduler(
...
+            }
+        }
+      }
+
+      runBlocking { job.join() }
+      log.debug("Scheduled environment verification complete")
+    } else {
+      log.debug("Scheduled environment verification disabled")
+    }
+  }
```
robfletcher, line 206: New scheduled action

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlStorageContext.kt
```diff
@@ -0,0 +1,20 @@
...
+  internal val jooq: DSLContext,
+  internal val clock: Clock,
+  internal val sqlRetry: SqlRetry,
+  internal val objectMapper: ObjectMapper,
+  internal val resourceSpecIdentifier: ResourceSpecIdentifier,
+  internal val artifactSuppliers: List<ArtifactSupplier<*, *>>,
+  internal val specMigrators: List<SpecMigrator<*, *>>
+) {
+  internal val resourceFactory = ResourceFactory(objectMapper, resourceSpecIdentifier, specMigrators)
+}
```
robfletcher, line 20: This is the start of some refactoring I've been talking about. For now I have some of the repositories extend this, which allows them to share some JOOQ / SQL code that is useful in multiple repository contexts. Eventually I'd like to get rid of the repositories altogether and replace them with a library of extension methods on this.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -1,25 +1,120 @@
...
+            VerificationContext(
+              deliveryConfigByName(deliveryConfigName),
+              environmentName,
+              artifactReference,
+              artifactVersion
+            )
+          }
+      }
+    }
+  }
```
robfletcher, line 117: This query is not quite as horrible as it looks at first glance. It also manages to avoid the need to pre-seed the "last checked" table whenever we create an environment (see the left join bit). I'd like to migrate that pattern to the similar queries we do elsewhere (separate PR).

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -64,22 +162,26 @@ class SqlVerificationRepository(
   private inline fun <reified RESULT> ResultQuery<*>.fetchOneInto() =
     fetchOneInto(RESULT::class.java)
 
+  /**
+   * Defaults a timestamp field to [EPOCH] if it is null.
+   */
+  private fun Field<LocalDateTime>.orEpoch(): Field<LocalDateTime> =
+    isnull(this, EPOCH.atZone(UTC).toLocalDateTime())
```
robfletcher, line 169: This is what makes the left join to the "last checked" table work. If the timestamp is `null` (because there is no corresponding row in the last checked table) then we get the default `EPOCH` meaning anything that was never checked will order first and definitely be older than the cutoff.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -1,25 +1,122 @@
...
+          .join(DELIVERY_ARTIFACT)
+          .on(DELIVERY_ARTIFACT.UID.eq(ENVIRONMENT_ARTIFACT_VERSIONS.ARTIFACT_UID))
+          // left join so we get results even if there is no row in ENVIRONMENT_LAST_VERIFIED
+          .leftJoin(ENVIRONMENT_LAST_VERIFIED)
+          .on(ENVIRONMENT_LAST_VERIFIED.ENVIRONMENT_UID.eq(ENVIRONMENT.UID))
+          .and(ENVIRONMENT_LAST_VERIFIED.ARTIFACT_UID.eq(DELIVERY_ARTIFACT.UID))
+          // has not been checked recently (or has never been checked)
+          .where(ENVIRONMENT_LAST_VERIFIED.AT.orEpoch().lessOrEqual(cutoff))
+          // order by last time checked with things never checked coming first
+          .orderBy(ENVIRONMENT_LAST_VERIFIED.AT.orEpoch())
```
lorin, line 97: I'm having trouble wrapping my head around this `orderBy` call. What is the SQL that it translates to?

robfletcher: ```order by ifnull(environment_last_verified.at, EPOCH)```

so basically if the left join to `environment_last_verified` did not find a row (hence `at` will be null) it uses epoch which will sort first.

lorin (approved):
# My understanding

## Problem 

In order to support the verifyWith functionality, keel needs to ensure that verifications specified in delivery configs run against the associated environment.

## Proposed solution

This PR implements the top-level verification loop that is responsible for invoking th VerificationRunner to kick off the verifications. 

This loops iterates over all of the environments being managed, looking for (artifact, environment) pairs that haven't been checked in the last 60 seconds (configurable).

The loop runs every 1 second (configurable), and retrieves a batch size of 1 (configurable) verifications that are old enough that they need checking.

The loop processes environments in order of "last checked" time, from most ancient to most recent.

# Database changes

## New table: `environment_last_verified`

Table tracks the last time the verification loop retrieved the (environment, artifact) pair.

* `environment_uid`
* `artifact_uid`
* `timestamp`

# Code changes

## Verification loop

The `CheckScheduler.verifyEnvironments` function.

## Database-related

The main change is the new: `SqlVerificationRepository.nextEnvironmentsForVerification` method.

## Refactor

Some DB methods were pulled into a new package: `com.netflix.spinnaker.keel.sql.deliveryconfigs`

* deliveryConfigByName
* attachDependents
* resourcesForEnvironment

## Minor changes

`VerificationContext` has a new `artifactReference` field.

I couldn't tell if this field is consumed by code yet, or is being added in anticipation of being consumed in a future PR.

# Feedback

Looks good overall. The only comment I have is around multiple artifact behavior.

## Multiple artifact behavior

This implementation uses (artifact, environment) pairs as the "thing we validated recently enough"
This results in a scenario that seem counter-intuitive to me when there are multiple artifacts. Consider:

* one verification: V
* Two artifacts: A, B
* One environment: staging

The same verification will run twice in succession against the same environment:

t=0s: (A, staging) hasn't been run lately, run V against staging
t=1s: (B, staging) hasn't been run lately, run V against staging

On the other hand, if keel pushes a new artifact version or makes another type of change (e.g., instance type, baseos) to the environment,
the verification check won't happen until the minimum amount of time elapses.

The weirdness here in my mind is around the relationship between a Verification and an Artifact. Right now we don't have an explicit relation between these two entities in our spec for the delivery config. Should we? 

That being said, I think multiple artifacts is not going to be the common case for the initial adopters, and so we can probably put it off.

robfletcher:
> `VerificationContext` has a new `artifactReference` field.
> 
> I couldn't tell if this field is consumed by code yet, or is being added in anticipation of being consumed in a future PR.

It's used by `VerificationContext` itself in its `artifact` property.

> This implementation uses (artifact, environment) pairs as the "thing we validated recently enough"
> This results in a scenario that seem counter-intuitive to me when there are multiple artifacts. Consider:
> 
>     * one verification: V
>     * Two artifacts: A, B
>     * One environment: staging
> 
> The same verification will run twice in succession against the same environment:
> 
> t=0s: (A, staging) hasn't been run lately, run V against staging
> t=1s: (B, staging) hasn't been run lately, run V against staging
> 

Oh boy, you're right and I'm not sure how to handle that. I guess we'd want to insert/update _n_ rows to `environment_last_verified` -- one per artifact. Then if _any_ of the artifact versions changed we'd run the verification once.

That would also require `VerificationContext` to change such that it had a collection of artifact versions.

lorin:
> Oh boy, you're right and I'm not sure how to handle that. I guess we'd want to insert/update _n_ rows to `environment_last_verified` -- one per artifact. Then if _any_ of the artifact versions changed we'd run the verification once.

That reminds me, another issue we'll (eventually) need to deal with is identifying when a resource changes that isn't related to an artifact version changing. (e.g., new version of baseos, change of instance type, addition of security group). We want to re-run the verification when that happens.

robfletcher:
> That reminds me, another issue we'll (eventually) need to deal with is identifying when a resource changes that isn't related to an artifact version changing. (e.g., new version of baseos, change of instance type, addition of security group). We want to re-run the verification when that happens.

Yeah, that makes the situation even worse. It almost feels like we'd need to fingerprint the environment in order to know whether we've already run the verification against the current configuration.

luispollo:
Holy smokes, @lorin! What a review! Now I understand why you asked me the other day how I was so quick to submit a review: I'm obviously not spending anywhere near the amount of time and attention you are on them! 😄 

Looking at this particular PR, I can see that your approach paid off as you caught something meaty! Having said that, I think there's an argument for both kinds of reviews, because statistically the number of severe issues caught in reviews is relatively small. Where I think they help the most is in catching major disagreements on the approach or design, as well as the organization and readability of the code, which are much quicker to review and comment on, in comparison. That's the stuff I tend to focus on.

I'm delighted that we have someone like you on our team because you elevate any activity you participate in, including code reviews! I like to think that the different approaches each of us takes in reviewing PRs complement each other well.

robfletcher:
> * one verification: V
> * Two artifacts: A, B
> * One environment: staging

Thinking a little more about this… is running the verification twice _wrong_? It's maybe overkill if you synchronize releases of the two artifacts, but if they're deployed independently (and currently MD has no mechanism to prevent that) then it makes sense to run the verification on _any_ deployment.

robfletcher:
> The weirdness here in my mind is around the relationship between a Verification and an Artifact. Right now we don't have an explicit relation between these two entities in our spec for the delivery config. Should we?

I'm not sure. In my mind the relationships are between verification and environment, and artifact and environment. There's not a direct relationship between verification and artifact.

lorin:
> Thinking a little more about this… is running the verification twice _wrong_? It's maybe overkill if you synchronize releases of the two artifacts, but if they're deployed independently (and currently MD has no mechanism to prevent that) then it makes sense to run the verification on _any_ deployment.

I agree that it's safe to run the verification twice, but I don't think the current implementation actually has the desired behavior of resetting the min age check on a new deployment. The current implementation uses (artifact, environment) pairs, and I think you'd need (artifact **version**, environment) pairs to achieve that behavior.

PR #1682, opened 2020-12-02
feat(lifecycle): refactor start monitoring events

emjburns:
After adding a second type of event I realized that it is a lot easier to use if the event can indicate whether it kicks off monitoring. I added a `monitor: Boolean` field to indicate whether or not the event should kick off a monitoring cycle. Follow-up events need to set this to `false`. 

I also refactored the publishing event functions in each monitor so that it was easy to enforce setting `monitor = false`.

No change in behavior except that the first `build` event is now of status `RUNNING`.

### DIFF
--- a/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
+++ b/keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
@@ -14,7 +14,7 @@ import com.netflix.spinnaker.keel.api.plugins.supporting
 import com.netflix.spinnaker.keel.exceptions.InvalidSystemStateException
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEvent
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEventScope.PRE_DEPLOYMENT
-import com.netflix.spinnaker.keel.lifecycle.LifecycleEventStatus.NOT_STARTED
+import com.netflix.spinnaker.keel.lifecycle.LifecycleEventStatus.RUNNING
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEventType.BUILD
 import com.netflix.spinnaker.keel.persistence.KeelRepository
 import com.netflix.spinnaker.keel.telemetry.ArtifactVersionUpdated
@@ -80,12 +80,9 @@ class ArtifactListener(
   /**
    * Finds the delivery configs that are using an artifact,
    * and publishes a build lifecycle event for them.
-   *
-   * todo eb: If the status is complete, publish a NOT_STARTED and a completed event
-   *  maybe add a "monitor me" flag? that seems better
    */
-  fun createBuildLifecycleEvent(artifact: PublishedArtifact) {
-    log.debug("Publishing build lifecycle event for artifact $artifact")
+  fun publishBuildLifecycleEvent(artifact: PublishedArtifact) {
+    log.debug("Publishing build lifecycle event for published artifact $artifact")
     artifact.buildMetadata
       ?.let { buildMetadata ->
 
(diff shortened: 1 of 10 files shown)


keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -108,11 +105,15 @@ class ArtifactListener(
                 artifactVersion = artifact.version,
                 type = BUILD,
                 id = "build-${artifact.version}",
-                status = NOT_STARTED,
+                // the build has already started, and is maybe complete.
+                // We use running to convey that to users, and allow the [BuildLifecycleMonitor] to immediately
+                // update the status
```
gal-yardeni, line 110: Can you explain a bit more on how the build monitor will update the status immediately if the build completed?

emjburns: The build monitor does a status check immediately after the event published (that's just how it works because the event has never been monitored). When it does it'll figure out what the actual status is and then publish another lifecycle event with the correct status and user link.

gal-yardeni: I see the logic there, but as we discussed in your previous PR, the status can be terminal here. Isn't it worth passing here the complete build metadata object, and then determinate if to check the status again?

emjburns: Alright, I made that change!

gal-yardeni (approved):
💯  thank you for addressing my comments! this is so great.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
```diff
@@ -80,26 +80,23 @@ class ArtifactListener(
   /**
    * Finds the delivery configs that are using an artifact,
    * and publishes a build lifecycle event for them.
-   *
-   * todo eb: If the status is complete, publish a NOT_STARTED and a completed event
-   *  maybe add a "monitor me" flag? that seems better
    */
-  fun createBuildLifecycleEvent(artifact: PublishedArtifact) {
-    log.debug("Publishing build lifecycle event for artifact $artifact")
+  fun publishBuildLifecycleEvent(artifact: PublishedArtifact) {
```
luispollo, line 84: Thank you for changing the name! ❤️

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
```diff
@@ -21,6 +21,8 @@ import java.time.Instant
  *   the user.
  * [timestamp] is nullable so that the repository can insert the timestamp when
  *   the event is stored.
+ * [monitor] true if this event is the first in the chain and needs to be monitored.
```
luispollo, line 24: There's _a lot_ of detail encoded in this name. How about `isMonitoringTrigger` instead, for a bit more context?

luispollo (approved):
Thanks for the quick follow-up with this refactor, @emjburns. Nice improvements! I only had one suggestion on naming, not blocking but hope you'll consider it.

keel-artifact/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactListener.kt
emjburns: I told you I would! :D

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEvent.kt
emjburns: how about: `startMonitoring: Boolean`?

luispollo: Sure!

PR #1684, opened 2020-12-03
fix(sql): Cascade deletion of delivery config to event and paused records

### LINKED ISSUE #1467: Deleting delivery config does not cleanup related records
While testing with a delivery config I had to delete/recreate, I noticed the corresponding records from the following tables are not deleted:
- `artifact_versions`
- `diff_fingerprint`
- `event` (I forget if this was by design or not)
- `resource_last_checked`
- `unhappy_veto`

Cc: @emjburns

luispollo:
Complements #1660 to close #1467.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
@@ -87,34 +87,6 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
       from = ArtifactOriginFilterSpec(branch = BranchFilterSpec(name = "main"))
     )
 
-    val completeDeliveryConfig: DeliveryConfig = deliveryConfig.copy(
-      artifacts = setOf(
-        artifact, artifactFromBranch
-      ),
-      environments = setOf(
-        Environment(
-          name = "test",
-          resources = setOf(
-            resource(kind = parseKind("ec2/cluster@v1")),
-            resource(kind = parseKind("ec2/security-group@v1"))
-          )
-        ),
-        Environment(
-          name = "staging",
-          constraints = setOf(
-            DependsOnConstraint(
-              environment = "test"
-            ),
-            ManualJudgementConstraint()
-          ),
-          resources = setOf(
-            resource(kind = parseKind("ec2/cluster@v1")),
-            resource(kind = parseKind("ec2/security-group@v1"))
-          )
-        )
-      )
-    )
-
     fun getByName() = expectCatching {
       repository.get(deliveryConfig.name)
     }
@@ -238,7 +210,33 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
     context("storing a delivery config with artifacts and environments") {
       deriveFixture {
         copy(
-          deliveryConfig = completeDeliveryConfig
+          deliveryConfig = deliveryConfig.copy(
+            artifacts = setOf(
+              artifact, artifactFromBranch
+            ),
+            environments = setOf(
+              Environment(
+                name = "test",
+                resources = setOf(
+                  resource(kind = parseKind("ec2/cluster@v1")),
+                  resource(kind = parseKind("ec2/security-group@v1"))
+                )
+              ),
+              Environment(
+                name = "staging",
+                constraints = setOf(
+                  DependsOnConstraint(
+                    environment = "test"
+                  ),
+                  ManualJudgementConstraint()
+                ),
+                resources = setOf(
+                  resource(kind = parseKind("ec2/cluster@v1")),
+                  resource(kind = parseKind("ec2/security-group@v1"))
+                )
+              )
+            )
+          )
         )
       }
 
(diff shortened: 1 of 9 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/PersistentEvent.kt
```diff
@@ -29,7 +29,7 @@ abstract class PersistentEvent {
     val clock: Clock = Clock.systemUTC()
   }
 
-  enum class Scope {
+  enum class EventScope {
```
luispollo, line 32: Renamed this to avoid confusion with other "scopes" in the codebase.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -184,6 +210,15 @@ class SqlDeliveryConfigRepository(
...
     }
 
+  private fun getResourceIDs(application: String): List<String> =
+    sqlRetry.withRetry(READ) {
+      jooq
+        .select(RESOURCE.ID)
+        .from(RESOURCE)
+        .where(RESOURCE.APPLICATION.eq(application))
+        .fetch(RESOURCE.ID)
+    }
```
robfletcher, line 220: Seems to me it would be pretty easy to inline this into the queries and delete everything at once rather than having to do a `forEach` over all the resources

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -201,33 +238,7 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
...
-                  ManualJudgementConstraint()
-                ),
-                resources = setOf(
-                  resource(kind = parseKind("ec2/cluster@v1")),
-                  resource(kind = parseKind("ec2/security-group@v1"))
-                )
-              )
-            )
-          )
+          deliveryConfig = completeDeliveryConfig
```
robfletcher, line 241: What's the thinking behind making this a top-level val? It seems to only be used here.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: Ya, good point. I just got lazy and followed the pattern. Will change it!

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
luispollo: I kept this change by accident. Was planning on using that elsewhere and never did. Will undo.

keel-sql/src/main/resources/db/changelog/20201203-cleanup-event-and-paused-orhpans.yml
```diff
@@ -0,0 +1,25 @@
+databaseChangeLog:
+- changeSet:
```
luispollo, line 2: Cleans up any existing orphan records from the database.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -166,31 +188,28 @@ class SqlDeliveryConfigRepository(
         .fetchOne(DELIVERY_CONFIG.UID)
     } ?: throw NoDeliveryConfigForApplication(application)
 
-  private fun getEnvironmentUIDs(deliveryConfigUIDs: List<String>): List<String> =
+  private fun getResourceUIDs(application: String): SelectConditionStep<Record1<String>>? =
     sqlRetry.withRetry(READ) {
```
robfletcher, line 192: It doesn't make sense to have a retry in here if this is returning the query itself rather than executing it

luispollo: True. Will fix.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -166,31 +188,28 @@ class SqlDeliveryConfigRepository(
...
       jooq
-        .select(ENVIRONMENT.UID)
```
robfletcher, line 172: You also don't need to call this on a `DSLContext` instance, there's a static function `DSL.select` you can use when you want to create a sub-query like this.

luispollo: Yeah, but it doesn't hurt, does it?

luispollo:
@robfletcher I think I've addressed your feedback. Anything else you'd like to see changed before approving?

robfletcher (approved):

PR #1686, opened 2020-12-04
fix(lifecycle): don't copy time forward

emjburns:
Fix for the bakery lifecycle monitor to not copy time forward.

Previously, this didn't matter, because we didn't care about the time an event added. But, when I updated that logic so the build monitor could actually pass in the correct timing I didn't realize it would cause the bakery monitor to continually pass forward the start time.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
@@ -123,7 +123,8 @@ class BakeryLifecycleMonitor(
       status = status,
       link = orcaTaskIdToLink(this),
       text = text,
-      startMonitoring = false
+      startMonitoring = false,
+      timestamp = null // let repository record the current time
     ))
   }
 }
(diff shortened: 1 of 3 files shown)


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/LifecycleEventRepositoryTests.kt
```diff
@@ -131,5 +132,27 @@ abstract class LifecycleEventRepositoryTests<T: LifecycleEventRepository> : JUni
         }
       }
     }
```
emjburns, line 134: This is a related test, but wouldn't have caught the problem But I added it because it seems worth it to test.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -123,7 +123,8 @@ class BakeryLifecycleMonitor(
       status = status,
       link = orcaTaskIdToLink(this),
       text = text,
-      startMonitoring = false
+      startMonitoring = false,
+      timestamp = null // let repository record the current time
```
robfletcher, line 127: Not sure I understand the practical difference here

emjburns: Otherwise we were copying forward the start time of the event. This lets the timestamp be calculated when the event is saved into the database

PR #1690, opened 2020-12-05
docs(api): Re-enable Swagger UI

luispollo:
Re-enable the Swagger UI by adding a dependency to springfox.

### DIFF
--- a/keel-web/keel-web.gradle.kts
+++ b/keel-web/keel-web.gradle.kts
@@ -40,6 +40,7 @@ dependencies {
     exclude(mapOf("group" to "com.netflix.spinnaker.kork", "module" to "kork-secrets-aws"))
     exclude(mapOf("group" to "com.netflix.spinnaker.kork", "module" to "kork-secrets-gcp"))
   }
+  runtimeOnly("io.springfox:springfox-boot-starter:3.0.0")
 
   testImplementation("io.strikt:strikt-jackson")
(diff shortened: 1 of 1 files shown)


lorin (approved):
# My understanding

## Problem

Not all keel REST API calls are exposed via gate. As operators, we sometimes need to make REST calls directly against keel. Today, the only way for us to figure out the arguments for the REST call (path, args) is to look through the keel controller code.

## Implemented solution

Use [SpringFox][1] to automatically generate Swagger documentation for the REST APIs that keel exposes.

# Feedback

Looks good to me: it looks like there should be no behavioral changes to keel here other than exposing the Swagger stuff.

## Questions
Do we need to make changes to gate to serve the swagger documentation? Or will we access it by hitting keel directly?

[1]: https://springfox.github.io/springfox/

luispollo:
> Do we need to make changes to gate to serve the swagger documentation? Or will we access it by hitting keel directly?

I had previously integrated the two in https://github.com/spinnaker/gate/pull/1211, but when @robfletcher removed Swagger from keel in favor of exposing only the delivery config schema via the new `GET /managed/delivery-configs/schema` API (plus the existing APIs exposed under the `/managed` path), he also removed that integration from gate. I think no one outside keel contributors would have an interest in the APIs that are not exposed through gate, so my suggestion would be to just access the docs directly from keel when needed.

PR #1691, opened 2020-12-07
fix(builds): calculate steps more correctly

emjburns:
(1) correctly calculates build steps: I changed the requirement to have the first event be NOT_STARTED, then I forgot to actually make that true. This PR fixes that.

(2) changes duplicate events to update the timestamp instead of saving a new event. Why? Last week I had a bug in my code where we never stopped monitoring failed builds. We got a LOT of events saved about this. It seems useless. Really, I just want to know the last seen time for the status. So, I made the saving this way.

(3) I put a retry around grabbing the event uuid for the foreign key.

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleEventRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleEventRepository.kt
@@ -41,7 +41,7 @@ class SqlLifecycleEventRepository(
       jooq.transaction { config ->
         val txn = DSL.using(config)
         // if event exists, only update timestamp
-        var uid: String? = null
+        var eventExists = false
         txn.select(LIFECYCLE_EVENT.UID, LIFECYCLE_EVENT.JSON)
           .from(LIFECYCLE_EVENT)
           .where(LIFECYCLE_EVENT.SCOPE.eq(event.scope.name))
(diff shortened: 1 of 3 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleEventRepository.kt
```diff
@@ -30,20 +31,56 @@ class SqlLifecycleEventRepository(
...
+          .and(LIFECYCLE_EVENT.STATUS.eq(event.status.name))
+          .limit(1)
+          .fetch()
+          .map { (savedUid: String, savedEvent: String) ->
+            uid = savedUid
+            objectMapper.readValue<LifecycleEvent>(savedEvent)
+          }
+          .firstOrNull()
+          ?.let { existingEvent: LifecycleEvent ->
+            if (existingEvent.copy(timestamp = null) == event.copy(timestamp = null)) {
```
luispollo, line 61: You could simplify this like so:
```suggestion
            if (event == existingEvent.copy(timestamp = event.timestamp)) {
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleEventRepository.kt
```diff
@@ -30,20 +31,56 @@ class SqlLifecycleEventRepository(
...
+              // events are the same except time
+              txn.update(LIFECYCLE_EVENT)
+                .set(LIFECYCLE_EVENT.TIMESTAMP, timestamp)
+                .where(LIFECYCLE_EVENT.UID.eq(uid))
+                .execute()
+            }
+          }
+
+        if (uid == null) {
```
luispollo, line 70: I think this would be easier to read if instead of saving the `uid` in a `var`, you just saved `existingEvent`, then here you could write `if (existingEvent == null)`, which is more self-explanatory, IMHO.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleMonitorRepository.kt
```diff
@@ -78,20 +78,23 @@ class SqlLifecycleMonitorRepository(
     }
 
   fun LifecycleEvent.getUid(jooq: DSLContext): String {
```
luispollo, line 80: This is an extension function, so you don't have to pass in `jooq`...

luispollo (approved):
Minor suggestions, otherwise LGTM.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleMonitorRepository.kt
emjburns: Oh nice! :D

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleEventRepository.kt
emjburns: yes nice, that is better

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleEventRepository.kt
emjburns: Okay, let me rethink the flow a little bit to try to be clearer

emjburns: Alright, done!

PR #1692, opened 2020-12-08
fix(ec2): gracefully handle dangling security group ref

lorin:
## Context

Every AWS security group specifies a set of *inbound rules*. There are different types of inbound rules. One particular type, "a security group rule reference", refers to a second security group. Essentially, it says "Allow instances in that security group to talk to me."

## Problem

Generally, AWS does not allow a user to delete a security group if other security groups refer to it in an inbound rule. However, there are edge cases where this can happen. As a result, a security group might have an inbound rule that refers to another security group that no longer exists. I call this a *dangling SG reference*.

If you query clouddriver for information about a security group that has a dangling SG reference, the referenced security group will have a missing "name" field in the payload, for example:

```javascript
{
 "class": "com.netflix.spinnaker.clouddriver.aws.model.AmazonSecurityGroup",
 "id": "sg-1234567890123",
 "name": "foo-bar",
  ...
  "inboundRules": [
    {
      "class": "com.netflix.spinnaker.clouddriver.model.securitygroups.SecurityGroupRule",
      "securityGroup": {
        "class": "com.netflix.spinnaker.clouddriver.aws.model.AmazonSecurityGroup",
	"id": "sg-xxxxxxxxxxxxxxxxx" // <--- sg that no longer exists
	// Missing "name" field that would normally go here
```

When keel tries to deserialize the above object, it throws an exception because the `name` field is required:

```
Instantiation of [simple type, class com.netflix.spinnaker.keel.clouddriver.model.SecurityGroupModel$SecurityGroupRuleReference] value failed for JSON property name due to missing (therefore NULL) value for creator parameter name which is a non-nullable type
 at [Source: (okhttp3.ResponseBody$BomAwareReader); line: 1, column: 1317] (through reference chain: com.netflix.spinnaker.keel.clouddriver.model.SecurityGroupModel["inboundRules"]->java.util.HashSet[1]->com.netflix.spinnaker.keel.clouddriver.model.SecurityGroupModel$SecurityGroupRule["securityGroup"]->com.netflix.spinnaker.keel.clouddriver.model.SecurityGroupModel$SecurityGroupRuleReference["name"])
```

## Proposed solution

Allow the `SecurityGroupRuleReference.name` field to be null to avoid the deserialization error.

In the SecurityGroupHandler logic, when converting from `com.netflix.spinnaker.keel.clouddriver.model.SecurityGroupModel` to `com.netflix.spinnaker.keel.api.ec2.SecurityGroup`, if the name field is null, then omit that inbound rule.

This should not affect behavior since an inbound rule that points to a non-existent security group has no effect.

We also now log a warning when encountering an inbound rule to a non-existent security group. This log includes the invalid security group id, to make it easier for an operator to find this rule if they wanted to look for it.


emjburns (approved):

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -275,6 +275,19 @@ class SecurityGroupHandler(
...
         val ingressRange = rule.range
         val protocol = rule.protocol!!.clouddriverProtocolToKeel()
+
+        /**
+         * Due to edge cases in AWS, it is possible for an inbound rule to reference a security group
+         * that no longer exists. When this happens, the clouddriver response won't have a [name]
+         * field. We silently ignore these inbound rules, since they have no effect.
+         */
+        ingressGroup?.let { group ->
+          if(group.name == null) {
```
luispollo, line 285: nit:
```suggestion
          if (group.name == null) {
```

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -275,6 +275,19 @@ class SecurityGroupHandler(
...
+            log.warn("security group $name ($accountName, $region) has inbound rule that references non-existent security group ${group.id} (${group.accountName}, ${group.region}, ${group.vpcId})")
+            return@flatMap emptyList()
```
luispollo, line 287: Isn't this going to cause us to ignore all other ingress rules as well? I thought the idea was just to skip this particular one (which you could do with a `.filter`)?

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt
```diff
@@ -391,6 +392,44 @@ internal class SecurityGroupHandlerTests : JUnit5Minutests {
           .hasSize(cloudDriverResponse1.inboundRules.size + 1)
       }
     }
+
+    context("only inbound rule points to a non-existent security group") {
```
luispollo, line 396: As I mentioned above, I think if you create a response with more than 1 inbound rule, where one of them has a missing name for a referenced SG, then we'll ignore all other rules with your current implementation, which doesn't seem correct.

luispollo (changes requested):
Please see inline comments.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
lorin: I agree that `.filter` is clearer, but I didn't go with it here because I wanted to log the warning, and I didn't see a nice way to combine the two.

Note that the`return@flatMap` call does not short-circuit processing of elements in the `inboundRules` list: it returns from one block iteration, but the block still gets called for subsequent elements of the list.

I was a little confused about this behavior as well, but I confirmed it with a simple example. Check out this [Kotlin playground snippet](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS40LjAiLCJjb2RlIjoiLyoqXG4gKiBGaWx0ZXIgb3V0IHRoZSBldmVuIG51bWJlcnNcbiAqL1xuZnVuIG9kZHMobHN0IDogTGlzdDxJbnQ+KSA9IFxuICAgIGxzdC5mbGF0TWFwIHsgbiAtPlxuICAgICAgXG4gICAgICAgIGlmKG4gJSAyID09IDApIHsgXG4gICAgICAgICAgICByZXR1cm5AZmxhdE1hcCBlbXB0eUxpc3QoKVxuICAgICAgICB9IFxuXG4gICAgICAgIGxpc3RPZihuKVxuICAgIH1cbiBcbmZ1biBtYWluKCkge1xuICAgIHByaW50bG4ob2RkcyhsaXN0T2YoMSwyLDMpKSlcbn0iLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIn0=) I wrote to confirm the behavior.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt
lorin: @luispollo I updated the tests to try to show that the behavior doesn't skip subsequent rules. However, because the underlying collection is a set and not a list, the test doesn't guarantee the behavior is correct (I could just be getting lucky). But my other comment had a link to some code to demonstrate that the other rules don't get ignored.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
luispollo: Thanks for checking! I still think this would be easier to read without the return. Something like:
```kotlin
inboundRules = inboundRules.flatMap { rule ->
  val ingressGroup = rule.securityGroup
  // if the ingress rule references another security group, the the referenced SG's name is null, ignore it with a warning
  if (ingressGroup != null && ingressGroup.name == null) {
    log.warn(...)
    null
  } else {
    // ...
  }
}.filterNotNull()

File: keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt (hunk 3 above)
Comment (luispollo) [reply keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt]: You could use `sortedSetOf`.

**Review by luispollo — APPROVED** [review]
Just a suggestion on form, but nothing blocking.

File: keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt (hunk 3 above)
Comment (lorin) [reply keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt]: TIL. Updated.

Comment (lorin) [reply keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt]: Unfortunately, `sortedSetOf` didn't do what I thought it would. I thought it would preserve insertion ordering. Instead, it sorts the inputs (which should have been obvious based on the name). Since the security group rules aren't comparable, they can't be stored in a sorted set.

File: keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt (hunk 2 above)
Comment (lorin) [reply keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt]: I refactored the code to improve readability.

File: keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt (hunk 3 above)
Comment (luispollo) [reply keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt]: Yeah, sorry, I didn't think to mentioned it, but there's a version of the function that takes a `Comparator` as a parameter, which is what I assumed you would use.

File: keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt [hunk 4]
```diff
@@ -305,6 +305,15 @@ class SecurityGroupHandler(
                   ingressRange.ip + ingressRange.cidr
                 )
               } ?: emptyList()
+          ingressGroup != null && ingressGroup.name == null -> {
```
luispollo, line 308: TIL. I hadn't seen this form of a lambda before.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt
lorin: Unfortunately, I'm not as smart as you think I am. 😝

luispollo: Yeah, right! 😆

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandlerTests.kt
```diff
@@ -393,17 +393,31 @@ internal class SecurityGroupHandlerTests : JUnit5Minutests {
       }
     }
 
-    context("only inbound rule points to a non-existent security group") {
+    context("one of the three inbound rules points to a non-existent security group") {
 
       deriveFixture {
         copy(
           cloudDriverResponse1 = cloudDriverResponse1.copy(
             inboundRules = setOf(
```
luispollo, line 401: If you want to go back to it later, this should work:
```suggestion
            inboundRules = sortedSetOf(
              compareBy { it.securityGroup!!.id },
```
...and then you change the IDs down below to make sure they have the order you like.

PR #1694, opened 2020-12-08
add missing cascade to foreign keys on environment_last_verified table

robfletcher:
Missing cascade causes deletion of artifacts / environment elsewhere to fail.

### DIFF
--- a/keel-sql/src/main/resources/db/changelog/20201208-fix-environment-last-verified-cascade.yml
+++ b/keel-sql/src/main/resources/db/changelog/20201208-fix-environment-last-verified-cascade.yml
@@ -1,0 +1,27 @@
+databaseChangeLog:
+- changeSet:
+    id: fix-environment-last-verified-cascade
+    author: fletch
+    changes:
+    - dropForeignKeyConstraint:
+        baseTableName: environment_last_verified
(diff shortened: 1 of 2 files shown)


emjburns (approved):
Approving but I don't know enough about what the actual problem is to verify that this fixes it :D

keel-sql/src/main/resources/db/changelog/20201208-fix-environment-last-verified-cascade.yml
```diff
@@ -0,0 +1,27 @@
...
+        baseTableName: environment_last_verified
+        constraintName: fk_environment_last_verified_environment
+    - addForeignKeyConstraint:
+        baseTableName: environment_last_verified
+        baseColumnNames: environment_uid
+        constraintName: fk_environment_last_verified_environment
+        referencedTableName: environment
+        referencedColumnNames: uid
+        referencesUniqueColumn: true
+        onDelete: CASCADE
```
robfletcher, line 16: @emjburns this is the important bit, without the cascade the foreign key prevents us deleting rows from `delivery_artifact` or `environment` that are referenced from this table. With the cascade in place the rows on this table will get deleted automatically when the parent row is deleted.

lorin (approved):

PR #1695, opened 2020-12-08
fix(pipeline): only run if version is deployed

emjburns:
We should only run our hacky pipeline constraint if the version it's evaluating is actually deployed. otherwise, fail the constraint.

### DIFF
--- a/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluator.kt
+++ b/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluator.kt
@@ -15,6 +15,7 @@ import com.netflix.spinnaker.keel.core.api.PipelineConstraint
 import com.netflix.spinnaker.keel.model.toOrcaNotification
 import com.netflix.spinnaker.keel.orca.OrcaExecutionStatus
 import com.netflix.spinnaker.keel.orca.OrcaService
+import com.netflix.spinnaker.keel.persistence.KeelRepository
 import java.time.Clock
 import java.util.HashMap
 import kotlinx.coroutines.runBlocking
@@ -43,7 +44,8 @@ class PipelineConstraintEvaluator(
   private val orcaService: OrcaService,
   repository: ConstraintRepository,
   override val eventPublisher: EventPublisher,
-  private val clock: Clock
+  private val clock: Clock,
+  private val keelRepository: KeelRepository
 ) : StatefulConstraintEvaluator<PipelineConstraint, PipelineConstraintStateAttributes>(repository) {
   override val supportedType = SupportedConstraintType<PipelineConstraint>("pipeline")
   override val attributeType = SupportedConstraintAttributesType<PipelineConstraintStateAttributes>("pipeline")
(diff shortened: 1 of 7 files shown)


keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluator.kt
```diff
@@ -67,6 +69,23 @@ class PipelineConstraintEvaluator(
...
+          state.copy(
+            status = FAIL,
+            comment = "Pipeline constraint wants to run against $version, " +
+              "but the current version in env ${targetEnvironment.name} is $currentVersion.",
+            judgedAt = clock.instant(),
+            judgedBy = "keel"
+          )
+        )
+      return false
+    }
```
emjburns, line 87: @lorin should we fail the constraint here? I thought it was an ok thing to do, since this pipeline constraint will get replaced soon. But I'm not sure.

lorin (approved):
# My understanding

## The problem

We had an operational surprise where the pipeline constraint kept running against older versions that were not currently deployed in the environment.

While the pipeline execution behavior was not the only factor in the surprise (it was related to the behavior of admin endpoint logic), this behavior increased the blast radius of the problem.

## Proposed solution

When the pipeline constraint evaluator fires, check to see if the `CURRENT` artifact version in the environment is the same as the artifact version associated with the constraint.

If this is not the case, then:
* do not execute the pipeline
* mark the constraint as having failed

# My feedback

This looks like a reasonable solution to me. Since the `ConstraintState.attributes` field remains unchanged when this happens, I don't see any unwanted side effects resulting from this change.

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluator.kt
```diff
@@ -67,6 +69,23 @@ class PipelineConstraintEvaluator(
 
     val judge = "pipeline:${constraint.pipelineId}"
 
+    val currentVersion = keelRepository
+      .getCurrentVersionInEnv(deliveryConfig, targetEnvironment.name, artifact.reference)
+
+    if (currentVersion != version) {
+      repository
```
lorin, line 76: I'd add a log.warn statement here, since we don't typically expect this to happen.

keel-orca/src/test/kotlin/com/netflix/spinnaker/keel/constraints/PipelineConstraintEvaluatorTests.kt
```diff
@@ -252,7 +286,7 @@ internal class PipelineConstraintEvaluatorTests : JUnit5Minutests {
     }
   }
 
-  fun MockKAnswerScope<*, *>.canaryConstraintState(
+  fun MockKAnswerScope<*, *>.pipelineConstraintState(
```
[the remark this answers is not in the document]
emjburns: Ha yes.

luispollo:
@emjburns Can we close this now that we've replaced our pipeline constraint with `verifyWith`?

luispollo:
> @emjburns Can we close this now that we've replaced our pipeline constraint with `verifyWith`?

Closing this in the absence of a response. Feel free to reopen if needed.

PR #1698, opened 2020-12-08
fix(lifecycle): start monitoring kicked off after saving event

emjburns:
**Problem**
We currently have two listeners for lifecycle events: one that saves the event (this is the "triggering event"), and one that saves the monitor instructions for that event. We're seeing some instances where the triggering event has not been saved when we try to monitor it. Because there is a foreign keys from the monitor table to the lifecycle event table, this causes problems. 

**Proposed solution**
I've changed the behavior slightly in this pr so that the lifecycle event is saved, and then a "start monitoring event" is published with that event's uid (which will be the foreign key in the monitoring table). This guarantees that the saving of the monitor event will happen only once the original event is saved. However, I've sort of 'leaked' the fact that we save the event with a uuid so that it's easier to save the monitor event. I feel this is fine but not great, and I'm happy to rethink it if anyone has suggestions.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventHandler.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventHandler.kt
@@ -1,18 +1,27 @@
 package com.netflix.spinnaker.keel.lifecycle
 
+import org.springframework.context.ApplicationEventPublisher
 import org.springframework.context.event.EventListener
 import org.springframework.stereotype.Component
 
 /**
- * Receives lifecycle events and stores them
+ * Receives lifecycle events and stores them.
  */
 @Component
 class LifecycleEventHandler(
-  val repository: LifecycleEventRepository
+  val repository: LifecycleEventRepository,
+  val publisher: ApplicationEventPublisher
 ) {
 
+  /**
+   * Saves all lifecycle events.
+   * If they should be monitored, publish a [StartMonitoringEvent]
+   */
   @EventListener(LifecycleEvent::class)
   fun handleEvent(event: LifecycleEvent) {
-    repository.saveEvent(event)
+    val eventId = repository.saveEvent(event)
+    if (event.startMonitoring) {
+      publisher.publishEvent(StartMonitoringEvent(eventId, event))
+    }
   }
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/MonitoredTask.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/MonitoredTask.kt
@@ -15,6 +15,7 @@ package com.netflix.spinnaker.keel.lifecycle
 data class MonitoredTask(
   val triggeringEvent: LifecycleEvent,
   val link: String,
+  val triggeringEventUid: String,
   val type: LifecycleEventType = triggeringEvent.type,
   val numFailures: Int = 0
 )
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleMonitorRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleMonitorRepository.kt
@@ -1,11 +1,13 @@
 package com.netflix.spinnaker.keel.sql
 
+import com.fasterxml.jackson.databind.JsonMappingException
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEvent
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEventType
 import com.netflix.spinnaker.keel.lifecycle.LifecycleMonitorRepository
 import com.netflix.spinnaker.keel.lifecycle.MonitoredTask
+import com.netflix.spinnaker.keel.lifecycle.StartMonitoringEvent
 import com.netflix.spinnaker.keel.persistence.metamodel.Tables.LIFECYCLE_EVENT
 import com.netflix.spinnaker.keel.persistence.metamodel.tables.LifecycleMonitor.LIFECYCLE_MONITOR
 import com.netflix.spinnaker.keel.sql.RetryCategory.READ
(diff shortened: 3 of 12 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/LifecycleEventHandler.kt
```diff
@@ -1,18 +1,27 @@
...
+   * If they should be monitored, publish a [StartMonitoringEvent]
+   */
   @EventListener(LifecycleEvent::class)
   fun handleEvent(event: LifecycleEvent) {
-    repository.saveEvent(event)
+    val eventId = repository.saveEvent(event)
+    if (event.startMonitoring) {
+      publisher.publishEvent(StartMonitoringEvent(eventId, event))
+    }
   }
```
emjburns, line 26: now we publish a `StartMonitoringEvent` to kick off monitoring instead of relying on another handler to also be listening to `LifecyceEvent`s

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/lifecycle/MonitoredTask.kt
```diff
@@ -15,6 +15,7 @@ package com.netflix.spinnaker.keel.lifecycle
 data class MonitoredTask(
   val triggeringEvent: LifecycleEvent,
   val link: String,
+  val triggeringEventUid: String,
   val type: LifecycleEventType = triggeringEvent.type,
   val numFailures: Int = 0
```
emjburns, line 20: We already store the triggering event uid, let's just return it here so we don't have to look it up again later!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlLifecycleMonitorRepository.kt
```diff
@@ -77,62 +80,43 @@ class SqlLifecycleMonitorRepository(
         }
     }
```
emjburns, line 82: No longer needed since we just look up the uid from the database.

lorin (approved):
## My understanding

### Context

In the current implementation, there are two separate event listeners for `LifecycleEvent` evnts:

1. LifecycleEventHandler.handleEvent
2. LifecycleMonitorScheduler.onLifecycleEvent

Both event handlers write to the database. The first one writes to the `lifecycle_event` table and the second to the `lifecycle_monitor` table.

### Problem 

The second handler first does a lookup in the `lifecycle_event` table to get the primary key associated with the lifecycle event, using information encoded in the LifecycleEvent to look up the record. However, because Spring does not enforce ordering of event listeners, if the second listener fires first, it will query the database for a record that does not exist yet. The result is an exception with a stack trace that looks like this:

```
java.util.NoSuchElementException: List is empty.
  at kotlin.collections.CollectionsKt___CollectionsKt.first(_Collections.kt:212)
  at com.netflix.spinnaker.keel.sql.SqlLifecycleMonitorRepository$getUid$1.invoke(SqlLifecycleMonitorRepository.kt:91)
  at com.netflix.spinnaker.keel.sql.SqlLifecycleMonitorRepository$getUid$1.invoke(SqlLifecycleMonitorRepository.kt:21)
  at com.netflix.spinnaker.keel.sql.SqlRetry$sam$java_util_function_Supplier$0.get(SqlRetry.kt:-1)
  at io.github.resilience4j.retry.Retry.lambda$decorateSupplier$2(Retry.java:177)
  at io.vavr.control.Try.of(Try.java:75)
  at io.vavr.control.Try.ofSupplier(Try.java:92)
  at com.netflix.spinnaker.keel.sql.SqlRetry.withRetry(SqlRetry.kt:35)
  at com.netflix.spinnaker.keel.sql.SqlLifecycleMonitorRepository.getUid(SqlLifecycleMonitorRepository.kt:81)
  at com.netflix.spinnaker.keel.sql.SqlLifecycleMonitorRepository$save$1.invoke(SqlLifecycleMonitorRepository.kt:97)
  at com.netflix.spinnaker.keel.sql.SqlLifecycleMonitorRepository$save$1.invoke(SqlLifecycleMonitorRepository.kt:21)
  at com.netflix.spinnaker.keel.sql.SqlRetry$sam$java_util_function_Supplier$0.get(SqlRetry.kt:-1)
  at io.github.resilience4j.retry.Retry.lambda$decorateSupplier$2(Retry.java:177)
  at io.vavr.control.Try.of(Try.java:75)
  at io.vavr.control.Try.ofSupplier(Try.java:92)
  at com.netflix.spinnaker.keel.sql.SqlRetry.withRetry(SqlRetry.kt:24)
  at com.netflix.spinnaker.keel.sql.SqlLifecycleMonitorRepository.save(SqlLifecycleMonitorRepository.kt:96)
  at com.netflix.spinnaker.keel.lifecycle.LifecycleMonitorScheduler.onLifecycleEvent(LifecycleMonitorScheduler.kt:59)
  ```

### Proposed solution

Enforcing ordering of the logic in the two listeners by using two different event types: the first listener emits a new event type, StartMonitoringEvent, that is consumed by the second listener.

#### New concepts

* StartMonitoringEvent, a new event class that is distinct from LifecycleEvent.
 

#### Other changes

MonitoredTask now stores the UID of the triggering event database entry.


## Feedback


I like the strategy of enforcing ordering of events by emitting different event types: it makes the code easier to reason about.
In particular, I don't see a problem with exposing the UID in the business logic.

### Comments


#### LifecycleEvent.id

The LifecycleEvent class currently has an `id` field, with different semantics. I can imagine a casual code reader getting confused between the id and the uid. Not sure what to do about this.

#### Naming

The StartMonitoringEvent class is named after the desired behavior when the event is fired. Typically we name events based on what happened, not what we want to happen as a result of the event. I can't think of a good name, though.


### Questions not related to changes in the PR

What happens if the instance gets terminated after the first database write but before the second one? It looks like in that case keel will fail to monitor the lifecycle task (but the task itself will still happen).

emjburns:
@lorin thank you for your detailed review. Re your comments:

- I think having `id` and `uid` is fine, not great but fine. 
- Interesting comment about naming. I also can't think of something better.
- If the instance gets terminated in the middle we won't monitor. This has the same result of the current behavior (i.e. when we get the list empty, we don't monitor). I'm thinking we might need a way to backfill or remediate based on the number of events we've seen that monitoring got dropped because of the bug this pr is fixing. In that case, this would help. However, I don't have a solution yet.

PR #1699, opened 2020-12-09
Test container verification

robfletcher:
Implementation of a verification that will run a Titus test job


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
```diff
@@ -10,12 +11,19 @@ interface VerificationEvaluator<VERIFICATION: Verification> {
...
   /**
+   * @param metadata as returned by [start].
    * @return the current status of the verification.
    */
-  fun evaluate() : VerificationStatus
+  fun evaluate(
+    context: VerificationContext,
+    verification: Verification,
+    metadata: Map<String, Any?>
+  ): VerificationStatus
```
robfletcher, line 21: Added parameters here so that the implementation has the context it needs

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
```diff
@@ -10,12 +11,19 @@ interface VerificationEvaluator<VERIFICATION: Verification> {
...
 
   /**
    * Start running [verification].
+   *
+   * @return any metadata needed to [evaluate] the verification in future.
    */
-  fun start(verification: Verification)
+  fun start(context: VerificationContext, verification: Verification): Map<String, Any?>
```
robfletcher, line 28: The `start` method can now return metadata (e.g. orca task id) that will get persisted and then passed to future calls to `evaluate`.

keel-sql/src/main/resources/db/changelog/20201208-add-verification-state-metadata.yml
```diff
@@ -0,0 +1,12 @@
...
+    id: add-verification-state-metadata
+    author: fletch
+    changes:
+    - addColumn:
+        tableName: verification_state
+        columns:
+        - name: metadata
+          type: json
+          constraints:
+            nullable: false
```
robfletcher, line 12: New column to store verification metadata

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
-  TAG,
-  DIGEST
-}
+const val RUN_JOB_TYPE: String = "runJob"
 
 /**
  * Holds the config info needed to execute a batch job on Titus via the Run Job stage
  */
 data class ContainerJobConfig(
-  val name : String,
```
robfletcher, line 21: I removed this because it's just the name of the stage and is defaulted to the stage type

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
-  val region: String,
-  val organization: String,
   /**
    * Repository name associated with the container, e.g.: "acme/widget"
    */
   val repository: String,
+  val tag: String? = "latest",
+  val digest: String? = null,
+  val application: String,
+  val location: TitusServerGroup.Location,
```
robfletcher, line 21: There's a lot of commonality between this class and `TitusServerGroup` so I re-used some structures from there. I feel like we can probably extract a common `TitusContainer` that is implemented by both `TitusServerGroup` and `TitusJob` or something like that.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
+  val resources: TitusServerGroup.Resources = TitusServerGroup.Resources(
+    cpu = 2,
+    memory = 2048,
+    disk = 256,
+    networkMbps = 1024
+  ),
```
robfletcher, line 27: Again re-using something from `TitusServerGroup` rather than having individual properties

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
+  val capacity: Capacity = Capacity(min = 1, max = 1, desired = 1),
```
robfletcher, line 28: re-use standard `Capacity` definition

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
   val serviceAccount: String,
   val credentials: String,
-  val registry : String,
```
robfletcher, line 32: I believe this field is just there for Deck to look up the docker images. Job runs fine without it.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
-  val type: ImageVersionReference = TAG,
-  /**
-   * Either a tag (e.g., "latest") or a digest (e.g., "sha256:be93efc727ba59813adc896859bccf32fb0f02202fe0526a9cd76326b9729cb3),
-   * depending on what [type] is
-   */
-  val versionIdentifier: String = "latest",
```
robfletcher, line 38: I changed this to `tag` and `digest` properties (one of which must be non-null) because it felt easier for users to understand

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
-  /**
-   * ID that identifies an image, e.g.:
-   *
-   * acme/widget:latest
-   * acme/widget:sha256:780f11bfc03495da29f9e2d25bf55123330715fb494ac27f45c96f808fd2d4c5
-   */
-  val imageId: String = "$repository:$versionIdentifier",
```
robfletcher, line 45: This is a derived property so no need to have it in the constructor

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
-  val maxCapacity: Int = 1,
-  val minCapacity: Int = 1,
-  val cpus: Int = 2,
-  val memoryMb: Int = 2048,
-  val diskMb: Int = 256,
-  val networkMbps: Int = 1024,
-  val gpus: Int = 0,
   val entrypoint: String = "",
-  val runtimeLimitSeconds: Int = 2700,
+  val runtimeLimit: Duration = Duration.ofSeconds(2700),
```
robfletcher, line 32: Used a `Duration` which we've tried to adhere to in the API

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
   val retries: Int = 0,
-  val cloudProvider : String = "titus",
-  val cloudProviderType: String = "aws",
```
robfletcher, line 58: Again these fields don't seem to be necessary for the job to work. I think they're there because Orca offers other ways to run a job than Titus

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
   val waitForCompletion: Boolean = true
 ) {
+  init {
+    require((tag == null) xor (digest == null)) {
+      "One, and only one, of digest or tag must be supplied"
+    }
+    require(retries >= 0) {
+      "Retries must be positive or zero"
+    }
+  }
```
robfletcher, line 50: Validation of fields in the constructor

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerification.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.netflix.spinnaker.keel.api.titus.TitusServerGroup.Location
+
+data class ContainerTestVerification(
+  val repository: String,
+  val tag: String = "latest",
+  val location: Location
+) : Verification {
+  override val type: String = "container-tests"
+  override val id: String = "$repository/$tag"
+}
```
robfletcher, line 13: This is the API class that will appear in delivery config. I wonder if we could merge this with the `ContainerJobConfig` class? I wanted to keep it as simple as possible though. I haven't (yet) added the capability to run digests but that would be easy enough.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
```diff
@@ -0,0 +1,84 @@
...
+      }
+        .let { response ->
+          when {
+            response.status.isSuccess() -> PASSED
+            response.status.isIncomplete() -> RUNNING
+            else -> FAILED
+          }
+        }
+    }
+  }
```
robfletcher, line 50: Call to orca to see if the task is complete yet. We'll need to extract some context from the task response here if it fails.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
```diff
@@ -0,0 +1,84 @@
...
+          user = context.deliveryConfig.serviceAccount,
+          application = context.deliveryConfig.application,
+          notifications = emptySet(),
+          stages = listOf(
+            ContainerJobConfig(
+              application = context.deliveryConfig.application,
+              location = verification.location,
+              repository = verification.repository,
+              serviceAccount = context.deliveryConfig.serviceAccount,
+              credentials = verification.location.account,
```
robfletcher, line 72: I'm not sure if there's any reason this would ever be a different value, but perhaps that's just within Netflix?

gal-yardeni:
The approach makes sense to me. If we can add some log statements in `ContainerTestVerificationEvaluator` I believe it'll make our lives easier when trying to debug a failure.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
```diff
@@ -0,0 +1,84 @@
...
+
+  override val supportedVerification: Pair<String, Class<ContainerTestVerification>> =
+    "container-tests" to ContainerTestVerification::class.java
+
+  override fun evaluate(
+    context: VerificationContext,
+    verification: Verification,
+    metadata: Map<String, Any?>
+  ): VerificationStatus {
+    val taskId = metadata["taskId"]
```
lorin, line 33: I think a constant (e.g., `TASK_ID`) would be better here.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
```diff
@@ -0,0 +1,84 @@
...
+    require(verification is ContainerTestVerification) {
+      "Expected a ${ContainerTestVerification::class.simpleName} but received a ${verification.javaClass.simpleName}"
+    }
+
+    return runBlocking {
+      withContext(IO) {
+        taskLauncher.submitJob(
+          subject = "container integration test",
+          description = "testing running a container",
+          correlationId = "keel-run-container-test",
```
lorin, line 62: We should change the subject/description/correlationId at some point to make them more meaningful.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerification.kt
lorin: Merging with `ContainerJobConfig` sounds reasonable to me.

As for digests, I'm tempted to not expose support for them in the delivery config until users come to us with an explicit use case.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
+    ),
+
+    "waitForCompletion" to waitForCompletion,
+    "cloudProvider" to cloudProvider,
+    "deferredInitialization" to deferredInitialization,
+    "credentials" to credentials,
+    "account" to location.account,
+    "region" to location.region,
+  ).also {
+    println(ObjectMapper().writeValueAsString(it))
```
lorin, line 111: This println looks like a leftover from development.

lorin (approved):
## My understanding

### Purpose

Implement a Titus-container-based verifier.

### Major changes

#### Add metadata to Verification objects

* Database: Adds a metadata column to the verification_state table
* jooq: Deserialize `VERIFICATION_STATE.METADATA` to `Map<String, ?>`
* VerificationEvaluator.start/evaluate methods now take additional arguments (context, metadata).
* VerificationState now contains metadata object
* VerificationRepository.updateState takes metadata argument
* SqlVerificationRepository: reads/writes metadata

#### Support executing Titus containers

New class: ContainerTestVerficationEvaluator

* start -> runs a new container via orca
* evaluate -> checks the status of the orca task

### Minor changes

* TestContainerVerification renamed to ContainerTestVerification (w/some field changes)
* ContainerJobConfig: removed unnecessary fields, abstracted some commonality with other Titus code
* Moved `ResultQuery<*>.fetchOneInto` extension function from SqlVerificationRepository.kt to Jooq.kt

## My feedback

Changes look great. I picked a couple of nits.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/VerificationEvaluator.kt
```diff
@@ -10,12 +11,19 @@ interface VerificationEvaluator<VERIFICATION: Verification> {
...
+  fun evaluate(
+    context: VerificationContext,
+    verification: Verification,
+    metadata: Map<String, Any?>
+  ): VerificationStatus
 
   /**
    * Start running [verification].
+   *
+   * @return any metadata needed to [evaluate] the verification in future.
```
luispollo, line 26:
```suggestion
   * @return any metadata needed to [evaluate] the verification in the future.
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -137,34 +136,38 @@ class SqlVerificationRepository(
...
+          VERIFICATION_STATE.STARTED_AT,
+          VERIFICATION_STATE.ENDED_AT
+        )
+          .from(VERIFICATION_STATE)
+          .where(VERIFICATION_STATE.ENVIRONMENT_UID.eq(environmentUid))
+          .and(VERIFICATION_STATE.ARTIFACT_UID.eq(artifact.uid))
+          .and(VERIFICATION_STATE.ARTIFACT_VERSION.eq(version))
+          .fetch()
+          .associate { (id, status, started_at, ended_at) ->
+            id to VerificationState(status, started_at, ended_at)
```
luispollo, line 155: Metadata is not relevant when retrieving multiple state records?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -1,123 +1,112 @@
...
   val capacityGroup: String = application,
   val containerAttributes: Map<String, String> = emptyMap(),
   val environmentVariables: Map<String, String> = emptyMap(),
   val labels: Map<String, String> = emptyMap(),
-  val deferredInitialization : Boolean = true,
+  val deferredInitialization: Boolean = true,
   val waitForCompletion: Boolean = true
 ) {
+  init {
+    require((tag == null) xor (digest == null)) {
```
luispollo, line 44: `xor` -- TIL!

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerification.kt
```diff
@@ -0,0 +1,13 @@
+package com.netflix.spinnaker.keel.titus.verification
+
+import com.netflix.spinnaker.keel.api.Verification
+import com.netflix.spinnaker.keel.api.titus.TitusServerGroup.Location
+
+data class ContainerTestVerification(
```
luispollo, line 6: This reads backwards to me as is (as in "a container test", not "a test that runs in a container"). Suggestion:
```suggestion
data class ContainerBasedTestVerification(
```

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerification.kt
luispollo: +1 to Lorin's suggestion.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.netflix.spinnaker.keel.orca.OrcaService
+import com.netflix.spinnaker.keel.titus.batch.ContainerJobConfig
+import com.netflix.spinnaker.keel.titus.batch.createRunJobStage
+import kotlinx.coroutines.Dispatchers.IO
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
+import org.springframework.stereotype.Component
+
+@Component
+class ContainerTestVerificationEvaluator(
```
luispollo, line 20:
```suggestion
/**
 * A [VerificationEvaluator] that runs a test container to verify an environment.
 */
class ContainerTestVerificationEvaluator(
```

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
luispollo: Just a suggestion to see what can be common between this implementation and `OrcaTaskMonitoringAgent`, which might be worth extracting into shared functions.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
```diff
@@ -0,0 +1,84 @@
...
+          when {
+            response.status.isSuccess() -> PASSED
+            response.status.isIncomplete() -> RUNNING
+            else -> FAILED
+          }
+        }
+    }
+  }
+
+  override fun start(context: VerificationContext, verification: Verification): Map<String, Any?> {
```
luispollo, line 52: Shouldn't these functions be `suspend` so that the orchestration in the upper layer can decide how to run these consistently across different types of evaluators?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
```diff
@@ -0,0 +1,84 @@
...
+    require(verification is ContainerTestVerification) {
+      "Expected a ${ContainerTestVerification::class.simpleName} but received a ${verification.javaClass.simpleName}"
+    }
+
+    return runBlocking {
+      withContext(IO) {
+        taskLauncher.submitJob(
+          subject = "container integration test",
```
luispollo, line 60: `subject` will end up in the task event record we store in the DB, prefixed by a `type`, which we're not passing here and defaults to `CONSTRAINT` in this variant of the function. I think you'll need to pass the type. Also, it's meant as an internal identifier, so I'd suggest passing the container image here.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
```diff
@@ -0,0 +1,84 @@
...
+          description = "testing running a container",
```
luispollo, line 61: Suggestion: 
```suggestion
          description = "Verifying environment ${context.environmentName} with container ${verification.container}",
```

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
luispollo: +1. Made some suggestions above.

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluatorTests.kt
```diff
@@ -0,0 +1,164 @@
...
+  private val verification = ContainerTestVerification(
+    repository = "illuminati/fnord",
+    location = Location(
+      account = "titustestvpc",
+      region = "ap-south-1"
+    )
+  )
+
+  @Test
+  fun `starting verification launches a container job via orca`() {
```
luispollo, line 51:
```suggestion
  fun `starting verification launches a container job via the task launcher`() {
```

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluatorTests.kt
```diff
@@ -0,0 +1,164 @@
...
+    verify {
+      taskLauncher.submitJob(
+        user = any(),
+        application = any(),
+        notifications = any(),
+        subject = any(),
+        description = any(),
+        correlationId = any(),
+        stages = match {
+          it.first()["type"] == RUN_JOB_TYPE
```
luispollo, line 79: It'd be nice to check that at least the container image matches what was specified. 🙂

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluatorTests.kt
```diff
@@ -0,0 +1,164 @@
...
+  @EnumSource(
+    OrcaExecutionStatus::class,
+    names = ["NOT_STARTED", "RUNNING", "PAUSED", "SUSPENDED", "BUFFERED"]
+  )
+  fun `verification is considered still running if orca task is running`(taskStatus: OrcaExecutionStatus) {
+    val taskId = ULID().nextULID()
+    val previousState = runningState(taskId)
+
+    every {
+      orca.getOrchestrationExecution(taskId, any())
```
luispollo, line 95: Side-comment: this just made me realize that we're leaking from the `TaskLauncher` abstraction. You can launch jobs using the abstract type, but not query for status...

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluatorTests.kt
```diff
@@ -0,0 +1,164 @@
...
+      status = taskStatus
+    )
+
+    expectThat(subject.evaluate(context, verification, previousState.metadata)) isEqualTo PASSED
+  }
+
+  @ParameterizedTest(name = "verification is considered failed if orca status is {0}")
+  @EnumSource(
+    OrcaExecutionStatus::class,
+    names = ["TERMINAL", "FAILED_CONTINUE", "STOPPED", "CANCELED"]
```
luispollo, line 136: You sure about the spelling of `CANCELED` in orca? I didn't look, but caught my eye... 🙂

luispollo: Never mind, you got it right.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
robfletcher: Oh yeah, I forgot about that. I'll take a look.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
robfletcher: D'oh yes

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
robfletcher: I think this is because we merged the change that added this query after I'd added metadata. I'll add it here.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
robfletcher: Yeah, no operator for it like in Groovy where it's `^`

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluatorTests.kt
robfletcher: Yeah, I copied these from the enum. It's very annoying you can't use an enum in annotations.

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluatorTests.kt
robfletcher: Is it really testing any logic or just that we can copy fields from one place to another?

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluator.kt
robfletcher: I figured that the implementation itself was best placed to know if the work it was doing needed to run off thread and could handle it within the method. If we wanted to async multiple parallel method calls it might make sense to do it externally but I'm not sure it does here.

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluatorTests.kt
luispollo: Not really, but you could say the same about the check for the stage type you have there. I just think it helps the test tell the story, is all. Not critical.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -137,34 +136,39 @@ class SqlVerificationRepository(
...
     verification: Verification,
-    status: VerificationStatus
+    status: VerificationStatus,
+    metadata: Map<String, Any?>?
   ) {
     with(context) {
       jooq
         .insertInto(VERIFICATION_STATE)
         .set(VERIFICATION_STATE.STATUS, status)
+        .set(VERIFICATION_STATE.METADATA, metadata)
```
luispollo, line 171: I thought I saw a comment somewhere that we'd only update the metadata if the parameter is not null?
```suggestion
        .apply { if (metadata != null) set(VERIFICATION_STATE.METADATA, metadata) }
```

robfletcher: `JSONConverter` implements that behavior already

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerificationEvaluatorTests.kt
robfletcher: Yes. I think this requires some thought in terms of handling the response coming back from Orca. There's limited value if we just pass it thru. If we can marshal it into a useful format that exposes error information in a consistent way it will make a lot of sense. I think to do it properly is outside of the scope of this PR tho. Maybe in the next one which will involve surfacing information about test results to the user.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/ContainerTestVerification.kt
robfletcher: Changing this now

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/actuation/SubjectType.kt
```diff
@@ -1,5 +1,5 @@
 package com.netflix.spinnaker.keel.api.actuation
 
 enum class SubjectType {
-  RESOURCE, CONSTRAINT
+  RESOURCE, CONSTRAINT, VERIFICATION
 }
```
robfletcher, line 5: New task subject type as suggested by @luispollo

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -137,34 +136,39 @@ class SqlVerificationRepository(
...
+  override fun getStates(context: VerificationContext): Map<String, VerificationState> =
+    with(context) {
+      when {
+        verifications.isEmpty() -> emptyMap() // Optimization: don't hit the db if we know there are no entries
+        else -> jooq.select(
+          VERIFICATION_STATE.VERIFICATION_ID,
+          VERIFICATION_STATE.STATUS,
+          VERIFICATION_STATE.STARTED_AT,
+          VERIFICATION_STATE.ENDED_AT,
+          VERIFICATION_STATE.METADATA
```
robfletcher, line 148: I just added this one select column

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/TestContainerVerificationEvaluator.kt
```diff
@@ -0,0 +1,96 @@
...
+
+  override fun start(context: VerificationContext, verification: Verification): Map<String, Any?> {
+    require(verification is TestContainerVerification) {
+      "Expected a ${TestContainerVerification::class.simpleName} but received a ${verification.javaClass.simpleName}"
+    }
+
+    return runBlocking {
+      withContext(IO) {
+        taskLauncher.submitJob(
+          type = VERIFICATION,
```
robfletcher, line 66: Now passing new `SubjectType` enum here

luispollo (approved):
LGTM. Good stuff, @robfletcher!

PR #1702, opened 2020-12-09
use Instant for timestamp columns

robfletcher:
Finally figured out how to configure this


keel-sql/src/main/resources/jooqConfig.xml
```diff
@@ -20,6 +20,11 @@
           <name>LONGVARCHAR</name>
           <includeTypes>JSON</includeTypes>
         </forcedType>
+        <forcedType>
+          <userType>java.time.Instant</userType>
+          <includeTypes>TIMESTAMP</includeTypes>
+          <converter>com.netflix.spinnaker.jooq.TimestampToInstantConverter</converter>
+        </forcedType>
```
robfletcher, line 27: Here's the bit of JOOQ configuration magic required

keel-sql/src/main/resources/jooqConfig.xml
```diff
@@ -20,6 +20,11 @@
           <name>LONGVARCHAR</name>
           <includeTypes>JSON</includeTypes>
         </forcedType>
+        <forcedType>
+          <userType>java.time.Instant</userType>
+          <includeTypes>TIMESTAMP</includeTypes>
+          <converter>com.netflix.spinnaker.jooq.LocalDateTimeToInstantConverter</converter>
+        </forcedType>
```
robfletcher, line 27: Here's the bit of JOOQ configuration magic required

keel-sql/src/main/kotlin/com/netflix/spinnaker/jooq/LocalDateTimeToInstantConverter.kt
```diff
@@ -0,0 +1,17 @@
...
+class LocalDateTimeToInstantConverter : AbstractConverter<LocalDateTime, Instant>(
+  LocalDateTime::class.java,
+  Instant::class.java
+) {
+  override fun from(databaseObject: LocalDateTime?): Instant? =
+    databaseObject?.toInstant(UTC)
+
+  override fun to(userObject: Instant?): LocalDateTime? =
+    userObject?.atZone(UTC)?.toLocalDateTime()
+}
```
robfletcher, line 17: Converter that converts between the type JOOQ wants to use by default (`LocalDateTime`) and the one we want to use (`Instant`)

keel-sql/src/main/resources/jooqConfig.xml
emjburns: LOVE

emjburns (approved):

luispollo:
> Finally figured out how to configure this

So glad you did! I tried this before and couldn't make it work, but I think it was back when we had half the columns using `BIGINT` and half using `TIMESTAMP`. Thanks for taking the time to improve this!

PR #1703, opened 2020-12-09
fix(builds): follow up on things we may have missed

emjburns:
**Problem**
Because of my coding, we missed monitoring some events (fixed in https://github.com/spinnaker/keel/pull/1698).

**Solution**
This pr looks for single events with a non-ending status (like `NOT_STARTED` or `RUNNING`) that have asked to be monitored (`startMonitoring  == true`) and have a timestamp of more than 5 minutes ago. If found, it triggers monitoring again. This should pick out all the events we missed.

This is kicked off when a user loads the `managed/application/{application}?entities=artifacts` page. I'm going to slyly visit every app once this is rolled out so it triggers the monitoring w/o our users knowing.

Once this is done I'll watch for it to re-surface (indicating that we're missing monitoring of events for some other reason...). If it doesn't, I"ll remove this code.


gal-yardeni:
talked with Emily offline about making it a potentially an admin endpoint. We agree of merging it for now and convert it into an endpoint in the future.

gal-yardeni (approved):
LGTM

PR #1704, opened 2020-12-10
JOOQ enum columns

robfletcher:
This change makes JOOQ use the actual enum types for columns that we model with enums


keel-sql/src/main/kotlin/com/netflix/spinnaker/jooq/enumConverters.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.netflix.spinnaker.keel.core.api.PromotionStatus
+import com.netflix.spinnaker.keel.events.PersistentEvent
+import com.netflix.spinnaker.keel.lifecycle.LifecycleEventStatus
+import com.netflix.spinnaker.keel.lifecycle.LifecycleEventType
+import com.netflix.spinnaker.keel.notifications.NotificationScope
+import com.netflix.spinnaker.keel.notifications.NotificationType
+import com.netflix.spinnaker.keel.pause.PauseScope
+import org.jooq.impl.EnumConverter
+
+class ArtifactStatusConverter : EnumConverter<String, ArtifactStatus>(
```
luispollo, line 15: Too bad we can't do this with a generic type to avoid the duplication. Oh well.

robfletcher: We could if the XML would allow for method references. There is a configuration DSL but I think we'd get into circular dependency issues because we're using the code generation stuff.

luispollo (approved):
Very, very nice clean-up! Thanks for doing this! 🙏

keel-sql/src/main/kotlin/com/netflix/spinnaker/jooq/enumConverters.kt
luispollo: It's alright, this is a great improvement as it is.

PR #1705, opened 2020-12-10
feat(verify): only promote if verifications succeed

lorin:
This is work towards supporting verification.

# Problem

We don't want to promote a version to the next environment if there are verifications that haven't succeeded yet.

# Proposed solution

Extend the behavior of the DependsOnConstraintEvaluator to enforce that all verifications have passed in the required environment before allowing promotion.

The DependsOnConstraintEvaluator seemed like a natural place to put this logic, since verification is associated with dependent environments. 

# Future work

Provide feedback about the status of verifications. The code doesn't (yet) provide any user/operator feedback about the status of verifications.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
```diff
@@ -27,6 +27,14 @@ interface VerificationRepository {
...
 
   fun nextEnvironmentsForVerification(minTimeSinceLastCheck: Duration, limit: Int) : Collection<VerificationContext>
+
+  fun wasSuccessfullyVerifiedIn(
+    deliveryConfig: DeliveryConfig,
+    artifactReference: String,
+    version: String,
+    environmentName: String,
+    verification : Verification
+  ): Boolean
```
emjburns, line 37: Is this actually implemented somewhere?

emjburns (approved):
Looks good to me! I agree with your placing of this in the depends on constraint.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
emjburns: I mean it must be implemented somewhere, so that this compiles... :D

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -117,6 +118,25 @@ class SqlVerificationRepository(
     }
   }
 
+  override fun wasSuccessfullyVerifiedIn(
+    deliveryConfig: DeliveryConfig,
+    artifactReference: String,
+    version: String,
+    environmentName: String,
```
robfletcher, line 125: Can we re-use the `VerificationContext` object like on the other methods' signatures?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
luispollo: Only if there is an implementation of the interface...

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
```diff
@@ -38,6 +40,27 @@ class DependsOnConstraintEvaluator(
       artifact,
       version,
       requiredEnvironment.name
+    ) && allVerificationsSucceededIn(
```
gal-yardeni, line 43: can we add a log statement here, in case the verification failed?

gal-yardeni: or even in both cases (succeeded/failed)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
```diff
@@ -38,6 +40,27 @@ class DependsOnConstraintEvaluator(
...
+  ): Boolean =
+    environment.verifyWith.all { verification ->
+      verificationRepository.wasSuccessfullyVerifiedIn(
+        deliveryConfig,
+        artifact.reference,
+        version,
+        environment.name,
+        verification
+      )
+    }
```
robfletcher, line 65: This may be premature optimization but I wonder if we could query the status of all the verifications for the context and check that a) none are missing and b) they all passed. That way a single query could do the job.

robfletcher (commented):
I really like the idea of making it part of the depends on constraint

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
lorin: Yeah, the method is implemented in SqlVerificationRepository.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
luispollo, line 43: It's funny that this question _just_ came up in our all hands meeting and @robfletcher was saying that verification would not necessarily block promotion of the artifact to the next environment, but I think this makes total sense. My recollection is that we had agreed on this behavior, right?

gal-yardeni (approved):
makes sense to me! I think more log statements is needed here, but we can add those later as well!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
robfletcher: Yeah probably. I wasn't sure even as I was saying that.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
```diff
@@ -27,6 +27,14 @@ interface VerificationRepository {
   )
 
   fun nextEnvironmentsForVerification(minTimeSinceLastCheck: Duration, limit: Int) : Collection<VerificationContext>
+
+  fun wasSuccessfullyVerifiedIn(
+    deliveryConfig: DeliveryConfig,
+    artifactReference: String,
+    version: String,
+    environmentName: String,
```
luispollo, line 35: Consider passing a `DeliveryArtifact` and an `Environment` objects for parameters instead to avoid unnecessary changes to the signature later if we change the identifiers. Particularly for the artifact, where the reference alone as an identifier has been cause for debate (it also helps by allowing access to more details when logging).

robfletcher: we've already got a `VerificationContext` object that's designed to abstract the details of these parameters that basically any method on this repo is going to need.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -117,6 +118,25 @@ class SqlVerificationRepository(
...
+    verification : Verification
+  ) =
+    getState(
+      VerificationContext(
+        deliveryConfig,
+        environmentName,
+        artifactReference,
+        version
+      ),
```
luispollo, line 134: Side-note: I just realized this is precisely what a `Deployment` class might model if we were to introduce that. It would make a bunch of our method signatures shorter and more self-explanatory (like in this case: "was deployment X successfully verified with verification Y?").
Cc: @robfletcher

luispollo (approved):
This looks great, @lorin! Simple and easy to follow! Love it! ❤️

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
luispollo: P.S. @robfletcher Any chance you might agree with just renaming that class?

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
luispollo: Right, but that still has just the environment name and artifact reference, so if you wanted to access other properties you'd need to query for them (which we do all over our repositories and is wasteful).

robfletcher: that's an implementation detail of that context class that we could change outside of the context of these methods

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
robfletcher: Yeah, if we are considering using it in a wider context

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
lorin: Added!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
lorin: Done.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
lorin: Done. I changed the verification repository method to `getStates`, which retrieves all of the states in associated with the context, and the DependsOnConstraintEvaluator does the check to see if any are missing.

erikmunson:
late to the party, but love the approach of building into dependsOn ❤️  in regards to this note on future work:

> Provide feedback about the status of verifications. The code doesn't (yet) provide any user/operator feedback about the status of verifications.

one other thing to consider: will customers need additional information in the depends on constraint experience to inform them of what specifically is blocking that constraint from passing? before this change, there was only one reason (hasn't come up healthy in the target environment). Now, there's two reasons a constraint would be blocked from passing. Do we need to explicitly disambiguate in the UI, or can saying something generic still get us the right results? I don't really have an opinion yet, just food for thought.

PR #1706, opened 2020-12-10
automatically handle JOOQ enum columns

robfletcher:
Turns out you don't need to implement a dedicated converter for each enum


luispollo (approved):
Ha! Nice!

PR #1708, opened 2020-12-10
parse/convert some JSON columns automatically

robfletcher:
I didn't do all the columns as some are more complex than others, but this is a start.


gal-yardeni (approved):
LGTM. Does this mean that from now on, when adding a new object which is intended to be stored/read from the DB, we would have to configure it in `jooqConfig`?

robfletcher:
> LGTM. Does this mean that from now on, when adding a new object which is intended to be stored/read from the DB, we would have to configure it in `jooqConfig`?

We don't have to, but I think in cases where there are many queries dealing with the same table columns it makes it easier to set up the parsing in one place.

PR #1710, opened 2020-12-11
disable @Scheduled things in integration tests

robfletcher:
I thought we'd already accomplished this with the `@DisableSpringScheduling` but using a debugger I could see it wasn't actually working. This really does turn scheduling off when running tests. No more big explosion of stack traces because mock beans called in scheduled actions aren't set up. Inline comments explain how it works.


keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/tx/DeliveryConfigDeletionTests.kt
```diff
@@ -24,16 +23,8 @@ import strikt.assertions.isEmpty
...
 @SpringBootTest(
   classes = [KeelApplication::class, TestConfiguration::class],
-  webEnvironment = MOCK,
-  properties = [
-    "spinnaker.baseUrl=http://spinnaker",
-    "sql.enabled=true",
-    "sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename",
-    "sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename",
-    "spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver"
-  ]
```
robfletcher, line 34: I just made all these properties consistent in `keel-sql/src/test/resources/application.properties` there's no need to specify them in each class as they're always the same.

keel-sql/src/test/resources/application.properties
```diff
@@ -6,5 +7,10 @@ orca.baseUrl=https://localhost:8081
 front50.baseUrl=https://localhost:8082
 igor.baseUrl=https://localhost:8083
 echo.baseUrl=https://localhost:8086
+spinnaker.baseUrl=http://spinnaker
 services.fiat.baseUrl=https://fiat.net
 spring.application.name=keel
+sql.enabled=true
+sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename
+sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename
+spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver
```
robfletcher, line 16: Properties moved over from individual tests

keel-sql/src/test/resources/application.properties
```diff
@@ -1,3 +1,4 @@
+scheduling.enabled=false
```
robfletcher, line 1: Explicitly disable scheduling in `keel-sql` tests

keel-web/src/main/kotlin/com/netflix/spinnaker/config/SchedulingConfiguration.kt
```diff
@@ -0,0 +1,10 @@
+package com.netflix.spinnaker.config
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.context.annotation.Configuration
+import org.springframework.scheduling.annotation.EnableScheduling
+
+@Configuration
+@ConditionalOnProperty("scheduling.enabled", matchIfMissing = true)
+@EnableScheduling
+class SchedulingConfiguration
```
robfletcher, line 10: This is a new configuration class that conditionally enables scheduling (previously we just had `@EnableScheduling` on `KeelApplication` so it was always on). If the `scheduling.enabled` property is missing, or equal to `true` scheduling is enabled. So tests can set it to `false` to disable it.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/Main.kt
```diff
@@ -47,7 +46,6 @@ private val DEFAULT_PROPS = mapOf(
   PluginsAutoConfiguration::class
 )
 @EnableAsync
-@EnableScheduling
 class KeelApplication
```
robfletcher, line 49: No longer enable scheduling on the Boot app directly as we want to do it conditionally

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ArtifactControllerTests.kt
```diff
@@ -1,41 +0,0 @@
...
-
-  @MockkBean
-  lateinit var authorizationSupport: AuthorizationSupport
-
-  fun tests() = rootContext {
-    after {
-      clearAllMocks()
-    }
-  }
-}
```
robfletcher, line 41: This class didn't actually have any tests so I deleted it

keel-web/src/test/resources/application.properties
```diff
@@ -1,3 +1,4 @@
+scheduling.enabled=false
```
robfletcher, line 1: Explicitly disable scheduling in `keel-web` tests

keel-sql/src/test/resources/application.properties
luispollo: I remember trying to debug this myself several months ago and seeing that the annotations didn't take effect, but I also have this vague memory that even the config property wasn't enough, because kork was enabling scheduling programmatically and it was hard-coded. Are you positive this does the trick? If not, I can go dig that code in kork again.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/SchedulingConfiguration.kt
luispollo: I remember trying to debug this myself several months ago and seeing that the annotations didn't take effect, but I also have this vague memory that even the config property wasn't enough, because kork was enabling scheduling programmatically and it was hard-coded. Are you positive this does the trick? If not, I can go dig that code in kork again.

luispollo: Looks like that was removed in this PR: https://github.com/spinnaker/kork/pull/743 🙏

emjburns (approved):

keel-web/src/main/kotlin/com/netflix/spinnaker/config/SchedulingConfiguration.kt
luispollo: Original line was here: https://github.com/spinnaker/kork/blob/44e1536e59fe19aa461b03e73438b18597408ec8/kork-web/src/main/groovy/com/netflix/spinnaker/config/TomcatConfiguration.groovy#L40

PR #1712, opened 2020-12-14
Fixed JOOQ bump from Kork

robfletcher:
Had to make the Gradle plugin use the same version of JOOQ as Kork. Also JOOQ has added `@Nullable` and `@NotNull` annotations to method signatures so there were a couple of minor updates required.

### DIFF
--- a/keel-sql/keel-sql.gradle.kts
+++ b/keel-sql/keel-sql.gradle.kts
@@ -6,7 +6,7 @@ val buildingInDocker = project.properties["buildingInDocker"]?.toString().let { 
 plugins {
   `java-library`
   id("kotlin-spring")
-  id("nu.studer.jooq") version "5.0.1"
+  id("nu.studer.jooq") version "5.2"
   id("org.liquibase.gradle") version "2.0.4"
 }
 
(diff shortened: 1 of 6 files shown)


keel-sql/keel-sql.gradle.kts
```diff
@@ -120,6 +121,10 @@ afterEvaluate {
   }
 }
 
+jooq {
+  version.set("3.14.4")
+}
```
robfletcher, line 126: This aligns the Gradle plugin with Kork

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -184,7 +185,7 @@ class SqlArtifactRepository(
         .from(DELIVERY_ARTIFACT)
         .where(DELIVERY_ARTIFACT.NAME.eq(name))
         .and(DELIVERY_ARTIFACT.TYPE.eq(type))
-        .fetchOne()
+        .fetchSingle()
```
robfletcher, line 188: `fetchOne` is `@Nullable`

emjburns (approved):

luispollo (approved):
LGTM!

PR #1714, opened 2020-12-18
chore(tests): add verifyWith parsing test, use spring for plumbing

lorin:
This PR has two changes:

1. Add a test that parses delivery config with `verifyWith`
2. Use Spring framework to instantiate the beans necessary for configuring the Jackson deserialization

The motivation for 2 is to help us catch potential incorrect Spring configuration issues, by having the delivery-config-loading tests execute the same startup code as keel uses when it starts up on its own. The downside is that these tests will run a little slower.

I had to explicitly enable the ec2 and titus plugins in the application.properties used by the tests so that the ec2 and titus resource handlers and spec migrators would register the appropriate types with the Jackson object mapper.

### DIFF
--- a/keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ConvertExampleFilesTest.kt
+++ b/keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ConvertExampleFilesTest.kt
@@ -17,47 +17,32 @@
  */
 package com.netflix.spinnaker.keel.rest
 
-import com.fasterxml.jackson.databind.jsontype.NamedType
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec
 import com.netflix.spinnaker.keel.api.ec2.ClassicLoadBalancerSpec
 import com.netflix.spinnaker.keel.api.ec2.ClusterSpec
-import com.netflix.spinnaker.keel.api.ec2.EC2_APPLICATION_LOAD_BALANCER_V1_1
-import com.netflix.spinnaker.keel.api.ec2.EC2_CLASSIC_LOAD_BALANCER_V1
-import com.netflix.spinnaker.keel.api.ec2.EC2_CLUSTER_V1
-import com.netflix.spinnaker.keel.api.ec2.EC2_CLUSTER_V1_1
-import com.netflix.spinnaker.keel.api.ec2.EC2_SECURITY_GROUP_V1
 import com.netflix.spinnaker.keel.api.ec2.SecurityGroupSpec
 import com.netflix.spinnaker.keel.api.ec2.old.ClusterV1Spec
-import com.netflix.spinnaker.keel.api.plugins.SupportedKind
+import com.netflix.spinnaker.keel.api.titus.TestContainerVerification
 import com.netflix.spinnaker.keel.api.titus.TitusClusterSpec
+import com.netflix.spinnaker.keel.api.titus.TitusServerGroup.Location
 import com.netflix.spinnaker.keel.core.api.SubmittedDeliveryConfig
-import com.netflix.spinnaker.keel.ec2.jackson.registerKeelEc2ApiModule
-import com.netflix.spinnaker.keel.test.configuredTestYamlMapper
-import com.netflix.spinnaker.keel.titus.TITUS_CLUSTER_V1
-import com.netflix.spinnaker.keel.titus.jackson.registerKeelTitusApiModule
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
 import strikt.api.expectCatching
 import strikt.assertions.isA
+import strikt.assertions.isEqualTo
 import strikt.assertions.isSuccess
 
+@SpringBootTest
 class ConvertExampleFilesTest : JUnit5Minutests {
-  private val mapper = configuredTestYamlMapper()
-    .registerKeelEc2ApiModule()
-    .registerKeelTitusApiModule()
+  @Autowired
+  lateinit var mapper : YAMLMapper
 
   fun tests() = rootContext<Unit> {
-    before {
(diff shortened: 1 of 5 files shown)


keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ConvertExampleFilesTest.kt
```diff
@@ -17,47 +17,32 @@
...
 import dev.minutest.rootContext
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
 import strikt.api.expectCatching
 import strikt.assertions.isA
+import strikt.assertions.isEqualTo
 import strikt.assertions.isSuccess
 
+@SpringBootTest
 class ConvertExampleFilesTest : JUnit5Minutests {
```
robfletcher, line 41:
```suggestion
class ConvertExampleFilesTest @Autowired constructor(
  private val mapper: YAMLMapper
) : JUnit5Minutests {
```

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ConvertExampleFilesTest.kt
```diff
@@ -17,47 +17,32 @@
...
-  private val mapper = configuredTestYamlMapper()
-    .registerKeelEc2ApiModule()
-    .registerKeelTitusApiModule()
+  @Autowired
+  lateinit var mapper : YAMLMapper
```
robfletcher, line 43:
```suggestion
```

robfletcher (approved):
Looks good. Much neater to have it using Spring rather than hand-wiring the configuration

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ConvertExampleFilesTest.kt
```diff
@@ -161,7 +146,22 @@ class ConvertExampleFilesTest : JUnit5Minutests {
           .isA<TitusClusterSpec>()
       }
     }
-  }
 
-  private fun SupportedKind<*>.toNamedType(): NamedType = NamedType(specClass, kind.toString())
+    context("titus cluster with test container") {
+      val file = this.javaClass.getResource("/examples/titus-cluster-with-test-container.yml").readText()
+
+      test("yml can be parsed") {
```
luispollo, line 153:
```suggestion
      test("yaml can be parsed") {
```

keel-web/src/test/resources/application.properties
```diff
@@ -14,3 +14,5 @@ sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:somepo
 sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename
 spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver
 spring.application.name=keel
+keel.plugins.ec2.enabled=true
+keel.plugins.titus.enabled=true
```
robfletcher, line 18: There are a few other tests where we explicitly enable these. Maybe we can not bother doing so any more?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ConvertExampleFilesTest.kt
```diff
@@ -17,47 +17,32 @@
...
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
 import strikt.api.expectCatching
 import strikt.assertions.isA
+import strikt.assertions.isEqualTo
 import strikt.assertions.isSuccess
 
+@SpringBootTest
```
robfletcher, line 40:
```suggestion
@SpringBootTest(webEnvironment = NONE)
```

Probably want this to avoid it starting up the embedded web server since the tests don't go via the REST API.

keel-web/src/test/resources/application.properties
luispollo, line 18: Aren't these going to apply to, and potentially affect, other tests? You can specify specific properties for the test with the `@SpringBootTest` annotation.

keel-web/src/test/resources/examples/titus-cluster-with-test-container.yml
```diff
@@ -0,0 +1,26 @@
...
+    type: docker
+    reference: rocketskates
+environments:
+  - name: test
+    locations:
+      account: test
+      regions:
+        - name: us-east-1
+    resources:
+      - kind: ec2/cluster@v1.1
```
luispollo, line 14:
```suggestion
      - kind: titus/cluster@v1
```

luispollo: I don't understand how the tests are passing with the delivery config in this state... 🤔

lorin: > I don't understand how the tests are passing with the delivery config in this state... 🤔

Do you mean, because the artifact type doesn't match the resource type?

I'd like to build on this to write a more extensive set of tests against delivery configs to catch more invalid configsthat we aren't checking for yet.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/ConvertExampleFilesTest.kt
lorin: Unfortunately, I still don't have a good enough mental model about how Spring's DI implementation works to understand why we don't need to specify the main class. All I know is that it seems to work.

keel-web/src/test/resources/application.properties
lorin: It looks like all of the tests are passing. My instinct is that it's better to leave these enabled for all tests, and so I'm going to remove the cases where we explicitly enable these in tests and leave them set here.

keel-web/src/test/resources/examples/titus-cluster-with-test-container.yml
luispollo: No, I mean because that `container` field you have there is not part of `ClusterSpec` (EC2), so I don't understand how that would parse correctly.

PR #1716, opened 2020-12-18
Persist Environment.verifyWith property

robfletcher:
Managed to neglect this in previous PR. I had assumed there was a single JSON column that captured all of the Environment data.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
@@ -4,6 +4,7 @@ import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.Resource
 import com.netflix.spinnaker.keel.api.ResourceKind.Companion.parseKind
+import com.netflix.spinnaker.keel.api.Verification
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactOriginFilterSpec
 import com.netflix.spinnaker.keel.api.artifacts.BranchFilterSpec
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
@@ -50,6 +51,12 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
 
   open fun flush() {}
 
+  data class DummyVerification(
+    override val id: String = "whatever"
+  ) : Verification {
+    override val type = "verification"
+  }
+
   data class Fixture<T : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository, P : PausedRepository>(
     val deliveryConfigRepositoryProvider: (ResourceSpecIdentifier) -> T,
     val resourceRepositoryProvider: (ResourceSpecIdentifier) -> R,
@@ -75,7 +82,7 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
     internal val artifactRepository: A = artifactRepositoryProvider()
     internal val clock = MutableClock()
 
-    val artifact: DebianArtifact =  DebianArtifact(
+    val artifact: DebianArtifact = DebianArtifact(
       name = "keel",
       deliveryConfigName = deliveryConfig.name,
       vmOptions = VirtualMachineOptions(baseOs = "bionic", regions = setOf("us-west-2"))
@@ -229,6 +236,9 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
                     environment = "test"
                   ),
                   ManualJudgementConstraint()
+                ),
+                verifyWith = setOf(
+                  DummyVerification()
                 ),
                 resources = setOf(
                   resource(kind = parseKind("ec2/cluster@v1")),
@@ -526,7 +536,7 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
             .resources
             .filter { it.kind == parseKind("ec2/security-group@v1") }
             .toSet()
-          val newEnvironments= deliveryConfig
+          val newEnvironments = deliveryConfig
             .environments
             .map {
               it.copy(resources = it.resources - movedResources)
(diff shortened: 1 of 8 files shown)


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -483,7 +531,12 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
 
           test("deletes related application pause records") {
             repository.deleteByApplication(deliveryConfig.application)
-            expectThat(pausedRepository.getPause(PauseScope.APPLICATION, deliveryConfig.application)).isNull()
+            expectThat(
```
luispollo, line 534: Just out of curiosity, do you have something configured in IntelliJ that's auto-applying formatting rules in the files you touch? I'd love to turn that on too.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -230,6 +247,9 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
                   ),
                   ManualJudgementConstraint()
                 ),
+                verifyWith = setOf(
```
luispollo, line 250: It doesn't look like this is being used in the test at all?

keel-sql/src/main/resources/db/changelog/20201218-add-environment-verify-with.yml
```diff
@@ -0,0 +1,13 @@
+databaseChangeLog:
+- changeSet:
+    id: add-environment-verify-with
+    author: fletch
+    changes:
+    - addColumn:
+        tableName: environment
+        columns:
+        - column:
+            name: verify_with
```
luispollo, line 10: I was torn between sticking with a field -> column naming convention here, or just use `verifications` to follow the pattern of `contraints` and `notifications` (although those do match their field names). I guess it doesn't matter that much, never mind.

gal-yardeni: +1 to `verifications`

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
robfletcher: No I think I accidentally did the format command. You can set it to automatically apply on commit but I don't like to because it creates a big delta. I'd rather isolate formatting changes into their own commit. I'll back the unnecessary stuff out here.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
robfletcher: Line 324

keel-sql/src/main/resources/db/changelog/20201218-add-environment-verify-with.yml
robfletcher: I was thinking we might want to unify everything into a single column so we don't fall into this trap again

robfletcher: Renamed the column as suggested

keel-sql/src/main/resources/db/changelog/20201218-add-environment-verify-with.yml
```diff
@@ -0,0 +1,21 @@
...
+    - update:
+        tableName: environment
+        columns:
+        - column:
+            name: verifications
+            value: []
+    - addNotNullConstraint:
+        tableName: environment
+        columnName: verifications
+        columnDataType: json
```
robfletcher, line 21: This caters for existing data by only adding the not-null constraint after updating existing rows. You can't have a `defaultValue` on a JSON column, apparently.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
luispollo: Ah, right!

keel-sql/src/main/resources/db/changelog/20201218-add-environment-verify-with.yml
luispollo: Why not null, though?

robfletcher: It's never null in the `Environment` class. It's an empty set.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/deliveryconfigs/deliveryconfigs.kt
```diff
@@ -74,16 +74,20 @@ internal fun SqlStorageContext.attachDependents(deliveryConfig: DeliveryConfig):
...
-            .fetch { (environmentUid, name, constraintsJson, notificationsJson) ->
+            .fetch { (environmentUid, name, constraintsJson, notificationsJson, verifyWithJson) ->
               Environment(
                 name = name,
                 resources = resourcesForEnvironment(environmentUid),
                 constraints = objectMapper.readValue(constraintsJson),
-                notifications = objectMapper.readValue(notificationsJson ?: "[]")
+                notifications = notificationsJson?.let { objectMapper.readValue(it) } ?: emptySet(),
+                verifyWith = verifyWithJson?.let {
+                  objectMapper.readValue(it)
```
lorin, line 89: Looking at this code makes me think: we could use some tests to ensure we can deserialize verification objects.

lorin (approved):
Looks good to me. But I'm starting to think that in the delivery config we should use `verifications` instead of `verifyWith`, to improve consistency.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/deliveryconfigs/deliveryconfigs.kt
robfletcher: Yeah, I wonder if we could automate that in a spring based test

PR #1719, opened 2020-12-19
fix sub-query that did not return a unique row

robfletcher:
I incorrectly assumed that `name` / `type` was a unique index on `delivery_artifact` but it is not. I've changed the query to use `delivery_config_name` and `reference` instead. I also upgraded that index to a unique one as it was not before, but it should be.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
@@ -58,7 +58,12 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
           DockerArtifact(
             name = "fnord",
             deliveryConfigName = "fnord-manifest",
-            reference = "fnord-docker"
+            reference = "fnord-docker-stable"
+          ),
+          DockerArtifact(
+            name = "fnord",
+            deliveryConfigName = "fnord-manifest",
+            reference = "fnord-docker-unstable"
           )
         ),
         environments = setOf(
@@ -69,7 +74,7 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
         )
       ),
       environmentName = "test",
-      artifactReference = "fnord-docker",
+      artifactReference = "fnord-docker-stable",
       version = "fnord-0.190.0-h378.eacb135"
     )
 
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
@@ -2,7 +2,6 @@ package com.netflix.spinnaker.keel.sql
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.keel.api.Verification
-import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.plugins.ArtifactSupplier
 import com.netflix.spinnaker.keel.api.verification.VerificationContext
 import com.netflix.spinnaker.keel.api.verification.VerificationRepository
@@ -130,7 +129,7 @@ class SqlVerificationRepository(
         )
         .from(VERIFICATION_STATE)
         .where(VERIFICATION_STATE.ENVIRONMENT_UID.eq(environmentUid))
-        .and(VERIFICATION_STATE.ARTIFACT_UID.eq(artifact.uid))
+        .and(VERIFICATION_STATE.ARTIFACT_UID.eq(artifactUid))
         .and(VERIFICATION_STATE.ARTIFACT_VERSION.eq(version))
         .and(VERIFICATION_STATE.VERIFICATION_ID.eq(verification.id))
         .fetchOneInto<VerificationState>()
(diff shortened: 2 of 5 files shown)


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
```diff
@@ -58,7 +58,12 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
...
             name = "fnord",
             deliveryConfigName = "fnord-manifest",
-            reference = "fnord-docker"
+            reference = "fnord-docker-stable"
+          ),
+          DockerArtifact(
+            name = "fnord",
+            deliveryConfigName = "fnord-manifest",
+            reference = "fnord-docker-unstable"
           )
```
robfletcher, line 67: Tested by adding multiple artifacts to the delivery config

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepositoryTests.kt
```diff
@@ -62,7 +62,7 @@ internal class SqlVerificationRepositoryTests :
     )
 
   override fun VerificationContext.setup() {
-    artifactRepository.register(artifact)
+    deliveryConfig.artifacts.forEach(artifactRepository::register)
```
robfletcher, line 65: This change makes it persist all artifacts in the delivery config rather than just the one we know we're interested in

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -193,9 +192,9 @@ class SqlVerificationRepository(
       .and(ENVIRONMENT.NAME.eq(environment.name))
       .and(ENVIRONMENT.DELIVERY_CONFIG_UID.eq(DELIVERY_CONFIG.UID))
 
-  private val DeliveryArtifact.uid: Select<Record1<String>>
+  private val VerificationContext.artifactUid: Select<Record1<String>>
     get() = select(DELIVERY_ARTIFACT.UID)
       .from(DELIVERY_ARTIFACT)
-      .where(DELIVERY_ARTIFACT.NAME.eq(name))
-      .and(DELIVERY_ARTIFACT.TYPE.eq(type))
+      .where(DELIVERY_ARTIFACT.DELIVERY_CONFIG_NAME.eq(deliveryConfig.name))
```
luispollo, line 198: We should really be using the delivery config uid in these tables consistently. Maybe not worth doing now, but thought I'd mention it while you're at it.

luispollo (approved):
LGTM, but I think we should replace the delivery config name with the uid in `delivery_artifacts` sooner rather than later (and make that an FK).

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
robfletcher: not sure what you mean, that's the thing we're trying to select here

robfletcher: Yeah, I think @emjburns had some reason why there's not a direct foreign key on this table. Can't remember now.

luispollo: My understanding was that that was only for `artifact_versions`, since those can be reused across multiple `deliver_artifact` records, but a `delivery_artifact` record is always tied to a specific `delivery_config` (hence the "soft" back reference here based on the name).

keel-sql/src/main/resources/db/changelog/20201219-add-artifact-unique-index.yml
```diff
@@ -0,0 +1,12 @@
+databaseChangeLog:
+- changeSet:
+    id: add-artifact-unique-index
+    author: fletch
+    changes:
+    - dropIndex:
+        tableName: delivery_artifact
+        indexName: delivery_artifact_byref_idx
```
lorin, line 8: I don't think this index should be dropped, it's doing what we want `VerificationContext.artifactUid` to do, indexing on `delivery_config_name` and `reference`. Here's the original: https://github.com/spinnaker/keel/blob/master/keel-sql/src/main/resources/db/changelog/20200127-environment-artifact-pin-table.yml#L49-L56

keel-sql/src/main/resources/db/changelog/20201219-add-artifact-unique-index.yml
```diff
@@ -0,0 +1,12 @@
...
+    - addUniqueConstraint:
+        tableName: delivery_artifact
+        columnNames: delivery_config_name, reference
+        constraintName: delivery_artifact_byref_idx
```
lorin, line 12: Since this is a constraint and not an index, it shouldn't have an `_idx` suffix.

lorin (changes requested):
I think the database migration is dropping an index that would make the new query faster. Everything else looks good to me.

keel-sql/src/main/resources/db/changelog/20201219-add-artifact-unique-index.yml
robfletcher: Doesn't adding a unique constraint automatically also create an index?

keel-sql/src/main/resources/db/changelog/20201219-add-artifact-unique-index.yml
robfletcher: It just needs to be `unique: true`. I updated the PR.

keel-sql/src/main/resources/db/changelog/20201219-add-artifact-unique-index.yml
```diff
@@ -0,0 +1,17 @@
...
+        indexName: delivery_artifact_byref_idx
+    - createIndex:
+        tableName: delivery_artifact
+        indexName: delivery_artifact_byref_idx
+        columns:
+        - column:
+            name: delivery_config_name
+        - column:
+            name: reference
+        unique: true
```
robfletcher, line 17: @lorin changed the constraint to just recreate the index, but with `unique: true`

lorin (approved):

PR #1721, opened 2020-12-20
prevent writing null verification state metadata

robfletcher:
Avoid accidentally writing `null` to the `verification_state.metadata` column (which does not allow null values). Also ensure that subsequent updates will add metadata.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
@@ -28,13 +28,13 @@ interface VerificationRepository {
   /**
    * Updates the state of [verification] as run against [context].
    *
-   * @param metadata `null` means "do not update the metadata".
+   * @param metadata if non-empty this will overwrite any existing metadata.
    */
   fun updateState(
     context: VerificationContext,
     verification: Verification,
     status: VerificationStatus,
-    metadata: Map<String, Any?>? = null
+    metadata: Map<String, Any?> = emptyMap()
   )
 
   fun nextEnvironmentsForVerification(minTimeSinceLastCheck: Duration, limit: Int) : Collection<VerificationContext>
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
@@ -161,7 +161,7 @@ class SqlVerificationRepository(
     context: VerificationContext,
     verification: Verification,
     status: VerificationStatus,
-    metadata: Map<String, Any?>?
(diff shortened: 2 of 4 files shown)


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
```diff
@@ -34,7 +34,7 @@ interface VerificationRepository {
     context: VerificationContext,
     verification: Verification,
     status: VerificationStatus,
-    metadata: Map<String, Any?>? = null
+    metadata: Map<String, Any?> = emptyMap()
```
robfletcher, line 37: made this parameter non-null so it doesn't get propagated to the database as `null`

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -175,6 +175,13 @@ class SqlVerificationRepository(
         .set(VERIFICATION_STATE.VERIFICATION_ID, verification.id)
         .onDuplicateKeyUpdate()
         .set(VERIFICATION_STATE.STATUS, status)
+        .run {
+          if (metadata.isNotEmpty()) {
+            set(VERIFICATION_STATE.METADATA, metadata)
+          } else {
+            this
+          }
+        }
```
robfletcher, line 184: This field was never getting set on update before. Ideally I'd like to use MySQL's `json_merge` function here so that it updates the document but I'm not sure how to do that with JOOQ. It's moot at the moment since the only thing currently written into this column is the Orca task id when running a container test job.

luispollo (approved):
Looks good to me.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
lorin: Comment at line 31 also needs updating. (Alas, GH doesn't let me annotate non-changed lines)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
lorin: I'd add a reminder comment here about the semantics of an empty metadata object, to help a future reader.

This is one of the things I find frustrating about the interaction of docstrings and subclassing. The semantics of an empty metadata object has already been documented on the `VerificationRepository` interface, but those comments aren't visible in this file.

lorin (approved):

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
robfletcher: Okay I changed it

PR #1722, opened 2020-12-21
still update environment_last_verified if delivery config load fails

robfletcher:
If the delivery config failed to load it would roll back the transaction and cause the verification loop to get stuck.

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
@@ -54,6 +54,7 @@ class SqlVerificationRepository(
     val now = clock.instant()
     val cutoff = now.minus(minTimeSinceLastCheck)
     return sqlRetry.withRetry(WRITE) {
+      // TODO: only consider environments that have verifications
       jooq.inTransaction {
         select(
           DELIVERY_CONFIG.UID,
@@ -100,19 +101,20 @@ class SqlVerificationRepository(
               .set(ENVIRONMENT_LAST_VERIFIED.ARTIFACT_VERSION, artifactVersion)
               .set(ENVIRONMENT_LAST_VERIFIED.AT, now)
               .onDuplicateKeyUpdate()
+              .set(ENVIRONMENT_LAST_VERIFIED.ARTIFACT_VERSION, artifactVersion)
               .set(ENVIRONMENT_LAST_VERIFIED.AT, now)
               .execute()
           }
-          .map { (_, deliveryConfigName, _, environmentName, _, artifactReference, artifactVersion) ->
-            VerificationContext(
-              deliveryConfigByName(deliveryConfigName),
-              environmentName,
-              artifactReference,
-              artifactVersion
-            )
-          }
       }
     }
+      .map { (_, deliveryConfigName, _, environmentName, _, artifactReference, artifactVersion) ->
+        VerificationContext(
+          deliveryConfigByName(deliveryConfigName),
+          environmentName,
+          artifactReference,
+          artifactVersion
+        )
+      }
   }
 
   override fun getState(
--- a/keel-sql/src/main/resources/db/changelog/20201221-shorten-environment-last-verified-pk.yml
+++ b/keel-sql/src/main/resources/db/changelog/20201221-shorten-environment-last-verified-pk.yml
@@ -1,0 +1,30 @@
+databaseChangeLog:
+- changeSet:
+    id: shorten-environment-last-verified-pk
+    author: fletch
+    changes:
+    - dropAllForeignKeyConstraints:
+        baseTableName: environment_last_verified
+    - delete:
+        tableName: environment_last_verified
+    - dropPrimaryKey:
+        tableName: environment_last_verified
+    - addForeignKeyConstraint:
+        baseTableName: environment_last_verified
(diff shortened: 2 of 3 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -100,19 +101,20 @@ class SqlVerificationRepository(
               .set(ENVIRONMENT_LAST_VERIFIED.ARTIFACT_VERSION, artifactVersion)
               .set(ENVIRONMENT_LAST_VERIFIED.AT, now)
               .onDuplicateKeyUpdate()
+              .set(ENVIRONMENT_LAST_VERIFIED.ARTIFACT_VERSION, artifactVersion)
```
robfletcher, line 104: because I removed this from the PK we need to update this column

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -100,19 +101,20 @@ class SqlVerificationRepository(
...
       }
     }
+      .map { (_, deliveryConfigName, _, environmentName, _, artifactReference, artifactVersion) ->
+        VerificationContext(
+          deliveryConfigByName(deliveryConfigName),
+          environmentName,
+          artifactReference,
+          artifactVersion
+        )
+      }
```
robfletcher, line 117: Moved this outside of the transaction block so if it throws an exception it doesn't cause the transaction to roll back.

keel-sql/src/main/resources/db/changelog/20201221-shorten-environment-last-verified-pk.yml
```diff
@@ -0,0 +1,30 @@
+databaseChangeLog:
+- changeSet:
+    id: shorten-environment-last-verified-pk
```
robfletcher, line 3: It doesn't make a lot of sense to include `artifact_version` in the PK for this table as it's always just the `CURRENT` version in the environment.

keel-sql/src/main/resources/db/changelog/20201221-shorten-environment-last-verified-pk.yml
```diff
@@ -0,0 +1,30 @@
...
+    author: fletch
+    changes:
+    - dropAllForeignKeyConstraints:
+        baseTableName: environment_last_verified
+    - delete:
+        tableName: environment_last_verified
+    - dropPrimaryKey:
+        tableName: environment_last_verified
```
robfletcher, line 11: Need to delete the content of the table as it may violate the shortened primary key. Also, I got an error dropping the PK if I didn't drop the FKs first.

keel-sql/src/main/resources/db/changelog/20201221-shorten-environment-last-verified-pk.yml
```diff
@@ -0,0 +1,30 @@
...
+        referencesUniqueColumn: true
+        onDelete: CASCADE
+    - addForeignKeyConstraint:
+        baseTableName: environment_last_verified
+        baseColumnNames: artifact_uid
+        constraintName: fk_environment_last_verified_delivery_artifact
+        referencedTableName: delivery_artifact
+        referencedColumnNames: uid
+        referencesUniqueColumn: true
+        onDelete: CASCADE
```
robfletcher, line 27: restore the FKs as they were

keel-sql/src/main/resources/db/changelog/20201221-shorten-environment-last-verified-pk.yml
```diff
@@ -0,0 +1,30 @@
...
+    - addPrimaryKey:
+        tableName: environment_last_verified
+        columnNames: environment_uid, artifact_uid
```
robfletcher, line 30: create a shorter PK

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
luispollo: Good catch!

robfletcher: Trying to figure out a good way to test it

luispollo (approved):
LGTM!

lorin (approved):
My understanding

## Problem

The `SqlVerificationRepository.nextEnvironmentsForVerification` method would do an insert and then a select within the same transaction. If the select fails, then the transaction would abort, and the `environment_last_verified` table would not get updated.

This is problematic because it means keel can get stuck, where it keeps trying to update a record for an environment, and there's a failure, and then it tries again, so it can never make progress.

We hit this in a situation where the select was failing because of a plugin that wasn't loaded.

## Solution

Move the select query (`SqlStorageContext.deliveryConfigByName` extension function) outside of the transaction so that the insert getes committed even if the select fails.

This means that when the select query fails for a particular delivery config, the environment is still marked as being updated, which enables keel to make progress.

## Other changes

The artifact version is no longer a component of the primary key of the `environment_last_verified` table. To make this change, there's a migration that drops the contents of the table, foreign keys, and primary key, and then recreates the foreign keys and the primary key.

## Feedback

looks good to me.

PR #1723, opened 2020-12-22
fix insert statement that doesn't always set required column

robfletcher:
Not clear why this doesn't happen in tests using MySQL testcontainer. It seems like an insert statement that does not set the required column `started_at` is rejected. Even though in practice if we weren't setting `started_at` we'd always be falling back to the `on duplicate key update` clause.

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
@@ -170,13 +170,27 @@ class SqlVerificationRepository(
         .insertInto(VERIFICATION_STATE)
         .set(VERIFICATION_STATE.STATUS, status)
         .set(VERIFICATION_STATE.METADATA, metadata)
-        .set(status.timestampColumn, currentTimestamp())
+        .set(VERIFICATION_STATE.STARTED_AT, currentTimestamp())
+        .run {
+          if (status.complete) {
+            set(VERIFICATION_STATE.ENDED_AT, currentTimestamp())
+          } else {
+            setNull(VERIFICATION_STATE.ENDED_AT)
+          }
+        }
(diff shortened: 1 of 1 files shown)


luispollo (approved):
LGTM. It feels like you're close, but I'd again suggest doing some end-to-end local testing before rolling out, in case there's anything else missing or left to fix.

robfletcher:
Local testing with a clone of the prestaging database did not exhibit this issue. I can only think it’s an issue specific to Aurora.

lorin (approved):
Looks fine to me.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -170,13 +170,27 @@ class SqlVerificationRepository(
         .insertInto(VERIFICATION_STATE)
         .set(VERIFICATION_STATE.STATUS, status)
         .set(VERIFICATION_STATE.METADATA, metadata)
-        .set(status.timestampColumn, currentTimestamp())
```
luispollo, line 173: On second thought, shouldn't this be set only in the insert (never in an update)? BTW, I didn't quite follow the comment on the duplicate key, since the insert would never even run without a value for the `started_at` column. My guess is that it passed in the tests because the way you set them up you'd never do an insert where that column would not be set (the first call to `updateState` always has a non-complete status, so the value was set).

luispollo (commented):
Late comment, sorry.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
luispollo: Wrong line, should be on line below.

luispollo:
> Local testing with a clone of the prestaging database did not exhibit this issue. I can only think it’s an issue specific to Aurora. 

You mentioned `testcontainers` in your comments, so I wasn't sure.

PR #1725, opened 2021-01-04
allow specification of application for test container cluster

robfletcher:
Our test suite fails because it runs using its own application rather than as part of keel.

### DIFF
--- a/keel-titus-api/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/TestContainerVerification.kt
+++ b/keel-titus-api/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/TestContainerVerification.kt
@@ -10,7 +10,7 @@ data class TestContainerVerification(
   val application: String? = null
 ) : Verification {
   override val type = TYPE
-  override val id = "$repository/$tag"
+  override val id = "$repository:$tag"
 
   companion object {
     const val TYPE = "test-container"
--- a/keel-titus-api/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/TestContainerVerification.kt
+++ b/keel-titus-api/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/TestContainerVerification.kt
@@ -6,7 +6,8 @@ import com.netflix.spinnaker.keel.api.titus.TitusServerGroup.Location
 data class TestContainerVerification(
   val repository: String,
   val tag: String = "latest",
-  val location: Location
+  val location: Location,
+  val application: String? = null
 ) : Verification {
   override val type = TYPE
   override val id = "$repository/$tag"
--- a/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
+++ b/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
@@ -25,7 +25,6 @@ data class ContainerJobConfig(
     networkMbps = 1024
   ),
   val capacity: Capacity = Capacity(min = 1, max = 1, desired = 1),
-  val serviceAccount: String,
   val credentials: String,
   val entrypoint: String = "",
   val runtimeLimit: Duration = Duration.ofSeconds(2700),
--- a/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/TestContainerVerificationEvaluator.kt
+++ b/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/TestContainerVerificationEvaluator.kt
@@ -66,17 +66,16 @@ class TestContainerVerificationEvaluator(
         taskLauncher.submitJob(
           type = VERIFICATION,
           subject = "container integration test for ${context.deliveryConfig.application}.${context.environmentName}",
-          description = "Verifying ${context.version} in environment ${context.environmentName} with test container ${verification.repository}/${verification.tag}",
+          description = "Verifying ${context.version} in environment ${context.environmentName} with test container ${verification.repository}:${verification.tag}",
           correlationId = "${supportedVerification.first}:${context.deliveryConfig.application}.${context.environmentName}",
           user = context.deliveryConfig.serviceAccount,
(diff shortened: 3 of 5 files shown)


keel-titus-api/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/TestContainerVerification.kt
```diff
@@ -6,7 +6,8 @@ import com.netflix.spinnaker.keel.api.titus.TitusServerGroup.Location
 data class TestContainerVerification(
   val repository: String,
   val tag: String = "latest",
-  val location: Location
+  val location: Location,
+  val application: String? = null
```
robfletcher, line 10: new optional property on the verification itself

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -25,7 +25,6 @@ data class ContainerJobConfig(
     networkMbps = 1024
   ),
   val capacity: Capacity = Capacity(min = 1, max = 1, desired = 1),
-  val serviceAccount: String,
```
robfletcher, line 28: I noticed this was never propagated to Orca so I figured we can remove it

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/TestContainerVerificationEvaluator.kt
```diff
@@ -66,17 +66,16 @@ class TestContainerVerificationEvaluator(
...
-          description = "Verifying ${context.version} in environment ${context.environmentName} with test container ${verification.repository}/${verification.tag}",
+          description = "Verifying ${context.version} in environment ${context.environmentName} with test container ${verification.repository}:${verification.tag}",
           correlationId = "${supportedVerification.first}:${context.deliveryConfig.application}.${context.environmentName}",
           user = context.deliveryConfig.serviceAccount,
           application = context.deliveryConfig.application,
           notifications = emptySet(),
           stages = listOf(
             ContainerJobConfig(
-              application = context.deliveryConfig.application,
+              application = verification.application ?: context.deliveryConfig.application,
```
robfletcher, line 76: use verification's application or default back to the one for the delivery config

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/TestContainerVerificationEvaluator.kt
```diff
@@ -66,17 +66,16 @@ class TestContainerVerificationEvaluator(
         taskLauncher.submitJob(
           type = VERIFICATION,
           subject = "container integration test for ${context.deliveryConfig.application}.${context.environmentName}",
-          description = "Verifying ${context.version} in environment ${context.environmentName} with test container ${verification.repository}/${verification.tag}",
+          description = "Verifying ${context.version} in environment ${context.environmentName} with test container ${verification.repository}:${verification.tag}",
```
lorin, line 69: Nit: note that the `TestContainerVerification.id` field uses a slash to delimit repository and tag.

```
  override val id = "$repository/$tag"
```

Would be good to be consistent.

lorin (approved):
## My understanding

### Context

The internal container we use for testing keel (ballast) uploads delivery configs to keel in our staging environment.

### Problem

When we executed this container via the test-container verification mechanism, the calls from our container to keel failed authorization checks, because it was trying to upsert the delivery config to the "ballast" app. However, when keel launches a container via Orca, it uses the application under test as the application name (in this case, keel was testing itself, so "keel" was the application name). Because the keel app doesn't have permissions to upsert delivery configs to the ballast app, when executing the ballast container, when ballast tried to upsert a delivery config for the "ballast" appp, keel returned an error that looks like this:

```
User keel-xxxxxxxxxxxx does not have access to the ballast-xxxxxxxxxxxx service account specified in the request
```

### Proposed solution

Add support for overriding the application that keel associates with the container when launching the container via orca. This will allow us to associate ballast container test runs with the ballast account, even though the verification is beign done against the keel app.

## Feedback

I picked a very small nit, looks good otherwise.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/TestContainerVerificationEvaluator.kt
robfletcher: Good spot. I changed it because of the way Deck renders it in the run job stage configuration view. I'll change it there as well so it's consistent in all 3 places.

luispollo (approved):
LGTM. Would be nice to add test coverage.

keel-titus-plugin/src/test/kotlin/com/netflix/spinnaker/keel/titus/verification/TestContainerVerificationEvaluatorTests.kt
```diff
@@ -86,6 +80,34 @@ internal class TestContainerVerificationEvaluatorTests {
...
+
+    subject.start(context, verification.copy(application = null))
+
+    expectThat(job).withCaptured {
+      hasSize(1)
+        .first()
+        .get("cluster").isA<Map<String, Any?>>()
+        .get("application") isEqualTo context.deliveryConfig.application
+    }
+  }
```
robfletcher, line 109: Test that `cluster/application` is set correctly in the job submitted to Orca

keel-titus-api/src/main/kotlin/com/netflix/spinnaker/keel/api/titus/TestContainerVerification.kt
```diff
@@ -6,10 +6,11 @@ import com.netflix.spinnaker.keel.api.titus.TitusServerGroup.Location
...
 ) : Verification {
   override val type = TYPE
-  override val id = "$repository/$tag"
+  override val id = "$repository:$tag"
```
robfletcher, line 13: Changed for consistency as pointed out by @lorin

PR #1727, opened 2021-01-05
Fix autobumped kork version

robfletcher:
Fixes #1726


keel-sql/keel-sql.gradle.kts
```diff
@@ -122,7 +121,7 @@ afterEvaluate {
 }
 
 jooq {
-  version.set("3.14.4")
+  version.set("3.13.2")
```
luispollo, line 124: Isn't that the same version we were using before the latest upgrade? Any incompatibilities moving back?

robfletcher: That's what the other changes fix. Nothing we were using. @ajordens rolled it back because it was causing an issue in Clouddriver.

PR #1729, opened 2021-01-08
feat(delivery-config): use list instead of set for verifications

lorin:
## Context

The delivery config supports specifying multiple verificiations. Here's a hypothetical example with two verifications, `test-container` and `wizbang`.

```
environments:
- name: staging
  verifyWith:
  - type: test-container
    repository: acme/mytests
    tag: stable
  - type: wizbang
    power: max
```

## Problem

We expect that there will be scenarios where users will want the verifications to execute sequentially, in the order in which they are specified in the delivery config.

Currently, keel internally represents the collection of verifications as a set, so the ordering is lost.

## Proposed solution

Use a list instead of a set so the ordering in the delivery config is preserved.


robfletcher (approved):
Makes sense to me

PR #1730, opened 2021-01-08
feat(status): add waiting status

emjburns:
Back end for https://github.com/spinnaker/deck/pull/8836

Introducing a new status called "waiting" for when a resource was created and then has an resource check unresolvable status. This happens primarily (maybe only) to clusters when there is no approved image. I am updating the status so new customers don't immediately see an error.

I'll re-evaluate this in the future if it turns out that other resources w/o artifacts also hit this condition.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
@@ -30,6 +30,7 @@ import com.netflix.spinnaker.keel.persistence.ResourceStatus.PAUSED
 import com.netflix.spinnaker.keel.persistence.ResourceStatus.RESUMED
 import com.netflix.spinnaker.keel.persistence.ResourceStatus.UNHAPPY
 import com.netflix.spinnaker.keel.persistence.ResourceStatus.UNKNOWN
+import com.netflix.spinnaker.keel.persistence.ResourceStatus.WAITING
 import org.springframework.stereotype.Component
 
 /**
@@ -66,6 +67,7 @@ class ResourceStatusService(
       history.isError() -> ERROR
       history.isCreated() -> CREATED
       history.isResumed() -> RESUMED
+      history.isWaiting() -> WAITING // must be before CURRENTLY_UNRESOLVABLE because it's a special case of that status
       history.isCurrentlyUnresolvable() -> CURRENTLY_UNRESOLVABLE
       else -> UNKNOWN
     }
@@ -91,6 +93,12 @@ class ResourceStatusService(
 
   private fun List<ResourceHistoryEvent>.isCreated(): Boolean {
     return first() is ResourceCreated
+  }
+
+  private fun List<ResourceHistoryEvent>.isWaiting(): Boolean {
+    // we expect to have only two events, but we will accept several different "unresolvable" events
+    // in order to be less brittle and show the user this status instead of an error
+    return size < 5 && all { it is ResourceCreated || it is ResourceCheckUnresolvable }
   }
 
   private fun List<ResourceHistoryEvent>.isDiff(): Boolean {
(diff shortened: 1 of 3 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
```diff
@@ -93,6 +95,12 @@ class ResourceStatusService(
     return first() is ResourceCreated
   }
 
+  private fun List<ResourceHistoryEvent>.isWaiting(): Boolean {
+    // we expect to have only two events, but we will accept several different "unresolvable" events
+    // in order to be less brittle and show the user this status instead of an error
+    return all { it is ResourceCreated || it is ResourceCheckUnresolvable } && size < 5
```
luispollo, line 101: Maybe reorder this so you don't actually look at a bunch of events unnecessarily?
```suggestion
    return size < 5 && all { it is ResourceCreated || it is ResourceCheckUnresolvable } 
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusServiceTests.kt
```diff
@@ -214,12 +215,27 @@ class ResourceStatusServiceTests : JUnit5Minutests {
...
+              }
+
+              test("returns WAITING") {
+                expectThat(subject.getStatus(resource.id)).isEqualTo(WAITING)
+              }
             }
 
-            test("returns CURRENTLY_UNRESOLVABLE") {
-              expectThat(subject.getStatus(resource.id)).isEqualTo(CURRENTLY_UNRESOLVABLE)
+            context("more than created and unresolvable events") {
```
luispollo, line 230: You're missing a test for your `size < 5` condition if you want to cover all cases.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
```diff
@@ -93,6 +95,12 @@ class ResourceStatusService(
     return first() is ResourceCreated
   }
 
+  private fun List<ResourceHistoryEvent>.isWaiting(): Boolean {
+    // we expect to have only two events, but we will accept several different "unresolvable" events
+    // in order to be less brittle and show the user this status instead of an error
+    return size < 5 && all { it is ResourceCreated || it is ResourceCheckUnresolvable }
```
robfletcher, line 101: The `5` seems arbitrary here. What's the thinking there?

emjburns:
@luispollo addressed your comments! Thanks.

robfletcher (approved):
I'm curious why 5, but otherwise LGTM

lorin (approved):
My understanding

## Context

If there are no approved images associated with a managed cluster, the cluster will be in a `CURRENTLY_UNRESOLVABLE` state.

## Problem

When a user deploys a new app using Managed Delivery and looks at the environment view, their first environment will have a cluster with zero approved images, which will put it in a `CURRENTLY_UNRESOLVABLE` state. 

This is a poor user experience for new users, because the way that `CURRENTLY_UNRESOLVABLE` is rendered in the UI, it looks like something is broken, even though this is expected behavior.

## Proposed solution

When a cluster resource has zero approved images, show its status as `WAITING` instead of `CURRENTLY_UNRESOLVABLE`. The new `WAITING` status is rendered in the UI in such a way that it communicates that this is expected behavior.

The way this is implemented is that a cluster resource is in the WAITING state if the only events in its history are:
* resource was created
* resource check is unresolvable


## Feedback

I agree with @robfletcher's comment that the threshold of `5` events feels arbitrary, otherwise looks good to me.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
luispollo: +1. Maybe make it a constant with a self-explanatory name?

luispollo (approved):
Other than explaining why the `5` to cap events as @robfletcher mentioned, LGTM.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
emjburns: Totally arbitrary. I feel like if the environment is actually new it'll only have a couple of events. I wanted to make sure this new status only shows up in cases where it is definitely new. So, I chose 5. Does that make sense / seem reasonable?

gal-yardeni (approved):
LGTM! my only comment is to remember adding a doc explaining this new status :)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ResourceStatusService.kt
emjburns: @robfletcher @luispollo I was hoping my comment of `// we expect to have only two events, but we will accept several different "unresolvable" events` would convey why I chose 5, is there a way I could make this comment more clear?

PR #1732, opened 2021-01-11
use existing ConstraintStatus enum for verifications

robfletcher:
In the interests of keeping things small, I thought I'd break this work up. First step is to re-use the enum we already use for constraints. I'm contemplating renaming it but haven't decided to what yet.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
@@ -3,12 +3,12 @@ package com.netflix.spinnaker.keel.persistence
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.Verification
+import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus
+import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus.PASS
+import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus.PENDING
 import com.netflix.spinnaker.keel.api.verification.VerificationContext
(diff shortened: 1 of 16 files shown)


emjburns (approved):

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
```diff
@@ -158,17 +158,17 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
 
     context.setup()
 
-    subject.updateState(context, verification, RUNNING)
+    subject.updateState(context, verification, PENDING)
```
luispollo, line 161: Hmm, while I like the idea of DRYing things up, I think `RUNNING` and `PENDING` mean two different things.

luispollo:
> I'm contemplating renaming it but haven't decided to what yet.

`EXECUTION_STATUS`?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
robfletcher: That would mean changing existing constraints to also use that new status consistently

lorin (approved):
My understanding

## Context

We eventually want to expose a set of API calls to allow users to override verification behaviors, such as:

* skipping a verification
* re-running a failed verification
* canceling a verification

We currently support API calls for constraints (e.g., overriding), and API calls for updating the state of verifications will be similar.

## Work in this PR

This PR replaces the VerificationStatus enum by reusing the existing ConstraintStatus enum. This is groundwork for exposing the APIs discussed above. 

The values for the two enums aren't identical, so this PR maps from VerificationStatus to ConstraintStatus, as follows:

```
PASSED -> PASS
RUNNING -> PENDING
FAILED -> FAIL
```

## Feedback

looks good to me

PR #1733, opened 2021-01-12
feat(slack): adding more notifications using slack SDK

gal-yardeni:
**Context**

Currently, we utilize `echo` to send slack notifications to users, and we don't have much control over the payload, interactive notifications or the new features slack provides -- which makes it very hard to add/modify notifications related to `keel`.

**Work in this PR**

1. Adding the new slack SDK to `keel` so we can build notifications when needed, upon demand. 
The current model, in this PR, will send an event like `SlackPinnedNotification`, `SlackUnpinnedNotification` etc. whenever the event is happening (can be anywhere in the code). We'll keep all those event listeners under the new `slack` module, where we will construct the actual notification.

2. Adding new notifications for `pinning` and `unpinning` to verify this new model + added more repository functions as needed.

3. [new] added a `NotifiactionEventListener`, which will encapsulate all events, and fetch the right information prior to constructing the actual slack notifications.

**Future work** (in following PRs)

- [ ] Deprecate notifications coming from `orca`
- [ ] Adding more `keel` related notifications (like: app is paused, deployment succeeded, bake/build failed..) + 
fix the code whenever we see common patterns when fetching data
- [ ] Migrating the "manual judgment" (or: interactive notifications) to the new implementation

**Looking for feedback**
Is the new implementation makes sense? (sending events from all over, listen in the slack module, and construct the notification there)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/SlackNotificationEvent.kt
```diff
@@ -0,0 +1,62 @@
+package com.netflix.spinnaker.keel.events
+
+import com.netflix.spinnaker.keel.api.NotificationConfig
+import com.netflix.spinnaker.keel.api.artifacts.PublishedArtifact
+import com.netflix.spinnaker.keel.core.api.EnvironmentArtifactPin
+import com.netflix.spinnaker.keel.notifications.NotificationType
+import java.time.Instant
+
+abstract class SlackNotificationEvent {
```
gal-yardeni, line 9: All slack related events will be defined here.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -104,14 +109,55 @@ class ApplicationService(
 
   fun pin(user: String, application: String, pin: EnvironmentArtifactPin) {
     val config = repository.getDeliveryConfigForApplication(application)
-    repository.pinEnvironment(config, pin.copy(pinnedBy = user))
-    // TODO: publish ArtifactPinnedEvent
+    val time = repository.pinEnvironment(config, pin.copy(pinnedBy = user))
```
gal-yardeni, line 112: I might be able to refactor notifications when there are common things as adding more notifications.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -204,7 +207,7 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
   /**
    * Pin an environment to only deploy a specific DeliveryArtifact version
    */
-  fun pinEnvironment(deliveryConfig: DeliveryConfig, environmentArtifactPin: EnvironmentArtifactPin)
+  fun pinEnvironment(deliveryConfig: DeliveryConfig, environmentArtifactPin: EnvironmentArtifactPin): Instant
```
gal-yardeni, line 210: Changed some method signatures to get information back from the DB, to construct the notifications.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/notifications/PinnedNotification.kt
```diff
@@ -0,0 +1,85 @@
...
+          text("$env is pinned", emoji = true)
+        }
+
+        section {
+          markdownText("*Version:* ~#${event.currentArtifact?.buildMetadata?.number}~ → <$pinnedArtifactUrl|#${pinnedArtifact?.buildMetadata?.number}> " +
+            "by ${pinnedArtifact?.gitMetadata?.author}\n " +
+            "*Where:* $env\n\n " +
+            "${pinnedArtifact?.gitMetadata?.commitInfo?.message}")
+          accessory {
+            image(imageUrl = "https://raw.githubusercontent.com/gcomstock/managed.delivery/master/src/icons/pinned.png", altText = "pinned")
```
lorin, line 44: My instinct is that we should avoid hotlinking to GitHub like this, but I don't know what the common pattern is for linking images in Slack notifications.

lorin (commented):
# My understanding

## Context

Slack notifications are currently implemented in echo: keel sends an event to echo, and echo calls the Slack API. 

## Problem

The current implementation makes it harder to iterate quickly on changes to Slack notifications, because it involves changing two codebases: keel and echo.

## Proposed solution

Move the logic that calls the Slack API from echo to keel.

## Code changes in this PR

Changes in this PR:

* Add logic for calling the Slack API
* Add two new events (pin, unpin) that get published by Spring framework when those actions are taken
* Send a notification to Slack when pin happens

### New module: keel-slack

The new keel-slack module is responsible for:

* listening for events that we want to notify users about via Slack
* calling the Slack API with the information about the notifications

There is a separate class per event handler. These classes contain the logic for constructing the Slack messages.
* SlackPinnedNotification event handler: PinnedNotification ()
* SlackUnpinnedNotification event handler: UnpinnedNoitification (placeholder for now)

* config/SlackConfiguration.kt - auth info needed to call the API
* keel/slack/SlackNotifier.kt - thin wrapper around Slack's Kotlin API

### Repository changes

* Repository classes have been augmented to return timestamp information that keel needs for populating the pinned/unpinned events.
* DeliveryConfig repository returns notification configs, which have information about what events to notify on and what channels to notify on

### Core/event changes

* services/ApplicationService.kt - pin/deletePin methods publish events
events/SlackNotificationEvent.kt - new event classes (SlackPinnedNotification, SlackUnpinnedNotification)
notifications/NotificationType.kt - adds PINNED,UNPINNED as new types of notifications

# My feedback

My main concern with the implementation is that new event classes (SlackPinnedNotification, SlackUnpinnedNotification) mix two concerns that should be separate:
* what the event is (e.g., pin/unpin)
* the channel used to notify the user of the event (e.g., Slack)

I don't think these events should have Slack-specific information in them. Instead, I think the event handlers that call the Slack API should be responsible for owning the Slack configuration details.

I also left some inline comments, but they're all minor things.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/persistence/KeelReadOnlyRepository.kt
```diff
@@ -19,6 +20,8 @@ interface KeelReadOnlyRepository {
 
   fun environmentFor(resourceId: String): Environment
 
+  fun environmentNotifications(deliveryConfigName: String, environmentName: String): Set<NotificationConfig>?
```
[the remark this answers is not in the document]
gal-yardeni: good question. it's up to the implementation I guess - are you thinking that we should return null instead of an empty set if it does not exist?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/SlackNotificationEvent.kt
```diff
@@ -0,0 +1,62 @@
...
+  open val notificationConfig: NotificationConfig? = null
+  abstract val type: NotificationType
+}
+
+
+data class SlackPinnedNotification(
+  override val notificationConfig: NotificationConfig,
+  val pin: EnvironmentArtifactPin,
+  val currentArtifact: PublishedArtifact?,
+  val pinnedArtifact: PublishedArtifact?,
```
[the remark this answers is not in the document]
gal-yardeni: nope, good point. will change it.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/notifications/PinnedNotification.kt
gal-yardeni: good point, let's ask @erikmunson :)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/SlackNotificationEvent.kt
```diff
@@ -0,0 +1,62 @@
...
+  val time: Instant,
+  val application: String,
+  override val type: NotificationType
+) :SlackNotificationEvent() {
+  constructor(notificationConfig: NotificationConfig,
```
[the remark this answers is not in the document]
gal-yardeni: thanks, made the change!

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -1156,5 +1157,37 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
...
+          reference = versionedReleaseDebian.reference,
+          version = version1,
+          null,
+          null
+          ))
+      }
+
+      test("delete pin returns the uid, and successfully retrieve a delivery artifact object for existing artifactUid") {
+        val pair = subject.deletePin(manifest, testEnvironment.name)
+        if (pair != null) {
```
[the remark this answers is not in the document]
gal-yardeni: done

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/persistence/KeelReadOnlyRepository.kt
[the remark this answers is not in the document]
gal-yardeni: If we look at the `Environment` class, we'll see that we don't have any nullable fields there:
```
data class Environment(
  val name: String,
  val resources: Set<Resource<*>> = emptySet(),
  val verifyWith: List<Verification> = emptyList(),
  val notifications: Set<NotificationConfig> = emptySet() // applies to each resource
)
```
So I think returning an empty set makes sense in this case.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/SlackNotificationEvent.kt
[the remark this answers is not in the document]
emjburns: I would expect the second. pinning should succeed and be completely decoupled from sending the notification

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -104,14 +109,55 @@ class ApplicationService(
...
+    val envNotifications = repository.environmentNotifications(config.name, pin.targetEnvironment)
+    val deliveryArtifact = pin.reference.let {
+      repository.getArtifact(config.name, it)
+    }
+    val pinnedArtifact = repository.getArtifactVersion(deliveryArtifact, pin.version, null)
+    val currentArtifact = repository.getArtifactVersionByPromotionStatus(config, pin.targetEnvironment, deliveryArtifact, CURRENT)
+
+    envNotifications?.forEach { notificationConfig ->
+      if (notificationConfig.type == NotificationType.slack) {
+        publisher.publishEvent(SlackPinnedNotification(
```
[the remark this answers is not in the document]
emjburns: Agreed. I think that a `PinnedNotification` should have just the information needed to look up the other stuff (like the `EnvironmentArtifactPin` and the application) and then there should be another service that that does all the translating. The patter we have in other places is that there is some general thing that looks up the information (like the extra details for the notification) and then has a list of specific things (in this case, the notification providers) that can take the details and apply them to the specific quirks of the implementation (like the correct slack fields).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/SlackNotificationEvent.kt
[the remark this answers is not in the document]
lorin: I agree with @emjburns: I see notifications as a "best effort" thing. We should never fail an operation just because notification failed.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEventListener.kt
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.context.event.EventListener
+import org.springframework.stereotype.Component
+import java.time.Clock
+
+/**
+ * Responsible to listen to notification events, and fetch the information needed
+ * for sending a notification, based on NotificationType.
+ */
+@Component
+class NotificationEventListener(
```
gal-yardeni, line 17: encapsulate all notifications related information into this class.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/SlackNotificationEvent.kt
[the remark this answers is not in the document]
gal-yardeni: done!

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
```diff
@@ -75,12 +75,12 @@ class AdminController(
...
 
-  @PostMapping(
-    path = ["/notifications/test"]
-  )
-  fun sendNotificationEvent(
-    @RequestBody event: NotificationEvent
-  ) {
-    adminService.sendNotificationEvent(event)
-  }
+//  @PostMapping(
```
gal-yardeni, line 78: @emjburns can we remove it for now? I feel it's a bit confusing with the new implemention.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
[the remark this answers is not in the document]
gal-yardeni: Done!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEvent.kt
```diff
@@ -1,13 +1,39 @@
...
-  val message: Notification
-)
+abstract class NotificationEvent{
+  abstract val scope: NotificationScope
+  open val ref: String? = null
+  abstract val type: NotificationType
+  open val message: Notification? = null
+}
 
+data class UnhealthyNotificationEvent(
```
gal-yardeni, line 17: @emjburns calling for your attention here -- I've created a new data class specific to the `UnhealthyNotificationEvent`. I might change the implementation in the `Notifier` class when I'll remove the echo dependencies later on.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEventListener.kt
lorin: I think this is definitely an improvement.

The question I have is: why not move this logic directly to the `keel.slack.notifications.*` event listeners functions, so that those functions listen directly for `PinnedNotification` events?

gal-yardeni: I thought it might be better to separate the logic between two pieces - one (NotificationEventListener) will be responsible to fetch the relevant data, and the second (events under the `slack` module) will just construct the slack specific notification per event. 
I'm thinking of a couple of things that might support this theory - 
1. we can utilize it to send emails in the future
2. we need a central place to control which notifications are being sent according to user-defined frequency, and I _think_ having it under `keel-core` will make more sense than under `keel-slack` module.

I can also refactor it in the near future if we feel it's overhead, as I still need to do a bunch of changes in the code. Let me know if it makes sense to you, we can also have a quick chat about it.

gal-yardeni: I thought about it a lot and I agree. It doesn't make sense to store in two places the data needed for each event. Will move it directly to the `keel-slack` module.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -105,13 +110,18 @@ class ApplicationService(
   fun pin(user: String, application: String, pin: EnvironmentArtifactPin) {
     val config = repository.getDeliveryConfigForApplication(application)
     repository.pinEnvironment(config, pin.copy(pinnedBy = user))
-    // TODO: publish ArtifactPinnedEvent
+    publisher.publishEvent(PinnedNotification(config, pin))
```
emjburns, line 113: so much better! love.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/AdminController.kt
emjburns: Sure!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/SlackNotificationHandler.kt
```diff
@@ -0,0 +1,11 @@
+package com.netflix.spinnaker.keel.slack
+
+import com.netflix.spinnaker.keel.events.SlackNotificationEvent
+import com.netflix.spinnaker.keel.notifications.NotificationType
+
+
```
emjburns, line 6: Could you add class comments here?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEvent.kt
```diff
@@ -1,13 +1,40 @@
...
 
-data class NotificationEvent(
-  val scope: NotificationScope,
-  val ref: String,
-  val type: NotificationType,
-  val message: Notification
-)
+abstract class NotificationEvent{
+  abstract val scope: NotificationScope
+  open val ref: String? = null
```
emjburns, line 12: why the change to `ref: String?`?

emjburns: same question for message

gal-yardeni: I've changed it to fit the new notifiactionEvent module, where you don't necessarily have the ref and message (we construct the message in the slack module, for instance).

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/SlackNotificationHandler.kt
gal-yardeni: done!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -0,0 +1,87 @@
...
+  fun onPinnedNotification(notification: PinnedNotification) {
+    with(notification) {
+      val envNotifications = repository.environmentNotifications(config.name, pin.targetEnvironment)
+      val deliveryArtifact = pin.reference.let {
+        repository.getArtifact(config.name, it)
+      }
+      val pinnedArtifact = repository.getArtifactVersion(deliveryArtifact, pin.version, null)
+      val currentArtifact = repository.getArtifactVersionByPromotionStatus(config, pin.targetEnvironment, deliveryArtifact, PromotionStatus.CURRENT)
+
+      //handlers.supporting<SlackPinnedNotification>(type)
```
emjburns, line 38: remove?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/SlackNotificationHandler.kt
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Implement this interface to send different types of slack notifications. Each notification is being consturct by [constructMessage]
+ * See: [PinnedNotificationHandler] for example
+ */
+
+interface SlackNotificationHandler <T: SlackNotificationEvent> {
+  val type: NotificationType
+
+  fun constructMessage(notification: T)
+
```
emjburns, line 15: remove?

keel-slack/src/test/kotlin/com/netflix/spinnaker/keel/slack/SlackNotifierTest.kt
```diff
@@ -0,0 +1,73 @@
...
+      )
+    }
+
+
+  fun tests() = rootContext<Fixture> {
+    fixture {
+      Fixture()
+    }
+
+    //TODO: fix test
```
emjburns, line 30: fix?

keel-slack/src/test/kotlin/com/netflix/spinnaker/keel/slack/handlers/NotificationEventListenerTests.kt
```diff
@@ -0,0 +1,103 @@
...
+        every {
+          repository.getArtifactVersionByPromotionStatus(any(), any(), any(), any())
+        } returns versions.toArtifactVersions(releaseArtifact).first()
+
+        every {
+          repository.latestVersionApprovedIn(any(), any(), any())
+        } returns versions.last()
+
+      }
+
```
emjburns, line 96: fix/remove?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/notifications/PinnedNotification.kt
luispollo: We could hotlink to Spinnaker UI assets from deck instead. 🙂

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/persistence/KeelReadOnlyRepository.kt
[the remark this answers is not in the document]
luispollo: Just a reminder to return an empty set and make this not nullable here as discussed above.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -575,6 +578,41 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
...
+          ))
+        }
+        test ("get stored notifications") {
+          expectThat(repository.environmentNotifications(deliveryConfig.name, "prod")?.size)
+            .isEqualTo(2)
+        }
+
+        test ("environment without notifications will return an empty set") {
+          expectThat(repository.environmentNotifications(deliveryConfig.name, "staging"))
+            .isNull()
```
luispollo, line 613: As mentioned above, this should be `.isEmpty()` so it will match the test description. 🙂

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEvent.kt
luispollo: You could pass the value of the `type` enum from the sub-classes for `ref` here, that way you wouldn't need to make it nullable. Same with `message`. In short, I don't think those fields should be made nullable just to fit the new use case.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEvent.kt
```diff
@@ -1,13 +1,40 @@
...
+  abstract val type: NotificationType
+  open val message: Notification? = null
+}
 
+data class UnhealthyNotification(
+  override val scope: NotificationScope,
+  override val ref: String?,
+  override val type: NotificationType,
+  override val message: Notification?
+  ) :NotificationEvent()
```
luispollo, line 22:
```suggestion
  ) : NotificationEvent()
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEvent.kt
```diff
@@ -1,13 +1,40 @@
...
+
+data class PinnedNotification(
+  val config: DeliveryConfig,
+  val pin: EnvironmentArtifactPin
+): NotificationEvent() {
```
luispollo, line 27:
```suggestion
) : NotificationEvent() {
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/notifications/PinnedNotification.kt
erikmunson: I originally thought the same Luis, but then later discovered that the public assets (i.e. available without auth) we serve from deck are all behind the Netflix internal VPN. That would mean that notifications would look broken for anyone on the public internet.

I'd love it if we could figure out a way to expose stuff outside the Netflix VPN without jumping through too many hoops, but also haven't had the time / mental focus to go any further looking into it. If anybody has thoughts on it please raise them so we can discuss!

For now I think having something that's hacky but publicly accessible is good enough to ship v0 of this feature, given its level of importance. Still worth improving it though.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEvent.kt
```diff
@@ -1,13 +1,40 @@
...
+  override val type = NotificationType.PINNED
+  override val scope = NotificationScope.ARTIFACT
+}
+
+data class UnpinnedNotification(
+  val config: DeliveryConfig,
+  val pinnedEnvironment: PinnedEnvironment?,
+  val targetEnvironment: String,
+  val user: String
+): NotificationEvent() {
```
luispollo, line 37:
```suggestion
) : NotificationEvent() {
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/NotificationType.kt
```diff
@@ -4,5 +4,7 @@ package com.netflix.spinnaker.keel.notifications
  * All valid notifiers
  */
 enum class NotificationType {
-  UNHEALTHY_RESOURCE
+  UNHEALTHY_RESOURCE,
+  PINNED,
```
luispollo, line 8: Maybe `ARTIFACT_PINNED` and `ARTIFACT_UNPINNED` to follow the naming pattern of the resource-related ones above?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
```diff
@@ -196,6 +197,9 @@ class CombinedRepository(
   override fun environmentFor(resourceId: String): Environment =
     deliveryConfigRepository.environmentFor(resourceId)
 
+  override fun environmentNotifications(deliveryConfigName: String, environmentName: String): Set<NotificationConfig>? =
```
luispollo, line 200:
```suggestion
  override fun environmentNotifications(deliveryConfigName: String, environmentName: String): Set<NotificationConfig> =
```

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/persistence/KeelReadOnlyRepository.kt
luispollo, line 23:
```suggestion
  fun environmentNotifications(deliveryConfigName: String, environmentName: String): Set<NotificationConfig>
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepository.kt
```diff
@@ -32,6 +33,11 @@ interface DeliveryConfigRepository : PeriodicallyCheckedRepository<DeliveryConfi
    */
   fun environmentFor(resourceId: String): Environment
 
+  /**
+   * Retrieve the [NotificationConfig] for an environment by environment name and delivery config name.
+   */
+  fun environmentNotifications(deliveryConfigName: String, environmentName: String): Set<NotificationConfig>?
```
luispollo, line 39:
```suggestion
  fun environmentNotifications(deliveryConfigName: String, environmentName: String): Set<NotificationConfig>
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -105,13 +110,18 @@ class ApplicationService(
...
   fun deletePin(user: String, application: String, targetEnvironment: String, reference: String? = null) {
     val config = repository.getDeliveryConfigForApplication(application)
+    val pinnedEnvironment = repository.pinnedEnvironments(config).find { it.targetEnvironment == targetEnvironment }
     repository.deletePin(config, targetEnvironment, reference)
-    // TODO: publish ArtifactUnpinnedEvent
+
+    publisher.publishEvent(UnpinnedNotification(config,
+      pinnedEnvironment,
+      targetEnvironment,
+      user))
```
luispollo, line 124:
```suggestion
      user)
    )
```

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListener.kt
```diff
@@ -48,7 +48,7 @@ class UnhealthyNotificationListener(
         val unhealthyDuration = unhealthyRepository.durationUnhealthy(event.resource)
         if (unhealthyDuration > config.minUnhealthyDuration) {
           publisher.publishEvent(
-            NotificationEvent(RESOURCE, event.resource.id, UNHEALTHY_RESOURCE, message(event, unhealthyDuration))
+            UnhealthyNotification(RESOURCE, event.resource.id, UNHEALTHY_RESOURCE, message(event, unhealthyDuration))
```
luispollo, line 51: Passing the `UNHEALTHY_RESOURCE` type is redundant with the constructor here, right?

keel-echo/src/test/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListenerTests.kt
```diff
@@ -91,7 +91,7 @@ class UnhealthyNotificationListenerTests : JUnit5Minutests {
           subject.onResourceHealthEvent(event)
           verify(exactly = 1) {
             publisher.publishEvent(
-              NotificationEvent(RESOURCE, r.id, UNHEALTHY_RESOURCE, subject.message(event, Duration.ofMinutes(6)))
+              UnhealthyNotification(RESOURCE, r.id, UNHEALTHY_RESOURCE, subject.message(event, Duration.ofMinutes(6)))
```
luispollo, line 94: Passing the `UNHEALTHY_RESOURCE` type is redundant with the constructor here, right?

keel-slack/keel-slack.gradle.kts
```diff
@@ -0,0 +1,18 @@
...
+  id("kotlin-spring")
+}
+
+dependencies {
+  implementation(project(":keel-retrofit"))
+  implementation(project(":keel-core"))
+  implementation("org.springframework.boot:spring-boot-autoconfigure")
+
+  implementation("com.slack.api:slack-api-model-kotlin-extension:1.4.1")
+  implementation("com.slack.api:slack-api-client-kotlin-extension:1.4.1")
```
luispollo, line 12: Nice!!! 🎉

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -0,0 +1,87 @@
...
+import com.netflix.spinnaker.keel.persistence.KeelRepository
+import com.netflix.spinnaker.keel.slack.handlers.SlackNotificationHandler
+import org.slf4j.LoggerFactory
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.context.event.EventListener
+import org.springframework.stereotype.Component
+import java.time.Clock
+
+/**
+ * Responsible to listen to notification events, and fetch the information needed
```
luispollo, line 16:
```suggestion
 * Responsible for listening to notification events, and fetching the information needed
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEvent.kt
emjburns: there is also some other wonky spacing in this file

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/Notifier.kt
```diff
@@ -56,8 +56,8 @@ class Notifier(
   private val notificationsEnabled: Boolean
     get() = springEnv.getProperty("keel.notifications.resource", Boolean::class.java, true)
 
-  @EventListener(NotificationEvent::class)
-  fun onResourceNotificationEvent(event: NotificationEvent) {
+  @EventListener(RepeatedNotificationEvent::class)
```
emjburns, line 59: <3 yasssss

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListener.kt
```diff
@@ -48,7 +48,7 @@ class UnhealthyNotificationListener(
         val unhealthyDuration = unhealthyRepository.durationUnhealthy(event.resource)
         if (unhealthyDuration > config.minUnhealthyDuration) {
           publisher.publishEvent(
-            NotificationEvent(RESOURCE, event.resource.id, UNHEALTHY_RESOURCE, message(event, unhealthyDuration))
+            UnhealthyNotification(event.resource.id,  message(event, unhealthyDuration))
```
emjburns, line 51: I think this needs to be `RepeatedNotifaction...` or whatever the class is called, because `Notifier` is listening for the generic version of the event

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
luispollo: If you want to try again, this works:
```kotlin
  fun <T : SlackNotificationEvent> Collection<SlackNotificationHandler<*>>.supporting(
    type: com.netflix.spinnaker.keel.notifications.NotificationType
  ): SlackNotificationHandler<T>? =
    handlers.find { hand ->
      hand.type == type
    } as? SlackNotificationHandler<T>
```

keel-echo/src/test/kotlin/com/netflix/spinnaker/keel/echo/NotifierTests.kt
```diff
@@ -78,7 +76,7 @@ class NotifierTests : JUnit5Minutests {
 
       context("new notification") {
         before {
-          every { notificationRepository.addNotification(event.scope, event.ref, event.type)} returns true
+          every { notificationRepository.addNotification(event.scope, event.ref!!, event.type)} returns true
```
emjburns, line 79: I think you can revert these `!!` since ref is required in the `RepeatedNotification`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/NotificationEvent.kt
gal-yardeni: @luispollo Yep I just pushed a change for it -- after talking with @emjburns , we've realized that we can have both repeated notifications (like unhealthy, which requires being persisted) and single notifications. I've decided to separate those two for now. I will definitely refactor it in the near future.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
luispollo: BTW, we ended up with 2 different `NotificationType` enums, hence the fully-qualified name above. You might want to rename one of them to avoid confusion.

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListener.kt
gal-yardeni: yes, removed it!

keel-echo/src/test/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListenerTests.kt
gal-yardeni: removed it!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -0,0 +1,87 @@
...
+    with(notification) {
+      val envNotifications = repository.environmentNotifications(config.name, targetEnvironment)
+
+      if (pinnedEnvironment != null) {
+        val latestApprovedArtifactVersion = repository.latestVersionApprovedIn(config, pinnedEnvironment!!.artifact, targetEnvironment)
+        val latestArtifact = latestApprovedArtifactVersion?.let { repository.getArtifactVersion(pinnedEnvironment!!.artifact, it, null) }
+
+        envNotifications?.forEach { notificationConfig ->
+          if (notificationConfig.type == NotificationType.slack) {
+            publisher.publishEvent(SlackUnpinnedNotification(
```
luispollo, line 71: What's the purpose of this event listener raising this other event here?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/SlackNotifier.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.slack.api.Slack
+import com.slack.api.model.block.LayoutBlock
+import com.slack.api.model.kotlin_extension.block.dsl.LayoutBlockDsl
+import org.slf4j.LoggerFactory
+import org.springframework.boot.context.properties.EnableConfigurationProperties
+import org.springframework.core.env.Environment
+import org.springframework.stereotype.Component
+
+/**
+ * This notifier is responsible on actually sending the slack notification,
```
luispollo, line 13:
```suggestion
 * This notifier is responsible for actually sending the Slack notification,
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: oh this is the old implementation -- it won't send an event, rather it will call the right handler. unpinning is still in TODO :)

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/SlackNotificationHandler.kt
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Implement this interface to send different types of slack notifications. Each notification is being consturct by [constructMessage]
+ * See: [PinnedNotificationHandler] for example
+ */
+
+interface SlackNotificationHandler <T: SlackNotificationEvent> {
+  val type: NotificationType
+
+  fun constructMessage(notification: T)
```
luispollo, line 14: This is not only constructing the message, but sending it out. Maybe `constructAndSendMessage`, or just `sendMessage`?

keel-slack/src/test/kotlin/com/netflix/spinnaker/keel/slack/handlers/NotificationEventListenerTests.kt
luispollo: Fix, please. 🙏

keel-echo/src/main/kotlin/com/netflix/spinnaker/keel/echo/UnhealthyNotificationListener.kt
gal-yardeni: I think the listener will to all `RepeatedNotifaction` events, will double-check.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -353,6 +354,26 @@ class SqlDeliveryConfigRepository(
         }
     } ?: throw OrphanedResourceException(resourceId)
 
+  override fun environmentNotifications(deliveryConfigName: String, environmentName: String): Set<NotificationConfig>? =
```
luispollo, line 357:
```suggestion
  override fun environmentNotifications(deliveryConfigName: String, environmentName: String): Set<NotificationConfig> =
```

emjburns (approved):
Changes look good to me after you remove the commented out code

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/SlackNotificationHandler.kt
gal-yardeni: sounds good!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/notifications/PinnedNotification.kt
luispollo: Ah, right, doh! [spinnaker.io](https://spinnaker.io)? 🙂

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
luispollo: What do you mean it won't send an event? Isn't that what the code is doing here? You mean you'll replace that with new code in a separate PR? If that's the case, it might be good to just remove this code from here, because it's confusing IMHO.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/verification/VerificationRepository.kt
```diff
@@ -64,4 +64,6 @@ data class VerificationContext(
     deliveryConfig.artifacts.first { it.reference == artifactReference }
 
   val verifications: Collection<Verification> = environment.verifyWith
+
+  fun verification(id: String): Verification = verifications.first { it.id == id }
```
luispollo, line 68: Unrelated? Need to rebase, maybe?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
```diff
@@ -172,6 +174,50 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
       .with(VerificationState::metadata) { isEqualTo(metadata) }
   }
 
+  @Test
+  fun `successive updates can add and update metadata`() {
```
luispollo, line 178: Unrelated? Rebase?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/SlackNotifier.kt
```diff
@@ -0,0 +1,49 @@
...
+
+
+  fun sendSlackNotification(channel: String, blocks: List<LayoutBlock>, token: String? = null) {
+    if (isSlackEnabled) {
+      log.debug("starting slack notification")
+      val slack = Slack.getInstance()
+
+      val actualToken = token ?: slackConfig.token
+
+      val response = slack.methods(actualToken).chatPostMessage { req ->
```
luispollo, line 36: This is soooo much easier than all that crap we have in Echo! ❤️

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
luispollo: Never mind, I see the new code now!

PR #1736, opened 2021-01-14
Model certificates for HTTPS ALBs more like Clouddriver

robfletcher:
This is one approach to fixing the issue we have with setting the certificate on ALB HTTPS listeners. What I've done here is create a new v1.2 of the ALB spec that changes the single `certificateArn` field on `Listener` to a set of objects, each of which may have a `certificateArn` field. This is the same as Clouddriver's model.

Long term I'd like to figure out if there's a better way to specify the cert than using the ARN, that seems not very user-friendly.

### DIFF
--- a/keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ApplicationLoadBalancerSpec.kt
+++ b/keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ApplicationLoadBalancerSpec.kt
@@ -3,9 +3,6 @@ package com.netflix.spinnaker.keel.api.ec2
 import com.netflix.spinnaker.keel.api.Moniker
 import com.netflix.spinnaker.keel.api.SubnetAwareLocations
 import com.netflix.spinnaker.keel.api.UnhappyControl
-import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec.Action
-import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec.Listener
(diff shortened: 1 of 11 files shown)


keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/ApplicationLoadBalancerSpec.kt
```diff
@@ -39,11 +36,15 @@ data class ApplicationLoadBalancerSpec(
...
     val protocol: String,
-    val certificateArn: String?,
+    val certificates: Set<Certificate> = emptySet(),
     val rules: Set<Rule> = emptySet(),
     val defaultActions: Set<Action> = emptySet()
   )
 
+  data class Certificate(
+    val certificateArn: String,
+  )
```
robfletcher, line 46: This is the model change

keel-web/src/test/resources/examples/alb-example.yml
```diff
@@ -4,14 +4,15 @@ serviceAccount: delivery-engineering@netflix.com
...
         app: fnord
       listeners:
       - port: 443
         protocol: HTTPS
-        certificateArn: arn:aws:iam::111111111111:server-certificate/fnord.prod.illuminati.org-DigiCertSHA2SecureServerCA-20200205-20210205
+        certificates:
+        - certificateArn: arn:aws:iam::111111111111:server-certificate/fnord.prod.illuminati.org-DigiCertSHA2SecureServerCA-20200205-20210205
         defaultActions:
         - type: forward
           order: 1
```
robfletcher, line 18: Here's what the change looks like in a delivery config

PR #1737, opened 2021-01-14
pass list of certificates to Orca rather than single certificateArn

robfletcher:
This is an alternate implementation to #1736 where I'm not changing the ALB spec model, just amending how we pass the cert to Orca.

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
@@ -294,7 +294,20 @@ class ApplicationLoadBalancerHandler(
           "isInternal" to internal,
           "idleTimeout" to idleTimeout.seconds,
           "securityGroups" to dependencies.securityGroupNames,
-          "listeners" to listeners,
+          "listeners" to listeners.map {
+            mapOf(
+              "port" to it.port,
+              "protocol" to it.protocol,
+              "rules" to it.rules,
+              "defaultActions" to it.defaultActions,
+            ).run {
+              if (it.certificateArn == null) {
+                this
+              } else {
+                this + mapOf("certificates" to listOf(mapOf("certificateArn" to it.certificateArn)))
+              }
+            }
+          },
           "targetGroups" to targetGroups.map {
             mapOf(
               "name" to it.name,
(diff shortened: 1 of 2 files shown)


keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandlerTests.kt
```diff
@@ -92,13 +92,14 @@ internal class ApplicationLoadBalancerHandlerTests : JUnit5Minutests {
     |    - us-east-1d
     |listeners:
     | - port: 80
-    |   protocol: HTTP
+    |   protocol: HTTPS
+    |   certificateArn: this-is-my-certificate
```
emjburns, line 96: I think i like this change better. there will be no change in the config, just a change on how we call orca so things work?

robfletcher: Correct

robfletcher: I would still like to figure out how to better specify a cert (if possible), but this should get us unblocked

lorin:
I agree with @emjburns, I prefer this approach to #1736. This one:

* has a simpler delivery config format
* the change is much less intrusive

The only downside I see compared to #1736 is that this approach doesn't support multiple certificates. However, I'd prefer to wait until we actually get a request for that functionality, since YAGNI.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ApplicationLoadBalancerHandler.kt
```diff
@@ -294,7 +294,20 @@ class ApplicationLoadBalancerHandler(
...
           "securityGroups" to dependencies.securityGroupNames,
-          "listeners" to listeners,
+          "listeners" to listeners.map {
+            mapOf(
+              "port" to it.port,
+              "protocol" to it.protocol,
+              "rules" to it.rules,
+              "defaultActions" to it.defaultActions,
+            ).run {
+              if (it.certificateArn == null) {
```
lorin, line 304: I love that `if` statements are expressions in Kotlin.

lorin (approved):
LGTM

emjburns (approved):

PR #1738, opened 2021-01-14
feat(docker): allow for optional container details in a resource

nimakaviani:
when resolving k8s resources, not every resource needs to have a container but all resources sharing the same resource version should still be able to go through a docker image resolver (among other resolvers) and be resolved successfully, regardless of whether or not they have container references specified in their spec.

This PR allows the decision for making containers optional to be made by the container resolvers.

/cc @luispollo @robfletcher


emjburns:
It would be helpful to have an example of what you're trying to do here. In general I don't have a problem with making things nullable when they need to be, but I'm confused about the need here

nimakaviani:
sure. We discussed the structure of keel's k8s resources during the SIG call and also in [this proposal](https://docs.google.com/document/d/1HKZyLNv9XZsLx6IGA2TbWNeLABvkk2MU6C5K1ipdDKk/edit#heading=h.7kajxsk7pi1i). basically, a given k8s resource of `kind: k8s/resource@v1` in keel can embed any of the vanilla Kubernetes resources, some of which will have artifacts associated with them while some of which won't. 

Consider the simple example below, where a k8s `Deployment` will deploy a contianer while a k8s `Service` won't have a container artifact associated with it:

```yaml
  - kind: k8s/resource@v1
      spec:
        metadata:
          application: spinmd
        template:
          apiVersion: v1
          kind: Service
          metadata:
            name: hello-kubernetes
            namespace: default
          spec:  { ... } # removed for brevity

    - kind: k8s/resource@v1
      spec:
        metadata:
          application: spinmd
        container:
          reference: my-docker-artifact
        template:
          apiVersion: apps/v1
          kind: Deployment
          metadata:
            name: sample-testing
            namespace: default
          spec: { ... } # removed for brevity
           
```

When these two resources are marked under the same keel resource `kind`, they end up going through the same set of resolvers (including the artifact resolver), and unless we make association of container artifacts to the resources of the given kind optional, the resolution will fail. 

Other solution would be to support different `kind`  for different resource types but that is less user-friendly and less intuitive IMO. thoughts?

queueburt:
> Other solution would be to support different `kind` for different resource types but that is less user-friendly and less intuitive IMO. thoughts?

I'd personally be worried about different `kind` resources in a k8s context creating a "chase the API" problem, not to mention limiting (or complicating) functionality for cool stuff like API extensibility, CRDs, and those sorts of things.  This is likely unique to the `k8s/resource@v1` just due to the fact that multiple instances of the type can create radically different constructs, but the flow you outlined in the post above makes sense to me, for what little that's worth.  :-)

luispollo:
> Other solution would be to support different `kind` for different resource types but that is less user-friendly and less intuitive IMO. thoughts?

Am I correct that, if we removed the need to wrap Kubernetes resources in MD resources, this would cease to be a problem? I'm wondering whether it would make more sense for us to invest in getting _that_ problem resolved instead...
Cc: @robfletcher

nimakaviani:
We still need separate resolvers for different `kinds` as part of adhering to MD abstractions, whether those kinds are vanilla k8s kinds or wrapped MD kinds. So, I'd say it falls under the category of _"chase the API" problems_ that @queueburt mentioned. Plus it requires some extra workaround for things like associating resources to a spinnaker application, etc.

Even though I was also in the boat of "lets support k8s resources as first level citizens in MD" initially, the more I think about it, the more I like the idea of generic wrappers for vanilla k8s resources. It also helps with more consistent CRDs when we introduce other types of resources like kustomize and helm, etc.

luispollo:
Fair enough. As a reminder, though, the current 1:1 mapping between a resource `kind` and a `ResourceHandler` is a design decision we could change. We could totally have a single handler for multiple types. The only downside in my mind being that, for Kubernetes resources, we'd need to make changes to the mapping every time we introduce a new K8s resource, whereas with the wrapper that's not needed.

nimakaviani:
yup exactly! Add CRDs to the mix, and it becomes impossible to maintain.

keel-docker/src/test/kotlin/com/netflix/spinnaker/keel/docker/SampleDockerImageResolverTests.kt
```diff
@@ -58,13 +58,20 @@ class SampleDockerImageResolverTests : JUnit5Minutests {
...
+  )
+
   val versionedContainerResource = generateResource(oldStyleSpec)
 
   val referenceResource = generateResource(referenceSpec)
   val versionedDeliveryConfig = generateDeliveryConfig(versionedContainerResource, artifact)
-
   val referenceDeliveryConfig = generateDeliveryConfig(referenceResource, artifact)
 
+  private val nullableResource = generateResource(nullable)
```
luispollo, line 72: Suggestion for clarity:
```suggestion
  private val nullableContainerResource = generateResource(nullable)
```

keel-docker/src/test/kotlin/com/netflix/spinnaker/keel/docker/SampleDockerImageResolverTests.kt
```diff
@@ -58,13 +58,20 @@ class SampleDockerImageResolverTests : JUnit5Minutests {
...
+  private val nullableDeliveryConfig = generateDeliveryConfig(nullableResource, artifact)
```
luispollo, line 73: Suggestion for clarity:
```suggestion
  private val nullContainerDeliveryConfig = generateDeliveryConfig(nullableResource, artifact)
```

keel-docker/src/test/kotlin/com/netflix/spinnaker/keel/docker/SampleDockerImageResolverTests.kt
```diff
@@ -58,13 +58,20 @@ class SampleDockerImageResolverTests : JUnit5Minutests {
     ),
     account = "test"
   )
+
+  private val nullable = SampleSpecWithContainer(
```
luispollo, line 62: Suggestion for clarity:
```suggestion
  private val nullContainerSpec = SampleSpecWithContainer(
```

keel-docker/src/test/kotlin/com/netflix/spinnaker/keel/docker/SampleDockerImageResolverTests.kt
```diff
@@ -156,5 +163,19 @@ class SampleDockerImageResolverTests : JUnit5Minutests {
         }
       }
     }
+
+    context("resolving from nullable resource") {
```
luispollo, line 167:
```suggestion
    context("resolving from resource with no container info") {
```

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/DockerImageResolver.kt
```diff
@@ -72,7 +72,7 @@ abstract class DockerImageResolver<T : ResourceSpec>(
 
   override fun invoke(resource: Resource<T>): Resource<T> {
     val container = getContainerFromSpec(resource)
-    if (container is DigestProvider) {
+    if (container is DigestProvider || container is NullableReferenceProvider) {
```
luispollo, line 75: IIUC, this comes into play in your plugin because your `ResourceSpec` has a `ContainerProvider` field (which might need to be empty if the K8s resources within don't care about a container), correct?

I was remembering that you had the opposite example in your proposal, where you might have, say, 2 different `Deployment` objects within the same MD resource spec, and would need a way to reference 2 different artifacts. That got me wondering whether perhaps a better approach might be to create a new sub-type of `ContainerProvider` that can handle multiple references (`MultiReferenceContainerProvider`?), which might look something like this:

```yaml
kind: k8s/resources@v1
spec:
  container:
    references:
      - my-first-image
      - my-second-image
   # omitted for brevity
```

...which then would also allow you do this when you don't care about containers, instead of omitting the property altogether:
```yaml
kind: k8s/resources@v1
spec:
  container:
    references: []
   # omitted for brevity
```

Thoughts?

robfletcher:
> We still need separate resolvers for different `kinds` as part of adhering to MD abstractions, whether those kinds are vanilla k8s kinds or wrapped MD kinds. So, I'd say it falls under the category of _"chase the API" problems_ that @queueburt mentioned. Plus it requires some extra workaround for things like associating resources to a spinnaker application, etc.
> 

If that's the issue we could certainly allow handlers to accept multiple resource kinds

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/DockerImageResolver.kt
nimakaviani: hmmm, yeah! love the idea of combining the two into one type. Had that multi-reference idea somehow parked when I did this. will update the PR. thanks for bringing it to my attention :)

nimakaviani:
@luispollo this latest change should capture the better of the two worlds, i.e. empty references and multiple references

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/ContainerProvider.kt
```diff
@@ -29,6 +29,10 @@ data class ReferenceProvider(
   val reference: String
 ) : ContainerProvider()
 
+// used for resources with multiple container references
+@JsonDeserialize(using = JsonDeserializer.None::class)
+data class ReferenceProviders(val references : Set<String> = HashSet()) : ContainerProvider()
```
luispollo, line 34:
```suggestion
data class MultiReferenceContainerProvider(val references : Set<String> = HashSet()) : ContainerProvider()
```

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/DockerImageResolver.kt
```diff
@@ -72,17 +72,30 @@ abstract class DockerImageResolver<T : ResourceSpec>(
 
   override fun invoke(resource: Resource<T>): Resource<T> {
     val container = getContainerFromSpec(resource)
-    if (container is DigestProvider) {
+    if (container is DigestProvider || (container is ReferenceProviders && container.references.isEmpty())) {
```
luispollo, line 75:
```suggestion
    if (container is DigestProvider || (container is MultiReferenceContainerProvider && container.references.isEmpty())) {
```

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/DockerImageResolver.kt
```diff
@@ -72,17 +72,30 @@ abstract class DockerImageResolver<T : ResourceSpec>(
...
     val deliveryConfig = repository.deliveryConfigFor(resource.id)
     val environment = repository.environmentFor(resource.id)
-    val artifact = getArtifact(container, deliveryConfig)
     val account = getAccountFromSpec(resource)
-    val tag: String = findTagGivenDeliveryConfig(deliveryConfig, environment, artifact)
 
-    val newContainer = getContainer(account, artifact, tag)
-    return updateContainerInSpec(resource, newContainer, artifact, tag)
+    val containers = mutableListOf<ContainerProvider>()
+    if (container is ReferenceProviders) {
```
luispollo, line 84:
```suggestion
    if (container is MultiReferenceContainerProvider) {
```

keel-docker/src/main/kotlin/com/netflix/spinnaker/keel/docker/DockerImageResolver.kt
```diff
@@ -72,17 +72,30 @@ abstract class DockerImageResolver<T : ResourceSpec>(
...
+      container.references.forEach{ containers.add(ReferenceProvider(it))}
```
luispollo, line 85:
```suggestion
      container.references.forEach {
        containers.add(ReferenceProvider(it))
      }
```

keel-docker/src/test/kotlin/com/netflix/spinnaker/keel/docker/SampleDockerImageResolver.kt
```diff
@@ -33,7 +33,7 @@ class SampleDockerImageResolver(
   override val supportedKind = kind<SampleSpecWithContainer>("sample.resource/sample@v1")
 
   override fun getContainerFromSpec(resource: Resource<SampleSpecWithContainer>) =
-    resource.spec.container
+    if (resource.spec.container != null) resource.spec.container as ContainerProvider else ReferenceProviders()
```
luispollo, line 36:
```suggestion
    if (resource.spec.container != null) resource.spec.container as ContainerProvider else MultiReferenceContainerProvider()
```

keel-docker/src/test/kotlin/com/netflix/spinnaker/keel/docker/SampleDockerImageResolverTests.kt
```diff
@@ -58,12 +62,27 @@ class SampleDockerImageResolverTests : JUnit5Minutests {
     ),
     account = "test"
   )
-  val versionedContainerResource = generateResource(oldStyleSpec)
 
+  private val multiReferenceSpec = SampleSpecWithContainer(
+    container = ReferenceProviders(
```
luispollo, line 67:
```suggestion
    container = MultiReferenceContainerProvider(
```

keel-docker/src/test/kotlin/com/netflix/spinnaker/keel/docker/SampleDockerImageResolverTests.kt
```diff
@@ -134,27 +153,74 @@ class SampleDockerImageResolverTests : JUnit5Minutests {
...
+          every { repository.latestVersionApprovedIn(multiReferenceDeliveryConfig, artifacts.elementAt(0), "test") } returns "v0.0.1"
+          every { repository.latestVersionApprovedIn(multiReferenceDeliveryConfig, artifacts.elementAt(1), "test") } returns "v0.0.2"
+        }
+
+        test("the resolver overrides to the latest artifact") {
+          val resolvedResource = this.invoke(multiReferenceResource)
+          expect {
+            that(resolvedResource.spec.container).isA<DigestProvider>()
+            that(resolvedResource.spec.container as DigestProvider).get { digest }.isEqualTo("sha256:b4857d7596462aeb1977e6e5d1e31b20a5b5eecf890cd64ac62f145b3839ee97")
+          }
```
luispollo, line 207: Hmm, this test is a bit confusing to me. I was expecting the resolver to translate each of the artifact references to a digest, but the end result here is that the last artifact's digest takes precedence. In a real-world scenario, I'd imagine you'd have a Kubernetes wrapper resource with, say, 2 different `PodSpec` objects each pointing to a different artifact. My guess is you'll override `updateContainerInSpec` for that purpose, but because the resource spec here contains a single `container` field, the test looks a bit weird. If nothing else, maybe just add a comment explaining why?

luispollo (commented):
Hi @nimakaviani, thanks for making the changes we discussed. I have a suggestion on naming and a question on the tests, otherwise looks good.

keel-docker/src/test/kotlin/com/netflix/spinnaker/keel/docker/SampleDockerImageResolverTests.kt
nimakaviani: Thanks for the comments @luispollo. The name change looks good to me. will update the PR.

As for the test, yeah your guess is spot on. The k8s plugin will override different container images in the PodSpec by cross-referencing them against artifact references. I have a full-blown set of tests in the k8s plugin ([see here](https://github.com/nimakaviani/managed-delivery-k8s-plugin/blob/f6fefb02a0105a9f49f382e7497f9fbd4a837f49/k8s-plugin/src/test/kotlin/com/amazon/spinnaker/keel/tests/DockerImageResolverTest.kt)) that exercise different permutations. Particularly for multiple containers in the spec [it verifies that the references get written correctly](https://github.com/nimakaviani/managed-delivery-k8s-plugin/blob/f6fefb02a0105a9f49f382e7497f9fbd4a837f49/k8s-plugin/src/test/kotlin/com/amazon/spinnaker/keel/tests/DockerImageResolverTest.kt#L171-L219). 

Because of the exhaustive testing in the plugin, I kept it relatively minimal in the Keel test. I didn't want to duplicate the testing behavior or make significant changes to the existing Sample Resolver; hence just verifying that the resource goes through the DockerImageResolver twice and that eventually the value for the container gets overriden by the second artifact that comes through. 

If you think the explanation is sufficient, I can leave it only as a comment in the test. If it needs further clarification, I'd be happy to make the necessary changes where needed.

luispollo: Makes sense, thanks for the details!

luispollo (approved):
Thanks for making the suggested changes. LGTM!

PR #1739, opened 2021-01-14
feat(ec2): Allow extension of security group handling

luispollo:
Problem: we want to extend the EC2 `SecurityGroupHandler` in a Netflix-internal implementation which handles Netflix-specific ingress rules, including the ability to modify the Orca job payload to include additional fields.

This PR introduces the following changes to enable this use case:
* Replaces custom registration of Jackson serializers and deserializers with usage of Spring's `JsonComponentModule` and corresponding `@JsonComponent` annotations. The existing serializers and deserializers are scanned and registered by default, but `SecurityGroupRuleDeserializer` is made conditional so that we can replace it with our own implementation in `keel-nflx`.
* Makes a couple of methods in `SecurityGroupHandler` `protected` instead of `private` so they can be overridden in subclasses.
* Adds a couple of missing fields from models required for the Netflix internal handler to make decisions.


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/support/Tag.kt
```diff
@@ -0,0 +1,6 @@
+package com.netflix.spinnaker.keel.api.support
+
+data class Tag(
```
luispollo, line 3: Moved here to be accessible from the EC2 API module.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/config/ClouddriverConfiguration.kt
```diff
@@ -43,6 +43,7 @@ class ClouddriverConfiguration {
       ?: throw BeanCreationException("Invalid URL: $clouddriverBaseUrl")
 
   @Bean
+  @ConditionalOnMissingBean(CloudDriverService::class)
```
luispollo, line 46: This is so we can replace the default implementation with the Netflix internal one.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/SecurityGroupModel.kt
```diff
@@ -26,13 +27,15 @@ data class SecurityGroupModel(
   val region: String,
   val vpcId: String?,
   val inboundRules: Set<SecurityGroupRule> = setOf(),
-  val moniker: Moniker
+  val moniker: Moniker,
+  val tags: Set<Tag> = setOf()
```
luispollo, line 31: Tags are required for the Netflix internal implementation to make determinations about the security group.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/SecurityGroupModel.kt
```diff
@@ -26,13 +27,15 @@ data class SecurityGroupModel(
...
 ) {
   data class SecurityGroupRule(
     val protocol: String?,
     val portRanges: List<SecurityGroupRulePortRange>?,
     val securityGroup: SecurityGroupRuleReference?,
-    val range: SecurityGroupRuleCidr?
+    val range: SecurityGroupRuleCidr?,
+    val description: String?
```
luispollo, line 38: Required for the Netflix internal implementation to make determinations about the security group rules.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/caffeine/CacheFactory.kt
```diff
@@ -52,7 +52,7 @@ class CacheFactory(
       .buildAsync(loader.toAsyncCacheLoader())
       .monitor(cacheName)
 
-  private fun builder(
+  fun builder(
```
luispollo, line 55: This could remain private. Will fix.

keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroup.kt
```diff
@@ -2,13 +2,16 @@ package com.netflix.spinnaker.keel.api.ec2
...
 import com.netflix.spinnaker.keel.api.Moniker
+import com.netflix.spinnaker.keel.api.support.Tag
 
 data class SecurityGroup(
   val moniker: Moniker,
   val location: Location,
   @get:ExcludedFromDiff
-  val description: String?,
-  val inboundRules: Set<SecurityGroupRule> = emptySet()
+  val description: String? = moniker.toString(),
```
luispollo, line 11: The description is apparently mandatory now, but I didn't want to make a ton of unrelated changes, so decided to just set a default if not specified.
image

keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupSpec.kt
```diff
@@ -21,7 +21,7 @@ import com.netflix.spinnaker.keel.api.Monikered
 import com.netflix.spinnaker.keel.api.SimpleLocations
 import com.netflix.spinnaker.keel.api.schema.Optional
 
-data class SecurityGroupSpec(
+open class SecurityGroupSpec(
```
luispollo, line 24: Also not necessary. Will undo.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/KeelEc2ApiModule.kt
```diff
@@ -51,14 +51,28 @@ import com.netflix.spinnaker.keel.ec2.jackson.mixins.StepAdjustmentMixin
 import com.netflix.spinnaker.keel.ec2.jackson.mixins.StepScalingPolicyMixin
 import com.netflix.spinnaker.keel.ec2.jackson.mixins.TargetGroupAttributesMixin
 import com.netflix.spinnaker.keel.ec2.jackson.mixins.TargetTrackingPolicyMixin
+import com.netflix.spinnaker.keel.jackson.SerializationExtensionRegistry
 
-fun ObjectMapper.registerKeelEc2ApiModule(): ObjectMapper = registerModule(KeelEc2ApiModule)
+fun ObjectMapper.registerKeelEc2ApiModule(serializationExtensionRegistry: SerializationExtensionRegistry): ObjectMapper {
+  with(serializationExtensionRegistry) {
```
luispollo, line 57: This is where the default serializers and deserializers are now registered. By the time we get here, the Netflix components will have already have initialized and had a chance to add customizations to the registry.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -328,60 +330,64 @@ class SecurityGroupHandler(
...
-        "vpcId" to cloudDriverCache.networkBy(location.vpc, location.account, location.region).id,
-        "description" to description,
-        "securityGroupIngress" to inboundRules
-          // we have to do a 2-phase create for self-referencing ingress rules as the referenced
-          // security group must exist prior to the rule being applied. We filter then out here and
-          // the subsequent diff will apply the additional group(s).
-          .filterNot { it is ReferenceRule && it.name == moniker.name }
-          .mapNotNull {
-            it.referenceRuleToJob(this)
+  protected open fun generateCreateJob(securityGroup: SecurityGroup): Job =
```
luispollo, line 333: `protected open` so it can be overridden in sub-classes.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -328,60 +330,64 @@ class SecurityGroupHandler(
...
-        "vpcId" to cloudDriverCache.networkBy(location.vpc, location.account, location.region).id,
-        "description" to description,
-        "securityGroupIngress" to inboundRules.mapNotNull {
-          it.referenceRuleToJob(this)
-        },
-        "ipIngress" to inboundRules.mapNotNull {
-          it.cidrRuleToJob()
-        },
-        "accountName" to location.account
+  protected open fun generateUpdateJob(securityGroup: SecurityGroup): Job =
```
luispollo, line 361: `protected open` so it can be overridden in sub-classes.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -328,60 +330,64 @@ class SecurityGroupHandler(
...
+    }
 
   private fun SecurityGroupRule.referenceRuleToJob(securityGroup: SecurityGroup): Map<String, Any?>? =
     when (this) {
       is ReferenceRule -> mapOf(
         "type" to protocol.type,
         "startPort" to (portRange as? PortRange)?.startPort,
         "endPort" to (portRange as? PortRange)?.endPort,
-        "name" to (name ?: securityGroup.moniker.toString())
+        "name" to name
```
luispollo, line 390: I honestly forget why I made this change. Will undo.

keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupRule.kt
```diff
@@ -18,9 +18,10 @@ package com.netflix.spinnaker.keel.api.ec2
 import com.netflix.spinnaker.keel.api.schema.Literal
 import com.netflix.spinnaker.keel.api.schema.Optional
 
-sealed class SecurityGroupRule {
+abstract class SecurityGroupRule {
```
luispollo, line 21: @robfletcher I have a suspicion that this change is what's causing this kind of error in the API schema tests:
```
  org.opentest4j.AssertionFailedError: ▼ Expect that validation messages for security-group-example.yml:
    ✗ is valid
      found 2 validation errors: [$.environments[0].resources[0].spec.inboundRules[0].protocol: is not defined in the schema and the schema does not allow additional properties, $.environments[0].resources[0].spec.inboundRules[0].portRange: is not defined in the schema and the schema does not allow additional properties]
```
Any ideas?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/SerializationExtensionRegistry.kt
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * A registry to allow customization of JSON serializers and deserializers .
+ */
+@Component
+class SerializationExtensionRegistry {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+  private val _deserializers: MutableMap<Class<*>, JsonDeserializer<*>> = mutableMapOf()
+  private val _serializers: MutableMap<Class<*>, JsonSerializer<*>> = mutableMapOf()
+
+  val deserializers: Map<Class<*>, JsonDeserializer<*>>
```
luispollo, line 17: I didn't want to make the exposed maps mutable.

keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupRule.kt
luispollo: Confirmed: if I change this back to `sealed`, the tests pass. Not sure how to proceed here.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/SerializationExtensionRegistry.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.fasterxml.jackson.databind.JsonDeserializer
+import com.fasterxml.jackson.databind.JsonSerializer
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+
+/**
+ * A registry to allow customization of JSON serializers and deserializers .
+ */
+@Component
+class SerializationExtensionRegistry {
```
luispollo, line 12: @robfletcher I debated with myself whether to just add these new methods to `ExtensionRegistry` or create a separate class. In the end I chose a separate class trying to stick with the single-responsibility principle, but it may be overkill. If nothing else, the name is kinda long and it's another thing to wire in, in various places. Thoughts?

keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroupRule.kt
luispollo: Fixed by manually registering the sub-types of `SecurityGroupRule` with the `ExtensionRegistry`.

luispollo, line 21: This change is what caused me to have to manually register subtypes for `SecurityGroupRule` with the object mappers and the `ExtensionRegistry`.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/SecurityGroupRuleDeserializer.kt
```diff
@@ -6,7 +6,7 @@ import com.netflix.spinnaker.keel.api.ec2.ReferenceRule
 import com.netflix.spinnaker.keel.api.ec2.SecurityGroupRule
 import com.netflix.spinnaker.keel.jackson.PropertyNamePolymorphicDeserializer
 
-internal class SecurityGroupRuleDeserializer :
+open class SecurityGroupRuleDeserializer :
```
luispollo, line 9: `open` so it can be extended in the Netflix wrapper.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/DeserializationInjectionTests.kt
```diff
@@ -17,17 +18,9 @@ import strikt.assertions.isNotNull
 import strikt.assertions.isSuccess
 
 internal class DeserializationInjectionTests {
-
-  val mapper = configuredYamlMapper()
+  private val mapper = configuredYamlMapper()
     .registerKeelApiModule()
-    .registerKeelEc2ApiModule()
-  val extensionRegistry = DefaultExtensionRegistry(listOf(mapper)).apply {
```
luispollo, line 24: This was unused, as far as I could see.

lorin (approved):
My understanding

## Context

Netflix's internal Spinnaker deployment has some custom functionality (called "IP objects"), which is implemented using security group  `tags` fields (and maybe `description` as well? I wasn't sure aobut htis).

## Problem

MD doesn't currently allow users to use this functionality. We could expose security group tags, but we want to avoid that since it's an implementation detail that we don't want to expose to users.

## Proposed solution

Support specifying IP objects in the internal deployment of keel by enabling the internal version of keel to customize how the JSON payload for security groups get generated when making calls to orca.

## My feedback

LGTM, although I'm not familiar enough with how Jackson works to have a lot of confidence in my ability to review code like this.

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/SecurityGroupModel.kt
```diff
@@ -26,13 +27,15 @@ data class SecurityGroupModel(
...
+  val moniker: Moniker,
+  val tags: Set<Tag> = setOf()
 ) {
   data class SecurityGroupRule(
     val protocol: String?,
     val portRanges: List<SecurityGroupRulePortRange>?,
     val securityGroup: SecurityGroupRuleReference?,
-    val range: SecurityGroupRuleCidr?
+    val range: SecurityGroupRuleCidr?,
+    val description: String? = null
```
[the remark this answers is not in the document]
robfletcher: I wasn't aware you can have a description on a rule. Is that really a thing?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/SerializationExtensionRegistry.kt
```diff
@@ -0,0 +1,40 @@
...
+    get() = _deserializers.toMap()
+
+  val serializers: Map<Class<*>, JsonSerializer<*>>
+    get() = _serializers.toMap()
+
+  fun <T : JsonDeserializer<*>> register(javaClass: Class<*>, deserializer: T) {
+    _deserializers.putIfAbsent(javaClass, deserializer)
+      ?.let {
+        if (it.javaClass != deserializer.javaClass) {
+          log.warn("Attempt to register deserializer ${deserializer.javaClass.name} for type ${javaClass.name}, but ${it.javaClass.name} is already registered.")
```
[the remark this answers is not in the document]
robfletcher: Doesn't that prevent the method from being idempotent if you call it with the same parameters though?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/SerializationExtensionRegistry.kt
emjburns: I personally like it when config stuff is all together in one file or package because sometimes you're searching for the config that you vaguely know exists but you can't for the life of you remember what it is called. 

That's not a reason to change it here, I don't mind either way. But since people rarely remember that there are settings controlling things 6 months down the line it's just a thought.

emjburns (approved):
The inline comments you added are really useful, thanks. Approving this as I have no concerns, but I also don't know enough to find a bug in Jackson stuff.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/SerializationExtensionRegistry.kt
robfletcher: I'm struggling to see that this class provides much value at all. Isn't it just adding indirection to the calls the `*Module` classes make? Where modules were previously calling the Jackson API directly (which seems reasonable since modules are a Jackson concept) now they're calling this and then you have to loop over all the things that were registered later to make the same calls to the Jackson API. It doesn't look like this is queried anywhere else. 

The purpose of the original `ExtensionRegistry` was that we needed to have a way to a) register subtypes with Jackson, and b) make the schema generator aware of those same subtypes. The `ExtensionRegistry` allows us to do both those things with one call so you don't accidentally do one and not the other. I'm not sure this class provides any similar value.

I guess what I'm saying is why is having this better than just calling the Jackson API directly?

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/SecurityGroupModel.kt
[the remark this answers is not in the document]
luispollo: It is. It's the thing that carries the IP Object ID, believe it of not. 😄

keel-ec2-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ec2/SecurityGroup.kt
```diff
@@ -2,13 +2,16 @@ package com.netflix.spinnaker.keel.api.ec2
...
+  val inboundRules: Set<SecurityGroupRule> = emptySet(),
+  @get:ExcludedFromDiff
```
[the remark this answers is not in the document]
luispollo: My rationale was that I don't expect users to specify tags in the delivery config, but I needed them because that's the second place where we store the Object IP IDs.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/SerializationExtensionRegistry.kt
luispollo: My thinking (which connects with Emily's comment above) was that I wanted the registration of serializers/deserializers to happen in a single place, specifically in that `registerKeelEc2ApiModule` function, but in order to do that I needed some place for the components in `keel-nflx` to register our own (de)serializers. The other challenge was that I wanted `SecurityGroupRuleDeserializer` to be registered by default in the OSS implementation, but have a way for `keel-nflx` to register a different deserializer first to replace it.

I guess I could wire in the list of `ObjectMapper` in the internal implementation and look for that particular one and replace it with ours, if that's available in the Jackson API? I dunno, I wanted to create something that wasn't serving just this one single use case, but maybe it's overkill.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/SerializationExtensionRegistry.kt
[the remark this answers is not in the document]
luispollo: No, because the warning is only printed if the class of the parameter is different (line 26). If you pass a different type, then it's ignored because I wanted the first registration to take precedence, hence the warning in that case.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/SerializationExtensionRegistry.kt
robfletcher: Would just having a module in keel-nflx address that, or would something still be missing?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/jackson/SerializationExtensionRegistry.kt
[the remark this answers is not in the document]
robfletcher: ok, I see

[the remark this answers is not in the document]
robfletcher: ok, I see

luispollo:
@robfletcher As discussed over chat, I've made changes to the PR to leverage Spring's  `JsonComponentModule` to auto-wire custom (de-)serializers in https://github.com/spinnaker/keel/pull/1739/commits/348c91c2a973e4c731f8293ea14bd1678a623590. I'm struggling to get all the tests adjusted, and could use some help. 🙏

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/ActiveServerGroupImageDeserializer.kt
```diff
@@ -5,7 +5,9 @@ import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.deser.std.StdNodeBasedDeserializer
 import com.fasterxml.jackson.databind.node.ArrayNode
 import com.netflix.spinnaker.keel.api.ec2.ServerGroup.ActiveServerGroupImage
+import org.springframework.boot.jackson.JsonComponent
 
+@JsonComponent
```
luispollo, line 10: Example of using `@JsonComponent` to auto-wire a custom deserializer.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/KeelEc2ApiModule.kt
```diff
@@ -52,14 +46,31 @@ import com.netflix.spinnaker.keel.ec2.jackson.mixins.StepScalingPolicyMixin
...
+  ReferenceRule::class.java to "reference",
+  CrossAccountReferenceRule::class.java to "cross-account",
+  CidrRule::class.java to "cidr"
+)
 
-object KeelEc2ApiModule : SimpleModule("Keel EC2 API") {
+fun ObjectMapper.registerKeelEc2ApiModule(): ObjectMapper {
+  return registerModule(KeelEc2ApiModule)
+    .apply {
+      SECURITY_GROUP_RULE_SUBTYPES.forEach { (subType, discriminator) ->
```
luispollo, line 58: Required due to making `SecurityGroupRule` abstract.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/KeelEc2ApiModule.kt
```diff
@@ -52,14 +46,31 @@ import com.netflix.spinnaker.keel.ec2.jackson.mixins.StepScalingPolicyMixin
...
+        registerSubtypes(NamedType(subType, discriminator))
+      }
+    }
+}
+
+fun ExtensionRegistry.registerEc2Subtypes() {
+  // Note that the discriminators below are not used as sub-types are determined by the custom deserializer above
+  SECURITY_GROUP_RULE_SUBTYPES.forEach { (subType, discriminator) ->
```
luispollo, line 66: Required due to making `SecurityGroupRule` abstract.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/KeelEc2ApiModule.kt
```diff
@@ -52,14 +46,31 @@ import com.netflix.spinnaker.keel.ec2.jackson.mixins.StepScalingPolicyMixin
...
+    register(SecurityGroupRule::class.java, subType, discriminator)
+  }
+}
+
+internal object KeelEc2ApiModule : SimpleModule("Keel EC2 API") {
   override fun setupModule(context: SetupContext) {
     with(context) {
-      addSerializers(KeelEc2ApiSerializers)
-      addDeserializers(KeelEc2ApiDeserializers)
```
luispollo, line 61: These are no longer required with `JsonComponentModule`.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/SecurityGroupRuleDeserializer.kt
```diff
@@ -5,8 +5,12 @@ import com.netflix.spinnaker.keel.api.ec2.CrossAccountReferenceRule
 import com.netflix.spinnaker.keel.api.ec2.ReferenceRule
 import com.netflix.spinnaker.keel.api.ec2.SecurityGroupRule
 import com.netflix.spinnaker.keel.jackson.PropertyNamePolymorphicDeserializer
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
+import org.springframework.boot.jackson.JsonComponent
 
-internal class SecurityGroupRuleDeserializer :
+@JsonComponent
+@ConditionalOnMissingBean(SecurityGroupRuleDeserializer::class)
```
luispollo, line 12: `SecurityGroupRuleDeserializer` is made conditional on any sub-classes such that we can replace it with our own in `keel-nflx`.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -52,11 +52,11 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import retrofit2.HttpException
 
-class SecurityGroupHandler(
-  private val cloudDriverService: CloudDriverService,
-  private val cloudDriverCache: CloudDriverCache,
-  private val orcaService: OrcaService,
-  private val taskLauncher: TaskLauncher,
+open class SecurityGroupHandler(
```
luispollo, line 55: `open` so it can be sub-classed in `keel-nflx`.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/DeserializationInjectionTests.kt
```diff
@@ -16,18 +15,14 @@ import strikt.assertions.isEqualTo
 import strikt.assertions.isNotNull
 import strikt.assertions.isSuccess
 
+@SpringBootTest(
+  classes = [Ec2JsonTestConfiguration::class],
+  webEnvironment = NONE
+)
```
luispollo, line 21: Example of replacing custom configuration of the `ObjectMapper` with Spring auto-wiring in tests.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/Ec2JsonTestConfiguration.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.context.annotation.Configuration
+import org.springframework.context.annotation.Primary
+
+/**
+ * Configures the primary object mapper with auto-wired custom JSON serializers and deserializers
+ * via [JsonComponentModule].
+ */
+@Configuration
+@ComponentScan(basePackages = ["com.netflix.spinnaker.keel.ec2.jackson"])
+class Ec2JsonTestConfiguration {
```
luispollo, line 21: @robfletcher I've attempted to create a new module `keel-ec2-test-support` with this class, and add a dependency in the other modules, so it could be reused, however I ran into Spring app initialization errors because the new module depended on `keel-ec2-plugin` which brought in a bunch of undesired Spring beans. If you have any ideas on how to approach this to avoid the redundancy, I'd love your thoughts.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/DefaultConfiguration.kt
```diff
@@ -67,12 +68,17 @@ class DefaultConfiguration(
...
   fun idGenerator(): ULID = ULID()
 
+  @Bean
+  fun jsonComponentModule() = JsonComponentModule()
+
   @Bean(name = ["jsonMapper", "objectMapper"])
   @Primary
-  fun objectMapper(): ObjectMapper = configuredObjectMapper()
+  fun objectMapper(jsonComponentModule: JsonComponentModule): ObjectMapper =
+    configuredObjectMapper().registerModule(jsonComponentModule)
```
luispollo, line 77: Example of wiring an `ObjectMapper` with `JsonComponentModule`, which itself needs to be a `@Bean` so it can find and register (de-)serializer beans.

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigYamlParsingFilterTests.kt
```diff
@@ -17,26 +23,67 @@ import io.mockk.just
...
 import strikt.assertions.isEqualTo
+import javax.servlet.FilterChain
+import javax.servlet.ServletRequest
+import javax.servlet.http.HttpServletRequestWrapper
 
+@SpringBootTest(
+  classes = [Ec2JsonTestConfiguration::class],
+  webEnvironment = NONE
+)
 class DeliveryConfigYamlParsingFilterTests : JUnit5Minutests {
```
luispollo, line 48: @robfletcher This test is currently failing, even with the configuration below., with this message:
```
com.fasterxml.jackson.databind.JsonMappingException: Cannot construct instance of `com.netflix.spinnaker.keel.core.api.SubmittedResource`, problem: Cannot construct instance of `com.netflix.spinnaker.keel.api.ec2.SecurityGroupRule` (no Creators, like default constructor, exist): abstract types either need to be mapped to concrete types, have custom deserializer, or contain additional type information
```
I've tried adding a `testRuntimeOnly` dependency to `keel-ec2-plugin`, since the package scanned for deserializers is in there, but no luck. Any ideas?

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/Ec2JsonTestConfiguration.kt
robfletcher: I don't really get what this class is for. Wouldn't any Spring tests that needed the EC2 plugin stuff to be in place be getting all of this from main configuration? Why do you need a special test configuration class?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigYamlParsingFilterTests.kt
luispollo: Issue was the usage of `@ConditionalOnMissingBean` on `SecurityGroupRuleDeserializer`. It doesn't work like I thought it did, and the bean is never instantiated even when there's no other one of the same type.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/Ec2JsonTestConfiguration.kt
luispollo: As discussed in chat, I had incorrectly assumed, based on other tests annotated with `@SpringBootTest`, that you had to manually specify `KeelApplication::class` as one of the configuration classes for it to load, and my rationale was to create a configuration that was much smaller, containing essentially the `ObjectMapper` that these kinds of tests need, especially since they were not loading a Spring application context before (I was wary of extra time to run the tests).

I'll try to use the main configuration and create an `application.properties` config file for each module where we have these such that I can at least disable certain things, especially liquibase migrations.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/DefaultSecurityGroupRuleDeserializer.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.netflix.spinnaker.keel.api.ec2.CidrRule
+import com.netflix.spinnaker.keel.api.ec2.CrossAccountReferenceRule
+import com.netflix.spinnaker.keel.api.ec2.ReferenceRule
+import com.netflix.spinnaker.keel.api.ec2.SecurityGroupRule
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.boot.jackson.JsonComponent
+
+@JsonComponent
+//@ConditionalOnMissingBean(SecurityGroupRuleDeserializer::class)
```
luispollo, line 12: Unfortunately, this did not work. Spring never instantiates this bean, even when there's not another one that's a sub-class of `SecurityGroupRuleDeserializer`. It's something to do with timing/ordering, as far as I understand from the docs and StackOverflow questions.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/DefaultSecurityGroupRuleDeserializer.kt
```diff
@@ -0,0 +1,21 @@
...
+@ConditionalOnProperty("keel.plugins.ec2.defaultSerializers", matchIfMissing = true)
```
luispollo, line 13: This was the best alternative I could think of for now...

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/Ec2JsonTestConfiguration.kt
luispollo: PR updated to use `KeelApplication` configuration for all new `SpringBootTest` tests.

keel-ec2-plugin/keel-ec2-plugin.gradle.kts
```diff
@@ -22,10 +22,20 @@ dependencies {
...
   testImplementation("org.junit.jupiter:junit-jupiter-params")
+
+  // the following are needed to use keel's real(-ish) Spring configuration
+  testImplementation(project(":keel-web")) {
+    // avoid circular dependency
+    exclude(module = "keel-ec2-plugin")
+  }
+  testImplementation("org.testcontainers:mysql:${property("testContainersVersion")}")
+  testImplementation(project(":keel-sql"))
+  testImplementation("org.liquibase:liquibase-core")
```
robfletcher, line 40: It does feel like it would be easier to move the one test to `keel-web`

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/DeserializationInjectionTests.kt
```diff
@@ -16,13 +18,20 @@ import strikt.assertions.isNotNull
...
   webEnvironment = NONE
 )
 internal class DeserializationInjectionTests {
 
   @Autowired
-  lateinit var mapper: ObjectMapper
+  lateinit var mapper: YAMLMapper
+
+  @Autowired
+  lateinit var liquibase: SpringLiquibase
```
robfletcher, line 34: Why are you wiring this in? Doesn't look like it's used

robfletcher (approved):

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/api/ec2/DeserializationInjectionTests.kt
luispollo: Left-over from trying to disable migrations. Will remove!

keel-ec2-plugin/keel-ec2-plugin.gradle.kts
luispollo: In hindsight, yes it does. I'll try to tackle that in a separate PR.

PR #1744, opened 2021-01-20
fix(titus): add new env variable

emjburns:
We now add a `SPINNAKER_ACCOUNT` env variable for titus (https://github.com/spinnaker/clouddriver/pull/5185/files). This PR introduces a resolver to add it to specs so that we don't diff forever on it.


lorin (approved):
My understanding

## Problem

Clouddriver automatically adds a `SPINNAKER_ACCOUNT` environment variable to Titus ASGs. Keel sees this variable set in the current state of a Titus cluster, but not on the desired state, which triggers a diff that never gets resolved.

## Proposed solution

Add a new Resolver class that injects the `SPINNAKER_ACCOUNT` environment variable into Titus cluster specs.

## Feedback

LGTM

robfletcher (approved):
👍

PR #1748, opened 2021-01-21
fix(titus): Fix interestingHealthProviderNames in deploy stage

luispollo:
We were passing `"Amazon"` instead of `"Titus"` for `interestingHealthProviderNames` in the Titus deploy stage. This fixes that and adds a small test.


emjburns (approved):
oopsie. nice!

PR #1750, opened 2021-01-22
chore(logs & metrics): remove debug statements, add metrics

emjburns:
removing some debug statements that print a lot and instead creating some more metrics so we can alert when things don't happen.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -96,16 +100,12 @@ class CheckScheduler(
         }
       }
       runBlocking { job.join() }
-      log.debug("Scheduled resource validation complete")
```
robfletcher, line 99: The value (such as it is) of these logs is in just knowing the loops didn't get stuck. We'll be able to see that from telemetry, right?

robfletcher (approved):
I like this. It is good.

gal-yardeni (approved):
LGTM

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
emjburns: Yep, I added a duration metric, so if we see that increasing we will know that something is funky. Or at least, that is my theory.

luispollo (approved):
Good cleanup!

While you're at it, I found the other message that is always confusing to me, if you wanna change that. It's this line: https://github.com/spinnaker/keel/blob/66d7fac0a1a6901c6165f1ccf1f3d3c6074ec326/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/ResourceActuator.kt#L113

To me, that always reads backwards: if the check is stale, we should check again, not skip checks, but that's obviously not what this is doing, so maybe we can just improve the message?

PR #1751, opened 2021-01-22
feat(slack): more slack notifications

gal-yardeni:
**Context**

Currently, we utilize echo to send slack notifications to users, and we don't have much control over the payload, interactive notifications or the new features slack provides -- which makes it very hard to add/modify notifications related to keel.

**Work in this PR**

Improved the existing design + added new notifications, like:
- pinning
- unpinning
- pause management
- resume management
- bake/build failed
- mark artifact as bad (example:)
Screen Shot 2021-01-21 at 7 49 53 PM

**Future work (in following PRs)**
- [ ] Deprecate notifications coming from orca and replace them with new notifications, like "deployment failed/ succeeded."
- [ ] Migrating the "manual judgment" (or: interactive notifications) to the new implementation
- [ ] Update notifications like pinning & paused notifications --> if it was unpinned/resumed, update the original notification.

**Looking for feedback**
Any feedback about the design is desired!


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/NotificationType.kt
```diff
@@ -6,5 +6,9 @@ package com.netflix.spinnaker.keel.notifications
 enum class NotificationType {
```
luispollo, line 6: Total nit, but it might make the enum values easier to read here if we reversed the order of the words so that the scope appears first:
```kotlin
enum class NotificationType {
  RESOURCE_UNHEALTHY,
  ARTIFACT_PINNED,
  ARTIFACT_UNPINNED,
  ARTIFACT_MARKED_AS_BAD,
  APPLICATION_PAUSED,
  APPLICATION_RESUMED,
  LIFECYCLE_EVENT
}
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -26,21 +32,26 @@ class NotificationEventListener(
...
-        if (notificationConfig.type == NotificationType.slack) {
-          val handler = handlers.find {
-            it.type == notification.type
-          } as? SlackNotificationHandler<SlackPinnedNotification>
-          handler?.let {
-            it.sendMessage(SlackPinnedNotification(
-              channel = notificationConfig.address,
+      val handler = handlers.find {
+        it.type == notification.type
+      } as? SlackNotificationHandler<SlackPinnedNotification> ?: return
```
luispollo, line 44: Since this is showing up in more places now, might be worth extracting it into that `supporting` function as suggested earlier:
```kotlin
fun <T : SlackNotificationEvent> Collection<SlackNotificationHandler<*>>.supporting(type: NotificationType): SlackNotificationHandler<T> =
  this.find { it. type == type } as? SlackNotificationHandler<T>
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
-          } else {
-            log.debug("no pinned artifacts exists for application ${config.application} and environment $targetEnvironment")
+      if (pinnedEnvironment == null) {
+        log.warn("no pinned artifacts exists for application ${config.application} and environment $targetEnvironment")
+        return
+      }
+
+      val latestApprovedArtifactVersion = repository.latestVersionApprovedIn(config, pinnedEnvironment!!.artifact, targetEnvironment)
+      if (latestApprovedArtifactVersion == null) {
+        log.warn("latestApprovedArtifactVersion is null for application ${config.application}, env $targetEnvironment. Can't send UnpinnedNotification")
```
luispollo, line 77:
```suggestion
        log.debug("latestApprovedArtifactVersion is null for application ${config.application}, env $targetEnvironment. Can't send UnpinnedNotification")
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
-                application = config.application,
-                time = clock.instant(),
-                user = user
-              ))
-            }
-
-          } else {
-            log.debug("no pinned artifacts exists for application ${config.application} and environment $targetEnvironment")
+      if (pinnedEnvironment == null) {
+        log.warn("no pinned artifacts exists for application ${config.application} and environment $targetEnvironment")
```
luispollo, line 71:
```suggestion
        log.debug("no pinned artifacts exists for application ${config.application} and environment $targetEnvironment")
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -26,21 +32,26 @@ class NotificationEventListener(
...
-            it.sendMessage(SlackPinnedNotification(
-              channel = notificationConfig.address,
+      val handler = handlers.find {
+        it.type == notification.type
+      } as? SlackNotificationHandler<SlackPinnedNotification> ?: return
+
+      if (pinnedArtifact == null || currentArtifact == null) {
+        log.warn("can't send notification as either pinned artifact or current artifacts information is missing")
+      } else {
+        config.environments.forEach { environment ->
```
luispollo, line 49: Why are we looping on every environment now? The pin is always for a specific environment, right?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
+      }
+
+      val latestArtifact = repository.getArtifactVersion(pinnedEnvironment!!.artifact, latestApprovedArtifactVersion, null)
+      val pinnedArtifact = repository.getArtifactVersion(pinnedEnvironment!!.artifact, pinnedEnvironment!!.version, null)
+
+      val handler = handlers.find {
+        it.type == notification.type
+      } as? SlackNotificationHandler<SlackUnpinnedNotification> ?: return
+
+      config.environments.forEach { environment ->
```
luispollo, line 88: Same question on looping over environments.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
+      }
+    }
+  }
+
+  @EventListener(MarkAsBadNotification::class)
+  fun onMarkAsBadNotification(notification: MarkAsBadNotification) {
+    with(notification) {
+      val deliveryArtifact = config.artifacts.find {
+        it.reference == veto.reference
+      } ?: return
```
luispollo, line 111: Debug log before returning?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
+        log.warn("latestApprovedArtifactVersion is null for application ${config.application}, env $targetEnvironment. Can't send UnpinnedNotification")
+        return
+      }
+
+      val latestArtifact = repository.getArtifactVersion(pinnedEnvironment!!.artifact, latestApprovedArtifactVersion, null)
+      val pinnedArtifact = repository.getArtifactVersion(pinnedEnvironment!!.artifact, pinnedEnvironment!!.version, null)
+
+      val handler = handlers.find {
+        it.type == notification.type
+      } as? SlackNotificationHandler<SlackUnpinnedNotification> ?: return
```
luispollo, line 86: Debug log before returning?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
luispollo, line 44: Debug log before returning?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
+  @EventListener(MarkAsBadNotification::class)
+  fun onMarkAsBadNotification(notification: MarkAsBadNotification) {
+    with(notification) {
+      val deliveryArtifact = config.artifacts.find {
+        it.reference == veto.reference
+      } ?: return
+
+      val vetoedArtifact = repository.getArtifactVersion(deliveryArtifact, veto.version, null)
+      if (vetoedArtifact == null) {
+        log.warn("vetoedArtifact is null for application ${config.application}. Can't send MarkAsBadNotification notification")
```
luispollo, line 115:
```suggestion
        log.debug("vetoedArtifact is null for application ${config.application}. Can't send MarkAsBadNotification notification")
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
+        return
+      }
+
+      val handler = handlers.find {
+        it.type == notification.type
+      } as? SlackNotificationHandler<SlackMarkAsBadNotification> ?: return
```
luispollo, line 121: Debug log before returning?

luispollo: P.S. One more reason for that `supporting` function: you can log in a single place when the handler is not found.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
+
+      config.environments.forEach { environment ->
```
luispollo, line 123: Same question on looping over environments.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
+  @EventListener(ApplicationActuationPaused::class)
+  fun onApplicationActuationPaused(notification: ApplicationActuationPaused) {
+    with(notification) {
+      val config = repository.getDeliveryConfigForApplication(application)
+
+      val handler = handlers.find {
+        it.type == Type.PAUSED_APPLICATION
+      } as? SlackNotificationHandler<SlackPausedNotification> ?: return
+
+      config.environments.forEach { environment ->
```
luispollo, line 150: Same question on looping over environments. BTW, this block of code repeats many places, so it's a good candidate to extract into a function. It could look something like:

```kotlin
fun Environment.sendSlackMessage(message: SlackNotificationEvent) {
  this.notifications.filter {
    it.type == NotificationType.slack
  }.forEach {
    handlers.supporting(message.type)?.sendMessage(message)
  }
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
+  @EventListener(ApplicationActuationResumed::class)
+  fun onApplicationActuationResumed(notification: ApplicationActuationResumed) {
+    with(notification) {
+      val config = repository.getDeliveryConfigForApplication(application)
+
+      val handler = handlers.find {
+        it.type == Type.PAUSED_APPLICATION
+      } as? SlackNotificationHandler<SlackResumedNotification> ?: return
+
+      config.environments.forEach { environment ->
```
luispollo, line 174: Same comment/suggestion on looping over environments.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
+      if (artifact == null) {
+        log.warn("vetoedArtifact is null for application ${config.application}. Can't send MarkAsBadNotification notification")
+        return
+      }
+
+      val handler = handlers.find {
+        it.type == Type.LIFECYCLE_EVENT
+      } as? SlackNotificationHandler<SlackLifecycleNotification>? ?: return
+
+      config.environments.forEach { environment ->
```
luispollo, line 207: Same comment/suggestion on looping over environments.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -56,30 +67,156 @@ class NotificationEventListener(
...
+          ))
+        }
+      }
+    }
+  }
+
+  @EventListener(LifecycleEvent::class)
+  fun onLifecycleEvent(notification: LifecycleEvent) {
+    with(notification) {
+      val config = repository.getDeliveryConfig(artifactRef.split(":")[0])
```
luispollo, line 192: This logic of splitting the artifact reference and picking the right piece looks like a good candidate for functions/getters in `LifecyleEvent`.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/SlackService.kt
```diff
@@ -45,4 +46,17 @@ class SlackNotifier(
...
+    log.debug("lookup user id for email $email")
+    val response = slack.methods(configToken).usersLookupByEmail { req ->
+      req.email(email)
+    }
+    log.debug("slack returned ${response.isOk}")
+
+    if (response.user != null && response.user.name != null) {
+      return "@${response.user.name}"
+    }
+    return email
```
luispollo, line 60: Looks like you're falling back to the email if you can't retrieve the user name, so the return type doesn't have to be nullable.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/SlackService.kt
```diff
@@ -45,4 +46,17 @@ class SlackNotifier(
       log.debug("new slack integration is not enabled")
     }
   }
+
+  fun getUserInfoByEmail(email: String): String? {
```
luispollo, line 50: This is not really returning user info but a user "handle". Consider renaming:
```suggestion
  fun getUserHandleByEmail(email: String): String? {
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.netflix.spinnaker.keel.api.artifacts.GitMetadata
+import com.netflix.spinnaker.keel.api.artifacts.PublishedArtifact
+import com.netflix.spinnaker.keel.artifacts.getScmBaseLink
+import com.slack.api.model.kotlin_extension.block.SectionBlockBuilder
+import org.apache.logging.log4j.util.Strings
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+
+@Component
+class GitDataGenerator(
```
luispollo, line 13: Please document the class.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: good call! I've refactored that piece of the code to get the notifications from the config, instead of making another call to the repository to fetch the notifications only. Will fix it!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
```diff
@@ -0,0 +1,65 @@
...
+  private val scmInfo: ScmInfo,
+  @Value("\${spinnaker.baseUrl}") private val spinnakerBaseUrl: String
+) {
+
+  fun generateRepoLink(gitMetadata: GitMetadata?): String {
```
luispollo, line 18: I find these functions with a single nullable parameter a bit weird. Since the function cannot do anything if the parameter is null, in my opinion it should not be nullable, and the caller should be responsible for checking that it has the data needed to make the call.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
```diff
@@ -0,0 +1,65 @@
...
+    if (gitMetadata != null) {
+      val baseScmUrl = gitMetadata.commitInfo?.link?.let { getScmBaseLink(scmInfo, it) }
+      return "$baseScmUrl/projects/${gitMetadata.project}/repos/${gitMetadata.repo?.name}"
```
luispollo, line 21: This doesn't make a lot of sense to me because, even though you are determining the base URL dynamically, the rest of the URL is hard-coded (presumably for Stash?). This kind of function should be defined in an interface/abstract class and implemented for each SCM type.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /**
+   * generateGitData will create a slack section blocks, which looks like:
+   * "spkr/keel › PR#7 › master › c25a357"
+   * Or: "spkr/keel › master › c25a358" (no PR data)
+   * Each component will have the corresponding link attached to SCM
+   */
+  fun generateData(sectionBlockBuilder: SectionBlockBuilder, application: String, artifact: PublishedArtifact): SectionBlockBuilder {
+//  protected fun SectionBlockBuilder.generateGitData(application: String, artifact: PublishedArtifact) {
```
luispollo, line 32: Remove?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
```diff
@@ -0,0 +1,65 @@
...
+    with(sectionBlockBuilder) {
+      var details = ""
+      val artifactUrl = "$spinnakerBaseUrl/#/applications/${application}/environments/${artifact.reference}/${artifact.version}"
+
+      if (artifact.gitMetadata != null) {
+        val repoLink = generateRepoLink(artifact.gitMetadata)
+        with(artifact.gitMetadata) {
+
+          details += "<$repoLink|${this?.project}/" +
+            "${this?.repo?.name}> › " +
```
luispollo, line 42: I can't tell what exactly will end up in the string if any of these things are null. Are you sure this will render as intended in those cases?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
```diff
@@ -0,0 +1,65 @@
...
+            "<$repoLink/branches|${this?.branch}> › "
+
+          //if the commit is not a part of a PR, don't include PR's data
+          if (Strings.isNotEmpty(this?.pullRequest?.number)) {
+            details += "<${this?.pullRequest?.url}|PR#${this?.pullRequest?.number}> ›"
+          }
+          markdownText(details +
+            "<${this?.commitInfo?.link}|${this?.commitInfo?.sha?.substring(0, 7)}>")
```
luispollo, line 50: Same question on nullable properties when rendering the link.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
```diff
@@ -0,0 +1,65 @@
...
+        }
+      }
+      accessory {
+        button {
+          text("More...")
+          //TODO: figure out which action id to send here
```
luispollo, line 56: Since the button has a fully parameterized URL, my guess is that the `actionId` doesn't matter here.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/LifecycleEventNotificationHandler.kt
```diff
@@ -0,0 +1,62 @@
...
+
+  override val type: NotificationType = NotificationType.LIFECYCLE_EVENT
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  override fun sendMessage(notification: SlackLifecycleNotification) {
+    log.debug("Sending pinned artifact notification for application ${notification.application}")
+
+    with(notification) {
+      val imageUrl = when (type) {
+          LifecycleEventType.BAKE -> "https://raw.githubusercontent.com/gcomstock/managed.delivery/master/src/icons/bake_fail.png"
```
luispollo, line 32: We could host these in `spinnaker.io` if you don't want to use Greg's personal GitHub URL.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
luispollo: Sounds good. I made a suggestion for how to refactor this into a shared function that you can use for all notifications here: https://github.com/spinnaker/keel/pull/1751#discussion_r562822794

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/MarkAsBadNotificationHandler.kt
```diff
@@ -0,0 +1,61 @@
...
+
+      val vetoedArtifactUrl = "$spinnakerBaseUrl/#/applications/${application}/environments/${vetoedArtifact.reference}/${vetoedArtifact.version}"
+      val blocks = withBlocks {
+        header {
+          text("#${vetoedArtifact.buildMetadata?.number} Marked as bad in $env", emoji = true)
+        }
+
+        section {
+          markdownText("*Version:* <$vetoedArtifactUrl|#${vetoedArtifact.buildMetadata?.number}> " +
+            "by @${vetoedArtifact.gitMetadata?.author}\n " +
```
luispollo, line 40: I thought this is one place where we would use your new `getUserInfoByEmail` function to get the user `@` handle?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/PinnedNotificationHandler.kt
```diff
@@ -1,86 +1,62 @@
...
       val blocks = withBlocks {
         header {
           text("$env is pinned", emoji = true)
         }
 
         section {
-          markdownText("*Version:* ~#${notification.currentArtifact?.buildMetadata?.number}~ → <$pinnedArtifactUrl|#${pinnedArtifact?.buildMetadata?.number}> " +
-            "by ${pinnedArtifact?.gitMetadata?.author}\n " +
+          markdownText("*Version:* ~#${notification.currentArtifact.buildMetadata?.number}~ → <$pinnedArtifactUrl|#${pinnedArtifact.buildMetadata?.number}> " +
+            "by @${pinnedArtifact.gitMetadata?.author}\n " +
```
luispollo, line 40:
```suggestion
            "by $username\n " +
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/ResumedNotificationHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.netflix.spinnaker.keel.notifications.NotificationType
+import com.netflix.spinnaker.keel.slack.SlackService
+import com.netflix.spinnaker.keel.slack.SlackResumedNotification
+import com.slack.api.model.kotlin_extension.block.withBlocks
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+
+/**
+ * Sends notification when resuming managment for an application
```
luispollo, line 12:
```suggestion
 * Sends notification when resuming management for an application
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/UnpinnedNotificationHandler.kt
```diff
@@ -2,21 +2,63 @@ package com.netflix.spinnaker.keel.slack.handlers
...
+      val username = slackService.getUserInfoByEmail(user)
+
+
+      val blocks = withBlocks {
+        header {
+          text("$env was unpinned", emoji = true)
+        }
+
+        section {
+          markdownText("*Version:* ~#${pinnedArtifact?.buildMetadata?.number}~ → <$artifactUrl|#${latestArtifact?.buildMetadata?.number}> " +
```
luispollo, line 40: Same comment on how this will render when any of these things are null. This question applies everywhere where you're using the null-safe operator `?.` in these notifications. I [tested this in the Kotlin playground](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS40LjAiLCJjb2RlIjoiXG5mdW4gbWFpbigpIHtcbiAgdmFsIHNvbWV0aGluZ051bGxhYmxlID0gbnVsbFxuICB2YWwgc3RyaW5nV2l0aFNhZmVOdWxsQ2hlY2sgPSBcIldoYXQgaGFwcGVucyBpZiAke3NvbWV0aGluZ051bGxhYmxlPy50b1N0cmluZygpfSBpcyBwYXNzZWQ/XCJcbiAgcHJpbnRsbihzdHJpbmdXaXRoU2FmZU51bGxDaGVjaylcbn0iLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIn0=) and it renders `null` in the middle of the string, which I'm not sure is what you want.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/SlackService.kt
gal-yardeni: what does `user handle` means? 🤔

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
gal-yardeni: will renaming it to stash link would help? I don't think of a way (or reason, for now) to make it more generic.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
gal-yardeni: I agree. Will think about it

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/LifecycleEventNotificationHandler.kt
gal-yardeni: I just spoke with @erikmunson about it. We agreed it can be hosted on GitHub for now (as it's not on VPN, and does not require auth), and we can refactor it later as we think of where can be the best place that we could host those images. I'll mark at as a todo.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/MarkAsBadNotificationHandler.kt
gal-yardeni: nope -- this should show the author of the commit, not the person who actually marks the artifact as bad.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/PinnedNotificationHandler.kt
gal-yardeni: see my comment above

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: done!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: done!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: fixed

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: done!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: fixed

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: DONE! thank you!!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: we should loop on the entire environments here, as it applies for all of them (correct me if I'm wrong)

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: same for here - we should notify all environments

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/SlackService.kt
gal-yardeni: done

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
gal-yardeni: I believe the action id is applicable for interactive notifications only. For non-interactive notifications, it is probably doesn't matter (however it is still mandatory for the API call 🤷 )

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
gal-yardeni: adressed it

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
gal-yardeni: added more null checks on top.

lorin (commented):
Still in the process of reviewing, but I wanted to get out my [comment](https://github.com/spinnaker/keel/pull/1751#discussion_r564861902)  about modifying an enum that's stored in the database.

keel-slack/src/test/kotlin/com/netflix/spinnaker/keel/slack/SlackNotifierTest.kt
```diff
@@ -15,7 +16,7 @@ class SlackNotifierTest: JUnit5Minutests {
     val slackConfiguration = SlackConfiguration()
     val springEnv: Environment = mockk(relaxed = true)
 
-    val subject = SlackNotifier(
+    val subject = SlackService(
```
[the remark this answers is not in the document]
gal-yardeni: done!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/NotificationType.kt
```diff
@@ -4,7 +4,11 @@ package com.netflix.spinnaker.keel.notifications
  * All valid notifiers
  */
 enum class NotificationType {
-  UNHEALTHY_RESOURCE,
-  PINNED_ARTIFACT,
-  UNPINNED_ARTIFACT
+  RESOURCE_UNHEALTHY,
```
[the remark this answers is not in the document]
gal-yardeni: good call!
we had only `UNHEALTHY_RESOURCE` prior to my change. As I know from @emjburns, these notifications are currently shut down. I do see 7 records in the `notification` table in keel main db though-- Emily, should we delete those?

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/SlackService.kt
```diff
@@ -14,23 +14,24 @@ import org.springframework.stereotype.Component
...
+
   private val isSlackEnabled: Boolean
     get() = springEnv.getProperty("keel.notifications.slack", Boolean::class.java, true)
 
-
   fun sendSlackNotification(channel: String, blocks: List<LayoutBlock>, token: String? = null) {
     if (isSlackEnabled) {
-      log.debug("starting slack notification")
-      val slack = Slack.getInstance()
+      log.debug("sending slack notification for channel $channel")
```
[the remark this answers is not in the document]
gal-yardeni: what do you mean by ` increment a counter metric`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/NotificationType.kt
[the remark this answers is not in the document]
emjburns: @gal-yardeni yep we can delete those. Good catch @lorin !

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -134,7 +135,12 @@ class ApplicationService(
     if (!succeeded) {
       throw InvalidVetoException(application, veto.targetEnvironment, veto.reference, veto.version)
     }
-    // TODO: publish ArtifactVetoedEvent
```
emjburns, line 137: 👏  ayyyy! love deleting todos

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -26,64 +33,188 @@ class NotificationEventListener(
...
+            currentArtifact = currentArtifact,
+            pinnedArtifact = pinnedArtifact,
+            application = config.application,
+            time = clock.instant()
+          ),
+            Type.ARTIFACT_PINNED)
         }
+      } catch (ex: NoSuchElementException) {
+        log.debug("no environment ${pin.targetEnvironment} was found in the config named ${config.name}")
+        throw ex
```
emjburns, line 63: Does this warrant throwing an exception? Or would it be better to log an error and alert on this? 

(I don't know the answer. I just tend to question throwing exceptions because we don't always handle that nicely downstream)

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/SlackService.kt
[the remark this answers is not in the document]
emjburns: Here's an example: https://github.com/spinnaker/keel/blob/master/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryListener.kt#L34-L45

Lorin means publishing a metric for this so that we can graph it in atlas / alert on things like a huge increase in # of messages sent

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: hmm good question. honestly, this whole notification functionality won't really affect the user if something fails, as the user will get an error only if the actual pinning/ unpinning/ etc. will fail. This exception is mainly for us and won't be returned to the user. Not sure what are we considered as a best practice here..

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/SlackService.kt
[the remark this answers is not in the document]
gal-yardeni: ok i've added some counter metrics, hope it looks ok. will add more when I'll add more logic!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/notifications/NotificationType.kt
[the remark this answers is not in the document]
luispollo: Good catch indeed. I'll try to pay more attention when making these suggestions for renaming in the future!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
luispollo: I'd recommend just logging an error. Also, you could simplify this as follows:
```kotlin
config.environments.find { it.name == pin.targetEnvironment }
  .also { if (it == null) log.error(...) }
  ?.sendSlackMessage(...)
```
That way you don't have to mess with exception handling in the first place.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/UnpinnedNotificationHandler.kt
```diff
@@ -2,21 +2,67 @@ package com.netflix.spinnaker.keel.slack.handlers
...
+      val artifactUrl = "$spinnakerBaseUrl/#/applications/${application}/environments/${latestArtifact?.reference}/${latestArtifact?.version}"
+      val username = slackService.getUsernameByEmail(user)
+
+
+      val blocks = withBlocks {
+        header {
+          text("$env was unpinned", emoji = true)
+        }
+
+        section {
```
emjburns, line 39: nit: weird spacing below. Idk why spotless didn't fix it for you..

emjburns (approved):

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
luispollo: Oh, right.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
luispollo: Hmm, I would think that the lifecycle events should only notify for the environments in which the corresponding artifact is used. I believe there are some utility functions to find out which artifact belongs to which environment.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
luispollo: It's just that it's inconsistent with the fact that we keep everything else about SCM providers generic. This wouldn't work for anyone trying to use this code from the OSS community unless they happen to be using Stash, for instance. Having said that, given that we're by far the primary users of keel (if not the only ones), I'd say the priority of refactoring is lower and leave it up to you to decide when to address it.

luispollo: P.S. At the very least, I'd recommend adding a `TODO` stating that this only works for Stash at the moment.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
```diff
@@ -29,27 +33,33 @@ class GitDataGenerator(
...
+            }
+            //if the commit is not a part of a PR, don't include PR's data
+            if (Strings.isNotEmpty(pullRequest?.number)) {
+              details += "<${pullRequest?.url}|PR#${pullRequest?.number}> ›"
+            }
 
-          //if the commit is not a part of a PR, don't include PR's data
-          if (Strings.isNotEmpty(this?.pullRequest?.number)) {
-            details += "<${this?.pullRequest?.url}|PR#${this?.pullRequest?.number}> ›"
+            if (commitInfo != null && commitInfo!!.sha != null && commitInfo!!.sha?.length!! > 7) {
```
luispollo, line 54:
```suggestion
            if (commitInfo != null && commitInfo!!.sha != null && commitInfo!!.sha?.length!! >= 7) {
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/SlackNotificationHandler.kt
```diff
@@ -11,12 +11,19 @@ import com.netflix.spinnaker.keel.slack.SlackNotificationEvent
...
+): SlackNotificationHandler<T>? =
+  this.find { it.type == type } as? SlackNotificationHandler<T>
+
+
+fun <T : SlackNotificationEvent> Collection<SlackNotificationHandler<*>>.supporting(
+  type: Class<T>
+): SlackNotificationHandler<T>? =
+  this.find { handler ->
+    handler.javaClass == type }
+    as? SlackNotificationHandler<T>
```
luispollo, line 29:
```suggestion
  this.find { it.javaClass == type } as? SlackNotificationHandler<T>
```

keel-slack/src/test/kotlin/com/netflix/spinnaker/keel/slack/handlers/NotificationEventListenerTests.kt
```diff
@@ -93,10 +157,42 @@ class NotificationEventListenerTests : JUnit5Minutests {
...
+        subject.onPinnedNotification(pinnedNotification)
+        verify(exactly = 1) {
+          pinnedNotificationHandler.sendMessage(any(), any())
+        }
+        verify(exactly = 0) {
+          unpinnedNotificationHandler.sendMessage(any(), any())
+        }
+      }
+
+      test("only slack notifications are sending out") {
```
luispollo, line 170:
```suggestion
      test("only slack notifications are sent out") {
```

luispollo (commented):
This is looking much better, @gal-yardeni. There are a couple of things to clean up still which I've pointed out in separate comments.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/MarkAsBadNotificationHandler.kt
luispollo: Ah, right. My bad.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: I prefer to keep it as is for now.. I think it's more readable.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/SlackNotificationHandler.kt
gal-yardeni: I removed it for now as I'm not using it in the end :)

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: oops, you're right @luispollo . will change it.

gal-yardeni: done

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -26,64 +35,180 @@ class NotificationEventListener(
...
+
+    if (handler == null) {
+      log.debug("no handler was found for notification type ${T::class.java}. Can't send slack notification.")
+      return
+    }
+
+    //if targetEnvironment is not null, use only its notifications. Else, use all notifications configured for all environments.
+    val environments: Set<Environment> = config.environments.filter {
+      targetEnvironment == null || it.name == targetEnvironment
+    }.also { it ->
```
gal-yardeni, line 200: @luispollo @emjburns I added these filters as part of your comments, but I think I have a small bug here.. `artifact.isUsedIn(it)` is returning true even if the `resource` is empty 🤷 , and the test below is failing. any ideas?

luispollo: I'm looking, but in the meantime, you could simplify this by removing the `.also` block and just chaining the other `.filter`.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: fixed!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: @luispollo that was it!!! fixed.

PR #1752, opened 2021-01-23
feat(db): batch query of verification states

lorin:
Add support for querying verification states in batch

# Context

This is work towards supporting showing verification status in the environments view.

# Problem

The environments view needs to display verification states for a collection of different artifact versions and environments. The current implementation of the version repository only allows us to query one *VerificationContext* (artifact version, environment) at a time. This means that rendering the environments view would require a separate database query for every (version, environment) pair. 

Today the UI defaults to 30 versions. If there are 3 environments, that would mean 90 database calls.

# Proposed solution

Provide a batch query that API allows the caller to pass a collection of contexts at once, so we can get all of the verification info we need in a single query.

# Future work (fast follow PR)

Populate the verification data in the response payload of the API call that the environments view uses (`ApplicationController.get`)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -162,6 +168,77 @@ class SqlVerificationRepository(
...
+   * @param contexts a list of verification contexts to query for state
+   * @return a list of maps of verification ids to states, in the same order as the contexts. If there are no
+   *         verification states associated with a context, the resulting map will be empty.
+   */
+  override fun getStatesBatch(contexts: List<VerificationContext>) : List<Map<String, VerificationState>> {
+    /**
+     * In-memory database table representation of the set of contexts we want to query for
+     *
+     * Columns:
+     *   ind - index that encodes the original list order, to ensure results are in same order
```
robfletcher, line 188: Does it matter that it's in the same order?

robfletcher (approved):
I think this looks good. I'm not sure I 100% understand the in-memory table part but it seems like it's basically just a data class with some hooks that enable JOOQ to treat it like a table. If I'm right about that it's not as complex as it first appeared.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
```diff
@@ -377,4 +382,101 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
...
+    subject.updateState(c1, v2, FAIL)
+
+    // Second context: v1: PENDING, v2: PENDING
+    subject.updateState(c2, v1, PENDING)
+    subject.updateState(c2, v2, PENDING)
+
+
+    //
+    // Code under test
+    //
```
luispollo, line 450: Unnecessary?
```suggestion
```

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
```diff
@@ -377,4 +382,101 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
...
+    expectThat(stateMap[verification.id])
+      .isNotNull()
+      .with(VerificationState::status) { isEqualTo(PENDING) }
+  }
+
+  @Test
+  fun `multiple contexts, multiple verifications`() {
+    //
+    // setup
+    //
```
luispollo, line 424: Unnecessary?
```suggestion
```

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
```diff
@@ -377,4 +382,101 @@ abstract class VerificationRepositoryTests<IMPLEMENTATION : VerificationReposito
...
+      that(result[1])
+        .and {
+          get { get(v1.id) }
+            .isNotNull()
+            .with(VerificationState::status) { isEqualTo(PENDING) }
+        }
+        .and {
+          get { get(v2.id) }
+            .isNotNull()
+            .with(VerificationState::status) { isEqualTo(PENDING) }
```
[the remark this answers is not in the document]
luispollo: Or even simpler: 😊
```suggestion
            .get { status }.isEqualTo(PENDING)
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -162,6 +168,77 @@ class SqlVerificationRepository(
...
+
+      contextTable.IND,
+      VERIFICATION_STATE.VERIFICATION_ID,
+      VERIFICATION_STATE.STATUS,
+      VERIFICATION_STATE.STARTED_AT,
+      VERIFICATION_STATE.ENDED_AT,
+      VERIFICATION_STATE.METADATA)
+      .from(contextTable.table)
+      .leftJoin(ENVIRONMENT)
+      .on(ENVIRONMENT.NAME.eq(contextTable.ENVIRONMENT_NAME))
```
luispollo, line 209: Side-note: we should really stop using environment names as FKs...

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -162,6 +168,77 @@ class SqlVerificationRepository(
...
+   * This call is semantically equivalent to
+   *    contexts.map { context -> this.getStates(context) }
+   *
+   * However, it's implemented as a single query for efficiency.
+   *
+   * @param contexts a list of verification contexts to query for state
+   * @return a list of maps of verification ids to states, in the same order as the contexts. If there are no
+   *         verification states associated with a context, the resulting map will be empty.
+   */
+  override fun getStatesBatch(contexts: List<VerificationContext>) : List<Map<String, VerificationState>> {
```
luispollo, line 183: I was a bit confused by this return type. Can a `VerificationContext` have more than one verification ID? In other words, could it just be either a map of IDs to states, or a list in the same order as the contexts, instead of both?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -238,4 +315,41 @@ class SqlVerificationRepository(
...
+    val jooq: DSLContext
+  ) {
+    val alias = "verification_contexts"
+
+    private val ind = "ind"
+    private val environmentName = "environment_name"
+    private val artifactReference = "artifact_reference"
+    private val artifactVersion = "artifact_version"
+
+    fun <T> f(s : String, t: Class<T>) : Field<T> = field(name(alias, s), t)
```
luispollo, line 335:
```suggestion
    fun <T> typedField(s : String, t: Class<T>) : Field<T> = field(name(alias, s), t)
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -238,4 +315,41 @@ class SqlVerificationRepository(
...
+
+    /**
+     * return a jOOQ table that contains the [contexts] data represented as a table that can be selected against
+     */
+    val table : Table<Record4<Int, String, String, String>>
+      get() =
+        contexts
+          .mapIndexed { idx, v -> jooq.select(inline(idx), inline(v.environmentName), inline(v.artifactReference), inline(v.version)) as SelectOrderByStep<Record4<Int, String, String, String>> }
+          .reduce { s1, s2 -> s1.unionAll(s2) }
+          .asTable(alias, ind, environmentName, artifactReference, artifactVersion)
```
luispollo, line 352: I can't say I fully understand this block of code. Do you think it might be worth expanding the comment above to add a bit more detail about what this does, especially the map/reduce part?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -162,6 +168,77 @@ class SqlVerificationRepository(
...
+      // sort the results by the "ind" (index) column, so that outputs are same order as inputs
+      .groupBy { (index, _, _, _, _, _) -> index as Long }
+      .toSortedMap()
+      .values
+
+      // convert List<Record> to Map<String,VerificationState>, where the string is the verification id
+      .map { records ->
+        records
+          // since we do a left join, there may be rows where there is no corresponding records in the
+          // verification_state database, so we filter them out, which will result in an empty map
```
luispollo, line 232: > which will result in an empty map

Should we additionally filter out those empty maps from the list, then?

luispollo (approved):
Overall, this looks good to me, @lorin. Very clever usage of jooq extensibility -- I could never have come up with that myself, which made it a bit harder to review, so take my approval with a grain of salt.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
robfletcher: One verification id can apply to many artifact versions

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/VerificationRepositoryTests.kt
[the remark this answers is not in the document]
robfletcher: That's actually considerably slower because it does some code reading tomfuckery to generate the string representation of the path (that will get used if the assertion fails). When using a method reference it can just use its name. Unfortunately there's not really a way round that as you can't use reflection to parse the content of the lambda. You can do `get("status") { status }` which will avoid the code reading trick, but that's kinda ugly.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
luispollo: Ah. Got it.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
lorin: Yeah, the problem is that a `VerificationState` object doesn't carry information about which `VerificationContext` it's associated with, so if we just return a set of states, we don't know what they refer to.

We could modify `VerificationState` to have a reference to the context it's associated with.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
lorin: No, because the input list has to be the exact same size as the output list so we can match up VerificationContext objects with VerificationState objects. (See [this earlier comment](https://github.com/spinnaker/keel/pull/1752#discussion_r564813023)).

PR #1757, opened 2021-01-24
fix(diff): remove description from SecurityGroupRule

lorin:
Keel is still detecting diffs in security groups that haven't changed, despite the fix introduced in #1755.

This PR removes the description field that was added to a security group model in the hopes that this resolves the regression, so that we can deploy #1754.


lorin:
I didn't understand the change enough to do the reversion, closing this for now.

PR #1758, opened 2021-01-24
fix(diff): Revert recent security group changes

lorin:
# Problem

Recent changes to security group models have resulted in flapping security groups.

# Proposed solution

This PR reverts PRs #1755 and #1739 to put master back in a deployable state.


lorin:
The issue was fixed in #1759, so we don't need this reversion any more.

PR #1759, opened 2021-01-25
fix(diff): Workaround for diff library's handling of SecurityGroup.inboundRules

luispollo:
It turns out that the [diff library](https://github.com/SQiShER/java-object-diff/) we use looks at object equality when determining changes in collections, so even though the `description` property of `CidrRule` was being excluded from the comparison, the entire object was considered different in the collection when there was a difference in the value of that property, due to the default implementation of `equals()` by data classes, which compares all fields.

This explains why we saw this kind of diff, where all the properties appear removed/added with the exact same values:
image

I opened an issue for the library here: https://github.com/SQiShER/java-object-diff/issues/216. In the meantime, the workaround is to override `hashCode()` and `equals()` in `CidrRule` to also ignore the `description` property.


lorin (approved):
+1 for linking to the issue

PR #1762, opened 2021-01-25
use bulk caches where appropriate

robfletcher:
Some CloudDriver endpoints return all data at once, rather than a specific item. This PR updates our CloudDriver caches to take that into account. This should result in more efficient cache utilization as we won't keep making the same CloudDriver call the first time each key is hit -- we'll make the call just once for the first key.

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
@@ -67,10 +67,11 @@ interface CloudDriverService {
     @Header("X-SPINNAKER-USER") user: String = DEFAULT_SERVICE_ACCOUNT
   ): SecurityGroupSummary
 
-  @GET("/networks")
+  @GET("/networks/{cloudProvider}")
   suspend fun listNetworks(
+    @Path("cloudProvider") cloudProvider: String,
     @Header("X-SPINNAKER-USER") user: String = DEFAULT_SERVICE_ACCOUNT
-  ): Map<String, Set<Network>>
+  ): Set<Network>
 
   @GET("/subnets/{cloudProvider}")
   suspend fun listSubnets(
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
@@ -57,7 +57,6 @@ class MemoryCloudDriverCache(
           }
       }
         .handleNotFound()
-        ?: notFound("Security group with id $id not found in the $account account and $region region")
     }
 
   private val securityGroupsByName: AsyncLoadingCache<Triple<String, String, String>, SecurityGroupSummary> = cacheFactory
@@ -73,34 +72,31 @@ class MemoryCloudDriverCache(
           }
       }
         .handleNotFound()
-        ?: notFound("Security group with name $name not found in the $account account and $region region")
     }
 
   private val networksById: AsyncLoadingCache<String, Network> = cacheFactory
-    .asyncLoadingCache(cacheName = "networksById") { id ->
+    .asyncBulkLoadingCache(cacheName = "networksById") {
       runCatching {
-        cloudDriver.listNetworks(DEFAULT_SERVICE_ACCOUNT)["aws"]
-          ?.firstOrNull { it.id == id }
-          ?.also {
-            networksByName.put(Triple(it.account, it.region, it.name), completedFuture(it))
-          }
+        cloudDriver.listNetworks("aws", DEFAULT_SERVICE_ACCOUNT)
+          .associateBy { it.id }
       }
-        .handleNotFound()
-        ?: notFound("VPC network with id $id not found")
+        .getOrElse { ex ->
+          throw CacheLoadingException("Error loading networksById cache", ex)
+        }
     }
 
   private val networksByName: AsyncLoadingCache<Triple<String, String, String?>, Network> = cacheFactory
-    .asyncLoadingCache(cacheName = "networksByName") { (account, region, name) ->
+    .asyncBulkLoadingCache(cacheName = "networksByName") {
       runCatching {
         cloudDriver
-          .listNetworks(DEFAULT_SERVICE_ACCOUNT)["aws"]
-          ?.firstOrNull { it.name == name && it.account == account && it.region == region }
-          ?.also {
-            networksById.put(it.id, completedFuture(it))
+          .listNetworks("aws", DEFAULT_SERVICE_ACCOUNT)
+          .associateBy {
+            Triple(it.account, it.region, it.name)
           }
       }
-        .handleNotFound()
-        ?: notFound("VPC network named $name not found in $region")
+        .getOrElse { ex ->
+          throw CacheLoadingException("Error loading networksByName cache", ex)
+        }
     }
 
   private data class AvailabilityZoneKey(
@@ -122,7 +118,7 @@ class MemoryCloudDriverCache(
           .toSet()
       }
         .getOrElse { ex ->
-          throw CacheLoadingException("Error loading cache", ex)
+          throw CacheLoadingException("Error loading availabilityZones cache", ex)
         }
     }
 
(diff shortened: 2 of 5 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/CloudDriverService.kt
```diff
@@ -67,10 +67,11 @@ interface CloudDriverService {
     @Header("X-SPINNAKER-USER") user: String = DEFAULT_SERVICE_ACCOUNT
   ): SecurityGroupSummary
 
-  @GET("/networks")
+  @GET("/networks/{cloudProvider}")
   suspend fun listNetworks(
+    @Path("cloudProvider") cloudProvider: String,
     @Header("X-SPINNAKER-USER") user: String = DEFAULT_SERVICE_ACCOUNT
-  ): Map<String, Set<Network>>
+  ): Set<Network>
```
robfletcher, line 74: I found that this different endpoint works fine and since we always use `aws` as the cloud provider…

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
```diff
@@ -73,34 +72,31 @@ class MemoryCloudDriverCache(
           }
       }
         .handleNotFound()
-        ?: notFound("Security group with name $name not found in the $account account and $region region")
     }
 
   private val networksById: AsyncLoadingCache<String, Network> = cacheFactory
-    .asyncLoadingCache(cacheName = "networksById") { id ->
+    .asyncBulkLoadingCache(cacheName = "networksById") {
```
robfletcher, line 78: Any time we load all the data at once we can use this factory method instead

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCacheTest.kt
```diff
@@ -246,6 +246,16 @@ internal class MemoryCloudDriverCacheTest {
...
 
+  @Test
+  fun `all certs are cached at once when requested by name`() {
+    every { cloudDriver.getCertificates() } returns certificates
+
+    listOf("cert-1", "cert-2")
+      .forEach(subject::certificateByName)
+
+    verify(exactly = 1) { cloudDriver.getCertificates() }
+  }
```
robfletcher, line 257: Test that cache is only hit once, even though multiple keys are queried

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/caffeine/CacheFactory.kt
```diff
@@ -66,16 +88,50 @@ class CacheFactory(
...
+      throw UnsupportedOperationException()
+    }
+
+    override fun asyncLoadAll(
+      keys: Iterable<K>,
+      executor: Executor
+    ): CompletableFuture<Map<K, V>> =
+      CoroutineScope(executor.asCoroutineDispatcher())
+        .future { this@toAsyncBulkCacheLoader.invoke() }
+  }
```
robfletcher, line 113: this type of cache loader only supports `asyncLoadAll` instead of `asyncLoad`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/caffeine/CacheFactory.kt
```diff
@@ -66,16 +88,50 @@ class CacheFactory(
...
+
+/**
+ * An implementation of [AsyncLoadingCache] that _always_ uses [AsyncCacheLoader.asyncLoadAll] to
+ * populate the cache.
+ */
+private class AsyncBulkLoadingCache<K : Any, V>(delegate: AsyncLoadingCache<K, V>) :
+  AsyncLoadingCache<K, V> by delegate {
+  override fun get(key: K): CompletableFuture<V> = getAll(listOf(key)).thenApply { it[key] }
```
robfletcher, line 121: Cache implementation that always maps calls to `get` to `getAll`. This way we can treat all caches the same at the points where we call them and don't need to think about whether they use bulk loading or regular loading.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/caffeine/CacheFactory.kt
```diff
@@ -66,16 +88,50 @@ class CacheFactory(
...
+  override fun get(key: K, mappingFunction: Function<in K, out V>): CompletableFuture<V> {
+    throw UnsupportedOperationException()
+  }
+
+  override fun get(
+    key: K,
+    mappingFunction: BiFunction<in K, Executor, CompletableFuture<V>>
+  ): CompletableFuture<V> {
+    throw UnsupportedOperationException()
+  }
```
robfletcher, line 132: We never use these `get` variants so I didn't bother to implement them and I don't think it really makes sense to anyway.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/caffeine/CacheFactory.kt
```diff
@@ -66,16 +88,50 @@ class CacheFactory(
...
+    override fun asyncLoad(key: K, executor: Executor): CompletableFuture<V> {
+      throw UnsupportedOperationException()
+    }
+
+    override fun asyncLoadAll(
+      keys: Iterable<K>,
+      executor: Executor
+    ): CompletableFuture<Map<K, V>> =
+      CoroutineScope(executor.asCoroutineDispatcher())
+        .future { this@toAsyncBulkCacheLoader.invoke() }
```
luispollo, line 112: This is throwing me out for a loop... It looks reentrant, but I know you know what you're doing. TBH I can't say I immediately understand most of the kotlinism-heavy code in this PR. 😓

robfletcher: This is an extension method on a function with a particular signature that wraps it up in a coroutine and returns the function's result in a `Future`. `this@toAsynBulkCacheLoader` is a reference to the function itself. It's not re-entrant, just decorates the function.

luispollo (approved):
LGTM! Hope that test doesn't stay flaky!

robfletcher:
> LGTM! Hope that test doesn't stay flaky!

Good job caches are a really simple thing in programming that never cause any unpredictability

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
```diff
@@ -57,7 +57,6 @@ class MemoryCloudDriverCache(
           }
       }
         .handleNotFound()
-        ?: notFound("Security group with id $id not found in the $account account and $region region")
```
lorin, line 60: What's the cache behavior when `handleNotFound()` returns null? Does it keep the cache unpopulated, or does it cache a miss (`null`)?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/caffeine/CacheFactory.kt
```diff
@@ -66,16 +88,50 @@ class CacheFactory(
...
+  }
+
+/**
+ * An implementation of [AsyncLoadingCache] that _always_ uses [AsyncCacheLoader.asyncLoadAll] to
+ * populate the cache.
+ */
+private class AsyncBulkLoadingCache<K : Any, V>(delegate: AsyncLoadingCache<K, V>) :
+  AsyncLoadingCache<K, V> by delegate {
```
lorin, line 120: This is neat, I hadn't seen this [Kotlin feature](https://kotlinlang.org/docs/reference/delegation.html#implementation-by-delegation) before.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/caffeine/CacheFactory.kt
```diff
@@ -66,16 +88,50 @@ class CacheFactory(
...
+  object : AsyncCacheLoader<K, V> {
+    override fun asyncLoad(key: K, executor: Executor): CompletableFuture<V> {
+      throw UnsupportedOperationException()
+    }
+
+    override fun asyncLoadAll(
+      keys: Iterable<K>,
+      executor: Executor
+    ): CompletableFuture<Map<K, V>> =
+      CoroutineScope(executor.asCoroutineDispatcher())
```
lorin, line 111: The Caffeine library could really use kotlin support so we don't have to do this suspend-function-to-Java future dance.

lorin (approved):
My understanding

## Context

For several API calls against river, the API call itself returns a collection of values, even though the caller only cares about one of them.

## Problem

For the API calls referenced above, Keel only caches the value that the caller was interested in. This is wasteful, it means we call river more than we have to, slowing keel down and increasing the load on river.

## Proposed solution

Cache all of the returned values from the river API calls that return multiple values.

## Implementation details

Adds a new class, `AsyncBulkLoadingCache`, which assumes that the supplied loader returns a map with all of the values to be cached. Overrides the `get` method to call `getAll` to ensure that Caffeine always calls `asyncLoadAll` on it.


## Feedback

LGTM

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/MemoryCloudDriverCache.kt
robfletcher: Keeps the cache unpopulated

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/caffeine/CacheFactory.kt
robfletcher: I think this is pretty much the first time I've found a good reason to use it

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/caffeine/CacheFactory.kt
robfletcher: Yeah, I wonder if it would be worth us contributing that

PR #1766, opened 2021-01-27
Fix stuck promotion checks

robfletcher:
Prior to this change any exception thrown trying to load delivery configs from the database as part of the `DeliveyrConfigRepository.itemsDueForCheck` method would roll back the timestamp update to indicate when that delivery config was last checked. That would mean we'd get stuck trying to read the same one over and over. Specifically the issue was observed due to a failing `SpecMigrator` which is what the test in this PR simulates.

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
@@ -1022,11 +1022,11 @@ class SqlDeliveryConfigRepository(
                 .execute()
             }
           }
-          .map { (_, name) ->
-            get(name)
-          }
       }
     }
+      .map { (_, name) ->
+        get(name)
+      }
   }
 
   override fun markCheckComplete(deliveryConfig: DeliveryConfig) {
--- a/keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/MigratorErrorTests.kt
+++ b/keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/MigratorErrorTests.kt
@@ -1,0 +1,125 @@
+package com.netflix.spinnaker.keel.sql
+
+import com.netflix.spinnaker.keel.api.DeliveryConfig
+import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.api.ResourceKind.Companion.parseKind
+import com.netflix.spinnaker.keel.api.plugins.kind
+import com.netflix.spinnaker.keel.resources.ResourceSpecIdentifier
+import com.netflix.spinnaker.keel.resources.SpecMigrator
+import com.netflix.spinnaker.keel.test.DummyResourceSpec
+import com.netflix.spinnaker.keel.test.configuredTestObjectMapper
+import com.netflix.spinnaker.keel.test.defaultArtifactSuppliers
+import com.netflix.spinnaker.keel.test.resource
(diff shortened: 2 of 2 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -1022,11 +1022,11 @@ class SqlDeliveryConfigRepository(
                 .execute()
             }
           }
-          .map { (_, name) ->
-            get(name)
-          }
       }
```
emjburns, line 1025: 🤣  of course

emjburns (approved):

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -1022,11 +1022,11 @@ class SqlDeliveryConfigRepository(
...
     }
+      .map { (_, name) ->
+        get(name)
+      }
```
robfletcher, line 1029: This is the fix -- once we have the candidates we want to load, do so _after_ the transaction has completed, otherwise the write to `DELIVERY_CONFIG_LAST_CHECKED.AT` will be rolled back if the read fails.

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/MigratorErrorTests.kt
```diff
@@ -0,0 +1,128 @@
...
+      .isFailure()
+      .isA<RuntimeException>()
+
+    // Subsequently we should not try to re-check the same delivery config (within the time window)
+    expectCatching {
+      deliveryConfigRepository.itemsDueForCheck(ofMinutes(1), 10)
+    }
+      .isSuccess()
+      .isEmpty()
+  }
```
robfletcher, line 112: Here's the test that confirms the behavior works as desired (and which failed before moving the load outside the tx).

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/MigratorErrorTests.kt
```diff
@@ -0,0 +1,128 @@
...
+    environments = setOf(
+      Environment(
+        name = "test",
+        resources = setOf(
+          resource(kind = kindV1.kind),
+          resource(kind = kindV2.kind)
+        )
+      )
+    )
+  )
```
robfletcher, line 87: Pretty simple test data -- a delivery config with 2 resources, one of which will fail to load because the migrator for its kind will throw an exception

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/MigratorErrorTests.kt
```diff
@@ -0,0 +1,125 @@
...
+
+  private val multiVersionResourceSpecIdentifier = ResourceSpecIdentifier(kindV1, kindV2)
+
+  private val bedShittingSpecMigrator =
+    object : SpecMigrator<DummyResourceSpec, DummyResourceSpec> {
+      override val input = kindV1
+      override val output = kindV2
+
+      override fun migrate(spec: DummyResourceSpec): DummyResourceSpec {
+        throw RuntimeException("💩🛏")
```
luispollo, line 48: 😆

luispollo (approved):
Great detective work and super nice tests, as usual! ❤️

PR #1767, opened 2021-01-28
Check certificate validity against Lemur and update if expired

robfletcher:
This PR restores the ALB spec v1.2 that lets users specify certificates for ALB listeners by name. It also:

- derives the name directly from the ARN when migrating older resource records on DB read, rather than querying CloudDriver. This is more resilient to error, since expired certificates may not be returned by CloudDriver and we'll get an error that potentially prevents an entire delivery config getting read.
- adds a resolver that checks certificate validity against Lemur and updates the name if the certificate has expired and a replacement exists. This means users do not need to update their delivery config when Lemur rotates certificates.

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/migrators/ApplicationLoadBalancerV1_1ToV1_2Migrator.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/migrators/ApplicationLoadBalancerV1_1ToV1_2Migrator.kt
@@ -7,17 +7,13 @@ import com.netflix.spinnaker.keel.api.ec2.EC2_APPLICATION_LOAD_BALANCER_V1_1
 import com.netflix.spinnaker.keel.api.ec2.EC2_APPLICATION_LOAD_BALANCER_V1_2
 import com.netflix.spinnaker.keel.api.ec2.old.ApplicationLoadBalancerV1_1Spec
 import com.netflix.spinnaker.keel.api.ec2.old.ApplicationLoadBalancerV1_1Spec.ListenerV1_1
-import com.netflix.spinnaker.keel.clouddriver.CloudDriverCache
 import com.netflix.spinnaker.keel.resources.SpecMigrator
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.stereotype.Component
 
 @Component
 @ConditionalOnProperty("keel.plugins.ec2.enabled")
-class ApplicationLoadBalancerV1_1ToV1_2Migrator(
-  private val cloudDriverCache: CloudDriverCache
-) :
-  SpecMigrator<ApplicationLoadBalancerV1_1Spec, ApplicationLoadBalancerSpec> {
+class ApplicationLoadBalancerV1_1ToV1_2Migrator : SpecMigrator<ApplicationLoadBalancerV1_1Spec, ApplicationLoadBalancerSpec> {
   @Suppress("DEPRECATION")
   override val input = EC2_APPLICATION_LOAD_BALANCER_V1_1
   override val output = EC2_APPLICATION_LOAD_BALANCER_V1_2
@@ -45,7 +41,7 @@ class ApplicationLoadBalancerV1_1ToV1_2Migrator(
       Listener(
         port = listener.port,
         protocol = listener.protocol,
-        certificate = listener.certificateArn?.let { cloudDriverCache.certificateByArn(it).serverCertificateName },
+        certificate = listener.certificateArn?.substringAfter(":server-certificate/"),
         rules = listener.rules,
         defaultActions = listener.defaultActions
       )
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/migrators/ApplicationLoadBalancerV1_1ToV1_2Migrator.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/migrators/ApplicationLoadBalancerV1_1ToV1_2Migrator.kt
@@ -1,0 +1,53 @@
+package com.netflix.spinnaker.keel.ec2.migrators
+
+import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec
+import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec.ApplicationLoadBalancerOverride
+import com.netflix.spinnaker.keel.api.ec2.ApplicationLoadBalancerSpec.Listener
+import com.netflix.spinnaker.keel.api.ec2.EC2_APPLICATION_LOAD_BALANCER_V1_1
+import com.netflix.spinnaker.keel.api.ec2.EC2_APPLICATION_LOAD_BALANCER_V1_2
+import com.netflix.spinnaker.keel.api.ec2.old.ApplicationLoadBalancerV1_1Spec
+import com.netflix.spinnaker.keel.api.ec2.old.ApplicationLoadBalancerV1_1Spec.ListenerV1_1
+import com.netflix.spinnaker.keel.clouddriver.CloudDriverCache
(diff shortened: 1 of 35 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/migrators/ApplicationLoadBalancerV1_1ToV1_2Migrator.kt
```diff
@@ -0,0 +1,49 @@
...
+        )
+      }
+    )
+
+  private fun Set<ListenerV1_1>.migrate(): Set<Listener> =
+    mapTo(mutableSetOf()) { listener ->
+      Listener(
+        port = listener.port,
+        protocol = listener.protocol,
+        certificate = listener.certificateArn?.substringAfter(":server-certificate/"),
```
robfletcher, line 44: Derive the certificate name from the ARN used in older ALB specs

keel-lemur/src/main/kotlin/com/netflix/spinnaker/config/LemurConfiguration.kt
```diff
@@ -0,0 +1,52 @@
...
+        )
+      }
+      .build()
+    return Retrofit.Builder()
+      .addConverterFactory(JacksonConverterFactory.create(objectMapper))
+      .baseUrl(lemurEndpoint)
+      .client(client)
+      .build()
+      .create(LemurService::class.java)
+  }
```
robfletcher, line 51: Access to the Lemur endpoint requires a token, which is already available & used by Orca & Gate

keel-lemur/src/main/kotlin/com/netflix/spinnaker/keel/lemur/LemurCertificateResolver.kt
```diff
@@ -0,0 +1,40 @@
...
+
+  private fun findCurrentCertificate(name: String) =
+    runBlocking {
+      val certificate = lemurService.certificateByName(name).items.first()
+      if (certificate.active) {
+        certificate.name
+      } else {
+        certificate.replacedBy.first { it.active }.name
+      }
+    }
```
robfletcher, line 39: Get the newer cert from lemur if the current one is no longer active

keel-lemur/src/main/kotlin/com/netflix/spinnaker/config/LemurConfiguration.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.http.HttpHeaders.AUTHORIZATION
+import retrofit2.Retrofit
+import retrofit2.converter.jackson.JacksonConverterFactory
+
+@Configuration
+@ConditionalOnProperty("lemur.base-url")
+class LemurConfiguration {
+  @Bean
+  fun lemurEndpoint(
+    @Value("\${lemur.base-url}") lemurBaseUrl: String
```
emjburns, line 22: I thought the prefered pattern for this stuff is to create a `LemurConfig` class with the properties in it, and wire that in via the `@EnableConfigurationProperties` annotation. I think it would be good to switch to that format so that we don't have the `@Value` annotations all over the place here.

luispollo: +1.

keel-lemur/src/main/kotlin/com/netflix/spinnaker/config/LemurConfiguration.kt
```diff
@@ -0,0 +1,52 @@
...
+    @Value("\${lemur.token}") token: String
+  ): LemurService {
+    val client = OkHttpClient
+      .Builder()
+      .addInterceptor { chain ->
+        chain.proceed(
+          chain
+            .request()
+            .newBuilder()
+            .header(AUTHORIZATION, "Bearer $token")
```
luispollo, line 40: That's nice and easy. Cool!

keel-lemur/src/main/kotlin/com/netflix/spinnaker/keel/lemur/LemurCertificateResolver.kt
```diff
@@ -0,0 +1,40 @@
...
+              certificate = findCurrentCertificate(name)
+            )
+          } ?: listener
+        }
+      )
+    )
+
+  private fun findCurrentCertificate(name: String) =
+    runBlocking {
+      val certificate = lemurService.certificateByName(name).items.first()
```
luispollo, line 33: I'm guessing both the list here and `certificate.replacedBy` can be empty, so perhaps use `firstOrNull()` and check the result so this doesn't throw `NoSuchElementException`?

keel-lemur/src/main/kotlin/com/netflix/spinnaker/config/LemurConfiguration.kt
robfletcher: Sure. I guess I felt like there was maybe a threshold of a number of properties where that became worthwhile, but I'm happy to do it.

keel-lemur/src/main/kotlin/com/netflix/spinnaker/config/LemurConfiguration.kt
robfletcher: Stole it from Orca, and just updated to the Retrofit 2 API

luispollo (approved):
Other than the config thing and the calls to `first()` that I pointed out, LGTM, so giving you an early approval. 🙂

keel-lemur/src/main/kotlin/com/netflix/spinnaker/keel/lemur/LemurCertificateResolver.kt
robfletcher: Yeah, I also wonder if it's potentially necessary to follow a chain longer than one link. I wasn't able to find any example in Lemur where that was the case, but just in case I'm going to add some further capabilities here.

keel-lemur/src/main/kotlin/com/netflix/spinnaker/keel/lemur/LemurCertificateResolver.kt
```diff
@@ -0,0 +1,49 @@
...
+    } else {
+      val replacement = certificate.replacedBy.firstOrNull { it.active }
+      replacement?.name
+        ?: if (certificate.replacedBy.isNotEmpty()) {
+          findCurrentCertificate(certificate.replacedBy.first().name)
+        } else {
+          throw CertificateExpired(certificate)
+        }
+    }
+  }
```
robfletcher, line 45: Find the cert by name, and return it if it's active, otherwise return the replacement if _that_ is active. If necessary recurse to look up further replacements.

robfletcher: @luispollo I revised the certificate lookup so that it throws a proper exception, and if necessary recurses to find the current certificate

keel-lemur/src/main/kotlin/com/netflix/spinnaker/keel/lemur/LemurCertificateResolver.kt
```diff
@@ -21,20 +22,28 @@ class LemurCertificateResolver(
...
-  private fun findCurrentCertificate(name: String) =
-    runBlocking {
-      val certificate = lemurService.certificateByName(name).items.first()
-      if (certificate.active) {
-        certificate.name
-      } else {
-        certificate.replacedBy.first { it.active }.name
-      }
+  private suspend fun findCurrentCertificate(name: String): String {
+    val certificate = lemurService.certificateByName(name).items.first()
```
luispollo, line 33: `lemurService.certificateByName(name).items` could also be empty here and so `.first()` would throw `NoSuchElementException`, right? I know you mentioned you checked the current certs, but I think it wouldn't hurt to add a check here.

robfletcher: Yeah, you're right. Fixed

PR #1769, opened 2021-01-29
chore(allowed-times): consistent numeric representation

emjburns:
We need a consistent representation of the allowed-times constraint (that doesn't rely on words) to power UI visualization of it. 

Before:
```
{
                  "type": "allowed-times",
                  "currentlyPassing": false,
                  "attributes": {
                    "windows": [
                      {
                        "days": "Monday-Friday",
                        "hours": "9-16"
                      }
                    ]
                  }
```

After:
```{
                  "type": "allowed-times",
                  "currentlyPassing": true,
                  "attributes": {
                    "windows": [
                      {
                        "days": [
                          1,
                          2,
                          3,
                          4,
                          5
                        ],
                        "hours": [
                          9,
                          10,
                          11,
                          12,
                          13,
                          14,
                          15,
                          16
                        ]
                      }
                    ]
                  }
                }```

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/TimeWindowConstraint.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/TimeWindowConstraint.kt
@@ -31,6 +31,15 @@ data class TimeWindowConstraint(
   }
 }
 
+/**
+ * Numeric day and hour representation of a time window.
+ * This format is consistent regardless of how the window is defined in a string.
+ */
+data class TimeWindowNumeric(
+  val days: Set<Int>,
+  val hours: Set<Int>,
+)
+
 data class TimeWindow(
   val days: String? = null,
   val hours: String? = null
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -3,7 +3,6 @@ package com.netflix.spinnaker.keel.services
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.Locatable
-import com.netflix.spinnaker.keel.api.NotificationType
 import com.netflix.spinnaker.keel.api.Resource
 import com.netflix.spinnaker.keel.api.ScmInfo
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt [hunk 1]
```diff
@@ -397,6 +397,15 @@ class ApplicationService(
...
   }
 
+  /**
+   * Translates the stored time window constraint format to numeric format for display
+   */
+  private fun translateAllowedTimesConstraint(constraint: TimeWindowConstraint) =
+    AllowedTimesConstraintMetadata(
+      AllowedTimesConstraintEvaluator.toNumericTimeWindows(constraint),
+      constraint.tz
+    )
```
luispollo, line 407: Maybe change the constructor of `AllowedTimesConstraintMetadata` to  just take the constraint as a parameter?

emjburns: Done.

gal-yardeni:
@emjburns can you elaborate more on how the user will configure it in his config? does he need to specify each day/time?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/TimeWindowConstraint.kt
```diff
@@ -31,6 +31,15 @@ data class TimeWindowConstraint(
...
 }
 
+/**
+ * Numeric day and hour representation of a time window.
+ * This format is consistent regardless of how the window is defined in a string.
+ */
+data class TimeWindowNumeric(
+  val days: Set<Int>,
+  val hours: Set<Int>,
+)
```
robfletcher, line 41: Does this representation allow me to have different hours on different days? Is that even a thing we want to support?

emjburns: Yep! It does. You just have several different windows.

However, this is _not_ user config. This is just what we translate the config into for the ui

emjburns:
@gal-yardeni @robfletcher this is *not* changing the user config. This is only changing the API response from the `/managed/application/APP?entities=artifacts` endpoint (the one that powers the environment view) so that we can get a UI for the allowed-times constraint.

robfletcher (approved):
That translation code terrifies me because it's date/time related but I think the format you're aiming for makes sense

PR #1771, opened 2021-01-29
chore(config): fast property-ing min age duration

emjburns:
Fast property-ing the minimum age for checking things. This metric controls how often we check the resource, artifact, or environment. I'm adding a fast property so that we can experiment with a smaller duration between checks (60s -> 30s) and have a way to quickly roll back changes.

Note: we use the `resource` specific config for all entities we check. I'm not changing that in this PR. I can change it in the future if we decide we need to tune each entity separately.


robfletcher (approved):
I definitely think we should separate those 3 settings but I support you not doing it in this PR

lorin (approved):

PR #1772, opened 2021-01-29
feat(slack): adding artifact deployment notifications

gal-yardeni:
In this PR 2 new notifications are added:
1. artifact version deployed successfully --> sent when we mark artifact version as successfully deployed
2. artifact version failed to deploy --> sent when we automatically mark as bad an artifact version (when we try to deploy twice with no success).
Screen Shot 2021-01-29 at 1 40 38 PM
Screen Shot 2021-01-29 at 1 40 47 PM

We used to rely on orca tasks to determine if a deployment is succeeded or failed, but we have all the data we need in keel's DB with all the artifact related data, and we can remove that orca dependency.

Future tasks/PRs:
- [ ] Remove the orca dependency (we can separate those as the new slack feature is currently disabled)
- [ ] Decide if/how we want to handle email notifications
- [ ] Add interactive notifications


emjburns (approved):
Seems reasonable to me!

keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/model/OrcaNotification.kt
```diff
@@ -60,6 +60,7 @@ data class NotificationMessage(
 const val RAINBOW = "\uD83C\uDF08"
 const val THUNDERCLOUD = "\u26c8\ufe0f"
 
+//TODO [gyardeni]: figure out if we want to keep it around, or deprecate it
```
luispollo, line 63: I think you need to remove the actual call to this function here to avoid duplicate notifications from Orca and Keel:
https://github.com/spinnaker/keel/blob/9db12ff2949db3c38501747e6cae3de3ec11fc6f/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/OrcaTaskLauncher.kt#L97

gal-yardeni: I'm going to address it in my next PR.

PR #1773, opened 2021-02-01
Try to speed up integration tests

robfletcher:
This PR

1. makes testcontainers re-use the same container across all tests

For this to work you **must have** a `~/.testcontainers.properties` file containing `testcontainers.reuse.enable=true`. This is not something you can switch on inside the project unfortunately.

### DIFF
--- a/keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/TestContainer.kt
+++ b/keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/TestContainer.kt
@@ -5,16 +5,17 @@ import org.jooq.SQLDialect.MYSQL
 import org.testcontainers.containers.JdbcDatabaseContainer
 import org.testcontainers.containers.MySQLContainerProvider
 
-internal fun initTestDatabase() = initDatabase(
-  mySQLContainer.authenticatedJdbcUrl,
-  MYSQL
-)
+internal val testDatabase by lazy {
+  initDatabase(mySQLContainer.authenticatedJdbcUrl, MYSQL)
+}
 
-private val mySQLContainer = MySQLContainerProvider()
+internal val mySQLContainer = MySQLContainerProvider()
   .newInstance("5.7.22")
   .withDatabaseName("keel")
   .withUsername("keel_service")
   .withPassword("whatever")
+  .withReuse(true)
+  .withNetwork(null)
   .also { it.start() }
 
 @Suppress("UsePropertyAccessSyntax")
--- a/keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/TestContainer.kt
+++ b/keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/TestContainer.kt
@@ -6,7 +6,7 @@ import org.testcontainers.containers.JdbcDatabaseContainer
 import org.testcontainers.containers.MySQLContainerProvider
 
 internal val testDatabase by lazy {
-  initDatabase(mySQLContainer.authenticatedJdbcUrl, MYSQL)
+  initDatabase(mySQLContainer.authenticatedJdbcUrl, MYSQL, "keel")
 }
 
 internal val mySQLContainer = MySQLContainerProvider()
@@ -15,7 +15,6 @@ internal val mySQLContainer = MySQLContainerProvider()
   .withUsername("keel_service")
   .withPassword("whatever")
   .withReuse(true)
-  .withNetwork(null)
   .also { it.start() }
 
 @Suppress("UsePropertyAccessSyntax")
--- a/keel-sql/src/test/resources/application.properties
+++ b/keel-sql/src/test/resources/application.properties
@@ -11,6 +11,6 @@ spinnaker.baseUrl=http://spinnaker
 services.fiat.baseUrl=https://fiat.net
 spring.application.name=keel
 sql.enabled=true
-sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename
-sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename
+sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/keel&TC_REUSABLE=true
+sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/keel&TC_REUSABLE=true
 spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver
--- a/keel-sql/src/test/resources/application.properties
+++ b/keel-sql/src/test/resources/application.properties
@@ -11,6 +11,6 @@ spinnaker.baseUrl=http://spinnaker
 services.fiat.baseUrl=https://fiat.net
 spring.application.name=keel
 sql.enabled=true
-sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/keel&TC_REUSABLE=true
-sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/keel&TC_REUSABLE=true
+sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22:///keel&TC_REUSABLE=true
+sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22:///keel&TC_REUSABLE=true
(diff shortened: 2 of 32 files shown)


emjburns (approved):

keel-sql/src/test/resources/application.properties
```diff
@@ -11,6 +11,6 @@ spinnaker.baseUrl=http://spinnaker
 services.fiat.baseUrl=https://fiat.net
 spring.application.name=keel
 sql.enabled=true
-sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename
-sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename
+sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/keel&TC_REUSABLE=true
+sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/keel&TC_REUSABLE=true
```
robfletcher, line 15: That `TC_REUSABLE` flag is how you tell testcontainers to re-use the container via the connection string

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/TestContainer.kt
```diff
@@ -5,16 +5,17 @@ import org.jooq.SQLDialect.MYSQL
...
+}
 
-private val mySQLContainer = MySQLContainerProvider()
+internal val mySQLContainer = MySQLContainerProvider()
   .newInstance("5.7.22")
   .withDatabaseName("keel")
   .withUsername("keel_service")
   .withPassword("whatever")
+  .withReuse(true)
+  .withNetwork(null)
```
robfletcher, line 18: the `withReuse(true)` does the same as `TC_REUSABLE=true` for programatically created containers (e.g. most of the integration tests in `keel-sql`). I'm still trying to figure out how I can get both Spring and non-Spring tests to share the _same_ container. But at least this way we only start up 2.

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/TestContainer.kt
```diff
@@ -5,16 +5,17 @@ import org.jooq.SQLDialect.MYSQL
 import org.testcontainers.containers.JdbcDatabaseContainer
 import org.testcontainers.containers.MySQLContainerProvider
 
-internal fun initTestDatabase() = initDatabase(
-  mySQLContainer.authenticatedJdbcUrl,
-  MYSQL
-)
+internal val testDatabase by lazy {
+  initDatabase(mySQLContainer.authenticatedJdbcUrl, MYSQL)
+}
```
robfletcher, line 10: This gives us a shared `testDatabase` across all the tests that gets lazy initialized

luispollo (approved):
Not sure why the tests are failing, but looks great to me! Excellent cleanup/optimization! ❤️

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/TestContainer.kt
bsideup: FYI I would recommend using `new MySQLContainer("mysql:5.7.22")` instead of `MySQLContainerProvider().newInstance("5.7.22")`, as `MySQLContainerProvider` is just for the JDBC infrastructure.

Also, unless the network is explicitly set earlier, `.withNetwork(null)` is not needed (even in Kafka starting with Testcontainers 1.15.x)

bsideup: >  I'm still trying to figure out how I can get both Spring and non-Spring tests to share the same container

Some implementation detail:
The reusable mode is using hashing. If both instances have the same hash, the container is reused, meaning that setting the same database name / username / password and other settings would result in the same hash.
Double check that _all_ parameters are the same.

bsideup: Also, since `mySQLContainer` variable is private, I got a feeling that you may as well use the JDBC-style integration and pass static `jdbc:tc:mysql:5.7.22:///keel&TC_REUSABLE=true` to `initDatabase`

keel-sql/src/test/resources/application.properties
bsideup, line 15: tiny cleanup (as per https://www.testcontainers.org/modules/databases/jdbc/):
```suggestion
sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22:///keel&TC_REUSABLE=true
sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22:///keel&TC_REUSABLE=true
```

Since host & port are assigned dynamically, TC's JDBC proxy driver ignores passed values, and JDBC URL allows providing empty values, hence `:///` (`://` + emptyness + `/`)

keel-web/src/test/resources/application.properties
```diff
@@ -10,8 +10,8 @@ igor.baseUrl=https://localhost:8083
 echo.baseUrl=https://localhost:8086
 services.fiat.baseUrl=https://fiat.net
 sql.enabled=true
-sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename
-sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/databasename
+sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/keel&TC_REUSABLE=true
+sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22://somehostname:someport/keel&TC_REUSABLE=true
```
bsideup, line 14: ditto

```suggestion
sql.connection-pools.default.jdbc-url=jdbc:tc:mysql:5.7.22:///keel&TC_REUSABLE=true
sql.migration.jdbc-url=jdbc:tc:mysql:5.7.22:///keel&TC_REUSABLE=true
```

bsideup:
> This is not something you can switch on inside the project unfortunately.

Since reusable mode (by definition) leaves the containers running, we decided that it should be an opt-in-able feature that is disabled by default **and** something that should be enabled _per environment_. This is to ensure that the mode is not enabled by accident on CIs, for example.

That said, we're currently evaluating options of how to make it more user friendly.
Also, rumours say that there are projects like https://github.com/stefanbirkner/system-rules that allow you to programmatically set environment variables... Maybe it even works with `TESTCONTAINERS_REUSE_ENABLE` 😉

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/TestContainer.kt
robfletcher: Yeah, I did try doing that and still ended up with multiple containers running (checking via `docker ps -a` afterward). I figured something was throwing the hash off but I wasn't able to pin it down.

robfletcher: > FYI I would recommend using `new MySQLContainer("mysql:5.7.22")` instead of `MySQLContainerProvider().newInstance("5.7.22")`, as `MySQLContainerProvider` is just for the JDBC infrastructure.

It doesn't seem like I can call this constructor from Kotlin due to the recursive generic type. 
> 
> Also, unless the network is explicitly set earlier, `.withNetwork(null)` is not needed (even in Kafka starting with Testcontainers 1.15.x)

That was just a desperate attempt to see if that was the issue preventing the hashes from matching.

bsideup: > It doesn't seem like I can call this constructor from Kotlin due to the recursive generic type.

Yes, this is a known limitation of Kotlin, but you can use `<Nothing>` + `apply {}`:
https://github.com/testcontainers/testcontainers-java/blob/97493b112f76711adc8f431f6f8e6f3245f05c47/examples/spring-boot-kotlin-redis/src/test/kotlin/com/example/redis/AbstractIntegrationTest.kt#L21

> Yeah, I did try doing that and still ended up with multiple containers running (checking via docker ps -a afterward). I figured something was throwing the hash off but I wasn't able to pin it down.

With MySQL? Wow. I would highly appreciate something runnable so that I can debug & help figuring out why the hash wasn't the same! (Although I understand that you got it working and most probably already switched so something else 😅  I will try to reproduce it myself, but any help would be appreciated!)

PR #1774, opened 2021-02-01
fix(export): handle nulls

emjburns:
a user hit an NPE here, so making it more safe

### DIFF
--- a/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/ClusterExportHelper.kt
+++ b/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/ClusterExportHelper.kt
@@ -129,8 +129,8 @@ class ClusterExportHelper(
           it as Map<String, Any>
         }
         ?.get("onFailure") as Boolean?,
-      resizePreviousToZero = this["scaleDown"] as Boolean,
-      maxServerGroups = this["maxRemainingAsgs"].toString().toInt(),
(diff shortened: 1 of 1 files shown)


keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/ClusterExportHelper.kt
```diff
@@ -129,8 +129,10 @@ class ClusterExportHelper(
           it as Map<String, Any>
         }
         ?.get("onFailure") as Boolean?,
-      resizePreviousToZero = this["scaleDown"] as Boolean,
-      maxServerGroups = this["maxRemainingAsgs"].toString().toInt(),
+      resizePreviousToZero = this["scaleDown"]?.let {
+        it as Boolean?
```
luispollo, line 133: Could this suffer from the same issue we had with `Int` below, which led to using `toString().toInt()` instead of `as Int`?

Also, I think you could simplify this as: 
```kotlin
resizePreviousToZero = this["scaleDown"] as? Boolean
```

emjburns: I am not sure - the issue here was it was null, so I think the `as?` might do the job!

robfletcher (approved):

luispollo (approved):

PR #1776, opened 2021-02-02
chore(release): Setup weekly releases for the benefit of OSS users

luispollo:
Adds a new GitHub Action triggered on a cron schedule to create a weekly release, so that OSS users can pull in the latest changes as a dependency during installation.

I couldn't think of a way to test this locally. If you have any ideas, please let me know.

Cc: @nimakaviani

### DIFF
--- a/.github/workflows/periodic_release.yml
+++ b/.github/workflows/periodic_release.yml
@@ -1,0 +1,60 @@
+name: Weekly Release
+
+on:
+  schedule:
+    # Every Tuesday at 6 PM UTC (10 AM PST)
+    - cron: "0 18 * * 2"
+
+env:
+  GRADLE_OPTS: -Dorg.gradle.daemon=false -Xmx2g -Xms2g
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: git fetch --prune --unshallow
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - uses: actions/cache@v1
+        with:
+          path: ~/.gradle
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle.kts') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Write current version
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/periodic_release.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          path: ~/.gradle
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle.kts') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Write current version
+        run: |
+          export CURRENT_TAG=`curl --silent "https://api.github.com/repos/$1/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'`
+          echo "version: $CURRENT_TAG" > package.json
```
luispollo, line 30: That's a hack, but the best I could come up with to make this work without this off-the-shelf action, which requires the current version to be written to a file.

.github/workflows/periodic_release.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Prepare changelog
+        id: changelog
+        uses: TriPSs/conventional-changelog-action@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          output-file: "false"
+          skip-version-file: "true"
```
luispollo, line 38: This skips updating the version file.

.github/workflows/periodic_release.yml
```diff
@@ -0,0 +1,59 @@
...
+          skip-commit: "true"
```
luispollo, line 39: Skips making a commit after bumping the version.

luispollo:
P.S. I could change the day of the month to `2` so we can see if this works tomorrow...

nimakaviani:
This is great, thanks @luispollo ... yeah I dont know of any good way to test github actions other than maybe on a separate git repo that can be tossed away. You might be able to try it in the test repo on an every minute schedule and once happy, configure the cron schedule accordingly.

luispollo:
> This is great, thanks @luispollo ... yeah I dont know of any good way to test github actions other than maybe on a separate git repo that can be tossed away. You might be able to try it in the test repo on an every minute schedule and once happy, configure the cron schedule accordingly.

Yeah, I thought about turning up the frequency but was afraid I was gonna annoy the heck out of everybody with notifications... 😅

PR #1777, opened 2021-02-02
feat(api): work towards showing verifications

lorin:
# Problem

Provide the UI with information about the state of verifications for artifact versions in each environment. (For example, what verifications have been run against keel-0.850.0-h1020.a525f99 in the prestaging environment)?

# Proposed solution

In the `/application/{application}` endpoint (called by deck in the environments view), add additional fields to artifact version objects which encode verification summary notification.

In *jq* notation, these verification summaries are found in: `artifacts[].versions[].environments[].verifications`.

## Example output

For the `fnord` app, the API call would be: 

http://gate/managed/application/fnord?entities=resources&entities=artifacts&entities=environments

The  output would looke like this:

```json
{
  "artifacts": [
    {
      "name": "fnord",
      "type": "deb",
      "reference": "fnord",
      "versions": [
        {
          "version": "fnord-0.850.0-h1021.b527f39",
          "environments": [
            {
              "name": "staging",
              "verifications": [
                {
                  "id": "test-container:acme/testcon:stable",
                  "type": "test-container",
                  "status": "PASSED",
                  "startedAt": "2021-01-04T22:31:12.142Z",
                  "completedAt": "2021-01-04T22:52:07.358Z",
                  "link": "http://www.example.com/test-output/abc123"
                }
              ]
            }
          ]
```

# Work in this PR

This PR contains plumbing work to implement the above functionality. It does not yet populate the `verifications` list above, but it creates the necessary classes to do so.

In particular:

ArtifactSummary.kt - the new `ArtifactVersionEnvironmentVerificationSummary` data class maps to the payload in the `verifications` field shown above. 

Verification.kt - Verification interface now has a `link` function for generating a URL where users can learn more about the current state of a verification. This hasn't been implemented yet for test-container verifications (see future work).

The other changes are just plumbing and updating of tests to accomodate the new verifications-related arguments.

**Note: there is no (intentional) behavior change introduced by this PR.**

# Fast-follow work

An upcoming PR will modify ApplicationService.kt to populate the list of `verifications` objects. I wanted to break it up into two PRs to make them easier to review.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Verification.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Verification.kt
@@ -18,5 +18,5 @@ interface Verification {
   /**
    * Generate a URL that a user can be directed to in order to view the current state of a verification
    */
-  fun link(state: VerificationState) : String? = null
+  fun getLink(state: VerificationState) : String? = null
 }
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Verification.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Verification.kt
@@ -1,6 +1,7 @@
 package com.netflix.spinnaker.keel.api
 
 import com.netflix.spinnaker.keel.api.schema.Discriminator
+import com.netflix.spinnaker.keel.api.verification.VerificationState
 
 interface Verification {
   @Discriminator
@@ -13,4 +14,9 @@ interface Verification {
    * in the database.
    */
   val id: String
+
+  /**
+   * Generate a URL that a user can be directed to in order to view the current state of a verification
+   */
+  fun link(state: VerificationState) : String? = null
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
@@ -51,7 +51,17 @@ data class ArtifactSummaryInEnvironment(
   val vetoed: ActionMetadata? = null,
   val statefulConstraints: List<StatefulConstraintSummary> = emptyList(),
   val statelessConstraints: List<StatelessConstraintSummary> = emptyList(),
-  val compareLink: String? = null
+  val compareLink: String? = null,
+  val verifications : List<ArtifactVersionEnvironmentVerificationSummary> = emptyList()
+)
+
+data class ArtifactVersionEnvironmentVerificationSummary(
+  val id: String,
+  val type: String,
+  val status: String,
+  val startedAt: Instant? = null,
+  val completedAt: Instant? = null,
+  val link: String? = null
 )
 
 data class ActionMetadata(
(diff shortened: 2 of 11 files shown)


lorin:
The next PR looks like this (pending some cleanup after I go over it again today): https://github.com/lorin/keel/pull/1

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Verification.kt
```diff
@@ -13,4 +14,9 @@ interface Verification {
    * in the database.
    */
   val id: String
+
+  /**
+   * Generate a URL that a user can be directed to in order to view the current state of a verification
+   */
+  fun link(state: VerificationState) : String? = null
```
luispollo, line 21: nit: should this be a `val` instead, since you're not naming it with a verb (`getLink`)?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
```diff
@@ -51,7 +51,17 @@ data class ArtifactSummaryInEnvironment(
   val vetoed: ActionMetadata? = null,
   val statefulConstraints: List<StatefulConstraintSummary> = emptyList(),
   val statelessConstraints: List<StatelessConstraintSummary> = emptyList(),
-  val compareLink: String? = null
+  val compareLink: String? = null,
+  val verifications : List<ArtifactVersionEnvironmentVerificationSummary> = emptyList()
+)
+
+data class ArtifactVersionEnvironmentVerificationSummary(
```
luispollo, line 58: _Man_, that's a long name, haha! 😄

I've struggled to come up with better names myself in a number of similar cases in keel. I think the problem is that we want to convey all the context of the wrapper objects in these types of classes. In this particular case though, I feel like, since a `Verification`, by definition, only exists for a specific artifact version in a specific environment, maybe we could ditch all that context and just call it `VerificationSummary`?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -1210,7 +1211,8 @@ class SqlArtifactRepository(
     deliveryConfig: DeliveryConfig,
     environmentName: String,
     artifactReference: String,
-    version: String
+    version: String,
+    verifications: List<ArtifactVersionEnvironmentVerificationSummary>
```
luispollo, line 1215: Instead of passing this as a parameter, I think it might be better to just add that via `copy()` to the return object in the caller, since this method is not doing anything with the parameter besides sticking it in there.

emjburns, line 1215: Why are you passing this in here only to just return it? where will this actually get calculated?

lorin: That's a good idea. Will do.

lorin: I'm going to change this to the copy pattern that @luispollo mentioned in [his comment](https://github.com/spinnaker/keel/pull/1777#discussion_r568965217).

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/Verification.kt
lorin: I ended up renaming this to `getLink` instead

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
lorin: Yeah, that was a crazy long name. Renamed.

PR #1778, opened 2021-02-02
fix(baking): Handle no versions gracefully in ImageHandler

luispollo:
We've been getting alerts for `NoKnownArtifactVersions` exceptions bubbling up all the way to the root exception handler from `ImageHandler`. This PR modifies the behavior to handle the exception gracefully by logging it and returning early.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
@@ -45,7 +45,12 @@ class ImageHandler(
           ArtifactCheckSkipped(artifact.type, artifact.name, "ActuationInProgress")
         )
       } else {
-        val latestArtifactVersion = artifact.findLatestArtifactVersion()
+        val latestArtifactVersion = try {
+          artifact.findLatestArtifactVersion()
+        } catch (e: NoKnownArtifactVersions) {
+          log.debug("Cannot resolve AMI for artifact $artifact: $e")
+          return
+        }
         val latestBaseAmiVersion = artifact.findLatestBaseAmiVersion()
 
         val desired = Image(latestBaseAmiVersion, latestArtifactVersion, artifact.vmOptions.regions)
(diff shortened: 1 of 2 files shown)


gal-yardeni (approved):

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -45,7 +45,12 @@ class ImageHandler(
           ArtifactCheckSkipped(artifact.type, artifact.name, "ActuationInProgress")
         )
       } else {
-        val latestArtifactVersion = artifact.findLatestArtifactVersion()
+        val latestArtifactVersion = try {
+          artifact.findLatestArtifactVersion()
+        } catch (e: NoKnownArtifactVersions) {
+          log.debug(e.message)
```
robfletcher, line 51: Is this really a debug situation? It seems like it shouldn't happen.

lorin, line 51: Would be good to increment a metric here as well.

lorin (approved):

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
luispollo: Why? When the artifact is first registered, you may not have any versions. That's what I'm seeing with new apps.

luispollo:
> > We've been getting alerts for `NoKnownArtifactVersions` exceptions bubbling up all the way to the root exception handler from `ImageHandler`
> 
> Why is that bad?

Because there's no action we can take on that alert, since it's a valid scenario. I'm seeing this with every new app I create that uses Debian. In my mind, this is equivalent to the exception we throw for Docker, which we're handling as a `ResourceCurrentlyUnresolvable` event, instead of letting the exception bubble up all the way as this was doing. Unfortunately for EC2 clusters there's no way to do that because this happens in the artifact check loop, not the resource check loop.

Does that make sense?

robfletcher (approved):

PR #1780, opened 2021-02-02
fix(events): Fix race condition persisting resource events

luispollo:
**Problem**
We're consistently seeing resources being created when upserting a delivery config and the matching `ResourceCreated` events failing to persist. There's a race condition between those two flows because the first runs a transaction to persist all the records associated with the delivery config, while the second is triggered asynchronously and attempts to read the UID of the resource to create the event record.

**Proposed solution**
This PR replaces the asynchronous events published in `CombinedRepository.upsertResource` (which is only used in this particular code path) with repository calls to persist the event records as part of the transaction in `CombinedRepository.upsertDeliveryConfig`.

**Other alternatives considered**
I've considered:
* Doubling the number of read retries in that second flow. This is pretty flaky.
* Refactoring the call to `upsertResource` to return any events that need publishing, and publishing the events as the very last thing in `upsertDeliveryConfig`, which would decrease the time window of the race condition, but it seemed ugly.
* Same as above, but bubbling up the events one call higher in the stack to publish the events outside of `upsertDeliveryConfig`. This was even uglier.

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
@@ -27,11 +27,7 @@ import com.netflix.spinnaker.keel.sql.RetryCategory.READ
 import com.netflix.spinnaker.keel.sql.RetryCategory.WRITE
 import de.huxhorn.sulky.ulid.ULID
 import org.jooq.DSLContext
-import org.jooq.impl.DSL
 import org.slf4j.LoggerFactory
-import org.springframework.transaction.annotation.Propagation.NESTED
-import org.springframework.transaction.annotation.Propagation.SUPPORTS
-import org.springframework.transaction.annotation.Transactional
 import java.time.Clock
 import java.time.Duration
 import java.time.Instant
@@ -247,17 +243,17 @@ open class SqlResourceRepository(
       }
 
       if (event.javaClass == previousEvent?.javaClass) return
+    }
 
-      sqlRetry.withRetry(WRITE) {
-        jooq
-          .insertInto(EVENT)
-          .set(EVENT.UID, ULID().nextULID(event.timestamp.toEpochMilli()))
-          .set(EVENT.SCOPE, event.scope)
-          .set(EVENT.REF, ref)
-          .set(EVENT.TIMESTAMP, event.timestamp)
-          .set(EVENT.JSON, event)
-          .execute()
-      }
+    sqlRetry.withRetry(WRITE) {
+      jooq
+        .insertInto(EVENT)
+        .set(EVENT.UID, ULID().nextULID(event.timestamp.toEpochMilli()))
+        .set(EVENT.SCOPE, event.scope)
+        .set(EVENT.REF, ref)
+        .set(EVENT.TIMESTAMP, event.timestamp)
+        .set(EVENT.JSON, event)
+        .execute()
     }
   }
 
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
@@ -29,6 +29,9 @@ import de.huxhorn.sulky.ulid.ULID
 import org.jooq.DSLContext
 import org.jooq.impl.DSL
 import org.slf4j.LoggerFactory
+import org.springframework.transaction.annotation.Propagation.NESTED
+import org.springframework.transaction.annotation.Propagation.SUPPORTS
+import org.springframework.transaction.annotation.Transactional
 import java.time.Clock
 import java.time.Duration
 import java.time.Instant
@@ -222,11 +225,10 @@ open class SqlResourceRepository(
 
   private fun doAppendHistory(event: PersistentEvent, ref: String) {
     log.debug("Appending event: $event")
-    jooq.transaction { config ->
-      val txn = DSL.using(config)
 
-      if (event.ignoreRepeatedInHistory) {
-        val previousEvent = txn
+    if (event.ignoreRepeatedInHistory) {
+      val previousEvent = sqlRetry.withRetry(READ) {
+        jooq
           .select(EVENT.JSON)
           .from(EVENT)
           // look for resource events that match the resource...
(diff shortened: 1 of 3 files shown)


emjburns:
At first glance I like the last approach you listed here: publishing the events after the method has successfully saved the config. To me, that seems like a more understandable and easy to reason about approach rather than increasing the number of retries. Why do you think it seems ugly?

robfletcher:
Couldn't you just make the triggering of the event happen in response to the transaction completing? That seems appropriate anyway.

luispollo:
> Couldn't you just make the triggering of the event happen in response to the transaction completing? That seems appropriate anyway.

You mean as in after the method marked `@Transactional` returns, or is there a way to hook into the transaction handling machinery? I can, but would need to bubble up all the events for all the resources as I mentioned. If you and @emjburns feel that's the right way to go, I'm happy to do that.

robfletcher:
I like it for a couple of reasons:

1. as Emily says it seems easier to reason about
2. it's more deterministic -- even with an increased retry count this could still fail, right?

lorin:
I agree that it would be cleaner if the code was structured so that the transaction completed before the ResourceCreated event was emitted.

My first instinct would be to figure out a way to queue up the resource events in such a way that they get processed after the transaction completes.

luispollo:
> I agree that it would be cleaner if the code was structured so that the transaction completed before the ResourceCreated event was emitted.
> 
> My first instinct would be to figure out a way to queue up the resource events in such a way that they get processed after the transaction completes.

Yup. Will update the PR with that approach.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -222,11 +225,10 @@ open class SqlResourceRepository(
 
   private fun doAppendHistory(event: PersistentEvent, ref: String) {
     log.debug("Appending event: $event")
-    jooq.transaction { config ->
```
luispollo, line 225: The new transaction here was unnecessary, and was breaking the expected transactional behavior of `CombinedRepository.upsertDeliveryConfig`.

gradle.properties
```diff
@@ -6,7 +6,9 @@ org.gradle.parallel=true
 spinnakerGradleVersion=8.10.1
 buildingInDocker=false
 targetJava11=true
-testContainersVersion=1.15.0
+testContainersVersion=1.15.1
```
luispollo, line 9: Fix for: 
```
   com.github.dockerjava.api.exception.NotFoundException: Status 404: {"message":"No such image: testcontainers/ryuk:0.3.0"}
```

keel-sql/src/test/resources/testcontainers.properties
```diff
@@ -0,0 +1 @@
+ryuk.container.image = public.ecr.aws/s4w6t4b6/testcontainers/ryuk:0.3.0
```
luispollo, line 1: Preemptive fix for Docker Hub rate limiting.

lorin (approved):

lorin (commented):
(accidentally approved, changing this back to "comment" status while I do a proper review)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -29,6 +29,9 @@ import de.huxhorn.sulky.ulid.ULID
 import org.jooq.DSLContext
 import org.jooq.impl.DSL
 import org.slf4j.LoggerFactory
+import org.springframework.transaction.annotation.Propagation.NESTED
```
lorin, line 32: It looks like these new imports aren't used.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -242,18 +244,20 @@ open class SqlResourceRepository(
...
-        .set(EVENT.SCOPE, event.scope)
-        .set(EVENT.REF, ref)
-        .set(EVENT.TIMESTAMP, event.timestamp)
-        .set(EVENT.JSON, event)
-        .execute()
+      if (event.javaClass == previousEvent?.javaClass) return
+
+      sqlRetry.withRetry(WRITE) {
+        jooq
+          .insertInto(EVENT)
```
lorin, line 253: It looks like the behavior of this block has changed if `event.ignoreRepeatedInHistory` is false. 

Old code: when condition is false, code jumps to line 249, inserts the event

New code: when condition is false, code jumps to line 261, does not insert the event.

lorin (changes requested):
It looks to me like there was an accidental behavioral change introduced here. Can you take a closer look to confirm?

lorin:
Overall, though, this is an elegant solution to this problem.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
lorin: I went back and forth on this. It looks like executing those two queries outside of a transaction should be harmless, but it's always difficult to reason about such things.

luispollo: Yeah, I discussed this a bit with @robfletcher and it seemed like an OK approach (we do reads with follow-up writes outside transactions in other places).

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: Ahhh! You're right! What was I thinking! Good spot! 🙏

luispollo: Actually, the qualified was returning to the end of the scope of `jooq.transaction`, which is the last curly brace in the function body before the end of the function, so it was essentially returning from the function altogether if the condition met. Stop scaring me, @lorin! 🤣

luispollo: P.S. But my curly braces were messed up, so thanks! 😄

lorin (approved):
Looks good now with the braces fix.

PR #1787, opened 2021-02-04
merge override inbound rules instead of replacing

robfletcher:
Merges `inboundRules` from region overrides with the common `inboundRules` rather than replacing them.

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
@@ -73,7 +73,7 @@ open class SecurityGroupHandler(
             region = region.name
           ),
           description = overrides[region.name]?.description ?: description,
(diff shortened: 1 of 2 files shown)


lorin (commented):
The implementation looks good. I had two questions.

1. Can you provide some more context into the scenario that motivated this change? 

I ask because I've run into the converse scenario:  a user expected overrides to replace but they merged.   (That other scenario also involved security groups in some way).

2. When this change is deployed, will it trigger diffs/actuations for apps that have inbound rule overrides?

I assume this is the case, that we would get some actuations triggered by this deploy if anyone happens to have overrides on inbound rules today.

robfletcher:
The situation was a case where there were many common ingress rules and an additional one in a single region. For non-overlapping ingress rules you can specify everything as overrides, but I think that's a less common scenario than having most of the rules be common and a small number of additional rules in individual regions.

luispollo (approved):
LGTM!

lorin (approved):
Makes sense. 

In the future, I wonder if we should support more explicit override commands (e.g., `override_replace`, `override_merge`). It gives me the willies that the same field, `override`, has different semantics depending on the resource.

PR #1788, opened 2021-02-04
chore(tests): disable flaky cert cache tests

lorin:
# Problem

Some tests are failing intermittently when run in GitHub Actions, which blocks PR merges.

Attempts to diagnose the issue so far have been unsuccessful.

# Proposed solution

Mark these tests as disabled for now so that they don't block development.


robfletcher (approved):
This makes me sad but I guess. The methods called twice are both blocking so the only way I can imagine it's getting into a race condition is if the cache writes behind (i.e. returns the value before / in parallel with putting that value in the cache). I'm going to quickly try to see if there's a way to control that from the test, but in the mean time this will get us unblocked.

PR #1791, opened 2021-02-05
feat(ec2): tag image after verification

emjburns:
This pr supports an experiment: a hook point that allows managed delivery to hand off to pipelines. 

Right before a verification starts the verification runner finds all the images that are currently in an environment. This is done by calling the cluster handlers (we ignore other resources) and getting the image that's running in that cluster (using the `current` function). We save that in the verification metadata.

If the verification passes, we dig into that metadata and issue a "tag image" job for each ami.
Screen Shot 2021-02-04 at 3 59 45 PM

This is meant to be a quick and dirty experiment to allow our partners (danny) to test out some new templates. If it turns out that handing off between md and pipelines is useful, I might implement v2 of this in a different way.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
@@ -125,6 +125,7 @@ data class VerificationCompleted(
   val artifactType: ArtifactType,
   val artifactVersion: String,
   val verificationType: String,
+  val verificationId: String,
   val status: ConstraintStatus,
   val metadata: Map<String,Any?>
 ) : TelemetryEvent() {
@@ -141,6 +142,7 @@ data class VerificationCompleted(
     context.artifact.type,
     context.version,
     verification.type,
+    verification.id,
     status,
     metadata
   )
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
@@ -121,25 +121,28 @@ data class VerificationCompleted(
   val application: String,
   val deliveryConfigName: String,
   val environmentName: String,
-  val artifactName: String,
+  val artifactReference: String,
   val artifactType: ArtifactType,
   val artifactVersion: String,
   val verificationType: String,
-  val status: ConstraintStatus
+  val status: ConstraintStatus,
+  val metadata: Map<String,Any?>
 ) : TelemetryEvent() {
   constructor(
     context: VerificationContext,
     verification: Verification,
-    status: ConstraintStatus
+    status: ConstraintStatus,
+    metadata: Map<String, Any?>
   ) : this(
     context.deliveryConfig.application,
     context.deliveryConfig.name,
     context.environmentName,
-    context.artifact.name,
+    context.artifact.reference,
     context.artifact.type,
     context.version,
     verification.type,
-    status
+    status,
+    metadata
   )
 }
 
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
@@ -4,6 +4,7 @@ import com.netflix.spinnaker.keel.api.Verification
 import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus
 import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus.NOT_EVALUATED
 import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus.PENDING
+import com.netflix.spinnaker.keel.api.plugins.CurrentImages
 import com.netflix.spinnaker.keel.api.plugins.VerificationEvaluator
 import com.netflix.spinnaker.keel.api.verification.VerificationContext
 import com.netflix.spinnaker.keel.api.verification.VerificationRepository
@@ -17,7 +18,8 @@ import org.springframework.stereotype.Component
 class VerificationRunner(
   private val verificationRepository: VerificationRepository,
   private val evaluators: List<VerificationEvaluator<*>>,
-  private val eventPublisher: ApplicationEventPublisher
+  private val eventPublisher: ApplicationEventPublisher,
+  private val imageFinder: ImageFinder
 ) {
   /**
    * Evaluates the state of any currently running verifications and launches the next, against a
@@ -37,13 +39,13 @@ class VerificationRunner(
       }
 
       statuses.firstOutstanding?.let { verification ->
-        start(verification)
+        start(verification, imageFinder.getImages(context.deliveryConfig, context.environmentName))
       } ?: log.debug("Verification complete for {}", environment.name)
     }
   }
 
-  private fun VerificationContext.start(verification: Verification) {
-    val metadata = evaluators.start(this, verification)
+  private fun VerificationContext.start(verification: Verification, images: List<CurrentImages>) {
+    val metadata = evaluators.start(this, verification) + mapOf("images" to images)
     markAsRunning(verification, metadata)
     eventPublisher.publishEvent(VerificationStarted(this, verification))
   }
@@ -56,7 +58,7 @@ class VerificationRunner(
           if (newStatus.complete) {
             log.debug("Verification {} completed with status {} for {}", verification, newStatus, environment.name)
             markAs(verification, newStatus)
-            eventPublisher.publishEvent(VerificationCompleted(this, verification, newStatus))
+            eventPublisher.publishEvent(VerificationCompleted(this, verification, newStatus, state.metadata))
           }
         }
     } else {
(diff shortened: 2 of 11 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryEvent.kt
```diff
@@ -121,25 +121,28 @@ data class VerificationCompleted(
   val application: String,
   val deliveryConfigName: String,
   val environmentName: String,
-  val artifactName: String,
+  val artifactReference: String,
```
emjburns, line 124: @lorin @robfletcher we should always be using reference and not name. Nothing was using this event, so I changed it here. But a note for the future.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -37,13 +39,13 @@ class VerificationRunner(
       }
 
       statuses.firstOutstanding?.let { verification ->
-        start(verification)
```
emjburns, line 40: Right before we start the verification we grab the current images

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -37,13 +39,13 @@ class VerificationRunner(
...
+        start(verification, imageFinder.getImages(context.deliveryConfig, context.environmentName))
       } ?: log.debug("Verification complete for {}", environment.name)
     }
   }
 
-  private fun VerificationContext.start(verification: Verification) {
-    val metadata = evaluators.start(this, verification)
+  private fun VerificationContext.start(verification: Verification, images: List<CurrentImages>) {
+    val metadata = evaluators.start(this, verification) + mapOf("images" to images)
```
emjburns, line 48: save the images in the metadata

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -142,6 +144,20 @@ class ClusterHandler(
       resource.spec.resolveCapacity(serverGroup.location.region)
     ) == true
```
emjburns, line 146: You have to tag amis by name and not id, so we have to make an additional call to clouddriver to get the info about the running ami.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ImageTagger.kt
```diff
@@ -0,0 +1,90 @@
...
+  private val springEnv: Environment,
+  private val spectator: Registry
+) {
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  private val TAG_AMI_JOB_LAUNCHED = "keel.image.tag"
+
+  private val shouldTagImages: Boolean
+    get() = springEnv.getProperty("keel.image.tagging.enabled", Boolean::class.java, true)
+
```
emjburns, line 40: Here's where we generate the tag task.

emjburns: I use the default service account for the tagging since it's a background task that is not user configurable. Also, I don't want to have to read the delivery config again.

keel-ec2-plugin/src/test/kotlin/com/netflix/spinnaker/keel/ec2/ImageTaggerTests.kt
```diff
@@ -0,0 +1,132 @@
...
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.slot
+import io.mockk.verify
+import org.springframework.core.env.Environment
+import strikt.api.expect
+import strikt.assertions.isA
+import strikt.assertions.isEqualTo
+
+class ImageTaggerTests : JUnit5Minutests {
```
lorin, line 29: A+ tests

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ImageTagger.kt
```diff
@@ -0,0 +1,90 @@
...
+  @EventListener(VerificationCompleted::class)
+  fun onVerificationCompleted(event: VerificationCompleted) {
+    if (event.status.failed() || !shouldTagImages) {
+      return
+    }
+
+    val imagesString: Any = event.metadata["images"] ?: emptyList<CurrentImages>()
```
lorin, line 47: This code looked a little weird to me on first glance, because:

* there's a variable with the name `imagesString` that doesn't always contain a string
* I didn't know you could pass an empty list as an argument to `mapper.convertValue`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -56,7 +58,7 @@ class VerificationRunner(
           if (newStatus.complete) {
             log.debug("Verification {} completed with status {} for {}", verification, newStatus, environment.name)
             markAs(verification, newStatus)
-            eventPublisher.publishEvent(VerificationCompleted(this, verification, newStatus))
+            eventPublisher.publishEvent(VerificationCompleted(this, verification, newStatus, state.metadata))
```
lorin, line 61: It looks like this fires whenever any verification completes, which means that if there are multiple verifications, and it passes the first one but fails the second, it will still be tagged.

lorin (commented):
Still reviewing, but here's one thing I noticed:

It looks like the tagging happens whenever any verification succeeds in an environment. If it happens that there are multiple verifications associated with an environment, then this tagging logic will fire after the first verification passes, even if the other ones fail.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ImageTagger.kt
emjburns: Ya, I'll make this better. But I totally have a test for the empty images case and it works! lol

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
emjburns: I...uh...forgot you could have multiple verifications. Good catch.

lorin (approved):
## Context

An internal team in our org is partially migrating some deployments to Managed Delivery. They want to be able to run the test pipelines on MD and keep the production pipelines as is. 

## Problem

For applications that use MD for deploying to test, and pipelines for deploying to prod, they need some mechanism for identifying which image to deploy to prod.

## Proposed solution

Automatically add AMI tags to all images that have passed verification in an environment. This will enable a production pipeline to deploy using the latest AMI that has the tag.

For example, if verification succeeded against the `test` environemnt. 

```
latest tested: true
environment: test
```

## Feedback

Other than the issue with multiple verifications, looks good to me.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ImageTagger.kt
```diff
@@ -0,0 +1,90 @@
...
+
+  private val TAG_AMI_JOB_LAUNCHED = "keel.image.tag"
+
+  private val shouldTagImages: Boolean
+    get() = springEnv.getProperty("keel.image.tagging.enabled", Boolean::class.java, true)
+
+  @EventListener(VerificationCompleted::class)
+  fun onVerificationCompleted(event: VerificationCompleted) {
+    if (event.status.failed() || !shouldTagImages) {
+      return
```
gal-yardeni, line 44: can you add a log message here?

emjburns: I'm hesitant to because I think it'll just clutter up the logs. What case are you imagining that a debug log would be useful?

gal-yardeni: hmm good question - in which cases `shouldTagImages` will be false?

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ImageTagger.kt
```diff
@@ -0,0 +1,91 @@
...
+  fun CurrentImages.toJob(env: String, verificationId: String): Map<String, Any?> =
+    mapOf(
+      "type" to "upsertImageTags",
+      "imageNames" to images.map { it.imageName },
+      "regions" to images.map { it.region }.toSet(),
+      "tags" to mapOf(
+        "latest tested" to true,
+        env to "environment:passed",
+        verificationId to "passed"
+      )
```
emjburns, line 89: Adding tags so that environments and verification ids don't collide

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/ImageTagger.kt
emjburns: fixed the name.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
emjburns: Added a new tag with the verification id to support looking for several tags in the "find image from tags" stage.

PR #1793, opened 2021-02-09
Versioned resources

robfletcher:
The aim of this PR is to start the implementation of versioned environments by making resources versioned.

* Each resource `id` may have 1..n versions
* The combination of `id` and `version` is unique
* Each version has a unique `uid`
* The resource with the highest `version` for a given `id` is the newest
* For now we will always deal with the newest version, but in future an environment version will reference particular resource versions

To-do
- [x] Test that old resource versions are no longer checked
- [x] Change resource deletion so that it deletes _all_ versions
- [x] Re-link environment to new resource version

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
@@ -294,9 +294,11 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
           test("stores the resource") {
             val persistedResource = subject.getResource(resource.id)
             expectThat(persistedResource) {
-              isA<Resource<DummyResourceSpec>>()
-              get { id }.isEqualTo(resource.id)
-              get { (spec as DummyResourceSpec).data }.isEqualTo("o hai")
+              get(Resource<*>::id) isEqualTo resource.id
+              get(Resource<*>::version) isEqualTo 1
+              get(Resource<*>::spec)
+                .isA<DummyResourceSpec>()
+                  .get(DummyResourceSpec::data) isEqualTo "o hai"
             }
           }
 
@@ -327,16 +329,17 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
             }
 
             test("stores the updated resource") {
-              expectThat(subject.getResource(resource.id))
-                .get { spec }
+              expectThat(subject.getResource(resource.id)) {
+                get(Resource<*>::version) isEqualTo 2
+                get(Resource<*>::spec)
                 .isA<DummyResourceSpec>()
-                .get { data }
-                .isEqualTo("kthxbye")
+                .get(DummyResourceSpec::data) isEqualTo "kthxbye"
+              }
             }
 
             test("records that the resource was updated") {
               verify {
-                resourceRepository.appendHistory(ofType<ResourceCreated>())
+                resourceRepository.appendHistory(ofType<ResourceUpdated>())
               }
             }
 
@@ -352,6 +355,12 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
             before {
               resourcesDueForCheck()
               subject.upsertResource(resource, deliveryConfig.name)
+            }
+
+            test("does not update the resource version") {
+              expectThat(subject.getResource(resource.id)) {
+                get(Resource<*>::version) isEqualTo 1
+              }
             }
 
             test("does not record that the resource was updated") {
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
@@ -599,6 +599,29 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
         }
       }
 
+      context("updating an existing delivery config with a new version of a resource") {
+        before {
+          storeArtifacts()
+          storeResources()
+          store()
+
+          val resource = deliveryConfig.environments.first().resources.first { it.kind == parseKind("ec2/cluster@v1") }
+            .run {
+              copy(spec = DummyResourceSpec(id = id, application = application))
+            }
+
+          resourceRepository.store(resource)
+        }
+
+        test("retrieving the delivery config includes the new version of the resource") {
+          getByName()
+            .isSuccess()
+            .get { resources.map(Resource<*>::version) }
+            .hasSize(deliveryConfig.resources.size)
+            .contains(2)
+        }
+      }
+
       context("notifications") {
         before {
           repository.store(deliveryConfig.copy(
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
@@ -17,6 +17,7 @@ package com.netflix.spinnaker.keel.persistence
 
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Resource
+import com.netflix.spinnaker.keel.api.ResourceSpec
 import com.netflix.spinnaker.keel.events.ApplicationActuationPaused
 import com.netflix.spinnaker.keel.events.ApplicationActuationResumed
 import com.netflix.spinnaker.keel.events.ResourceActuationLaunched
@@ -181,6 +182,20 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
         test("the resource version is incremented") {
           expectThat(subject.get<DummyResourceSpec>(resource.id))
             .get(Resource<*>::version) isEqualTo 2
+        }
+
+        context("when deleted") {
+          before {
+            subject.delete(resource.id)
+          }
+
+          test("all versions of the resource are deleted") {
+            expectCatching {
+              subject.get(resource.id)
+            }
+              .isFailure()
+              .isA<NoSuchResourceException>()
+          }
         }
       }
 
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
@@ -38,10 +38,6 @@ import io.mockk.confirmVerified
 import io.mockk.mockk
 import io.mockk.verify
 import io.mockk.verifyAll
-import java.time.Clock
-import java.time.Duration
-import java.time.Period
-import java.util.UUID.randomUUID
 import strikt.api.Assertion
 import strikt.api.expectCatching
 import strikt.api.expectThat
@@ -55,6 +51,10 @@ import strikt.assertions.isFailure
 import strikt.assertions.isGreaterThanOrEqualTo
 import strikt.assertions.isNotEmpty
 import strikt.assertions.map
+import java.time.Clock
+import java.time.Duration
+import java.time.Period
+import java.util.UUID.randomUUID
 
 abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests {
 
@@ -176,6 +176,11 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
           expectThat(subject.get<DummyResourceSpec>(resource.id))
             .get(Resource<*>::spec)
             .isEqualTo(updatedResource.spec)
+        }
+
+        test("the resource version is incremented") {
+          expectThat(subject.get<DummyResourceSpec>(resource.id))
+            .get(Resource<*>::version) isEqualTo 2
         }
       }
 
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
@@ -68,13 +68,15 @@ interface KeelRepository : KeelReadOnlyRepository {
       val diff = DefaultResourceDiff(resource.spec, existingResource.spec)
       if (diff.hasChanges() || resource.kind.version != existingResource.kind.version) {
         log.debug("Updating ${resource.id}")
-        storeResource(resource)
-        appendResourceHistory(ResourceUpdated(resource, diff.toDeltaJson(), clock))
+        storeResource(resource).also { updatedResource ->
+          appendResourceHistory(ResourceUpdated(updatedResource, diff.toDeltaJson(), clock))
+        }
       }
     } else {
       log.debug("Creating $resource")
-      storeResource(resource)
-      appendResourceHistory(ResourceCreated(resource, clock))
+      storeResource(resource).also { updatedResource ->
+        appendResourceHistory(ResourceCreated(updatedResource, clock))
+      }
     }
   }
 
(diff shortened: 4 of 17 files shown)


keel-sql/src/main/resources/db/changelog/20210208-versioned-resources.yml
```diff
@@ -0,0 +1,47 @@
...
+    changes:
+    - addColumn:
+        tableName: resource
+        columns:
+        - name: version
+          type: integer
+          defaultValueNumeric: 1
+          afterColumn: id
+          constraints:
+            nullable: false
```
robfletcher, line 14: Add the version column to resource

keel-sql/src/main/resources/db/changelog/20210208-versioned-resources.yml
```diff
@@ -0,0 +1,47 @@
...
+    - dropUniqueConstraint:
+        tableName: resource
+        constraintName: name
+    - addUniqueConstraint:
+        tableName: resource
+        constraintName: resource_id_version_idx
+        columnNames: id, version
```
robfletcher, line 21: Replace the existing (weirdly named) unique index with a new compound one

keel-sql/src/main/resources/db/changelog/20210208-versioned-resources.yml
```diff
@@ -0,0 +1,47 @@
...
+              'environmentName', environment.name,
+              'deliveryConfig', delivery_config.uid,
+              'serviceAccount', delivery_config.service_account
+            ) as metadata,
+            resource.spec
+          from resource
+          left outer join environment_resource on resource.uid = environment_resource.resource_uid
+          left outer join environment on environment.uid = environment_resource.environment_uid
+          left outer join delivery_config on delivery_config.uid = environment.delivery_config_uid
+          where resource.version = (select max(r2.version) from resource r2 where resource.id = r2.id)
```
robfletcher, line 47: Replace the `resource_with_metadata` view so that it only shows the latest resource version for a given id, and the version itself appears in the metadata

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
```diff
@@ -327,16 +329,17 @@ abstract class CombinedRepositoryTests<D : DeliveryConfigRepository, R : Resourc
...
+                .get(DummyResourceSpec::data) isEqualTo "kthxbye"
+              }
             }
 
             test("records that the resource was updated") {
               verify {
-                resourceRepository.appendHistory(ofType<ResourceCreated>())
+                resourceRepository.appendHistory(ofType<ResourceUpdated>())
               }
             }
```
robfletcher, line 344: this test was wrong 😬

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/KeelRepository.kt
```diff
@@ -121,7 +123,7 @@ interface KeelRepository : KeelReadOnlyRepository {
   // START ResourceRepository methods
   fun allResources(callback: (ResourceHeader) -> Unit)
 
-  fun storeResource(resource: Resource<*>)
+  fun storeResource(resource: Resource<*>): Resource<*>
```
robfletcher, line 126: This method now returns the resource as persisted, i.e. with a `uid` and `version` in the metadata

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -121,35 +125,36 @@ open class SqlResourceRepository(
...
-    val uid = jooq.select(RESOURCE.UID)
+  override fun store(resource: Resource<*>): Resource<*> {
+    val uid = randomUID().toString()
+    val version = jooq
+      .select(
+        coalesce(
+          max(RESOURCE.VERSION).plus(1),
+          value(1)
+        )
+      )
```
robfletcher, line 136: When updating we want current version + 1, when creating for the first time we want 1

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -121,35 +125,36 @@ open class SqlResourceRepository(
...
-      .onDuplicateKeyUpdate()
-      .set(updatePairs)
+    jooq.insertInto(RESOURCE)
+      .set(RESOURCE.UID, uid)
+      .set(RESOURCE.KIND, resource.kind.toString())
+      .set(RESOURCE.ID, resource.id)
+      .set(RESOURCE.VERSION, version)
+      .set(RESOURCE.SPEC, objectMapper.writeValueAsString(resource.spec))
+      .set(RESOURCE.APPLICATION, resource.application)
       .execute()
```
robfletcher, line 148: We no longer update resources, just insert a new version

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -121,35 +125,36 @@ open class SqlResourceRepository(
...
+
     jooq.insertInto(RESOURCE_LAST_CHECKED)
       .set(RESOURCE_LAST_CHECKED.RESOURCE_UID, uid)
       .set(RESOURCE_LAST_CHECKED.AT, EPOCH.plusSeconds(1))
-      .onDuplicateKeyUpdate()
-      .set(RESOURCE_LAST_CHECKED.AT, EPOCH.plusSeconds(1))
-      .set(RESOURCE_LAST_CHECKED.IGNORE, false)
       .execute()
```
robfletcher, line 153: Reminder to self: I need to verify that we do _not_ try to check old versions of resources. So long as we read from the `resource_with_metadata` view that won't be the case but having a test in place would be a good idea.

keel-sql/src/main/resources/db/changelog/20210208-versioned-resources.yml
emjburns: So basically the change in view here is the thing that makes sure the latest version is returned?

robfletcher: Note, the only differences here are the `version` property in the metadata, and the `where` clause on the last line of the view definition

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
robfletcher: Okay I added a test the checks this

keel-sql/src/main/resources/db/changelog/20210208-versioned-resources.yml
robfletcher: @emjburns exactly, yes

emjburns (approved):

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
luispollo: Oops, my bad. How was it passing?! 😬

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: What's the purpose of `max()` in the expression?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: Should we be worried about the size of the table after this change, and whether to cap the max number of versions for a resource (in a future PR, if at all)?

keel-sql/src/main/resources/db/changelog/20210208-versioned-resources.yml
```diff
@@ -0,0 +1,47 @@
...
+        columns:
+        - name: version
+          type: integer
+          defaultValueNumeric: 1
+          afterColumn: id
+          constraints:
+            nullable: false
+    - dropUniqueConstraint:
+        tableName: resource
+        constraintName: name
```
luispollo, line 17: So I take it we didn't have a unique constraint on `id`?

luispollo: Oh, never mind, I see the comment on the name below.

keel-sql/src/main/resources/db/changelog/20210208-versioned-resources.yml
```diff
@@ -0,0 +1,47 @@
...
+              'version', resource.version,
+              'application', resource.application,
+              'environment', environment.uid,
+              'environmentName', environment.name,
+              'deliveryConfig', delivery_config.uid,
+              'serviceAccount', delivery_config.service_account
+            ) as metadata,
+            resource.spec
+          from resource
+          left outer join environment_resource on resource.uid = environment_resource.resource_uid
```
luispollo, line 44: Hmm, this implies that the relationship is between an environment and specific versions of a resource, which is not the case. While I don't like using `resource.id` as an FK, shouldn't that be what we use here instead of `resource.uid`?

luispollo (approved):
I only had a question on the join columns used in the view, otherwise looks good!

luispollo:
Sorry, I just noted the open TO-DO items in the description. Are you planning to address those in this PR (in which case I'll wait to do a final review)?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
robfletcher: Very good question. I presume because that type of event was added as well (but that was already checked in an earlier test)

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -599,6 +599,29 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
...
+        }
+
+        test("retrieving the delivery config includes the new version of the resource") {
+          getByName()
+            .isSuccess()
+            .get { resources.map(Resource<*>::version) }
+            .hasSize(deliveryConfig.resources.size)
+            .contains(2)
+        }
+      }
```
robfletcher, line 623: This tests that when we create a v n+1 of a resource, the foreign key linking it to the environment updates so that the environment now contains the new resource version

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -177,6 +178,25 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
...
+          }
+
+          test("all versions of the resource are deleted") {
+            expectCatching {
+              subject.get(resource.id)
+            }
+              .isFailure()
+              .isA<NoSuchResourceException>()
+          }
+        }
```
robfletcher, line 199: Deleting a resource should now delete _all_ versions of it

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -121,35 +126,50 @@ open class SqlResourceRepository(
...
+
+    // This is somewhat temporary because eventually we'll define a new environment version
+    if (oldUid != null) {
+      jooq
+        .update(ENVIRONMENT_RESOURCE)
+        .set(ENVIRONMENT_RESOURCE.RESOURCE_UID, uid)
+        .where(ENVIRONMENT_RESOURCE.RESOURCE_UID.eq(oldUid))
+        .execute()
+    }
```
robfletcher, line 168: If this is an update, re-link the new version of the resource to any environments it belongs to

keel-sql/src/main/resources/db/changelog/20210208-versioned-resources.yml
robfletcher: That _is_ the relationship. An environment doesn't contain all the versions of a resource, just the latest one.

Future work will define multiple versions of environments, so if a resource updates it will trigger the creation of a new environment version.

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepositoryPeriodicallyCheckedTests.kt
```diff
@@ -122,6 +115,29 @@ internal object SqlResourceRepositoryPeriodicallyCheckedTests :
...
+        expectThat(nextResults())
+          .hasSize(1)
+          .first()
+          .get(Resource<*>::version) isEqualTo 2
+
+        expectThat(nextResults())
+          .isEmpty()
+      }
+    }
+  }
```
robfletcher, line 140: Our resource check loop should only check the latest versions of resources, not older ones

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepositoryTests.kt
luispollo: Argh. Thanks for fixing it! Damn copy/🍝.

keel-sql/src/main/resources/db/changelog/20210208-versioned-resources.yml
luispollo: Ah, got it. Makes sense. I was reading this thinking of the relationships as defined in the delivery config (where there's no concept of versions).

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
luispollo: Is there a chance this might break code that was previously using `deliveryConfig.resources` or `environment.resources` to loop through resources, assuming there would be a single object for each resource ID?

luispollo: For example, when generating resource summaries for the UI:
https://github.com/spinnaker/keel/blob/37ea959427af8fa2c2e9ce2d39d2f07f3c9ac3d5/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt#L184-L186

robfletcher:
> Sorry, I just noted the open TO-DO items in the description. Are you planning to address those in this PR (in which case I'll wait to do a final review)?

Yep, done now.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
robfletcher: > (I'm assuming we now have all versions of each resource in that list)

We don't. The environment will only contain the latest version of the resource. In the code where the new version of the resource is persisted I break the existing `environment_resource` link and replace it with a new one.

luispollo (approved):
Double thumbs-up! 🙂

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
robfletcher: SQL's `max` function will return the highest value of that field from all the selected rows. So in this case the highest version number for any resource with the same `id`.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
robfletcher: Still gonna be a lot smaller than the `event` table. I'm not worried just yet. But, yes I did suggest in the proposal doc that we could prune older versions after a while.

gal-yardeni (approved):
makes sense to me!

PR #1795, opened 2021-02-09
chore(veto/pin): reflect new desired artifact more quickly

emjburns:
When someone vetos an artifact or pins an artifact (or deletes either of those) we want to make sure that we reflect this really quickly. It's a bad experience to have to wait a minute for that change to get picked up. This PR triggers a recheck of all the environments in a config every time one of those things happens. This means that the desired version for each resource gets reflected more quickly. 

My goal is to make the mark-as-bad -> rollback action happen as quickly as possible so the UI feels more responsive.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
@@ -29,6 +29,7 @@ import com.netflix.spinnaker.keel.test.resource
 import com.netflix.spinnaker.time.MutableClock
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
+import strikt.api.expect
 import strikt.api.expectCatching
 import strikt.api.expectThat
 import strikt.api.expectThrows
@@ -43,6 +44,7 @@ import strikt.assertions.isNotNull
 import strikt.assertions.isNull
 import strikt.assertions.isSuccess
 import strikt.assertions.none
+import java.time.Duration
 
 abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository, P : PausedRepository> :
   JUnit5Minutests {
@@ -189,6 +191,10 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
           .isFailure()
           .isA<NoSuchDeliveryConfigException>()
       }
+
+      test("recheck does not throw an exception") {
+        expectCatching { repository.triggerRecheck("whatever") }.isSuccess()
+      }
     }
 
     context("storing a delivery config with no artifacts or environments") {
@@ -214,6 +220,20 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
             get { application }.isEqualTo(deliveryConfig.application)
             get { metadata }.isNotEmpty()
           }
+      }
+
+      test("config can be rechecked") {
+        val items = repository.itemsDueForCheck(Duration.ofMinutes(2), 1)
+        val items2 = repository.itemsDueForCheck(Duration.ofMinutes(2), 1)
+        repository.triggerRecheck(deliveryConfig.application)
+        val items3 = repository.itemsDueForCheck(Duration.ofMinutes(2), 1)
+
+        expect {
+          that(items.size).isEqualTo(1)
+          that(items2.size).isEqualTo(0)
+          that(items3.size).isEqualTo(1)
+          that(items3.first().application).isEqualTo(deliveryConfig.application)
+        }
       }
     }
 
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
@@ -983,6 +983,29 @@ class SqlDeliveryConfigRepository(
     }
   }
 
+  override fun triggerRecheck(application: String) {
+    val uid = sqlRetry.withRetry(READ) {
+      jooq.select(DELIVERY_CONFIG.UID)
+        .from(DELIVERY_CONFIG)
+        .where(DELIVERY_CONFIG.APPLICATION.eq(application))
+        .fetchOne(DELIVERY_CONFIG.UID)
+    }
+
+    if (uid == null) {
+      log.error("Config for app $application does not exist, cannot recheck it")
+      return
+    }
+
+    sqlRetry.withRetry(WRITE) {
+      jooq.update(DELIVERY_CONFIG_LAST_CHECKED)
+        .set(DELIVERY_CONFIG_LAST_CHECKED.AT, EPOCH.plusSeconds(1))
+        .set(DELIVERY_CONFIG_LAST_CHECKED.LEASED_AT, EPOCH.plusSeconds(1))
+        .setNull(DELIVERY_CONFIG_LAST_CHECKED.LEASED_BY)
+        .where(DELIVERY_CONFIG_LAST_CHECKED.DELIVERY_CONFIG_UID.eq(uid))
+        .execute()
+    }
+  }
+
   override fun itemsDueForCheck(
     minTimeSinceLastCheck: Duration,
     limit: Int
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/CombinedRepository.kt
@@ -255,6 +255,9 @@ class CombinedRepository(
 
   override fun getApplicationSummaries(): Collection<ApplicationSummary> =
     deliveryConfigRepository.getApplicationSummaries()
+
+  override fun triggerRecheck(application: String) =
+    deliveryConfigRepository.triggerRecheck(application)
   // END DeliveryConfigRepository methods
 
   // START ResourceRepository methods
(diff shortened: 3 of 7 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -983,6 +983,29 @@ class SqlDeliveryConfigRepository(
...
+        .from(DELIVERY_CONFIG)
+        .where(DELIVERY_CONFIG.APPLICATION.eq(application))
+        .fetchOne(DELIVERY_CONFIG.UID)
+    }
+
+    if (uid == null) {
+      log.error("Config for app $application does not exist, cannot recheck it")
+      return
+    }
+
```
emjburns, line 998: @robfletcher in order to actually make this env immediately recheck I need to clear the leased at field. But, that means that if another instance is checking the env there's a potential to have some cross-contamination. What do you think the best way is to get the environment to be rechecked as soon as possible?

erikmunson:
I have no comment on the code itself, but BOY OH BOY is this a great behavior change. wowowowowow.

oh my god

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -983,6 +983,29 @@ class SqlDeliveryConfigRepository(
     }
   }
 
+  override fun triggerRecheck(application: String) {
```
gal-yardeni, line 986: one small (maybe improvement) - you can pass the delivery config itself here from the caller functions (all of them already fetching it) instead of fetching it by app name.

emjburns: We need the UID of the delivery config because that's what the checked at table uses. I'm not loading the entire delivery config here. And we have to fetch that UID because it doesn't come with the delivery config object :(

gal-yardeni: ohhh in that case, forget it 😅

gal-yardeni (approved):
thanks for making this improvement!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
robfletcher: Yeah, I think this is legit, although I wonder if it makes sense to clear the `LEASED_AT` field as well? It seems like this will leave it (however briefly) in a weird state where it was leased at time x, but we don't know who by.

emjburns: Okay, I'll clear that field as well for consistency.

lorin (approved):
My understanding.

## Problem

When the user takes certain actions (pin, unpin, mark as bad, unmark as bad),
it can take a while between the time they initiate the request and the time that keel actually starts to take action. That's because if the environment being changed was checked recently, then keel won't take action on it until the next cycle when it gets checked.

## Implemented solution

When the user does one of (un)pin, (un)mark as bad, reset the environment's last checked time to ensure that it gets prioritized in the next environment check cycle.

## Code changes

New method `SqlDeliveryConfigRepository.triggerRecheck` - resets the last checked at time to epoch+1 to ensure that it will get checked on next cycle

This method is invoked in the following ApplicationService methods:

* pin
* deletePin
* markasVetoedIn
* deleteVeto


## Feedback

I do see one race condition which could lead to keel behaving as if the `triggerRecheck` call did not take effect. This can happen due to interleaving of the CheckScheduler.checkEnvironments code with when the triggerRecheck is called.

Here's a scenario that illustrates it for a hypothetical app named "fnord".

I'll use t1 to refer to the check scheduler thread, and t2 to refer to the thread executed by a "pin" API call.

1. t1: [checkEnvironments](https://github.com/spinnaker/keel/blob/ff12ef877d6251892235d2a5891ae72b94662b1b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt#L113) fires
1. t1: [repository.deliveryConfigsDueForCheck](https://github.com/spinnaker/keel/blob/ff12ef877d6251892235d2a5891ae72b94662b1b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt#L120) returns a list of configs, including app "fnord"
1. t1: [environmentPromotionChecker.checkEnvironments](https://github.com/spinnaker/keel/blob/ff12ef877d6251892235d2a5891ae72b94662b1b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt#L131) gets called for all of the 1. environments (environments have now all been checked)
1. t2: user initiates a pin for the fnord app
1. t2: The triggerRecheck method is called, which sets the `DELIVERY_CONFIG_LAST_CHECKED.AT` column to epoch+1 for the fnord app.
1. t1: [repository.markDeliveryConfigCheckComplete](https://github.com/spinnaker/keel/blob/ff12ef877d6251892235d2a5891ae72b94662b1b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt#L137) is called, which in turn calls  [SqlDeliveryConfigRepository.markCheckComplete](https://github.com/spinnaker/keel/blob/9354802bfd08ba3e67b32cdf658427181c3052f6/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt#L1032), which sets `DELIVERY_CONFIG_LAST_CHECKED.AT` to [*now*](https://github.com/spinnaker/keel/blob/9354802bfd08ba3e67b32cdf658427181c3052f6/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt#L1037) for the fnord app.

The `markDeliveryConfigCheckComplete` call has clobbered the value in `AT` column.

The current implementation is strictly better than we have right now, and won't lead to any correctness issues, so I'm going to mark as approve.

lorin:
You could eliminate the race condition by modifying the [SqlDeliveryConfigRepository.markCheckComplete](https://github.com/spinnaker/keel/blob/9354802bfd08ba3e67b32cdf658427181c3052f6/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt#L1032) method so that it doesn't do the update if *triggerRecheck* was called. 

For example, you could update the `WHERE` clause to only write if `LEASED_BY` is not null.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -215,6 +221,20 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
             get { metadata }.isNotEmpty()
           }
       }
+
+      test("config can be rechecked") {
+        val items = repository.itemsDueForCheck(Duration.ofMinutes(2), 1)
+        val items2 = repository.itemsDueForCheck(Duration.ofMinutes(2), 1)
+        repository.triggerRecheck(deliveryConfig.application)
+        val items3 = repository.itemsDueForCheck(Duration.ofMinutes(2), 1)
```
luispollo, line 229: Consider renaming for clarity:
```suggestion
        val firstCheck = repository.itemsDueForCheck(Duration.ofMinutes(2), 1)
        val secondCheck = repository.itemsDueForCheck(Duration.ofMinutes(2), 1)
        repository.triggerRecheck(deliveryConfig.application)
        val afterRecheck = repository.itemsDueForCheck(Duration.ofMinutes(2), 1)
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: I think @gal-yardeni's suggestion is legit. In fact, I think we should standardize on passing the objects to internal APIs instead of keys, as this kind of sub-optimal, redundant look up is all over the place. We could just add `uid` with a `@JsonIgnore` to `DeliveryConfig`. We already do that for `ConstraintState`, as an example.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: Isn't setting it to `EPOCH` the equivalent of clearing it, though, since that is the default value? I don't think that column is nullable (or maybe I'm confusing with `AT`?).

luispollo (approved):
Just a couple of minor suggestions, otherwise looks good to me! Thanks for taking this up!

emjburns:
@lorin great idea! I'm going to update the `triggerRecheck` function to put a special "leased_by" value. Then, if that value is present when we go to mark the check complete we won't clobber it. This makes the recheck case distinctly different from the new delivery config case for those values.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
emjburns: @luispollo I think that would be better done as a separate change, but I agree with you and Gal. Seems like it would be an improvement.

emjburns:
Well, I'm somehow in a weird git state. I'll make the test naming suggestion and lorin's race condition suggestion in a separate PR and merge that in.

PR #1797, opened 2021-02-09
fix(api): verification summary issues

lorin:
This fixes two bugs in the API used by the environments view that displays verifications data.

## Problems

Two separate cases weren't being handled properly

* when the artifact reference does not match artifact name
* when the list of verification contexts is empty in the `getStatesBatch` call.

### Copy objects to add artifact reference

When querying the artifact repository for a list of published artifact versions, we now add the artifact references by copying each element of the list. 

This isn't the most efficient way of doing this, but code-wise it was the smallest change.

### Use reduceOrNull instead of reduce

Kotlin's `reduce` throws an exception if you give it an empty list. Change to using `reduceOrNull` and then do null handling.

Note: You'll want to hide whitespace when looking at the diff.

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
@@ -221,7 +221,7 @@ class SqlArtifactRepository(
         .where(ARTIFACT_VERSIONS.NAME.eq(artifact.name))
         .and(ARTIFACT_VERSIONS.TYPE.eq(artifact.type))
         .fetchSortedArtifactVersions(artifact, limit)
-        .map { it.copy(reference=artifact.reference) }
+        .map { it.copy(reference = artifact.reference) }
     }
   }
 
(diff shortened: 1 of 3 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -221,6 +221,7 @@ class SqlArtifactRepository(
         .where(ARTIFACT_VERSIONS.NAME.eq(artifact.name))
         .and(ARTIFACT_VERSIONS.TYPE.eq(artifact.type))
         .fetchSortedArtifactVersions(artifact, limit)
+        .map { it.copy(reference=artifact.reference) }
```
luispollo, line 224: Formatting:
```suggestion
        .map { it.copy(reference = artifact.reference) }
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -196,46 +196,48 @@ class SqlVerificationRepository(
      * This function guarantees that the number of output elements match the number of input elements.
      * So we use left joins on the givenVersions table
      */
-    return jooq.select(
+    return contextTable.table?.let { ctxTable ->
```
luispollo, line 199: nit: I think this might be a bit easier to reason about if you added an early return here like so:
```kotlin
return if contextTable.table == null {
  emptyList()
} else {
  // select code
}
```

luispollo (approved):
Couple nits, otherwise LGTM.

PR #1802, opened 2021-02-11
feat(verif): more robust handling of invalid verification id

lorin:
# Problem

Environments view was blowing up for a test app in our staging environment because the `verification_state` table had an invalid verification id in one of the entries.

The invalid data looked like this: `spkr/mytest/stable` where the valid id was `spkr/mytest:stable`.

We had changed the convention for the id a while back from `/` to `:` for delimiting the tag of the container, but there was some data that hadn't been updated.

While this could have been fixed by just repairing the data in the database, having the environments view break entirely because of one invalid verificiations record is an undesirable user experience.

# Implemented solution

Explicitly detect and handle invalid verification ids.

The `VerificationContext.verification(id: String)` function now returns a null if the id doesn't exist, which forces the caller to explicitly handle it.

## Environment view

The ApplicationService, which computes the environments view, now handles this by skipping the invalid record. It also logs an error message and emits a metric: `keel.verification.invalid.id.seen`

## Verification controller

The `VerificationController` returns a meaningful error message if a client calls it with an invalid id. Here's example output

```yaml
timestamp: "2021-02-11T17:51:50.148+0000"
status: 404
error: "Not Found"
message: "Unknown verification id: spkr/fakeetest:fail. Expecting one of: [spkr/faketest:fail]"
```

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -578,7 +578,7 @@ class ApplicationService(
       .mapNotNull { (vId: String, state: VerificationState) ->
         ctx.verification(vId) // Verification?
           ?.let { verification -> verification to state } // Pair<Verification, VerificationState>?
-          .also { if (it == null) { onInvalidVerificationId(vId, deliveryConfig, ctx) }}
+          .also { if (it == null) { onInvalidVerificationId(vId, deliveryConfig, ctx) } }
       }
       .toMap()
 
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -53,6 +53,7 @@ import com.netflix.spinnaker.keel.lifecycle.LifecycleEventRepository
 import com.netflix.spinnaker.keel.persistence.ArtifactNotFoundException
 import com.netflix.spinnaker.keel.persistence.KeelRepository
 import com.netflix.spinnaker.keel.persistence.NoSuchDeliveryConfigException
+import com.netflix.spinnaker.keel.telemetry.InvalidVerificationIdSeen
 import org.slf4j.LoggerFactory
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.core.env.Environment as SpringEnvironment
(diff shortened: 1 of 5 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -552,25 +553,52 @@ class ApplicationService(
...
     return repository.getArtifactVersion(artifact, version, releaseStatus)
   }
-}
 
-/**
- * Query the repository for all of the verification states associated with [versions]
- *
- * This just calls [KeelRepository.getVerificationStatesBatch] and reshapes the returned value to a map
- */
-fun KeelRepository.getVerificationStates(
```
lorin, line 562: This function is a bit of a hairball. I made an effort ti refactor it into smaller functions.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -552,25 +553,52 @@ class ApplicationService(
...
+
+  /**
+   * Convert a (verification id -> verification state) map to a (verification -> verification state) map
+   *
+   * Most of the logic in this method is to deal with the case where the verification id is invalid
+   */
+  fun Map<String, VerificationState>.toVerificationMap(deliveryConfig: DeliveryConfig, ctx: VerificationContext) : Map<Verification, VerificationState> =
+    entries
+      .mapNotNull { (vId: String, state: VerificationState) ->
+        ctx.verification(vId) // Verification?
```
luispollo, line 579: nit: I think the EOL comments with the types are unnecessary. The IDE will give us those hints when we need them.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -552,25 +553,52 @@ class ApplicationService(
...
+          ?.let { verification -> verification to state } // Pair<Verification, VerificationState>?
+          .also { if (it == null) { onInvalidVerificationId(vId, deliveryConfig, ctx) }}
```
luispollo, line 581:
```suggestion
          .also { if (it == null) { onInvalidVerificationId(vId, deliveryConfig, ctx) } }
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -552,25 +553,52 @@ class ApplicationService(
...
+    }
+
+  /**
+   * Convert a (verification id -> verification state) map to a (verification -> verification state) map
+   *
+   * Most of the logic in this method is to deal with the case where the verification id is invalid
+   */
+  fun Map<String, VerificationState>.toVerificationMap(deliveryConfig: DeliveryConfig, ctx: VerificationContext) : Map<Verification, VerificationState> =
+    entries
+      .mapNotNull { (vId: String, state: VerificationState) ->
```
luispollo, line 578: I think a slightly more idiomatic way of doing this would be using `mapKeys` here and a `filterNotNull` at the end (avoids the need to convert back to a map at the end).

luispollo (approved):
Couple nits, otherwise LGTM!

robfletcher (approved):
Don't think I have anything to add beyond what Luis said. Looks good.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
lorin: It looks like there is no  built-in `filterNotNull` function defined on maps, only `filterNotNullValues`, where here what I want to do is filter out null keys.

PR #1803, opened 2021-02-11
fix sub-queries that assume resource id is unique

robfletcher:
Bug detected in prestaging


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -307,13 +307,15 @@ class SqlDeliveryConfigRepository(
...
+          // are moving it from one environment to another)
           jooq.deleteFrom(ENVIRONMENT_RESOURCE)
             .where(ENVIRONMENT_RESOURCE.ENVIRONMENT_UID.notEqual(environmentUID))
             .and(
-              ENVIRONMENT_RESOURCE.RESOURCE_UID.equal(
-                select(RESOURCE.UID).from(RESOURCE).where(RESOURCE.ID.eq(resource.id))
+              ENVIRONMENT_RESOURCE.RESOURCE_UID.`in`(
+                select(RESOURCE.UID)
+                  .from(RESOURCE)
+                  .where(RESOURCE.ID.eq(resource.id))
```
robfletcher, line 318: I could have added `.and(RESOURCE.VERSION.eq(resource.version))` to this sub-query but that version may have changed, so I think it makes more sense to remove _all_ references regardless of the version.

lorin:
This looks good to me so far, but I know there's more coming, so I'll wait for the additional code before giving it the ol' 👍

PR #1804, opened 2021-02-12
fix(metrics): add missing increments

lorin:
# Problem

While we were creating spectator counters for verification metrics, we weren't actually incrementing them, so no metrics were reported.

# Implemented solution

Increment the danged counters.


luispollo:
> Increment the danged counters.

🤣

luispollo (approved):

emjburns (approved):

PR #1806, opened 2021-02-13
chore(perf): eliminate a lot of redundant calls for env api

emjburns:
When we were generating the artifacts portion of the environments view we were making lots of redundant calls to get the artifact versions. This pr fetches data for all versions and status info in all environments and then passes that info in for later functions to use.

I am 99% sure this will help significantly with the loading time for the environments view. Locally it helped a lot in a simple scenario. But I have timing metrics to measure this in our actual environments. 

I'd appreciate a review on the "compare link tests" specifically. Because of my refactoring I think it's easier to see what is happening. But, I had to change the tests to be more "realistic" so the versions changed. I think I did this right! But I'd appreciate a second look.


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -91,9 +91,8 @@ class ApplicationService(
   private val RESOURCE_SUMMARY_CONSTRUCT_DURATION_ID = "keel.api.resource.summary.duration"
   private val ENV_SUMMARY_CONSTRUCT_DURATION_ID = "keel.api.environment.summary.duration"
   private val ARTIFACT_SUMMARY_CONSTRUCT_DURATION_ID = "keel.api.artifact.summary.duration"
-  private val CONSTRAINT_SUMMARY_CONSTRUCT_DURATION_ID = "keel.api.constraint.summary.duration"
+  private val ARTIFACT_IN_ENV_SUMMARY_CONSTRUCT_DURATION = "keel.api.artifact.in.environment.summary.duration"
```
luispollo, line 94: Will you be able to compare this new metric with the other two that it's replacing? Just trying to understand if you'll be able to do a before/after comparison given that you're changing the metrics.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -277,18 +276,18 @@ class ApplicationService(
     }
 
     val artifactSummaries = deliveryConfig.artifacts.map { artifact ->
+      // We calculate a bunch of data for each artifact, so pre-load the bulk of it here and pass that info into
+      //  later functions to save redundant database calls.
```
luispollo, line 280:
```suggestion
      // later functions to save redundant database calls.
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -277,18 +276,18 @@ class ApplicationService(
...
       val artifactVersions = repository.artifactVersions(artifact, limit)
+      val artifactVersionInfoInEnvironment: Map<String, List<StatusInfoForArtifactInEnvironment>> = deliveryConfig.environments.map { env ->
+        env.name to repository.getVersionInfoInEnvironment(deliveryConfig, env.name, artifact)
+      }.toMap()
```
luispollo, line 284:
```suggestion
      val artifactVersionInfoInEnvironment = deliveryConfig.environments.associate { env ->
        env.name to repository.getVersionInfoInEnvironment(deliveryConfig, env.name, artifact)
      }
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -373,33 +380,42 @@ class ApplicationService(
...
       ?.copy(verifications = verifications)
 
-  private fun buildArtifactSummaryInEnvironment(deliveryConfig: DeliveryConfig, environmentName: String, artifact: DeliveryArtifact, version: String, status: PromotionStatus): ArtifactSummaryInEnvironment? {
-    val currentArtifact = getArtifactInstance(artifact, version)
-
+  private fun buildArtifactSummaryInEnvironment(
+    deliveryConfig: DeliveryConfig,
+    environmentName: String,
+    artifact: DeliveryArtifact,
+    currentArtifact: PublishedArtifact,
```
luispollo, line 396: Consider standardizing on one name for this object. You have `currentArtifact` here and `artifactVersion` on line 383.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -443,15 +459,45 @@ class ApplicationService(
     }
   }
 
+  private fun getArtifactVersionByPromotionStatus(
+    environmentName: String,
+    promotionStatus: PromotionStatus,
+    version: String?,
+    artifactInfoInEnvironment: Map<String, List<StatusInfoForArtifactInEnvironment>>,
+    allVersions: List<PublishedArtifact>
```
luispollo, line 467: Consider standardizing on a name for this object. You have `allVersions` here and `versions` above. I prefer `allVersions`.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -443,15 +459,45 @@ class ApplicationService(
...
+  ): PublishedArtifact? {
+    //only CURRENT and PREVIOUS are supported, as they can be sorted by deploy_at
+    require(promotionStatus in listOf(CURRENT, PREVIOUS)) { "Invalid promotion status used to query" }
+    val ourEnv = artifactInfoInEnvironment[environmentName] ?: return null
```
luispollo, line 471:
```suggestion
    val envArtifactVersions = artifactInfoInEnvironment[environmentName] ?: return null
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -443,15 +459,45 @@ class ApplicationService(
...
+    // we sort the versions by their deployed at time in descending order, then we take
+    // the first one where all the conditions match.
+    val chosenVersion = if (version == null) {
+      ourEnv.sortedByDescending { it.deployedAt }.firstOrNull { it.status == promotionStatus }
```
luispollo, line 475:
```suggestion
      envArtifactVersions.sortedByDescending { it.deployedAt }.firstOrNull { it.status == promotionStatus }
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -443,15 +459,45 @@ class ApplicationService(
...
+    } else {
+      ourEnv.sortedByDescending { it.deployedAt }.firstOrNull { it.status == promotionStatus && it.replacedByVersion == version}
```
luispollo, line 477:
```suggestion
      envArtifactVersions.sortedByDescending { it.deployedAt }.firstOrNull { it.status == promotionStatus && it.replacedByVersion == version}
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -443,15 +459,45 @@ class ApplicationService(
...
+    } ?: return null
+    return allVersions.find { it.version == chosenVersion.version }
+  }
+
   // Pinning is a special case when is coming to creating a compare link between versions.
   // If there is a pinned version, which is not the same as the current version, we need
   // to make sure we are creating the comparable link with reference to the pinned version.
```
luispollo, line 484:
```suggestion
  // Pinning is a special case when it comes to creating a compare link between versions.
  // If there is a pinned version, which is not the same as the current version, we need
  // to make sure we are creating the compare link with reference to the pinned version.
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -443,15 +459,45 @@ class ApplicationService(
...
     val pinnedVersion = repository.getPinnedVersion(deliveryConfig, environmentName, artifact.reference)
-     return if (pinnedVersion != version)
-      pinnedVersion?.let { getArtifactInstance(artifact, it) }
-    else { //if pinnedVersion == current version, fetch the version which that the pinned version replaced
-       repository.getArtifactVersionByPromotionStatus(deliveryConfig, environmentName, artifact, PREVIOUS, pinnedVersion)
+     return if (pinnedVersion != version) {
+       pinnedVersion?.let {
+         allVersions.find { it.version == pinnedVersion }
+       }
+     } else { //if pinnedVersion == current version, fetch the version which that the pinned version replaced
```
luispollo, line 498:
```suggestion
     } else { //if pinnedVersion == current version, fetch the version which the pinned version replaced
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -611,3 +650,15 @@ fun DeliveryConfig.contexts(
   versions.flatMap { version ->
     environments.map { env -> VerificationContext(this, env, version) }
   }
+
+/**
+ * Holds the info we need about artifacts in an environment for building the UI view.
+ *
+ * This is used in a list of versions pertaining to a specific delivery artifact.
+ */
+data class StatusInfoForArtifactInEnvironment(
```
luispollo, line 659: Should this class maybe be defined in `keel-sql` instead, since that's where the data is coming from and `keel-core` already has a dependency on it?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ApplicationServiceTests.kt
```diff
@@ -232,7 +236,14 @@ class ApplicationServiceTests : JUnit5Minutests {
...
     fun Collection<String>.toArtifactVersions(artifact: DeliveryArtifact) =
-      map { PublishedArtifact(artifact.name, artifact.type, artifact.reference, it) }
+      map { PublishedArtifact(
+        name = artifact.name,
+        type = artifact.type,
+        reference = artifact.reference,
+        version = it,
+        gitMetadata = gitMetadata,
+        buildMetadata = buildMetadata
+      ) }
```
luispollo, line 246:
```suggestion
      )
    }
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
```diff
@@ -147,21 +150,31 @@ class ComparableLinksTests : JUnit5Minutests {
...
-      map { PublishedArtifact(artifact.name, artifact.type, artifact.reference, it) }
+      map { version ->
+        PublishedArtifact(
+        name = artifact.name,
+        type = artifact.type,
+        reference = artifact.reference,
+        version = version,
+        gitMetadata = git(version),
+        buildMetadata = buildMetadata,
+      ) }
```
luispollo, line 177:
```suggestion
      )
    }
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
```diff
@@ -339,82 +319,49 @@ class ComparableLinksTests : JUnit5Minutests {
...
+          )
         }
+
         test ("get the correct compare link when pinning forward (current)") {
+          // pinning to version 3 here
           val summaries = applicationService.getArtifactSummariesFor(application1)
           expectThat(summaries.first())
             .withVersionInEnvironment(version1, "staging") {
-              compareLink.isEqualTo("https://stash/projects/spkr/repos/keel/compare/commits?targetBranch=fnord-1.0.1-h1.b1b1b1b&sourceBranch=pinnedVersion")
+              compareLink.isEqualTo("https://stash/projects/spkr/repos/keel/compare/commits?targetBranch=${version1}&sourceBranch=${version3}")
```
luispollo, line 340: nit: The compiler will give you warnings for all these unnecessary curly braces. I think that's annoying, personally (I don't mind the curly braces, makes things easier to read!), but thought I'd point it out.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
luispollo: P.S. Seems so obvious to load all the data upfront in hindsight, doesn't it? 🙂  Thanks for making this optimization! ❤️

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -299,31 +298,39 @@ class ApplicationService(
           environmentSummary.getArtifactPromotionStatus(artifact, artifactVersion.version)
             ?.let { status ->
               if (artifact.isUsedIn(environment)) { // only add a summary if the artifact is used in the environment
+                val artifactInEnvStartTime = now
                 buildArtifactSummaryInEnvironment(
```
luispollo, line 302: I find the number of parameters to this function to be unwieldy at this point. It reminded of the verification stuff where we debated this in a PR review and Lorin introduced a `VerificationContext`, which I thought was create. I wonder if we could generalize that class a bit to be used elsewhere and hold these related context bits (delivery config, environment, artifact at least, but possibly other bits as well).

I also wondered if we could simplify this a bit further by using just `artifactVersionInfoInEnvironment` and ditching `artifactVersions`. I wasn't sure why you're filtering for only `CURRENT` and `PREVIOUS` in the query to build the artifact info, but maybe you could consider removing that filter from the query so you could use a single list here?

luispollo (approved):
Even though I really don't think the additional queries were causing a major performance impact to the API call (UI slowness in the environments view was not caused by the backend, as we've determined), this looks like an excellent optimization to me! I had a few suggestions on naming and potential further simplifications, but nothing blocking. Curious to see what the before/after times look like!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
```diff
@@ -250,6 +250,15 @@ interface ArtifactRepository : PeriodicallyCheckedRepository<DeliveryArtifact> {
...
 
+  /**
+   * Get a map of the versions and status for an environmnet
+   */
+  fun getVersionInfoInEnvironment(
+    deliveryConfig: DeliveryConfig,
+    environmentName: String,
+    artifact: DeliveryArtifact
+  ): List<StatusInfoForArtifactInEnvironment>
+
```
emjburns, line 261: We query this information _a lot_ of times, and previously we were just looking it up for each individual version. The main change I made in this PR is to pull info for all versions at once and then pass that into subsequent function calls so they look it up from the list we pulled all at once. This allows us to eliminate a bunch of calls to the db.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: Good question. The metric I removed was useless to measure (it was taking 25millis each time, and was measuring constructing things for each version). So I removed it. I added another metric now that I better understand how the function works. The main metric I'm measuring (`keel.api.artifact.summary.duration`, with an approximate average of 6 seconds) will remain the same so that it's useful.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: I purposefully put spaces on the front of continuation comments so they're visually grouped

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: It took me quite a lot of staring to realize how much this would save in db calls! I'm so thrilled to open this PR.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -277,18 +276,18 @@ class ApplicationService(
...
-      val verificationStartTime = now
       // A verification context identifies an artifact version in an environment
       // For each context, there may be multiple verifications (e.g., test-container, canary)
       //
       // This map associates a context with this collection of verifications and their states
       val verificationStateMap = getVerificationStates(deliveryConfig, artifactVersions)
-      spectator.timer(
-        VERIFICATION_SUMMARY_CONSTRUCT_DURATION_ID,
-        listOf(BasicTag("application", application))
-      ).record(Duration.between(verificationStartTime, now))
```
emjburns, line 291: This didn't seem worthwhile to measure (so little time) so I'm removing the metric.

emjburns:
@luispollo Rani and I were wrong this week, the slowness is coming from the backend. This call which I am optimizing here is taking around 6 seconds on average (`keel.api.artifact.summary.duration` is the metric to look at). I'm hoping this makes a significant performance improvement because I'm eliminating a lot of calls.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: changing to `currentEnvVersionInfo`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: not changed in my PR, not going to touch.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: 🤷 I like defining it here because it's only used here. I think it really matters.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: I don't think there's enough benefit to doing the kind of refactoring you're mentioning. The goal of this PR is to change the minimum possible things and see if it improves our API time. I'll likely refactor this whole method in the near future to support the API work that Rani wants. 

Some of this PR (like the `getArtifactVersionByPromotionStatus(...)` functions) have been lifted from the SQL repository to optimize. For this case (filtering by current and previous) there's a clear function comment as to why (`//only CURRENT and PREVIOUS are supported, as they can be sorted by deploy_at`).

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
```diff
@@ -147,21 +151,30 @@ class ComparableLinksTests : JUnit5Minutests {
       number = "1",
     )
```
emjburns, line 153: Because I changed the way we interact with the database (we pull the artifact versions info as one list and the status info as a separate list with just the version) we now get the artifact details from the list of artifact versions. Because of that, I needed to change the way this test was structured. The `git(...)` function here is responsible for making that list of artifacts.

I was able to remove a lot of mocked calls because we are no longer making redundant db calls for every single version. But, because of that, the versions in the test changed because the pinned version became a real version and not just the version string "pinned".

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
```diff
@@ -245,87 +244,88 @@ class ComparableLinksTests : JUnit5Minutests {
           emptyList()
         }
 
-        every { repository.artifactVersions(releaseArtifact) } returns versions.toArtifactVersions(releaseArtifact)
```
emjburns, line 248: This data structure is a shortened version of the `ENVIRONMENT_ARTIFACT_VERSIONS` table. As I mentioned above, we pull this once and use this list to avoid querying the db excessively for each version. As a result I think the test is a little easier to read (you can see the "set up" - here version 0 was replaced by version 1, version 1 replaced by version 2, version 2 is current).

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
```diff
@@ -245,87 +244,88 @@ class ComparableLinksTests : JUnit5Minutests {
...
+            PREVIOUS,
+            version1,
+            clock.instant().minus(Duration.ofHours(4))
+          ),
+          StatusInfoForArtifactInEnvironment(version1, CURRENT, null, clock.instant().minus(Duration.ofHours(2))),
+        )
       }
 
       context("stash") {
         before {
```
emjburns, line 279: Removed a lot of mocks like this because we no longer call the db for this info, we pull from the list of artifact version info that we load once.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
gal-yardeni: what happens if a new version comes in in the meantime?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
gal-yardeni: woohoo!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepository.kt
emjburns: we will get that version on the next load of the environments UI. The perf gain is definitely worth it here - i'd rather the env load be snappy

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -1206,6 +1206,33 @@ class SqlArtifactRepository(
...
   }
 
+  override fun getVersionInfoInEnvironment(
+    deliveryConfig: DeliveryConfig,
+    environmentName: String,
+    artifact: DeliveryArtifact
+  ): List<StatusInfoForArtifactInEnvironment> =
+    sqlRetry.withRetry(READ) {
+      jooq
+        .select(
```
lorin, line 1216: My instinct is to add a limit to this query. I can imagine this returning a very large number of records.

gal-yardeni (approved):
this refactor makes sense to me! I hope it'll make a difference with performance!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -443,15 +461,45 @@ class ApplicationService(
...
+    require(promotionStatus in listOf(CURRENT, PREVIOUS)) { "Invalid promotion status used to query" }
+    val currentEnvVersionInfo = artifactInfoInEnvironment[environmentName] ?: return null
+    // we sort the versions by their deployed at time in descending order, then we take
+    // the first one where all the conditions match.
+    val chosenVersion = if (version == null) {
+      currentEnvVersionInfo.sortedByDescending { it.deployedAt }.firstOrNull { it.status == promotionStatus }
+    } else {
+      currentEnvVersionInfo.sortedByDescending { it.deployedAt }.firstOrNull { it.status == promotionStatus && it.replacedByVersion == version}
+    } ?: return null
+    return allVersions.find { it.version == chosenVersion.version }
```
lorin, line 481: Might be worth checking if binary search gives us a performance boost here (although we would need to pass in the sorting strategy).

lorin (commented):
I didn't get a chance to finish my review, I had a couple of comments inline.

PR #1809, opened 2021-02-16
chore(test): rename test label so 'failed' doesn't appear

lorin:
## Problem

When tests fail on GitHub Actions, I search for the string `" failed"` to jump to the failing test lines.

When I do that, the first search hit is a successful test that has the string `"failed"` in its description.
```
BakeryLifecycleMonitorTests publishes failed event and marks failure PASSED
```

## Implemented solution

Change the test description text from `failed` to `fail(ed)` so it doesn't match a `failed` string search.


emjburns (approved):
short and sweet and useful, just the way I like it

PR #1813, opened 2021-02-17
fix(api): remove 3 sql queries per artifact version

emjburns:
This pr batch loads artifact evironment version data which was previously loaded by doing 3 SQL queries per artifact version requested. So, if we have 1 artifact in one environment and we load 30 versions for the environments UI, this takes 30*3 = 90 SQL queries and reduces it to 3.

I could further optimize this API by bulk loading constraints data (we are also making 1 SQL query per version and environment also) but I decided not to do that in this PR.

I also created a context object for the data loaded (like Luis had previously suggested) because I was adding even more data to each function.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -278,11 +278,26 @@ class ApplicationService(
     val artifactSummaries = deliveryConfig.artifacts.map { artifact ->
       // We calculate a bunch of data for each artifact, so pre-load the bulk of it here and pass that info into
       //  later functions to save redundant database calls.
+
+      // Load all artifact version information
       val artifactVersions = repository.artifactVersions(artifact, limit)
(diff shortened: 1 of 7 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -379,54 +402,31 @@ class ApplicationService(
...
   ): ArtifactSummaryInEnvironment? {
     // some environments contain relevant info for skipped artifacts, so
     // try and find that summary before defaulting to less information
-    val potentialSummary =
-      repository.getArtifactSummaryInEnvironment(
-        deliveryConfig = deliveryConfig,
-        environmentName = environmentName,
-        artifactReference = artifact.reference,
-        version = currentArtifact.version
-      ) //todo eb: can we eliminate this call?
```
emjburns, line 411: spoiler alert! we can eliminate this call :D

gal-yardeni (approved):
LGTM!

PR #1815, opened 2021-02-19
feat(slack): add a warning if there's a pinned version

gal-yardeni:
This PR will add a warning to the manual judgment notification, in case there's a pinned version in the environment.
(As a reminder, if there's a pinned version, even if the user will click on `approve`, nothing will happen).
It will look like this:
Screen Shot 2021-02-19 at 11 23 16 AM

### DIFF
--- a/keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
+++ b/keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
@@ -273,9 +273,16 @@ class NotificationEventListener(
 
         val deliveryArtifact = configAndArtifact.first.artifacts.find {
           it.reference == currentState.artifactReference
-        }?: return
+        } ?: return
 
         val currentArtifact = repository.getArtifactVersionByPromotionStatus(configAndArtifact.first, currentState.environmentName, deliveryArtifact, PromotionStatus.CURRENT)
+
+        // fetch the pinned artifact, if exists
+        val pinnedArtifact =
+          repository.getPinnedVersion(configAndArtifact.first, currentState.environmentName, deliveryArtifact.reference)?.let {
+            repository.getArtifactVersion(deliveryArtifact, it, null)
+              ?.copy(reference = deliveryArtifact.reference)
+          }
 
         sendSlackMessage(
           configAndArtifact.first,
@@ -286,6 +293,7 @@ class NotificationEventListener(
             targetEnvironment = currentState.environmentName,
             currentArtifact = currentArtifact,
             deliveryArtifact = deliveryArtifact,
+            pinnedArtifact = pinnedArtifact,
             stateUid = currentState.uid
           ),
           MANUAL_JUDGMENT_AWAIT,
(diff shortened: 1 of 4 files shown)


lorin (approved):
love it

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/ManualJudgmentNotificationHandler.kt
```diff
@@ -50,6 +50,20 @@ class ManualJudgmentNotificationHandler(
...
 
+        // Add a warning section in case there's a pinned artifact
+        if (pinnedArtifact != null){
+          section {
+            markdownText(":warning: Another version is pinned here. You will need to unpin it first to promote this version.")
+            accessory {
+              button {
+                text("See pinned version", emoji = true)
+                actionId("button:url:pinned")
+                url(gitDataGenerator.generateArtifactUrl(application, pinnedArtifact.reference, pinnedArtifact.version))
```
luispollo, line 61: It looks like `generateArtifactUrl` is only used in this class and so should probably go here instead of in `GitDataGenerator`.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
```diff
@@ -273,10 +273,17 @@ class NotificationEventListener(
...
         val deliveryArtifact = configAndArtifact.first.artifacts.find {
           it.reference == currentState.artifactReference
-        }?: return
+        } ?: return
 
         val currentArtifact = repository.getArtifactVersionByPromotionStatus(configAndArtifact.first, currentState.environmentName, deliveryArtifact, PromotionStatus.CURRENT)
 
+        // fetch the pinned artifact, if exists
+        val pinnedArtifact =
+          repository.getPinnedVersion(configAndArtifact.first, currentState.environmentName, deliveryArtifact.reference)?.let {
```
luispollo, line 282: Small suggestion: I find the `configAndArtifact.first` and `configAndArtifact.second` expressions confusing. You could replace that with one variable for each by doing this on line 266:
```kotlin
val (config, artifact) = currentState.artifactReference?.let {
  // ...
}
```

luispollo (approved):
Couple small suggestions, otherwise looks good to me. Great little addition to the notification! ❤️

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/ManualJudgmentNotificationHandler.kt
gal-yardeni: all the relevant data to constructing the URL is there, and we will probably need it for future use :)

gal-yardeni: and i also added some other usage for it

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/NotificationEventListener.kt
gal-yardeni: ohhhh fancy! will make the change, thanks

gal-yardeni: done

gcomstock:
Love it, thanks Gal!

PR #1816, opened 2021-02-19
Fix orphaned old resource versions

robfletcher:
Deleting a delivery config would leave behind non-current versions of resources

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
@@ -34,6 +34,7 @@ import strikt.api.expectCatching
 import strikt.api.expectThat
 import strikt.api.expectThrows
 import strikt.assertions.contains
+import strikt.assertions.containsExactlyInAnyOrder
 import strikt.assertions.hasSize
 import strikt.assertions.isA
 import strikt.assertions.isEmpty
@@ -43,6 +44,7 @@ import strikt.assertions.isNotEmpty
 import strikt.assertions.isNotNull
 import strikt.assertions.isNull
 import strikt.assertions.isSuccess
+import strikt.assertions.map
 import strikt.assertions.none
 import java.time.Duration
 
@@ -486,8 +488,8 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
 
       context("deleting a delivery config") {
         before {
-          store()
           storeResources()
+          store()
           resourceRepository.appendHistory(ApplicationActuationPaused(deliveryConfig.application, "test"))
           pausedRepository.pauseApplication(deliveryConfig.application, "test")
           deliveryConfig.resources.forEach { resource ->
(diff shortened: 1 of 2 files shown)


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -486,8 +488,8 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
 
       context("deleting a delivery config") {
         before {
-          store()
           storeResources()
+          store()
```
robfletcher, line 492: The order of calls is important here. If the resources are not persisted before the delivery config, they do not get attached to it. This was never working.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -532,6 +534,26 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
...
+
+          test("deletes any older versions of resources") {
+            with(deliveryConfig.resources.first()) {
+              val newResourceVersion = resource(
+                kind = kind,
+                id = spec.id,
+                application = application
+              )
+              resourceRepository.store(newResourceVersion)
+            }
```
robfletcher, line 546: Modify and save one of the resources so a version 2 will now be attached to the delivery config

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -532,6 +534,26 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
...
+
+            expectThat(repository.get(deliveryConfig.name).resources)
+              .hasSize(4)
+              .map { it.version }
+              .containsExactlyInAnyOrder(1, 1, 1, 2)
```
robfletcher, line 551: I could probably remove this, but it was helpful in just ensuring that the version change worked

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -532,6 +534,26 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
...
+
+            repository.deleteByApplication(deliveryConfig.application)
+
+            expectThat(resourceRepository.getResourcesByApplication(deliveryConfig.application)).isEmpty()
```
robfletcher, line 555: Delete the delivery config and check there are now zero resources associated with that application. This failed before because the version 1 of the updated resource would not get deleted

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -202,19 +202,15 @@ class SqlDeliveryConfigRepository(
...
-      .innerJoin(
-        select(RESOURCE.ID, max(RESOURCE.VERSION).`as`("MAX_VERSION"))
-          .from(RESOURCE)
-          .where(RESOURCE.APPLICATION.eq(application))
-          .groupBy(RESOURCE.ID)
-          .asTable("GROUPED_RESOURCE")
-      )
-      .on(RESOURCE.ID.eq(field("GROUPED_RESOURCE.ID")))
-      .and(RESOURCE.VERSION.eq(field("GROUPED_RESOURCE.MAX_VERSION")))
+      .where(RESOURCE.APPLICATION.eq(application))
```
robfletcher, line 213: I like not needing clever code. This private method is only used from the `deleteByApplication` method.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
luispollo: Oh wow! Good find! 😬

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
luispollo: Couldn't agree more. Much clearer now. Good simplification.

luispollo (approved):
LGTM! Thanks for fixing this!

PR #1818, opened 2021-02-22
feat(reports): Add Front50 API to retrieve all apps + fields in Application model

luispollo:
Enablement work for an internal PR to report on Managed Delivery onboarding, which leverages metadata persisted in Front50.

### DIFF
--- a/keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Application.kt
+++ b/keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Application.kt
@@ -9,11 +9,11 @@ import com.fasterxml.jackson.annotation.JsonAnySetter
 data class Application(
   val name: String,
   val email: String,
-  val createTs: String,
   val dataSources: DataSources?,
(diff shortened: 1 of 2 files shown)


keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/model/Application.kt
```diff
@@ -1,13 +1,26 @@
...
   val name: String,
   val email: String,
+  val createTs: String,
   val dataSources: DataSources?,
   val repoProjectKey: String? = null,
   val repoSlug: String? = null,
-  val repoType: String? = null
-)
+  val repoType: String? = null,
+  @get:JsonAnyGetter val details: MutableMap<String, Any?> = mutableMapOf()
```
luispollo, line 17: The `details` field is a "catch-all" map of all non-explicitly-mapped properties that appear in the JSON.

lorin (approved):
Note that the any getter/setter Jackson annotations are magic that I'm not familiar with, so I can't vouch for that part of the code. It looks correct to me, though.

PR #1819, opened 2021-02-23
feat(verif): in delivery config, use "image" instead of "repository/tag"

lorin:
# Problem

Currently, the container image specified in a `verifyWith` section is identified by "repository" and "tag" fields, for example:

```
verifyWith:
- type: test-container
  repository: acme/mytests
  tag: stable
  ...
```

However, `repository`, while used by the spinnaker UI, isn't standard industry terminology. Instead, people generally use the term "image". In addition, the common convention for dealing with tags is to append them to the end of the image name.

# Proposed solution

Enable users to specify the image name and tag using a field named `image`.

```
verifyWith:
- type: test-container
  image: acme/mytests:stable
```

## Deprecating the older syntax

Note: to avoid breaking existing apps, this PR supports both the new and old syntaxes.  However, the older syntax should be considered deprecated, and we are going to remove it once we track down all apps that are using it and convert them over.

### TestContainerVerification

TestContainerVerification is the class that is deserialized with the `verifyWith` entry in the delivery config.  It now takes a nullable `image` field, which will become non-nullable in a future PR. The `repository` and `tag` fields are now also nullable. Those will go away in a future PR.

 now has an `imageId` property, backed by a getter function. This is responsible for identifying whether the newer (image) field is in use or the old (repository/tag) is in use, and computing the correct `imageId` field that will ultimately gets passed to orca.

### ContainerJobConfig

ContainerJobConfig now just takes an `image` parameter instead of repository/tag/digest.

### DIFF
--- a/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
+++ b/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
@@ -47,7 +47,6 @@ data class ContainerJobConfig(
 
   val cloudProvider: String = "titus"
   val cloudProviderType: String = "aws"
-
 }
 
 /**
--- a/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
+++ b/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
@@ -16,19 +16,7 @@ data class ContainerJobConfig(
    *   acme/widget
    *   acme/widget:latest
    */
-  val image: String? = null,
-
-  /**
-   * Repository name associated with the container, e.g.: "acme/widget"
-   *
-   * Deprecated: will be removed once all deployments switch to [image]
-   */
-  val repository: String? = null,
-
-  /**
-   * Deprecated. will be removed once all deployments switch to [image]
-   */
-  val tag: String? = null,
+  val image: String,
 
   val application: String,
   val location: TitusServerGroup.Location,
@@ -53,10 +41,6 @@ data class ContainerJobConfig(
   val waitForCompletion: Boolean = true
 ) {
   init {
-    require((image == null) xor (repository == null)) {
-      "One, and only one, of image or repository must be supplied"
-    }
-
     require(retries >= 0) {
       "Retries must be positive or zero"
     }
@@ -65,19 +49,6 @@ data class ContainerJobConfig(
   val cloudProvider: String = "titus"
   val cloudProviderType: String = "aws"
 
-  /**
-   * Determine imageId depending on the newer field (image) or the deprecated fields (repository, tag)
-   */
-  val imageId : String
-  get() =
-    when {
-      image != null &&  image.contains(":") -> image
-      image != null && !image.contains(":") -> "${image}:latest"
-
-      repository != null && tag != null -> "${repository}:${tag}"
-      repository != null && tag == null -> "${repository}:latest"
-      else -> error("no container image specified")
-    }
 }
 
 /**
@@ -117,7 +88,7 @@ fun ContainerJobConfig.createRunJobStage() =
       "iamProfile" to iamInstanceProfile,
       "region" to location.region,
       "capacityGroup" to capacityGroup,
-      "imageId" to imageId,
+      "imageId" to image,
       "entryPoint" to entrypoint
     ),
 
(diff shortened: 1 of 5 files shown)


keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -39,23 +53,31 @@ data class ContainerJobConfig(
   val waitForCompletion: Boolean = true
 ) {
   init {
-    require((tag == null) xor (digest == null)) {
```
lorin, line 42: zapped digest, so this check is no longer necessary

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -39,23 +53,31 @@ data class ContainerJobConfig(
...
     }
   }
 
-  /**
-   * ID that identifies an image, e.g.:
-   *
-   * acme/widget:latest
-   * acme/widget:sha256:780f11bfc03495da29f9e2d25bf55123330715fb494ac27f45c96f808fd2d4c5
-   */
-  val imageId: String = "$repository:${tag ?: digest}"
```
lorin, line 56: The `imageId` field has changed from a variable to a property. It is used by the `ContainerJobConfig.createRunJobStage` extension function, which is defined at the bottom of this file, and was not modified by this PR.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -10,12 +10,26 @@ const val RUN_JOB_TYPE: String = "runJob"
...
+   *
+   * Deprecated: will be removed once all deployments switch to [image]
    */
-  val repository: String,
-  val tag: String? = "latest",
-  val digest: String? = null,
+  val repository: String? = null,
+
+  /**
+   * Deprecated. will be removed once all deployments switch to [image]
```
luispollo, line 29: Maybe use the `@Deprecated` annotation instead on all these fields? It would give us the extra benefit of compiler warnings as reminders.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
```diff
@@ -39,23 +53,31 @@ data class ContainerJobConfig(
...
-   */
-  val imageId: String = "$repository:${tag ?: digest}"
   val cloudProvider: String = "titus"
   val cloudProviderType: String = "aws"
+
+  /**
+   * Determine imageId depending on the newer field (image) or the deprecated fields (repository, tag)
+   */
+  val imageId : String
+  get() =
```
luispollo, line 72: Indentation is off here.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/verification/TestContainerVerificationEvaluator.kt
```diff
@@ -64,22 +64,26 @@ class TestContainerVerificationEvaluator(
...
     return runBlocking {
       withContext(IO) {
+        val containerJobConfig = ContainerJobConfig(
+          application = verification.application ?: context.deliveryConfig.application,
+          location = verification.location,
+          credentials = verification.location.account,
+          image = verification.image,
+
+          // These are deprecated and will eventually go away, since users now pass the info in the image field
+          repository = verification.repository,
```
luispollo, line 74: It seems to me that, by the time we get here, we could already determine the image from the different optional fields, and so `ContainerJobConfig` should only have an `image` field. The only place we should keep the old fields for backwards compatibility is in the `TestContainerVerification` class.

lorin: That's a good idea. I'm going to move the imageId property to TestContainerVerification.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/batch/ContainerJobConfig.kt
lorin: I really wish spotless would catch this sort of thing. It passes spotlessCheck on my machine.

Issue #1821, opened 2021-02-24
java.sql.SQLSyntaxErrorException: Duplicate column name


lorin:
If an artifact reference and environment happen to have the same name (e.g., `dev`), then the SqlVerificationRepository.getStatesBatch method throws the following exception:

```
java.sql.SQLSyntaxErrorException: Duplicate column name 'dev'
  at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:120)
  at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97)
  at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
  at com.mysql.cj.jdbc.ClientPreparedStatement.executeInternal(ClientPreparedStatement.java:953)
  at com.mysql.cj.jdbc.ClientPreparedStatement.execute(ClientPreparedStatement.java:370)
  at com.zaxxer.hikari.pool.ProxyPreparedStatement.execute(ProxyPreparedStatement.java:44)
  at com.zaxxer.hikari.pool.HikariProxyPreparedStatement.execute(HikariProxyPreparedStatement.java:-1)
  at org.jooq.tools.jdbc.DefaultPreparedStatement.execute(DefaultPreparedStatement.java:209)
  at org.jooq.impl.Tools.executeStatementAndGetFirstResultSet(Tools.java:3956)
  at org.jooq.impl.AbstractResultQuery.execute(AbstractResultQuery.java:295)
  at org.jooq.impl.AbstractQuery.execute(AbstractQuery.java:371)
  at org.jooq.impl.AbstractResultQuery.fetch(AbstractResultQuery.java:354)
  at org.jooq.impl.SelectImpl.fetch(SelectImpl.java:2690)
  at com.netflix.spinnaker.keel.sql.SqlVerificationRepository.getStatesBatch(SqlVerificationRepository.kt:222)
  at com.netflix.spinnaker.keel.persistence.CombinedRepository.getVerificationStatesBatch(CombinedRepository.kt:458)
```

Consider this query, which is valid:

```
mysql> select 0, 'dev', 'dev', 'myartifact-1.0.0' from dual
+---+-----+-----+------------------+
| 0 | dev | dev | myartifact-1.0.0 |
+---+-----+-----+------------------+
| 0 | dev | dev | myartifact-1.0.0 |
+---+-----+-----+------------------+
```

[...]

PR #1822, opened 2021-02-24
fix(api): don't blow up the api if can't query for verifications

lorin:
# Problem

We hit an issue (documented as as #1821) where the batch verification query was throwing an exception, which broke the view.

# Implemented solution

Degrade gracefully by catching the exception, logging the error, and continuing on as if there is no verification info to display.

This ensures that the UI remains functional even if that particular query breaks.


emjburns (approved):
Love

robfletcher (approved):

PR #1824, opened 2021-02-25
feat(api): add basic environments api

emjburns:
A new api for returning the resources and current artifacts for an application, by environment.


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/EnvironmentController.kt
```diff
@@ -11,7 +11,7 @@ import org.springframework.web.bind.annotation.RestController
 @RestController
 @RequestMapping(path = ["/environments"])
 class EnvironmentController(
-  private val repository: KeelRepository
+  private val environmentService: EnvironmentService
 ) {
```
emjburns, line 16: Reusing this method `/environments/{application}` since I'm like 99% confident that it is not used.

emjburns:
Working on fixing the tests - who knew that adding a new constructor would conflict with so many tests..

PR #1828, opened 2021-02-26
Ensure events for all version of a resource are included in history

robfletcher:
Because of the poor performance of the initial attempt at this functionality, this PR includes revisions to the way the `event` table is structured. The most important is that `event.ref` will now be the `id` of a resource for rows where `scope = 'RESOURCE'`. In addition I re-ordered the columns for better indexing performance, and now derive the `ref`, `application`, and `timestamp` columns from `json` rather than requiring us to explicitly insert data. For now these are implemented as duplicate columns but I'll follow up with another PR to remove the old ones once this change is live.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
@@ -43,9 +43,11 @@ import strikt.api.expectCatching
 import strikt.api.expectThat
 import strikt.api.expectThrows
 import strikt.assertions.all
+import strikt.assertions.contains
 import strikt.assertions.first
 import strikt.assertions.hasSize
 import strikt.assertions.isA
+import strikt.assertions.isEmpty
 import strikt.assertions.isEqualTo
 import strikt.assertions.isFailure
 import strikt.assertions.isGreaterThanOrEqualTo
@@ -91,10 +93,9 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
         verify { callback wasNot Called }
       }
 
-      test("getting state history throws an exception") {
-        expectThrows<NoSuchResourceId> {
-          subject.eventHistory("whatever")
-        }
+      test("getting state history returns an empty list") {
+        expectThat(subject.eventHistory("whatever"))
+          .isEmpty()
       }
 
       test("deleting a non-existent resource throws an exception") {
@@ -126,11 +127,6 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
       }
 
       test("it can be retrieved by id") {
-        val retrieved = subject.get<DummyResourceSpec>(resource.id)
-        expectThat(retrieved).isEqualTo(resource)
-      }
-
-      test("it can be retrieved by uid") {
         val retrieved = subject.get<DummyResourceSpec>(resource.id)
         expectThat(retrieved).isEqualTo(resource)
       }
@@ -174,7 +170,11 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
 
         before {
           tick()
-          subject.store(updatedResource)
+          with(subject) {
+            store(updatedResource).also {
+              appendHistory(ResourceUpdated(it, mapOf("delta" to "some-difference"), clock))
+            }
+          }
         }
 
         test("it replaces the original resource") {
@@ -186,6 +186,12 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
         test("the resource version is incremented") {
           expectThat(subject.get<DummyResourceSpec>(resource.id))
             .get(Resource<*>::version) isEqualTo 2
+        }
+
+        test("history includes events for the previous version of the resource") {
+          expectThat(subject.eventHistory(resource.id) as List<ResourceEvent>)
+            .map(ResourceEvent::version)
+            .contains(1, 2)
         }
 
         context("when deleted") {
@@ -326,9 +332,8 @@ abstract class ResourceRepositoryTests<T : ResourceRepository> : JUnit5Minutests
         }
 
         test("events for the resource are also deleted") {
-          expectThrows<NoSuchResourceException> {
-            subject.eventHistory(resource.id)
-          }
+          expectThat(subject.eventHistory(resource.id))
+            .isEmpty()
         }
 
         test("events for the resource's parent application remain") {
@@ -422,6 +427,10 @@ fun <T : List<E>, E> Assertion.Builder<T>.third(): Assertion.Builder<E> =
 
 fun <T : List<E>, E> Assertion.Builder<T>.fourth(): Assertion.Builder<E> =
   get("fourth element %s") { this[3] }
+
+@Suppress("UNCHECKED_CAST")
+inline fun <reified T> Assertion.Builder<List<*>>.areAll() : Assertion.Builder<List<T>> =
+  all { isA<T>() } as Assertion.Builder<List<T>>
 
 fun randomString(length: Int = 8) =
   randomUUID()
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
@@ -428,10 +428,6 @@ fun <T : List<E>, E> Assertion.Builder<T>.third(): Assertion.Builder<E> =
 fun <T : List<E>, E> Assertion.Builder<T>.fourth(): Assertion.Builder<E> =
   get("fourth element %s") { this[3] }
 
-@Suppress("UNCHECKED_CAST")
-inline fun <reified T> Assertion.Builder<List<*>>.areAll() : Assertion.Builder<List<T>> =
-  all { isA<T>() } as Assertion.Builder<List<T>>
-
 fun randomString(length: Int = 8) =
   randomUUID()
     .toString()
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
@@ -165,11 +165,11 @@ class SqlDeliveryConfigRepository(
         // delete events
         txn.deleteFrom(EVENT)
           .where(EVENT.SCOPE.eq(EventScope.RESOURCE))
-          .and(EVENT.REF.`in`(resourceIds))
+          .and(EVENT.REF_GEN.`in`(resourceIds))
           .execute()
         txn.deleteFrom(EVENT)
           .where(EVENT.SCOPE.eq(EventScope.APPLICATION))
-          .and(EVENT.REF.eq(application))
+          .and(EVENT.REF_GEN.eq(application))
           .execute()
         // delete pause records
         txn.deleteFrom(PAUSED)
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
@@ -165,7 +165,7 @@ class SqlDeliveryConfigRepository(
         // delete events
         txn.deleteFrom(EVENT)
           .where(EVENT.SCOPE.eq(EventScope.RESOURCE))
-          .and(EVENT.REF.`in`(resourceUids))
+          .and(EVENT.REF.`in`(resourceIds))
           .execute()
         txn.deleteFrom(EVENT)
           .where(EVENT.SCOPE.eq(EventScope.APPLICATION))
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
@@ -30,9 +30,12 @@ import com.netflix.spinnaker.keel.sql.RetryCategory.READ
 import com.netflix.spinnaker.keel.sql.RetryCategory.WRITE
 import de.huxhorn.sulky.ulid.ULID
 import org.jooq.DSLContext
+import org.jooq.Record1
+import org.jooq.Select
 import org.jooq.impl.DSL
 import org.jooq.impl.DSL.coalesce
 import org.jooq.impl.DSL.max
+import org.jooq.impl.DSL.select
 import org.jooq.impl.DSL.value
 import org.slf4j.LoggerFactory
 import java.time.Clock
@@ -206,8 +209,6 @@ open class SqlResourceRepository(
   override fun eventHistory(id: String, limit: Int): List<ResourceHistoryEvent> {
     require(limit > 0) { "limit must be a positive integer" }
 
-    val resource = get(id)
-
     return sqlRetry.withRetry(READ) {
       jooq
         .select(EVENT.JSON)
@@ -215,12 +216,12 @@ open class SqlResourceRepository(
         // look for resource events that match the resource...
         .where(
           EVENT.SCOPE.eq(EventScope.RESOURCE)
-            .and(EVENT.REF.eq(resource.uid))
+            .and(EVENT.REF.`in`(uidsForId(id)))
         )
         // ...or application events that match the application as they apply to all resources
         .or(
           EVENT.SCOPE.eq(EventScope.APPLICATION)
-            .and(EVENT.APPLICATION.eq(resource.application))
+            .and(EVENT.APPLICATION.eq(applicationForId(id)))
         )
         .orderBy(EVENT.TIMESTAMP.desc())
         .limit(limit)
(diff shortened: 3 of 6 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -162,7 +162,7 @@ class SqlDeliveryConfigRepository(
         // delete events
         txn.deleteFrom(EVENT)
           .where(EVENT.SCOPE.eq(EventScope.RESOURCE))
-          .and(EVENT.REF.`in`(resourceUids))
+          .and(EVENT.REF.`in`(resourceIds))
```
robfletcher, line 165: `EVENT.REF` now contains resource id, not uid

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -203,23 +206,22 @@ open class SqlResourceRepository(
...
-
     return sqlRetry.withRetry(READ) {
       jooq
         .select(EVENT.JSON)
         .from(EVENT)
         // look for resource events that match the resource...
         .where(
           EVENT.SCOPE.eq(EventScope.RESOURCE)
-            .and(EVENT.REF.eq(resource.uid))
+            .and(EVENT.REF.eq(id))
```
robfletcher, line 216: I no longer have to do an `in` with a sub-select here which was terrible for performance

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -229,16 +230,14 @@ open class SqlResourceRepository(
 
   // todo: add sql retries once we've rethought repository structure: https://github.com/spinnaker/keel/issues/740
   override fun appendHistory(event: ResourceEvent) {
-    // for historical reasons, we use the resource UID (not the ID) as an identifier in resource events
-    val ref = getResourceUid(event.ref, event.version)
-    doAppendHistory(event, ref)
+    doAppendHistory(event)
```
robfletcher, line 233: not any more

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -269,50 +268,39 @@ open class SqlResourceRepository(
         .insertInto(EVENT)
         .set(EVENT.UID, ULID().nextULID(event.timestamp.toEpochMilli()))
         .set(EVENT.SCOPE, event.scope)
-        .set(EVENT.REF, ref)
-        .set(EVENT.TIMESTAMP, event.timestamp)
```
robfletcher, line 273: No need to set these values any more as the columns are just derived from the JSON column

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -269,50 +268,39 @@ open class SqlResourceRepository(
...
+        .and(EVENT.REF.eq(id))
+        .execute()
     }
-
-    uids.forEach { uid ->
-      sqlRetry.withRetry(WRITE) {
-        jooq.deleteFrom(RESOURCE)
-          .where(RESOURCE.UID.eq(uid.toString()))
-          .execute()
-      }
```
robfletcher, line 298: Don't need to do this in a loop any more as events will have the resource id, not one of the uids so we can delete them all at once

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -269,50 +268,39 @@ open class SqlResourceRepository(
...
-      sqlRetry.withRetry(WRITE) {
-        jooq.deleteFrom(EVENT)
-          .where(EVENT.SCOPE.eq(EventScope.RESOURCE))
-          .and(EVENT.REF.eq(uid.toString()))
-          .execute()
-      }
```
robfletcher, line 304: Don't need to do this in a loop any more as events will have the resource id, not one of the uids so we can delete them all at once

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -269,50 +268,39 @@ open class SqlResourceRepository(
...
-        jooq.deleteFrom(DIFF_FINGERPRINT)
-          .where(DIFF_FINGERPRINT.ENTITY_ID.eq(id))
-          .execute()
-      }
-      sqlRetry.withRetry(WRITE) {
-        jooq.deleteFrom(PAUSED)
-          .where(PAUSED.SCOPE.eq(PauseScope.RESOURCE))
-          .and(PAUSED.NAME.eq(id))
-          .execute()
-      }
```
robfletcher, line 315: Don't know why these were ever in the loop since they used the resource id and not uid anyway

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,80 @@
...
+    changes:
+    - dropIndex:
+        tableName: event
+        indexName: event_uid_timestamp_idx
+    - dropIndex:
+        tableName: event
+        indexName: event_scope_ref_timestamp_idx
+    - dropIndex:
+        tableName: event
+        indexName: event_scope_application_timestamp_idx
```
robfletcher, line 14: Drop all the indices that use the columns I'm re-defining

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,80 @@
...
+    - dropColumn:
+        tableName: event
+        columnName: ref
+    - dropColumn:
+        tableName: event
+        columnName: application
+    - dropColumn:
+        tableName: event
+        columnName: timestamp
```
robfletcher, line 23: Drop all the columns I'm re-defining

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,80 @@
...
+          alter table event
+          add column ref varchar(255) generated always as (json ->> '$.ref')
+          not null
+          after scope;
+    - sql:
+        sql: |
+          alter table event
+          add column application varchar(255) generated always as (json ->> '$.application')
+          not null
+          after ref;
```
robfletcher, line 35: `ref` and `application` are strings pulled from the JSON. Actually `application` already was but it was in a non-optimal position in the table, so I thought it made sense to move it.

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,80 @@
...
+    - sql:
+        sql: |
+          update event
+          set json = json_set(
+            json,
+            '$.timestamp',
+            concat(substring(json_unquote(json->'$.timestamp'), 1, 19), '.000Z')
+          )
+          where length(json -> '$.timestamp') = 22;
```
robfletcher, line 44: We had 57 stray timestamps with no millisecond precision, so this cleans them up

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,80 @@
...
+    - sql:
+        sql: |
+          alter table event
+          add column timestamp datetime(3) generated always as (str_to_date(json_unquote(json->'$.timestamp'), '%Y-%m-%dT%T.%fZ'))
+          after application;
```
robfletcher, line 49: derive the `timestamp` column from the JSON data. I couldn't make `timestamp(3)` work because it wouldn't then let me include the column in the indices. No idea why.

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,80 @@
...
+        tableName: event
+        indexName: event_scope_application_timestamp_idx
+        columns:
+        - column:
+            name: scope
+        - column:
+            name: application
+        - column:
+            name: timestamp
+            descending: true
```
robfletcher, line 80: Re-create all the indices that use the re-defined columns

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
```diff
@@ -423,6 +428,10 @@ fun <T : List<E>, E> Assertion.Builder<T>.third(): Assertion.Builder<E> =
 fun <T : List<E>, E> Assertion.Builder<T>.fourth(): Assertion.Builder<E> =
   get("fourth element %s") { this[3] }
 
+@Suppress("UNCHECKED_CAST")
+inline fun <reified T> Assertion.Builder<List<*>>.areAll() : Assertion.Builder<List<T>> =
```
luispollo, line 432: Looks like this is not used anywhere?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
luispollo: Oh, wow. Good find/fix!

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt
robfletcher: Yep. Fixed.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -397,6 +385,12 @@ open class SqlResourceRepository(
         ?: throw IllegalStateException("Resource with id $id not found. Retrying.")
     }
 
+  private fun applicationForId(id: String): Select<Record1<String>> =
```
luispollo, line 388: Just for my education, is there an advantage to returning the JOOQ select record here instead of doing the `fetch` and returning the value?

robfletcher: yes, because then you can use it to compose a larger query. If you return the value from the function you'll always be executing this as a separate query.

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,80 @@
...
+          set json = json_set(
+            json,
+            '$.timestamp',
+            concat(substring(json_unquote(json->'$.timestamp'), 1, 19), '.000Z')
+          )
+          where length(json -> '$.timestamp') = 22;
+    - sql:
+        sql: |
+          alter table event
+          add column timestamp datetime(3) generated always as (str_to_date(json_unquote(json->'$.timestamp'), '%Y-%m-%dT%T.%fZ'))
```
luispollo, line 48:
```suggestion
          add column timestamp datetime(3) generated always as (str_to_date(json->>'$.timestamp', '%Y-%m-%dT%T.%fZ'))
```

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
luispollo: Does this have any impact in the JOOQ automatic type conversions when reading, or are we always reading from the JSON?

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,80 @@
...
+          after application;
+    - createIndex:
+        tableName: event
+        indexName: event_uid_timestamp_idx
+        columns:
+        - column:
+            name: uid
+        - column:
+            name: timestamp
+            descending: true
```
luispollo, line 58: TIL! Nice!

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
luispollo: Never mind, I see you added a new JOOQ config for `DATETIME` conversion.

luispollo (approved):
LGTM!

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
robfletcher: huh, I didn't know that syntax. Nice.

lorin:
Question before I start reviewing: Are the the database changes backwards compatible with the previous version of the code (e.g., if we deploy and roll back, will the old code work against the new database changes)?

robfletcher:
> Question before I start reviewing: Are the the database changes backwards compatible with the previous version of the code (e.g., if we deploy and roll back, will the old code work against the new database changes)?

No, because it would be trying to insert into columns that are now generated

lorin (commented):
## My understanding

### Problem 

For a resource with multiple versions, the history of events only shows the events of the latest version, not older versions.

### Implemented solution

Event history is now queried by resource id instead of uid. Since the resource id (unlike resource uid) is the same across the multiple resource versions, this ensures that we retrieve all of the events associated with a resource.

#### Database changes

All changes are to the `event` table.

The `ref` column now encodes the resource id string, not the resource's uid. This necessitates a change to a query in SqlDeliveryConfigRepository.

Drops the columns: `ref`, `application`, `timestamp` and replaces them with [generated columns](https://dev.mysql.com/doc/refman/5.7/en/alter-table-generated-columns.html) using data from the `json` column

## My feedback

To increase the safety of the rollout, I'd prefer to see the database migrations done in a backwards compatible way.

The rest of the changes look good to me.

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
luispollo: I learned it from your PR, haha! 😄 
https://github.com/spinnaker/keel/blob/c93a4ce8cea1f7106153e0853be51b6a1fa0723d/keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml#L27

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,71 @@
...
+            json,
+            '$.timestamp',
+            concat(substring(json->>'$.timestamp', 1, 19), '.000Z')
+          )
+          where length(json -> '$.timestamp') = 22;
+    - sql:
+        sql: |
+          alter table event
+          add column timestamp_gen datetime(3) generated always as (str_to_date(json->>'$.timestamp', '%Y-%m-%dT%T.%fZ'))
+          after application_gen;
```
robfletcher, line 40: I added these as new columns rather than replacing the existing ones.

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,74 @@
...
+        indexName: event_scope_application_timestamp_idx
+    - dropColumn:
+        tableName: event
+        columnName: application
+    - sql:
+        sql: |
+          alter table event
+          add column ref_gen varchar(255) generated always as (json ->> '$.ref')
+          not null
+          after scope;
```
robfletcher, line 23: I added this as a new column instead of replacing the existing one

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,74 @@
...
+            json,
+            '$.timestamp',
+            concat(substring(json->>'$.timestamp', 1, 19), '.000Z')
+          )
+          where length(json -> '$.timestamp') = 22;
+    - sql:
+        sql: |
+          alter table event
+          add column timestamp_gen datetime(3) generated always as (str_to_date(json->>'$.timestamp', '%Y-%m-%dT%T.%fZ'))
+          after application;
```
robfletcher, line 43: Another new column

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
```diff
@@ -0,0 +1,74 @@
...
+          alter table event
+          add column ref_gen varchar(255) generated always as (json ->> '$.ref')
+          not null
+          after scope;
+    - sql:
+        sql: |
+          alter table event
+          add column application varchar(255) generated always as (json ->> '$.application')
+          not null
+          after ref_gen;
```
robfletcher, line 29: This column definition isn't changing, I'm just repositioning it in the table

robfletcher:
@lorin I updated this PR to make the database changes backward compatible. Instead of replacing the `ref` and `timestamp` columns on the `event` table, I'm adding new ones. All references in code are updated to use the new columns, except where we're inserting a row to that table, in which case the existing columns need to have values set.

I'll follow this with another PR that drops the old columns and renames the new ones to the old names.

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
lorin: My only theoretical concern is how long the event table will stay locked during the migration. Given that it's a generated column being dropped and re-added afterwards, I assume mysql is smart enough that this change happens pretty quickly?

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
lorin: I have to say, I know that the rename is just for backwards compatibility, but having the `_gen` suffix communicate that this is a generated column is actually nice.

lorin (approved):
Looks good to me.  The only change I suggest is updating the summary of the PR to reflect the most recent changes in the PR itself.

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
robfletcher: I tested it out locally with a dump of prod data and it takes barely any time at all. I was pretty surprised.

keel-sql/src/main/resources/db/changelog/20210226-generate-event-columns.yml
robfletcher: Are you saying you like Hungarian Notation? ¬_¬

PR #1829, opened 2021-02-27
chore(perf): speed up resource checking after new version is approved

emjburns:
We've gotten some user reports of things feeling slow and not responsive. We've also gotten reports like "what are you waiting for??". This pr hopes to make deployment happen much faster after a new version is "approved" for an environment (pin, unpin, mark as bad, and also new versions).

This pr introduces a "recheck resources in an environment" function that clears the "last checked" time for all resources in an environment. This makes them immediately due for check. My hope is this will mean we spend almost 0 seconds in the "approved" state. (unless your application is paused, which means that versions will stay in the approved state until you un pause).

Hopefully this is a big performance-feel win for a minimal amount of effort.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
@@ -275,6 +275,29 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
         )
       }
 
+
+      test("resources in an environment can be rechecked") {
+        // note: this test needs to be here even though it's testing a resource repository function
+        // because we need a valid config and environment for the resources to exist in, and those can only
+        // be saved with both a delivery config repository and a resource repository.
+        storeResources()
+        store()
+
+        val firstCheck = resourceRepository.itemsDueForCheck(Duration.ofMinutes(2), 4)
+        val secondCheck = resourceRepository.itemsDueForCheck(Duration.ofMinutes(2), 4)
+        resourceRepository.triggerResourceRecheck("test", deliveryConfig.application)
+        val afterRecheck = resourceRepository.itemsDueForCheck(Duration.ofMinutes(2), 4)
+        val testResources = deliveryConfig.environments.find { it.name == "test" }?.resourceIds ?: emptySet()
+
+        expect {
+          that(firstCheck.size).isEqualTo(4)
+          that(secondCheck.size).isEqualTo(0)
+          that(afterRecheck.size).isEqualTo(2)
+          that(afterRecheck.first().application).isEqualTo(deliveryConfig.application)
+          that(afterRecheck.map { it.id }.toList()).containsExactlyInAnyOrder(testResources)
+        }
+      }
+
       context("updating an existing delivery config") {
         deriveFixture {
           storeArtifacts()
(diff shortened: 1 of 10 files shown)


keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -275,6 +275,29 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
...
+        val firstCheck = resourceRepository.itemsDueForCheck(Duration.ofMinutes(2), 4)
+        val secondCheck = resourceRepository.itemsDueForCheck(Duration.ofMinutes(2), 4)
+        resourceRepository.triggerResourceRecheck("test", deliveryConfig.application)
+        val afterRecheck = resourceRepository.itemsDueForCheck(Duration.ofMinutes(2), 4)
+        val testResources = deliveryConfig.environments.find { it.name == "test" }?.resourceIds ?: emptySet()
+
+        expect {
+          that(firstCheck.size).isEqualTo(4)
+          that(secondCheck.size).isEqualTo(0)
+          that(afterRecheck.size).isEqualTo(2)
```
luispollo, line 295: I'm guessing the count here is 2 because there's more than one environment in the delivery config and you're only resetting "test", is that it? Might be worth adding a comment.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
```diff
@@ -357,6 +359,27 @@ open class SqlResourceRepository(
     }
   }
 
+  override fun triggerResourceRecheck(environmentName: String, application: String) {
+    val resourceUids = sqlRetry.withRetry(READ) {
```
luispollo, line 363: I think ideally you'd run the select and update in a transaction so that you don't run the (small) risk of the resources changing under you in between reading and updating.

luispollo (approved):
Great idea! Thanks for doing this. Left you a couple of suggestions, not blocking.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
emjburns: Yep, there are two envs with two resources each. Do comments on tests help you a lot? (trying to gauge when it is worth it to add comments vs when it just makes the tests cluttered. I'll add here but I'm curious your general thoughts about when you yourself would add comments to tests)

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepository.kt
emjburns: ugh you're right, good point. I'll update this

emjburns:
I checked this out locally and it seemed to speed things up.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
luispollo: In cases like these, when it's not obvious where certain values are coming from, or around logic that might not be self-explanatory, yes, I tend to add a comment. Actually I don't think of test code that much differently from "main" code. It's code that we have to maintain, and we actually have more test code than application code, so in these specific cases I find it worth the trouble/noise. 🤷‍♂️ 

Example from my open PR:
https://github.com/spinnaker/keel/blob/99cc244e387cc64165aac061714ba4d728fa4ab3/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt#L494-L495

https://github.com/spinnaker/keel/blob/99cc244e387cc64165aac061714ba4d728fa4ab3/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt#L746

PR #1833, opened 2021-03-01
feat(reports): Add fields to ApplicationSummary in support of reporting

luispollo:
Adds the following fields to `ApplicationSummary`, in support of reporting internally at Netflix:
- `createdAt`: the `Instant` at which the corresponding `DeliveryConfig` was created. This field is derived from `delivery_config.uid` (which is a ULID).
- `resourceCount`: count of resources in the `DeliveryConfig` at the time of querying.

### DIFF
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
@@ -18,6 +18,7 @@ import com.netflix.spinnaker.keel.api.constraints.ConstraintState
 import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus
 import com.netflix.spinnaker.keel.api.plugins.kind
 import com.netflix.spinnaker.keel.artifacts.DebianArtifact
+import com.netflix.spinnaker.keel.core.api.ApplicationSummary
 import com.netflix.spinnaker.keel.core.api.DependsOnConstraint
 import com.netflix.spinnaker.keel.core.api.ManualJudgementConstraint
 import com.netflix.spinnaker.keel.events.ApplicationActuationPaused
@@ -47,6 +48,7 @@ import strikt.assertions.isSuccess
 import strikt.assertions.map
 import strikt.assertions.none
 import java.time.Duration
+import java.time.Instant
 
 abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : ResourceRepository, A : ArtifactRepository, P : PausedRepository> :
   JUnit5Minutests {
@@ -109,11 +111,11 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
       repository.getByApplication(deliveryConfig.application)
     }
 
-    fun store() {
+    fun store(deliveryConfig: DeliveryConfig = this.deliveryConfig) {
       repository.store(deliveryConfig)
     }
 
-    fun storeResources() {
+    fun storeResources(deliveryConfig: DeliveryConfig = this.deliveryConfig) {
       deliveryConfig.environments.flatMap { it.resources }.forEach {
         resourceRepository.store(it)
       }
@@ -487,10 +489,9 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
 
         test("can retrieve the manifest for the resources") {
           val resource = deliveryConfig.resources.random()
-
-          getDeliveryConfig(resource)
-            .isSuccess()
-            .isEqualTo(deliveryConfig)
+          val persistedDeliveryConfig = repository.deliveryConfigFor(resource.id)
+          expectThat(persistedDeliveryConfig)
+            .isEqualTo(deliveryConfig.copy(metadata = persistedDeliveryConfig.metadata))
         }
       }
 
@@ -700,6 +701,66 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
           expectThat(repository.environmentNotifications(deliveryConfig.name, "staging"))
             .isEmpty()
         }
+      }
+    }
+
+    context("multiple delivery configs exist") {
+      before {
+        (1..5).forEach { index ->
+          val resources = setOf(
+            resource(
+              application = "${deliveryConfig.application}$index",
+              kind = parseKind("ec2/cluster@v1")
+            ),
+            resource(
+              application = "${deliveryConfig.application}$index",
+              kind = parseKind("ec2/security-group@v1")
+            )
+          )
+          val deliveryConfig = deliveryConfig.copy(
+            application = "${deliveryConfig.application}$index",
+            name = "${deliveryConfig.name}$index",
+            artifacts = setOf(
+              artifact, artifactFromBranch
+            ),
+            environments = setOf(
+              Environment(
+                name = "test",
+                resources = resources
+              ),
+              Environment(
+                name = "staging",
+                resources = resources
+              )
+            )
+          )
+          store(deliveryConfig)
+          storeResources(deliveryConfig)
+        }
+      }
+
+      test("application summaries can be retrieved successfully") {
+        val persistedDeliveryConfigs = (1..5).map { index ->
+          repository.get("${deliveryConfig.name}$index")
+        }
+
+        expectThat(repository.getApplicationSummaries())
+          .hasSize(5)
+          .containsExactlyInAnyOrder(
+            (0..4).map { index ->
+              with (persistedDeliveryConfigs[index]) {
+                ApplicationSummary(
+                  deliveryConfigName = name,
+                  application = application,
+                  serviceAccount = serviceAccount,
+                  apiVersion = apiVersion,
+                  createdAt = metadata["createdAt"] as Instant,
+                  resourceCount = 4, // 2 from test, 2 from staging
+                  isPaused = false
+                )
+              }
+            }
+          )
       }
     }
   }
--- a/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
+++ b/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
@@ -36,6 +36,7 @@ import strikt.api.expectThat
 import strikt.api.expectThrows
 import strikt.assertions.contains
 import strikt.assertions.containsExactlyInAnyOrder
+import strikt.assertions.first
 import strikt.assertions.hasSize
 import strikt.assertions.isA
 import strikt.assertions.isEmpty
@@ -111,11 +112,11 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
       repository.getByApplication(deliveryConfig.application)
     }
 
-    fun store(deliveryConfig: DeliveryConfig = this.deliveryConfig) {
+    fun store() {
       repository.store(deliveryConfig)
     }
 
-    fun storeResources(deliveryConfig: DeliveryConfig = this.deliveryConfig) {
+    fun storeResources() {
       deliveryConfig.environments.flatMap { it.resources }.forEach {
         resourceRepository.store(it)
       }
@@ -276,7 +277,6 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
           )
         )
       }
-
 
       test("resources in an environment can be rechecked") {
         // note: this test needs to be here even though it's testing a resource repository function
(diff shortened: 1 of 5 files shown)


keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
```diff
@@ -133,7 +140,7 @@ internal class AdminControllerTests
       fixture { emptyList() }
 
       before {
-        every { adminService.getManagedApplications() } returns this
+        every { adminService.getManagedApplications() }.returns(this)
```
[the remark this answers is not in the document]
robfletcher: It's legit either way (`returns` is an `infix fun`) but yeah, why the change?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
```diff
@@ -17,6 +17,7 @@ import org.springframework.test.web.servlet.MockMvc
 import org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get
 import org.springframework.test.web.servlet.result.MockMvcResultMatchers.content
 import org.springframework.test.web.servlet.result.MockMvcResultMatchers.status
+import java.time.Instant
```
emjburns, line 20: If you're mocking all the data, what value do you expect to get out of these tests? I feel like we don't need to test our object to json translation for every controller. Are you explicitly cutting out data from the kotlin object and testing that?

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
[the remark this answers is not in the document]
luispollo: I got a compiler warning (error?)... 🤷‍♂️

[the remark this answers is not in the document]
luispollo: image

keel-web/src/test/kotlin/com/netflix/spinnaker/keel/rest/AdminControllerTests.kt
luispollo: Yeah, good question... I landed here while looking for repository tests that read the data, which I wanted to extend to make sure the parsing of the ULID is working as expected, but there are none as far as I can see, so I ended up just adding this here which, you're right, doesn't test anything useful, instead of writing the repository tests. In other words, I did something stupid. 😆  Will add a repository test!

emjburns: okay cool, haha. I generally don't find a lot of value in controller tests (and find them hard to update) so I try to question them every chance I get. If you feel like this test might not be useful, feel free to delete it!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -96,7 +99,7 @@ class SqlDeliveryConfigRepository(
             name = name,
             application = application,
             serviceAccount = serviceAccount,
-            metadata = metadata ?: emptyMap()
+            metadata = (metadata ?: emptyMap()) + mapOf("createdAt" to ULID.parseULID(uid).timestampAsInstant())
```
luispollo, line 102: I decided to store this into `metadata` because I wanted to keep this PR small and refactoring `DeliveryConfig` to expose a new `createdAt` field would be a much larger change that's beyond my goal at the moment (and also not used anywhere besides tests).

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
```diff
@@ -1145,22 +1148,31 @@ class SqlDeliveryConfigRepository(
...
-        .fetch { (name, application, serviceAccount, apiVersion, paused) ->
+        .groupBy(DELIVERY_CONFIG.APPLICATION)
+        .orderBy(DELIVERY_CONFIG.APPLICATION)
+        .fetch { (uid, name, application, serviceAccount, apiVersion, resourceCount, paused) ->
           ApplicationSummary(
             deliveryConfigName = name,
             application = application,
             serviceAccount = serviceAccount,
             apiVersion = apiVersion,
+            createdAt = ULID.parseULID(uid).timestampAsInstant(),
```
luispollo, line 1174: Extracts timestamp from ULID.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -487,10 +489,9 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
 
         test("can retrieve the manifest for the resources") {
           val resource = deliveryConfig.resources.random()
-
-          getDeliveryConfig(resource)
-            .isSuccess()
-            .isEqualTo(deliveryConfig)
+          val persistedDeliveryConfig = repository.deliveryConfigFor(resource.id)
+          expectThat(persistedDeliveryConfig)
+            .isEqualTo(deliveryConfig.copy(metadata = persistedDeliveryConfig.metadata))
```
robfletcher, line 494: Is this really proving anything?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -702,5 +703,65 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
...
+                  deliveryConfigName = name,
+                  application = application,
+                  serviceAccount = serviceAccount,
+                  apiVersion = apiVersion,
+                  createdAt = metadata["createdAt"] as Instant,
+                  resourceCount = 4, // 2 from test, 2 from staging
+                  isPaused = false
+                )
+              }
+            }
```
robfletcher, line 762: Does this give you decent output if it fails? It feels like it could use boiling down to the thing you're actually trying to prove. It's not clear to me exactly what that is from looking here. How many results there are? The fact that each has a `createdAt` / `resourceCount` / `isPaused` value? Whether those values are correct? 

What value do you get from returning multiple results here? There's not any difference in behavior if there's one result or any number, no? They're all configured the same so I'm kind of confused what the point of this test is.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/deliveryconfigs/deliveryconfigs.kt
```diff
@@ -34,7 +36,7 @@ internal fun SqlStorageContext.deliveryConfigByName(name: String): DeliveryConfi
           name = name,
           application = application,
           serviceAccount = serviceAccount,
-          metadata = metadata as Map<String, Any?>? ?: emptyMap()
+          metadata = (metadata ?: emptyMap()) + mapOf("createdAt" to ULID.parseULID(uid).timestampAsInstant())
```
robfletcher, line 39: Is this to cater for existing records where `createdAt` wasn't persisted? If we're doing this on read, why bother to persist it at all?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
luispollo: This was me just basically trying to keep a check equivalent to what was there before (there was already an `isEqualTo` but the adding of `createdAt` to the metadata when reading makes the objects different only for that reason).

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/deliveryconfigs/deliveryconfigs.kt
luispollo: We're not persisting it. It's extracted from the ULID... The only reason I added this was so that I could have something to compare against in the tests, since `uid` is not exposed in `DeliveryConfig` and I didn't want to add `createdAt` as an extra field in there either.

luispollo: I could expose the method to read the UID instead in the interface if you think that's preferable.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
luispollo: My thinking was to try and verify that we can retrieve more than one `ApplicationSummary` and that it has the right properties (including `createdAt`, `resourceCount` and `isPaused`). Agreed that it shouldn't matter if it's one or 5. I'll make it a single object.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
robfletcher: so `expectThat(persistedDeliveryConfig.metadata["createdAt"]).isNotNull()`?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
luispollo: I simplified the test per the above. Let me know if it looks better now.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
luispollo: No, this test was unrelated to my change. It was just checking that you could retrieve a delivery config for a resource id. But I had to fix it after my change because the metadata wouldn't match. I changed the test a bit and added a comment, hopefully it's clearer now.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/deliveryconfigs/deliveryconfigs.kt
luispollo: P.S. I should have added this same comment here: https://github.com/spinnaker/keel/pull/1833/files#r585070191

emjburns (approved):

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -490,8 +490,12 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
         test("can retrieve the manifest for the resources") {
           val resource = deliveryConfig.resources.random()
           val persistedDeliveryConfig = repository.deliveryConfigFor(resource.id)
+            .let {
+              // disregard `createdAt` which is added when reading from the database
+              it.copy(metadata = it.metadata.filterKeys { k -> k != "createdAt" })
+            }
```
robfletcher, line 496: Isn't this what you're testing? Why do you disregard it?

luispollo: I've revisited the test per your suggestion in chat.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/DeliveryConfigRepositoryTests.kt
```diff
@@ -487,10 +489,10 @@ abstract class DeliveryConfigRepositoryTests<T : DeliveryConfigRepository, R : R
 
         test("can retrieve the manifest for the resources") {
           val resource = deliveryConfig.resources.random()
-
           getDeliveryConfig(resource)
             .isSuccess()
-            .isEqualTo(deliveryConfig)
+            .get { name }
+            .isEqualTo(deliveryConfig.name)
```
luispollo, line 495: Just because the `DeliveryConfig` object itself won't match due to the addition of `createdAt` in the `metadata` map when reading from the database.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlDeliveryConfigRepository.kt
lorin: Fun fact: you can also add a pair to a map (the `mapOf` isn't strictly necessary).

Issue #1831, opened 2021-03-01
Overlapping verification executions


lorin:
We currently don't have a mechanism for preventing overlapping executions of verifications in the same environment. For example:

1. User pushes version 1, which is deployed to staging
1. Keel starts to execute verification task V1 to verify version 1
1. User pushes version 2, which is deployed to staging
1. Keel starts to execute verification task V2 to verify verison 2

Currently, V2 can start even though V1 has not yet completed.  One problem we've seen with this is that one of the versions completes successfully but is not properly recorded.

We should probably prevent concurrent executions of the same verification. At least, this should be the default behavior.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

PR #1834, opened 2021-03-02
Delay promotion if verifications are still running

robfletcher:
If a previous deployment has verifications still running, this PR makes it so that promotion of a new version will be delayed.

Relevant to #1831

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
@@ -26,6 +26,25 @@ class VerificationRunner(
    * particular environment and artifact version.
    */
   fun runVerificationsFor(context: VerificationContext) {
+    if (context.environment.verifyWith.isEmpty()) {
+      return
+    }
+
+    verificationRepository
+      .pendingInEnvironment(context.deliveryConfig, context.environmentName)
+      .filter { it.context.version != context.version }
+      .map {
+        Triple(it.context.version, it.verification.id, it.context.latestStatus(it.verification))
+      }
+      .let {
+        if (it.isNotEmpty()) {
+          it.forEach { (version, id, status)->
+            log.debug("Previous verification {} for {} is still {}", id, version, status)
+          }
+          return
+        }
+      }
+
     with(context) {
       val statuses = environment
         .verifyWith
@@ -67,7 +86,7 @@ class VerificationRunner(
   }
 
   private val Collection<Pair<*, ConstraintStatus?>>.anyStillRunning: Boolean
-    get() = any { (_, status) -> status == PENDING }
+    get() = !none { (_, status) -> status == PENDING }
 
   private val Collection<Pair<Verification, ConstraintStatus?>>.firstOutstanding: Verification?
     get() = firstOrNull { (_, status) -> status in listOf(null, NOT_EVALUATED) }?.first
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
@@ -91,8 +91,11 @@ class VerificationRunner(
     }
   }
 
+  /**
+   * `true` if any of the statuses is [PENDING], `false` if none are or the collection is empty.
+   */
   private val Collection<Pair<*, ConstraintStatus?>>.anyStillRunning: Boolean
-    get() = !none { (_, status) -> status == PENDING }
+    get() = any { (_, status) -> status == PENDING }
 
   private val Collection<Pair<Verification, ConstraintStatus?>>.firstOutstanding: Verification?
     get() = firstOrNull { (_, status) -> status in listOf(null, NOT_EVALUATED) }?.first
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
@@ -36,6 +36,9 @@ class VerificationRunner(
       .map {
         Triple(it.context.version, it.verification.id, it.context.latestStatus(it.verification))
       }
+      .filterNot { (_, _, status) ->
+        status?.complete ?: false
+      }
       .let {
         if (it.isNotEmpty()) {
           it.forEach { (version, id, status)->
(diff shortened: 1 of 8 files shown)


robfletcher:
@lorin I know we discussed adding events from this constraint evaluator but I don't think that's the mechanism for getting info to the user. The environment summary endpoint details which constraints are failing so it should be easy enough to tie this to a "verifications are still running for a previous version" type message.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -67,7 +89,7 @@ class VerificationRunner(
   }
 
   private val Collection<Pair<*, ConstraintStatus?>>.anyStillRunning: Boolean
-    get() = any { (_, status) -> status == PENDING }
+    get() = !none { (_, status) -> status == PENDING }
```
emjburns, line 92: Does `NOT_STARTED` count as 'still running'? i thought it did in some other conditions

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -26,6 +26,28 @@ class VerificationRunner(
    * particular environment and artifact version.
    */
   fun runVerificationsFor(context: VerificationContext) {
+    if (context.environment.verifyWith.isEmpty()) {
+      return
+    }
+
```
emjburns, line 32: I am having a hard time absorbing what this chain of things is doing. Could you add a quick comment(in the filter not block, the let block, and/or the return statement in the let), or change some variable names away from `it`, to help clarify what you're doing? I feel like if I had to debug this statement I'd shake my fist at you :D

emjburns (approved):
I like this solution! Good use of an implicit constraint. Two small comments but otherwise looks good to me.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
robfletcher, line 92: Fun fact: `any` returns `false` for an empty collection which is not what we want here. "Is anything still running?" should be `true` if there are no things at all.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
robfletcher: I tidied it up a bit & added some comments

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
robfletcher: Yeah, I think so. I fixed this to check for `status?.completed`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
emjburns: This is super helpful, thank you!!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
robfletcher: In this context we're only interested in ones that are actually running -- any that are `NOT_EVALUATED` are the one we're potentially about to launch

emjburns: LOL well, who knew.

lorin (commented):
## My feedback

Overall, looks good.

The one thing I don't fully understand is what happens for verifications in the `NOT_EVALUATED` state, since this is a kind of implicit state (i.e., it doesn't get explicitly stored in the `verification_state` table, instead it is encoded by an absence of a record).

I assume the desired behavior is that we still want to allow launching a new verification task in this case. (We're basically skipping over verification of a version that we clobbered).

It would be useful to have a VerificationRunner test that explicitly evaluated that scenario.

(Marking as "Comment" instead of "Approve" just to get clarity on what happens with not evaluated).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
lorin: What's the reasoning behind treating empty collection as true instead of false? When would we even call this on an empty collection? Looking at the code, it doesn't seem like should ever happen since we return early if there are no verifications defined on an environment..

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -91,8 +91,11 @@ class VerificationRunner(
     }
   }
 
+  /**
+   * `true` if any of the statuses is [PENDING], `false` if none are or the collection is empty.
+   */
   private val Collection<Pair<*, ConstraintStatus?>>.anyStillRunning: Boolean
-    get() = !none { (_, status) -> status == PENDING }
+    get() = isNotEmpty() && any { (_, status) -> status == PENDING }
```
lorin, line 98: A wise man once noted that `any` will return false if the collection is empty, so the `isNotEmpty` isn't necessary.

https://github.com/spinnaker/keel/pull/1834#discussion_r585986054

robfletcher: Dammit. Why can't I get a simple bit of logic right. At least that explains why it didn't break a test.

robfletcher:
> The one thing I don't fully understand is what happens for verifications in the `NOT_EVALUATED` state, since this is a kind of implicit state (i.e., it doesn't get explicitly stored in the `verification_state` table, instead it is encoded by an absence of a record).
> 
> I assume the desired behavior is that we still want to allow launching a new verification task in this case. (We're basically skipping over verification of a version that we clobbered).

Yes, exactly. If a verification for a prior artifact version was in a state equivalent to `NOT_EVALUATED` it would mean artifact version _n+1_ had somehow been evaluated by `VerificationRunner` before version _n_ or so close behind it that it was almost simultaneous. I don't think that can happen in reality as we cannot execute concurrent tasks on the same resource (i.e. deploying version _n_ and _n+1_ so close together that verifications for each would be evaluated within milliseconds).

PR #1835, opened 2021-03-02
feat(api): add displayName and level to events history

ranihorev:
The goal is to remove the allow-list from the frontend, and allow backend changes without the need to update the frontend.

https://github.com/spinnaker/deck/blob/5a9768bc6db2f527a73d6b1f5fb3120c101e094b/app/scripts/modules/core/src/managed/resourceHistory/ManagedResourceHistoryModal.tsx#L34

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
@@ -32,7 +32,9 @@ abstract class ApplicationEvent(
 data class ApplicationActuationPaused(
   override val application: String,
   override val timestamp: Instant,
-  override val triggeredBy: String?
+  override val triggeredBy: String?,
+  override val level: EventLevel = EventLevel.WARNING,
+  override val displayName: String = "Application management paused",
 ) : ApplicationEvent(), ResourceHistoryEvent {
   @JsonIgnore
   override val ignoreRepeatedInHistory = true
@@ -50,7 +52,9 @@ data class ApplicationActuationPaused(
 data class ApplicationActuationResumed(
   override val application: String,
   override val triggeredBy: String?,
-  override val timestamp: Instant
+  override val timestamp: Instant,
+  override val level: EventLevel = EventLevel.INFO,
+  override val displayName: String = "Application management resumed",
 ) : ApplicationEvent(), ResourceHistoryEvent {
   @JsonIgnore override val ignoreRepeatedInHistory = true
 
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.annotation.JsonSubTypes
 import com.fasterxml.jackson.annotation.JsonTypeInfo
 import java.time.Clock
 import java.time.Instant
+import com.netflix.spinnaker.keel.events.EventLevel.WARNING
 
 @JsonTypeInfo(
   use = JsonTypeInfo.Id.NAME,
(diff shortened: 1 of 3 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -404,7 +430,9 @@ data class ResourceCheckUnresolvable(
   override val version: Int,
   override val application: String,
   override val timestamp: Instant,
-  override val message: String?
+  override val message: String?,
+  override val level: EventLevel = EventLevel.WARNING,
+  override val displayName: String = "Failed to resolved state diff",
```
luispollo, line 435:
```suggestion
  override val displayName: String = "Failed to resolve state diff",
```

luispollo (approved):
Other than a little typo, looks good to me. Please reference this PR on the corresponding `deck` PR when you get to it, for context.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
```diff
@@ -32,7 +32,9 @@ abstract class ApplicationEvent(
 data class ApplicationActuationPaused(
   override val application: String,
   override val timestamp: Instant,
-  override val triggeredBy: String?
+  override val triggeredBy: String?,
+  override val level: EventLevel = EventLevel.WARNING,
```
emjburns, line 36: nit: can you import these all so that they just show up as `WARNING`? In intellij you can option + click to show context actions, and it is something like "add static import"

emjburns: (for all the enum values)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -404,7 +430,9 @@ data class ResourceCheckUnresolvable(
   override val version: Int,
   override val application: String,
   override val timestamp: Instant,
-  override val message: String?
+  override val message: String?,
+  override val level: EventLevel = EventLevel.WARNING,
+  override val displayName: String = "Failed to resolve state diff",
```
emjburns, line 435: You're using "difference" and "diff" - do you mean to use both forms? do you think it would be better to pick just one? I'm not sure.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt
```diff
@@ -205,7 +215,9 @@ data class ResourceActuationLaunched(
   override val application: String,
   val plugin: String,
   val tasks: List<Task>,
-  override val timestamp: Instant
+  override val timestamp: Instant,
+  override val level: EventLevel = EventLevel.INFO,
+  override val displayName: String = "Updating resource state",
```
emjburns, line 220: "updating resource state" is a bit confusing for me -  it doesn't convey that we started a task. But I don't have a better suggestion, so this comment is not that actionable lol

emjburns (approved):
A couple of small things. If you're just porting the wording over from the front end then ignore my wording suggestions, it's not worth enough to change them. Also the enum import preference is something that I wish we could set our code formatter to do but I don't know how - so sorry for the formatting nit!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ApplicationEvent.kt
ranihorev: Not sure that I understand :)

ranihorev: `import com.netflix.spinnaker.keel.events.EventLevel.*;` or one by one?

emjburns: one by one

PR #1836, opened 2021-03-02
fix(db): handle verif query when artifact ref and env name are the same

lorin:
## Problem

If a delivery config contains an artifact reference and environment that both happen to have the same name (e.g., an artifact reference named `"dev"` and an environment named `"dev"`), then the [`SqlVerificationRepository.getStatesBatch`][1] method throws a SQLSyntaxErrorException.

For details, see #1821

## Implemented solution

Add field aliases to prevent the name collision.

[1]: https://github.com/spinnaker/keel/blob/d87062779779c383685a2f495c1ce2dc719dbf6f/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt#L183-L241

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
@@ -378,7 +378,10 @@ class SqlVerificationRepository(
         contexts // List<VerificationContext>
           // Creates a SELECT statement from each element of [contexts], where every column is a constant. e.g.:
           // SELECT 0, "staging", "myapp", "myapp-h123-v23.4" FROM dual
-          .mapIndexed { idx, v -> jooq.select(inline(idx), inline(v.environmentName), inline(v.artifactReference), inline(v.version)) as SelectOrderByStep<Record4<Int, String, String, String>> } // List<SelectOrderByStep<Record4<Int, String, String, String>>>
+          .mapIndexed { idx, v -> jooq.select(inline(idx).`as`(ind),
+                                              inline(v.environmentName).`as`(environmentName),
+                                              inline(v.artifactReference).`as`(artifactReference),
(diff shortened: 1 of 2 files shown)


emjburns (approved):

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -378,7 +378,10 @@ class SqlVerificationRepository(
         contexts // List<VerificationContext>
           // Creates a SELECT statement from each element of [contexts], where every column is a constant. e.g.:
           // SELECT 0, "staging", "myapp", "myapp-h123-v23.4" FROM dual
-          .mapIndexed { idx, v -> jooq.select(inline(idx), inline(v.environmentName), inline(v.artifactReference), inline(v.version)) as SelectOrderByStep<Record4<Int, String, String, String>> } // List<SelectOrderByStep<Record4<Int, String, String, String>>>
+          .mapIndexed { idx, v -> jooq.select(inline(idx).`as`(ind),
+                                              inline(v.environmentName).`as`(environmentName),
+                                              inline(v.artifactReference).`as`(artifactReference),
+                                              inline(v.version).`as`(artifactVersion)) as SelectOrderByStep<Record4<Int, String, String, String>> }
```
robfletcher, line 384: Why was this failing? Because it would generate something like `SELECT 0, "test", "test", "myapp-h123-v23.4" FROM dual`… what's wrong with that?

robfletcher (approved):
I trust you this fixes the issue, but I'm curious about it.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
lorin: It's because of what happens when it's used in a subselect. This fails:

```
select * from (select 0, 'dev', 'dev', 'myartifact-1.0.0' from dual) x;
```

I have a documented example in #1821

PR #1837, opened 2021-03-02
fix(verification): you can't have a path variable with a slash

emjburns:
I have also done this, accidentally. And it is no bueno. If the URL value might have a slash it needs to go into the request body.


luispollo:
> I have also done this, accidentally. And it is no bueno. If the URL value might have a slash it needs to go into the request body.

The caller could just URL-encode the `verificationId` in the path, no?

luispollo (approved):
This looks fine to me, but it's something that I think the client (deck, I assume?) could have handled by URL-encoding that piece of the request path.

emjburns:
I've had lots of trouble with that working, and I honestly don't think it does. I spent like 45 min investigating this for a different API (mark good) and just couldn't get it to work in any way. This is why I changed this api.

emjburns:
ya we reject the url encoded docker container.

Issue #1832, opened 2021-03-01
Race leads to environments checked for verifications too quickly


lorin:
## Incorrect behavior

We've seen cases in deployment where two nodes check the same environments for verification in less than 60 seconds. This, in combination with #1830, has led to concurrent verification executions.

## Hypothesis: phantom reads

The [SqlVerificationRepository.nextEnvironmentsForVerification][1] code should ensure that we don't check the same environment more often than once every 60 seconds (this is the [environmentVerificationMinAge][2]).

The nextEnvironmentsForVerification method contains multiple SQL queries wrapped in a transaction. The behavior looks like this:

1. select all (environment, artifact versions) records which were last verified at least 60 seconds ago, or have never been verified at all
2. for each record:
    update its last verified time to now if it was previously verified
    add a last verified row if it was never previously verified

I believe we're hitting a database consistency anomaly called [phantom reads][3]. This can happen at the MySQL isolation level that we use: ([*repeatable reads*][4]).

The problem is that the repeatable read isolation level doesn't protect you against a new record being added by a concurrent transaction: it only protects you against concurrent modifications to existing records. 

Because concurrent execution of the nextEnvironmentsForVerification method can lead to new entries in the `ENVIRONMENT_LAST_VERIFIED` table, the repeatable read isolation level isn't strong enough to protect us.

## Potential solution: locking reads

We really need the equivalent of serializable level isolation here. 

I think if we modify the select statement to add [`LOCK IN SHARE MODE`][5], that would solve the problem. (That's effectively what MySQL does if you specify *serializable* transaction level).

I believe that JOOQ supports this, but the syntax changed from MySQL 5.7 to MySQL 8.0. I'm not sure if JOOQ generates the MySQL 8.0 syntax (8.0 introduced `FOR SHARE` as alternate syntax for `LOCK IN SHARE MODE`) or the backward-compatible 5.7 syntax.

[1]: https://github.com/spinnaker/keel/blob/d87062779779c383685a2f495c1ce2dc719dbf6f/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt#L59-L127
[2]:https://github.com/spinnaker/keel/blob/ff12ef877d6251892235d2a5891ae72b94662b1b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
[3]: https://en.wikipedia.org/wiki/Isolation_(database_systems)#Phantom_reads
[...]

PR #1838, opened 2021-03-02
fix(db): add read lock to avoid phantom read

lorin:
## Summary

Use a locking read to eliminate a race condition in a database transaction related to checking verifications.

## Problem

We've seen cases where two nodes check the same environments for verification within about the same second, even though verification checks are only supposed to have for an environment once a minute.

Here's an example of timestamps from two keel instances checking the same environments:

```
02/23/2021 12:46:35.613 -0800
02/23/2021 12:46:35.869 -0800
```

### Hypothesis: phantom reads

I belive the problem is related to the database consistency level used in the [SqlVerificationRepository.nextEnvironmentsForVerification][1] check. This is a databasde anomaly known as *phantom reads*.

See #1832 for details.

## Implemented solution: locking reads

This PR adds a [`LOCK IN SHARE MODE`][2] directive to the relevant select statement to ensure that phantom reads can't happen.

## Concerns: performance impact

I don't know how this lock will impact database performance. As a precaution, I put in a feature toggle so we can turn it off quickly: `keel.verifications.db.lock.reads.enabled`

I set it to be on by default, so we can see what effect it has as soon as it rolls out.

[1]: https://github.com/spinnaker/keel/blob/d87062779779c383685a2f495c1ce2dc719dbf6f/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt#L59-L127
[2]: https://dev.mysql.com/doc/refman/5.7/en/innodb-locking-reads.html

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
@@ -20,20 +20,15 @@ import com.netflix.spinnaker.keel.resources.ResourceSpecIdentifier
 import com.netflix.spinnaker.keel.resources.SpecMigrator
 import com.netflix.spinnaker.keel.sql.RetryCategory.WRITE
 import com.netflix.spinnaker.keel.sql.deliveryconfigs.deliveryConfigByName
-import org.jooq.DSLContext
-import org.jooq.Field
-import org.jooq.Record1
-import org.jooq.Record4
-import org.jooq.Select
-import org.jooq.SelectOrderByStep
+import org.jooq.*
 import org.jooq.impl.DSL.function
-import org.jooq.Table
 import org.jooq.impl.DSL.field
 import org.jooq.impl.DSL.inline
 import org.jooq.impl.DSL.isnull
 import org.jooq.impl.DSL.name
 import org.jooq.impl.DSL.select
 import org.jooq.impl.DSL.value
+import org.springframework.core.env.Environment
 import java.time.Clock
 import java.time.Duration
 import java.time.Instant.EPOCH
@@ -45,7 +40,8 @@ class SqlVerificationRepository(
   objectMapper: ObjectMapper,
   sqlRetry: SqlRetry,
   artifactSuppliers: List<ArtifactSupplier<*, *>> = emptyList(),
-  specMigrators: List<SpecMigrator<*, *>> = emptyList()
+  specMigrators: List<SpecMigrator<*, *>> = emptyList(),
+  private val environment: Environment
 ) : SqlStorageContext(
   jooq,
   clock,
@@ -55,6 +51,9 @@ class SqlVerificationRepository(
   artifactSuppliers,
   specMigrators
 ), VerificationRepository {
+
+  private val useLockingRead : Boolean
+    get() = environment.getProperty("keel.verifications.db.lock.reads.enabled", Boolean::class.java, true)
 
   override fun nextEnvironmentsForVerification(
     minTimeSinceLastCheck: Duration,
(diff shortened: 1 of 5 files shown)


keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -20,20 +20,15 @@ import com.netflix.spinnaker.keel.resources.ResourceSpecIdentifier
 import com.netflix.spinnaker.keel.resources.SpecMigrator
 import com.netflix.spinnaker.keel.sql.RetryCategory.WRITE
 import com.netflix.spinnaker.keel.sql.deliveryconfigs.deliveryConfigByName
-import org.jooq.DSLContext
-import org.jooq.Field
-import org.jooq.Record1
-import org.jooq.Record4
-import org.jooq.Select
-import org.jooq.SelectOrderByStep
+import org.jooq.*
```
lorin, line 23: I tried and failed to configure IntelliJ to stop doing the wildcard thing, so I just gave up.

emjburns (approved):
Thanks for adding in a property to control this! Looking forward to seeing if this works.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -56,6 +52,9 @@ class SqlVerificationRepository(
   specMigrators
 ), VerificationRepository {
 
+  private val useLockingRead : Boolean
+    get() = environment.getProperty("keel.verifications.db.lock.reads.enabled", Boolean::class.java, true)
```
robfletcher, line 56: Shouldn't we inject this using `@Value` rather than using the low-level Spring API?

keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepositoryTests.kt
```diff
@@ -50,14 +53,31 @@ internal class SqlVerificationRepositoryTests :
...
+    val environment: Environment = mockk()
+
+    every {
+      environment.getProperty(any(), Boolean::class.java, capture(defaultValue))
+    } answers {
+      defaultValue.captured
+    }
+
+    return environment
+  }
```
robfletcher, line 70: … it would really simplify all this stuff as you could just use `true` / `false`

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -386,4 +386,16 @@ class SqlVerificationRepository(
...
+   *
+   * In MySQL 5.7, this is `LOCK IN SHARE MODE`
+   * See https://dev.mysql.com/doc/refman/5.7/en/innodb-locking-reads.html
+   */
+  private fun <R : Record?> SelectForUpdateStep<R>.lockInShareMode(): SelectOptionStep<R> =
+    if(useLockingRead) {
+      this.forShare()
+    } else {
+      this
+    }
```
robfletcher, line 400: This is going to lock across the cluster? What's the scope of that lock (query / table?) What happens if the instance holding the lock terminates unexpectedly?

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
lorin: I can't use a `@Value` here because (AFAIK) that doesn't let us change the value of a property at runtime via fast properties. Using the `getProperty` method is the only mechanism I know of that will enable the code to respect a fast property change.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
lorin: I communicated with @robfletcher about this out of band. This is a read-write lock, meaning that it won't block reads. The lock is scoped to the transaction, so if an instance terminates during the transaction then mysql will release the locks once it detects the connection has been torn down.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
luispollo: image

image

PR #1839, opened 2021-03-02
fix(bake): use artifact version in correlation id

emjburns:
Include the version in the correlation id we send to orca so we can bake two different versions at once. This fixes the annoying situation where you push two versions in quick succession but we bake them serially instead of in parallel.


robfletcher (approved):
I like it. That was dumb before.

PR #1840, opened 2021-03-03
refactor(export): Refactor security group export to facilitate customization by sub-classes

luispollo:
Problem: we want to customize the behavior of `SecurityGroupHandler.export` in Netflix's internal implementation to support IP Objects, but there's not a currently good place to hook in to the default implementation to achieve that.

This PR extracts part of the logic in the `export` method into a new `protected` method that can be overridden by sub-classes to transform the `SecurityGroup` objects as needed before the default implementation calculates the default and override properties of the `SecurityGroupSpec`.

The only behavior changes in this PR are:
1. Simplification of how to select the `SecurityGroupRuleDeserializer` bean.
2. Export now picks the security group with the fewest inbound rules as the "base" for the properties at the top-level of the spec, to account for how we merge lists in overrides.

### DIFF
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/DefaultSecurityGroupRuleDeserializer.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/DefaultSecurityGroupRuleDeserializer.kt
@@ -5,12 +5,10 @@ import com.netflix.spinnaker.keel.api.ec2.CrossAccountReferenceRule
 import com.netflix.spinnaker.keel.api.ec2.ReferenceRule
 import com.netflix.spinnaker.keel.api.ec2.SecurityGroupRule
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.boot.jackson.JsonComponent
 
 @JsonComponent
-//@ConditionalOnMissingBean(SecurityGroupRuleDeserializer::class)
-@ConditionalOnProperty("keel.plugins.ec2.defaultSerializers", matchIfMissing = true)
+@ConditionalOnMissingBean(name = ["securityGroupRuleDeserializer"])
 class DefaultSecurityGroupRuleDeserializer : SecurityGroupRuleDeserializer() {
   override fun identifySubType(fieldNames: Collection<String>): Class<out SecurityGroupRule> =
     when {
--- a/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
+++ b/keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
@@ -119,46 +119,20 @@ open class SecurityGroupHandler(
         .map { it.await() }
     }
 
-  override suspend fun export(exportable: Exportable): SecurityGroupSpec {
-    val summaries = exportable.regions.associateWith { region ->
-      try {
-        cloudDriverCache.securityGroupByName(
-          account = exportable.account,
-          region = region,
-          name = exportable.moniker.toString()
-        )
-      } catch (e: ResourceNotFound) {
-        null
+  override suspend fun actuationInProgress(resource: Resource<SecurityGroupSpec>): Boolean =
+    resource
+      .spec
+      .locations
+      .regions
+      .map { it.name }
+      .any { region ->
+        orcaService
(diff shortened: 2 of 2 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/DefaultSecurityGroupRuleDeserializer.kt
```diff
@@ -5,12 +5,10 @@ import com.netflix.spinnaker.keel.api.ec2.CrossAccountReferenceRule
 import com.netflix.spinnaker.keel.api.ec2.ReferenceRule
 import com.netflix.spinnaker.keel.api.ec2.SecurityGroupRule
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.boot.jackson.JsonComponent
 
 @JsonComponent
-//@ConditionalOnMissingBean(SecurityGroupRuleDeserializer::class)
-@ConditionalOnProperty("keel.plugins.ec2.defaultSerializers", matchIfMissing = true)
+@ConditionalOnMissingBean(name = ["securityGroupRuleDeserializer"])
```
luispollo, line 11: Bonus: I found a simpler way to toggle between the internal and default implementation via the bean name.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -119,46 +119,20 @@ open class SecurityGroupHandler(
...
-    val summaries = exportable.regions.associateWith { region ->
-      try {
-        cloudDriverCache.securityGroupByName(
-          account = exportable.account,
-          region = region,
-          name = exportable.moniker.toString()
-        )
-      } catch (e: ResourceNotFound) {
-        null
+  override suspend fun actuationInProgress(resource: Resource<SecurityGroupSpec>): Boolean =
```
luispollo, line 122: No change here, I just moved it so that the export-related methods sit together in the file.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -167,7 +141,7 @@ open class SecurityGroupHandler(
       )
     }
 
-    val base = securityGroups.values.first()
+    val base = securityGroups.values.minByOrNull { it.inboundRules.size }!!
```
luispollo, line 144: This makes it so that the group with the fewest inbound rules is used as the default at the top-level of the spec, with overrides having the extra rules, since there's no way for an override to remove a rule (we merge them together).

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
```diff
@@ -219,17 +193,49 @@ open class SecurityGroupHandler(
...
-      .regions
-      .map { it.name }
-      .any { region ->
-        orcaService
-          .getCorrelatedExecutions("${resource.id}:$region")
-          .isNotEmpty()
+  /**
+   * Returns a map of region names to the corresponding [SecurityGroup] objects to be used in [export].
+   */
+  protected open suspend fun getSecurityGroupsByRegion(exportable: Exportable): Map<String, SecurityGroup> {
```
luispollo, line 199: This is the new method, basically just extracting this bit of the logic out of `export` into a separate method that can be overridden by the sub-class.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
robfletcher: Is there any guarantee the smallest set is common to all though?

robfletcher (approved):
Small question about the logic of using the fewest inbound rules as the base, however that's probably very much an edge case.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/SecurityGroupHandler.kt
luispollo: No, but neither is any other random pick, and definitely not the largest. 🙂 
So I think you just identified a bug in our logic for calculating overrides... We can look into that in a separate PR.

PR #1841, opened 2021-03-03
chore(deps): Bump kork and Kotlin

luispollo:
Bumps kork to pick up https://github.com/spinnaker/kork/pull/850 and aligns Kotlin version in keel.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -17,8 +17,8 @@ buildscript {
 }
 
 plugins {
-  kotlin("jvm") version "1.4.21" apply false
-  id("org.jetbrains.kotlin.plugin.allopen") version "1.4.21" apply false
+  kotlin("jvm") version "1.4.31" apply false
+  id("org.jetbrains.kotlin.plugin.allopen") version "1.4.31" apply false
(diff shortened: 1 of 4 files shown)


build.gradle.kts
```diff
@@ -17,8 +17,8 @@ buildscript {
 }
 
 plugins {
-  kotlin("jvm") version "1.4.21" apply false
-  id("org.jetbrains.kotlin.plugin.allopen") version "1.4.21" apply false
+  kotlin("jvm") version "1.4.31" apply false
```
luispollo, line 20: @robfletcher Is there a way we could avoid specifying the version here and just pick up the version defined in kork?

lorin (approved):

build.gradle.kts
robfletcher: No

luispollo: K

robfletcher: It's annoying but it's not possible to specify Gradle plugin versions with a variable / property because they get loaded before anything else (including project properties). It's a long-standing gripe.

PR #1844, opened 2021-03-04
fix(perf): eliminate caching delay between bake and deploy

emjburns:
*Summary*
Currently, there is a delay between baking and deploying because we wait for the baked image to appear in the clouddriver cache (this is where we get all our data about what exists in the world). Bake and deploy pipelines don't have this delay because the deploy stage grabs the ami id from the bake stage.

In order to bring this lightning fast speed to keel, we need to store the ami information from the bake task we launch in order to use it right away. This allows us to deploy before the ami is seen in clouddriver's cache.

*Background knowledge*
We have an implicit constraint `ImageExistsConstraint` that prevents a debian artifact version from being approved in an environment before an image exists for that version. This is the thing that ultimately makes the delay. I modified this constraint to read from the new store of ami information.

We have an `ImageResolver` that takes the desired state (version + base ami version) and calculates the ami that matches that. I modified this to also read from the new store of ami information.

*New things*
We already monitor the bake task (to produce lifecycle events, the underlying date for the cute cake bubble that appears while a version is baking). I am using this mechanism to store ami information from _successful_ bake tasks. This data is stored in a new repository called `BakedImageRepository`.

*Resiliency*
What happens if the data is malformed, or we can't get information about the ami from the bake stage? That's fine, we just ignore this optimization and we will wait for clouddriver information. We also default to the information clouddriver has, and only use our stored information if the information is not present in the clouddriver response.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/UnknownBaseImage.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/UnknownBaseImage.kt
@@ -5,3 +5,6 @@ import com.netflix.spinnaker.kork.exceptions.SystemException
 
 class UnknownBaseImage(os: String, label: BaseLabel) :
   SystemException("Could not identify base image for os $os and label $label")
+
+class UnknownBaseImageId(id: String):
+  SystemException("Could not identify base image for id $id")
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
@@ -5,7 +5,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.convertValue
 import com.netflix.spinnaker.config.LifecycleConfig
 import com.netflix.spinnaker.keel.artifacts.BakedImage
-import com.netflix.spinnaker.keel.bakery.BaseImageCache
+import com.netflix.spinnaker.keel.clouddriver.ImageService
 import com.netflix.spinnaker.keel.core.api.DEFAULT_SERVICE_ACCOUNT
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEvent
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEventStatus
@@ -48,7 +48,7 @@ class BakeryLifecycleMonitor(
   private val orcaService: OrcaService,
   @Value("\${spinnaker.baseUrl}") private val spinnakerBaseUrl: String,
   private val bakedImageRepository: BakedImageRepository,
-  private val baseImageCache: BaseImageCache,
+  private val imageService: ImageService,
   private val objectMapper: ObjectMapper,
   private val clock: Clock
 ): LifecycleMonitor(monitorRepository, publisher, lifecycleConfig) {
@@ -92,7 +92,7 @@ class BakeryLifecycleMonitor(
    * If there's a problem parsing we log an error and move on, because we will see
    * the image in clouddriver eventually.
    */
-  fun storeBakedImage(execution: ExecutionDetailResponse) {
+  suspend fun storeBakedImage(execution: ExecutionDetailResponse) {
     // we know we launch a bake task with a stage called "bake".
     val bakeStageRaw = execution.execution.stages?.find { it["type"] == "bake" }
     if (bakeStageRaw == null) {
@@ -115,7 +115,7 @@ class BakeryLifecycleMonitor(
         vmType = details.first().vmType,
         cloudProvider = details.first().cloudProviderType,
         appVersion = details.first().`package`.substringBefore("_all.deb").replaceFirst("_", "-"),
-        baseAmiVersion = baseImageCache.getVersionByImageId(details.first().baseAmiId, details.first().region),
+        baseAmiVersion = imageService.findBaseAmiVersion(details.first().baseAmiId, details.first().region),
         timestamp = execution.endTime ?: clock.instant(),
         amiIdsByRegion = details.associate { regionDetail -> regionDetail.region to regionDetail.imageId }
       )
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
@@ -96,37 +96,37 @@ class BakeryLifecycleMonitor(
     // we know we launch a bake task with a stage called "bake".
     val bakeStageRaw = execution.execution.stages?.find { it["type"] == "bake" }
     if (bakeStageRaw == null) {
-      log.error("Trying to find baked ami information, but can't find a bake stage in execution ${execution.id}")
+      log.error("Trying to find baked ami information, but can't find a bake stage for app ${execution.application} in execution ${execution.id}")
       return
     }
     try {
       val bakeStage = objectMapper.convertValue<BakeStage>(bakeStageRaw)
       val details = bakeStage.outputs.deploymentDetails
       if (details.isEmpty()) {
-        log.error("No bake details in the bake stage in execution ${execution.id}")
+        log.error("No bake details in the bake stage for app ${execution.application} in execution ${execution.id}")
+        return
       }
 
       // use the first region present because the only difference should be in ami id and region name
       // even the base ami app version will be the same across regions
+      val detail = details.first()
       val bakedImage = BakedImage(
-        name = details.first().imageName,
-        baseLabel = details.first().baseLabel,
-        baseOs = details.first().baseOs,
-        vmType = details.first().vmType,
-        cloudProvider = details.first().cloudProviderType,
-        appVersion = details.first().`package`.substringBefore("_all.deb").replaceFirst("_", "-"),
-        baseAmiVersion = imageService.findBaseAmiVersion(details.first().baseAmiId, details.first().region),
+        name = detail.imageName,
+        baseLabel = detail.baseLabel,
+        baseOs = detail.baseOs,
+        vmType = detail.vmType,
+        cloudProvider = detail.cloudProviderType,
+        appVersion = detail.`package`.substringBefore("_all.deb").replaceFirst("_", "-"),
+        baseAmiVersion = imageService.findBaseAmiVersion(detail.baseAmiId, detail.region),
         timestamp = execution.endTime ?: clock.instant(),
         amiIdsByRegion = details.associate { regionDetail -> regionDetail.region to regionDetail.imageId }
       )
       bakedImageRepository.store(bakedImage)
     } catch (e: JsonMappingException) {
-      log.error("Error converting bake stage to kotlin object in execution ${execution.id}", e)
-      return
+      log.error("Error converting bake stage to kotlin object for app ${execution.application} in execution ${execution.id}", e)
     } catch (e: Exception) {
       // if there's an error that's fine, we will move on.
-      log.error("Error finding baked image information in execution ${execution.id}", e)
-      return
+      log.error("Error finding baked image information for app ${execution.application} in execution ${execution.id}", e)
     }
   }
 
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
@@ -1,6 +1,11 @@
 package com.netflix.spinnaker.keel.bakery.artifact
 
+import com.fasterxml.jackson.databind.JsonMappingException
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.convertValue
 import com.netflix.spinnaker.config.LifecycleConfig
+import com.netflix.spinnaker.keel.artifacts.BakedImage
+import com.netflix.spinnaker.keel.bakery.BaseImageCache
 import com.netflix.spinnaker.keel.core.api.DEFAULT_SERVICE_ACCOUNT
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEvent
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEventStatus
@@ -16,11 +21,13 @@ import com.netflix.spinnaker.keel.orca.ExecutionDetailResponse
 import com.netflix.spinnaker.keel.orca.OrcaExecutionStatus.BUFFERED
 import com.netflix.spinnaker.keel.orca.OrcaExecutionStatus.RUNNING
 import com.netflix.spinnaker.keel.orca.OrcaService
+import com.netflix.spinnaker.keel.persistence.BakedImageRepository
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Component
+import java.time.Clock
 
 /**
  * A monitor for bake tasks.
@@ -39,7 +46,11 @@ class BakeryLifecycleMonitor(
   override val publisher: ApplicationEventPublisher,
   override val lifecycleConfig: LifecycleConfig,
   private val orcaService: OrcaService,
-  @Value("\${spinnaker.baseUrl}") private val spinnakerBaseUrl: String
+  @Value("\${spinnaker.baseUrl}") private val spinnakerBaseUrl: String,
+  private val bakedImageRepository: BakedImageRepository,
+  private val baseImageCache: BaseImageCache,
+  private val objectMapper: ObjectMapper,
+  private val clock: Clock
 ): LifecycleMonitor(monitorRepository, publisher, lifecycleConfig) {
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
@@ -54,7 +65,10 @@ class BakeryLifecycleMonitor(
         when {
           execution.status == BUFFERED -> publishCorrectLink(task)
           execution.status == RUNNING -> publishRunningEvent(task)
-          execution.status.isSuccess() -> publishSucceededEvent(task)
+          execution.status.isSuccess() -> {
+            publishSucceededEvent(task)
+            storeBakedImage(execution)
+          }
           execution.status.isFailure() -> publishFailedEvent(task)
           else -> publishUnknownEvent(task, execution)
         }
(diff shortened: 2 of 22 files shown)


keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -79,44 +80,66 @@ class ImageResolver(
       deliveryConfig,
       artifact,
       environment.name
-    ) ?: throw NoImageSatisfiesConstraints(artifact.name, environment.name)
+    ) ?: throw NoArtifactVersionHasBeenApproved(artifact.name, environment.name)
```
emjburns, line 83: Throwing a more specific error here because I believe this propagates to the history view/users

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -71,6 +85,51 @@ class BakeryLifecycleMonitor(
...
+   * details to appear in the clouddriver cache (~5 minute delay every time).
+   *
+   * If there's a problem parsing we log an error and move on, because we will see
+   * the image in clouddriver eventually.
+   */
+  suspend fun storeBakedImage(execution: ExecutionDetailResponse) {
+    // we know we launch a bake task with a stage called "bake".
+    val bakeStageRaw = execution.execution.stages?.find { it["type"] == "bake" }
+    if (bakeStageRaw == null) {
+      log.error("Trying to find baked ami information, but can't find a bake stage in execution ${execution.id}")
```
luispollo, line 99: Do we have the application name in the response at least to include in this log message?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -71,6 +85,51 @@ class BakeryLifecycleMonitor(
...
+      return
+    }
+    try {
+      val bakeStage = objectMapper.convertValue<BakeStage>(bakeStageRaw)
+      val details = bakeStage.outputs.deploymentDetails
+      if (details.isEmpty()) {
+        log.error("No bake details in the bake stage in execution ${execution.id}")
+      }
```
luispollo, line 107: Missing a `return` here?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -71,6 +85,51 @@ class BakeryLifecycleMonitor(
...
+
+      // use the first region present because the only difference should be in ami id and region name
+      // even the base ami app version will be the same across regions
+      val bakedImage = BakedImage(
+        name = details.first().imageName,
```
luispollo, line 112: Since you're using `details.first` everywhere, you could replace lines 104-105 with the following to make the code a bit cleaner:
```kotlin
val detail = bakeStage.outputs.deploymentDetails.firstOrNull()
if (detail == null) {
```

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -71,6 +85,51 @@ class BakeryLifecycleMonitor(
...
+        baseLabel = details.first().baseLabel,
+        baseOs = details.first().baseOs,
+        vmType = details.first().vmType,
+        cloudProvider = details.first().cloudProviderType,
+        appVersion = details.first().`package`.substringBefore("_all.deb").replaceFirst("_", "-"),
```
luispollo, line 117: Could use utility function `parseAppVersionOrNull` here.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -71,6 +85,51 @@ class BakeryLifecycleMonitor(
...
+        baseAmiVersion = imageService.findBaseAmiVersion(details.first().baseAmiId, details.first().region),
+        timestamp = execution.endTime ?: clock.instant(),
+        amiIdsByRegion = details.associate { regionDetail -> regionDetail.region to regionDetail.imageId }
+      )
+      bakedImageRepository.store(bakedImage)
+    } catch (e: JsonMappingException) {
+      log.error("Error converting bake stage to kotlin object in execution ${execution.id}", e)
+      return
```
luispollo, line 125: Not needed.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -71,6 +85,51 @@ class BakeryLifecycleMonitor(
...
+    } catch (e: Exception) {
+      // if there's an error that's fine, we will move on.
+      log.error("Error finding baked image information in execution ${execution.id}", e)
+      return
```
luispollo, line 129: Not needed.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
```diff
@@ -41,25 +43,38 @@ class ImageExistsConstraintEvaluator(
...
-        if (it.keys.containsAll(vmOptions.regions)) {
-          log.info("Found AMIs for all desired regions for {}", version)
-        } else {
-          log.warn("Missing regions {} for {}", (vmOptions.regions - it.keys).sorted().joinToString(), version)
-          eventPublisher.publishEvent(MissingRegionsDetected(version))
-        }
-      }
-      .keys.containsAll(vmOptions.regions)
+
+    val bakedImage = bakedImageRepository.getLatestByArtfiactVerstion(version, artifact)
```
luispollo, line 61: Typo in `getLatestArtifactVersion`

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/UnknownBaseImage.kt
```diff
@@ -5,3 +5,6 @@ import com.netflix.spinnaker.kork.exceptions.SystemException
 
 class UnknownBaseImage(os: String, label: BaseLabel) :
   SystemException("Could not identify base image for os $os and label $label")
+
+class UnknownBaseImageId(id: String):
+  SystemException("Could not identify base image for id $id")
```
robfletcher, line 10: Yes! My convention of not smurf-naming exception types is catching on!

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
```diff
@@ -41,25 +43,38 @@ class ImageExistsConstraintEvaluator(
...
+
+    if (clouddriverImages.keys.containsAll(vmOptions.regions)) {
+      log.info("Found AMIs for all desired regions in clouddriver cache for {}", version)
+    } else {
+      log.warn("Missing regions {} clouddriver cache for {}", (vmOptions.regions - clouddriverImages.keys).sorted().joinToString(), version)
+      eventPublisher.publishEvent(MissingRegionsDetected(version))
+    }
+
+    if (!clouddriverImages.keys.containsAll(vmOptions.regions) && bakedImage != null) {
```
luispollo, line 70: The first part of this condition is the same as the `else` condition above, so you could nest an `if (bakedImage != null)` there instead to simplify.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
```diff
@@ -41,25 +43,38 @@ class ImageExistsConstraintEvaluator(
...
+      log.debug("Found AMIs for regions {} in baked image list for {}", bakedImage.amiIdsByRegion.keys, version)
+      // if we can see them from the bake, we know they exist, so we can approve the version.
+      return bakedImage.amiIdsByRegion.keys.containsAll(vmOptions.regions)
```
luispollo, line 73: Just a suggestion, but maybe add a helper method in `BakedImage` like `presentInAllRegions(regions)` to make these checks a bit easier to read?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
robfletcher: or use a `with` block

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -157,15 +157,30 @@ class ImageService(
...
+  suspend fun findBaseAmiVersion(baseImageId: String, region: String): String {
+    return cloudDriverService.getImage("test", region, baseImageId)
+      .lastOrNull()
+      ?.let { namedImage ->
+        findBaseAmiVersion(namedImage)
+      } ?: throw BaseAmiNotFound(baseImageId)
+  }
+
+  /**
+   * Extracts the base ami version from the tages of a named image
```
robfletcher, line 173:
```suggestion
   * Extracts the base ami version from the tags of a named image
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/BakedImageRepository.kt
```diff
@@ -0,0 +1,21 @@
...
+  /**
+   * Stores information about an image that keel successfully baked
+   */
+  fun store(image: BakedImage)
+
+  /**
+   * Retrieves the information we have about baked images for an artifact versions,
+   * or null.
+   */
+  fun getLatestByArtfiactVerstion(version: String, artifact: DebianArtifact): BakedImage?
```
robfletcher, line 20:
```suggestion
  fun getLatestByArtifactVersion(version: String, artifact: DebianArtifact): BakedImage?
```

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
luispollo: Nice!!! ❤️

keel-sql/src/main/resources/db/changelog/202010303-baked-images.yml
```diff
@@ -0,0 +1,46 @@
...
+                    nullable: false
+              - column:
+                  name: base_label
+                  type: varchar(50)
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: time_detected
+                  type: timestamp
```
robfletcher, line 35:
```suggestion
                  type: timestamp(3)
```

keel-sql/src/main/resources/db/changelog/202010303-baked-images.yml
```diff
@@ -0,0 +1,46 @@
...
+                  type: varchar(50)
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: base_label
+                  type: varchar(50)
+                  constraints:
+                    primaryKey: true
+                    nullable: false
```
robfletcher, line 32: Seems like we could generate these columns based on the `json` column since all the values in the insert are just derived from that object

robfletcher (commented):
I like the approach and this will be a huge improvement. Just found a couple of typos, and have a question about the table structure (which could be handled by a future optimization if you want to get this out ASAP).

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -79,44 +80,66 @@ class ImageResolver(
...
+      .getLatestByArtfiactVerstion(artifactVersion, artifact)
+      ?.let { bakedImage ->
+        bakedImage.amiIdsByRegion.forEach{ (region, amiId) ->
+          bakedVmImages[region] = VirtualMachineImage(
+            id = amiId,
+            appVersion = bakedImage.appVersion,
+            baseImageVersion = bakedImage.baseAmiVersion
+          )
+        }
+      }
```
luispollo, line 115: You could simplify this a bit:
```suggestion
    // we might have pre-knowledge of baked images here, so let's load that.
    val bakedVmImages = bakedImageRepository
      .getLatestByArtfiactVerstion(artifactVersion, artifact)
      ?.let { bakedImage ->
        bakedImage.amiIdsByRegion.map{ (region, amiId) ->
          VirtualMachineImage(
            id = amiId,
            appVersion = bakedImage.appVersion,
            baseImageVersion = bakedImage.baseAmiVersion
          )
      }
      ?: emptyList()
```

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlBakedImageRepository.kt
```diff
@@ -0,0 +1,54 @@
...
+        .execute()
+    }
+  }
+
+  override fun getLatestByArtfiactVerstion(version: String, artifact: DebianArtifact): BakedImage? {
+     return sqlRetry.withRetry(READ) {
+      jooq
+        .select(BAKED_IMAGES.IMAGE)
+        .from(BAKED_IMAGES)
+        .where(BAKED_IMAGES.PACKAGE_VERSION.eq(version))
```
luispollo, line 44: The only condition here is matching on the version, so the `latest` in the function name was a bit misleading, I thought. Is there ever a case where we may have a dupe with a different timestamp, in which case we should sort by that and really pick the latest? Otherwise, I'd suggest dropping the `Latest` from the name.

luispollo (commented):
This looks great, @emjburns! Great job digging and optimizing! I found a couple things that I think need correction, pointed out in inline comments. Happy to look over it again later.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
luispollo: Also, same comment on including the app name for context in the log.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
emjburns: we do! i'll add it to all them. good call

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
emjburns: yuuuup

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
emjburns: good call I'll do that

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
emjburns: nooooo i'm so bad at spelling and then intellij propagates it everywhere

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/UnknownBaseImage.kt
emjburns: What is smurf naming??

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
emjburns: I was totally trying to get this to work, but I couldn't. then I forgot about it. the snippet you created is a list, but I need a map. This prompted me to figure it out using `bakedImage.amiIdsByRegion.mapValues {`. Thanks for the prompt!

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlBakedImageRepository.kt
emjburns: Oh good point. There won't be a duplicate, so I'll drop latest from the name so it's more clear.

keel-sql/src/main/resources/db/changelog/202010303-baked-images.yml
emjburns: Done!

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
```diff
@@ -41,25 +43,38 @@ class ImageExistsConstraintEvaluator(
...
-          log.warn("Missing regions {} for {}", (vmOptions.regions - it.keys).sorted().joinToString(), version)
-          eventPublisher.publishEvent(MissingRegionsDetected(version))
-        }
+
+    val bakedImage = bakedImageRepository.getByArtifactVersion(version, artifact)
+
+    if (clouddriverImages.keys.containsAll(vmOptions.regions)) {
+      log.info("Found AMIs for all desired regions in clouddriver cache for {}", version)
+    } else {
+      log.warn("Missing regions {} clouddriver cache for {}", (vmOptions.regions - clouddriverImages.keys).sorted().joinToString(), version)
```
luispollo, line 66:
```suggestion
      log.warn("Missing regions {} in clouddriver cache for {}", (vmOptions.regions - clouddriverImages.keys).sorted().joinToString(), version)
```

keel-sql/src/main/resources/db/changelog/202010303-baked-images.yml
```diff
@@ -22,32 +27,27 @@ databaseChangeLog:
...
             sql: |
               alter table baked_images
-              add column base_os varchar(150) generated always as (image ->> '$.baseOs')
+              add column base_os varchar(50) generated always as (image ->> '$.baseOs')
               not null
               after base_label;
-        - sql:
-            sql: |
-              alter table baked_images
-              add column time_detected datetime(3) generated always as (str_to_date(image->>'$.timestamp', '%Y-%m-%dT%T.%fZ'))
```
luispollo: Did you maybe have left-over data in your local database generated when you only had `timestamp` (smaller precision), which would break the `datetime(3)` here? Just a guess, but I remember @robfletcher had to add the missing precision in the migration because we had a few records like in the DB.

luispollo (approved):
LGTM! Left you one last comment on the generated datetime column, otherwise looks great!

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/UnknownBaseImage.kt
lorin: My guess is that this refers to having `Exception` at the end of an exception class name (e.g., `ThisIsBrokenException`) vs just calling it `ThisIsBroken`.

keel-sql/src/main/resources/db/changelog/202010303-baked-images.yml
emjburns: No, I had to wipe my local database everytime I changed the sql. I didn't feel like it was worth figuring out why I couldn't get this fancy bit to work, discussed with rob offline.

PR #1845, opened 2021-03-04
fix(tracing): Ensure clean MDC context at top coroutine scopes

### LINKED ISSUE #1122: X-SPINNAKER-RESOURCE-ID correlates to wrong resource in logs
Recently, @emjburns reported the following:
> I’m filtering on `x_spinnaker_resource_id="titus:cluster:titustestvpc:mddemotitus-prestaging"`, and I see log message like `Latest image for keeldemo is Image(baseAmiVersion=nflx-base-5.538.0-h1704.e0fadae, appVersion=keeldemo-0.0.1~dev.131-h128.4478aa1, regions=[us-east-1, us-west-2])`.

It looks like there might something wrong with our passing of the resource ID across coroutines and threads with MDC.

luispollo:
Follow-up to #1841. Following the discussion and recommendations in https://github.com/Kotlin/kotlinx.coroutines/issues/985, this PR adds an empty `MDCContext` to all top-level coroutine scopes where `withTracingContext` is used, in an attempt to remove left-over `X-SPINNAKER-RESOURCE-ID`s when switching threads.

Hopefully fixes #1122.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/logging/TracingSupport.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/logging/TracingSupport.kt
@@ -15,7 +15,7 @@ class TracingSupport {
   companion object {
     const val X_SPINNAKER_RESOURCE_ID = "X-SPINNAKER-RESOURCE-ID"
 
-    val clearMDC: MDCContext = MDCContext(emptyMap())
+    val blankMDC: MDCContext = MDCContext(emptyMap())
 
     suspend fun <T : ResourceSpec, R> withTracingContext(
       resource: Resource<T>,
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/logging/TracingSupport.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/logging/TracingSupport.kt
@@ -15,6 +15,8 @@ class TracingSupport {
   companion object {
     const val X_SPINNAKER_RESOURCE_ID = "X-SPINNAKER-RESOURCE-ID"
 
+    val clearMDC: MDCContext = MDCContext(emptyMap())
+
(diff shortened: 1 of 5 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/logging/TracingSupport.kt
```diff
@@ -15,6 +15,8 @@ class TracingSupport {
   companion object {
     const val X_SPINNAKER_RESOURCE_ID = "X-SPINNAKER-RESOURCE-ID"
 
+    val clearMDC: MDCContext = MDCContext(emptyMap())
```
emjburns, line 18: I think this should be `emptyMDC` - when I read it as a parameter I thought it was a function that cleared the mdc context.

Actually, I think any wording for this - "empty", "clean", "clear" would have given me the same thought. Maybe `blankMDC`? Change if you think that is more helpful or clear, if you don't think it adds anything feel free to ignore.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -34,13 +34,13 @@ class VerificationRunner(
         }
 
       if (statuses.anyStillRunning) {
-        log.debug("Verification already running for {}", environment.name)
+        log.debug("Verification already running for environment {} of application {}", environment.name, deliveryConfig.application)
```
emjburns, line 37: ooo thank you thank you

emjburns (approved):
I hope this fixes things!!!!!!!!!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/logging/TracingSupport.kt
luispollo: Renamed to `blankMDC`.

PR #1847, opened 2021-03-04
Constraint to prevent deployment if verifications are pending

robfletcher:
Re-introducing pending verifications constraint

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
@@ -212,7 +212,7 @@ class EnvironmentConstraintRunner(
     environment: Environment
   ): Boolean =
     evaluators.all { evaluator ->
-      evaluator.canPromote(artifact, version, deliveryConfig, environment)
+      evaluator.testConstraint(artifact, version, deliveryConfig, environment)
     }
 
   /**
@@ -229,8 +229,28 @@ class EnvironmentConstraintRunner(
   ): Boolean =
     evaluators.all { evaluator ->
       !environment.hasSupportedConstraint(evaluator) ||
-        evaluator.canPromote(artifact, version, deliveryConfig, environment)
+        evaluator.testConstraint(artifact, version, deliveryConfig, environment)
     }
+
+  private fun ConstraintEvaluator<*>.testConstraint(
+    artifact: DeliveryArtifact,
+    version: String,
+    deliveryConfig: DeliveryConfig,
+    environment: Environment
+  ) =
+    canPromote(artifact, version, deliveryConfig, environment)
+      .also { result ->
+        if (!result) {
+          log.debug(
+            "{} constraint prevents promotion of {} {} to {}'s {} environment",
+            supportedType.name,
+            artifact.name,
+            version,
+            deliveryConfig.application,
+            environment.name
+          )
+        }
+  }
 
   private fun Environment.hasSupportedConstraint(constraintEvaluator: ConstraintEvaluator<*>) =
     constraints.any { it.javaClass.isAssignableFrom(constraintEvaluator.supportedType.type) }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
@@ -30,26 +30,9 @@ class VerificationRunner(
       return
     }
 
-    verificationRepository
-      .pendingInEnvironment(context.deliveryConfig, context.environmentName)
-      // only consider other versions, we'll handle verifications for the version in context later
-      .filterNot { it.context.version == context.version }
-      // get the latest status by re-evaluating each one (which will update in the database)
-      .associateWith { it.context.latestStatus(it.verification) }
-      // filter out things that have now completed (since we last checked)
-      .filterNot { (_,  status) ->
-        status?.complete ?: false
-      }
-      .let { pendingVerifications ->
-        // if we still have any pending verifications then something is still running for a previous
-        // version of the artifact -- we should wait
-        if (pendingVerifications.isNotEmpty()) {
-          pendingVerifications.forEach { (pendingVerification, status)->
-            log.debug("Previous verification {} for {} is still {}", pendingVerification.verification.id, pendingVerification.context.version, status)
-          }
-          return
-        }
-      }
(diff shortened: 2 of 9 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -44,6 +52,35 @@ class VerificationRunner(
...
+    verificationRepository
+      .pendingInEnvironment(context.deliveryConfig, context.environmentName)
+      // only consider other versions, we'll handle verifications for the version in context later
+      .filterNot { it.context.version == context.version }
+      // get the latest status by re-evaluating each one (which will update in the database)
+      .associateWith { it.context.latestStatus(it.verification) }
+      // we're only interested in any verifications that are actually running, if the last step has
+      // found they are now complete, or if they never even started, that should not stop us from
+      // launching new verifications
+      .filter { (_, status) -> status == PENDING }
```
robfletcher, line 65: This is the crucial bit I changed. Previously I was looking at any status that was incomplete, but we may have old verifications that never got launched for whatever reason. We really only care about ones that are actually running.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/PendingVerificationsConstraintEvaluatorTests.kt
```diff
@@ -0,0 +1,113 @@
...
+        )
+      )
+
+      expectCatching {
+        subject.canPromote(artifact, "v1", this, environmentWithVerification)
+      }
+        .isSuccess()
+        .isTrue()
+    }
+  }
```
robfletcher, line 89: Additional test for that change

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
```diff
@@ -229,9 +229,29 @@ class EnvironmentConstraintRunner(
...
+        if (!result) {
+          log.debug(
+            "{} constraint prevents promotion of {} {} to {}'s {} environment",
+            supportedType.name,
+            artifact.name,
+            version,
+            deliveryConfig.application,
+            environment.name
+          )
+        }
```
robfletcher, line 252: It was driving me bananas that I couldn't see _which_ constraints were blocking promotion. At least with this we can find it in logs. We really need to figure out a way to enhance the API response for environments, but I figure that would make sense as a separate PR.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
```diff
@@ -284,6 +287,48 @@ class SqlVerificationRepository(
...
+      .from(VERIFICATION_STATE)
+      .join(ENVIRONMENT)
+      .on(ENVIRONMENT.UID.eq(VERIFICATION_STATE.ENVIRONMENT_UID))
+      .and(ENVIRONMENT.NAME.eq(environmentName))
+      .join(DELIVERY_CONFIG)
+      .on(DELIVERY_CONFIG.UID.eq(ENVIRONMENT.DELIVERY_CONFIG_UID))
+      .and(DELIVERY_CONFIG.NAME.eq(deliveryConfig.name))
+      .join(DELIVERY_ARTIFACT)
+      .on(DELIVERY_ARTIFACT.UID.eq(VERIFICATION_STATE.ARTIFACT_UID))
+      .where(VERIFICATION_STATE.STATUS.eq(PENDING))
```
robfletcher, line 313: I changed the query in this repository method to only return `PENDING` verifications. We may have old ones that are `NOT_EVALUATED` and _never_ launched, but they should not block new verifications from happening, or prevent promotion of a new artifact version.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -44,6 +52,33 @@ class VerificationRunner(
...
 
+  private fun anyPendingVerificationsForPreviousVersions(context: VerificationContext) =
+    verificationRepository
+      .pendingInEnvironment(context.deliveryConfig, context.environmentName)
+      // only consider other versions, we'll handle verifications for the version in context later
+      .filterNot { it.context.version == context.version }
+      // get the latest status by re-evaluating each one (which will update in the database)
+      .associateWith { it.context.latestStatus(it.verification) }
+      // we're only concerned with anything that is still running after we have re-evaluated it
+      .filter { (_, status) -> status == PENDING }
```
robfletcher, line 63: This is a slight change as we'd only have been looking at verifications that were `PENDING` and they can't go back from `PENDING` to `NOT_EVALUATED`.

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlVerificationRepository.kt
lorin: We need to remember that `PENDING` now has explicit semantics in keel, it means "actively running". If we ever add an additional state that is also true when actively running, it could bite us if we don't remember to update this query.

Might be a good idea to abstract out the where clause into a function to make this more explicit, e.g.:

```kotlin
fun Select.whereVerificationIsRunning() ...
```

lorin (approved):
### My understanding

*Context*

Desired safety properties:

1. we never run two of the same verifications at the same time in the same environment
2. we never promote a new version to an environment while a verification is actively running.

*Problem*
keel currently doesn't enforce these two safety properties. 

*Implemented solution*
Add a new implicit constraint that prevents promotion into an environment if there's a verification in the `PENDING` state in that environment.

(For safety property 1, this PR focuses on verifications against different versions).

### My feedback

This PR is strictly better than the current implementation, so I think we should merge it.

However, I don't think it guarantees the safety properties because it doesn't completely eliminate race conditions.

*Race condition example*

If you had two processes (P1, P2), you could have something like this:

P1: VerificationRunner.anyPendingVerificationsForPreviousVerions → false
P2: launches a verification for V1 in env E
    marks as pending
P1: launches a verification for V2 in env E
    marks as pending

or:

P1: PendingVerificationsConstraintEvaluator calls verificationRepository.pendingInEnvironment() -> nothing
P1: starts to promote V2 to env E
P2: launches verification V1 in env E
    marks as pending
    

*Leases*

Given that we're only marking as pending after launching a verification, I don't think there's any way to avoid these types of races without introducing some sort of "I'm about to launch a verification" lease that we take out using transactions.

PR #1848, opened 2021-03-05
feat(export): Add support for branch-based artifacts

luispollo:
Problem: we're promoting the use of "source-driven artifacts" (i.e. artifacts filtered by branch/PR information), which avoids the need for users to worry about version strings/patterns, however we were lacking support for exporting artifacts in that format, which is useful when migrating existing apps to Managed Delivery.

This PR updates the export code for both EC2 and Titus clusters to look up the artifact information from CloudDriver and Igor, respectively, and generate an artifact with the new format. If we can't find the branch metadata for any reason, the code still falls back to the original behavior, generating artifacts with release statuses for Debian and a tag version strategy for Docker.

I recommend viewing with the "Ignore whitespace" option turned on.

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
@@ -26,7 +26,7 @@ enum class ArtifactStatus {
  * @param startsWith Match branches starting with this string.
  * @param regex A regular expression to match against (e.g. "feature-.*").
  */
-data class BranchFilterSpec(
+data class BranchFilter(
   val name: String? = null,
   val startsWith: String? = null,
   val regex: String? = null
@@ -45,11 +45,11 @@ data class BranchFilterSpec(
 /**
  * Filters for the origin of an artifact in source control.
  *
- * @param branch A [BranchFilterSpec] with branch filters.
+ * @param branch A [BranchFilter] with branch filters.
  * @param pullRequestOnly Whether to include only artifacts built from pull requests.
  */
-data class ArtifactOriginFilterSpec(
-  val branch: BranchFilterSpec? = null,
+data class ArtifactOriginFilter(
+  val branch: BranchFilter? = null,
   val pullRequestOnly: Boolean? = false
 )
 
@@ -72,7 +72,7 @@ abstract class DeliveryArtifact {
   /** A set of release statuses to filter by. Mutually exclusive with [from] filters. */
   open val statuses: Set<ArtifactStatus> = emptySet()
   /** Filters for the artifact origin in source control. */
-  open val from: ArtifactOriginFilterSpec? = null
+  open val from: ArtifactOriginFilter? = null
 
   val filteredByBranch: Boolean
     get() = from?.branch != null
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
@@ -76,7 +76,7 @@ data class PublishedArtifact(
 
   val createdAt: Instant?
     get() = (metadata["createdAt"]
-    //docker artifact createdAt time is under date field
+    // docker artifact createdAt time is under date field
       ?: metadata["date"])
       ?.let {
         when (it) {
@@ -90,6 +90,15 @@ data class PublishedArtifact(
           else -> null
         }
       }
+
+  val branch: String?
+    get() = gitMetadata?.branch ?: metadata["branch"] as? String
+
+  val commitHash: String?
+    get() = gitMetadata?.commit ?: metadata["commitId"] as? String
+
+  val buildNumber: String?
+    get() = buildMetadata?.number ?: metadata["buildNumber"] as? String
 
   fun normalized() = copy(
     type = type.toLowerCase(),
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifact.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifact.kt
@@ -2,6 +2,7 @@ package com.netflix.spinnaker.keel.artifacts
 
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactOriginFilterSpec
 import com.netflix.spinnaker.keel.api.artifacts.ArtifactStatus
+import com.netflix.spinnaker.keel.api.artifacts.BranchFilterSpec
 import com.netflix.spinnaker.keel.api.artifacts.DEBIAN
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.artifacts.SortingStrategy
@@ -14,9 +15,11 @@ data class DebianArtifact(
   override val name: String,
   override val deliveryConfigName: String? = null,
   override val reference: String = name,
-  val vmOptions: VirtualMachineOptions,
   override val statuses: Set<ArtifactStatus> = emptySet(),
-  override val from: ArtifactOriginFilterSpec? = null
+  val vmOptions: VirtualMachineOptions,
+  val branch: String? = null,
+  override val from: ArtifactOriginFilterSpec? =
+    branch?.let { ArtifactOriginFilterSpec(BranchFilterSpec(name = branch)) }
 ) : DeliveryArtifact() {
   override val type = DEBIAN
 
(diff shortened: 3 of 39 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifact.kt
```diff
@@ -16,14 +18,19 @@ data class DockerArtifact(
...
-  override val from: ArtifactOriginFilterSpec? = null
+  val branch: String? = null,
+  override val from: ArtifactOriginFilterSpec? =
+    branch?.let { ArtifactOriginFilterSpec(BranchFilterSpec(name = branch)) }
 ) : DeliveryArtifact() {
   init {
     require(name.count { it == '/' } <= 1) {
       "Docker image name has more than one slash, which is not Docker convention. Please convert to `organization/image-name` format."
     }
+    require(from == null || tagVersionStrategy == null) {
```
luispollo, line 31: This should prevent people from submitting delivery configs with the two formats mixed going forward. There are occurrences in the database that I plan to fix (could add a migration to this PR too).

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/config/EC2Config.kt
```diff
@@ -31,6 +31,8 @@ import com.netflix.spinnaker.keel.ec2.resource.BlockDeviceConfig
 import com.netflix.spinnaker.keel.ec2.resource.ClassicLoadBalancerHandler
 import com.netflix.spinnaker.keel.ec2.resource.ClusterHandler
 import com.netflix.spinnaker.keel.ec2.resource.SecurityGroupHandler
+import com.netflix.spinnaker.keel.igor.artifact.ArtifactMetadataService
```
luispollo, line 34:
```suggestion
```

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -45,11 +45,11 @@ data class BranchFilterSpec(
...
  * Filters for the origin of an artifact in source control.
  *
- * @param branch A [BranchFilterSpec] with branch filters.
+ * @param branch A [BranchFilter] with branch filters.
  * @param pullRequestOnly Whether to include only artifacts built from pull requests.
  */
-data class ArtifactOriginFilterSpec(
-  val branch: BranchFilterSpec? = null,
+data class ArtifactOriginFilter(
+  val branch: BranchFilter? = null,
```
luispollo, line 52: Just shortening these names a bit.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifact.kt
lorin: You can use `xor` instead of `||` to enforce exactly one here.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/PublishedArtifact.kt
```diff
@@ -91,6 +91,15 @@ data class PublishedArtifact(
...
 
+  val branch: String?
+    get() = gitMetadata?.branch ?: metadata["branch"] as? String
+
+  val commitHash: String?
+    get() = gitMetadata?.commit ?: metadata["commitId"] as? String
+
+  val buildNumber: String?
+    get() = buildMetadata?.number ?: metadata["buildNumber"] as? String
+
```
luispollo, line 102: I found that Igor already provided all this metadata when we first get information about the artifacts (i.e. before we call again to retrieve additional git and build metadata). These accessors expose the corresponding bits of metadata delegating to the generic `metadata` from Igor first, and then fallback to the `gitMetadata` and `buildMetadata`.

luispollo: Correction: the only thing missing was the branch, which I've added in Igor.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DebianArtifact.kt
```diff
@@ -14,9 +15,11 @@ data class DebianArtifact(
   override val name: String,
   override val deliveryConfigName: String? = null,
   override val reference: String = name,
-  val vmOptions: VirtualMachineOptions,
   override val statuses: Set<ArtifactStatus> = emptySet(),
-  override val from: ArtifactOriginFilterSpec? = null
+  val vmOptions: VirtualMachineOptions,
+  val branch: String? = null,
```
luispollo, line 20: Convenience property to indirectly set the `ArtifactOriginFilter` below without typing a bunch of stuff.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifact.kt
```diff
@@ -16,14 +17,19 @@ data class DockerArtifact(
   override val name: String,
   override val deliveryConfigName: String? = null,
   override val reference: String = name,
-  val tagVersionStrategy: TagVersionStrategy = SEMVER_TAG,
+  val tagVersionStrategy: TagVersionStrategy? = null,
   val captureGroupRegex: String? = null,
-  override val from: ArtifactOriginFilterSpec? = null
+  val branch: String? = null,
```
luispollo, line 22: Convenience property to indirectly set the `ArtifactOriginFilter` below without typing a bunch of stuff.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resource/ClusterHandler.kt
```diff
@@ -483,25 +489,48 @@ class ClusterHandler(
...
-    val artifactName = checkNotNull(base.launchConfiguration.appVersion).parseAppVersion().packageName
-
-    val status = debianArtifactParser.parseStatus(base.launchConfiguration.appVersion?.substringAfter("$artifactName-"))
-    if (status == UNKNOWN) {
-      throw ExportError("Unable to determine release status from appVersion ${base.launchConfiguration.appVersion}, you'll have to configure this artifact manually.")
+    val appVersion = checkNotNull(base.launchConfiguration.appVersion).parseAppVersion()
+    val artifactName = appVersion.packageName
+    val artifactVersion = base.launchConfiguration.appVersion!!.removePrefix("${artifactName}-")
+    val artifact = try {
+      artifactService.getArtifact(artifactName, artifactVersion, DEBIAN)
```
luispollo, line 496: Here's where we try to retrieve basic artifact information from Igor for the artifact, which we use to fill in the branch in the artifact constructor below.

keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/TitusClusterHandler.kt
```diff
@@ -283,49 +285,32 @@ class TitusClusterHandler(
...
-    }
-    return when (versioningStrategies.size) {
-      1 -> versioningStrategies.first()
-      0 -> null
-      else -> {
-        log.warn("Multiple versioning strategies apply for image, returning first")
-        versioningStrategies.first()
-      }
+    // prefer branch-based artifact spec, fallback to tag version strategy
+    return if (matchingImage.branch != null) {
```
luispollo, line 303: The `DockerImage` returned by CloudDriver already contained the branch name, so we're just checking for its presence here and using that to construct the artifact.

lorin (approved):
### My understanding

*Problem*

When a user exports an existing app to managed delivery, they can't export their artifacts to use source-driven artifacts (the new `from` scheme that specifies a branch the code comes from).

*Implemented solution*

During export of Titus/EC2 resources, if keel can determine the branch associated with the currently deployed artifact version, it will export artifacts with the `from` field populated.

*Implementation details*

The DockerArtifact and DebianArtifact constructors now take a nullable `branch` argument. If a branch is specified, then it is used to populate the `from` property of these artifacts.

EC2: In the handler's export method, call out to Igor (ArtifactService) with the artifact version information in order to retrieve branch info if it's available. If Igor doesn't have the branch information, fall back to the existing behavior

Titus: In the handler's export method, extract the branch directly from the image metadata, which is
provided by clouddriver.

### Feedback

Looks good to me!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/DockerArtifact.kt
luispollo: I remembered that! I tried it, but I think I didn't get the syntax right. How would that expression look?

luispollo: Done!

PR #1849, opened 2021-03-06
veto comment can be longer than 255 characters

robfletcher:
Deck allows any length of comment and users get an error if they enter more than 255 characters.

### DIFF
--- a/keel-sql/src/main/resources/db/databaseChangeLog.yml
+++ b/keel-sql/src/main/resources/db/databaseChangeLog.yml
@@ -245,3 +245,6 @@ databaseChangeLog:
   - include:
       file: changelog/20210305-fix-artifacts.yml
       relativeToChangelogFile: true
+  - include:
+      file: changelog/20210305-embiggen-reason-column.yml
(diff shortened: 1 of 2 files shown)


keel-sql/src/main/resources/db/databaseChangeLog.yml
```diff
@@ -245,3 +245,6 @@ databaseChangeLog:
   - include:
       file: changelog/20210305-fix-artifacts.yml
       relativeToChangelogFile: true
+  - include:
+      file: changelog/202010305-embiggen-reason-column.yml
```
lorin, line 249: Extra 0 in the year (`20201`)

lorin (commented):
Looks good (other than the comment about the file name). But I have no intuition about how long mysql will lock the table for in order to do this migration. Any idea about how fast this will be?

keel-sql/src/main/resources/db/databaseChangeLog.yml
robfletcher: d'oh copied 'n pasted

PR #1850, opened 2021-03-08
feat(constraints): freeze status of stateless constraints when artifact approved

emjburns:
**Problem**
Stateless constraint status is always evaluated on the fly. So if you have an `allowed-times` constraint, we will evaluate whether it passes at this exact moment. That's fine when we haven't approved an artifact. But, when we've approved an artifact, that constant evaluation becomes misleading if you were to look at the non-existant UI at a time when the constraint doesn't pass.

**Solution** 
I'm saving the constraint status to the database when an artifact version is approved (`EnvironmentPromotionChecker`). We only approve a version when all constraints pass. So, at that moment we can capture the state of all the stateless constraints and save it.

**Implementation details**
You can think of the constraints like this:
- Stateful constraints need to check database information to decide if they pass or fail.
- Stateless constraints do not need to check the database.

I'm taking advantage of the fact that the two kinds of constraints are evaluated differently. Since stateless constraints don't read from the database while being evaluated it's okay that we save information about the stateless constraint in the database (or "freeze it" as I like to say). This information will only be looked at when we generate the application summary information.

**Gross things that I feel bad about**
I'm sort of misusing the types of constraints. This makes it harder to understand where information is coming from. I feel kinda bad about it. But my thought is that the distinction between the types of constraints is fine at evaluation time, it's just confusing when we generate the data. And we can hide this by fixing the API. Which I'm doing by adding a new list of `constraintSummaries` that duplicates the data of `statefulConstraintSummaries` and `StatelessConstraintSummaries`. We can switch to using that and then I can remove the duplication. This new list uses the data type of the stateful summaries. For stateless constraints, I'm now representing whether they're currently passing by adding a field to the attributes.

Api examples:

```"environments": [
  {
    "name": "testing",
    "state": "pending",
    "statefulConstraints": [
      {
        "type": "manual-judgement",
        "status": "PENDING",
        "startedAt": "2021-03-08T22:25:32.029Z"
      }
    ],
    "statelessConstraints": [
      {
        "type": "allowed-times",
        "currentlyPassing": true,
        "attributes": {
          "windows": [
            {
              "days": [1,2,3,4,5],
              "hours": [9,10,11,12,13,14,15,16]
            }
          ]
        }}],
    "constraints": [
      {
        "type": "allowed-times",
        "status": "PASS",
        "startedAt": "2021-03-08T22:26:02.165Z",
        "judgedBy": "Spinnaker",
        "judgedAt": "2021-03-08T22:26:02.165Z",
        "attributes": {
          "allowedTimes": [
            {
              "days": [1,2,3,4,5],
              "hours": [16,9,10,11,12,13,14,15]
            }
          ],
          "timezone": "America/Los_Angeles",
          "currentlyPassing": true,
          "type": "allowed-times"
        }
      },
      {
        "type": "manual-judgement",
        "status": "OVERRIDE_PASS",
        "startedAt": "2021-03-08T06:25:45.503Z",
        "judgedBy": "emburns@netflix.com",
        "judgedAt": "2021-03-08T22:25:45.498Z"
      }
    ]
  }
]```

### DIFF
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ConstraintEvaluator.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ConstraintEvaluator.kt
@@ -75,19 +75,4 @@ interface ConstraintEvaluator<T : Constraint> : SpinnakerExtensionPoint {
     deliveryConfig: DeliveryConfig,
     targetEnvironment: Environment
   ): Boolean
-
-  /**
-   * @return a constraint state object that captures the current state of the constraint.
-   * This will be used for stateless constraints to construct and save a summary of the state
-   * when they are passing.
-   *
-   * If a summary of the constraint will not be shown to the user (like with an implicit constraint)
-   * this function does not need to be implemented.
-   */
-  fun generateConstraintState(
-    artifact: DeliveryArtifact,
-    version: String,
-    deliveryConfig: DeliveryConfig,
-    targetEnvironment: Environment
-  ): ConstraintState? = null
 }
--- a/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ConstraintEvaluator.kt
+++ b/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ConstraintEvaluator.kt
@@ -19,6 +19,7 @@ import com.netflix.spinnaker.keel.api.Constraint
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import com.netflix.spinnaker.keel.api.constraints.ConstraintState
 import com.netflix.spinnaker.keel.api.constraints.SupportedConstraintType
 import com.netflix.spinnaker.keel.api.support.EventPublisher
 import com.netflix.spinnaker.kork.plugins.api.internal.SpinnakerExtensionPoint
@@ -60,13 +61,13 @@ interface ConstraintEvaluator<T : Constraint> : SpinnakerExtensionPoint {
   val eventPublisher: EventPublisher
 
   /**
-   * returns true if a constraint should be run for every environment in every delivery config, without being
+   * @return true if a constraint should be run for every environment in every delivery config, without being
    * exposed to delivery config author.
    */
   fun isImplicit(): Boolean = false
 
   /**
-   * TODO: Docs
+   * @return true if the constraint passes, false otherwise.
    */
   fun canPromote(
     artifact: DeliveryArtifact,
@@ -74,4 +75,19 @@ interface ConstraintEvaluator<T : Constraint> : SpinnakerExtensionPoint {
     deliveryConfig: DeliveryConfig,
     targetEnvironment: Environment
   ): Boolean
+
+  /**
+   * @return a constraint state object that captures the current state of the constraint.
+   * This will be used for stateless constraints to construct and save a summary of the state
+   * when they are passing.
+   *
+   * If a summary of the constraint will not be shown to the user (like with an implicit constraint)
+   * this function does not need to be implemented.
+   */
+  fun generateConstraintState(
+    artifact: DeliveryArtifact,
+    version: String,
+    deliveryConfig: DeliveryConfig,
+    targetEnvironment: Environment
+  ): ConstraintState? = null
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
@@ -8,7 +8,6 @@ import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.artifacts.PublishedArtifact
 import com.netflix.spinnaker.keel.api.constraints.ConstraintState
 import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus
-import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus.PASS
 import com.netflix.spinnaker.keel.api.constraints.StatefulConstraintEvaluator
 import com.netflix.spinnaker.keel.api.constraints.StatelessConstraintEvaluator
 import com.netflix.spinnaker.keel.api.plugins.ConstraintEvaluator
@@ -185,7 +184,7 @@ class EnvironmentConstraintRunner(
     }
   }
 
-  fun generateStatelessConstraintState(
+  fun getStatelessConstraintSnapshots(
     artifact: DeliveryArtifact,
     deliveryConfig: DeliveryConfig,
     version: String,
@@ -195,7 +194,7 @@ class EnvironmentConstraintRunner(
     environment.constraints.mapNotNull { constraint ->
       constraint
         .findStatelessEvaluator()
-        ?.generateConstraintState(
+        ?.generateConstraintStateSnapshot(
           artifact = artifact,
           deliveryConfig = deliveryConfig,
           version = version,
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
@@ -1,10 +1,12 @@
 package com.netflix.spinnaker.keel.actuation
 
+import com.netflix.spinnaker.keel.api.Constraint
 import com.netflix.spinnaker.keel.api.DeliveryConfig
 import com.netflix.spinnaker.keel.api.Environment
 import com.netflix.spinnaker.keel.api.anyStateful
 import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.artifacts.PublishedArtifact
+import com.netflix.spinnaker.keel.api.constraints.ConstraintState
 import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus
 import com.netflix.spinnaker.keel.api.constraints.StatefulConstraintEvaluator
 import com.netflix.spinnaker.keel.api.plugins.ConstraintEvaluator
@@ -181,6 +183,23 @@ class EnvironmentConstraintRunner(
     }
   }
 
+  fun generateStatelessConstraintState(
+    artifact: DeliveryArtifact,
+    deliveryConfig: DeliveryConfig,
+    version: String,
+    environment: Environment
+  ): List<ConstraintState> =
+    environment.constraints.mapNotNull { constraint ->
+      constraint
+        .findStatelessEvaluator()
+        ?.generateConstraintState(
+          artifact = artifact,
+          deliveryConfig = deliveryConfig,
+          version = version,
+          targetEnvironment = environment
+        )
+    }
+
   fun checkStatelessConstraints(
     artifact: DeliveryArtifact,
     deliveryConfig: DeliveryConfig,
@@ -234,4 +253,7 @@ class EnvironmentConstraintRunner(
 
   private fun Environment.hasSupportedConstraint(constraintEvaluator: ConstraintEvaluator<*>) =
     constraints.any { it.javaClass.isAssignableFrom(constraintEvaluator.supportedType.type) }
+
+  private fun Constraint.findStatelessEvaluator(): ConstraintEvaluator<*>? =
+    statelessEvaluators.firstOrNull { javaClass.isAssignableFrom(it.supportedType.type) }
 }
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
@@ -8,7 +8,9 @@ import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
 import com.netflix.spinnaker.keel.api.artifacts.PublishedArtifact
 import com.netflix.spinnaker.keel.api.constraints.ConstraintState
 import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus
+import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus.PASS
 import com.netflix.spinnaker.keel.api.constraints.StatefulConstraintEvaluator
+import com.netflix.spinnaker.keel.api.constraints.StatelessConstraintEvaluator
 import com.netflix.spinnaker.keel.api.plugins.ConstraintEvaluator
 import com.netflix.spinnaker.keel.persistence.KeelRepository
 import org.slf4j.LoggerFactory
@@ -187,7 +189,8 @@ class EnvironmentConstraintRunner(
     artifact: DeliveryArtifact,
     deliveryConfig: DeliveryConfig,
     version: String,
-    environment: Environment
+    environment: Environment,
+    currentStatus: ConstraintStatus?
   ): List<ConstraintState> =
     environment.constraints.mapNotNull { constraint ->
       constraint
@@ -196,7 +199,8 @@ class EnvironmentConstraintRunner(
           artifact = artifact,
           deliveryConfig = deliveryConfig,
           version = version,
-          targetEnvironment = environment
+          targetEnvironment = environment,
+          currentStatus = currentStatus
         )
     }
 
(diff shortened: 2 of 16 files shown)

### CODE REVIEW

File: keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt [hunk 1]
```diff
@@ -96,6 +98,9 @@ class KeelConfigurationFinalizer(
         log.info("Registering Constraint Attributes sub-type {}: {}", attributeType.name, attributeType.type.simpleName)
         extensionRegistry.register(attributeType.type, attributeType.name)
       }
+
+    extensionRegistry.register(DependsOnConstraintAttributes::class.java, "depends-on")
```
lorin, line 102: I assume this is to support deserializing these objects from the database?

lorin (approved):
My understanding

*Problem*

Currently, the keel API shows stateless information about an environment version based on the result of evaluating the constraint at the time of the API call.

The problem with this is, for versions that have already been deployed, what users really care about was the state of the constraints when keel do the evaluation. For example, It's confusing if a user is looking at a version that had been successfully deployed, but keel says the allowed-times constraint is not satisfied. 

*Implemented solution*

For certain stateless constraints (allowed-times, depends-on), keel now "freezes"  their state to the database when a version is approved.

The main changes are:
* EnvironmentPromotionChecker - writes out to the database the state of stateless constraints whose evaluators implement the generateConstraintState method.
* ApplicationService - handle frozen stateless constraints when queried for info about constraints

*My feedback*

I had a few minor comments, but otherwise looks good to me.

keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
emjburns: yeah, unfortunately. The way to 'automatically' do this is right above, but they need to be a part of a stateless evaluator. I felt this was easier but grosser.

luispollo:
> But, when we've approved an artifact, that constant evaluation becomes misleading if you were to look at the non-existant UI at a time when the constraint doesn't pass.

Can you elaborate on that a bit, please? How would this manifest in the non-existent UI that would be misleading? 🙂

luispollo:
@emjburns never mind, @lorin's articulation of the problem helped!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ConstraintEvaluator.kt
```diff
@@ -60,18 +61,33 @@ interface ConstraintEvaluator<T : Constraint> : SpinnakerExtensionPoint {
...
   fun canPromote(
     artifact: DeliveryArtifact,
     version: String,
     deliveryConfig: DeliveryConfig,
     targetEnvironment: Environment
   ): Boolean
+
+  /**
+   * @return a constraint state object that captures the current state of the constraint.
+   * This will be used for stateless constraints to construct and save a summary of the state
```
luispollo, line 81: Since this only applies to stateless constraints, would it make sense to create a new `StatelessConstraintEvaluator` base class to host this function, and refactor the implementations to extend it?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -124,38 +124,71 @@ class EnvironmentPromotionChecker(
     deliveryConfig: DeliveryConfig,
     artifact: DeliveryArtifact,
     version: String,
-    targetEnvironment: String
+    targetEnvironment: Environment
```
luispollo, line 127: Every time I see us replace a string parameter with the actual object it warms my heart. ❤️ 🙂

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -124,38 +124,71 @@ class EnvironmentPromotionChecker(
...
+
       // recheck all resources in an environment, so action can be taken right away
-      repository.triggerResourceRecheck(targetEnvironment, deliveryConfig.application)
+      repository.triggerResourceRecheck(targetEnvironment.name, deliveryConfig.application)
     }
   }
 
+  /**
+   * Save the passing status of all stateless constraints when a version is approved so that
+   * their status stays the same forever. We don't want them to be evaluated anymore.
```
luispollo, line 164: It occurred to me reading this that we have an API (mostly for internal use, I think?) that clears the constraint states so they can be re-evaluated. Do we have to worry about this new state in that API?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
```diff
@@ -181,6 +183,23 @@ class EnvironmentConstraintRunner(
     }
   }
 
+  fun generateStatelessConstraintState(
```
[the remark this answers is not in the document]
luispollo: P.S. If you're OK with that, then maybe you could also rename `freezeStatelessConstraintStatus` to `snapshotStatelessConstraints`?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
```diff
@@ -124,38 +124,71 @@ class EnvironmentPromotionChecker(
...
+    ).forEach { constraintState ->
+      log.debug("Storing final constraint state for {} constraint for artifact {} {} version {} for {} environment {} in {}",
+        constraintState.type,
+        artifact.name,
+        artifact.type,
+        version,
+        deliveryConfig.name,
+        environment.name,
+        deliveryConfig.application
+      )
```
luispollo, line 186:
```suggestion
      log.debug("Storing final state for constraint {} for {} version {} for {} environment {} in {}",
        constraintState.type,
        artifact,
        version,
        deliveryConfig.name,
        environment.name,
        deliveryConfig.application
      )
```
`DeliveryArtifact.toString()` already prints a nice representation of the artifact including the type.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
```diff
@@ -51,8 +51,9 @@ data class ArtifactSummaryInEnvironment(
   val replacedBy: String? = null,
   val pinned: ActionMetadata? = null,
   val vetoed: ActionMetadata? = null,
-  val statefulConstraints: List<StatefulConstraintSummary> = emptyList(),
-  val statelessConstraints: List<StatelessConstraintSummary> = emptyList(),
+  val statefulConstraints: List<StatefulConstraintSummary> = emptyList(), //todo eb: remove in favor of [constraints]
+  val statelessConstraints: List<StatelessConstraintSummary> = emptyList(), //todo eb: remove in favor of [constraints]
+  val constraints: List<StatefulConstraintSummary> = emptyList(),
```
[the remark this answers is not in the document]
luispollo: I agree. I was actually thinking that maybe we ought to decouple the base class names from any notion of persisted state since they all have persisted state now. Something that conveys that one type is always re-evaluated on-the-fly, while the other uses stored information to determine the result, but I couldn't come up with any names today (tired 🧠).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -103,7 +107,13 @@ class ApplicationService(
 
   fun deleteConfigByApp(application: String) = repository.deleteDeliveryConfigByApplication(application)
 
-  fun getConstraintStatesFor(application: String) = repository.constraintStateFor(application)
+  fun getConstraintStatesFor(application: String): List<ConstraintState> =
+    repository
+      .constraintStateFor(application)
+      .filterNot {
+        // remove frozen "stateless" constraints from this list
+        listOf("allowed-times", "depends-on").contains(it.type)
```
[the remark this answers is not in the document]
luispollo: If we create a new base class for the evaluators like I suggested in another comment, then we could use that to filter here without hard-coding. We could do the same to the actual constraint classes as well (make them extend `StatelessConstraint`).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -344,9 +354,11 @@ class ApplicationService(
                   status
                 )
                   ?.also {
+                    //todo eb: use only the last one
```
luispollo, line 357: I had to read this a couple of times to understand what it meant. Maybe something like "consolidate constraint summaries into a single list"?

keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
luispollo: Couldn't you just remove the filter above and check if `attributeType` is not null?

luispollo (commented):
Overall, looks good to me, but I think it'd be best if you could address some of the comments on hard-coding things. I left you a couple of suggestions that hopefully help with that.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/core/api/ArtifactSummary.kt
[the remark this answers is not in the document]
emjburns: Ya, I was going to do clean up once this rolls out and Rani starts using it, because I'll need to go back and clean up the API anyways.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/AllowedTimesConstraintEvaluator.kt
```diff
@@ -254,6 +256,32 @@ class AllowedTimesConstraintEvaluator(
...
+
+    val tz: String = constraint.tz ?: dynamicConfigService.getConfig(String::class.java, "default.time-zone", "America/Los_Angeles")
+
+    return ConstraintState(
+      deliveryConfigName = deliveryConfig.name,
+      environmentName = targetEnvironment.name,
+      artifactVersion = version,
+      artifactReference = artifact.reference,
+      type = CONSTRAINT_NAME,
+      status = PASS,
```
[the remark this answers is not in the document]
emjburns: @lorin What if I were to pass in the status I wanted? I want to avoid user confusion in the case of a race condition (like the artifact is approved at 9:59:59 and then we call the allowed times at 10:00:01 and it is now false). The change in assumption would be that the calling function should know what happened.

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ConstraintEvaluator.kt
emjburns: Hmm ya, it might. Let me test that out.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentConstraintRunner.kt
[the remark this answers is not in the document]
emjburns: Snapshot might be a better word - that makes more sense to me also. Let me make some updates!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
emjburns: No, I don't think we have to worry about that because this is just used for generating the API data. But good thought!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/EnvironmentPromotionChecker.kt
emjburns: oh nice

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/AllowedTimesConstraintEvaluator.kt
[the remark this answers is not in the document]
lorin: @emjburns Yeah, I think passing it in would be good.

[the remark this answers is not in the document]
emjburns: ok @lorin I've made this change, I'd love for you to take another look.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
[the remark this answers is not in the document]
emjburns: Done. I did this by adding a statelessConstraintEvaluator and just scraping the attr types of those. Worked nicely I think!

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/plugins/ConstraintEvaluator.kt
emjburns: This was sooooo much better, great suggestion @luispollo!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/DependsOnConstraintEvaluator.kt
```diff
@@ -19,18 +22,20 @@ import com.netflix.spinnaker.keel.core.api.DependsOnConstraint
...
 @Component
 class DependsOnConstraintEvaluator(
   private val artifactRepository: ArtifactRepository,
   private val verificationRepository: VerificationRepository,
-  override val eventPublisher: EventPublisher
-) : ConstraintEvaluator<DependsOnConstraint> {
-
+  override val eventPublisher: EventPublisher,
+  private val clock: Clock
+) : StatelessConstraintEvaluator<DependsOnConstraint, DependsOnConstraintAttributes>() {
```
luispollo, line 33: Thought for a future refactor: instead of having 2 generic types, which are intimately related, we could have a `attributeType` field in `Constraint` which returns that type. Then you'd only need the first generic type here. Same for `StatefulConstraintEvaluator` too.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -563,6 +584,61 @@ class ApplicationService(
...
+      // some/all stateless constraints might already have summary info, so filter those out.
+      constraint !is StatefulConstraint && persistedStates.none { it.type == constraint.type }
+    }.mapNotNull { constraint ->
+      statelessEvaluators.find { evaluator ->
+        evaluator.supportedType.name == constraint.type
+      }?.let { evaluator ->
+        val passes = evaluator.canPromote(artifact, version = version, deliveryConfig = deliveryConfig, targetEnvironment = environment)
+        StatefulConstraintSummary(
+          type = constraint.type,
+          status = if (passes) PASS else ConstraintStatus.PENDING,
```
luispollo, line 624: nit: import `ConstraintStatus.PENDING`

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -563,6 +584,61 @@ class ApplicationService(
...
+
+    val statelessConstraintsStates: List<StatefulConstraintSummary> = environment.constraints.filter { constraint ->
+      // some/all stateless constraints might already have summary info, so filter those out.
+      constraint !is StatefulConstraint && persistedStates.none { it.type == constraint.type }
+    }.mapNotNull { constraint ->
+      statelessEvaluators.find { evaluator ->
+        evaluator.supportedType.name == constraint.type
+      }?.let { evaluator ->
+        val passes = evaluator.canPromote(artifact, version = version, deliveryConfig = deliveryConfig, targetEnvironment = environment)
+        StatefulConstraintSummary(
```
luispollo, line 622: Am I correct in understanding that we only need to build these summary objects here for backwards compatibility with stateless constraints that have already passed but won't have state stored in the DB before this PR?

keel-web/src/main/kotlin/com/netflix/spinnaker/config/KeelConfigurationFinalizer.kt
```diff
@@ -96,6 +97,14 @@ class KeelConfigurationFinalizer(
         log.info("Registering Constraint Attributes sub-type {}: {}", attributeType.name, attributeType.type.simpleName)
         extensionRegistry.register(attributeType.type, attributeType.name)
       }
+
+    constraintEvaluators
+      .filterIsInstance<StatelessConstraintEvaluator<*, *>>()
+      .map { it.attributeType }
+      .forEach { attributeType ->
```
luispollo, line 104:
```suggestion
      .forEach {
        val attributeType = it.attributeType
```

luispollo (approved):
LGTM!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
emjburns: Yep!

PR #1851, opened 2021-03-09
back out pending verification constraint as we're going to try a different approach

robfletcher:
Lorin's "environment lease" proposal makes this redundant and is much less complex.


keel-core/src/test/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunnerTests.kt
```diff
@@ -67,7 +66,7 @@ internal class VerificationRunnerTests {
         name = "fnord-manifest",
         serviceAccount = "jamm@illuminati.org",
         artifacts = setOf(
-          DockerArtifact(name = "fnord", reference = "fnord-docker", branch = "main")
+          DockerArtifact(name = "fnord", reference = "fnord-docker")
```
lorin, line 69: The removal of the `branch` argument in this file looks like an accidental revert of unrelated code.

PR #1855, opened 2021-03-10
feat(slack): updating notifications based on user feedback

emjburns:
Updating the notifications based on gregs updated designs.

This biggest change is in the unpin notification, where we now show the pin reason and who pinned it. We also don't cross out the commit in the manual judgment after someone has taken action.


erikmunson:
👏 yay!

it's tough/impossible for me personally to give feedback on the product experience without screenshots, but can/will give feedback on them when they hit main if none are posted.

emjburns:
Screen Shot 2021-03-10 at 10 13 25 AM

emjburns:
Screen Shot 2021-03-10 at 10 16 07 AM
(I can't show what happens when you approve/deny the judgment because of the slack callback logic)

emjburns:
Screen Shot 2021-03-10 at 10 22 04 AM

erikmunson:
YOWZA those are delicious, the pinning ones in particular are 👨‍🍳 😘

PR #1857, opened 2021-03-10
fix(artifacts): Fix Docker artifact exclusion filter by tag

luispollo:
**Problem**
In #1848, I made [a change](https://github.com/spinnaker/keel/pull/1848/files#diff-9f58dc9351553d788680e001e102b4f871451189100572f06de9dc554acab0ebR198-R200) to the `ArtifactUtils.shouldInclude` function to account for the fact that I made `DockerArtifact.tagVersionStrategy` nullable. This function is called when building environment and artifact summaries for the UI to filter out docker tags that don't match the tag version strategy. However, when the artifact is filtered by source (branch, PR), that filter should not apply, but was incorrectly excluding all versions.

**Solution**
This PR fixes the bug in that function and replaces the condition with a more explicit check on `DockerArtifact.filteredBySource`.

### DIFF
--- a/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
+++ b/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
@@ -175,7 +175,7 @@ internal fun ArtifactVersionSelectStep.fetchSortedArtifactVersions(
   artifact: DeliveryArtifact,
   limit: Int? = null
 ): List<PublishedArtifact> {
-  return if (artifact.filteredByBranch || artifact.filteredByPullRequest) {
+  return if (artifact.filteredBySource) {
     fetchArtifactVersionsSortedWithQuery(artifact, limit)
   } else {
     fetchArtifactVersionsSortedWithComparator(artifact, limit)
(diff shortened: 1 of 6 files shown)


emjburns:
I'm confused about both the problem and the fix here - do you think you could post an example scenario and how it will change with this pr?

luispollo:
> I'm confused about both the problem and the fix here - do you think you could post an example scenario and how it will change with this pr?

Sure. The test I added intended to demonstrate the scenario where this broke things. If you have a Docker artifact with a `from` section (meaning it's filtered by branch or PR), then we should not attempt to filter versions based on the tag versioning strategy (because there won't be one). However, in the original code, I was returning `false` from the filter if that tag strategy was  `null`, which means I was filtering out all versions for the artifact. If you click on a version for any Docker artifact in the Environments view right now, you'll see the UI is broken because the API is not returning any artifact data due to this problem.

luispollo:
@Mergifyio refresh

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
```diff
@@ -189,15 +189,16 @@ internal fun ArtifactVersionSelectStep.fetchSortedArtifactVersions(
...
- * Returns true if a docker tag is not a match to the regex produces exactly one capture group on the tag, false otherwise.
+ * @return true if a docker tag is a match to the regex associated with the [TagVersionStrategy] and produces exactly
+ * one capture group on the tag, false otherwise. This only applies if the artifact is defined in terms of versions.
  */
-internal fun shouldInclude(tag: String, artifact: DockerArtifact) =
-  if (artifact.tagVersionStrategy == null) {
-    log.warn("Attempt to check Docker tag against unspecified tag version strategy for $artifact. Excluding.")
-    false
+internal fun filterDockerTag(tag: String, artifact: DockerArtifact) =
+  if (artifact.filteredBySource) {
```
luispollo, line 199: This was the bug. Instead of checking if the tag version strategy is `null`, I thought it was clearer to check if the artifact is filtered by source, which means the filter doesn't apply, and so we return `true`.

emjburns (approved):

keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/ArtifactUtils.kt
```diff
@@ -189,15 +189,16 @@ internal fun ArtifactVersionSelectStep.fetchSortedArtifactVersions(
...
+    log.debug("Ignoring Docker tag filter for $artifact, tag $tag since it's filtered by source.")
+    true
```
luispollo, line 201: This was the bug. We should be retuning `true` and not `false` inside this `if` statement. Instead of checking if the tag version strategy is `null`, I thought it was clearer to check if the artifact is filtered by source, which means the filter doesn't apply, and so we return `true`.

PR #1859, opened 2021-03-10
Sketch of proposed lease approach for preventing concurrent verifications/deployments

lorin:
Illustrates how we could use the proposed "environment leases" approach.

There's a low-level (Sql)EnvironmentLeaseRepository that just does the short-term locking on an environment.

There's an EnvironmentExclusionEnforcer which consumes the EnvironmentLeaseRepository and does some additional checking to see if there are active deployments or verifications, depending on why a lease is being taken.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcer.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcer.kt
@@ -1,0 +1,34 @@
+package com.netflix.spinnaker.keel.enforcers
+
+import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.api.Verification
+import com.netflix.spinnaker.keel.api.artifacts.PublishedArtifact
+import com.netflix.spinnaker.keel.api.verification.VerificationContext
+import org.springframework.stereotype.Component
+
+interface Lease {
+  fun <T> withLease(action: () -> T) : T
+
+}
+
+
+@Component
+class EnvironmentExclusionEnforcer(
+) {
+  /**
+   * This lease takes the simple repository lease as well as checking that
+   */
+  class CompoundLease : Lease {
+    override fun <T> withLease(action: () -> T): T {
+      /**
+       * Take simple lease
+       * Check if it's actually safe
+       */
+      return action.invoke()
+    }
+  }
+
+  fun requestVerificationLease(context: VerificationContext, verification: Verification) : Lease {
+    return CompoundLease()
+  }
+}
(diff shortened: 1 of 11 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcer.kt
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * take an action and then release the lease.
+   */
+  private fun <T> Lease.withLease(action: () -> T): T =
+    try {
+      action.invoke()
+    } catch (ex: Exception) {
+      throw ex
```
robfletcher, line 77: You can omit the `catch` if that's all it's going to do

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcer.kt
```diff
@@ -0,0 +1,93 @@
...
+    } finally {
+      environmentLeaseRepository.release(this)
+    }
+
+  /**
+   * take an action and then release the lease, suspend flavor
+   */
+  private suspend fun <T> Lease.withLeaseSuspend(action: suspend () -> T) : T =
```
robfletcher, line 85: I'm not sure there's any need for 2 versions. If the action doesn't require suspending it will still work fine.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/persistence/EnvironmentLeaseRepository.kt
```diff
@@ -0,0 +1,34 @@
...
+ *
+ */
+interface EnvironmentLeaseRepository {
+
+  /**
+   * Try to take a lease on an environment
+   *
+   * @returns a lease if nobody else has one
+   *
+   * @throws ActiveLeaseExists if someone else has a lease on the environment
```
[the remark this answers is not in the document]
lorin: This is handled in the enforcer, which now catches it and then throws an `EnvironmentCurrentlyBeingActedOn` exception . I forgot to put this handling in initially, just added it.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcer.kt
lorin: This version is to handle the case where the action does require suspend (note the `suspend` annotation on the action).

lorin:
Closing this for now, will submit as real PRs shortly.

PR #1862, opened 2021-03-10
feat(slack): hide long commit msgs, show in modal

emjburns:
Large commit messages were driving me crazy. This pr hides characters > 100 (configurable, but I picked 100 for now) behind a little "Show full commit" modal. Will add some photos shortly.

I wish I could make this code a bit nicer, but I could not figure out how to do the correct slack things in one function. So it's two for now. 

Deployd notification and a glimpse of the popup:
Screen Shot 2021-03-10 at 3 09 36 PM
Screen Shot 2021-03-10 at 3 09 44 PM

(there is an "Ok" button and a "close" button because I'm hacking the confirmation dialog page. If anyone knows how to just launch a modal let me know!)

pinned:
Screen Shot 2021-03-10 at 3 10 50 PM

example of a message that isn't too long:
Screen Shot 2021-03-10 at 3 11 46 PM


emjburns:
Looks like actual modals are much more complicated: https://api.slack.com/surfaces/modals/using

I'll come back to this dialog if it becomes a problem.

PR #1863, opened 2021-03-11
use base image name rather than version for diffs

robfletcher:
Currently we use the `base_ami_version` tag from the active server group to determine if its base AMI is up-to-date. However, since the versions for bionic and bionic-classic OS variants are identical, we have no way to detect a diff if a user changes from one OS to the other, and they will have to push a new version of their app to trigger a bake.

This PR changes things so that we use the base AMI _name_ rather than version. This has a number of advantages:
- It's unique to each OS variant so we'll detect a diff if the user changes from `bionic` to `bionic-classic` or vice-versa
- It's easier for us to get either via Bakery's API or via Bakery fast properties as we do now (currently we have to do an additional lookup via Clouddriver to turn the AMI name into a version)
- It's a more meaningful string to a user

@ajordens has made a corresponding change so that CloudDriver's AMI lookup endpoint `/aws/images/find?q=` will include the `base_ami_name` tag. This change should not go out until the CloudDriver change is live. However, it's a little fiddlier to get the AMI name from an active server group (see comments below) since it's not present as a tag there.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/BaseImageCache.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/BaseImageCache.kt
@@ -10,7 +10,7 @@ interface BaseImageCache {
    * image, if it exists.
    * @throws UnknownBaseImage if there is no known base image for [os] and [label].
    */
-  fun getBaseAmiVersion(os: String, label: BaseLabel): String
+  fun getBaseAmiName(os: String, label: BaseLabel): String
 
   val allVersions: Map<String, Map<String, String>>
 }
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
@@ -117,7 +117,7 @@ class BakeryLifecycleMonitor(
         vmType = detail.vmType,
         cloudProvider = detail.cloudProviderType,
         appVersion = detail.`package`.substringBefore("_all.deb").replaceFirst("_", "-"),
-        baseAmiVersion = imageService.findBaseAmiVersion(detail.baseAmiId, detail.region),
+        baseAmiName = imageService.findBaseAmiName(detail.baseAmiId, detail.region),
         timestamp = execution.endTime ?: clock.instant(),
         amiIdsByRegion = details.associate { regionDetail -> regionDetail.region to regionDetail.imageId }
       )
--- a/keel-bakery-plugin/src/test/kotlin/com/netflix/spinnaker/keel/bakery/DefaultBaseImageCacheTests.kt
+++ b/keel-bakery-plugin/src/test/kotlin/com/netflix/spinnaker/keel/bakery/DefaultBaseImageCacheTests.kt
@@ -16,14 +16,19 @@ internal class DefaultBaseImageCacheTests : JUnit5Minutests {
     val config =
       """
       |bionic:
-      |  candidate: nflx-base-5.375.0-h1224.8808866
-      |  unstable: nflx-base-5.375.1-h1225.8808866~unstable
-      |  release: nflx-base-5.365.0-h1191.6a005e3
+      |  candidate: bionicbase-x86_64-202103092356-ebs
+      |  unstable: bionicbase-unstable-x86_64-202103092010-ebs
+      |  release: bionicbase-x86_64-202103092356-ebs
+      |bionic-classic:
+      |  release: bionic-classicbase-x86_64-202102241716-ebs
+      |  candidate: bionic-classicbase-x86_64-202103092356-ebs
+      |  unstable: bionic-classicbase-unstable-x86_64-202103092010-ebs
+      |  previous: bionic-classicbase-x86_64-202101262358-ebs
       |xenial:
-      |  candidate: nflx-base-5.375.0-h1224.8808866
-      |  unstable: nflx-base-5.375.1-h1225.8808866~unstable
-      |  release: nflx-base-5.365.0-h1191.6a005e3
-      |  previous: nflx-base-5.344.0-h1137.cc92ef3
+      |  candidate: xenialbase-x86_64-202103092356-ebs
+      |  unstable: xenialbase-unstable-x86_64-202103092010-ebs
+      |  release: xenialbase-x86_64-202103092356-ebs
+      |  previous: xenialbase-x86_64-202101262358-ebs
       |""".trimMargin()
     val baseImageCache = DefaultBaseImageCache(YAMLMapper().readValue(config))
   }
@@ -31,20 +36,20 @@ internal class DefaultBaseImageCacheTests : JUnit5Minutests {
   fun tests() = rootContext<Fixture> {
     fixture { Fixture }
 
-    test("returns the base AMI version for a valid os/label") {
-      expectThat(baseImageCache.getBaseAmiVersion("bionic", RELEASE))
-        .isEqualTo("nflx-base-5.365.0-h1191.6a005e3")
+    test("returns the base AMI name for a valid os/label") {
+      expectThat(baseImageCache.getBaseAmiName("bionic", RELEASE))
+        .isEqualTo("bionicbase-x86_64-202103092356-ebs")
     }
 
     test("throw an exception for an invalid label") {
       expectThrows<UnknownBaseImage> {
-        baseImageCache.getBaseAmiVersion("bionic", PREVIOUS)
+        baseImageCache.getBaseAmiName("bionic", PREVIOUS)
       }
     }
 
     test("throw an exception for an invalid os") {
       expectThrows<UnknownBaseImage> {
-        baseImageCache.getBaseAmiVersion("windows", RELEASE)
+        baseImageCache.getBaseAmiName("windows", RELEASE)
       }
     }
   }
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
@@ -25,10 +25,10 @@ import com.netflix.spinnaker.keel.clouddriver.model.Image
 import com.netflix.spinnaker.keel.clouddriver.model.NamedImage
 import com.netflix.spinnaker.keel.clouddriver.model.NamedImageComparator
 import com.netflix.spinnaker.keel.clouddriver.model.appVersion
-import com.netflix.spinnaker.keel.clouddriver.model.baseImageVersion
+import com.netflix.spinnaker.keel.clouddriver.model.baseImageName
 import com.netflix.spinnaker.keel.clouddriver.model.creationDate
 import com.netflix.spinnaker.keel.clouddriver.model.hasAppVersion
-import com.netflix.spinnaker.keel.clouddriver.model.hasBaseImageVersion
+import com.netflix.spinnaker.keel.clouddriver.model.hasBaseImageName
 import com.netflix.spinnaker.keel.core.api.DEFAULT_SERVICE_ACCOUNT
 import com.netflix.spinnaker.keel.filterNotNullValues
 import com.netflix.spinnaker.keel.parseAppVersion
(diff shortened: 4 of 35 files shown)


keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/BaseImageCache.kt
```diff
@@ -10,7 +10,7 @@ interface BaseImageCache {
    * image, if it exists.
    * @throws UnknownBaseImage if there is no known base image for [os] and [label].
    */
-  fun getBaseAmiVersion(os: String, label: BaseLabel): String
+  fun getBaseAmiName(os: String, label: BaseLabel): String
```
robfletcher, line 13: This will require a corresponding change in keel-nflx obvs

keel-bakery-plugin/src/test/kotlin/com/netflix/spinnaker/keel/bakery/DefaultBaseImageCacheTests.kt
```diff
@@ -16,35 +16,40 @@ internal class DefaultBaseImageCacheTests : JUnit5Minutests {
...
+      |  previous: bionic-classicbase-x86_64-202101262358-ebs
       |xenial:
-      |  candidate: nflx-base-5.375.0-h1224.8808866
-      |  unstable: nflx-base-5.375.1-h1225.8808866~unstable
-      |  release: nflx-base-5.365.0-h1191.6a005e3
-      |  previous: nflx-base-5.344.0-h1137.cc92ef3
+      |  candidate: xenialbase-x86_64-202103092356-ebs
+      |  unstable: xenialbase-unstable-x86_64-202103092010-ebs
+      |  release: xenialbase-x86_64-202103092356-ebs
+      |  previous: xenialbase-x86_64-202101262358-ebs
```
robfletcher, line 31: Here you can see the difference in what we'll be storing in the cache. Base AMI version e.g. `nflx-base-5.365.0-h1191.6a005e3` is the app version of the baseos repo, whereas the base AMI _name` e.g. `bionicbase-x86_64-202103092356-ebs` is the actual unique name of the AMI generated for each OS variant from that build.

keel-bakery-plugin/src/test/kotlin/com/netflix/spinnaker/keel/bakery/DefaultBaseImageCacheTests.kt
```diff
@@ -16,35 +16,40 @@ internal class DefaultBaseImageCacheTests : JUnit5Minutests {
...
 
   fun tests() = rootContext<Fixture> {
     fixture { Fixture }
 
-    test("returns the base AMI version for a valid os/label") {
-      expectThat(baseImageCache.getBaseAmiVersion("bionic", RELEASE))
-        .isEqualTo("nflx-base-5.365.0-h1191.6a005e3")
+    test("returns the base AMI name for a valid os/label") {
+      expectThat(baseImageCache.getBaseAmiName("bionic", RELEASE))
+        .isEqualTo("bionicbase-x86_64-202103092356-ebs")
```
robfletcher, line 41: I've tried to update tests to use the correct format to avoid any future confusion

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -153,32 +153,32 @@ class ImageService(
...
-  private fun findBaseAmiVersion(image: NamedImage): String? =
+  private fun findBaseAmiName(image: NamedImage): String? =
     image
       .tagsByImageId
       .values
       .filterNotNull()
-      .find { it.containsKey("base_ami_version") }
-      ?.getValue("base_ami_version")
+      .find { it.containsKey("base_ami_name") }
+      ?.getValue("base_ami_name")
```
robfletcher, line 181: This reflects the change @ajordens made in CloudDriver so that this new `base_ami_name` tag will be present

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -186,7 +186,7 @@ data class ActiveServerGroupImage(
   ) : this(
     imageId = imageId,
     appVersion = tags.getTag("appversion")?.substringBefore("/"),
-    baseImageVersion = tags.getTag("base_ami_version"),
+    baseImageName = description?.let{ Regex("""ancestor_name=(.+?),""").find(it)?.groupValues?.get(1) },
```
robfletcher, line 189: Sadly on the active server group we need to scrape the base AMI name out of the image description as there's no convenient tag available

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/BaseImageCache.kt
```diff
@@ -10,7 +10,7 @@ interface BaseImageCache {
...
 
-  val allVersions: Map<String, Map<String, String>>
+  val allVersions: Map<String, Map<BaseLabel, String>>
```
robfletcher, line 15: This is just used by the Keel endpoint that lists out the base AMIs. Using the enum helped me keep it straight in my head

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/BakedImage.kt
```diff
@@ -10,7 +10,7 @@ data class BakedImage(
   val vmType: String,
   val cloudProvider: String,
   val appVersion: String,
-  val baseAmiVersion: String,
+  val baseAmiName: String,
```
emjburns, line 13: You'll need to make the corresponding change in the [bakeryLifecycleMonitor](https://github.com/spinnaker/keel/blob/master/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt#L120), which means changing the function called there in [ImageService](https://github.com/spinnaker/keel/blob/master/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt#L164).

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
emjburns: yaaaaaay images

ajordens: I suspect the ordering of `description` is consistent _but_ `ancestor_name=([\w-]+)` might also work for the regex should there actually not be a trailing `,`?

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/BakeryLifecycleMonitor.kt
```diff
@@ -117,7 +117,7 @@ class BakeryLifecycleMonitor(
         vmType = detail.vmType,
         cloudProvider = detail.cloudProviderType,
         appVersion = detail.`package`.substringBefore("_all.deb").replaceFirst("_", "-"),
-        baseAmiVersion = imageService.findBaseAmiVersion(detail.baseAmiId, detail.region),
+        baseAmiName = imageService.findBaseAmiName(detail.baseAmiId, detail.region),
```
robfletcher, line 120: @emjburns I did change that bit

emjburns: oh cool I'm blind. thanks for pointing it out!

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
robfletcher: Hmm, that's a good point. Probably safer.

robfletcher: Fixed

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -186,7 +186,7 @@ data class ActiveServerGroupImage(
   ) : this(
     imageId = imageId,
     appVersion = tags.getTag("appversion")?.substringBefore("/"),
-    baseImageVersion = tags.getTag("base_ami_version"),
+    baseImageName = description?.let{ Regex("""ancestor_name=([\w-]+)""").find(it)?.groupValues?.get(1) },
```
lorin, line 189: As a reader, I'd really appreciate having an example of a `description` string here as a comment. 

There's a good example in `ActiveServerGroupTest`:

```
name=fnord, arch=x86_64, ancestor_name=xenialbase-x86_64-201811142132-ebs, ancestor_id=ami-0fe383ecea7f75eac, ancestor_version=nflx-base-5.308.0-h1044.b4b3f78
```

lorin: Note also that the `get` is going to throw an exception if the description exists but doesn't match the regex. I'm not sure what behavior we actually want in that situation.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/ActiveServerGroupImageDeserializer.kt
```diff
@@ -8,14 +8,17 @@ import com.netflix.spinnaker.keel.api.ec2.ServerGroup.ActiveServerGroupImage
...
+  StdNodeBasedDeserializer<ActiveServerGroupImage>(ActiveServerGroupImage::class.java) {
   override fun convert(root: JsonNode, ctxt: DeserializationContext): ActiveServerGroupImage {
     val tags = root.get("tags") as ArrayNode
 
     return ActiveServerGroupImage(
       imageId = root.get("imageId").textValue(),
       appVersion = tags.getTag("appversion")?.substringBefore("/"),
-      baseImageVersion = tags.getTag("base_ami_version"),
+      baseImageName = root.get("description").textValue()?.let {
+        Regex("""ancestor_name=([\w-]+)""").find(it)?.groupValues?.get(1)
```
lorin, line 20: Givent that the identical parsing logic is in the ActiveServerGroup.kt class, (and that this involves a regex), I'd recommend encapsulating this logic in a function.

lorin (approved):
#### My understanding

*Problem*

Currently, keel identifies base AMIs by the version number (e.g., "nflx-base-5.365.0-h1191.6a005e3"). The problem is that different variants of the same Ubuntu version (e.g., bionic, bionic-classic) can have the same version number, which means that keel won't detect a difference if a delivery config changes the variant (e.g., from bionic to bionic-classic).

*Implemented solution*

Use the AMI name instead of the version. The name uniquely identifies a variant (e.g., bionicbase-x86_64-202103092356-ebs vs bionic-classicbase-x86_64-202103092356-ebs).

keel determines the base AMI name is by parsing it out of the `description` field of an ASG. The logic for that is in the following two files:

```
keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/ActiveServerGroupImageDeserializer.kt
```

Most of the other changes are renames from `version` to `name` (e.g., of the form `s/getBaseAmiVersion/getBaseAmiName`.)

#### Feedback

Looks good to me. (some minor comments inline)

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
robfletcher: That's what all the null-safety is for. If it doesn't match `find` will return `null` (and `baseImageName` will end up as `null`)

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/ActiveServerGroupImageDeserializer.kt
robfletcher: I'm inclined to agree

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
lorin: Ah, thanks.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/jackson/ActiveServerGroupImageDeserializer.kt
robfletcher: Done

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/model/ActiveServerGroup.kt
```diff
@@ -201,6 +201,22 @@ private fun List<Map<String, Any?>>.getTag(key: String) =
...
+ * On a [NamedImage] this is available as a tag, but not on a server group.
+ */
+fun extractBaseImageName(description: String?): String? =
+  description?.let {
+    """ancestor_name=([\w-]+)"""
+      .toRegex()
+      .find(it)
+      ?.destructured
+      ?.let { (baseAmiName) -> baseAmiName }
+  }
```
robfletcher, line 218: @lorin here's the regex logic extracted as a function

PR #1864, opened 2021-03-11
feat(notifications): add delivery config updated notifications

emjburns:
I have an idea. 

I'd like to send delivery config updated notifications. I'd _like_ to include the git metadata, but we don't have a git integration. So my idea is that I'll modify the orca "Import delivery config" stage to scrape the rocket trigger and pass that information to `keel`.

To do that, I've created a new endpoint to submit delivery configs to. I'll update the stage to test this out.

Here are sample images with and without git data. Just like the updated commit messages, this notification also hides the long text behind a "show more" modal. If there is no commit data I'm linking to the API representation of the config that we have. If this is useful, we can think about showing this in a better way that's not a json blob.
Screen Shot 2021-03-11 at 4 14 35 PM

Screen Shot 2021-03-11 at 2 47 39 PM

This pr is kinda long because I needed to adjust some of the helper methods to take different parameters.

Bonus: my next idea is to send "validation failed" slack events so we can start showing users what is wrong with their delivery config!


keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -68,7 +72,39 @@ class DeliveryConfigController(
       .let { processedDeliveryConfig ->
         validator.validate(processedDeliveryConfig)
         log.debug("Upserting delivery config '${processedDeliveryConfig.name}' for app '${processedDeliveryConfig.application}'")
-        repository.upsertDeliveryConfig(processedDeliveryConfig)
+        val config = repository.upsertDeliveryConfig(processedDeliveryConfig)
+        publisher.publishEvent(DeliveryConfigChangedNotification(config))
+        return config
+      }
+
```
emjburns, line 79: here is the new endpoint that I'm experimenting with. It takes both a delivery config and also git metadata which I will scrape from the git trigger of pipelines (if it exists)

luispollo (commented):
I like the idea, but have a couple design questions:
- Instead of a separate endpoint and a formal object to map the git metadata, what about just using `DeliveryConfig.metadata`? That way, it'd be totally backwards compatible.
- Is the fact that the second rendering of the notification (with the git metadata) does not contain the "View as JSON" link just an accident or were you really not planning to expose it in that format? I think that link would be really valuable! Better yet if it was a link to the commit diff instead of just the file.

luispollo:
One more question: you mentioned we don't have a "git integration", but we kinda do:
* We already receive Rocket code events that we can cross-reference with the application's configured Stash repo. We could use this to detect changes to the delivery config even without the Orca changes.
* We already have functions in `IgorService` to fetch contents from a Stash repo. We could use that to fetch the JSON.
* I _think_ we may have APIs available from Rocket to get more metadata about a commit (e.g. which files were changed), which we could use in conjunction with code events to find out more about what changed in a commit.

emjburns:
@luis I'm going to address the comments about the git pieces first.

We definitely could take advantage of the various git pieces that we have. I even think that we could use the specific pieces you describe to build a git integration for Netflix. I've been idly thinking about whether we have all the pieces to build the git integration we want without too much work. For me, this would look like defining your repo/branch info in the delivery config. Keel would listen for rocket events like you're describing and react to them by importing the delivery config from a new commit. 

I believe that fully doing that git integration is somewhat costly. I don't have a time estimate, but I do know it would touch keel (decide on a delivery ux that is clear, get user feedback, update delivery config fields, build logic for calling out to igor to get the file and parsing that and submitting it to the api, handling errors nicely) and involve some exploration into igor/rocket apis, and involve porting the error handling from orca to keel. There are also a bunch of open questions around UX and visibility that I don't have an answer for. 
- How can we surface if the git integration is configured right? 
- How can we show errors to users? 
- How do we add this to newt? What questions would we ask? 
- Do we always have the right permissions on user repositories?
(to name a few that I've thought of)

I think that there is a lot of value in providing notification information to our users right now. People want a gut check that we received their delivery config (to build trust) and would love a link to the commit that pushed it. We've also had user feedback that there is a lack of visibility into config changes since the import pipeline is something people forget about. People _might_ also want a link to the json that we have (I'm not actually sure, but putting it in only one notification is a good way to test whether people want it I think!).

Because I think there is value in delivering config change notifications, and there is a cost to building a git nice integration that could eventually provide a similar experience, my current position is that there's value in rolling out this day's work. I don't think that a nice git integration would add a lot more value to this notification experience. I'm estimating that this hack can provide maybe 60-70% of the notification value we eventually want to have with a very low engineering cost.

Based on that I'd like to move quickly and test this out. If this is super valuable and our team thinks the actual git integration would make it more valuable and robust I'll do the work to build that and convert all these pieces over to use that.

Lastly, regarding using the `DeliveryConfig.metadata` field. That's currently a field that we don't accept user input on (technically you can submit the field, but we don't do anything with it). Do you think it's worth changing that assumption to be more like kubernetes? I know kubernetes powers a lot of flexibility with a metadata field. It would certainly make things like this neater.

luispollo:
Hi @emjburns, thanks for taking the time to write up this rationale. Super helpful.

I'll start backwards, but this is the most important feedback I wanted to give:
> Based on that I'd like to move quickly and test this out. If this is super valuable and our team thinks the actual git integration would make it more valuable and robust I'll do the work to build that and convert all these pieces over to use that.

I agree 100% on moving quickly, experimenting, and optimizing later once we've proven the benefit of any feature. The only reason I asked these questions was to make sure we were on the same page about the technical options available to make a good judgement on the implementation approach. I'll proceed to reviewing your PRs and have absolutely no objection to you moving forward with this approach.

Having said that, I wanted to clarify a couple of things:

> I believe that fully doing that git integration is somewhat costly. I don't have a time estimate, but I do know it would touch keel (decide on a delivery ux that is clear, get user feedback, update delivery config fields, build logic for calling out to igor to get the file and parsing that and submitting it to the api, handling errors nicely) and involve some exploration into igor/rocket apis, and involve porting the error handling from orca to keel. 

Just to make sure we're on the same page, we already have support in keel today for importing delivery configs from source via igor: https://github.com/spinnaker/keel/blob/master/keel-igor/src/main/kotlin/com/netflix/spinnaker/keel/igor/DeliveryConfigImporter.kt

We can always improve it, but it provides the basics, and -- what do you know! -- I had already added logic to add some metadata into `DeliveryConfig.metadata`! (At least I'm being consistent! 😄 )

> There are also a bunch of open questions around UX and visibility that I don't have an answer for.
> - How can we surface if the git integration is configured right?
> - How can we show errors to users?
> - How do we add this to newt? What questions would we ask?

To be clear, I wasn't advocating for building the full git integration experience (to replace the "Import Delivery Config" pipeline stage) now! I was just thinking that we could build these notifications without having to modify Orca.

> Do we always have the right permissions on user repositories?

Just FYI, the answer is yes. I've worked with the Stash team to ensure that every new repository includes a read permission to our automation user. Existing repos would already have the permission because they'd be using the pipeline.

Hope this helps! Now I'll review your PRs! 🙂

emjburns:
@luispollo 👍  I was not aware that you had already added a class to import a delivery config from source in `keel`. Thanks for the permissions clarifications as well.

luispollo:
> @luispollo 👍 I was not aware that you had already added a class to import a delivery config from source in `keel`. Thanks for the permissions clarifications as well.

Yup. I did that in #1261 to make progress towards #870.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
emjburns: Update - I'm going to move the gitMetadata to the "metadata" section of the delivery config like @luispollo suggested, and then strip it so we don't save it to the database.

lorin (approved):
### My understanding

#### Problem

If a user updates their delivery config with keel, they don't get any feedback that keel has seen that the update actually happened.

#### Implemented solution

Emit a Slack notification when keel receives an updated delivery config. This is at the "normal" level of notifications.

If the delivery config contains information about the git commit associated with the delivery config update, this will appear in the notification.

The git metadata is passed in the `metadata` map of the delivery config, it is a serialized version of the existing [GitMetadata][1] class.



[1]: https://github.com/spinnaker/keel/blob/9db12ff2949db3c38501747e6cae3de3ec11fc6f/keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/GitMetadata.kt#L6-L14

e.g., 

```yaml
metadata:
  gitMetaData:
    repo: 
      name: ...
      link: ...
    commit: 95f803c72
    ...
```


The event class is `DeliveryConfigChangedNotification`, which is emitted in `DeliveryConfigController.upsert`, and is handled by `NotificationEventListener.onDeliveryConfigChangedNotification`, which in turn calls `DeliveryConfigNotificationHandler.sendMessage` to actually send the slack message.


Some minor changes to multiple files due to:

* change of spinnaker url prefix config info
* change to GitDataGenerator method signature, to accomodate slack messages associated with git commits that are not specifcally tied to artifacts.



### My feedback

Looks good to me!

I'm going to miss good ol' `DELIVER_CONFIG_UPDATED` (sic).

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/DeliveryConfigNotificationHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+  private val gitDataGenerator: GitDataGenerator
+) : SlackNotificationHandler<SlackConfigNotification> {
+  override val supportedTypes = listOf(DELIVERY_CONFIG_CHANGED)
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  override fun sendMessage(notification: SlackConfigNotification, channel: String) {
+    with(notification) {
+      log.debug("Sending config changed notification for application ${notification.application}")
+      val imageUrl = "https://raw.githubusercontent.com/spinnaker/spinnaker.github.io/master/assets/images/md_icons/config_change.png"
```
luispollo, line 25: Just a small suggestion, since these appear in multiple places now: in a future PR, we could create a `IconRepository` class coupled with a well defined enum of icon types to make this cleaner. So you'd end up with something like this here:
```kotlin
val imageUrl = iconRepository.getUrlFor(DELIVERY_CONFIG_CHANGED)
```
Another reason I'm suggesting this is that @ranihorev has proposed that the backend starts returning icons for resources and other things so that we don't have to change deck every time there's a new thing to represent.

luispollo: Or it could it even be just config in `keel.yml`.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/DeliveryConfigNotificationHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+          }
+          gitDataGenerator.conditionallyAddFullCommitMsgButton(this, gitMetadata)
+          section {
+            gitDataGenerator.generateScmInfo(this, application, gitMetadata, null)
+          }
+        } else {
+          section {
+            markdownText("*App:* $application\n\n" +
+              "No commit info available.\n" +
+              "View the current <${gitDataGenerator.generateConfigUrl(application)}|config as json>."
```
luispollo, line 47:
```suggestion
              "View the current <${gitDataGenerator.generateConfigUrl(application)}|config as JSON>."
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/DeliveryConfigNotificationHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+          section {
+            gitDataGenerator.generateCommitInfoNoArtifact(this, application, imageUrl, altText, notification.gitMetadata)
+          }
+          gitDataGenerator.conditionallyAddFullCommitMsgButton(this, gitMetadata)
+          section {
+            gitDataGenerator.generateScmInfo(this, application, gitMetadata, null)
+          }
+        } else {
+          section {
+            markdownText("*App:* $application\n\n" +
```
luispollo, line 45:
```suggestion
            markdownText("*Application:* $application\n\n" +
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
```diff
@@ -160,6 +156,33 @@ class GitDataGenerator(
...
+   *
+   */
+  fun generateCommitInfoNoArtifact(sectionBlockBuilder: SectionBlockBuilder,
+                                   application: String,
+                                   imageUrl: String,
+                                   altText: String,
+                                   gitMetadata: GitMetadata): SectionBlockBuilder {
+
+    with(sectionBlockBuilder) {
+      markdownText("*App:* $application\n" +
```
luispollo, line 175:
```suggestion
      markdownText("*Application:* $application\n" +
```

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
```diff
@@ -160,6 +156,33 @@ class GitDataGenerator(
...
+        "Commit by @${gitMetadata.author}\n " +
```
luispollo, line 176: Just a heads-up that, if you want tagging to work in Slack, you can't just use the commit author here, you need to call the Slack API to translate the user email to a Slack username, like here: https://github.com/spinnaker/keel/pull/1864/files#diff-a3f4013f961d345a64b48d91be4d7f2f7c3952522e05f850d5b0ca8c30fb6d12R31

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -60,16 +67,37 @@ class DeliveryConfigController(
...
+      description = "The delivery config. If its `name` matches an existing delivery config the operation is " +
+        "an update, otherwise a new delivery config is created."
     )
     deliveryConfig: SubmittedDeliveryConfig
-  ): DeliveryConfig =
-    deliveryConfigProcessors.applyAll(deliveryConfig)
+  ): DeliveryConfig {
+    val metadata: MutableMap<String,Any?> = deliveryConfig.metadata?.toMutableMap() ?: mutableMapOf()
+    val gitMetadata: GitMetadata? = try {
+      val candidateMetadata = metadata.getOrDefault("gitMetadata", null)
```
luispollo, line 77: Nice! This is what I had in mind.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -60,16 +67,37 @@ class DeliveryConfigController(
...
+        null
+      }
+    } catch (e: IllegalArgumentException) {
+      log.debug("Error converting git metadata ${metadata.getOrDefault("gitMetadata", null)}: {}", e)
+      // not properly formed, so ignore the metadata and move on
+      null
+    }
+
+    // we don't save this key to the database, since we're programmatically adding it in orca.
+    metadata.remove("gitMetadata")
```
luispollo, line 90: Why remove? We already add a ton of metadata everywhere that is returned by the API, so if you call `GET /applications/{application}/config` what you get is much different from what's in your git repo already. I propose we keep it as it can be useful for troubleshooting.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
```diff
@@ -60,16 +67,37 @@ class DeliveryConfigController(
...
+
+    deliveryConfigProcessors.applyAll(deliveryConfig.copy(metadata = metadata))
       .let { processedDeliveryConfig ->
         validator.validate(processedDeliveryConfig)
         log.debug("Upserting delivery config '${processedDeliveryConfig.name}' for app '${processedDeliveryConfig.application}'")
-        repository.upsertDeliveryConfig(processedDeliveryConfig)
+        val config = repository.upsertDeliveryConfig(processedDeliveryConfig)
+        publisher.publishEvent(DeliveryConfigChangedNotification(config, gitMetadata))
```
luispollo, line 97: nit: Should this be encapsulated in a `DeliveryConfigProcessor` like other similar processing we already do?

luispollo (approved):
Overall, looks good to me. Left you some small suggestions on wording, the delivery config metadata, and a heads-up on tagging users in Slack.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/DeliveryConfigNotificationHandler.kt
emjburns: Greg's designs say "app" so I'm going to stick with that.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/DeliveryConfigNotificationHandler.kt
emjburns: I'll think about this in the future either when I need to make another icon change or when we update the API :P thanks for the suggestion

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
emjburns: oh right, good catch. I got thrown off because in my local tests this was working, but of course I was passing my slack handle and not the email.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
emjburns: Okay, I'll keep it. I wasn't sure what the best move was, but you're right it would be useful for troubleshooting.

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt
emjburns: I don't think so, since we're not modifying the delivery config or changing values. We're just gathering information to send a notification. I want to also add validation notifications here (so we can notify with a good message if there's a validation error). Maybe when I play with that it'll make more sense to refactor this area.

Issue #1830, opened 2021-03-01
Duplicate verification executions against the same version


lorin:
Due to a race condition, it is possible that two different keel instances launch the same verification task. We should never allow this to happen.

## Hypothesis about problem: we launch task before marking as running

The logic in [VerificationRunner.runVerifcationsFor][1] does:

1. get status (line 33)
2. if verification already running, don't run a new verification (lines 36-38)
3. start verification (line 48)
4. mark as running (line 49)

Two processes can both get to step 3 if we're unlucky with scheduling.

## Thinking about solution: change ordering

To ensure that only one verification can run at a time, we'd need to mark as running *before* starting verification.

Note that if we change the logic to mark before running, then we need some sort of lease mechanism to handle the case where:

1. instance mark as running
2. instance gets terminated before starting verification

[1]: https://github.com/spinnaker/keel/blob/7994aba537ebc6c5f4e9ce7dcc115dcc0dcb378e/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt#L28-L67

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

PR #1865, opened 2021-03-12
fix(envts): initial work to fix verification/deployment races

lorin:
## Summary

This is the first of a series of PRs to squish some race conditions.

This PR is a top-down start of a solution, with details to be filled in in future PRs.

There are no (intended) behavior changes in this PR.

## Context

Here are two [safety properties][1] of the verification behavior of Managed Delivery.

**P1: Two verifications should never execute concurrently against the same environment.**

For example, if the `acme/tests:stable` test container is currently running in the `staging` environment, keel should
not launch any other verifications in the `staging` environment.

**P2: Verifications against an environment should never happen concurrently with a deployment in an environment.**

For example, if the `fnord-v123.deb` artifact is being deployed to an EC2 cluster in the `staging` environment, keel should not launch verifications while this is happening.

## Problem

Because of race conditions, we don't properly enforce these safety properties today. See #1830, #1831.

## Sketch of overall solution

Add guard conditions before launching a verification or an actuation.

While it may not be strictly necessary to guard against *all* actuations, it simplifies the implementation to guard actuations instead of figuring out which actuations correspond to deployments.

### Enforcer and leases

These guard conditions are enforced by a new class named `EnvironmentExclusionEnforcer`.

The enforcer works by granting the client a *lease* if the guard conditions are met. A lease is a lock that has an expiration time. Leases expire to protect against the situation where an instance takes a lease and then terminates unexpectedly before releasing it.

If a client is unable to get a lease, the enforcer will throw a `EnvironmentCurrentlyBeingActedOn` exception.

Lease enforcement (including expiration) will be implemented in a future PR.

### Verification guard conditions

Before starting a verification in an environment E, ensure that:
* No other verifications are currently running in E
* No deployments are currently happening in E

Implemented by: `EnvironmentExclusionEnforcer.withVerificationLease`

### Actuation guard conditions

Before starting an actuation in an environment E, ensure that:

* No other verifications are running in E

Implemented by: `EnvironmentExclusionEnforcer.withActuationLease`

## Work in this PR

* Create the `EnvironmentExclusionEnforcer` class
* Call the enforcer before launching a verification
* Call the enforcer before actuating a resource
* Catch & log the `EnvironmentCurrentlyBeingActedOn` exception if fail to get a verification lease
* Catch & log the `EnvironmentCurrentlyBeingActedOn` exception if fail to get an actuation lease

All of the actual checks are no-ops and will be implemented in a future PR.

[1]: https://buttondown.email/hillelwayne/archive/safety-and-liveness-properties/


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -94,8 +95,12 @@ class CheckScheduler(
                    */
                   withTimeout(checkTimeout.toMillis()) {
                     launch {
```
emjburns, line 97: General comment - have you thought about how we communicate this to a user? since we will "freeze" the environment while it's being verified, is there a way we can let the user know? (maybe this is just UI treatment if a verification is running)

emjburns (approved):

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
luispollo: Thought: we could take advantage of `ApplicationEvent`s and start adding more sub-types, which we can then surface in the UI. We already have an API to query for them (`GET /application/{application}/events`).

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
```diff
@@ -94,8 +95,12 @@ class CheckScheduler(
...
-                      resourceActuator.checkResource(it)
-                      publisher.publishEvent(ResourceCheckCompleted(Duration.between(startTime, clock.instant())))
+                      try {
+                        resourceActuator.checkResource(it)
+                        publisher.publishEvent(ResourceCheckCompleted(Duration.between(startTime, clock.instant())))
+                      } catch (e: EnvironmentCurrentlyBeingActedOn) {
+                        log.warn("Couldn't actuate resource ${it.id} because environment is currently being acted on", e)
```
luispollo, line 102: Might be a nit, but I think words like "actuation" or "acted on" are kind of vague and not super user-friendly. If/when we expose a message to users in the UI, maybe we could say something like "...because resources in the environment are currently being created of modified"?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcer.kt
```diff
@@ -0,0 +1,62 @@
+package com.netflix.spinnaker.keel.enforcers
+
+import com.netflix.spinnaker.keel.api.Environment
+import com.netflix.spinnaker.keel.api.verification.VerificationContext
+import org.springframework.stereotype.Component
+
+/**
+ * Exception thrown when it's not safe to take action against the environment because
+ * something is is acting on it
```
luispollo, line 9:
```suggestion
 * Exception thrown when it's not safe to take action against the environment because
 * something is already acting on it.
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcer.kt
```diff
@@ -0,0 +1,62 @@
...
+ */
+class EnvironmentCurrentlyBeingActedOn(message: String) : Exception(message) { }
+
+@Component
+class EnvironmentExclusionEnforcer {
```
luispollo, line 14: Please consider adding a class doc. The description from your PR would be great here.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
```diff
@@ -39,7 +41,9 @@ class VerificationRunner(
       }
 
       statuses.firstOutstanding?.let { verification ->
-        start(verification, imageFinder.getImages(context.deliveryConfig, context.environmentName))
+        enforcer.withVerificationLease(this) {
```
luispollo, line 44: Why do we need this here in addition to the one in `CheckScheduler`?

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/actuation/IntermittentFailureTests.kt
```diff
@@ -70,6 +71,8 @@ class IntermittentFailureTests : JUnit5Minutests {
     val clock = MutableClock()
     val vetoRepository = mockk<UnhappyVetoRepository>(relaxUnitFun = true)
 
+    val environmentExclusionEnforcer = EnvironmentExclusionEnforcer()
```
luispollo, line 74: I'm guessing you'll want to mock responses from this object once the enforcement checks are implemented, so you might want to use mocks in the tests already.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
lorin: I like the idea of adding this via events.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/verification/VerificationRunner.kt
lorin: Because verifications aren't executed by the `CheckScheduler.checkResources` function, they're executed by the `CheckScheduler.verifyEnvironments` function, which is a different loop.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/actuation/CheckScheduler.kt
lorin: That's a good idea, will do this in a future PR.

PR #1866, opened 2021-03-12
get the base image name, not the base image's base image name

robfletcher:
This causes continual rebaking because the base_ami_name of the image we bake does not match the one this bit of code finds by looking up the base image by id.

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
@@ -153,33 +153,14 @@ class ImageService(
     }
   }
 
-  suspend fun findBaseAmiName(baseImageName: String): String {
-    return cloudDriverService.namedImages(DEFAULT_SERVICE_ACCOUNT, baseImageName, "test")
-      .lastOrNull()
-      ?.let { namedImage ->
-        findBaseAmiName(namedImage)
-      } ?: throw BaseAmiNotFound(baseImageName)
-  }
-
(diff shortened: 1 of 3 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -153,33 +153,14 @@ class ImageService(
...
-      .values
-      .filterNotNull()
-      .find { it.containsKey("base_ami_name") }
-      ?.getValue("base_ami_name")
-
+  suspend fun findBaseImageName(baseImageId: String, region: String): String =
+    cloudDriverService.getImage("test", region, baseImageId)
+      .lastOrNull()
+      ?.imageName
+      ?: throw BaseAmiNotFound(baseImageId)
```
robfletcher, line 163: This is the base image we're looking up, so we just want its name, not its own base image name

lorin: I know it's not directly related to this PR, but I figure I'll ask since I'm looking: do you know why it is that we hard-code `"test"` as the account here?

lorin (approved):
LGTM.

(I asked a background question about this code inline)

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
robfletcher: Because images are always baked it test

PR #1867, opened 2021-03-12
only consider images with the same base os when checking if we need to bake

robfletcher:
This is pretty hacky and I'd like to reconsider how we determine if we need to bake or not, but this fix should get us unblocked in the meantime.

### DIFF
--- a/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
+++ b/keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
@@ -19,7 +19,7 @@ package com.netflix.spinnaker.keel.clouddriver
 
 import com.github.benmanes.caffeine.cache.AsyncCache
 import com.netflix.frigga.ami.AppVersion
-import com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact
+import com.netflix.spinnaker.keel.artifacts.DebianArtifact
 import com.netflix.spinnaker.keel.caffeine.CacheFactory
 import com.netflix.spinnaker.keel.clouddriver.model.Image
 import com.netflix.spinnaker.keel.clouddriver.model.NamedImage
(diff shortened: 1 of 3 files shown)


keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -52,10 +52,10 @@ class ImageService(
    *
    * If no images at all exist for [artifact] this method returns `null`.
    */
-  suspend fun getLatestImage(artifact: DeliveryArtifact, account: String): Image? {
+  suspend fun getLatestImage(artifact: DebianArtifact, account: String): Image? {
     val candidateImages = cloudDriverService
       .namedImages(DEFAULT_SERVICE_ACCOUNT, artifact.name, account)
-      .filter { it.hasAppVersion && it.hasBaseImageName }
+      .filter { it.hasAppVersion && it.hasBaseImageName && it.baseImageName.startsWith("${artifact.vmOptions.baseOs}base") }
```
robfletcher, line 58: filter so that we only consider images with the same base OS

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageServiceTests.kt
```diff
@@ -49,7 +49,7 @@ class ImageServiceTests : JUnit5Minutests {
   class Fixture {
     val cloudDriver = mockk<CloudDriverService>()
     val subject = ImageService(cloudDriver, TEST_CACHE_FACTORY)
-    val artifact = DebianArtifact("my-package", vmOptions = VirtualMachineOptions(baseOs = "ignored", regions = emptySet()))
+    val artifact = DebianArtifact("my-package", vmOptions = VirtualMachineOptions(baseOs = "trusty", regions = emptySet()))
```
robfletcher, line 52: it's not ignored any more

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
robfletcher: I specifically made it a substring with `base` because `bionic` would match `bionic-classicbase` otherwise

emjburns (approved):

PR #1868, opened 2021-03-13
chore(notif): remove unhealthy notifier

emjburns:
Bye byyeeeee


emjburns:
The unhealthy notifier was an experiment to see if this kind of notification would add value. It did not add value (for a variety of reasons. I'm removing the classes here that do the notifying but not the actual data types. The actual notification is used in a test of some of the repeated notification stuff I built, and I suspect I might want to come back to that stuff if we decide to notify users repeatedly that their app is paused, or that their env has been pinned for a while. I didn't have the energy to refactor the tests so I just did not clean up the data types they used.

PR #1869, opened 2021-03-13
fix(mj + slack): log exception, guess at fixing

emjburns:
This week I added a feature to hide long commits in a pop up. Apparently, when we update the manual judgment notification we scrape the text from the original message and re-add it to the notification with the updated image and title. I broke that.

This PR takes a stab at fixing it, and also logs the exception we get so I can try and really fix it.

### DIFF
--- a/keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/callbacks/ManualJudgmentCallbackHandler.kt
+++ b/keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/callbacks/ManualJudgmentCallbackHandler.kt
@@ -66,21 +66,7 @@ class ManualJudgmentCallbackHandler(
       val originalCommitText = response.message.blocks[1].getText
       val action = actionsMap[response.actions.first().value]
 
-      // since we sometimes add a "show more" button, the position of this information changes.
-      //  I cannot figure out (yet) how to scrape the original section and just re-add it to the notification
-      val originalGitInfoText = if (response.message.blocks.size == 3) {
-        response.message.blocks[2].getText
-      } else {
-        response.message.blocks[3].getText
-      }
-
-      val originalUrl = if (response.message.blocks.size == 3) {
-        response.message.blocks[2].getUrl
-      } else {
-        response.message.blocks[3].getUrl
-      }
-
-      return withBlocks {
+      val newBlocks = withBlocks {
         header {
           when (action) {
             "approve" -> text("Manual judgement approved", emoji = true)
@@ -96,16 +82,6 @@ class ManualJudgmentCallbackHandler(
             }
           }
         }
-        section {
-          markdownText(originalGitInfoText)
-          accessory {
-            button {
-              text("More...")
-              actionId("button-action")
-              url(originalUrl)
-            }
-          }
-        }
 
         context {
           elements {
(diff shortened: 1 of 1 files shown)


keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/callbacks/ManualJudgmentCallbackHandler.kt
```diff
@@ -63,13 +64,25 @@ class ManualJudgmentCallbackHandler(
    * For example, if user gyardeni approved the notification, this function will add:
    * "@Gal Yardeni hit approve on 2021-02-12 11:05:57 AM" and will marked the original text with strikethrough.
    */
-  fun updateManualJudgementNotification(response: BlockActionPayload): List<LayoutBlock> {
+  fun oldUpdateManualJudgementNotification(response: BlockActionPayload): List<LayoutBlock> {
```
luispollo, line 67: Why the rename?

emjburns: arg oops, good catch. it was an accident from my other idea. I'll fix

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/callbacks/ManualJudgmentCallbackHandler.kt
```diff
@@ -63,13 +64,25 @@ class ManualJudgmentCallbackHandler(
...
     try {
       val originalCommitText = response.message.blocks[1].getText
-      val originalGitInfoText = response.message.blocks[2].getText
-      val originalUrl = response.message.blocks[2].getUrl
       val action = actionsMap[response.actions.first().value]
 
+      // since we sometimes add a "show more" button, the position of this information changes.
+      //  I cannot figure out (yet) how to scrape the original section and just re-add it to the notification
```
luispollo, line 73: Aren't the blocks all exactly the same except for the one added at the bottom to say "X person approved/rejected"? If that's the case, couldn't we just use the original array of blocks and tack that at the end? Why do we need to worry about these positions to begin with?

emjburns: We update the first block to have a different image.

I have been struggling with how to just grab the old blocks but actually, you just gave me an idea!

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/callbacks/ManualJudgmentCallbackHandler.kt
```diff
@@ -104,8 +90,13 @@ class ManualJudgmentCallbackHandler(
         }
       }
 
+      val originalBlocks = response.message.blocks
+      //remove the first two blocks because we're replacing them
+      originalBlocks.removeFirstOrNull()
+      originalBlocks.removeFirstOrNull()
+      return newBlocks + originalBlocks
```
luispollo, line 97: Perfect! Something like that is what I had in mind!

luispollo (approved):

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/callbacks/ManualJudgmentCallbackHandler.kt
emjburns: I was so stuck on working inside the `withBlocks` thing that I forgot it returned a list of blocks. 🤞

PR #1871, opened 2021-03-15
Disambiguate base OS of AMIs

robfletcher:
This PR ensures we always specify the base OS when looking for latest AMIs for a particular package. There are 4 situations where this is relevant:

1. when we check to see if something is already baked before launching a bake task
2. the implicit constraint that checks an appropriate AMI exists before promoting an artifact version to an environment
3. when materializing the desired state of a cluster
4. when launching a canary

In all those scenarios we want to be sure that we're selecting an image with the correct base OS according to the artifact definition in the delivery config. If a user updates their delivery config to use a new base OS we want to consider that a delta, re-bake, and re-deploy.

This also simplifies the logic around determining if we need to bake an AMI as there are basically only 3 situations where we should:

1. the user has a new release of their code (new artifact version)
2. there is a new base image
3. for some reason not all the regions we need are currently baked.

Previously we were pulling down all images for the artifact from CloudDriver which was sometimes quite slow and totally unnecessary since we know exactly what artifact version we need an image for.

### DIFF
--- a/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
+++ b/keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
@@ -1,7 +1,7 @@
 package com.netflix.spinnaker.keel.bakery.artifact
 
+import com.netflix.frigga.ami.AppVersion
 import com.netflix.spinnaker.keel.actuation.ArtifactHandler
-import com.netflix.spinnaker.keel.api.ResourceDiff
 import com.netflix.spinnaker.keel.api.actuation.Task
 import com.netflix.spinnaker.keel.api.actuation.TaskLauncher
 import com.netflix.spinnaker.keel.api.artifacts.DEBIAN
@@ -10,9 +10,9 @@ import com.netflix.spinnaker.keel.api.events.ArtifactRegisteredEvent
 import com.netflix.spinnaker.keel.artifacts.DebianArtifact
 import com.netflix.spinnaker.keel.bakery.BaseImageCache
 import com.netflix.spinnaker.keel.clouddriver.ImageService
-import com.netflix.spinnaker.keel.clouddriver.model.Image
+import com.netflix.spinnaker.keel.clouddriver.getLatestNamedImages
+import com.netflix.spinnaker.keel.clouddriver.model.baseImageName
 import com.netflix.spinnaker.keel.core.NoKnownArtifactVersions
-import com.netflix.spinnaker.keel.diff.DefaultResourceDiff
 import com.netflix.spinnaker.keel.igor.artifact.ArtifactService
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEvent
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEventScope.PRE_DEPLOYMENT
@@ -20,7 +20,7 @@ import com.netflix.spinnaker.keel.lifecycle.LifecycleEventStatus.NOT_STARTED
 import com.netflix.spinnaker.keel.lifecycle.LifecycleEventType.BAKE
 import com.netflix.spinnaker.keel.model.Job
 import com.netflix.spinnaker.keel.parseAppVersion
-import com.netflix.spinnaker.keel.persistence.DiffFingerprintRepository
+import com.netflix.spinnaker.keel.persistence.BakedImageRepository
 import com.netflix.spinnaker.keel.persistence.KeelRepository
 import com.netflix.spinnaker.keel.persistence.NoSuchArtifactException
 import com.netflix.spinnaker.keel.telemetry.ArtifactCheckSkipped
@@ -30,9 +30,9 @@ import org.springframework.context.ApplicationEventPublisher
 class ImageHandler(
   private val repository: KeelRepository,
   private val baseImageCache: BaseImageCache,
+  private val bakedImageRepository: BakedImageRepository,
   private val igorService: ArtifactService,
   private val imageService: ImageService,
-  private val diffFingerprintRepository: DiffFingerprintRepository,
   private val publisher: ApplicationEventPublisher,
   private val taskLauncher: TaskLauncher,
   private val defaultCredentials: BakeCredentials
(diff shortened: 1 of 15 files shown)


keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -32,66 +31,63 @@ class ImageHandler(
...
-          log.debug("Existing image for {} is up-to-date", artifact.name)
         }
       }
     }
   }
 
-  private suspend fun DebianArtifact.findLatestAmi() =
-    imageService.getLatestImage(this, "test")
+  private suspend fun DebianArtifact.findLatestAmi(desiredArtifactVersion: String) =
+    imageService.getLatestNamedImages(AppVersion.parseName(desiredArtifactVersion), "test", vmOptions.regions)
```
robfletcher, line 90: Now just using the same method that `ImageExistsConstraint` and `ImageResolver` use to get the AMIs for a specific app version

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -32,66 +31,63 @@ class ImageHandler(
...
-        if (current != null && diff.isRegionsOnly()) {
-          if (current.regions.containsAll(desired.regions)) {
-            log.debug("Image for ${current.appVersion} contains more regions than we need, which is fine")
-          } else {
-            log.warn("Detected a diff in the regions for ${current.appVersion}: ${diff.toDebug()}")
-            publisher.publishEvent(ImageRegionMismatchDetected(current, artifact.vmOptions.regions))
+        val desiredBaseAmiName = artifact.findLatestBaseAmiName()
+
+        val images = artifact.findLatestAmi(desiredAppVersion)
+        val olderBaseImages = images.filterValues { it.baseImageName != desiredBaseAmiName }
```
emjburns, line 56: the variable name is slightly misleading here - this is images for the artifact version where the base is old. maybe rename to `imagesWithOlderBaseImages`? when reading this just now I thought it might actually be the base images themselves.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -32,66 +31,63 @@ class ImageHandler(
...
+        when {
+          images.isEmpty() -> {
+            log.info("No AMI found for {}", desiredAppVersion)
+            launchBake(artifact, desiredAppVersion)
           }
-        } else if (diff.hasChanges()) {
-          if (current == null) {
-            log.info("No AMI found for {}", artifact.name)
-          } else {
-            log.info("Image for {} delta: {}", artifact.name, diff.toDebug())
```
emjburns, line 76: Won't this logic trigger if ever there is a version that was baked with two different base images? Because you're only checking if the older one is populated, not if the older one is populated and the newer one isn't?

robfletcher: I think you're right. Fixing

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -32,66 +31,63 @@ class ImageHandler(
...
-          } else {
-            log.warn("Detected a diff in the regions for ${current.appVersion}: ${diff.toDebug()}")
-            publisher.publishEvent(ImageRegionMismatchDetected(current, artifact.vmOptions.regions))
+        val desiredBaseAmiName = artifact.findLatestBaseAmiName()
+
+        val images = artifact.findLatestAmi(desiredAppVersion)
+        val olderBaseImages = images.filterValues { it.baseImageName != desiredBaseAmiName }
+        val missingRegions = artifact.vmOptions.regions - images.keys
+        when {
+          images.isEmpty() -> {
```
lorin, line 59: I don't think this clause of the `when` block is strictly needed, since when `images` is empty,  the `missingRegions` variable will be populated with the regions that need baking.

keel-clouddriver/src/test/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageServiceTests.kt
```diff
@@ -439,7 +465,55 @@ class ImageServiceTests : JUnit5Minutests {
...
+        }
+
+        expectThat(images)
+          .hasSize(2)
+          .get(Map<*, NamedImage>::values)
+          .map { it.baseImageName }
+          .all {
+            isEqualTo(imageWithNewerBaseImageName.baseImageName)
+          }
       }
```
robfletcher, line 517: Tests for filtering images by base OS

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -183,18 +193,23 @@ class ImageService(
...
+ * @param baseOs if supplied only images with that base OS are considered.
  */
 suspend fun ImageService.getLatestNamedImages(
   appVersion: AppVersion,
   account: String,
-  regions: Collection<String>
+  regions: Collection<String>,
+  // TODO: make this non-nullable
+  baseOs: String? = null
 ): Map<String, NamedImage> = coroutineScope {
```
robfletcher, line 205: Added a new `baseOs` param here. There are other places this method is used but I think all of them should specify a `baseOs`, hence the `TODO` comment

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -86,11 +86,12 @@ class ImageService(
   private data class NamedImageCacheKey(
     val appVersion: AppVersion,
     val account: String,
-    val region: String
+    val region: String,
+    val baseOs: String?
   )
```
robfletcher, line 91: base OS is now part of the cache key

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -102,6 +103,12 @@ class ImageService(
         .filter { image ->
           tagsExistForAllAmis(image.tagsByImageId) && image.hasAppVersion
         }
+        // filter to images with the correct base os
+        .filter { image ->
+          // using the xxxxbase pattern means we won't get matches for base OS values that are
+          // substrings of others -- e.g. bionic and bionic-classic
+          baseOs == null || image.baseImageName.startsWith("${baseOs}base")
+        }
```
robfletcher, line 111: This filters the images returned by River into only those with the specified base OS (if it _was_ specified). Sadly I don't think there's a better way to check the base OS than this. River is extracting base image name and base AMI id from the `description` field in the image itself but I don't see any clearly delimited base OS value in there otherwise I would add a new tag in River.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
robfletcher: Fixed

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
robfletcher: While that's true, we raise an event if there are just _some_ missing regions (because we want to know that's happening as it implies something other than MD is baking this app), and I customized the task description in that case too.

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -30,68 +30,87 @@ import org.springframework.context.ApplicationEventPublisher
...
-          if (current.regions.containsAll(desired.regions)) {
-            log.debug("Image for ${current.appVersion} contains more regions than we need, which is fine")
-          } else {
-            log.warn("Detected a diff in the regions for ${current.appVersion}: ${diff.toDebug()}")
-            publisher.publishEvent(ImageRegionMismatchDetected(current, artifact.vmOptions.regions))
+        val byArtifactVersion =
+          artifact.wasPreviouslyBakedWith(desiredAppVersion, desiredBaseAmiName)
+        if (byArtifactVersion) {
+          return
+        }
```
robfletcher, line 61: Just nope out of this whole check if we know we've already done this. There's a possibility that River's cache has yet to update fully and we might unintentionally launch additional tasks that will just no-op (in fact that's been a long-standing gripe with this class).

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
```diff
@@ -30,68 +30,87 @@ import org.springframework.context.ApplicationEventPublisher
...
-    imageService.getLatestImage(this, "test")
+  private fun DebianArtifact.wasPreviouslyBakedWith(
+    desiredAppVersion: String,
+    desiredBaseAmiName: String
+  ) =
+    bakedImageRepository
+      .getByArtifactVersion(desiredAppVersion, this)
+      ?.let {
+        it.baseAmiName == desiredBaseAmiName && it.amiIdsByRegion.keys.containsAll(vmOptions.regions)
+      } ?: false
```
robfletcher, line 105: Quick check against our `baked_images` table to see if we've already done combination of app version and base image in all the regions we need

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/artifact/ImageHandler.kt
robfletcher: Fixed

keel-bakery-plugin/src/main/kotlin/com/netflix/spinnaker/keel/bakery/constraint/ImageExistsConstraintEvaluator.kt
```diff
@@ -54,7 +54,8 @@ class ImageExistsConstraintEvaluator(
       imageService.getLatestNamedImages(
         appVersion = version.parseAppVersion(),
         account = defaultImageAccount,
-        regions = vmOptions.regions
+        regions = vmOptions.regions,
+        baseOs = vmOptions.baseOs
       )
     }
```
robfletcher, line 60: I've added `baseOs` wherever this method is used

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -86,11 +86,12 @@ class ImageService(
   private data class NamedImageCacheKey(
     val appVersion: AppVersion,
     val account: String,
-    val region: String
+    val region: String,
+    val baseOs: String
   )
```
robfletcher, line 91: Base OS is now part of this cache key

keel-clouddriver/src/main/kotlin/com/netflix/spinnaker/keel/clouddriver/ImageService.kt
```diff
@@ -102,6 +103,12 @@ class ImageService(
         .filter { image ->
           tagsExistForAllAmis(image.tagsByImageId) && image.hasAppVersion
         }
+        // filter to images with the correct base os
+        .filter { image ->
+          // using the xxxxbase pattern means we won't get matches for base OS values that are
+          // substrings of others -- e.g. bionic and bionic-classic
+          image.baseImageName.startsWith("${baseOs}base")
+        }
```
robfletcher, line 111: Filter images from River to only those that have the correct base OS

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/constraints/CanaryConstraintDeployHandler.kt
```diff
@@ -27,6 +28,7 @@ interface CanaryConstraintDeployHandler {
    */
   suspend fun deployCanary(
     constraint: CanaryConstraint,
+    artifact: DeliveryArtifact,
     version: String,
     deliveryConfig: DeliveryConfig,
     targetEnvironment: Environment,
```
robfletcher, line 34: had to add `artifact` to this method signature as in the EC2 case we'll need the base OS. It seems kinda weird that the artifact wasn't already being passed.

I considered making this class have a type parameter for the specific sub-type of `DeliveryArtifact` but it was a bit of a PITA.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/constraints/Ec2CanaryConstraintDeployHandler.kt
```diff
@@ -46,18 +48,24 @@ class Ec2CanaryConstraintDeployHandler(
...
     val scope = CoroutineScope(GlobalScope.coroutineContext)
     val judge = "canary:${deliveryConfig.application}:${targetEnvironment.name}:${constraint.canaryConfigId}"
 
     val images = imageService.getLatestNamedImages(
       appVersion = version.replace("~", "_").parseAppVersion(),
       account = imageResolver.defaultImageAccount,
-      regions = regions
+      regions = regions,
+      baseOs = artifact.vmOptions.baseOs
     )
```
robfletcher, line 69: Another place this method was used -- deploying canaries. Makes sense to get the right base OS.

keel-ec2-plugin/src/main/kotlin/com/netflix/spinnaker/keel/ec2/resolvers/ImageResolver.kt
```diff
@@ -85,7 +85,8 @@ class ImageResolver(
     val images = imageService.getLatestNamedImages(
       appVersion = artifactVersion.parseAppVersion(),
       account = account,
-      regions = regions
+      regions = regions,
+      baseOs = artifact.vmOptions.baseOs
     )
```
robfletcher, line 90: This is pretty critical -- when we resolve the desired state of a cluster, make sure we're looking for an image with the right base OS.

emjburns (approved):
LGTM, I like the choices you made around using the baked images information and adding baseos everywhere

PR #1873, opened 2021-03-15
chore(env): add timer for envt exclusion checks

lorin:
Use a percentile timer to record the time spent doing the EnvironmentExclusionEnforcer checks, so we can identify if this code leads to performance issues.

Other than the timer, this PR doesn't introduce any (intended) changes in behavior.

### DIFF
--- a/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcerTest.kt
+++ b/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcerTest.kt
@@ -1,5 +1,6 @@
 package com.netflix.spinnaker.keel.enforcers
 
+import com.netflix.spectator.api.NoopRegistry
 import io.mockk.coVerify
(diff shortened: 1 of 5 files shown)


keel-core/src/test/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcerTest.kt
```diff
@@ -30,6 +31,7 @@ internal class EnvironmentExclusionEnforcerTest {
     }
   }
 
+  @kotlinx.coroutines.ExperimentalCoroutinesApi
```
lorin, line 34: This annotation was added to suppress a warning because I'm using the `runBlockingTest` construct in this test.

emjburns (approved):
Do you think the code will spend a non minuscule amount of time here?

lorin:
> Do you think the code will spend a non minuscule amount of time here?

I hope it will be a small amount of time, but I worry about a database interaction spending way longer than I anticipated, once I add the actual lease logic.

PR #1875, opened 2021-03-16
chore(env): implement ensureNoActiveVerifications

lorin:
Implement the EnvironmentExclusionEnforcer.ensureNoActiveVerifications method. This is progress towards the work started in #1865.

## Main changes

* EnvironmentExclusionEnforcer.ensureNoActiveVerifications is now implemented
* new SqlVerificationRepository.countVerifications method
* EnvironmentExclusionEnforcer.withActuationLease now runs checks inside of IO dispatcher

## Other changes 

* Change how spectator (atlas) timer is called. I was previously passing a function to spectator, but now instead I'm using traditional start/end timestamps for invoking the timer.
* Switch from the experimental `runBlockingTest` to the regular `runBlocking` in test code

### DIFF
--- a/keel-core/keel-core.gradle.kts
+++ b/keel-core/keel-core.gradle.kts
@@ -47,6 +47,5 @@ dependencies {
   testImplementation("dev.minutest:minutest")
 
   testImplementation("org.assertj:assertj-core")
-  testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-test")
   testImplementation("org.springframework.boot:spring-boot-test-autoconfigure")
 }
--- a/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcerTest.kt
+++ b/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcerTest.kt
@@ -1,21 +1,64 @@
 package com.netflix.spinnaker.keel.enforcers
 
 import com.netflix.spectator.api.NoopRegistry
+import com.netflix.spinnaker.keel.api.constraints.ConstraintStatus
+import com.netflix.spinnaker.keel.api.verification.VerificationRepository
+import io.mockk.Called
 import io.mockk.coVerify
-import kotlinx.coroutines.test.runBlockingTest
 import io.mockk.every
 import io.mockk.mockk
 import io.mockk.verify
+import kotlinx.coroutines.runBlocking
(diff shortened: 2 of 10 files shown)


keel-core/src/test/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcerTest.kt
```diff
@@ -31,15 +74,14 @@ internal class EnvironmentExclusionEnforcerTest {
...
   }
 
-  @kotlinx.coroutines.ExperimentalCoroutinesApi
   @ParameterizedTest(name="withActuationLease executes action, enabled={0}")
   @ValueSource(booleans = [true, false])
   fun `withActuationLease invokes the action whether the feature flag is enabled or not`(enabled: Boolean) {
     every { springEnv.getProperty("keel.enforcement.environment-exclusion.enabled", Boolean::class.java, any())} returns enabled
 
     val action : suspend () -> Unit = mockk(relaxed=true)
-    runBlockingTest {
```
lorin, line 41: Was getting an occasional error with `runBlockingTest`, so switched to `runBlocking` instead.

keel-core/keel-core.gradle.kts
```diff
@@ -47,6 +47,5 @@ dependencies {
   testImplementation("dev.minutest:minutest")
 
   testImplementation("org.assertj:assertj-core")
-  testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-test")
```
lorin, line 50: Because `runBlockingTest` is no longer being used, don't need this anymore.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcerTest.kt
robfletcher: I've never seen `runBlockingTest` before. I'll have to look into that

lorin: It's like runBlocking, but it does things like ignore delays.

keel-core/keel-core.gradle.kts
emjburns: good riddance

emjburns (approved):

PR #1877, opened 2021-03-16
fix(plugins): Add default 'name' getter for ConstraintEvaluator

luispollo:
Attempt to fix exception loading external `ConstraintEvaluator` plugin:
```
java.lang.ClassCastException: class com.sun.proxy.$Proxy160 cannot be cast to class com.netflix.spinnaker.keel.api.plugins.ConstraintEvaluator (com.sun.proxy.$Proxy160 is in unnamed module of loader org.pf4j.PluginClassLoader @345292c1; com.netflix.spinnaker.keel.api.plugins.ConstraintEvaluator is in unnamed module of loader 'app')
```


emjburns (approved):

jonsie:
How/where is `ConstraintEvaluator#name` called in Keel?  Curious how this will resolve the issue?

luispollo:
> How/where is `ConstraintEvaluator#name` called in Keel? Curious how this will resolve the issue?

Uh... I was just following your suggestion. 😆

PR #1879, opened 2021-03-16
fix(constraints): use reference to find constraint when saving

emjburns:
we need to use the reference field here since version is not guaranteed to be unique across all artifacts


emjburns:
This pr is going to get much larger, the reference field needs to be non nullable and part of the primary key. what fun.

PR #1880, opened 2021-03-17
fix(constraints): require artifact reference in constraint state

emjburns:
An artifact version isn't unique across artifacts, but an artifact reference is unique within a delivery config. This pr addresses a longstanding bug where we didn't have the artifact reference as part of the unique constraint on the `environment_artifact_constraint` table.

We've been recording artifact reference in the database since august. In this pr I'm deleting all columns where artifact reference is null. I think this is fine since the data is months old. I tested the migrations locally and it worked just fine (I inserted one row where the artifact reference was null)!

### DIFF
--- a/keel-sql/src/main/resources/db/changelog/20210316-update-env-artifact-constr.yml
+++ b/keel-sql/src/main/resources/db/changelog/20210316-update-env-artifact-constr.yml
@@ -1,0 +1,28 @@
+databaseChangeLog:
+  - changeSet:
+      id: drop-null-artifact-reference
+      author: emjburns
+      changes:
+        - sql:
+            dbms: mysql
+            sql: DELETE FROM environment_artifact_constraint WHERE artifact_reference is NULL
+  - changeSet:
+      id: non-null-artifact-reference
+      author: embjurns
+      changes:
+        - addNotNullConstraint:
+            tableName: environment_artifact_constraint
+            columnName: artifact_reference
+            columnDataType: varchar(150)
+  - changeSet:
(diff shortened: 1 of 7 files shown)


keel-sql/src/main/resources/db/changelog/20210316-update-env-artifact-constr.yml
```diff
@@ -0,0 +1,28 @@
...
+            columnDataType: varchar(150)
+  - changeSet:
+      id: new-unique-constraint
+      author: emjburns
+      changes:
+        - dropUniqueConstraint:
+            tableName: environment_artifact_constraint
+            constraintName: constraint_pk
+            uniqueColumns: environment_uid, type, artifact_version
+        - addUniqueConstraint:
```
robfletcher, line 25: Shouldn't that be [`addPrimaryKey`](https://docs.liquibase.com/change-types/community/add-primary-key.html)?

robfletcher (approved):
One question but LGTM otherwise

keel-sql/src/main/resources/db/changelog/20210316-update-env-artifact-constr.yml
emjburns: no, the primary key is actually the uuid. it's the unique constraint i needed to update. i was wrong when I pinged you earlier

emjburns: updated to not have pk in the name

PR #1882, opened 2021-03-18
chore(env): implement ensureNoActiveDeployments

lorin:
Implement the EnvironmentExclusionEnforcer.ensureNoActiveDeployments method.

This is progress towards fixing verification race conditions, continuing of work from #1865, #1875.

## Changes

* EnvironmentExclusionEnforcer.ensureNoActiveDeployments is now implemented
* new SqlArtifactRepository.isDeployingTo method

## Notes

Something we have to watch out for: if an `environment_artifact_versions` record somehow gets stuck in the `DEPLOYING` state, it will block subsequent verifications.

Longer term, we'll want better visibility into this.


emjburns:
Re: your note. Do you think you could also create an admin API that updates a version in the `DEPLOYING` status to `SKIPPED`? I'd like to avoid a situation (if possible) where the only way to get a version deployed is by hand updating the database. I agree we will need to address this problem long term (maybe we add a `FAILED` status and something that marks the versions as failed? maybe we fix the vetoing?) but that is obviously outside the scope here

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcer.kt
```diff
@@ -110,12 +114,13 @@ class EnvironmentExclusionEnforcer(
...
+   * @throws ActiveDeployments  if there's an active deployment
    */
   private fun ensureNoActiveDeployments(deliveryConfig: DeliveryConfig, environment: Environment) {
-    /**
-     * To be implemented in a future PR.
-     */
+    artifactRepository.isDeployingTo(deliveryConfig, environment.name)
+    if(artifactRepository.isDeployingTo(deliveryConfig, environment.name)) {
+      throw ActiveDeployments(deliveryConfig, environment)
+    }
```
robfletcher, line 123: This feels like using exceptions for control flow. Is this really an exception?

lorin:
> Re: your note. Do you think you could also create an admin API that updates a version in the DEPLOYING status to SKIPPED? I'd like to avoid a situation (if possible) where the only way to get a version deployed is by hand updating the database. I agree we will need to address this problem long term (maybe we add a FAILED status and something that marks the versions as failed? maybe we fix the vetoing?) but that is obviously outside the scope here

Yeah, that's a great idea. I'll create an admin API for both `DEPLOYING` versions and `PENDING` verifications (the latter one just  bit us today)

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcer.kt
lorin: I think this gets at a deep philosophical question. Here's my argument for using an exception here.

While "a deployment is active" is not exceptional from the perspective of keel as a whole, I think it is an exceptional thing to try do a verification while a deployment is happening. It's not a combination of events we typically expect to happen. I

I think that the "happy path" is that the enforcer's methods (`withVerificationLease`, `withActuationLease`) succeed.

Aesthetically, I also prefer the code structure when throwing the exception because then we can pass the exception -related info up the stack to the caller in a nice way.

Note, though, that I was a Python programmer in a previous life, where exceptions for control flow are so common that they're used for terminating loops! On the other hand, before that I was a C++ programmer, where using exceptions for control flow is severley frowned upon.

lorin:
@emjburns I added new endpoints to the admin controller.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcer.kt
```diff
@@ -109,12 +114,13 @@ class EnvironmentExclusionEnforcer(
   }
 
   /**
-   * @throws EnvironmentCurrentlyBeingActedOn if there's an active deployment
+   * @throws ActiveDeployments  if there's an active deployment
    */
   private fun ensureNoActiveDeployments(deliveryConfig: DeliveryConfig, environment: Environment) {
-    /**
-     * To be implemented in a future PR.
-     */
```
emjburns, line 117: why are you calling `artifactRepository.isDeployingTo` twice instead of just saving the result?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
```diff
@@ -113,4 +116,37 @@ class AdminService(
...
 
+  fun forceSkipArtifactVersion(application: String, environment: String, artifactReference: String, version: String) {
+    val deliveryConfig = repository.getDeliveryConfigForApplication(application)
+    val artifact = deliveryConfig.matchingArtifactByReference(artifactReference)
+      ?: throw UserException("application $application contains no artifact ref $artifactReference. Artifact references are: ${deliveryConfig.artifacts.map { it.reference }}")
+
+     // Identify the current version in the environment
+    val currentVersion = repository.getCurrentArtifactVersions(deliveryConfig, environment)
+      .firstOrNull { it.reference == artifactReference }
+      ?: throw UserException("application $application has no artifact with reference $artifactReference in environment $environment in CURRENT status")
```
emjburns, line 127: Can we ever be in a state where there is no `current` artifact but one is stuck in deploying?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/enforcers/EnvironmentExclusionEnforcer.kt
lorin: Ugh, it's a typo (thinko?).  Thanks for spotting that!

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/AdminService.kt
lorin: I think it is possible for the system to be in that state (e.g., if the first deploy gets stuck). If that happens, I'm not sure what state we would want to force the stuck version into. "Skipped" doesn't feel like the right state.

I think what I'll do is also add an API endpoint that lets the admin choose what state to force the version into. That way, the operator can decide in the moment what makes the most sense.

emjburns: Okay - but this api call would fail because a precondition is not true (there is no current version). We can either deal with that later, or you could add a `force: Boolean` flag which does not care if there is no current version

emjburns (approved):

PR #1883, opened 2021-03-18
fix(plugins): Convert StatefulConstraintEvaluator and StatelessConstraintEvaluator to interfaces

luispollo:
As seen previously with `ResourceHandler` (#1404), the plugin framework cannot proxy abstract classes. This PR converts `StatefulConstraintEvaluator` and `StatelessConstraintEvaluator` to interfaces to allow for 3rd-party extensions.


emjburns (approved):
lgtm. no behavior change, right?

luispollo:
@emjburns That's right, no behavior change!

PR #1885, opened 2021-03-18
fix(resource): don't show error if verification prevents actuation

lorin:
## Problem

If a resource was prevented from being actuation because there was a verification running in the environment, we show the user an error in the event history for the resource. 

Pasted image 20210318120320

We shouldn't do this, because this isn't really an error state, it's something that will resolve on its own once the verification completes.

### What the user will see

If verification blocks actuation, the user will see this text in the event history at the INFO level:

> Waiting for verification to complete before updating can start

The detailed message in the event log looks like this:

> There is an active verification against version fnord-0.950.0-h1278.3cd2ac9.

or, if there are verifications running against multiple versions in the environment (which should never really happen, but we might get "stuck" verifications)

> There are active verifications against versions [fnord-0.950.0-h1278.3cd2ac9, fnord-0.951.0-h1279.4a20bb3].

### Implementation details

If the `ResourceActuator` tries to actuate a resource in an environment where verifications are running, the actuator emits a `VerificationBlockedActuation` event. This event subclasses from ResourceEvent, which ensures that the `ResourceHistoryListener`

Move the exception-handling logic so that it catches the exception that gets thrown when actuation fails due to verification. Don't treat this as an error anymore.

### New metrics

`keel.actuation.blocked` - counter that is incremented whenever a `VerificationBlockedActuation` event is emitted.  This metric is tagged with resourceId, resourceKind, and resourceApplication.

### Other changes

I changed the method we use to query the VerificationRepository (there's a new method, called `getContextsWithStatus`) so that we could report the versions being verified to the user.

### Docs

I've started to incrementally add developer docs to the docs directory as I work.


emjburns:
How will this show up to the user now?

robfletcher:
I'm not really clear from looking at the code how this changes things. It looks like you stopped an exception from propagating, does that prevent it from flagging an error to the user?

lorin:
@robfletcher asked:
> I'm not really clear from looking at the code how this changes things. It looks like you stopped an exception from propagating, does that prevent it from flagging an error to the user?

The main difference is that the code no longer publishes a `ResourceEvent` object.

The previous behavior was:

1. The `catch(e: Exception)` block publishes a `ResourceCheckError` object. 
2. Because `ResourceCheckError` is a subclass of `ResourceEvent`, it would get handled by `ResourceHistoryListener.onResourceEvent`.
3. This would append the event to the history.

With my change, the code no longer publishes a `ResourceEvent` object. (It only publishes a `TelemetryEvent` object to increment a metric).


@emjburns asked:

> How will this show up to the user now?

With this implementation, it actually doesn't show up at all, it's completely invisible to the user. Thinking about it, I think this behavior is bad. I think it would be better if we showed in the resource history an info level message like:

"Waiting for verifications to complete before updating can start."

That way, users can get some feedback in the history about what's keeping the resource from updating.

I'm going to update the PR to do that.

emjburns:
@lorin Thanks so much, I think the transparent behavior you described is much better.

lorin:
@emjburns I updated the PR description with details of what the user will see when verifications block actuations.

PR #1886, opened 2021-03-18
feat(mj) don't time out manual judgement

emjburns:
Timing out a manual judgement is left over from the pipeline implementation. with pipelines, we didn't want a stage running indefinitely. But MD doesn't have that technical constraint. So, let's just not time out manual judgements. It's frustrating and confusing to users and isn't getting us anything.

I've changed the default to be no time out. is that an appropriate default, or should we have a timeout as default?

### DIFF
--- a/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluatorTests.kt
+++ b/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluatorTests.kt
@@ -14,6 +14,8 @@ import com.netflix.spinnaker.keel.test.resource
 import com.netflix.spinnaker.time.MutableClock
 import dev.minutest.junit.JUnit5Minutests
 import dev.minutest.rootContext
+import io.mockk.clearAllMocks
+import io.mockk.clearMocks
 import io.mockk.mockk
 import io.mockk.verify
 import java.time.Duration
@@ -21,7 +23,7 @@ import strikt.api.expectThat
 import strikt.assertions.isFalse
 
 internal class ManualJudgementConstraintEvaluatorTests : JUnit5Minutests {
-  object Fixture {
+  class Fixture {
     val clock = MutableClock()
     val repository = mockk<ConstraintRepository>(relaxUnitFun = true)
     val publisher: EventPublisher = mockk()
@@ -63,7 +65,7 @@ internal class ManualJudgementConstraintEvaluatorTests : JUnit5Minutests {
   }
 
   fun tests() = rootContext<Fixture> {
-    fixture { Fixture }
+    fixture { Fixture() }
 
     context("evaluating state") {
       test("before timeout we don't pass") {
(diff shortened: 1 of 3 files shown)


lorin (approved):
I think "no timeout" should be the default.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluatorTests.kt
```diff
@@ -65,6 +67,8 @@ internal class ManualJudgementConstraintEvaluatorTests : JUnit5Minutests {
   fun tests() = rootContext<Fixture> {
     fixture { Fixture }
 
+    before { clearAllMocks() }
```
robfletcher, line 70: image

robfletcher (commented):
Makes sense

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluatorTests.kt
```diff
@@ -65,6 +67,8 @@ internal class ManualJudgementConstraintEvaluatorTests : JUnit5Minutests {
   fun tests() = rootContext<Fixture> {
     fixture { Fixture }
```
robfletcher, line 68:
```suggestion
    fixture { Fixture() }
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluatorTests.kt
```diff
@@ -65,6 +67,8 @@ internal class ManualJudgementConstraintEvaluatorTests : JUnit5Minutests {
...
 
+    before { clearAllMocks() }
+
```
robfletcher, line 71:
```suggestion
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluatorTests.kt
robfletcher: and change `object Fixture` above to `class Fixture` (can't figure out how to make suggestions on lines you didn't change in the PR)

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluatorTests.kt
emjburns: @robfletcher I thought of you when i typed this but it legit did not work before I did it and after I added it it worked. So what am I missing here????

robfletcher: I added suggestions below. It's because `Fixture` is an object rather than a class so all the tests end up using the same instance. Probably my fault.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluatorTests.kt
emjburns: oooo but that fixes my mocks problem! interesting

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/constraints/ManualJudgementConstraintEvaluatorTests.kt
emjburns: oh that's SOOOO great to finally understand.

robfletcher (approved):

PR #1887, opened 2021-03-18
fix(slack): launch a modal with the full commit

emjburns:
Launch an actual modal with the actual full commit (well, 2000 characters of it) instead of hacking the "close" dialog.
Screen Shot 2021-03-18 at 3 08 37 PM

### DIFF
--- a/keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/callbacks/CommitModalCallbackHandler.kt
+++ b/keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/callbacks/CommitModalCallbackHandler.kt
@@ -1,0 +1,24 @@
+package com.netflix.spinnaker.keel.slack.callbacks
+
+import com.netflix.spinnaker.keel.slack.handlers.GitDataGenerator
+import com.slack.api.app_backend.interactive_components.payload.BlockActionPayload
+import com.slack.api.model.view.View
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+
+/**
+ * A handler that builds the modal for launching the full commit modal
+ */
+@Component
+class CommitModalCallbackHandler(
+  private val gitDataGenerator: GitDataGenerator
+) {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  fun buildView(slackCallbackResponse: BlockActionPayload): View {
+    val message = slackCallbackResponse.actions.first().value
+    val hash = slackCallbackResponse.actions.first().actionId.split(":")[1]
+    return gitDataGenerator.buildFullCommitModal(message = message, hash = hash)
+  }
+}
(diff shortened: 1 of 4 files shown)


keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/callbacks/CommitModalCallbackHandler.kt
```diff
@@ -0,0 +1,24 @@
...
+@Component
+class CommitModalCallbackHandler(
+  private val gitDataGenerator: GitDataGenerator
+) {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  fun buildView(slackCallbackResponse: BlockActionPayload): View {
+    val message = slackCallbackResponse.actions.first().value
+    val hash = slackCallbackResponse.actions.first().actionId.split(":")[1]
```
luispollo, line 21: nit: Define an extension property like Gal did for the last piece, to make this easier to read.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
```diff
@@ -36,6 +41,25 @@ class GitDataGenerator(
...
+  /**
+   * Builds a modal with the full commit message
+   */
+  fun buildFullCommitModal(message: String, hash: String): View {
+    return view { thisView -> thisView
+      .callbackId("view:$hash:modal")
+      .type("modal")
+      .notifyOnClose(false)
+      .title(viewTitle { it.type("plain_text").text("Full commit").emoji(true) })
+      .close(viewClose { it.type("plain_text").text("Close").emoji(true) })
```
luispollo, line 53: You probably don't need `emoji(true)` since none of those have emojis. 🙂

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/SlackAppController.kt
```diff
@@ -34,12 +41,25 @@ class SlackAppController(
...
+        val modal: View = commitModalCallbackHandler.buildView(req.payload)
+        val response = ctx.client().viewsOpen { r ->
+          r
+            .token(slackConfiguration.token)
+            .triggerId(req.payload.triggerId)
+            .view(modal)
+        }
+        if (!response.isOk) {
+          log.error("Failed to send slack callback: {}", response)
+        }
```
luispollo, line 54: Kind of a nit, but I think the code would be cleaner if we encapsulated all this logic in `CommitModalCallbackHandler` in a method like `handleCallback` that takes the request and context and returns a response.

luispollo (approved):
Looks good! Gave you a couple of suggestions, nothing blocking.

keel-slack/src/main/kotlin/com/netflix/spinnaker/keel/slack/handlers/GitDataGenerator.kt
emjburns: oh, so that's what this was for! Just copying it. Will leave as is just in case I want to emojify it

keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/SlackAppController.kt
emjburns: extracted both this and the mj stuff into a method on the callback handlers

PR #1892, opened 2021-03-22
chore(artifacts): don't treat missing metadata on local artifact build as an error

lorin:
## Context

When a user builds a deb package locally, the build number is set to LOCAL. These builds are not tracked by the artifact metadata service. 

## Problem

For local builds, keel logs an ERROR level log message when it fails to retrieve the metadata.

## Implemented solution

When  keel tries and fails to retrieve artifact metadata for a local build, log at info level instead.

## Alternate solution

We could detect that it's a LOCAL build and not even make the artifact metadata call in the first place. The current implementation feels a little safer because there's no risk of not calling the service.


emjburns (approved):
Thanks!

PR #1915, opened 2021-04-09
feat(plugins): Add callback API for constraint state changes

luispollo:
Problem: plugin implementations of `StatefulConstraintEvaluator` may be interested in constraint state changes, but currently there's no contract defined on the interface to call them back when that happens. We've attempted to use the Spring `@EventListener` machinery with a `@PluginComponent` in the plugin but the callback method was not called, so we need an alternative solution.

Solution: this PR introduces a new `onConstraintStateChanged` method in `StatefulConstraintEvaluator` which receives a `ConstraintStateChanged` event when a stateful constraint's state is changed via the REST API.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
@@ -436,8 +436,8 @@ class ApplicationService(
   }
 
   private fun getVerificationStates(
-      deliveryConfig: DeliveryConfig,
-      artifactVersions: List<PublishedArtifact>
+    deliveryConfig: DeliveryConfig,
+    artifactVersions: List<PublishedArtifact>
   ) =
     if (verificationsEnabled) {
       repository.getVerificationStates(deliveryConfig, artifactVersions)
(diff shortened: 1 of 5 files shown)


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/services/ApplicationService.kt
```diff
@@ -165,10 +171,14 @@ class ApplicationService(
       environment,
       status.type
     )
```
emjburns, line 173: The way this is, you'll only notify plugins if a user changed the constraint state. We publish constraint state changed events in a couple of places, because sometimes the system changes the state.

I think it might be better to create a new component that listens to all `ConstraintStateChanged` events and then  notifies each stateful evaluator. That way you only have to add this logic in one place.

luispollo: Yup, good call! I had the same exact thought the minute after I pushed this. Will make that change!

luispollo: Updated the PR based on your suggestion.

emjburns (approved):

PR #1918, opened 2021-04-13
fix(bakery): remove hardcoded account name

dmart:
Replace the hardcoded "test" account with dynamicConfigService.


emjburns:
Ha! Yes thank you. 

We've switched to using the spring environment instead of the dynamic config service, could you update to use this format?

https://github.com/spinnaker/keel/blob/9363d1afc65b0708e1d13ef1663e85123b227ce2/keel-titus-plugin/src/main/kotlin/com/netflix/spinnaker/keel/titus/EnvironmentVariablesResolver.kt#L20-L21

It should do the same thing

dmart:
@emjburns something like this?

emjburns (approved):
Perfect!

PR #1926, opened 2021-04-28
fix(imageservice): replace hardcoded account name with springenv

dmart:


luispollo (approved):
Thanks for the contribution, @dmart. In the future, please follow the contributor guidelines and provide a description for your PRs.

PR #1954, opened 2021-06-30
fix(keel-web): rename keel-web -> keel

nimakaviani:
this name change happened when switch to
groovy-based gradle was made. keel-web breaks creating
the docker image for keel.

Also apply the packaging for springboot configuration
location to populate properly in the run script.

It might be that https://github.com/spinnaker/keel/commit/858c078faafc268ee628172cb9d675e0d66c2c33 impacted the way we used to package keel. 

@luispollo @lorin


luispollo (approved):
Sorry about that, and thanks for the fix!

PR #1966, opened 2021-07-14
fix(artifacts): remove hard coded artifact types

nabuskey:
This removes hard coded check for artifact types. Keel supports implementing custom artifact types except for changed part.

Removing the exception shouldn't be a problem since artifact types are validated during deserialization. 

```
➜  spinnaker/keel git:(bugfix/artifact-types) ✗ curl -X POST -d @test.json --header "Content-Type:application/json" localhost:7010/delivery-configs
---
message: "Could not resolve type id 'git' as a subtype of `com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact`:\
  \ known type ids = [deb, docker, npm] (for POJO property 'artifacts')\n at [Source:\
  \ (PushbackInputStream); line: 1, column: 138] (through reference chain: com.netflix.spinnaker.keel.core.api.SubmittedDeliveryConfig[\"\
  artifacts\"]->java.util.HashSet[0])"
details:
  error: "invalid_type"
  message: "Could not resolve type id 'git' as a subtype of `com.netflix.spinnaker.keel.api.artifacts.DeliveryArtifact`:\
    \ known type ids = [deb, docker, npm] (for POJO property 'artifacts')\n at [Source:\
    \ (PushbackInputStream); line: 1, column: 138] (through reference chain: com.netflix.spinnaker.keel.core.api.SubmittedDeliveryConfig[\"\
    artifacts\"]->java.util.HashSet[0])"
  location:
    line: 1
    column: 138
  path:
    field: "artifacts"
  - type: "array"
    index: 0
  type: "invalid_type"
  pathExpression: ".artifacts[0]"
timestamp: "2021-07-14T17:10:22.875733Z"
status: 400
error: "Bad Request"
```


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/artifacts/DeliveryArtifact.kt
```diff
@@ -96,6 +96,8 @@ abstract class DeliveryArtifact {
   @get:ExcludedFromDiff
   abstract val deliveryConfigName: String?
 
+  abstract fun copyWithName(name: String): DeliveryArtifact
```
luispollo, line 99: Consider renaming for accuracy:
```suggestion
  abstract fun withDeliveryConfigName(deliveryConfigName: String): DeliveryArtifact
```
Another alternative would be to just make `deliveryConfigName` a `var` instead of a `val`.

luispollo:
Good catch, @nabuskey! Just a small naming suggestion, otherwise looks good.

nabuskey:
Thanks for the review! Renamed the method and its argument as suggested.

luispollo (approved):

PR #1934, opened 2021-05-14
feat(multiple): batch of changes

lorin:
We recently changed our development model at Netflix, which means that we now need to remember to push our changes up to GitHub. For details on the change, see the blog post [Evolving How Netflix Builds, Maintains, and Operates Their Spinnaker Distribution](https://blog.spinnaker.io/evolving-how-netflix-builds-maintains-and-operates-their-spinnaker-distribution-33c844d0102c).

Our goal is generally to push up our changes incrementally, but we fell a bit behind, and so I'm pushing them up as a batch to bring us back into sync.

Recent changes: 

* dry run mode for testing
* slack mj notifications: add number behind current
* preview environment code handling
* start at environment versioning
* verifications -> actions
* support for post deploy actions
* api conversion to dgs progress
* verification log link shows while verification is running
* retrieve delivery configs from commits
* cache application data from front50
* thread pool moitoring metrics
* baking when base image changes fix
* flakey sql test fix (1000 != 999)
* evaluate fewer versions in the environment promotion checker
* evaluate only pending versions in the environment promotion checker (flagged to off)
* mark all past versions as skipped instead of leaving them as pending
* lifecycle events artifact key to two separate fields
* build url lifecycle event fix
* preview environments cluster VIP overrides support
* don't bake if the app is paused
* DELETE delivery config moved to async
* simplify artifact metadata scheduled backfill
* remove failed build notification
* support for authenticate-oidc actions on ALB listeners

PR #1967, opened 2021-07-20
fix(imageservice): remove hardcoded account name

dmart:
Replace the hardcoded account "test" with a spring environment reference.


dmart:
This was previously fixed in https://github.com/spinnaker/keel/pull/1926 and reverted in https://github.com/spinnaker/keel/pull/1934.

It is unclear to me if 1926 was intentionally reverted has there is no mention of it in the commit message.

luispollo:
Hi @dmart, thanks for the heads-up. We may have missed picking up the original change downstream. @lorin is looking at syncing our internal OSS fork with upstream this week, which hopefully will make things better (we might need to pick up this one too right after merging!).

luispollo (approved):
LGTM and this had been previously reviewed/merged.

PR #1989, opened 2021-08-21
chore(upstream): Upstream latest batch of changes from Netflix

luispollo:
Includes commits from the last several weeks. This is practically impossible to review -- the idea with these PRs is to trust that they've been properly vetted internally at Netflix before upstreaming. Not a great experience for the community, I know, but that's the best process we could come up with for the time being.

@spinnaker/managed-delivery If you could please just take a look over the commits to see if anything stands out to you as something that shouldn't be included, that would be helpful.

@nimakaviani @nabuskey FYI.


luispollo:
I'll resolve the conflicts as soon as possible.

PR #2000, opened 2021-10-11
feat(security groups): support security group rules that reference prefix lists

mrobinet:
AWS added a new feature called Managed Prefix Lists a little over a year ago that are lists of either IPv4 or IPv6 CIDR blocks that can be easily referenced in security groups and route tables.

* [Simplify network routing and security administration with VPC Prefix Lists](https://aws.amazon.com/blogs/networking-and-content-delivery/simplify-network-routing-and-security-administration-with-vpc-prefix-lists/) 
* [Prefix Lists Documentation](https://docs.aws.amazon.com/vpc/latest/userguide/managed-prefix-lists.html)

This PR adds the ability for Keel to create security group rules that reference managed prefix lists rather than other security groups or individual CIDR blocks.

I read the contributing guide after implementing this and probably should have started by filing a feature request. 😬 If you'd like me to close this and file a feature request first to discuss it I would be happy to. I'm hoping this change is fairly non-controversial though as it merely closes a gap between what AWS supports for security group rules and what Keel supports.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,72 @@
+pull_request_rules:
+  - name: Automatically merge on CI success and review
+    conditions:
+      - base=master
+      - status-success=build
+      - "label=ready to merge"
+      - "approved-reviews-by=@oss-approvers"
+    actions:
+      merge:
+        method: squash
+        strict: smart
+      label:
+        add: ["auto merged"]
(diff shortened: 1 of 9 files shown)


nabuskey:
@nimakaviani You may want to run the build before merging. It built just fine on my computer but just in case.

PR #2004, opened 2021-11-01
fix(artifacts): use normalized artifact versions

dmart:
Use normalized version when comparing latest available artifact and the current artifact from the DB.


dmart:
This manifested itself as large numbers of `SqlDeliveryConfigRepository` errors `Artifact <name> version <version> is already associated with <app/env>` and unnecessary calls to Igor to populate metadata for artifacts which already existed in the DB.

nimakaviani (approved):
LGTM

PR #2024, opened 2022-04-04
fix(artifacts): Add support for github enterprise

### LINKED ISSUE #2023: Support github enterprise
Right now only github.com is supported. Github enterprise has the same api's for diffs between commits. It should be easy to support if url matching was smarter.

https://github.com/spinnaker/keel/blob/9e228c182c2ac2cc52409512bf79fe6aea3869b0/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt#L40-L49

https://github.com/spinnaker/keel/blob/9e228c182c2ac2cc52409512bf79fe6aea3869b0/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt#L61-L69

jessesanford:
Signed-off-by: Jesse Sanford <jesse.sanford@autodesk.com>

Adds support for github enterprise. Currently the gui shows up broken because it doesn't understand URLs that don't contain "github". This will allow those as long as they are "mapped to github" in the list of scm accounts.

### DIFF
--- a/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
+++ b/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
@@ -35,14 +35,15 @@ class ArtifactVersionLinks(private val scmInfo: ScmInfo, private val cacheFactor
   //Generating a SCM compare link between source (new version) and target (old version) versions (the order matter!)
   private fun generateCompareLink(newerGitMetadata: GitMetadata?, olderGitMetadata: GitMetadata?): String? {
     val commitLink = newerGitMetadata?.commitInfo?.link ?: return null
+    val normScmType = getNormalizedScmType(commitLink)
     val baseScmUrl = getScmBaseLink(commitLink)
-    return if (baseScmUrl != null && olderGitMetadata != null && !(olderGitMetadata.commitInfo?.sha.isNullOrEmpty())) {
+    return if (normScmType != null && baseScmUrl != null && olderGitMetadata != null && !(olderGitMetadata.commitInfo?.sha.isNullOrEmpty())) {
       when {
-        "stash" in commitLink -> {
+        "stash" in normScmType -> {
           "$baseScmUrl/projects/${newerGitMetadata.project}/repos/${newerGitMetadata.repo?.name}/compare/commits?" +
             "targetBranch=${olderGitMetadata.commitInfo?.sha}&sourceBranch=${newerGitMetadata.commitInfo?.sha}"
         }
-        "github" in commitLink -> {
+        "github" in normScmType -> {
           "$baseScmUrl/${newerGitMetadata.project}/${newerGitMetadata.repo?.name}/compare/" +
             "${olderGitMetadata.commitInfo?.sha}...${newerGitMetadata.commitInfo?.sha}"
         }
@@ -51,22 +52,46 @@ class ArtifactVersionLinks(private val scmInfo: ScmInfo, private val cacheFactor
     } else null
   }
 
-  //Calling igor to fetch all base urls by SCM type, and returning the right one based on current commit link
-  fun getScmBaseLink(commitLink: String): String? {
+  fun getNormalizedScmType(commitLink: String): String? {
+    val scmType = getScmType(commitLink)
+    return scmType?.toLowerCase()
+  }
+
+  fun getScmType(commitLink: String): String? {
+    val commitURL = URL(commitLink)
 
     val scmBaseURLs = runBlocking {
       val cachedValue = cache[cacheName].get()
       cachedValue ?: scmInfo.getScmInfo()
     }
-    return when {
-      "stash" in commitLink ->
-        scmBaseURLs["stash"]
-      "github" in commitLink -> {
-        val url = URL(scmBaseURLs["gitHub"])
-        "${url.protocol}://${url.host}"
-      }
-      else ->
-        throw UnsupportedScmType(message = "Stash and GitHub are currently the only supported SCM types.")
+
+    val base = scmBaseURLs.filter { (_,baseUrl) ->
+      commitURL.host == URL(baseUrl).host
     }
+
+    return base.keys.toList().firstOrNull()
+  }
+
+  //Calling igor to fetch all base urls by SCM type, and returning the right one based on current commit link
+  fun getScmBaseLink(commitLink: String): String? {
+    val normScmType = getNormalizedScmType(commitLink)
+    val scmType = getScmType(commitLink)
+    val scmBaseURLs = runBlocking {
+      val cachedValue = cache[cacheName].get()
+      cachedValue ?: scmInfo.getScmInfo()
+    }
+
+    return if (normScmType != null) {
+      when {
+        "stash" in normScmType ->
+          scmBaseURLs["stash"]
+        "github" in normScmType -> {
+          val url = URL(scmBaseURLs[scmType])
+          "${url.protocol}://${url.host}"
+        }
+        else ->
+          throw UnsupportedScmType(message = "Stash and GitHub are currently the only supported SCM types.")
+      }
+    } else null
   }
 }
--- a/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
+++ b/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
@@ -465,6 +465,171 @@ class ComparableLinksTests : JUnit5Minutests {
         }
       }
 
+      context("github enterprise") {
+        before {
+          every { repository.artifactVersions(releaseArtifact, limit) } returns versions.toArtifactVersions(
+            releaseArtifact,
+            "https://git.foo.com"
+          )
+        }
+
+        test("compare links for previous-->current are generated as expected, in the correct env") {
+          // current is compared against what it replaced
+          val summaries = applicationService.getArtifactSummariesFor(application1, limit)
+          expectThat(summaries.first())
+            .withVersionInEnvironment(version2, "test") {
+              state.isEqualTo(CURRENT.name.toLowerCase())
+              compareLink
+                .isEqualTo("https://git.foo.com/spkr/keel/compare/${version1}...${version2}")
+            }
+            .withVersionInEnvironment(version1, "staging") {
+              state.isEqualTo(CURRENT.name.toLowerCase())
+              compareLink.isEqualTo("https://git.foo.com/spkr/keel/compare/${version0}...${version1}")
+            }
+        }
+
+        test("compare links for current --> deploying are generated as expected, in the correct env") {
+          // deploying is compared against the current version
+          val summaries = applicationService.getArtifactSummariesFor(application1, limit)
+          expectThat(summaries.first())
+            .withVersionInEnvironment(version3, "test") {
+              state.isEqualTo(DEPLOYING.name.toLowerCase())
+              compareLink
+                .isEqualTo("https://git.foo.com/spkr/keel/compare/${version2}...${version3}")
+            }
+        }
+
+        test("compare links for previous --> current  are generated as expected, in the correct env") {
+          // previous is compared to what replaced it, which we know.
+          val summaries = applicationService.getArtifactSummariesFor(application1, limit)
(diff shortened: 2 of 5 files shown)


keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
```diff
@@ -287,11 +287,11 @@ class ComparableLinksTests : JUnit5Minutests {
         before {
           every { repository.artifactVersions(releaseArtifact, limit) } returns versions.toArtifactVersions(
             releaseArtifact,
-            "stash"
+            "https://stash"
```
jessesanford, line 290: Someone please review if it was critical that this work with commit urls that are protocol less. I think that was simply an oversight in writing the test.

dbyron-sf: I don't know the code here to say one way or another, but putting this change in a separate commit would help clarify this....and make it easier to see that tests continue to pass before the other changes in the pr.

jessesanford: Good point. I will restructure this.

jessesanford: Now in explicit commit. Still calling attention to this though here so that it gets eyes :eyes:

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
```diff
@@ -35,14 +35,15 @@ class ArtifactVersionLinks(private val scmInfo: ScmInfo, private val cacheFactor
...
-    return if (baseScmUrl != null && olderGitMetadata != null && !(olderGitMetadata.commitInfo?.sha.isNullOrEmpty())) {
+    return if (scmType != null && baseScmUrl != null && olderGitMetadata != null && !(olderGitMetadata.commitInfo?.sha.isNullOrEmpty())) {
       when {
-        "stash" in commitLink -> {
+        "stash" in scmType -> {
           "$baseScmUrl/projects/${newerGitMetadata.project}/repos/${newerGitMetadata.repo?.name}/compare/commits?" +
             "targetBranch=${olderGitMetadata.commitInfo?.sha}&sourceBranch=${newerGitMetadata.commitInfo?.sha}"
         }
-        "github" in commitLink -> {
+        "gitHub" in scmType -> {
```
nimakaviani, line 46: can we make it case insensitive?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
```diff
@@ -51,22 +52,40 @@ class ArtifactVersionLinks(private val scmInfo: ScmInfo, private val cacheFactor
...
-        scmBaseURLs["stash"]
-      "github" in commitLink -> {
-        val url = URL(scmBaseURLs["gitHub"])
-        "${url.protocol}://${url.host}"
+
+    return if (scmType != null) {
+      when {
+        "stash" in scmType ->
+          scmBaseURLs["stash"]
+        "gitHub" in scmType -> {
```
nimakaviani, line 82: same here. problably better if `getScmType` converts the type to lower case to ensure consistency.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
nimakaviani: previously, we didnt parse the link to extract the type so https in the source type didnt matter. better version for this change to go is probably here, where the link to the git artifact gets created: https://github.com/spinnaker/keel/blob/cf94218db7fc757c260dd5d866d478857dae0292/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt#L157

nimakaviani (changes requested):

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
```diff
@@ -51,22 +52,40 @@ class ArtifactVersionLinks(private val scmInfo: ScmInfo, private val cacheFactor
     } else null
   }
 
+  fun getScmType(commitLink: String): String? {
+    val commitURL = URL(commitLink)
```
nabuskey, line 56: Do we need to worry about the potential exception being thrown here?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
```diff
@@ -51,22 +52,40 @@ class ArtifactVersionLinks(private val scmInfo: ScmInfo, private val cacheFactor
...
+    val scmBaseURLs = runBlocking {
+      val cachedValue = cache[cacheName].get()
+      cachedValue ?: scmInfo.getScmInfo()
+    }
+
+    val base = scmBaseURLs.filter { (_,baseUrl) ->
+      commitURL.host == URL(baseUrl).host
+    }
+
+    return base.keys.toList()[0]
```
nabuskey, line 67: This could throw index out of boundary exception. If base is not found after filter, we should return null and log.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
jessesanford: Yes absolutley.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
jessesanford: I would rather perform the normalization of the string closer to the site of the check to avoid ambiguity. Check my latest commits.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
jessesanford: Yes good call. I didn't like that. I fixed this by using getOrNull.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
jessesanford: didn't know about firstOrNull!, but I did use getOrNull let me know if you want me to change it.

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
jessesanford: Are you suggesting that we update the mock generation code to convert "stash" -> "https://stash" rather than just use the string literal  "https://stash" below? What is the real world situation? Are these artifacts going to have protocol less sources? I like modifying the literal because the test reads more explicitly.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
```diff
@@ -51,22 +53,41 @@ class ArtifactVersionLinks(private val scmInfo: ScmInfo, private val cacheFactor
...
+    val scmBaseURLs = runBlocking {
+      val cachedValue = cache[cacheName].get()
+      cachedValue ?: scmInfo.getScmInfo()
+    }
+
+    val base = scmBaseURLs.filter { (_,baseUrl) ->
+      commitURL.host == URL(baseUrl).host
+    }
+
+    return base.keys.toList().getOrNull(0)
```
nimakaviani, line 68: probably slightly easier if `getScmType` return the lowercase type.

```suggestion
    return base.keys.toList().getOrNull(0)?.let { it.lowercase() }
```

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
nabuskey: I was talking about the`java.net.URL` constructor there. It throws `MalformedURLException`. So if given commit link is somehow malformed or empty, the exception will bubble up. `newerGitMetadata?.commitInfo?.link?` will check if it's not null, but not empty / blank string.

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
```diff
@@ -51,22 +53,41 @@ class ArtifactVersionLinks(private val scmInfo: ScmInfo, private val cacheFactor
...
     }
-    return when {
-      "stash" in commitLink ->
-        scmBaseURLs["stash"]
-      "github" in commitLink -> {
-        val url = URL(scmBaseURLs["gitHub"])
-        "${url.protocol}://${url.host}"
+
+    return if (scmType != null) {
+      val scmTypeLower = scmType.lowercase()
```
nimakaviani, line 80:
```suggestion
```

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
nimakaviani: yeah either that or have `toArtifactVersions` and the `git` mock receive another optional parameter for `link`. Problem is, the second parameter for `toArtifactVersion` is the source type, and passing the protocol scheme as part of the source type makes it look weird.

jessesanford: I understand what you mean and I agree, however aren't the original github tests passing protocol in via the second parameter? https://github.com/spinnaker/keel/blob/cf94218db7fc757c260dd5d866d478857dae0292/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt#L620 should we clean all of these up for consistency in a separate PR?

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/artifacts/ArtifactVersionLinks.kt
jessesanford: right but per my earlier comment, the intent is less explicit even if is more dry. A compromise could be renaming getScmType to getNormalizedScmType or getLowerScmType, I like the former in case there is more work to normalize these strings in the future.

nimakaviani: yeah lets go with `getNormalizedScmType`. sounds good to me

keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt
nimakaviani: yep, I agree with the inconsistencies, and it is a good idea to fix it while we are at it.

nimakaviani:
build broken? 🤔

jessesanford:
I think we are good to go now. Your requested change can be marked resolved.

jessesanford:
Ok I removed the call to lowercase() because it's still marked as experimental (*shrug*) and instead am using toLowerCase (which is deprecated) but passes the tests. I also found an issue with the normalization that was causing the lookup to fail for github enterprise tests. So I added getScmType back in and dried up the code. Give a look please.

jessesanford:
Ok looks like builds are passing now. @nimakaviani @nabuskey

nimakaviani (approved):
LGTM

PR #2027, opened 2022-04-05
fix(artifacts): add missing protocol prefixes for mocks

### LINKED ISSUE #2026: Need to add protocol to source in order to create a proper "link"
Note: source may or may not contain a protocol, per the setup of the tests below. We should add a protocol prefix to the string passed in (if it needs it) before calling it a "link"

https://github.com/spinnaker/keel/blob/cf94218db7fc757c260dd5d866d478857dae0292/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/services/ComparableLinksTests.kt#L157

jessesanford:
Fixes #2026 by normalizing all sources to proper "links" with protocol prefixes before they are used by the gitMetadata mock generator.


nimakaviani:
LGTM.

mind fixing the commit msg?

jessesanford:
> mind fixing the commit msg?

Done

PR #2029, opened 2022-04-20

spinnakerbot:


dbyron-sf:
@mergifyio update

dbyron-sf:
@mergifyio update

nabuskey:
From what I have investigated, there are two issues:
1. Docker repository reference is now invalid. It is currently set to use the testcontainer images from a public ECR repository but it seems to not work anymore. I think we can updated it to either point to another ECR or point to Dockerhub. 
2. Spring tests takes a lot of time when ran together. If they are run individually, they work as expected. It seems it is related to DB connection handling during Spring bean refresh in between tests. I haven't had the time to really dig into this part.

PR #1995, opened 2021-09-06
chore(upstream): Upstream changes from Netflix

luispollo:
Upstreaming the latest batch of changes from the Netflix fork.

This is hopefully a much more manageable sized PR to review, but we haven't yet been able to automate PR creation for each internal PR. I'll try to look into that this week.

PR #2067, opened 2023-09-18
chore(upstream): Port upstream changes from Netflix

osoriano:
Port over the first ~60 changes from the [`netflix` branch](https://github.com/luispollo/keel/tree/netflix). Only ~60 were taken to avoid [backward-incompatible changes](https://github.com/luispollo/keel/commit/6344b9c61164830e91283d0e4aa9473eebb379db) and to avoid too many changes.

Works with the latest deck UI (Attached a screenshot).

~⚠️ Please DO NOT squash the commits in this PR when merging!~

The porting of the commits have been [done in the past](https://github.com/spinnaker/keel/pull/1995), but ~600 changes are remaining

Screenshot 2023-09-03 at 8 16 47 PM


keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ClusterDeployStrategy.kt
```diff
@@ -13,6 +13,10 @@ abstract class ClusterDeployStrategy {
 
   companion object {
     val DEFAULT_WAIT_FOR_INSTANCES_UP: Duration = Duration.ofMinutes(30)
+    const val RED_BLACK_STRATEGY = "red-black"
```
jasonmcintosh, line 16: Note, red-black has been.. "deprecated" in favor of standard terms "blue-green"

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceSpec.kt
```diff
@@ -30,4 +29,14 @@ interface ResourceSpec {
...
+
+  /**
+   * Applies the given [suffix] to the resource [id], and to all aggregate properties of the spec
+   * whose names are derived from the [id].
+   *
+   * @return a copy of the original [ResourceSpec] with the modified identifiers.
+   */
+  @JvmDefault
+  fun deepRename(suffix: String): ResourceSpec =
+    throw UnsupportedOperationException("Not implemented")
```
jasonmcintosh, line 41: Intentional unsupported I'd guess for now?

jvz: Kotlin nit: should use `TODO()`

keel-web/src/main/kotlin/com/netflix/spinnaker/config/SlackRequestSecurityConfiguration.kt
```diff
@@ -0,0 +1,29 @@
...
+  override fun configure(http: HttpSecurity) {
+    http
+      .antMatcher("/slack/notifications/callbacks")
+      .authorizeRequests()
+      .anyRequest()
+      .permitAll()
+      .and()
+      .csrf()
+      .ignoringAntMatchers("/slack/notifications/callbacks")
+  }
```
jasonmcintosh, line 28: I THINK this is problematic... as that'd allow non authenticated access to that URL which... would allow any slack notification.  I'm asking around... b/c I may have the order wrong atm...

keel-web/src/main/kotlin/com/netflix/spinnaker/config/SlackRequestSecurityConfiguration.kt
```diff
@@ -0,0 +1,29 @@
...
+@Order(1)
+@EnableWebSecurity
+class SlackRequestSecurityConfiguration : WebSecurityConfigurerAdapter() {
+
+  override fun configure(http: HttpSecurity) {
+    http
+      .antMatcher("/slack/notifications/callbacks")
+      .authorizeRequests()
+      .anyRequest()
+      .permitAll()
```
jvz, line 24: I should note that in Gate, we have a property, `security.webhooks.default-auth-enabled` (default to `false`), which toggles whether this should use `authenticated()` (when `true`) or `permitAll()` (when `false`). Might be useful to add.

osoriano: Good catch, I ended up reverting [this change](https://github.com/luispollo/keel/commit/b56ccbd5b0a91e30f6a24dfbd57609e8024c7cf5#diff-48ff33f739abaec5bfa33e61576ba2997c57035cc5f69aabdd283b98e9eb2f45) since it relied on some security workarounds. That commit could be revisited if we ever decide to port it over

keel-web/src/main/kotlin/com/netflix/spinnaker/config/SlackRequestSecurityConfiguration.kt
osoriano: Good catch, I ended up reverting [this change](https://github.com/luispollo/keel/commit/b56ccbd5b0a91e30f6a24dfbd57609e8024c7cf5#diff-48ff33f739abaec5bfa33e61576ba2997c57035cc5f69aabdd283b98e9eb2f45) since it relied on some security workarounds. That commit could be revisited if we ever decide to port it over

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ResourceSpec.kt
osoriano: Found this `deepRename` call is used for [Preview / PR environments](https://github.com/osoriano/keel/blob/port-netflix-changes/keel-scm/src/main/kotlin/com/netflix/spinnaker/keel/preview/PreviewEnvironmentCodeEventListener.kt#L324). I haven't used this feature yet though

I was tempted to remove the default `UnsupportedOperationException` here, but it could cause plugins to break if it's not implemented, so I moved to `TODO()` :+1:

keel-api/src/main/kotlin/com/netflix/spinnaker/keel/api/ClusterDeployStrategy.kt
osoriano: Seeing lots of existing "red-black" terms in this repo. Going to check if they can be migrated and follow up

jasonmcintosh (approved):

PR #2107, opened 2024-11-13
chore(java): Full java 17 only support

jasonmcintosh:


keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
```diff
@@ -29,18 +29,21 @@ fun configuredObjectMapper(): ObjectMapper = ObjectMapper().configureForKeel()
  * Factory method for [YAMLMapper]s configured how we like 'em.
  */
 fun configuredYamlMapper(): YAMLMapper = YAMLMapper().configureForKeel().disable(USE_NATIVE_TYPE_ID)
-
-fun <T : ObjectMapper> T.configureForKeel(): T =
-  apply {
+fun <T : ObjectMapper> T.configureForKeel(): T {
+  val javaTimeModule = JavaTimeModule()
+  javaTimeModule.addDeserializer(Instant::class.java, PrecisionSqlDeserializer())
```
jasonmcintosh, line 34: Arguably... not sure this custom DESERIALIZER is really needed...

keel-core/src/main/java/com/netflix/spinnaker/keel/serialization/PrecisionSqlSerializer.java
```diff
@@ -0,0 +1,24 @@
+package com.netflix.spinnaker.keel.serialization;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.datatype.jsr310.ser.InstantSerializerBase;
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+
+public class PrecisionSqlSerializer extends InstantSerializerBase<Instant> {
```
dbyron-sf, line 9: Please add some javadoc that explains what this does and why we need it.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -649,7 +650,7 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
     }
 
     context("artifact creation timestamp exists") {
-      val createdAt = Instant.now()
+      val createdAt = Instant.now().truncatedTo(ChronoUnit.MILLIS)
```
dbyron-sf, line 653: Do we still need this with PrecisionSqlSerializer?

jasonmcintosh: Short answer yes.  The test needs it because the test is comparing instant.now to the serialized version.  BUT the databases precision since we're limiting the precision doesn't match as a result post operation.  This truncates it to the same level as what the serialization formatter is doing.  I'll add a comment to this effect

keel-sql/src/test/resources/testcontainers.properties
```diff
@@ -1 +1 @@
-ryuk.container.image = public.ecr.aws/s4w6t4b6/testcontainers/ryuk:0.3.4
+ryuk.container.image = testcontainers/ryuk:0.11.0
```
jasonmcintosh, line 1: FYI the ECR image wasn't working in local testing - thus the need to switch.  PROBABLY should use docker.io/testcontainers/ryuk:0.11.0 but this works

keel-core/src/main/kotlin/com/netflix/spinnaker/keel/serialization/ObjectMappers.kt
```diff
@@ -29,18 +29,20 @@ fun configuredObjectMapper(): ObjectMapper = ObjectMapper().configureForKeel()
  * Factory method for [YAMLMapper]s configured how we like 'em.
  */
 fun configuredYamlMapper(): YAMLMapper = YAMLMapper().configureForKeel().disable(USE_NATIVE_TYPE_ID)
-
-fun <T : ObjectMapper> T.configureForKeel(): T =
-  apply {
+fun <T : ObjectMapper> T.configureForKeel(): T {
+  val javaTimeModule = JavaTimeModule()
+  javaTimeModule.addSerializer(Instant::class.java,PrecisionSqlSerializer())
```
xibz, line 34: Can we add a round trip test that uses the serializer and then deserializes to ensure precision is in fact kept? Other than that, this PR looks good

jasonmcintosh: It's actually tested indirectly via the ArtifactRepositoryTests file  but let me see about a direct test.  It'll only keep as much precision as it writes in the JSON fields.  So this WILL reduce precision on these columns from what the JVM could track - but it's never had that until java 17 it looks like so we're not losing anything - though I may try to do a "6" instead of "3" precision as docs say MySQL CAN handle that level.

jasonmcintosh: Added a test see what you think :)

xibz:
Looks good! One thing Im a little un sure on is the change from 3 to 6. I get it is for mysql, but does this cover everything? Im guessing yes but with sqlite, oracle, microsoft, my hope is this will cover all of those unless we explicitly say that we dont guarantee for those DBs.

xibz (approved):

jasonmcintosh:
SHOULD cover most of them if I recall.  TECHNICALLY Spinnaker supports MySQL & Postgresql :) I wouldn't mind oracle but definitely would be syntax changes.  AND this is really mostly impactful because it's auto generating a column of JSON.  IF You're persisting DIRECT to a datetime field, I THOUGHT the database drivers kick in on reducing the precision with warnings about it - which can have interesting impacts if you're NOT cognizant of the reduction.

A quick search says postgresql has the same precision of 6.  And... as I recalled oracle LOOKS like 9 aka much higher.   I'd BET SQLLite has a higher precision because of how it operates/stores things, but ... not dug as much there and probably very platform specific.

jasonmcintosh:
to note, the problem really is java.time.Instant usage - which... is NOT all that common. Particularly the toString on it - if we were using say OffSet or one of the others... it'd help.  OR we could have this as a standard in the kork serialization stuff.

keel-core/src/main/java/com/netflix/spinnaker/keel/serialization/PrecisionSqlSerializer.java
```diff
@@ -35,7 +35,7 @@ public PrecisionSqlSerializer() {
         Instant::toEpochMilli,
         Instant::getEpochSecond,
         Instant::getNano,
-        new DateTimeFormatterBuilder().appendInstant(3).toFormatter());
+        new DateTimeFormatterBuilder().appendInstant(6).toFormatter());
```
dbyron-sf, line 38: The comment above about using 3 digits could use a tweak.

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
```diff
@@ -649,7 +650,10 @@ abstract class ArtifactRepositoryTests<T : ArtifactRepository> : JUnit5Minutests
     }
 
     context("artifact creation timestamp exists") {
-      val createdAt = Instant.now()
+      // We truncate this since we're using a serialization to java that reduces the level of precision
+      // and later comparisons break otherwise.  This is needed to work with generated columns in
+      // certain databases.  See the PrecisionSqlSerializer class for more info
+      val createdAt = Instant.now().truncatedTo(ChronoUnit.MILLIS)
```
dbyron-sf, line 656: Now that we're using 6 digits, can we change it to ChronoUnit.MICROS?

keel-core/src/main/java/com/netflix/spinnaker/keel/serialization/PrecisionSqlSerializer.java
jasonmcintosh: Was waiting for tests to pass ;)  LOOKS Like they're happy with the 6 precision, though datetime(3) MAY truncate it more.  The question there to my mind is... do we want this to go to 3 or 6?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
jasonmcintosh: Key question I see is do we want 6 or 3?  Database for these columns is 3.  BUT mysql can support 6.  I'd lean towards leaving it 6, and just accept that those specific generated columns only do 3 (IF I'm reading the spec on the `datetime(3)` correctly

keel-core/src/main/java/com/netflix/spinnaker/keel/serialization/PrecisionSqlSerializer.java
dbyron-sf: What did we have with java 11?

keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ArtifactRepositoryTests.kt
jasonmcintosh: And I'd be good with MICROS :)

keel-core/src/main/java/com/netflix/spinnaker/keel/serialization/PrecisionSqlSerializer.java
xibz: Was the point for this to just allow for more precision? If that's the case, we may want to just leave it at 3. Like I said that was the only thing I was unsure of, but more precision also does not hurt, but Spinnaker also doesn't need ot be the most precise service either

jasonmcintosh: Yeah mostly trying to match the most precision to prevent any goofy collision issues we MIGHT hit.  I was originally "leave it at 3" but... if 6 works, AND we're consistent that's really the bigger key in my mind.

xibz: Yea, agreed. I think leaving it at 6 is fine.

dbyron-sf: Maybe I'm getting tripped up in the wording, but isn't there a difference between what java 11 supports, and what keel actually used?  I'd prefer to keep it the same as what we had if possible.

jasonmcintosh: SO not quite.  Java 11 supported and delivered 6 digits of precision at MOST.  Java 17 ALLOWED a higher level of precision if the system clock supported it.  Instant.toString was never consistent on the output - it would just output the MOST available.  This led to situations where when doing serialization to JSON, you'd get inconsistent results.  And on an OS with HIGHER precision, you'd go over the bounds that the database supported.  SO the primary impact and root of the problem is when Keel does auto generated fields OFF of the JSON string in date format, causing a failure.  Java was always able to handle it, but databases didn't handle the higher precision UNLESS they were storing things as a string (which most are).  Timestamp fields however would NOT work.

SO the key on this is to serialize at MAX precision for the storage layer.  NOTE:  There are POTENTIAL headaches on tests in other places if we do comparisons of what's written & read on unit tests in a real database.  SQLLite being Java MAY support a higher preceision - but I'd bet it depends on compatibility mode.  SO if we have a test that writes the timestamps using a mysql test container, then tries to do an equals on a read of that timestamp it will NOT work as mysql wouldn't support the full precision that the JVM supports.

jasonmcintosh: To note... Orca uses LONG's for persistence of timestamps - so it worked without issue.  It's JUST keel that stores times as Instants and in string format.  Least so far that I found.  That was the original headache with a simple replacement serialization implementation - I had to switch to the time base serializer which is smarter on it's handling.

dbyron-sf: OK, sounds like we're going with 6.  I guess we don't include keel in our release notes, so we're off the hook there :)

dbyron-sf (approved):

